PR #58015, opened 2023-10-31
nix: upgrade node and pnpm

burmudar:
Upgrade nodejs and pnpm to match what the rest of the repo is using.

We don't fetch typescript from npm registry since 5.2.2 has a missing `package-lock.json`. Fortunately `nixpkgs-unstable` has 5.2.2.
## Test plan
Tested locally
```
bash-5.2$ pnpm --version
8.9.2
bash-5.2$ node --version
v20.8.0
bash-5.2$ tsc --version
Version 5.2.2
bash-5.2$ exit

```


Strum355 (commented):
Is this tested on macos?

Strum355 (approved):
tested on linux too, all working :+1:

PR #58022, opened 2023-10-31
search jobs: fix query highlighting

stefanhengl:
`SyntaxHighlightedSearchQuery` does not render regexp ranges, such as `[a-z]`, correctly. 

input query: `context:global ghp_[A-Za-z0-9]{36} r:.* rev:*refs/heads/* patterntype:regexp`

Search jobs page:

image

I couldn't find a simple fix so I decided to replace it with a read-only CodeMirror editor.

I want to get this into the upcoming patch. However, we use `SyntaxHighlightedSearchQuery` in a lot of places and it seems risky to replace the component without baking it a bit on s2. Hence I only fix this bug for Search Jobs.

## Test plan
- This is how it looks locally after the fix:
image

### DIFF
--- a/client/shared/src/search/query/scanner.ts
+++ b/client/shared/src/search/query/scanner.ts
@@ -511,7 +511,7 @@ const scanStandard = (query: string): ScanResult<Token[]> => {
     return scan(query, 0)
 }
 
-function detectPatternType(query: string): SearchPatternType | undefined {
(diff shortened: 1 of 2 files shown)


fkling (approved):
Accepting to unblock (I'm happy the new `BaseCodeMirrorQueryInput` is of use), but I think we should really figure out how to fix that in `SyntaxHighlightedSearchQuery`.

PR #58023, opened 2023-10-31
Vscode extension: Delete vscode extension package 

vovakulikov:
This PR removes the whole vscode integration package from our codebase, it removes all readme, CI jobs, ...etc. It's been a while since we did anything about this part of our codebase, but it still require some attention when we do something about search result page. This PR is first part of change, later we in follow up we would be able to move search result UI from shared and simplify its API. 

## Test plan
- Check that CI is green


jasonhawkharris (approved):
LGTM! Do we need a change log entry?

PR #58025, opened 2023-10-31
Code Navigation: optimize repo link picker

jasonhawkharris:
Before, the RepoLinkPicker was initializing the first set of results by running a query with a value of `""`. While this doesn't affect small instances much, large instances, like dotcom, are affected by very slow initial loading times.  

Now, it's faster. 

This also fixes a bug where users without site-admin privileges were unable to use the feature due to getting a bunch of errors when clicking the dropdown instead of repo names. 
  
## Test plan
- Tested performance on local dev environment - **Negligible/No observable improvement**
- Tested performance on `k8s.sgdev.org` - **Negligible improvement**
- Tested performance on `dotcom` - **Markedly faster than before**
- Add unit test

### DIFF
--- a/client/web/src/repo/RepoLinkPicker.tsx
+++ b/client/web/src/repo/RepoLinkPicker.tsx
@@ -34,8 +34,8 @@ import {
 
 import { type RepositoriesSuggestionsResult, type RepositoriesSuggestionsVariables } from '../graphql-operations'
 
-import { CodeHostTypes } from './constants'
-import { getInitialRepoResults } from './utils'
+import { CodeHostType } from './constants'
+import { getInitialSearchTerm, stringToCodeHostType } from './utils'
 
 import styles from './RepoLinkPicker.module.scss'
 
@@ -46,6 +46,8 @@ const REPOSITORIES_QUERY = gql`
                 id
                 name
                 url
+                # We use externalRepository over externalServices here
+                # because externalServices is accessible only to admins.
                 externalRepository {
                     id
                     serviceType
@@ -83,7 +85,7 @@ export const RepoLinkPicker: FC<RepoLinkPickerProps> = props => {
     } = useQuery<RepositoriesSuggestionsResult, RepositoriesSuggestionsVariables>(REPOSITORIES_QUERY, {
         skip: !isSuggestionOpen,
         variables: {
-            query: searchTerm.length === 0 ? getInitialRepoResults(repositoryName) : debouncedSearchTerm,
+            query: searchTerm.length === 0 ? getInitialSearchTerm(repositoryName) : debouncedSearchTerm,
         },
         fetchPolicy: 'cache-first',
     })
@@ -91,7 +93,6 @@ export const RepoLinkPicker: FC<RepoLinkPickerProps> = props => {
     const handleSelect = (selectedValue: string): void => {
         navigate(`/${selectedValue}`)
         setSuggestionOpen(false)
-        setSearchTerm('')
     }
 
     const data = currentData ?? previousData
@@ -146,7 +147,9 @@ export const RepoLinkPicker: FC<RepoLinkPickerProps> = props => {
                                         className={styles.item}
                                     >
                                         <Icon
-                                            svgPath={getCodeHostIconPath(suggestion.externalRepository.serviceType)}
+                                            svgPath={getCodeHostIconPath(
+                                                stringToCodeHostType(suggestion.externalRepository.serviceType)
+                                            )}
                                             height="1rem"
                                             width="1rem"
                                             inline={false}
@@ -165,27 +168,27 @@ export const RepoLinkPicker: FC<RepoLinkPickerProps> = props => {
     )
 }
 
-export const getCodeHostIconPath = (codeHostType?: String): string => {
-    switch (codeHostType) {
-        case CodeHostTypes.GITHUB: {
+export const getCodeHostIconPath = (codehost: CodeHostType): string => {
+    switch (codehost) {
+        case CodeHostType.GITHUB: {
             return mdiGithub
         }
-        case CodeHostTypes.BITBUCKETCLOUD: {
+        case CodeHostType.BITBUCKETCLOUD: {
             return mdiBitbucket
         }
-        case CodeHostTypes.BITBUCKETSERVER: {
+        case CodeHostType.BITBUCKETSERVER: {
             return mdiBitbucket
         }
-        case CodeHostTypes.GITLAB: {
+        case CodeHostType.GITLAB: {
             return mdiGitlab
         }
-        case CodeHostTypes.GITOLITE: {
+        case CodeHostType.GITOLITE: {
             return mdiGit
         }
-        case CodeHostTypes.AWSCODECOMMIT: {
+        case CodeHostType.AWSCODECOMMIT: {
             return mdiAws
         }
-        case CodeHostTypes.AZUREDEVOPS: {
+        case CodeHostType.AZUREDEVOPS: {
             return mdiMicrosoftAzure
         }
         default: {
--- a/client/web/src/repo/RepoLinkPicker.tsx
+++ b/client/web/src/repo/RepoLinkPicker.tsx
@@ -88,14 +88,12 @@ export const RepoLinkPicker: FC<RepoLinkPickerProps> = props => {
         variables: {
             query: searchTerm.length === 0 ? getInitialSearchTerm(repositoryName) : debouncedSearchTerm,
         },
-        fetchPolicy: 'cache-and-network',
+        fetchPolicy: 'cache-first',
     })
 
     const handleSelect = (selectedValue: string): void => {
         navigate(`/${selectedValue}`)
         setSuggestionOpen(false)
-        // set search term back to an empty string, but replace
-        // new initial value for the gql request
         setSearchTerm('')
     }
 
--- a/client/web/src/repo/RepoLinkPicker.tsx
+++ b/client/web/src/repo/RepoLinkPicker.tsx
@@ -38,6 +38,8 @@ import {
     type RepositoriesSuggestionsVariables,
 } from '../graphql-operations'
 
+import { getInitialSearchTerm } from './utils'
+
 import styles from './RepoLinkPicker.module.scss'
 
 const REPOSITORIES_QUERY = gql`
@@ -195,9 +197,4 @@ export const getCodeHostIconPath = (codeHostType?: ExternalServiceKind): string 
             return mdiSourceRepository
         }
     }
-}
-
-const getInitialSearchTerm = (repo: string) => {
-    const r = repo.split('/')
-    return r[r.length - 1]
 }
--- a/client/web/src/repo/RepoLinkPicker.tsx
+++ b/client/web/src/repo/RepoLinkPicker.tsx
@@ -85,13 +85,18 @@ export const RepoLinkPicker: FC<RepoLinkPickerProps> = props => {
         loading,
     } = useQuery<RepositoriesSuggestionsResult, RepositoriesSuggestionsVariables>(REPOSITORIES_QUERY, {
         skip: !isSuggestionOpen,
-        variables: { query: debouncedSearchTerm },
+        variables: {
+            query: searchTerm.length === 0 ? getInitialSearchTerm(repositoryName) : debouncedSearchTerm,
+        },
         fetchPolicy: 'cache-and-network',
     })
 
     const handleSelect = (selectedValue: string): void => {
         navigate(`/${selectedValue}`)
         setSuggestionOpen(false)
+        // set search term back to an empty string, but replace
+        // new initial value for the gql request
+        setSearchTerm('')
     }
 
     const data = currentData ?? previousData
@@ -192,4 +197,9 @@ export const getCodeHostIconPath = (codeHostType?: ExternalServiceKind): string 
             return mdiSourceRepository
         }
     }
+}
+
+const getInitialSearchTerm = (repo: string) => {
+    const r = repo.split('/')
+    return r[r.length - 1]
 }
--- a/client/web/src/repo/RepoLinkPicker.tsx
+++ b/client/web/src/repo/RepoLinkPicker.tsx
@@ -32,13 +32,10 @@ import {
     useDebounce,
 } from '@sourcegraph/wildcard'
 
-import {
-    ExternalServiceKind,
-    type RepositoriesSuggestionsResult,
-    type RepositoriesSuggestionsVariables,
-} from '../graphql-operations'
+import { type RepositoriesSuggestionsResult, type RepositoriesSuggestionsVariables } from '../graphql-operations'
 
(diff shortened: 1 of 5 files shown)


jasonhawkharris:
@camdencheek, @vovakulikov: Before I move on to adding some tests, I'd like to get your thoughts on my approach here.
1. Create a new function that takes the repo name (passed into props and so always present) and returns a derived initial search term.
2. If the value of `searchTerm` is an empty string, pass the return value of `getInitialSearchTerm()` into the `$query` value of the `gql` call. This avoids using an empty string as the query while also, hopefully, giving somewhat relevant initial suggestions to the user. 
3. `searchTerm` remains an empty string even after the `getInitialSearchTerm` function is called, so as not to pre-populate the search input field.
4. When searched and selected, we reset the `$query` value and the search term (back to `""`) so that the input field is clear if the user wishes to search again. 

Do either of you have any thoughts on my approach here?

camdencheek:
Nice find! Approach looks solid to me. There's a separate question of "why is the backend so slow with an empty query string" that we should make a followup issue for. 

Probably the biggest question mark for me is what to populate the initial list with. I think there are a few options:
1) A search for the repo name, like you have. Downside is this is pretty likely to be unique, so there is often going to only be a single entry.
2) A search for the repo org (e.g. `github.com/sourcegraph/`). Downside is this list might still be pretty large, though it might practically be small enough. 
3) Nothing. Just don't suggest until the user starts typing. This probably isn't great because it wouldn't be clear to the user what they're filtering from.

I'd personally lean towards (2), but don't really feel strongly.

client/web/src/repo/RepoLinkPicker.tsx
```diff
@@ -85,13 +85,19 @@ export const RepoLinkPicker: FC<RepoLinkPickerProps> = props => {
...
         fetchPolicy: 'cache-and-network',
     })
 
     const handleSelect = (selectedValue: string): void => {
         navigate(`/${selectedValue}`)
         setSuggestionOpen(false)
+        // set search term back to an empty string, but replace
+        // new initial value for the gql request
+        setSearchTerm('')
+        getInitialSearchTerm(selectedValue)
```
vovakulikov, line 100: Not sure I understand this line, based on this function body it just return the last segment split by `/` symbol, but we don't set it or use it in anyway. Why do we need this?

client/web/src/repo/RepoLinkPicker.tsx
```diff
@@ -85,13 +85,19 @@ export const RepoLinkPicker: FC<RepoLinkPickerProps> = props => {
         loading,
     } = useQuery<RepositoriesSuggestionsResult, RepositoriesSuggestionsVariables>(REPOSITORIES_QUERY, {
         skip: !isSuggestionOpen,
-        variables: { query: debouncedSearchTerm },
+        variables: {
+            query: searchTerm.length === 0 ? getInitialSearchTerm(repositoryName) : debouncedSearchTerm,
+        },
         fetchPolicy: 'cache-and-network',
```
vovakulikov, line 91: I actually think that I made a mistake here, we probably could go with just cache-first strategy, I doubt that set of repositories will be changed between picker searches. Even if so it should be very very rare situation.

vovakulikov (commented):
Left a few comments, 😉 thanks for catching and fixing this thing up.

vovakulikov:
@jasonhawkharris besides perfomance problem you also noticed that this picker doesn't work with non admin users, this is because query for fetching repositories list tries to read external service info which is admin only. That's why we have api error if user isn't an admin. Thanks to @camdencheek advice I think we can use just externalRepository field instead of externalServices to read kind field (in externalRepository it's called serviceType I think) and use this info for rendering a code host icon. 

This is just for the context, I can fix it in a separate PR. 😉

client/web/src/repo/RepoLinkPicker.tsx
jasonhawkharris: Whoops! you're right! I'll remove it.

jasonhawkharris:
@camdencheek Would another approach be to create a new query that would fetch the users most recently visited repos? Do we have access to that information within the API?

camdencheek:
That would be really nice, but we don't currently have that info easily accessible.

client/web/src/repo/RepoLinkPicker.tsx
jasonhawkharris: Changed it for ya. 🥇

client/web/src/repo/RepoLinkPicker.tsx
```diff
@@ -165,27 +165,27 @@ export const RepoLinkPicker: FC<RepoLinkPickerProps> = props => {
     )
 }
 
-export const getCodeHostIconPath = (codeHostType?: ExternalServiceKind): string => {
+export const getCodeHostIconPath = (codeHostType?: String): string => {
```
camdencheek, line 168: Should we make an enum type for this string?

client/web/src/repo/utils.tsx
```diff
@@ -12,3 +12,8 @@ export const getCanonicalURL = (sourceType: RepositoryType | string, node: GitCo
     isPerforceChangelistMappingEnabled() && isPerforceDepotSource(sourceType) && node.perforceChangelist
         ? node.perforceChangelist.canonicalURL
         : node.canonicalURL
+
+export const getInitialSearchTerm = (repo: string) => {
+    const r = repo.split('/')
+    return r[r.length - 1]
```
camdencheek, line 18: javascript arrays can be negative indexed, right?

```suggestion
    return r[-1]
```

camdencheek (approved):

client/web/src/repo/utils.tsx
vovakulikov: @camdencheek No, you have to control range manually that it is within 0 and slice length - 1 range. You can use negative indexes when you work with operators though (like slice or splice)

client/web/src/repo/RepoLinkPicker.tsx
jasonhawkharris: Can you elaborate?

client/web/src/repo/utils.tsx
camdencheek: aah, gotcha. TIL, thanks!

jasonhawkharris: This is possible in TypeScript, but it becomes JavaScript incompatible if you take this route. I think it's safer to return `r[r.length - 1]` for that reason.

client/web/src/repo/RepoLinkPicker.tsx
camdencheek: I think it would be possible to type this more strongly as an enum so that we can guarantee every case is handled. However, we just get a string from the API, so we'd need to have a function that translates from a string at the boundary to the more strongly-typed enum. Might not be worth it

jasonhawkharris: But now I'm confused as to why the function signature doesn't need to change in order for this to work. If I leave it as string, and convert the group of constants to an enum object, it seems to work as intended. Is there any advantage to this?

camdencheek: > why the function signature doesn't need to change in order for this to work

It works because every member of the enum _is_ a string. The function signature accepts a more general type than what you're passing in.

> Is there any advantage to this?

The advantage comes when you update the function signature to only accept `CodeHostType` rather than `String`. Because right now, looking at the function signature, I could pass in any string. Maybe somewhere else, the github code host is `GitHub`, which would be accepted by the current function signature, but would be caught at compile time with a function signature that takes an enum. 

Additionally, Typescript will guarantee that a switch statement is exhaustive. So if we add support for, say, `gerrit`, when we update the type to include `gerrit`, Typescript will complain that the switch statement doesn't handle that enum value.

client/web/src/repo/utils.tsx
```diff
@@ -12,3 +12,8 @@ export const getCanonicalURL = (sourceType: RepositoryType | string, node: GitCo
     isPerforceChangelistMappingEnabled() && isPerforceDepotSource(sourceType) && node.perforceChangelist
         ? node.perforceChangelist.canonicalURL
         : node.canonicalURL
+
+export const getInitialRepoResults = (repo: string) => {
```
fkling, line 16: I find the name of the function confusing compared to what it does. Also have you looked whether we already have a utility function that gets the repo name?

client/web/src/repo/RepoLinkPicker.tsx
```diff
@@ -47,11 +46,9 @@ const REPOSITORIES_QUERY = gql`
...
                 name
                 url
-                externalServices(first: 1) {
-                    nodes {
-                        id
-                        kind
-                    }
+                externalRepository {
+                    id
+                    serviceType
```
fkling, line 51: What's the difference between those two fields?

jasonhawkharris: externalServices can only be accessed by site-admins. externalRepository is public. Before, non-site-admins were completely blocked from using this feature because it just returned a bunch of 403's in the dropdown.

client/web/src/repo/utils.tsx
jasonhawkharris: I'll see if there's a utility function already written. 

Good call on the name. I was trying to be too clever and ended up being incoherent. I'll change it.

client/web/src/repo/RepoLinkPicker.tsx
jasonhawkharris: Got it. I've implemented it, but I'm not sure if I've done it correctly. Can you take a look?

client/web/src/repo/utils.tsx
jasonhawkharris: No good utility for this. Every utility like it are methods on a specific class. I think this is good for our purposes here. Marking as resolved.

client/web/src/repo/RepoLinkPicker.tsx
jasonhawkharris: Will do.

client/web/src/repo/utils.tsx
```diff
@@ -12,3 +14,27 @@ export const getCanonicalURL = (sourceType: RepositoryType | string, node: GitCo
...
+            return CodeHostTypes.GITHUB
+        case 'gitlab':
+            return CodeHostTypes.GITLAB
+        case 'bitbucketCloud':
+            return CodeHostTypes.BITBUCKETCLOUD
+        case 'gitolite':
+            return CodeHostTypes.GITOLITE
+        case 'awsCodeCommit':
+            return CodeHostTypes.AWSCODECOMMIT
+        case 'asureDevOps':
```
camdencheek, line 35:
```suggestion
        case 'azureDevOps':
```

client/web/src/repo/utils.tsx
```diff
@@ -12,3 +14,27 @@ export const getCanonicalURL = (sourceType: RepositoryType | string, node: GitCo
...
+            return CodeHostTypes.AZUREDEVOPS
+        default:
+            throw new Error(`Invalid code host type: ${codeHostType}`)
```
camdencheek, line 38: Rather than throwing here, I think it's probably preferable to have a `CodeHostTypes.OTHER`. We can't control what the API spits back, so we don't want to error out the webapp if the backend decides to return a new code host.

client/web/src/repo/RepoLinkPicker.tsx
```diff
@@ -165,27 +167,27 @@ export const RepoLinkPicker: FC<RepoLinkPickerProps> = props => {
     )
 }
 
-export const getCodeHostIconPath = (codeHostType?: ExternalServiceKind): string => {
-    switch (codeHostType) {
-        case ExternalServiceKind.GITHUB: {
+export const getCodeHostIconPath = (codeHostType: string): string => {
```
camdencheek, line 170: Since we went through the effort of making a type for this, we should push that type as far upwards as we can to get the most leverage out of it. The goal is to constrain the inputs as much as possible so we can catch errors at compile time rather than runtime.

client/web/src/repo/constants.ts
```diff
@@ -2,3 +2,13 @@ export const LogsPageTabs = {
     COMMANDS: 0,
     SYNCLOGS: 1,
 } as const
+
+export enum CodeHostTypes {
```
camdencheek, line 6: nit: the type only represents a single code host type, so I think it makes sense for this to be singular

```suggestion
export enum CodeHostType {
```

client/web/src/repo/RepoLinkPicker.tsx
```diff
@@ -85,13 +84,16 @@ export const RepoLinkPicker: FC<RepoLinkPickerProps> = props => {
         loading,
     } = useQuery<RepositoriesSuggestionsResult, RepositoriesSuggestionsVariables>(REPOSITORIES_QUERY, {
         skip: !isSuggestionOpen,
-        variables: { query: debouncedSearchTerm },
-        fetchPolicy: 'cache-and-network',
+        variables: {
+            query: searchTerm.length === 0 ? getInitialSearchTerm(repositoryName) : debouncedSearchTerm,
```
vovakulikov, line 88: I don't think I like this logic from ux point of view. That means that user can't really understand based on what they have suggestion results (is it their search term or it's based on repo name. 

For example, if I'm looking at some specific repo its name won't give us much suggestions, 
it would make me think that I don't have any repositories on the instance or they are not available to me 

Screenshot 2023-11-01 at 15 01 54

@jasonhawkharris my suggestion here is still using repository name as initial search term value but make it more explicit and put this value in the search input that user could see why they suggestions have only these items they see. 

So it would be like this 
Screenshot 2023-11-01 at 15 05 39

vovakulikov (commented):
Left minor tech suggestions and one UX suggestion which I think important to do (or at least to consider), anyway thanks for fixing my mistakes @jasonhawkharris

client/web/src/repo/RepoLinkPicker.tsx
vovakulikov: Change for this is simple I think 

```ts
    const [searchTerm, setSearchTerm] = useState<string>(getInitialSearchTerm(repositoryName))
    const debouncedSearchTerm = useDebounce(searchTerm, 500)

    const {
        data: currentData,
        previousData,
        error,
        loading,
    } = useQuery<RepositoriesSuggestionsResult, RepositoriesSuggestionsVariables>(REPOSITORIES_QUERY, {
        skip: !isSuggestionOpen,
        variables: { query: debouncedSearchTerm },
        fetchPolicy: 'cache-first',
    })

    const handleSelect = (selectedValue: string): void => {
        navigate(`/${selectedValue}`)
        setSuggestionOpen(false)
        setSearchTerm(selectedValue)
    }
```

client/web/src/repo/utils.tsx
BolajiOlajide: There's `Array.at` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at) but that might return undefined if the array is empty.

Issue #57974, opened 2023-10-30
Code Intel is broken on file extensions that don't match case


RichardBelzer:
- **Sourcegraph version:** 5.2.1
- **Platform information:** Fedora Server

#### Steps to reproduce:

1. Have a file in the repo that is all caps. Example could be `MYCODE.CPP`.

#### Expected behavior:

Code intel works. When moving my mouse over a function call, I would get a popup that lets me either go to definition or see references.

#### Actual behavior:

Code intel doesn't function on the file. Looking in the browser logs, there is an error that states "unrecognized file extension"

![image](https://github.com/sourcegraph/sourcegraph/assets/69686380/b72abad7-a93c-4370-a907-7ba0fcd61035)

This is similar to what @slimsag worked in https://github.com/sourcegraph/sourcegraph/issues/13007. We fixed syntax highlighting a couple years ago, but at some point between `3.2.3` and the current version `5.2.1` we introduced a regression where code intel no longer works on those kind of files.

RichardBelzer:
Did some research and discovered that this issue occurs starting in `5.2.0`. It works fine in `5.1.9`.

It was triggered by the change in https://github.com/sourcegraph/sourcegraph/issues/51547 by @ggilmore

As a temporary workaround, you can disable gRPC by adding:

```
"experimentalFeatures": { "enableGRPC": false }
```

To the site configuration which will fix this problem in `5.2.0+`. Hopefully this issue gets fixed soon, or at least while the option to disable gRPC is still available.

Please allow users to continue to disable gRPC until this issue is fixed, thanks!

jdorfman:
Thanks @RichardBelzer! I notified @ggilmore

varungandhi-src:
Thanks for reporting the issue.

Reproduced: https://sourcegraph.com/github.com/electronicarts/CnC_Remastered_Collection/-/blob/REDALERT/MP.CPP

<img width="895" alt="CleanShot 2023-11-01 at 09 52 49@2x" src="https://github.com/sourcegraph/sourcegraph/assets/93103176/19c7cb8b-d7cc-4d1a-9ba6-f895711dd613">

varungandhi-src:
There are two different bugs here.

1. The squirrel code path doesn't handle `.CPP` as an extension. https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/cmd/symbols/squirrel/language-file-extensions.json?L45-57
[...]

PR #58030, opened 2023-11-01
fix: Check for known Squirrel errors inside GraphQL responses

varungandhi-src:
Check for known Squirrel errors in frontend emitted in gRPC code path and handle
them gracefully instead of panicking. Fixes https://github.com/sourcegraph/sourcegraph/issues/57974

There was a separate issue, where Squirrel does a case-sensitive comparison
of file extensions (alternatively, we can just list the uppercase file extensions
for C/C++), which I'll fix in a subsequent PR.

Question:
- [x] Technically, the use of an error in gRPC mode is a a breaking change, because
    the old code path would return an object with `null` for `SymbolInfo` and `"null"`
    for LocalCodeIntel inside a more deeply nested object.
    So the same GraphQL API doesn't return the exact same object with/without gRPC.
    Even though the newer behavior is arguably more correct (returning an error),
    is it OK to land it as-is given that it is technically a breaking change (which already
    shipped in 5.2.0)?

## Test plan

With this patch, we're able to successfully see search-based code intel for files
with uppercase extensions like `.H` and `.CPP` even with gRPC turned on.

CleanShot 2023-11-02 at 14 11 08@2x

Q: Is there a better way to test this?

### DIFF
--- a/cmd/symbols/internal/api/handler_cgo.go
+++ b/cmd/symbols/internal/api/handler_cgo.go
@@ -4,7 +4,6 @@ package api
 
 import (
 	"context"
-	"fmt"
 	"net/http"
 
 	"github.com/sourcegraph/sourcegraph/cmd/symbols/squirrel"
@@ -88,8 +87,6 @@ func (s *grpcService) SymbolInfo(ctx context.Context, request *proto.SymbolInfoR
 
 	args.RepoCommitPath = request.GetRepoCommitPath().ToInternal()
 	args.Point = request.GetPoint().ToInternal()
-
-	fmt.Println("gRPC calling Squirrel.SymbolInfo")
 
 	info, err := squirrelService.SymbolInfo(ctx, args)
 	if err != nil {
--- a/cmd/symbols/internal/api/handler_cgo.go
+++ b/cmd/symbols/internal/api/handler_cgo.go
@@ -27,6 +27,16 @@ func addHandlers(
 	mux.HandleFunc("/symbolInfo", squirrel.NewSymbolInfoHandler(searchFunc, readFileFunc))
 }
 
+func convertSquirrelErrorToGrpcError(err error) error {
+	if errors.Is(err, squirrel.UnrecognizedFileExtensionError) {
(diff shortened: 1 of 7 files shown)


olafurpg (approved):

cmd/symbols/internal/api/handler_cgo.go
```diff
@@ -26,6 +26,16 @@ func addHandlers(
 	mux.HandleFunc("/symbolInfo", squirrel.NewSymbolInfoHandler(searchFunc, readFileFunc))
 }
 
+func convertSquirrelErrorToGrpcError(err error) error {
```
ggilmore, line 29: Your intent might be clearer (this function attempts to lift a normal error into a gRPC status it can) if you returned a [status.Status](https://pkg.go.dev/google.golang.org/grpc/status) instead of an opaque error. 

Ex:
func convertSquirrelErrorToGrpcStatus(err error) *status.Status {
   if ... {
      return status.New(codes.InvalidArgument, err.Error())
   }
  

   return nil
}

...


payyload, err := squirrelService.LocalCodeIntel(ctx, args)
if err != nil {
	if s := convertSquirrelErrorToGrpcStatus(err); s != nil {
		return s.Err()
	}

   ... 
}

ggilmore (approved):

ggilmore:
>Technically, the use of an error in gRPC mode is a a breaking change, because
the old code path would return an object with null for SymbolInfo and "null"
for LocalCodeIntel inside a more deeply nested object.
So the same GraphQL API doesn't return the exact same object with/without gRPC.
Even though the newer behavior is arguably more correct (returning an error),
is it OK to land it as-is given that it is technically a breaking change (which already
shipped in 5.2.0)?

(I don't own this area of the product), but in my mind stability with the web app's behavior is more important than stability with our graphQL API. I think it's fine to ship this as is.

cmd/symbols/internal/api/handler_cgo.go
varungandhi-src: Fixed.

PR #58034, opened 2023-11-01
do not show "repo last fetched" if it is recent

sqs:
This is the icon in the top right of each search result container that shows the last time the repo was fetched.
    
Now, it's only shown if the last-fetched date is 1 day or more in the past. If it is more recent than 1 day, the icon is not shown. This reduces visual noise and simplifies the UI.

image

## Test plan

CI


fkling (approved):

camdencheek (approved):

camdencheek:
cc @rrhyne

PR #58035, opened 2023-11-01
remove "Copy full query" button from search box

sqs:
This button does not makes sense:

- Select-all in the new query input will select the `context:foo ` prefix, which was one (barely plausible) possible reason people might have wanted to use the copy query button.
- The URL is a better thing to copy and share than the query anyway.

Removing it frees up space and simplifies the UI.

image

## Test plan

CI


peterguy (approved):
Legit change. Agreed that the URL is (nearly?) always the preferred thing to copy.

Perhaps consider as a followup to add a hook to the Copy event in the query input so that if all of the query input is selected, add to it the pattern type and case setting. That would preserve the functionality of the copy query button without the UI clutter.

PR #58042, opened 2023-11-01
blob: Refactor blame decorations 

fkling:
This commit refactors blame decorations to be more configuration driven. Specific changes and fixes:

- Fixed "lock first line" behavior. It think this was because the lock line behavior was triggered multiple times due to data loading. Separating showing the column and showing the data seems to fix this.
- Wrong (no) blame decoration shown after a folded range. This was an interesting issue. The root issue was that the blame decoration component would only render something when it was shown in its start line. But when a range is folded that start line might be hidden. Removing this condition together with other changes fixes this issue.
- Previously every line was rendering a blame decoration component, but for most lines the component wouldn't display anything. Even without displaying anything, creating a react root per line contributed to unnecessary overhead.
  The implementation was changed to only render the a decoration in the first (visible) line of the hunk. In order to still show the recency indicator across all
lines it is moved into its own gutter and rendered directly via the DOM.
- Recency colors have been refactored to avoid having to use `useIsLightTheme`.
- `BlameColumn` wasn't used anymore and could be removed which also allowed changes to the `BlameDecoration` component itself.
- Instead of just highlighting the line when hovering over the blame information the whole hunk is highlighted. It should be possible to highlight the whole hunk when hovering anywhere in the blame column, but due to how we implement it (widget decorations) it would require some more work.

Additional notes:
- Using React for this, even with fewer occurrences, still has some performance penalty. This becomes apparent when scrolling in a larger document with blame enabled, and hovering over blame information. The more often this is done the slower the blame decorations seem to react (ha!) to hover events (to be fair, I assume it's due to React, it could also be for other reasons)
- Scenario: `Hunk 1/Commit A | Hunk 2/Commit B | Hunk 3/Commit A`
  If code is folded so that Hunk 2 is completely hidden, it would be nice if we only displayed `Hunk 1/Commit A` instead of `Hunk 1/Commit A | Hunk 3/Commit A`. That's actually relatively easy to do but getting the new "highlight hunk lines" to work properly then is more difficult.

## Test plan

Manual testing: 

https://github.com/sourcegraph/sourcegraph/assets/179026/6355d2fb-027b-45f0-863d-965968911802


client/web/src/repo/blob/blameRecency.ts
fkling, line 1: Most of this is copied from `blameRecency.tsx`.

client/shared/src/components/CodeMirrorEditor.tsx
fkling, line 1: Optional effects was added to `useCompartment` to support the 'lock first line' use case.

taras-yemets (approved):
Great work, @fkling! Thank you! 

On the demo video (sec 20-21) the version on the right (with current PR changes) has slightly misaligned background colors on hover. Is it fixed in the final version?

fkling:
@taras-yemets That's also an issue in the current version. It happens when selecting an empty line and is somehow related to us using a layer to mark selected lines. I haven't figured out how to fix that yet.

camdencheek (approved):
I at least read through this PR to try to get a feel for things, but I don't really know enough to provide any valuable feedback. Looks good though 😄

client/web/src/repo/blob/blameRecency.ts
```diff
@@ -0,0 +1,42 @@
...
+const THREE_YEARS_AGO = subYears(Date.now(), 3).getTime()
+
+export function getBlameRecencyColor(commit: Date | undefined, firstCommitDate: Date | undefined): string {
+    if (!commit) {
+        return COLORS[0]
+    }
+
+    // We create a recency range depending on the repo creation date. If the
+    // repo is newer than a year, we use the last year so that we don't have a
+    // scale that is too sensible.
```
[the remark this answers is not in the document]
fkling: In my defense that's copy and pasted :D

PR #58050, opened 2023-11-01
Chore: remove team notification every time "batch" is mentioned

camdencheek:
This removes the workflow that pings the team every time the word "batch" is mentioned. 

## Test plan

N/A


peterguy (approved):
Time to remove this notification: SNR is too low

Issue #57999, opened 2023-10-30
Directory page showing duplicate file names


peterguy:
- **Sourcegraph version:** sourcegraph.com

This looks like a bug in the directory page where it's showing two files with the same name while the tree view is showing only a single file.

#### Steps to reproduce:

Link to reproduce: https://sourcegraph.com/github.com/sourcegraph/scip-java/-/tree/semanticdb-gradle-plugin/src/main/scala

#### Expected behavior:

One file in the directory page.

#### Actual behavior:

Multiple files in the directory page.

![screenshot of bug](https://github.com/sourcegraph/sourcegraph/assets/129280/0c46cd55-0459-4e75-a908-7272d8364389)

[Internal conversation](https://sourcegraph.slack.com/archives/C05MHAP318B/p1698654812627649)

Issue #57936, opened 2023-10-26
File tree: forever loading icon


camdencheek:
[This directory](https://sourcegraph.com/github.com/backstage/backstage@HEAD/-/tree/contrib/chart) loads forever in the file tree. Not sure why yet. 

![Image](https://github.com/sourcegraph/sourcegraph/assets/12631702/c3627230-6230-402e-b24f-cf52f84a7b8d)

camdencheek:
@eseliger I think this might actually have to do with the recent changes you made to this resolver. Would you mind taking a look? It seems to happen consistently when expanding any directory that contains only a single child.

eseliger:
Duh, taking a look.

camdencheek:
Looks like [this one](https://github.com/sourcegraph/sourcegraph/issues/57999) might also be related. If you don't have the time to look at it, @BolajiOlajide and I are planning to tackle these on Friday, so no worries

PR #58058, opened 2023-11-01
Fixup GitTree behavior after adding test suite

eseliger:
There were two issues with that PR that introduced the test suite:
- When switching recursive on, we accidentally added more entries than expected, because git would also output parents of the requested subtree to list.
- I misunderstood what recursiveSingleChild is supposed to do. As a result, the code is a bunch simpler.

Closes https://github.com/sourcegraph/sourcegraph/issues/57999 Closes https://github.com/sourcegraph/sourcegraph/issues/57936

## Test plan

Added tests for both of these cases.


camdencheek (approved):
thank you! Much easier to understand as a bonus as well

PR #58061, opened 2023-11-01
Admin: add search indexing to repo menu

jtibshirani:
When debugging a search issue, it's common to want to check the status of
search indexing. Currently, you need to open the repo menu and click through to
'Settings' -> 'Search Indexing', which is not obvious. This change adds 'Search
indexing' directly to the repo menu, matching what we do for 'Permissions' and
'Last sync log'.

To counterbalance this, this change also removes the 'Embeddings Jobs' item. It
felt cluttered to have two 'Embeddings' items, and we want to direct people
towards policies instead of manually scheduling jobs.

## Test plan

Manual testing


jtibshirani:
Here's what it looks like now:
Screenshot 2023-11-01 at 4 00 26 PM

docker-images/syntax-highlighter/crates/scip-syntax/testdata/globals.ts
```diff
@@ -8,6 +8,19 @@ class MyClass {
...
+    return (
+        <OpenByDefaultWrapper>
+            {({ menuButtonRef }) => (
+        <WebStory>
+            {webProps => <UserNavItem {...props} {...webProps} menuButtonRef={menuButtonRef} />}
+    </WebStory>
+    )}
+    </OpenByDefaultWrapper>
+)
+}
```
camdencheek, line 22: Is this formatting funky, or is that just github?

camdencheek (approved):
Nice! LGTM.

docker-images/syntax-highlighter/crates/scip-syntax/testdata/globals.ts
camdencheek, line 22: This looks unrelated?

jtibshirani: Oof, thought I had fixed this! Thanks.

PR #58065, opened 2023-11-01
Release: add qdrant to images list

camdencheek:
Qdrant was missing, causing some incorrect behavior when updating images.

## Test plan

It got me another step closer to a clean release.


DaedalusG (approved):
Lgtm

PR #57774, opened 2023-10-20
telemetry-export: include interaction.traceId on events

bobheadxi:
This change will add OpenTelmetry trace IDs, which can be propagated from clients to Sourcegraph backends, to all events as they get exported if the trace ID is valid. This trace ID is then reported to Telemetry Gateway and our BigQuery data, allowing us to get per-interaction matching across different data sets, such as Cody Gateway's data, as was asked about [here in the "Gateway to eventlogger mapping" doc](https://docs.google.com/document/d/1V5MK_SI7OnvzVSi1AnhuX5RzvIZTNIZ3MdHACTBz0tE/edit?disco=AAAA67cXdpM).

I think standardizing on trace IDs for tracking user interactions is a good idea because it's an open standard for describing interactions already, and integrates with common observability tooling through trace exports, so investments made to track interactions for telemetry benefit observability and vice versa.

Trace IDs can be propagated via https://opentelemetry.io/docs/instrumentation/js/propagation/ - this was already set up in the web app in the past.

Multiple events can correspond to the same trace ID, so this is different from event ID.

## Test plan

Unit tests

PR #58068, opened 2023-11-02
tracer: enforce trace policy entirely via Sampler

bobheadxi:
This change updates our internal tracer to enforce policy only via a `Sampler` implementation. This has the following benefits:

1. Even when a trace should not be sampled, contexts are still populated with valid spans, rather than no-op ones. This is important to make use of trace IDs for non-tracing purposes, e.g. https://github.com/sourcegraph/sourcegraph/pull/57774 and https://github.com/sourcegraph/sourcegraph/pull/58060
2. We enforce trace policies the way they were meant to be enforced in OpenTelemetry: by simply indicating that the span should not be exported.

This was not possible before because OpenTracing did not use context propagation, hence we did not have a way to use trace policy flags set in context - but thanks to @camdencheek's work removing OpenTracing entirely, we can now do this in a more idiomatic fashion.

Thanks to this, I've removed a few places that prevented trace context from being populated based on trace policy (HTTP and GraphQL middleware, and `internal/trace`). This delegates sampling decisions to the sampler, and ensures we accept valid trace context everywhere.

## Test plan

Unit tests on a TracerProvider configured with the new sampler.

Manual testing:

```
sg run jaeger otel-collector 
sg start
```

Setting `observability.tracing.debug` to `true` we can see logs indicating the desired traits for non-`ShouldTrace` traces:

```
[         worker] INFO tracer tracer/logged_otel.go:63 Start {"spanName": "workerutil.dbworker.store.insights_query_runner_jobs_store.dequeue", "isRecording": false, "isSampled": false, "isValid": true}
```

With `observability.tracing.sampling` set to `none`, running a search with `&trace=1` only gives us spans from zoekt, which seems to have always been outside our rules here.

With `observability.tracing.sampling` set to `selective`, running a search with `&trace=1` gives us a full trace.

With `observability.tracing.sampling` set to `all`, Jaeger instantly gets loads of traces, and in logs we see:

```
[         worker] INFO tracer tracer/logged_otel.go:63 Start {"spanName": "workerutil.dbworker.store.exhaustive_search_worker_store.dequeue", "isRecording": true, "isSampled": true, "isValid": true}
```


internal/trace/policy/policy.go
```diff
@@ -38,9 +38,15 @@ type key int
 
 const shouldTraceKey key = iota
 
-// ShouldTrace returns true if the shouldTraceKey context value is true. It is used to
-// determine if a trace should be started by various middleware. If the value is not set
-// at all, we check if we should the global policy is set to TraceAll instead.
+// ShouldTrace returns true if the shouldTraceKey context value is true. If the
+// value is not set at all, we check if we should the global policy is set to
```
burmudar, line 42:
```suggestion
// value is not set at all, we check if what the global policy is set to
```

burmudar, line 42:
```suggestion
// value is not set at all, we check if the global policy is set to
```

burmudar (approved):
Nice!

cmd/frontend/graphqlbackend/graphqlbackend.go
```diff
@@ -182,10 +180,7 @@ func (requestTracer) TraceField(ctx context.Context, _, typeName, fieldName stri
 }
 
 func (t requestTracer) TraceValidation(ctx context.Context) func([]*gqlerrors.QueryError) {
-	var finish func([]*gqlerrors.QueryError)
-	if policy.ShouldTrace(ctx) {
-		finish = t.tracer.TraceValidation(ctx)
-	}
+	finish := t.tracer.TraceValidation(ctx)
 	return func(queryErrors []*gqlerrors.QueryError) {
 		if finish != nil {
```
camdencheek, line 185: In all these spots, `finish` will always be non-nil now, right?

camdencheek (approved):
very exciting!

cmd/frontend/graphqlbackend/graphqlbackend.go
bobheadxi: You're right, simplified it!

bobheadxi:
Backporting as bugfix for trace ID propagation

bobheadxi:
Manual backport: https://github.com/sourcegraph/sourcegraph/pull/58084

(logger diffs 😰 )

Issue #56379, opened 2023-09-06
Provide centralized Go package which handles language detection, file extension mappings etc.


varungandhi-src:

keegancsmith:
Can you take a look inside of zoekt as well.

The problem we ran into is when someone says lang:c or lang:c++ how do we return header files that make sense for that query. IE as you mention we can't treat files as 100% being a language. So when a user asks for all files that match a language how do we model that in our indexes?

varungandhi-src:
> IE as you mention we can't treat files as 100% being a language.

I don't think I quite said this (or if I did elsewhere, that's not quite so straightforward). From an API standpoint, I think it makes sense to expose multiple languages in our GraphQL API in case of ambiguity so that the client can properly handle ambiguity as desired.

From a search POV, I think such files are likely to be quite rare in practice, so modeling it as 1 language per file should match users' mental model closely. Copying my answer from [Slack](https://sourcegraph.slack.com/archives/C05EA9KQUTA/p1707191732763329?thread_ts=1706827628.336119&cid=C05EA9KQUTA) in a recent thread so that this is documented in GitHub and not just Slack:

> * For the immediate short term, the `search-content-based-lang-detection` setting does reliably distinguish between C and C++ files with the .h extension, so it makes sense to try to turn it on by default for indexed search (via Zoekt
> * Based on my experience working in C and C++, I think developers generally tend to think of a file as belonging to one language and it is not useful to show results from C when searching with `lang:C++` and vice-versa. This also reflects the 'By language' breakdown in the search results page, which bakes in the assumption that 1 file has 1 language. So for simplicity, it makes sense for Zoekt to continue storing one language per file as it does today. Users can combine lang:C OR lang:C++ if they want to mix languages (or that can be folded into a search context for ease of use).

PR #58069, opened 2023-11-02
squirrel: Do case-insensitive comparison of file extensions

varungandhi-src:
While investigating https://github.com/sourcegraph/sourcegraph/issues/57974, I noticed
that the list of file extensions in the Squirrel code for C and C++ is partly wrong
and partly incomplete. So this PR fixes that. Specifically:

There are a bunch of uppercase extensions in use:
- `.C`, `.CC` and so on -- confirmed via Sourcegraph.com using a `count:all` query.
- The `.pc` and `.pcc` files are just wrong. `pc` is most common for pkg-config files,
  and `pcc` is for PCGen files.

~We don't add a general lower-casing step before comparing the extension as it is
not common in other languages to have uppercase file extensions.~

~However, we could do that, since the current list of extensions is all lower-case.~

Ideally, https://github.com/sourcegraph/sourcegraph/issues/56379 would've been
fixed already, and we'd just use that, instead of fixing these stupid issues as one-offs.

## Test plan

n/a

### DIFF
--- a/cmd/symbols/squirrel/language-file-extensions.json
+++ b/cmd/symbols/squirrel/language-file-extensions.json
@@ -51,9 +51,7 @@
     "h++",
     "hh",
     "h",
-    "hpp",
-    "pc",
-    "pcc"
+    "hpp"
   ],
   "csharp": [
     "cs",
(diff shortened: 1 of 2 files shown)


keegancsmith:
I think doing a case insensitive comparison makes sense. Lots of people do dev on case insensitive file systems (windows/default darwin). I'm also not aware of any extensions for a language that mean different things depending on case?

keynmol (approved):

varungandhi-src:
Turns out, the `.pc` and `.pcc` extensions were marked as being valid for C++ for a customer. https://github.com/sourcegraph/customer/issues/124

PR #58072, opened 2023-11-02
blob: Simplify/cleanup some extension usage

fkling:
- Move the end-of-file decoration extension into its own module
- Remove blobInfo dependency of `buildLinks` extension (renamed to `linkify`)
- Use `useCompartment` for blobProps and wrap settings to simplify component structure.
- Remove theme extension. We are using our own CSS rules for dark/light theme styling.

## Test plan

Manual testing:

- Last line is hidden when empty
- Links are clickable

### DIFF
--- a/client/web/src/repo/blob/codemirror/eof.ts
+++ b/client/web/src/repo/blob/codemirror/eof.ts
@@ -48,5 +48,5 @@ export const hideEmptyLastLine: Extension = [
             fontStyle: 'italic',
             marginTop: '.2rem',
         },
-    })
+    }),
 ]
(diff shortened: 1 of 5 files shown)


client/web/src/repo/blob/codemirror/eof.ts
```diff
@@ -0,0 +1,52 @@
+import type { Extension } from '@codemirror/state'
+import { Decoration, EditorView, WidgetType } from '@codemirror/view'
+
+class NoLineBreakWidget extends WidgetType {
```
jasonhawkharris, line 4: Nit: When I was working on this it kind of bugged me that this needs to be class-based but I don't know whether it would be worth it to change it.

jasonhawkharris (approved):
LGTM. Just left a nit-note.

client/web/src/repo/blob/codemirror/eof.ts
fkling: I assume there is certain base functionality that widgets need to implement and having them extend a base class was the simplest approach? idk

PR #47290, opened 2023-02-01
grpc gitserver exec

camdencheek:
## Test plan

Unit tests

PR #58075, opened 2023-11-02
gitserver: include stderr in output error message

keegancsmith:
A long time ago in a galaxy far, far away we included stderr in the error messages returned from git. From reading our codepaths nearly everything goes via DividedOutput which throws the message away since CommandStatusError.Message is always set to a message like "exit status 128". This makes it really hard to understand why something isn't working.

A risk here is we intentionally got rid of stderr output due to security concerns? However, I think the more likely reason was accidental from when the whole RemoteGitCommand vs LocalGitCommand abstraction was introduced.

I noticed this when trying to debug why hybrid search was failing.

Test Plan: CI

### DIFF
--- a/internal/gitserver/git_command.go
+++ b/internal/gitserver/git_command.go
@@ -205,10 +205,20 @@ func (c *RemoteGitCommand) DividedOutput(ctx context.Context) ([]byte, []byte, e
 	return stdout, nil, nil
 }
 
-// Output runs the command and returns its standard output.
+// Output runs the command and returns its standard output. If the command
+// fails it usually returns CommandStatusError.
 func (c *RemoteGitCommand) Output(ctx context.Context) ([]byte, error) {
-	stdout, _, err := c.DividedOutput(ctx)
-	return stdout, err
+	// Note: we do not use DividedOutput because we don't want its behaviour
+	// where it throws away stderr in the error message. Stderr in error is
+	// useful to us because the client is not asking for it.
+
+	rc, err := c.sendExec(ctx)
+	if err != nil {
+		return nil, err
+	}
+	defer rc.Close()
+
+	return io.ReadAll(rc)
 }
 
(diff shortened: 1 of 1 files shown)


keegancsmith:
Here is some more context of why I was surprised stderr was missing from a DM with @pjlast.

When we have a non-zero exit code the server sends this https://github.com/sourcegraph/sourcegraph/blob/main/cmd/gitserver/internal/server_grpc.go#L211-L230
which should be converted into a gitdomain error here via the use of readCloseWrapper https://github.com/sourcegraph/sourcegraph/blob/main/internal/gitserver/client.go#L882-L887
which formats the string with stderr included here https://github.com/sourcegraph/sourcegraph/blob/main/internal/gitserver/client.go#L907-L919

Maybe the problem is this line which throws all the useful info away? https://github.com/sourcegraph/sourcegraph/blob/main/internal/gitserver/git_command.go#L196-L197
yes I think that is the problem.

keegancsmith:
@sourcegraph/source does this change feel to risky to merge? IE should I close?

pjlast:
@keegancsmith I actually thought it was merged already 😬 which means it's actually part of `main` now since I pulled it into the E2E test fix branch which I merged yesterday

pjlast (approved):

eseliger (approved):
Oops, sorry I skipped over this. Feels like a good change to me, not sure why it'd be risky. Feel free to merge it :)

Side-note: TIL that DividedOutput only gives you stderr when an error happened, which was quite unexpected. Alas.

keegancsmith:
> Side-note: TIL that DividedOutput only gives you stderr when an error happened, which was quite unexpected. Alas.

Urgh. Looks like this is a regression from introducing grpc to gitserver https://github.com/sourcegraph/sourcegraph/pull/47290/files#diff-5ab63931b9d3d41253df1859ecb9f6086ffce6511b05dd0079a42e1ed969056b

PR #58080, opened 2023-11-02
Code Navigation: fix jumpy revision selector

jasonhawkharris:
Before, the tabs and associated panels in the revision selector were jumpy. Though snappy, the tabs in the selector leave the user playing a game of whack-a-mole. 

https://github.com/sourcegraph/sourcegraph/assets/62355966/29437f21-4764-4bce-aa53-038fc4a9b800

Now, the number of displayed commits/branches/tags are set dynamically based on the browser window height, so that they don't immediately jump to the right of the selector. 

https://github.com/sourcegraph/sourcegraph/assets/62355966/2194d634-7e7e-4f45-91bb-bf5b5302fb76

In cases where the browser window is *very* small, the tabs will still float to the right of the selector. But only if window is so small that it renders the entire web app useless anyways.

As a follow-up, I'd like to consider removing the `show more...` button. I'm very skeptical of its utility IRL. Personally, I have never used it, and if I need to find something, I always default to the search bar above each panel. If someone is pressing the `show more...` button, their intention is to scroll. When they've hit the show more button so many times that the data displayed exceeds the size of the container, the tab _will_ jump to the right of the selector. 

https://github.com/sourcegraph/sourcegraph/assets/62355966/276f2cc1-0753-4902-ab5d-f4dab4dd88bd

Ultimately, I don't think this is a popular enough use case to justify having the `show more...` button at all, and I think we should consider removing it. 

## Test plan
Manually test against small, medium, and large instances.


jasonhawkharris:
@taiyab I'd love to get your thoughts on my note about the `show more...` button and whether it's really necessary, or whether we should remove it altogether.

client/web/src/repo/RevisionsPopover/RevisionsPopoverReferences.tsx
```diff
@@ -145,12 +144,22 @@ export const RevisionsPopoverReferences: React.FunctionComponent<
...
+    const { height } = useWindowSize()
+    const [batchCount, setBatchCount] = useState(getBatchCount(height))
+
+    /*
+     * Set batch count based on current screen height
+     * to avoid jumping tabs in the UI.
+     * */
+    useMemo(() => {
+        setBatchCount(getBatchCount(height))
+    }, [height])
```
camdencheek, line 156: react n00b here: couldn't we do this more simply?

```suggestion
    /*
     * Set batch count based on current screen height
     * to avoid jumping tabs in the UI.
     * */
    const batchCount = useMemo(() => getBatchCount(height), [height]) 
```

jasonhawkharris: I'll try it out. Maybe @felixfbecker or @vovakulikov will know if this will work. My guess is that it won't but that's more instinct that actual knowledge.

camdencheek: I think you're looking for @fkling

jasonhawkharris: Yup. This totally works.

taiyab:
@jasonhawkharris One approach would be to:
- Trigger loading more on-scroll to the bottom of the existing list of loaded items
- Add counts for Branches / Tags / Commits next to each tab label (using the `Badge` component) so we don't lose that

Is there a use case here I'm not considering @rrhyne?

vovakulikov (commented):
@jasonhawkharris I think our Popover component already includes API for controlling jumping and switching side when content within popover container requires more space 

So I think if you change props here https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/client/web/src/repo/RepoRevisionContainer.tsx?L140

to something like 
```tsx
 <PopoverContent
                position={Position.bottomStart}
                flipping={Flipping.opposite}
                className="pt-0 pb-0"
                aria-labelledby="repo-revision-popover"
            >
```
prop `flipping` with value `opposite` will prevent popover content positioning over side points of the target. Let me know if it doesn't work, 😉

vovakulikov (approved):
Nice and clean, thanks @jasonhawkharris

Issue #58024, opened 2023-10-31
Ranking: improve SCIP ctags output


jtibshirani:
In https://github.com/sourcegraph/sourcegraph/issues/57659 we noticed that SCIP ctags assigned very general kinds compared to universal-ctags, missing important information needed for search ranking. This is a follow-up issue to improve SCIP ctags for the remaining languages it supports:
* C
* C++
* Java
* Scala
* Zig

I'm spinning this out of the original issue #57659 since there's no search ranking regression for these languages: either universal-ctags never supported them (Scala, Zig), or we hadn't enabled SCIP ctags yet (C, C++, Java). Once we're confident in the output, we can also enable SCIP ctags for C, C++, Java.

PR #58082, opened 2023-11-02
scip-ctags: add `kinds` support for C

mrnugget:
This is part of https://github.com/sourcegraph/sourcegraph/issues/58024 and adds scip-ctags support for C.

## Test plan

- Unit tests


docker-images/syntax-highlighter/crates/scip-syntax/src/snapshots/scip_syntax__test__scip_snapshot_example.c.snap
```diff
@@ -0,0 +1,115 @@
...
+//      ^^^^^^^ definition(Field) scip-ctags connection#bufsize.
+    char *buffer;
+//        ^^^^^^ definition(Field) scip-ctags connection#buffer.
+    char *url;
+//        ^^^ definition(Field) scip-ctags connection#url.
+  };
+  
+  // Prototype
+  struct connection *new_connection(int fd);
+//                   ^^^^^^^^^^^^^^ definition(Function) scip-ctags new_connection().
```
mrnugget, line 70: @olafurpg is this the right kind?

olafurpg: Typically, we try to assign language-specific kinds. In Go, we distinguish between methods and method specs so it might make sense to do the same for C and distinguish between functions and their prototypes.

mrnugget: Ahh! That’s very helpful. Thanks. Then I guess I’ll need to add `FunctionPrototype` to the schema.

varungandhi-src: The correct thing to do here would be to add a new `SymbolRole` (https://github.com/sourcegraph/scip/issues/131) and not a new case to `Kind`. The reason being that, `SymbolRole`s can vary on a per-occurrence basis, but a `Kind` is attached to a single `SymbolInformation`.

However, the same `SymbolInformation` can have multiple occurrences, say 1 definition, 1 forward declaration, and 3 other references.

I'll submit a PR to the scip repo to add this SymbolRole

mrnugget: Thanks @varungandhi-src! I can try updating in here.

olafurpg: @varungandhi-src great point! Agree that the symbol role is more appropriate if these two have the same symbol. For Go method specs, the symbols are different so it makes sense to have a different kind as well.

varungandhi-src: PR with new SymbolRole: https://github.com/sourcegraph/scip/pull/217

docker-images/syntax-highlighter/crates/scip-syntax/src/ts_scip.rs
```diff
@@ -78,6 +79,7 @@ pub fn symbol_kind_to_ctags_kind(kind: &symbol_information::Kind) -> Option<&'st
         Struct => Some("struct"),
         TypeAlias => Some("typeAlias"),
         Trait => Some("trait"),
+        Variable => Some("variable"),
```
mrnugget, line 82: @varungandhi-src @olafurpg The changes in this file also change a lot of other snapshot tests, I assume that's okay?

varungandhi-src: That's fine.

varungandhi-src: I'm a little confused by this list to be honest, because I just checked the output of `ctags --list-kind-full` and a bunch of kinds on the RHS aren't actually supported by ctags, but this function's name is `symbol_kind_to_ctags_kind`.

mrnugget: I'll leave that for @jtibshirani, not sure yet how these are consumed on the zoekt side.

docker-images/syntax-highlighter/crates/scip-syntax/src/snapshots/scip_syntax__test__scip_snapshot_example.c.snap
```diff
@@ -0,0 +1,115 @@
...
+  const char *author = "Petri & Thorsten";
+//            ^^^^^^ definition(Constant) scip-ctags author.
+  const int count;
+//          ^^^^^ definition(Constant) scip-ctags count.
+  int count2;
+//    ^^^^^^ definition(Variable) scip-ctags count2.
+  const int age = 28;
+//          ^^^ definition(Constant) scip-ctags age.
+  static uint sweet_sweet_numbers[5] = {23, 420, 69, 42, 7};
+//            ^^^^^^^^^^^^^^^^^^^ definition(Variable) scip-ctags sweet_sweet_numbers.
```
varungandhi-src, line 21: There is a separate `StaticVariable` Kind here: https://sourcegraph.com/github.com/sourcegraph/scip@main/-/blob/bindings/typescript/scip.ts?L1353:13

docker-images/syntax-highlighter/crates/scip-syntax/src/snapshots/scip_syntax__test__scip_snapshot_example.c.snap
```diff
@@ -0,0 +1,115 @@
+---
+source: crates/scip-syntax/src/lib.rs
+expression: dumped
+---
+  #include <stdlib.h>
+  #include <sys/types.h>
+  
+  #define BUFSIZE 4096
```
varungandhi-src, line 8: There should be a `Macro` kind here. https://sourcegraph.com/github.com/sourcegraph/scip@main/-/blob/bindings/typescript/scip.ts?L1319

docker-images/syntax-highlighter/crates/scip-syntax/testdata/example.c
```diff
@@ -1,3 +1,78 @@
...
+
+enum { BLACK, RED };
+
+enum animal {
+  ANIMAL_TOUCAN,
+  ANIMAL_TIGER,
+  ANIMAL_HORSE,
+  ANIMAL_GIRAFFE,
+  ANIMAL_GOPHER,
+  ANIMAL_ORANGUTAN
```
varungandhi-src, line 21: It'd be more helpful to test different bits of syntax compared to adding more examples for the same kind of syntax.

```suggestion
  ANIMAL_TIGER = 1,
  ANIMAL_TIGGER = ANIMAL_TIGER,
```

docker-images/syntax-highlighter/crates/scip-syntax/testdata/example.c
```diff
@@ -1,3 +1,78 @@
...
+};
+
+typedef enum instrument {
+  INSTRUMENT_GUITAR,
+  INSTRUMENT_KEYTAR,
+  INSTRUMENT_SITAR
+} Instrument;
+
+struct connection {
```
varungandhi-src, line 30: Similar to the `typedef enum` case above, please add a test case for `typedef struct` as that is a common pattern.

varungandhi-src:
I've mentioned some missing test cases inline. Some more test cases to add:
- K&R style declarations. [example](https://jameshfisher.com/2016/11/27/c-k-and-r/)
- Union declarations
- Make sure that we're not emitting symbols for type declarations in nested contexts (such as inside another type definition, inside a function body etc, inside a parameter list) (as those aren't accessible from a global scope). E.g. `struct z { struct w { int x; int y; } b; }` -- we shouldn't emit a global descriptor for `w` here.

mrnugget:
Thanks @varungandhi-src! `union` popped into my head earlier, but not the other things. I'm on it.

docker-images/syntax-highlighter/crates/scip-syntax/src/snapshots/scip_syntax__test__scip_snapshot_example.c.snap
```diff
@@ -0,0 +1,115 @@
...
+//            ^^^^^^^^^^^^^^^^^^^ definition(Variable) scip-ctags sweet_sweet_numbers.
+  
+  enum { BLACK, RED };
+//       ^^^^^ definition(EnumMember) scip-ctags BLACK.
+//              ^^^ definition(EnumMember) scip-ctags RED.
+  
+  enum animal {
+//     ^^^^^^ definition(Enum) scip-ctags animal#
+    ANIMAL_TOUCAN,
+//  ^^^^^^^^^^^^^ definition(EnumMember) scip-ctags animal#ANIMAL_TOUCAN.
```
varungandhi-src, line 30: The name here should be `ANIMAL_TOUCAN.` not `animal#ANIMAL_TOUCAN.`. In C, enum constants are not scoped to the containing type (which is why C uses type prefixes in enum constants as a convention). E.g. this code will not compile:

```
enum X {
    A
};

enum X {
    A
 // ^ error: redeclaration of enumerator 'A'
};
```

docker-images/syntax-highlighter/crates/scip-syntax/queries/c/scip-tags.scm
```diff
@@ -1,11 +1,72 @@
...
+(struct_specifier
+  name: (type_identifier) @descriptor.type @kind.struct ;; <-- capture struct kidn
+  body: (field_declaration_list))
+
+(struct_specifier
+  name: (type_identifier) @descriptor.type
+  body: (field_declaration_list
+          (field_declaration declarator: [
+            (pointer_declarator (field_identifier) @descriptor.term @kind.field)
+            (field_identifier) @descriptor.term @kind.field]))) ;; <-- capture field kind
```
varungandhi-src, line 67: AIUI, this isn't handling forward declarations for types, so add some queries + test cases for those too.

docker-images/syntax-highlighter/crates/scip-syntax/src/snapshots/scip_syntax__test__scip_snapshot_example.c.snap
mrnugget: Done!

docker-images/syntax-highlighter/crates/scip-syntax/testdata/example.c
mrnugget: makes sense!

docker-images/syntax-highlighter/crates/scip-syntax/src/snapshots/scip_syntax__test__scip_snapshot_example.c.snap
mrnugget: Done

docker-images/syntax-highlighter/crates/scip-syntax/testdata/example.c
mrnugget: Done

docker-images/syntax-highlighter/crates/scip-syntax/src/snapshots/scip_syntax__test__scip_snapshot_example.c.snap
mrnugget: Okay, dumb question: how do I capture a `SymbolRole` here? And also: does it matter for this exercise where we just want "Kinds"?

mrnugget:
Another thing: pointers to pointers!

```c
struct connection **my_ptr_ptr;
```

docker-images/syntax-highlighter/crates/scip-syntax/src/lib.rs
```diff
@@ -115,6 +115,8 @@ mod test {
     generate_tags_and_snapshot!(Scip, test_scip_javascript, "globals.js");
     generate_tags_and_snapshot!(Scip, test_scip_javascript_object, "javascript-object.js");
 
+    generate_tags_and_snapshot!(Scip, test_scip_c_example, "example.c");
```
jtibshirani, line 118: It's a little redundant, but I like testing the tags output directly too, just in case there's an issue generating them (we're missing a kind conversion, etc.) This helped me catch some issues with other languages. It would look like this:

```
generate_tags_and_snapshot!(All, test_tags_c_example, test_scip_c_example, "example.c");
```

docker-images/syntax-highlighter/crates/scip-syntax/src/ts_scip.rs
jtibshirani: @varungandhi-src this is a good observation. To truly match the ctags output, the kind names would depend on the language: for example, Kotlin uses `typealias` while Go uses `talias`. We decided that it's cleaner to pass through the SCIP kind names, and rely on clients (like Zoekt) to handle the different naming conventions.

Overall, I don't think we should treat the universal-ctags output as a "spec" -- the names would be inconsistent and it may prove to be a moving target. Instead, we want to switch to consuming SCIP directly in Zoekt... so we'll end up mapping the messy universe of universal-ctags kinds to SCIP kinds before they're consumed.

Outcome from my long explanation: I'll update the method name and add a comment :)

docker-images/syntax-highlighter/crates/scip-syntax/src/snapshots/scip_syntax__test__scip_snapshot_example.c.snap
jtibshirani: I don't think universal-ctags has a robust notion of "role", and we don't consume that anywhere (to the best of my knowledge). So for now we can skip adding symbol role information.

docker-images/syntax-highlighter/crates/scip-syntax/src/lib.rs
mrnugget: Thanks! Will do.

jtibshirani:
@mrnugget the simplifications look great and seem like the right trade-off. Our intended use cases for symbol kinds don't require a high level of precision (it's just ranking and showing the kinds for symbol search). I'll review again shortly.

mrnugget:
Update: talked with @tjdevries a bit and he also confirmed that getting `const`/`static` working reliably and fixing the anonymous union thing would require more than just a couple of hours of query writing -- would need to change the Rust code. Happy to do that, of course, but wanted to check, @jtibshirani, whether that's even wanted or not.

jtibshirani (approved):
Looks good to me! I see all the important kinds to detect.

docker-images/syntax-highlighter/crates/scip-syntax/src/ts_scip.rs
jtibshirani: Update: opened https://github.com/sourcegraph/sourcegraph/pull/58128

PR #57939, opened 2023-10-27
client/web: add telemetry v2 client

bobheadxi:

PR #58085, opened 2023-11-02
[Backport 5.2] client/web: add telemetry v2 client (#57939)

bobheadxi:
Brings something like https://github.com/sourcegraph/cody/pull/1192 to Sourcegraph's web client, injected as a new `telemetryRecorder` in `PlatformContext`. Because we want to use the only non-deprecated way to make GraphQL requests, Apollo clients, a global events recorder is not provided ([thread](https://sourcegraph.slack.com/archives/C01C3NCGD40/p1698376463350449)) - the shared one can only retrieved from `PlatformContext`.

In non-`web` platforms (`client/browser`) the recorder is currently set to a no-op recorder that errors upon use, as for this PR intended for backport, we only want to add some usages to `web` for Cody events.

To start off, this change adds some new usage of this on a rather out-of-the-way section `permissions-center`, tested manually below.

Closes https://github.com/sourcegraph/sourcegraph/issues/56920

(cherry picked from commit 726a5aff953dee7a07c63a1687396044be361d4d)

## Test plan

See test plan in https://github.com/sourcegraph/sourcegraph/pull/57939

## Backport notes

This is required to enable the telemetry team to instrument updated events in 5.2.x as part of the event-logging-everywhere project.


unknwon (approved):

unknwon:
Assuming you will fix the CI 😂

Issue #57827, opened 2023-10-23
Tree view: more useful recent activity


camdencheek:
Our tree view has a "recent activity" column which shows an opaque red/green line that supposedly represents how much has changed in that entry recently. This is pretty useless in practice. 

<img width="800" alt="CleanShot 2023-10-23 at 13 39 34@2x" src="https://github.com/sourcegraph/sourcegraph/assets/12631702/8a89a191-daf4-4b7c-bfb1-6836b738740a">

Compare to the tree view of another well-known codehost, which shows the most recent commit message, how long ago it was edited, and a link to that change. This is significantly more useful when browsing code.

<img width="1024" alt="CleanShot 2023-10-23 at 13 37 52@2x" src="https://github.com/sourcegraph/sourcegraph/assets/12631702/516d1183-2eed-4d4e-a8b4-f34e42f5b52c">

________

### Preview

<img width="1119" alt="image" src="https://github.com/sourcegraph/sourcegraph/assets/5337876/3970556e-60c2-4dac-889e-c923d08e153b">

### Link

https://www.figma.com/file/pOuuecCujCnOfvaNEcws9z/Tree-View?type=design&node-id=69%3A1280&mode=design&t=gt9kf9NDtGUi5SWL-1

taiyab:
@fkling Happy for you to do a first pass on this without design whilst I figure out what we can do on hovering the commit message.

fkling:
After talking with @taiyab, my plan is to

- load commit information lazily (I think we do the same for the "activity" information).
- remove "Commits" panel to reduce the amount of seemingly duplicate information
- add the history action to tree pages, so that users can still access the full history if they need to. That's at least consistent with the file page.

taiyab:
Here's a visual of the proposal:
<img width="1439" alt="image" src="https://github.com/sourcegraph/sourcegraph/assets/5337876/a1aa6376-8c4e-4178-8942-ee22546bcdf1">

Does the Commit panel actually buy us anything here? We should make sure whatever it's giving us at this point, we're able to bake into the file list is a useful way.

It's not an ideal solution, but I think it works alright in the interim, unless people really like the commit list for some reason?

cc @rrhyne

camdencheek:
Removing the commit panel sounds good to me. Though, I agree with Felix that it would be nice to still be able to view history for a directory. A unified history panel across both the blob and tree views sounds excellent

taiyab:
[...]

PR #58088, opened 2023-11-02
GraphQL: add history to TreeEntry

camdencheek:
This adds a new `history` method to the `TreeEntry` GraphQL object. It returns the history of that entry as a set of `TreeEntry` nodes. This is to suport #57827. Previously, fetching the history would take a separate GraphQL request entry.

A sticky design choice was whether to make `history` return a connection of `TreeEntry` or `Commit` objects. I think that the choice to return `TreeEntry` is slightly more correct even though it requires another layer of querying to get to the underlying commit for two reasons:
1) If we just return the commits that modified the tree, if we wanted to fetch the content at the historical revision (say, to fetch a diff), we'd have to make a followup query for the path at the returned commit instead of just adding a field to the first query.
2) It makes it theoretically possible to follow renames. We don't do that right now because it wouldn't be super easy to implement, but if we returned commits that modified the path and followed renames, when we did a followup query to fetch the content at one of the historical commits, it might fail because the path doesn't exist (it was renamed). The returned tree entry naturally contains the path name, so we could return the changed path name along with the history entry.

Example query:

```graphql
query GetTreeHistory {
  repository(name: "github.com/sourcegraph/sourcegraph") {
    id
    name
    commit(rev: "HEAD") {
      tree {
        entries {
          history(first: 1) {
            nodes {
              commit {
                message
                author{
                  date
                }
              }
            }
          }
        }
      }
    }
  }
}
```

## Test plan

Add a functional test for the resolver. Manually checked the output of the GraphQL query.

### DIFF
--- a/cmd/gitserver/internal/integration_tests/archivereader_test.go
+++ b/cmd/gitserver/internal/integration_tests/archivereader_test.go
@@ -35,7 +35,7 @@ import (
 )
 
 func TestClient_ArchiveReader(t *testing.T) {
-	root := gitserver.CreateRepoDir(t)
+	root := t.TempDir()
 
 	type test struct {
 		name string
--- a/cmd/gitserver/internal/integration_tests/archivereader_test.go
+++ b/cmd/gitserver/internal/integration_tests/archivereader_test.go
@@ -35,7 +35,7 @@ import (
 )
 
 func TestClient_ArchiveReader(t *testing.T) {
-	root := t.TempDir()
+	root := gitserver.CreateRepoDir(t)
 
 	type test struct {
 		name string
--- a/cmd/frontend/graphqlbackend/BUILD.bazel
+++ b/cmd/frontend/graphqlbackend/BUILD.bazel
@@ -82,6 +82,7 @@ go_library(
         "git_revision.go",
         "git_tree.go",
         "git_tree_entry.go",
+        "git_tree_history.go",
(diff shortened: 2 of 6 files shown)


cmd/gitserver/internal/integration_tests/archivereader_test.go
```diff
@@ -35,7 +35,7 @@ import (
 )
 
 func TestClient_ArchiveReader(t *testing.T) {
-	root := gitserver.CreateRepoDir(t)
+	root := t.TempDir()
```
fkling, line 38: Is this intentional? Just double checking, since I don't know anything about our backend 😬

fkling (approved):
Nice! I guess we'll have to see how much more expensive it will be to fetch really large directories.

cmd/gitserver/internal/integration_tests/archivereader_test.go
BolajiOlajide: I believe it's a cheaper way of creating a directory - this gets cleaned up automatically AFAIK.

BolajiOlajide (approved):

cmd/gitserver/internal/integration_tests/archivereader_test.go
camdencheek: Oh, whoops, not intentional. Thanks!

PR #58092, opened 2023-11-03
[Backport 5.2] fix: Check for known Squirrel errors inside GraphQL responses (#58030)

varungandhi-src:
Cherry-pick #58030 to release branch.

Fixes https://github.com/sourcegraph/sourcegraph/issues/57974

## Test plan

Same as #58030

### DIFF
--- a/client/shared/src/codeintel/legacy-extensions/util/api.ts
+++ b/client/shared/src/codeintel/legacy-extensions/util/api.ts
@@ -1,6 +1,8 @@
 import { once } from 'lodash'
 import gql from 'tagged-template-noop'
 
+import { isErrorLike } from '@sourcegraph/common'
+
 import { searchContext } from '../../searchContext'
 import type * as sourcegraph from '../api'
 import { cache } from '../util'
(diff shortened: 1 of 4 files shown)


keynmol (approved):

client/shared/src/codeintel/legacy-extensions/util/api.ts
```diff
@@ -56,6 +58,14 @@ export interface RepoMeta {
     isArchived: boolean
 }
 
+function isKnownSquirrelErrorLike(value: unknown): boolean {
+    return (
+        isErrorLike(value) &&
+        'message' in value &&
+        (value.message.includes('unrecognized file extension') || value.message.includes('unsupported language'))
+    )
+}
```
keegancsmith, line 67: is there a way to make the backend return something more structured such that the client side code doesn't need changing? Or has this always been a problem, even before squirrel?

keegancsmith (approved):
random inline comment, although I know this is a backport.

client/shared/src/codeintel/legacy-extensions/util/api.ts
varungandhi-src: The error format here is coming from the gRPC layer, which returns the result in this format with the error string inside the `message` field. It's not really Squirrel specific.

PR #58093, opened 2023-11-03
client: Remove needless useMemo calls in ref panel code

varungandhi-src:
Following up from Slack thread: https://sourcegraph.slack.com/archives/C01C3NCGD40/p1698900105171149

## Test plan

n/a

### DIFF
--- a/client/web/src/codeintel/ReferencesPanel.tsx
+++ b/client/web/src/codeintel/ReferencesPanel.tsx
@@ -281,6 +281,8 @@ const SearchTokenFindingReferencesList: React.FunctionComponent<
 
 const SHOW_SPINNER_DELAY_MS = 100
 
+const empty: any[] = []
+
 const ReferencesList: React.FunctionComponent<
     React.PropsWithChildren<
         ReferencesPanelPropsWithToken & {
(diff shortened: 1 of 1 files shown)


client/web/src/codeintel/ReferencesPanel.tsx
```diff
@@ -350,10 +350,10 @@ const ReferencesList: React.FunctionComponent<
         // Make sure this effect only runs once
     }, [loading])
 
-    const references = useMemo(() => data?.references.nodes ?? [], [data])
-    const definitions = useMemo(() => data?.definitions.nodes ?? [], [data])
-    const implementations = useMemo(() => data?.implementations.nodes ?? [], [data])
-    const prototypes = useMemo(() => data?.prototypes.nodes ?? [], [data])
```
fkling, line 356: Those calls are potentially not unnecessary. It depends on how those values are used. The new code will create new arrays every time the component is updated and `data.X` isn't set. One way to work around this is to define an empty array outside the component and use that here, e.g.

```js
var empty: any[] = []

//...
const references = data?.references.nodes ?? empty
// ...
```

Though that might require typing `references` properly.

It might also not be a problem. As I said, it depends on how those values are used downstream.

fkling (approved):

client/web/src/codeintel/ReferencesPanel.tsx
varungandhi-src: Thanks, I was implicitly assuming that empty arrays had referential equality but that's not true. (`[] == []` returns `false` in Node.)

I've changed this to use your suggestion, albeit with `const empty` instead of `var empty`.

fkling: Oops, yeah `const` of course :) An object (which includes arrays in JS) is only equal to itself in JS.

PR #58100, opened 2023-11-03
Search blitz: increase poll interval to 5m

camdencheek:
This increases the poll interval of Search Blitz from 1m to 5m. We are currently running about 80 queries per minute, which is a non-insignificant amount of load on sourcegraph.com. This should cut down the load significantly while keeping a decent granularity for our charts.

[Slack thread](https://sourcegraph.slack.com/archives/C05EA9KQUTA/p1698962313071229)

## Test plan

N/A


eseliger (approved):
Remember to manually roll this out afterwards, I don't this updates automatically :)

PR #58106, opened 2023-11-03
Code Navigation: fix offset diff header

jasonhawkharris:
Small commit to align offset diff header with the rest of the page.

Before:
Screenshot 2023-11-03 at 2 49 43 PM

After:
Screenshot 2023-11-03 at 2 49 10 PM

## Test plan
Test manually


peterguy (approved):
great!

camdencheek (approved):
nice, LGTM! Have you checked what this looks like in split view?

PR #58109, opened 2023-11-05
Svelte: remove `extends` directive for a file that doesn't exist

camdencheek:
The svelte `tsconfig.json` file references another `tsconfig` that doesn't exist. This causes my editor to complain. I don't think this is specific to my environment because there is no `.gitignore` entry that would cause that file to be ignored. 

## Test plan

My editor no longer complains.

### DIFF
--- a/client/web-sveltekit/tsconfig.json
+++ b/client/web-sveltekit/tsconfig.json
@@ -1,5 +1,4 @@
 {
-  "extends": "./.svelte-kit/tsconfig.json",
   "compilerOptions": {
     "allowJs": true,
     "checkJs": true,


fkling:
`.sveltekit` gets generated during the build process. When exactly does your editor complain? You could also run `pnpm sync` inside `client/web-sveltekit/` to generate it once.

camdencheek:
🤦 Turns out, there _is_ a `.gitignore` entry for the `.svelte-kit` directory, it's just in `client/web-sveltekit/.gitignore` rather than the top-level `.gitignore`. Thanks for pointing me in the right direction!

PR #58114, opened 2023-11-06
WIP: remove dependency on `users.site_admin` column when checking for site_admin status

BolajiOlajide:


camdencheek:
@BolajiOlajide where are we at on this PR? IIRC, it was pretty close where we left offa

PR #58117, opened 2023-11-06
sec: Define GraphQL cost limits

evict:
This sets a maximum GraphQL query depth of 20 and sets a maximum number of aliases. This helps prevent queries that consume too many resources and lead to a Denial-of-Service.

For more information, or discussion on the change please refer to: 
- https://github.com/sourcegraph/security-issues/issues/359 for the the maxdepth definition;
- https://github.com/sourcegraph/security-issues/issues/358 for the max field count check;

## Test plan
- Tested on local instance
- CI tests

### DIFF
--- a/cmd/frontend/graphqlbackend/rate_limit.go
+++ b/cmd/frontend/graphqlbackend/rate_limit.go
@@ -22,9 +22,14 @@ import (
 // the algorithm
 const costEstimateVersion = 2
 
+const MaxAliasCount = 500        // SECURITY: prevent too many aliased queries
+const MaxFieldCount = 500 * 1000 // SECURITY: prevent deeply nested or overly broad queries
+const maxDepth = 30              // SECURITY: prevent deep queries that consume too many resources
+
 type QueryCost struct {
 	FieldCount int
 	MaxDepth   int
+	AliasCount int
 	Version    int
 }
 
@@ -126,6 +131,7 @@ func EstimateQueryCost(query string, variables map[string]any) (totalCost *Query
 			return nil, errors.Wrap(err, "calculating operation cost")
 		}
 		totalCost.FieldCount += cost.FieldCount
+		totalCost.AliasCount += cost.AliasCount
 		if totalCost.MaxDepth < cost.MaxDepth {
 			totalCost.MaxDepth = cost.MaxDepth
 		}
@@ -158,6 +164,7 @@ func calcNodeCost(def ast.Node, fragmentCosts map[string]int, variables map[stri
 	limitStack := make([]int, 0)
 	currentLimit := 1
 
+	aliasCount := 0
 	fieldCount := 0
 	depth := 0
 	maxDepth := 0
@@ -191,6 +198,9 @@ func calcNodeCost(def ast.Node, fragmentCosts map[string]int, variables map[stri
 				}
 				pushLimit()
 			case *ast.Field:
+				if node.Alias != nil {
+					aliasCount++
+				}
 				switch node.Name.Value {
 				// Values that won't appear in the result
 				case "nodes", "__typename":
@@ -306,6 +316,7 @@ func calcNodeCost(def ast.Node, fragmentCosts map[string]int, variables map[stri
 	return &QueryCost{
 		FieldCount: fieldCount + maxInlineFragmentCost,
 		MaxDepth:   maxDepth,
+		AliasCount: aliasCount,
 	}, visitErr
 }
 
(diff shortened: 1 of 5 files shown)


unknwon (approved):
Would like a review from @eseliger 😁

cmd/frontend/graphqlbackend/rate_limit.go
```diff
@@ -22,6 +22,9 @@ import (
 // the algorithm
 const costEstimateVersion = 2
 
+const MaxFieldCount = 500 // SECURITY: prevent queries that have too many fields
+const maxDepth = 20       // SECURITY: prevent deep queries that consume too many resources
+
```
evict, line 27: We should be able to get more accurate number from traces.

camdencheek:
Do we have any queries that traverse the commit graph or the tree graph recursively? If so, this will probably break things.

camdencheek:
I did a quick scan through our fragments for git commits and git trees, and did not see any that operate recursively, so I think we're good there. It's possible we have other things that could be queried recursively though.

evict:
> I did a quick scan through our fragments for git commits and git trees, and did not see any that operate recursively, so I think we're good there. It's possible we have other things that could be queried recursively though.

Alright! We can change the number if we notice issues on the internal instances. Currently it seems the max field count is an issue (tests), so I will try to find a more sane value.

evict:
> > I did a quick scan through our fragments for git commits and git trees, and did not see any that operate recursively, so I think we're good there. It's possible we have other things that could be queried recursively though.
> 
> Alright! We can change the number if we notice issues on the internal instances. Currently it seems the max field count is an issue (tests), so I will try to find a more sane value.

I will remove the maximum field count, as it will break permission syncing it seems. The alias count check will be sufficient to resolve the issues mentioned above.

eseliger:
> Do we have any queries that traverse the commit graph or the tree graph recursively? If so, this will probably break things.

Unless I'm absolutely missing something, I don't think GraphQL does not allow recursive queries. So for that cause it shouldn't be a problem.

eseliger (commented):
IMO, this is a bit of a duct tape fix. I don't want to block this if that's a serious security concern, but here's why I think it'll not get there all the way:

- When I use 100 aliases to get a very simple, non-nested top-level value, that actually wouldn't cause a lot of load at all, but would be blocked now
- Using maxDepth 20 _could_ in theory mean that some legit queries become forbidden (if not from out clients, then perhaps from implementors)
- Some deeply nested resolvers might still be dead cheap to compute
- There certainly are not-20-deep nested resolver paths that generate a massive amount of load causing another source of DoS

Also, now we limit the maxDepth to 20, but what if I combine my 10 aliases, multiply by 20 max depth, and use a resolver that is computationally expensive and expands to a connection at each level. 
Could probably still make 100k+ resolver handlers run, vs exactly 21 in other cases that are now forbidden.

A common approach for GraphQL (and we can still hard cap things at larger numbers, too, if we want to) calculate a query "cost", which conveniently is implemented already by the function we modify here 😬. That's what GitHub does, too. Queries above a certain cost will be rejected, which accounts for very deep queries and lots of aliases in a combined solution. 
So I wonder: Why was that causing a DoS? Because we didn't enable the cost rate limiter? 
If so, would enabling it fix the reported issue as well?

evict:
> IMO, this is a bit of a duct tape fix. I don't want to block this if that's a serious security concern, but here's why I think it'll not get there all the way:
> 
> * When I use 100 aliases to get a very simple, non-nested top-level value, that actually wouldn't cause a lot of load at all, but would be blocked now
> * Using maxDepth 20 _could_ in theory mean that some legit queries become forbidden (if not from out clients, then perhaps from implementors)
> * Some deeply nested resolvers might still be dead cheap to compute
> * There certainly are not-20-deep nested resolver paths that generate a massive amount of load causing another source of DoS
> 
> Also, now we limit the maxDepth to 20, but what if I combine my 10 aliases, multiply by 20 max depth, and use a resolver that is computationally expensive and expands to a connection at each level. Could probably still make 100k+ resolver handlers run, vs exactly 21 in other cases that are now forbidden.
> 
> A common approach for GraphQL (and we can still hard cap things at larger numbers, too, if we want to) calculate a query "cost", which conveniently is implemented already by the function we modify here 😬. That's what GitHub does, too. Queries above a certain cost will be rejected, which accounts for very deep queries and lots of aliases in a combined solution. So I wonder: Why was that causing a DoS? Because we didn't enable the cost rate limiter? If so, would enabling it fix the reported issue as well?

Yes, we can calculate the proper cost too. I was previously using the maximum field count, but it is hard to determine the right values here and integration tests were failing. That's when I decided to implement the `aliasCount` instead.

The DoS is indeed caused by lack of limiting. With huge queries (e.g. loads of objects or aliases) the instance runs out of memory.

Defining maximum depth and aliases are a common method that help reduce DoS risk: https://escape.tech/graphql-armor/docs/plugins/max-aliases#:~:text=Limit%20the%20number%20of%20aliases,DOS%20attack%20or%20heap%20overflow. We can definitely increase the numbers, I think even around 100-200 should be completely fine (more even).

PR #58128, opened 2023-11-06
SCIP ctags: explain the 'kind' naming

jtibshirani:
This change adds a comment explaining how the 'kind' names don't always match
the universal-ctags output exactly.

## Test plan

Comment-only change

### DIFF
--- a/docker-images/syntax-highlighter/crates/scip-syntax/src/ts_scip.rs
+++ b/docker-images/syntax-highlighter/crates/scip-syntax/src/ts_scip.rs
@@ -60,8 +60,11 @@ pub fn captures_to_kind(kind: &Option<&String>) -> symbol_information::Kind {
 // Universal-ctags uses inconsistent kind names for the same concept (for example typealias for
 // Kotlin vs. talias for Go). For simplicity, we don't try to match the universal-ctags output, we
 // just pass through the SCIP kinds and rely on clients like Zoekt to handle the different naming
-// conventions. When adding a new entry here, be sure to check if Zoekt search ranking needs to be
-// updated to recognize the new kind.
+// conventions.
+//
+// When adding a new entry here, be sure to check if Zoekt search ranking needs to be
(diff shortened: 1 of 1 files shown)


docker-images/syntax-highlighter/crates/scip-syntax/src/ts_scip.rs
```diff
@@ -54,6 +54,13 @@ pub fn captures_to_kind(kind: &Option<&String>) -> symbol_information::Kind {
     })
 }
 
+
+// Converts a SCIP symbol kind to the ctags format.
+//
+// Universal-ctags uses inconsistent kind names for the same concept (for example typealias for
+// Kotlin vs. talias for Go). For simplicity, we don't try to match the universal-ctags output, we
+// just pass through the SCIP kinds and rely on clients like Zoekt to handle the different naming
+// conventions.
```
varungandhi-src, line 63: Could you include a recommendation for someone who wants to add a new string in the output here? What are the steps they need to take? Does Zoekt also need to be updated when a new kind is added here?

jtibshirani: I added a note about checking if Zoekt search ranking needs to be updated to recognize the new kind. I don't want to be more specific here (mentioning a specific Zoekt file, method, etc.) as this can easily lead to comment drift.

> I'd assumed that the process to follow was to find the closest match from ctags --list-kinds-full and use that, but it seems like that's not the right thing to do here.

The process is just to pass through the SCIP kind name, as we chatted about here: https://github.com/sourcegraph/sourcegraph/pull/58082#discussion_r1381964036. The SCIP kinds map very well onto universal-ctags kinds, but their names are cleaner and consistent across languages. I also added a note about the plan to switch to SCIP to make this process less fragile.

varungandhi-src (approved):
LGTM, thanks!

PR #58134, opened 2023-11-06
Git blame: make ignoring whitespace configurable

camdencheek:
Currently, our `git blame` API endpoint adds the `-w` flag to `git blame`, which causes whitespace-only changes to be ignored when blaming a line. This is different than GitHub's behavior, and a confusing default.

This makes the ignoring whitespace configurable in the GraphQL enpdoint, and disables it by default.

## Test plan

Git blame for [this file](https://sourcegraph.test:3443/github.com/sourcegraph/sourcegraph/-/blob/client/web/src/cody/search/CodySearchPage.tsx?L122-129&trace=1) now matches what GitHub shows.

### DIFF
--- a/cmd/frontend/graphqlbackend/git_blob.go
+++ b/cmd/frontend/graphqlbackend/git_blob.go
@@ -5,30 +5,33 @@ import (
 
 	"github.com/sourcegraph/sourcegraph/internal/api"
 	"github.com/sourcegraph/sourcegraph/internal/gitserver"
+	"github.com/sourcegraph/sourcegraph/lib/pointers"
 )
 
 func (r *GitTreeEntryResolver) Blame(ctx context.Context,
 	args *struct {
-		StartLine int32
-		EndLine   int32
+		StartLine        int32
+		EndLine          int32
+		IgnoreWhitespace *bool
 	}) ([]*hunkResolver, error) {
 	hunks, err := r.gitserverClient.BlameFile(ctx, r.commit.repoResolver.RepoName(), r.Path(), &gitserver.BlameOptions{
-		NewestCommit: api.CommitID(r.commit.OID()),
-		StartLine:    int(args.StartLine),
-		EndLine:      int(args.EndLine),
+		NewestCommit:     api.CommitID(r.commit.OID()),
+		StartLine:        int(args.StartLine),
+		EndLine:          int(args.EndLine),
+		IgnoreWhitespace: pointers.Deref(args.IgnoreWhitespace, false),
 	})
 	if err != nil {
 		return nil, err
 	}
 
-	var hunksResolver []*hunkResolver
+	hunkResolvers := make([]*hunkResolver, 0, len(hunks))
 	for _, hunk := range hunks {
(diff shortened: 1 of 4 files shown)


cmd/frontend/graphqlbackend/git_blob.go
```diff
@@ -5,30 +5,33 @@ import (
...
-		hunksResolver = append(hunksResolver, &hunkResolver{
+		hunkResolvers = append(hunkResolvers, &hunkResolver{
 			db:   r.db,
 			repo: r.commit.repoResolver,
 			hunk: hunk,
 		})
 	}
 
-	return hunksResolver, nil
+	return hunkResolvers, nil
```
camdencheek, line 36: unrelated, just cleaning up some weird naming and extra allocations

cmd/frontend/graphqlbackend/git_blob.go
```diff
@@ -5,30 +5,33 @@ import (
...
+	"github.com/sourcegraph/sourcegraph/lib/pointers"
 )
 
 func (r *GitTreeEntryResolver) Blame(ctx context.Context,
 	args *struct {
-		StartLine int32
-		EndLine   int32
+		StartLine        int32
+		EndLine          int32
+		IgnoreWhitespace *bool
```
eseliger, line 15: inline structs 🚬

cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -6029,9 +6029,9 @@ type GitBlob implements TreeEntry & File2 {
     """
     externalURLs: [ExternalLink!]!
     """
-    Blame the blob.
+    Blame the blob. If unset, ignoreWhitespace defaults to false.
     """
-    blame(startLine: Int!, endLine: Int!): [Hunk!]!
+    blame(startLine: Int!, endLine: Int!, ignoreWhitespace: Boolean): [Hunk!]!
```
eseliger, line 6034: the GraphQL-esque way would be to provide the default, which would make it non-null, too:

```suggestion
    blame(startLine: Int!, endLine: Int!, ignoreWhitespace: Boolean = false): [Hunk!]!
```

eseliger: also, might be nice to add a quick changelog entry about that

eseliger (approved):

cmd/frontend/graphqlbackend/schema.graphql
camdencheek: Oh! I actually had no idea that GraphQL supported default values 😬 That's a big TIL for the day 😄

camdencheek: good call on the changelog too

fkling:
In case it's relevant: Currently the blame view uses a separate endpoint to get streamed blame information. I guess it uses this function under the hood too. Just mentioning it since you've been changing the GraphQL API.

fkling (approved):

Issue #57834, opened 2023-10-23
Commit view: when lines wrap, they start on a second line


camdencheek:
In the commit view, when a line wraps because it is too long, it starts on a second line, leaving an unnecessary empty line above it. 

https://github.com/sourcegraph/sourcegraph/assets/12631702/86d6f4e2-5d7d-480c-a8ef-17353b5fed48

Galindo-J:
Here is an additional instance of this occurring: 

When the added elements in a yaml array have a value that is long enough to wrap in the UI, the Sourcegraph version of the diff starts the new line visually one space below the new line, and the indentation is thrown off to the left. In the screenshot, line 5 was modified to a longer value. The Sourcegraph diff result shows line 5 with `+` only, and starts the actual value below line 5 visually as well as making the indentation in the same scope as the name of the yaml array.

Sourcegraph View:
<img width="1724" alt="Screenshot 2023-11-06 at 4 21 12 PM" src="https://github.com/sourcegraph/sourcegraph/assets/54120980/1b641f89-b3e5-4230-92c4-434d285179e6">

Gitlab View:
![Screenshot 2023-11-03 at 6 25 31 PM](https://github.com/sourcegraph/sourcegraph/assets/54120980/3f452905-6281-43bf-bb87-fb87074b2242)

PR #58138, opened 2023-11-07
Diff view: fix line wrapping

camdencheek:
This fixes line wrapping behavior in diff view. Previously, if a line didn't fit in the view, it would jump to a new line before properly wrapping. This made it look like the diff added a newline.

The problem was that we're adding a diff marker (the plus/minus) in the gutter, but outside the div with `inline-block`, so the whole div was being pushed to the next line. Instead, if we move the `::before` selector to target the div with the content, the issue goes away. Because we cannot use a data attribute from a parent element in CSS, I also had to move `data-diff-marker` to the div that's being targeted.

Fixes https://github.com/sourcegraph/sourcegraph/issues/57834

## Test plan

See the before in the [issue report](https://github.com/sourcegraph/sourcegraph/issues/57834). Here's the after:

https://github.com/sourcegraph/sourcegraph/assets/12631702/af6e521c-82ed-4f4d-a8d0-6c01d8e96c9c


peterguy (approved):
Nice! Such a simple solution for a nagging problem.

PR #58139, opened 2023-11-07
Web client: fix copyability of text

camdencheek:
This adds the `user-select: none` attribute to text in two places. The first is the `no newline at end of file` notice. It's really easy to accidentally copy that instead of the code, so this just makes it non-selectable. The second is the hidden text describing the types of diff lines. Previously, this would make text copied from diffs unusable, so this makes that text not show up when copied. 

## Test plan

Ensured that not copying the text succeeds in both firefox and safari.

### DIFF
--- a/client/web/src/components/diff/DiffHunk.module.scss
+++ b/client/web/src/components/diff/DiffHunk.module.scss
@@ -46,7 +46,6 @@
 }
 
 .diff-type-description {
-    -webkit-user-select: none;
     user-select: none;
 }
 
--- a/client/web/src/components/diff/DiffHunk.module.scss
+++ b/client/web/src/components/diff/DiffHunk.module.scss
@@ -45,6 +45,11 @@
     }
 }
 
+.diff-type-description {
+    -webkit-user-select: none;
+    user-select: none;
+}
+
 .num {
     background-color: var(--code-bg);
     min-width: 2.5rem;
--- a/client/web/src/repo/blob/codemirror/eof.ts
+++ b/client/web/src/repo/blob/codemirror/eof.ts
@@ -22,7 +22,7 @@ class NoLineBreakWidget extends WidgetType {
 // causes the last line to be hidden.
 const removeLastLineDeco = Decoration.replace({})
 const eofNoteDeco = Decoration.replace({
-    widget: new NoLineBreakWidget('(No new line at end of file)'),
+    widget: new NoLineBreakWidget('(No newline at end of file)'),
     block: true,
 })
 
@@ -47,6 +47,7 @@ export const hideEmptyLastLine: Extension = [
             color: 'var(--text-muted)',
             fontStyle: 'italic',
             marginTop: '.2rem',
+            userSelect: 'none',
         },
     }),
 ]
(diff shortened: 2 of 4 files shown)


client/web/src/components/diff/DiffHunk.module.scss
```diff
@@ -45,6 +45,11 @@
     }
 }
 
+.diff-type-description {
+    -webkit-user-select: none;
+    user-select: none;
```
camdencheek, line 50: Web people: how do I know which of these I need? There are also `-ms-user-select` and `-moz-user-select`. Is there a tool I should be using to determine what's still necessary in 2023?

client/web/src/repo/blob/codemirror/eof.ts
```diff
@@ -22,7 +22,7 @@ class NoLineBreakWidget extends WidgetType {
 // causes the last line to be hidden.
 const removeLastLineDeco = Decoration.replace({})
 const eofNoteDeco = Decoration.replace({
-    widget: new NoLineBreakWidget('(No new line at end of file)'),
+    widget: new NoLineBreakWidget('(No newline at end of file)'),
```
camdencheek, line 25: unrelated, but "newline" is one word in this context

client/web/src/repo/blob/codemirror/eof.ts
```diff
@@ -47,6 +47,8 @@ export const hideEmptyLastLine: Extension = [
             color: 'var(--text-muted)',
             fontStyle: 'italic',
             marginTop: '.2rem',
+            userSelect: 'none',
+            '-webkit-user-select': 'none',
```
camdencheek, line 51: This works in safari, but it still shows the selection pointer when hovered over. This doesn't happen in firefox. Is there something else I need to add to fix that in safari?

client/web/BUILD.bazel
```diff
@@ -2088,6 +2088,10 @@ jest_test(
     ],
     patch_node_fs = False,
     shard_count = 8,  # empirical
+    snapshots = glob(
+        ["**/__snapshots__"],
+        exclude_directories = 0,
+    ),
```
camdencheek, line 2094: cc @sourcegraph/team-dev-infra, this seemed to be necessary to get `bazel run //client/web:test_update_snapshots` to work. Does that seem right?

client/web/src/components/diff/DiffHunk.module.scss
fkling: I believe PostCSS will convert this as necessary. MDN and caniuse are great resources: https://developer.mozilla.org/en-US/docs/Web/CSS/user-select#browser_compatibility, https://caniuse.com/mdn-css_properties_user-select. If the information is correct it looks like Safari still requires the prefix. You could leave it off and inspect the element to see whether it gets added automatically.

client/web/src/repo/blob/codemirror/eof.ts
fkling: You can add `cursor: default` to show a normal cursor.

fkling (approved):

client/web/BUILD.bazel
camdencheek: nvm, we use vitest now

client/web/src/components/diff/DiffHunk.module.scss
camdencheek: aha! Sure enough, looks like `-webkit-user-select` gets added by PostCSS. Thanks for the pointer

PR #58141, opened 2023-11-07
codygateway: Expose allowed models in limiter status endpoint

eseliger:
This adds the set of allowed models to the `GET /v1/limits` API endpoint for feature detection.

## Test plan

Verified locally that the endpoint still works and that model names are returned correctly.


bobheadxi (approved):

abeatrix (approved):
Thank you for the quick turnaround!

PR #58153, opened 2023-11-07
Add gpt-4-turbo and new Fireworks model support

philipp-spiess:
Adds supports for new model identifiers:

- `gpt-4-1106-preview` as the preview for gpt-4-turbo
- `accounts/sourcegraph/models/starcoder-7b` and `accounts/sourcegraph/models/starcoder-16b` as new StarCoder deployments

## Test plan

- Tests green on CI

### DIFF
--- a/internal/licensing/codygateway_test.go
+++ b/internal/licensing/codygateway_test.go
@@ -80,7 +80,7 @@ func TestCodyGatewayCodeRateLimit(t *testing.T) {
 			userCount:   pointers.Ptr(50),
 			licenseTags: []string{GPTLLMAccessTag},
 			want: CodyGatewayRateLimit{
-				AllowedModels:   []string{"anthropic/claude-instant-v1", "anthropic/claude-instant-1", "openai/gpt-3.5-turbo", "openai/gpt-4-1106-preview"},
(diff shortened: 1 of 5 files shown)


internal/licensing/codygateway_test.go
```diff
@@ -80,7 +80,7 @@ func TestCodyGatewayCodeRateLimit(t *testing.T) {
 			userCount:   pointers.Ptr(50),
 			licenseTags: []string{GPTLLMAccessTag},
 			want: CodyGatewayRateLimit{
-				AllowedModels:   []string{"anthropic/claude-instant-v1", "anthropic/claude-instant-1", "openai/gpt-3.5-turbo"},
+				AllowedModels:   []string{"anthropic/claude-instant-v1", "anthropic/claude-instant-1", "openai/gpt-3.5-turbo", "openai/gpt-4-1106-preview"},
```
eseliger, line 83: That passes? Maybe I’m blind but I don’t see where in the diff it’s been added😅

eseliger (approved):

internal/licensing/codygateway_test.go
philipp-spiess: Oh that's my bad 🤦  I will remove it from this list, we need to test the model first before we add it to the gpt tag

PR #51572, opened 2023-05-08
batches: add fork attribute to changeset template

adeola-ak:
closes: #50849

Adding fork attribute to changeset template as a boolean for now
_True_ will force the default forking behavior of the code host" and _false_  will force not forking the repo, overriding what is set in `batchChanges.enforceForks` If nothing is set in the batch spec but the global setting is set, then it will fall back on that behaivor

In the future we can add the ability to specify custom namespaces and fork names 

## Test plan
tested that a spec with the fork attribute set to true, but global enforce forks setting off will create user fork
as well as if the fork attribute in the spec is not set at all, but gloabl enforce forks it set, it will

PR #58156, opened 2023-11-07
code-search: handle changeset fork when creating a batch change via src-cli

BolajiOlajide:
Part 1 of [#58150](https://github.com/sourcegraph/sourcegraph/issues/58150)

#51572 added the `fork` attribute to the changeset template. This allowed per-batch-change control for pushing to a fork of the upstream repository. However, this was only supported by SSBC, and the field was ignored by batch changes created via the `src-cli` tool.

This PR adds the attributes to the `/lib` so that `src-cli` can properly parse the fork attribute whenever it's specified in a batch spec.

## Test plan

* I did a couple of manual tests to ensure this change didn't introduce any regression.
* Added a test for fork support to the `ParseChangesetSpec` function.
* Batch changes created via `src-cli` with the fork attribute now have the proper `fork_namespace` set.

CleanShot 2023-11-07 at 14 08 42@2x

### DIFF
--- a/lib/batches/changeset_spec.go
+++ b/lib/batches/changeset_spec.go
@@ -103,6 +103,7 @@ func (c *ChangesetSpec) MarshalJSON() ([]byte, error) {
 		Body           string                 `json:"body,omitempty"`
 		Commits        []GitCommitDescription `json:"commits,omitempty"`
 		Published      *PublishedValue        `json:"published,omitempty"`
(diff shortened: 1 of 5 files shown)


camdencheek (approved):
Nice, LGTM!

peterguy:
Good job finding this so quickly!

PR #58158, opened 2023-11-07
feat: add custom chat model selection

abeatrix:
feat: add custom chat model selection

Allow dotcom clients to specify a custom chat model. Maintain an allowlist of supported models, and only allow custom models for dotcom clients.

Models can be specified in the request parameters, and will be returned directly if allowed. Otherwise default to the configured chat model.

## Test plan

Set the model field in [CompletionParameters](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/lib/shared/src/sourcegraph-api/completions/types.ts?L26%3A1-35%3A2) to a value that's not on the allowed list will always fallback to using the default model. 

When an allowed model is configured, it should use the correct model accordingly.

### DIFF
--- a/internal/completions/httpapi/chat.go
+++ b/internal/completions/httpapi/chat.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/sourcegraph/log"
 
+	"github.com/sourcegraph/sourcegraph/cmd/frontend/envvar"
 	"github.com/sourcegraph/sourcegraph/internal/completions/types"
 	"github.com/sourcegraph/sourcegraph/internal/conf/conftypes"
 	"github.com/sourcegraph/sourcegraph/internal/database"
@@ -24,6 +25,9 @@ func NewChatCompletionsStreamHandler(logger log.Logger, db database.DB) http.Han
 		rl,
 		"chat",
 		func(requestParams types.CodyCompletionRequestParameters, c *conftypes.CompletionsConfig) (string, error) {
+			if isAllowedCustomChatModel(requestParams.Model) {
+				return requestParams.Model, nil
+			}
 			// No user defined models for now.
 			if requestParams.Fast {
 				return c.FastChatModel, nil
@@ -31,4 +35,24 @@ func NewChatCompletionsStreamHandler(logger log.Logger, db database.DB) http.Han
 			return c.ChatModel, nil
 		},
 	)
+}
+
+// We only allow dotcom clients to select a custom chat model and maintain an allowlist for which
+// custom values we support
+func isAllowedCustomChatModel(model string) bool {
+	if !(envvar.SourcegraphDotComMode()) {
+		return false
+	}
+
+	switch model {
+	case "anthropic/claude-2",
+		"anthropic/claude-v1",
+		"anthropic/claude-instant-1",
+		"openai/gpt-4",
+		"openai/gpt-3.5-turbo",
+		"openai/gpt-4-1106-preview":
+		return true
+	}
+
+	return false
 }


internal/completions/httpapi/chat.go
```diff
@@ -32,3 +36,23 @@ func NewChatCompletionsStreamHandler(logger log.Logger, db database.DB) http.Han
...
+func isAllowedCustomChatModel(model string) bool {
+	if !(envvar.SourcegraphDotComMode()) {
+		return false
+	}
+
+	switch model {
+	case "anthropic/claude-2",
+		"anthropic/claude-v1",
+		"anthropic/claude-instant-1",
+		"openai/gpt-4",
```
philipp-spiess, line 51: Should we maybe wait before adding GPT4 here? It's a lot more expensive than the other providers. cc @rafax @eseliger

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ All notable changes to Sourcegraph are documented in this file.
 
 - The `has.topic` filter now supports filtering by Gitlab topics. [#57649](https://github.com/sourcegraph/sourcegraph/pull/57649)
 - Batch Changes now allows changesets to be exported in CSV and JSON format. [#56721](https://github.com/sourcegraph/sourcegraph/pull/56721)
+- Supports custom ChatCompletion models in Cody clients for dotcom users. [#58158](https://github.com/sourcegraph/sourcegraph/pull/58158)
```
philipp-spiess, line 22: The changelog is more used for on-prem and cloud users and these won't be affected by the change, wdyt?

philipp-spiess (approved):

internal/completions/httpapi/chat.go
abeatrix: Removed!

internal/completions/httpapi/chat.go
```diff
@@ -24,6 +25,9 @@ func NewChatCompletionsStreamHandler(logger log.Logger, db database.DB) http.Han
 		rl,
 		"chat",
 		func(requestParams types.CodyCompletionRequestParameters, c *conftypes.CompletionsConfig) (string, error) {
+			if isAllowedCustomChatModel(requestParams.Model) {
```
eseliger, line 28: For code completions this is currently gated to dotcom. Should we do the same here?

also, once we open this to all instances, I think it would be nice to remove the fast field and instead let the client decide 😬

eseliger (approved):

internal/completions/httpapi/chat.go
philipp-spiess: @eseliger There's a dotcom check inside the `isAllowedCustomChatModel` function 🙈

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ All notable changes to Sourcegraph are documented in this file.
...
+
```
eseliger, line 23:
```suggestion
```

internal/completions/httpapi/chat.go
eseliger: duh, sorry 😬

internal/completions/httpapi/chat.go
```diff
@@ -24,6 +25,9 @@ func NewChatCompletionsStreamHandler(logger log.Logger, db database.DB) http.Han
...
+				return requestParams.Model, nil
+			}
```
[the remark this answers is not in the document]
abeatrix: Makes sense, thanks @philipp-spiess !

CHANGELOG.md
abeatrix: chatted with erik, will keep this entry for record

philipp-spiess: Makes sense :)

PR #58161, opened 2023-11-07
Code Navigation: align EOF message with code mirror content

jasonhawkharris:
Fixes a layout bug that was causing EOF message in the blob view to align with git blame decorations instead of with code mirror content. Before and After video: 

https://github.com/sourcegraph/sourcegraph/assets/62355966/79dd6673-6f05-448d-a24c-7d08ce779fe0

## Test plan
Manual testing on files with and without the EOF decoration

### DIFF
--- a/client/web/src/repo/blob/codemirror/blame-decorations.tsx
+++ b/client/web/src/repo/blob/codemirror/blame-decorations.tsx
@@ -154,6 +154,10 @@ const blameDecorationTheme = EditorView.theme({
         // override default .cm-gutters z-index 200
         zIndex: 201,
     },
+    '.no-line-break-msg': {
+        marginLeft: 'calc(var(--blame-decoration-width))',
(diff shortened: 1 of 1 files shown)


client/web/src/repo/blob/codemirror/blame-decorations.tsx
```diff
@@ -154,6 +154,10 @@ const blameDecorationTheme = EditorView.theme({
         // override default .cm-gutters z-index 200
         zIndex: 201,
     },
+    '.no-line-break-msg': {
+        marginLeft: 'calc(var(--blame-decoration-width))',
+        zIndex: 201,
```
fkling, line 159: Don't think we need `calc` here since you are not doing any calculation. Also unsure whether you need the `zIndex` setting.

fkling (approved):

PR #58162, opened 2023-11-07
sec: improve outgoing webhook url validation

evict:
This improves the URL validation for outgoing webhooks, see: https://github.com/sourcegraph/security-issues/issues/363. It's not too restrictive we only don't allow localhost and the metadata instance. We also allow an additional env var setting for our cloud instances.

## Test plan
Tested locally and with unit tests.


internal/webhooks/outbound/outbound_test.go
```diff
@@ -81,10 +81,10 @@ func TestCheckAddress(t *testing.T) {
...
 			"https://my.webhooks.site/receiver",
 			"https://my.webhooks.site:3000/receiver",
-			"1.2.3.4",
-			"1.2.3.4:2000",
-			"2001:0db8:0000:0000:0000:8a2e:0370:7334",
-			"2001:db8::8a2e:370:7334",
+			"https://1.2.3.4",
+			"http://1.2.3.4:2000",
+			"https://2001:0db8:0000:0000:0000:8a2e:0370:7334",
+			"http://2001:db8::8a2e:370:7334",
```
andreeleuterio, line 87: would be good to add one with a redirect to prevent regressions of the patch

andreeleuterio (approved):

internal/httpcli/client.go
```diff
@@ -379,6 +380,11 @@ func ExternalTransportOpt(cli *http.Client) error {
 		return errors.Wrap(err, "httpcli.ExternalTransportOpt")
 	}
 
+	// default addresses to block
+	denyList.AppendBuiltin("loopback")
+	denyList.AppendPattern("169.254.169.254")
```
peterguy, line 385: Are there other special IP addresses or hostname we should also block for other cloud providers? Like `metadata.google.internal` for GCE?

peterguy, line 385: Should we consider blocking all reserved IP addresses? Or would it be legitimate to have a service listening for a webhook on a reserved IP address?

evict: The 169 one is the metadata one, it's the same for all providers.

evict: I don't want to block too much as it can be an issue for self-hosted customers. But having the environment variable will allow us to do that for Cloud instances.

internal/httpcli/client.go
```diff
@@ -366,9 +367,10 @@ func NewLoggingMiddleware(logger log.Logger) Middleware {
 	}
 }
 
-//
 // Common Opts
-//
+var hostList = env.Get("EXTERNAL_DENY_LIST", "", "Deny list for outgoing requests")
+var denyList = hostmatcher.ParseHostMatchList("", hostList)
+var once sync.Once
```
mrnugget, line 373: Makes it clearer what it's protecting and usually what we do:

```suggestion
var denyListOnce sync.Once
```

internal/httpcli/client.go
```diff
@@ -379,6 +381,13 @@ func ExternalTransportOpt(cli *http.Client) error {
 		return errors.Wrap(err, "httpcli.ExternalTransportOpt")
 	}
 
+	once.Do(func() {
+		denyList.AppendBuiltin("loopback")
+		denyList.AppendPattern("169.254.169.254")
```
mrnugget, line 386: (Continuing our conversation from Slack in here)

```go
Why don’t you have a

```
var defaultDenyList = []string{
}
```

And then add a

```go
func MockDefaultDenyList(t *testing.T, new []string) {
  old := denyList
  t.Cleanup(func() { denyList = old })

  denyList = hostmatcher.ParseHostMatchList("", new)
}
```

Something like this?

internal/httpcli/client.go
```diff
@@ -169,6 +179,18 @@ func newExternalClientFactory(cache bool, middleware ...Middleware) *Factory {
 // downloading arbitrarily sized files! See UncachedExternalClient instead.
 var ExternalDoer, _ = ExternalClientFactory.Doer()
 
+// TestExternalClientFactory is a httpcli.Factory with common options
+// and middleware pre-set for communicating with TestExternalClientFactory services.
+var TestExternalClientFactory = newExternalClientFactory(false, true)
+
+// TestExternalClientFactory is a httpcli.Factory with common options
```
evict, line 186: TODO: fix description

mrnugget (approved):
image

internal/webhooks/outbound/outbound.go
```diff
@@ -55,63 +53,105 @@ func (s *outboundWebhookService) Enqueue(
...
-			return errors.New("Must not be unspecified, private, link-local, or loopback address")
+	if ip != nil {
+		if isIllegalIp(ip, hostAllowList) {
+			return errIllegalAddr
 		}
-
-		return nil
-	} else if uErr != nil {
-		return errors.New("Could not parse address")
 	} else {
```
mrnugget, line 136: I would get rid of the `else` and basically do

```go
if ip != nil {
  if isIllegalIp(...) {
  }
return nil
}

....
```

internal/webhooks/outbound/outbound.go
```diff
@@ -55,63 +53,105 @@ func (s *outboundWebhookService) Enqueue(
...
+			}
 		}
-
-		return nil
 	}
+
+	return nil
+}
+
+func isIllegalIp(ip net.IP, hostAllowList *hostmatcher.HostMatchList) bool {
```
mrnugget, line 154:
```suggestion
func isIllegalIP(ip net.IP, hostAllowList *hostmatcher.HostMatchList) bool {
```

cmd/frontend/graphqlbackend/outbound_webhooks.go
```diff
@@ -136,12 +136,6 @@ func (r *schemaResolver) CreateOutboundWebhook(ctx context.Context, args CreateO
 		return nil, auth.ErrMustBeSiteAdmin
 	}
 
-	// Validate the URL
-	err = outbound.CheckAddress(args.Input.URL)
-	if err != nil {
-		return nil, errors.Wrap(err, "invalid webhook address")
-	}
-
```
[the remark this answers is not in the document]
evict: Correct, this is now handled by the dialer we've created in the client.

cmd/worker/internal/outboundwebhooks/handler.go
```diff
@@ -21,9 +21,10 @@ import (
 )
 
 type handler struct {
-	client   *http.Client
-	store    database.OutboundWebhookStore
-	logStore database.OutboundWebhookLogStore
+	client        *http.Client
+	store         database.OutboundWebhookStore
+	logStore      database.OutboundWebhookLogStore
+	hostMatchList string
```
mrnugget, line 27: Is this unused?

mrnugget (approved):

cmd/frontend/graphqlbackend/outbound_webhooks.go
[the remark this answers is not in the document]
evict: I made a mistake, that is the message I wanted to keep: https://github.com/sourcegraph/sourcegraph/pull/58535.

Issue #54794, opened 2023-07-11
Merge navigation bar and search bar into one component


vovakulikov:
At the moment our navigation and search bar are separate component visually, it's good because usually you can see the most recent query as you navigate to the repository files (because we populate query according to you viewed files) but at the same time it's problematic because it takes an additional space from actual content of the file/blob UI where we already have a lot of sticky elements. 

A main Idea of this issue is to try to merge these two elements into one (you can find a design [here](https://www.figma.com/file/ZrIRl5OchqCrj8d0kghLyV/Yet-another-search-design?type=design&node-id=18-2227&mode=design&t=p14UngcTUeelq9dR-0))

![Image](https://github.com/sourcegraph/sourcegraph/assets/18492575/8bf0592d-5032-4d8f-9ef7-0129a4a615b9)

This would give us more space for the content on the blob page and also gives more straightforward way to jump to the search page from pages like batch changes, code insights, monitoring, etc (basically any other page in the product)

toolmantim:
Nice! I did some [related design work](https://www.figma.com/file/QqXxwyBs0hUyNXivbfBfbh/Cody-Search?type=design&node-id=0-1&mode=design) a while back:

<img width="1190" alt="Cody Search" src="https://github.com/sourcegraph/sourcegraph/assets/153/0b9580c3-97d5-4663-86be-43e9609d50f8">

PR #58164, opened 2023-11-07
[Search UI]: Add new experimental search UI navigation

vovakulikov:
Closes https://github.com/sourcegraph/sourcegraph/issues/54794

## Problem 
The problem this PR is trying to solve is optimising/freeing vertical space (free it for actual content that matters but still manage our global navigation in a way that this is still visible and accessible). Vertical space becomes a bigger problem in places where we have a lot of sticky content on top (like blob UI page or even search result page), users who don't use big external monitors can't use Sourcegraph efficiently. 

## Solution
The solution is to merge search box UI and global navigation UI, similar how our code search product have it (see Google Code Search and Github navigation). 

Screenshot 2023-11-07 at 13 51 05 Screenshot 2023-11-07 at 13 51 07 

Screenshot 2023-11-07 at 13 51 18 Screenshot 2023-11-07 at 13 51 20

Note: that we still not sure about some decisions about this navigation, this might affect our feature visibility in unexpected way, so on pages where search isn't related directly (like code insights, batch changes we still show old global navigation UI) 

## Availability 
New navigation UI is under experimental feature flag `newSearchNavigationUI`.

## Other improvements 
- Fixed compact/full navigation transition breakpoint (now navigation switches only and only when there is no enough space for navigation items, previously it was static break point which didn't work properly for generic content)
- Search UI box now has plain query input fallback UI to show when we load search box UI component (it should help with UI thrashing) 

## Further steps 
It's in experimental stage, so implementation may lack some things in layout like Mobile layout might be weird, we will improve it in a follow up PR if this UI will work out.  

## Test plan
- First of all check old navigation, it should be no regression on mobile and desktop versions
- Turn on experimental navigation with experimentalFeatures: `newSearchNavigationUI`
- Check that you can go through search core flow with new navigation.

### DIFF
--- a/client/web/src/LegacyLayout.tsx
+++ b/client/web/src/LegacyLayout.tsx
@@ -222,6 +222,15 @@ export const LegacyLayout: FC<LegacyLayoutProps> = props => {
         return <Navigate to={PageRoutes.PostSignUp} replace={true} />
     }
 
+    const showNavigationSearchBox =
+        isSearchRelatedPage &&
+        !isSearchHomepage &&
+        !isCommunitySearchContextPage &&
+        !isSearchConsolePage &&
+        !isSearchNotebooksPage &&
+        !isCodySearchPage &&
+        !isSearchJobsPage
+
     return (
         <div
             className={classNames(
@@ -261,15 +270,7 @@ export const LegacyLayout: FC<LegacyLayoutProps> = props => {
                 <>
                     {newSearchNavigation ? (
                         <NewGlobalNavigationBar
-                            showSearchBox={
-                                isSearchRelatedPage &&
-                                !isSearchHomepage &&
-                                !isCommunitySearchContextPage &&
-                                !isSearchConsolePage &&
-                                !isSearchNotebooksPage &&
-                                !isCodySearchPage &&
-                                !isSearchJobsPage
-                            }
+                            showSearchBox={showNavigationSearchBox}
                             authenticatedUser={props.authenticatedUser}
                             isSourcegraphDotCom={props.isSourcegraphDotCom}
                             ownEnabled={props.ownEnabled}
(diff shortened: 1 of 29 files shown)


client/web/src/LegacyLayout.tsx
```diff
@@ -256,22 +258,48 @@ export const LegacyLayout: FC<LegacyLayoutProps> = props => {
...
+                        <NewGlobalNavigationBar
+                            showSearchBox={
+                                isSearchRelatedPage &&
+                                !isSearchHomepage &&
+                                !isCommunitySearchContextPage &&
+                                !isSearchConsolePage &&
+                                !isSearchNotebooksPage &&
+                                !isCodySearchPage &&
+                                !isSearchJobsPage
+                            }
```
fkling, line 272: This can be moved out into a separate variable.

client/web/src/nav/new-global-navigation/NewGlobalNavigationBar.tsx
```diff
@@ -0,0 +1,435 @@
...
+                    submitSearch={submitSearchOnChange}
+                />
+            </LazyV2SearchInput>
+
+            {isFocused && <div className={styles.overlay} />}
+        </>
+    )
+}
+
+interface SighInUpButtonsProps {
```
fkling, line 250: Typo
```suggestion
interface SignInUpButtonsProps {
```

client/web/src/nav/new-global-navigation/NewGlobalNavigationBar.tsx
```diff
@@ -0,0 +1,435 @@
...
+    isSourcegraphDotCom: boolean
+}
+
+const SighInUpButtons: FC<SighInUpButtonsProps> = props => {
```
fkling, line 254: Typo
```suggestion
const SignInUpButtons: FC<SignInUpButtonsProps> = props => {
```

fkling (approved):
LGTM

PR #58169, opened 2023-11-07
search: Refactor new search input

fkling:
This refactors some parts of the new search input, especially around the suggestions UI. There is no functionality change.

- Remove React dependencies from suggestions: 
  - Instead of having a view plugin create a React root and render the component in there, the component is now rendered directly in the component itself. The component is listening to CodeMirror suggestion state changes and updated its own state accordingly. 
  - Custom render logic was moved into the suggestions component itself.
- Remove legacy/deprecated dependencies, such as `platformContext`, and unused properties.
- Directly use GraphQL API to query for search contexts.

## Test plan

Enter a query into main search input and verify that suggestions are rendered correctly:
- Syntax highlighting for filters, filter+values and queries
- Substring matches are emphasized
- Value completion suggestions are shown for the `context:` filter


peterguy (approved):
Removing React dependencies and using CodeMirror state correctly - great cleanup!

PR #58187, opened 2023-11-08
Temporarily log costly queries

evict:
We want to log query violations, as this could indicate client code that uses the GraphQL API in a non-standard way. 

## Test plan
CI tests.

### DIFF
--- a/cmd/frontend/internal/httpapi/graphql.go
+++ b/cmd/frontend/internal/httpapi/graphql.go
@@ -4,7 +4,6 @@ import (
 	"compress/gzip"
 	"context"
 	"encoding/json"
-	"fmt"
 	"net/http"
 	"strconv"
 	"strings"
--- a/cmd/frontend/internal/httpapi/graphql.go
+++ b/cmd/frontend/internal/httpapi/graphql.go
@@ -4,6 +4,7 @@ import (
 	"compress/gzip"
 	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


cmd/frontend/internal/httpapi/graphql.go
```diff
@@ -125,6 +127,9 @@ func serveGraphQL(logger log.Logger, schema *graphql.Schema, rlw graphqlbackend.
 			}
 
 			if !isInternal && (cost.FieldCount > graphqlbackend.MaxFieldCount) {
+				if envvar.SourcegraphDotComMode() { // temporarily logging queries that exceed field count limit on Sourcegraph.com
+					logger.Warn(fmt.Sprintf("GQL cost limit exceeded. Query: (%s), Vars: (%s)", params.Query, params.Variables))
```
eseliger, line 131: The idiomatic way of logging parameters is to use this API, then we can also do proper JSON magic on the field values in GCP. 

```suggestion
					logger.Warn("GQL cost limit exceeded", log.String("query", params.Query))
```

In the suggestion I left out variables because it could contain secrets. I'm not sure if that's a problem.

cmd/frontend/internal/httpapi/graphql.go
```diff
@@ -4,6 +4,7 @@ import (
 	"compress/gzip"
 	"context"
 	"encoding/json"
+	"fmt"
```
eseliger, line 7: I think it'll yell at you for the unused import now 

```suggestion
```

eseliger (approved):

PR #58206, opened 2023-11-08
Code Navigation: cleans up blob view search

jasonhawkharris:
This PR cleans up the blob view search bar, which was obfuscated in a number of ways, especially in dark mode. 

Before: 
Screenshot 2023-11-08 at 11 10 34 AM
Screenshot 2023-11-08 at 11 11 09 AM

After: 
Screenshot 2023-11-08 at 11 11 27 AM
Screenshot 2023-11-08 at 11 11 57 AM

Some other changes: 
- Remove superfluous 'previous' and 'next' labels since carat up and down will suffice. 
- previous, next buttons, and results will show up dynamically based on matches. Matches must be greater than 1 for next and previous buttons to become visible.
Screenshot 2023-11-08 at 11 15 28 AM
Screenshot 2023-11-08 at 11 16 01 AM
Screenshot 2023-11-08 at 11 16 26 AM
- Remove superfluous label describing command on right side of search container.
- Lighten any items that were not easily visible in dark mode.
- Fix strange looking key icons

## Test plan
Manual testing

### DIFF
--- a/client/web/src/repo/blob/codemirror/search.tsx
+++ b/client/web/src/repo/blob/codemirror/search.tsx
@@ -222,21 +222,20 @@ class SearchPanel implements Panel {
                     {
                         /* commenting regex filter out since it is breaking everything
                          * TODO: fix it
-                         *
-                         * <QueryInputToggle
-                        //     isActive={searchQuery.regexp}
-                        //     onToggle={() => this.commit({ regexp: !searchQuery.regexp })}
-                        //     iconSvgPath={mdiRegex}
-                        //     title="Regular expression"
-                        //     className="cm-search-toggle test-blob-view-search-regexp"
-                        />*/
+                         */
                     }
+                    <QueryInputToggle
+                        isActive={searchQuery.regexp}
+                        onToggle={() => this.commit({ regexp: !searchQuery.regexp })}
+                        iconSvgPath={mdiRegex}
+                        title="Regular expression"
+                        className="cm-search-toggle test-blob-view-search-regexp"
+                    />
                 </div>
                 {totalMatches > 1 && (
                     <>
-
                         <Button
-                            className="mr-2"
+                            className="p-1 ml-2 mr-2 mt-0 mb-0"
                             type="button"
                             size="sm"
                             outline={true}
@@ -248,7 +247,7 @@ class SearchPanel implements Panel {
                         </Button>
 
                         <Button
-                            className="mr-3"
+                            className="p-1 mr-0 mt-0 mb-0"
                             type="button"
                             size="sm"
                             outline={true}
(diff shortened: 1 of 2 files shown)


client/web/src/repo/blob/codemirror/search.tsx
```diff
@@ -216,45 +216,53 @@ class SearchPanel implements Panel {
...
-                    variant="secondary"
-                    onClick={this.findNext}
-                    data-testid="blob-view-search-next"
-                >
-                    <Icon svgPath={mdiChevronDown} aria-hidden={true} />
-                    Next
-                </Button>
+                {totalMatches > 1 && (
+                    <>
+                        <Button
```
camdencheek, line 237: now that the button doesn't have text, should we add an aria attribute for accessibility?

jasonhawkharris: Yes. Good catch. 👍🏼

taiyab (commented):
@jasonhawkharris Some really nice changes here. Good stuff!

LGTM!

vovakulikov (approved):
Tech changes ok, UX suggestions (cc @taiyab)
- I think at the moment there is no controls to close this panel, we probably want to add some close panel button there
- Removing buttons when there is only one change is probably ok but other search panels always have them (Safari and Chrome) 
- Personal opinion (super minor) use of instead of '/' so it would be 1 of 4 matches

p.s. I really like how Safari search panel  looks like, very easy to use 
Screenshot 2023-11-08 at 14 45 20

jasonhawkharris:
@vovakulikov I can make a separate issue to add closing capabilities. This is really just to clean up what's currently there, which is not great, at the moment. I'll likely work on that and some other bugs right after I merge this.

jasonhawkharris:
> * Removing buttons when there is only one change is probably ok but other search panels always have them (Safari and Chrome)

I could add a conditional to gray them out if search results aren't greater than 1, but this also seems like adding more complexity than less. If you can't use them, why not just leave them out in the first place?

client/web/src/repo/blob/codemirror/search.tsx
```diff
@@ -448,6 +456,14 @@ const theme = EditorView.theme({
...
     },
     '.cm-sg-search-info': {
+        color: 'var(--body-color)',
+    },
+    '.cm-search-results': {
+        color: 'var(--gray-06)',
+        fontFamily: 'var(--font-family-base)',
+        marginLeft: '2rem',
+    },
+    '.cm-search-toggle': {
```
fkling, line 466: Nit: I'd prefix our custom classes with `sg-`, not `cm-`. This makes it clearer what are custom styles and what are things we override. I know we are currently doing this inconsistently.

Issue #57996, opened 2023-10-30
Notebooks: remove notepad


camdencheek:
The notepad feature currently contributes to some global state stuff and makes it harder to isolate the notebooks feature from the rest of the code search product. Its removal does not affect any notebooks that already exist, and new notebooks can still be created manually.

PR #58217, opened 2023-11-08
[Notebooks]: Remove notepad UI from notebooks feature

vovakulikov:
Closes https://github.com/sourcegraph/sourcegraph/issues/57996

## Test plan
- Check that notebooks still works as expected
- Check that search flow has no regressions

## Preview 🤩
[Preview Link](https://docs.sourcegraph.com/@vk/remove-notepad)


vovakulikov:
And all these code logic we had in the initial main chunk (so all users, whenever they use notepad or not load this logic to their clients)

camdencheek (approved):

PR #58218, opened 2023-11-08
Code Navigation: fix borked regex filter

jasonhawkharris:
Before, the regex filter on blobview search was crashing the application when selected with an undefined or empty value. 
This commit adds conditionals that check for valid regular expressions so that the feature is now useable, and no longer crashing the application.

There are still other improvements that need to be made regarding the regex filter, but this commit makes the core feature useable.

## Test plan
Manual testing

### DIFF
--- a/client/web/src/repo/blob/codemirror/search.tsx
+++ b/client/web/src/repo/blob/codemirror/search.tsx
@@ -217,7 +217,7 @@ class SearchPanel implements Panel {
                     />
                     <QueryInputToggle
                         isActive={searchQuery.regexp}
-                        onToggle={() => this.commit({ regexp: !searchQuery.regexp, })}
+                        onToggle={() => this.commit({ regexp: !searchQuery.regexp })}
                         iconSvgPath={mdiRegex}
                         title="Regular expression"
                         className="cm-search-toggle test-blob-view-search-regexp"
@@ -342,16 +342,11 @@ class SearchPanel implements Panel {
                     topLineBlock = this.view.lineBlockAtHeight(scrollTop + topLineBlock.height)
                 }
 
-                // conditional to check regex
-                if (query.regexp) {
-                    if (!query.valid) {
-                        console.error("invalid regular expression: ", query.search)
-                        return
-                    }
+                if (query.regexp && !query.valid) {
+                    return
                 }
 
                 let result = query.getCursor(this.view.state.doc, topLineBlock.from).next()
-
                 if (result.done) {
                     // No match in the remainder of the document, wrap around
                     result = query.getCursor(this.view.state.doc).next()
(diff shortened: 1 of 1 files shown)


client/web/src/repo/blob/codemirror/search.tsx
```diff
@@ -375,13 +375,29 @@ class SearchPanel implements Panel {
...
-    while (!result.done) {
+    // Regular expressions that result in matches with length 0 would
+    // cause an infinite loop. So we guard against that by verifying
+    // whether or not the cursor moves to the next match at a new position.
+    let prevValue: { from: number; to: number } | null = null
+
+    while (!result.done && result.value.from !== prevValue?.from) {
         newSearchMatches.set(result.value.from, index++)
+        prevValue = result.value
         result = query.getCursor(document, result.value.to).next()
```
camdencheek, line 393: Why are we recreating the iterator every time? Looks like `query.getCursor` returns an iterator and we can just repeatdly call next on that iterator rather than recreating it every time

jasonhawkharris: This is @fkling's solution to an issue we were having with the way codemirror calls `next()`. He will have more insight into why. In any case, may be a good place for a comment.

fkling: You are right, maybe the core problem is recreating the cursor. That line of code already existed. Worth looking into what happens if we persist the cursor.

vovakulikov: Agree I think we can reuse this iterator from getCursor without re-creating it within the loop
```ts
function calculateMatches(query: SearchQuery, document: CodeMirrorText): SearchMatches {
    const newSearchMatches: SearchMatches = new Map()

    if (!query.valid) {
        return newSearchMatches
    }

    let index = 1
    // Regular expressions that result in matches with length 0 would
    // cause an infinite loop. So we guard against that by verifying
    // whether the cursor moves to the next match at a new position.
    let prevValue: { from: number; to: number } | null = null
    const matches = query.getCursor(document)
    let result = matches.next()

    while (!result.done && result.value.from !== prevValue?.from) {
        newSearchMatches.set(result.value.from, index++)
        prevValue = result.value
        result = matches.next()
    }

    // If the result is not done, it detected an infinite loop, so we
    // do not have any matches.
    if (!result.done) {
        return new Map()
    }

    return newSearchMatches
}
``` 

cc @fkling curios why `getCursor` doesn't return `Iterable` type but `Iterator`, means we can't use `for of` loop here 🤔

vovakulikov (approved):
LGTM, left one comment about iterator (which should be easy to fix) approve in advance, thanks!

client/web/src/repo/blob/codemirror/search.tsx
vovakulikov: FYI I tested reusing cursor version and it did work out for me

vovakulikov:
Ah, I didn't realise it had auto merge label, sorry.

vovakulikov:
Made a PR with quick fix for iterator https://github.com/sourcegraph/sourcegraph/pull/58240

PR #58221, opened 2023-11-09
bzl: Remove library target from Rust crate

varungandhi-src:
Removes the extra `lib.rs` file and extra library target specific to Bazel.

I think the main "unusual" bit is the `mod foo;` declarations. Previously, those
were in `lib.rs`, but they can be added to `main.rs` directly.

The `mod X` declaration says "I have a module named X", so make sure
to look for it. If it is `mod X { ... }` then the body is right there. If it is
`mod X;`, then the lookup is done based on file system paths from the current
module. However, these `mod X;` declarations can live in any file, not just `lib.rs`.

## Test plan

`bazel build //docker-images/syntax-highlighter/crates/...` works fine except
for clippy errors


keynmol (approved):
Nice!

PR #58240, opened 2023-11-09
[Blob UI]: Re-use cursor iterator in file match search calculations

vovakulikov:
Follow up for https://github.com/sourcegraph/sourcegraph/pull/58218#discussion_r1388343328

## Test plan
- Check that regex document search still works properly

### DIFF
--- a/client/web/src/repo/blob/codemirror/search.tsx
+++ b/client/web/src/repo/blob/codemirror/search.tsx
@@ -420,23 +420,12 @@ function calculateMatches(query: SearchQuery, document: CodeMirrorText): SearchM
     }
 
     let index = 1
-    // Regular expressions that result in matches with length 0 would
-    // cause an infinite loop. So we guard against that by verifying
-    // whether the cursor moves to the next match at a new position.
-    let prevValue: { from: number; to: number } | null = null
     const matches = query.getCursor(document)
     let result = matches.next()
 
-    while (!result.done && result.value.from !== prevValue?.from) {
+    while (!result.done) {
         newSearchMatches.set(result.value.from, index++)
-        prevValue = result.value
         result = matches.next()
-    }
-
-    // If the result is not done, it detected an infinite loop, so we
-    // do not have any matches.
-    if (!result.done) {
-        return new Map()
     }
 
     return newSearchMatches
--- a/client/web/src/repo/blob/codemirror/search.tsx
+++ b/client/web/src/repo/blob/codemirror/search.tsx
@@ -413,22 +413,24 @@ class SearchPanel implements Panel {
 }
 
 function calculateMatches(query: SearchQuery, document: CodeMirrorText): SearchMatches {
+    const newSearchMatches: SearchMatches = new Map()
+
     if (!query.valid) {
-        return new Map()
+        return newSearchMatches
     }
 
-    const newSearchMatches: SearchMatches = new Map()
     let index = 1
-    let result = query.getCursor(document).next()
     // Regular expressions that result in matches with length 0 would
     // cause an infinite loop. So we guard against that by verifying
-    // whether or not the cursor moves to the next match at a new position.
+    // whether the cursor moves to the next match at a new position.
     let prevValue: { from: number; to: number } | null = null
+    const matches = query.getCursor(document)
+    let result = matches.next()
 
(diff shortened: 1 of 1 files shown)


client/web/src/repo/blob/codemirror/search.tsx
```diff
@@ -375,29 +375,32 @@ class SearchPanel implements Panel {
...
-    let result = query.getCursor(document).next()
     // Regular expressions that result in matches with length 0 would
     // cause an infinite loop. So we guard against that by verifying
-    // whether or not the cursor moves to the next match at a new position.
+    // whether the cursor moves to the next match at a new position.
     let prevValue: { from: number; to: number } | null = null
+    const matches = query.getCursor(document)
+    let result = matches.next()
 
     while (!result.done && result.value.from !== prevValue?.from) {
```
camdencheek, line 392: do we still need this extra loop logic now?

jasonhawkharris: Also, we're sure this is still working for regexp like `^` and `$`? I just want to make sure we test those strings specifically.

jasonhawkharris: By `^` and `$`, I mean literally those regex patterns. Not regex patterns that contain those characters, but those characters alone.

fkling: yeah I can totally see that we have the infinite loop problem because we were recreating the cursor at the same position. If we reuse the cursor maybe the problem goes away. Still need to check what happens with valid, zero-width patterns.

client/web/src/repo/blob/codemirror/search.tsx
```diff
@@ -375,29 +375,32 @@ class SearchPanel implements Panel {
...
-        result = query.getCursor(document, result.value.to).next()
+        result = matches.next()
     }
 
     // If the result is not done, it detected an infinite loop, so we
     // do not have any matches.
     if (!result.done) {
         return new Map()
     }
+
```
jasonhawkharris, line 403: For some reason this is no longer working with your changes. We added this conditional so that it would show zero results for searches like `^` and `$`. See below, it shows 385.
Screenshot 2023-11-09 at 12 39 27 PM

jasonhawkharris: Should say 0 results

camdencheek: Is that actually incorrect? I'd expect `^` to show one result per line

vovakulikov: Yeah, I think it counts just lines in this case, curios why this worked previously

client/web/src/repo/blob/codemirror/search.tsx
vovakulikov: I removed the checks from the loop and it seems working just fine but there is a problem with `^` regex count number

client/web/src/repo/blob/codemirror/search.tsx
fkling: I guess it's up to us to decide how these "weird" regular expressions should behave. I didn't actually consider that `^` and `$` would anchor to the whole line, but it makes sense. If it doesn't cause any issues anymore I'd show the actual match count.

vovakulikov: Ok, I added some check to skip empty matches, should be ok now

fkling: What happens to the next/previous buttons in this case though? Are they actionable?

jasonhawkharris: There has to be more than 1 match in order for the next/previous buttons to appear.

jasonhawkharris (approved):

client/web/src/repo/blob/codemirror/search.tsx
jasonhawkharris: next/previous will only appear if there is more than 1 result.

fkling: OK. I wasn't sure whether that is determined from the return value of this function (I now saw that it is) or not.

fkling (approved):

PR #58242, opened 2023-11-09
Add Protocol Check for `getReturnTo` method 

shivasurya:
Resolves: https://github.com/sourcegraph/security-issues/issues/362

For more context, check the above issue.

## Test plan

- CI 🟢 
- Redirects without any issue and defaults to "/search" when the check fails
- Tested with username, passwd auth
- redirects perfectly to returnTo param
- redirects to /search when custom schemes are used


client/web/src/auth/SignInSignUpCommon.tsx
```diff
@@ -75,5 +75,10 @@ export function getReturnTo(location: H.Location, defaultReturnTo: string = Page
     const returnTo = searchParameters.get('returnTo') || defaultReturnTo
     const newURL = new URL(returnTo, window.location.href)
 
-    return newURL.pathname + newURL.search + newURL.hash
```
evict, line 78: Did the XSS get through here because the protocol would be `x:`?

client/web/src/auth/SignInSignUpCommon.tsx
```diff
@@ -75,5 +75,10 @@ export function getReturnTo(location: H.Location, defaultReturnTo: string = Page
...
+    // 🚨 SECURITY: check newURL scheme http or https
+    if (newURL.protocol == 'http:' || newURL.protocol == 'https:') {
+        return newURL.pathname + newURL.search + newURL.hash
```
evict, line 80: It seems to me that `http:javascript:alert(document.domain)` might still work. Did you test that?

evict (changes requested):
See comment: https://github.com/sourcegraph/security-issues/issues/362#issuecomment-1805469880.

evict (approved):

PR #58249, opened 2023-11-09
Fix panic for empty blobs

camdencheek:
When `len(in) == 0`, this code panics because `in` will be indexed with a negative value. 

## Test plan

The file I was trying to view no longer panics.

### DIFF
--- a/cmd/frontend/graphqlbackend/git_tree_entry.go
+++ b/cmd/frontend/graphqlbackend/git_tree_entry.go
@@ -181,7 +181,7 @@ func pageContent(content []byte, startLine, endLine *int) []byte {
 func lineCount(in []byte) int {
 	c := bytes.Count(in, []byte("\n"))
 	// Final newline doesn't mark a new line.
-	if in[len(in)-1] != '\n' {
+	if len(in) > 0 && in[len(in)-1] != '\n' {
 		return c + 1
 	}
 	return c


unknwon (approved):

cmd/frontend/graphqlbackend/git_tree_entry.go
```diff
@@ -181,7 +181,7 @@ func pageContent(content []byte, startLine, endLine *int) []byte {
 func lineCount(in []byte) int {
 	c := bytes.Count(in, []byte("\n"))
 	// Final newline doesn't mark a new line.
-	if in[len(in)-1] != '\n' {
+	if len(in) > 0 && in[len(in)-1] != '\n' {
 		return c + 1
 	}
```
eseliger, line 186: could we be even cheaper here and do:

```
if len(in) == 0 { return 0 }
```

?

eseliger (approved):

cmd/frontend/graphqlbackend/git_tree_entry.go
camdencheek: I mean, if `len(in) == 0`, `bytes.Count` is going to be stupid cheap anyways. And given that, I'd personally prefer to keep the safety check alongside the condition that requires it (rather than moving it to the top of the function.

PR #58251, opened 2023-11-10
dev/msp: do not mutate package variable

bobheadxi:
The iam stack was mutating the shared `serviceAccountRoles` variable, which affects the output of other services.

## Test plan

```
sg msp generate -all
```

### DIFF
--- a/dev/managedservicesplatform/internal/stack/iam/iam.go
+++ b/dev/managedservicesplatform/internal/stack/iam/iam.go
@@ -32,9 +32,17 @@ type Variables struct {
 
 const StackName = "iam"
 
-var (
+func NewStack(stacks *stack.Set, vars Variables) (*CrossStackOutput, error) {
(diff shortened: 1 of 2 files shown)


dev/managedservicesplatform/internal/stack/iam/iam.go
```diff
@@ -32,9 +32,17 @@ type Variables struct {
...
+		googleprovider.With(vars.ProjectID))
+	if err != nil {
+		return nil, err
+	}
+
+	id := resourceid.New("iam")
+
 	// serviceAccountRoles are granted to the service account for the Cloud Run service.
-	serviceAccountRoles = []serviceaccount.Role{
+	serviceAccountRoles := []serviceaccount.Role{
```
bobheadxi, line 45: was original package-level, now scoped to the function

michaellzc (approved):

PR #58255, opened 2023-11-10
cleanup: Make evaluator code more idiomatic

varungandhi-src:
Major changes:

1. Use proper iteration structures like `into_iter()`/`iter()` in a bunch of places to
    avoid cloning `HashMap`/`Vec` structures (as those will be deep clones).
2. Uses string interning to avoid cloning `String`s and avoiding complicating
    lifetimes by passing `&str` everywhere. This allows marking all the interesting
    types as `Copy` so that `*` can perform a copy with needing `clone()`.
3. Introduces newtype wrappers to avoid confusing different kinds of symbols.
4. Uses more methods instead of free functions for better auto-complete and
    and code organization.

## Test plan

n/a


varungandhi-src:
I'm just going to go ahead and merge this @keynmol, but please feel free to ask questions for things that are unclear.

PR #58258, opened 2023-11-10
code intel: Don't rely on URL polyfill to correctly parse git: URIs

fkling:
This was a strange problem: I kept running into problems fetching code intel data in the Svelte prototype and it turned out that there was an issue parsing `git:` URIs. 
In the React app, when you pass a `git:` URI to `new URL`, the following is returned (showing only the relevant parts:

```js
// In:
new URL("git://github.com/microsoft/vscode?dbd76d987cf1a412401bdbd3fb785217ac94197e#src/vs/css.js")
// Out:
{
  hash: "#src/vs/css.js",
​  ​host: "github.com",
​​  hostname: "github.com",
​​  href: "git://github.com/microsoft/vscode?dbd76d987cf1a412401bdbd3fb785217ac94197e#src/vs/css.js",
​​  origin: "null",
​​  password: "",
​​  pathname: "/microsoft/vscode",
​​  port: "",
​​  protocol: "git:",
​​  search: "?dbd76d987cf1a412401bdbd3fb785217ac94197e",
}
```

On other sites, and the SvelteKit prototype, this was the result. Note how `host` and `pathname` are different:

```js
{
  hash: "#src/vs/css.js",
  host: "",
  hostname: "",
  href: "git://github.com/microsoft/vscode?dbd76d987cf1a412401bdbd3fb785217ac94197e#src/vs/css.js",
  origin: "null",
  password: "",
  pathname: "//github.com/microsoft/vscode",
  port: "",
  protocol: "git:",
  search: "?dbd76d987cf1a412401bdbd3fb785217ac94197e",
}
```

The return value is also displayed differently in the console. I assume something (maybe React Router or a custom bundler replacement rule) overwrites the native `URL` constructor with something that can parse `git:` URIs the way we expect it.

**EDIT:** It seems we are explicitly enabling polyfills for `URL` and `URLSearchParams` in [this file](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@2108b6250948ed072f87766a3775613c65f4e37c/-/blob/client/shared/src/polyfills/configure-core-js.ts?L5-7). It seems reasonable to do this, somehow it makes me uncomfortable to (implicitly) rely on polyfill to achieve the correct behavior. It makes things more difficult to debug and potentially difficult to share/migrate this logic.

~It looks like simply replacing `git:` with `http:` makes it work with the standard `URL` constructor.~

There are subtle differences between how a `git:` URI should be parsed and how a `http:` URI is passed. E.g. if the path is empty in `git:`, `pathname` will also be empty. When parsing as `http:` it will be `/`. 
To avoid any problems I decided to use a spec compliant URI parser in `parseRepoURI` and changed the implementation of `parseGitURI` to use that function (the structure of the return values are different and I didn't want to update all callsites of `parseRepoURI` too).
This also revealed that we passed some URIs to `parseRepoURI` that should not have worked.

## Test plan

Existing unit tests


varungandhi-src:
> I assume something (maybe React Router or a custom bundler replacement rule) overwrites the native URL constructor with something that can parse git: URIs the way we expect it.

Yikes, that seems bad/scary if there's something overriding a standard constructor (even if it's wrong...). 😬

varungandhi-src (approved):

fkling:
Yeah. I looked again just now and it seems to be a polyfill coming from core-js. And it seems we explicitly set it though I'm not sure the comment is correct. As I understand the spec it works as expected: https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@2108b6250948ed072f87766a3775613c65f4e37c/-/blob/client/shared/src/polyfills/configure-core-js.ts?L5-7

BolajiOlajide (approved):

PR #58260, opened 2023-11-10
search: add pattern type for new keyword search

stefanhengl:
This adds a new pattern type "newStandardRC1" which replaces "standard" if the feature flag "search-new-keyword" is set. Instead of calling Zoekt's parser, we concatenate literals with an `AND` operator, which should automatically provide 100% compatibility with our current search syntax. Previously we didn't support Sourcegraph specific parts of the query language, such as inline regexp, in our prototype.

To switch back and forth between "standard" and "newStandardRC1", I adapted the smart search selector like this (only visible if "search-new-keyword" is set) 

image

## Test plan
- I copied many of the tests for "standard" and adapted them for "newStandardRC1". Check out the new golden files and compare them to the corresponding golden files for "standard".
- manual testing
  -  tried out queries, such as `r:zoekt query parse`, and verified the results
  - mixed literals with regexp and verified that the results make sense.


stefanhengl:
@fkling tagging you for the client changes.

internal/search/job/jobutil/job.go
```diff
@@ -840,7 +840,7 @@ func (b *jobBuilder) newZoektSearch(typ search.IndexedRequestType) (job.Job, err
 		FileMatchLimit: b.fileMatchLimit,
 		Select:         b.selector,
 		Features:       *b.features,
-		KeywordScoring: b.patternType == query.SearchTypeKeyword,
+		PatternType:    b.patternType,
```
keegancsmith, line 843: intentional to remove KeywordScoring?

internal/search/client/client.go
```diff
@@ -231,6 +231,8 @@ func SearchTypeFromString(patternType string) (query.SearchType, error) {
 		return query.SearchTypeLucky, nil
 	case "keyword":
 		return query.SearchTypeKeyword, nil
+	case "standardv2":
```
keegancsmith, line 234: I'm wondering if while this is experimental we should include that in the wording. Less surpsing then when we rename/remove it.

internal/search/client/client.go
```diff
@@ -252,8 +254,10 @@ func detectSearchType(version string, patternType *string) (query.SearchType, er
 			searchType = query.SearchTypeLiteral
 		case "V3":
 			searchType = query.SearchTypeStandard
+		case "V4":
```
keegancsmith, line 257: does version have to look like v{num}. Could we do something like V4-rc1 or V4-exp or something?

keegancsmith (approved):
backend LGTM. Main feedback is using an experimental marker in what we marshal over the wire for search type and version.

internal/search/client/client.go
stefanhengl: Sure, that makes sense to me.

camdencheek (approved):
LGTM! Exciting stuff.

Unrelated to this PR, but commenting here so I don't forget: do we have plans to push down and/or support to `searcher`? If we make this search mode default, it will dramatically increase the number of AND queries, which are slow and unstable for unindexed search because searcher cannot handle them natively, so we do intersection/union in `frontend`.

stefanhengl:
> LGTM! Exciting stuff.
> 
> Unrelated to this PR, but commenting here so I don't forget: do we have plans to push down and/or support to `searcher`? If we make this search mode default, it will dramatically increase the number of AND queries, which are slow and unstable for unindexed search because searcher cannot handle them natively, so we do intersection/union in `frontend`.

Good point. We definitely have to look into that. I will put it on our board. Thanks for commenting here.

PR #58262, opened 2023-11-10
embeddings: fix default size limit

stefanhengl:
This fixes a bug where the default size limit for files was set to 0 if `FileFilters`  were defined.

Until this fix is released, a workaround is to always define `MaxFileSizeBytes` if `FileFilters` are defined.

Test plan:
- new unit test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ All notable changes to Sourcegraph are documented in this file.
 - Fixed a bug where typing in the GraphQL editor in the Site Admin API console could cause the cursor to jump to the start of the editor. [#57862](https://github.com/sourcegraph/sourcegraph/pull/57862)
 - The blame column no longer ignores whitespace-only changes by default. [#58134](https://github.com/sourcegraph/sourcegraph/pull/58134)
 - Long lines now wrap correctly in the diff view. [#58138](https://github.com/sourcegraph/sourcegraph/pull/58138)
+- Defining file filters for embeddings jobs no longer causes all files to be skipped if `MaxFileSizeBytes` isn't defined. [#58262](https://github.com/sourcegraph/sourcegraph/pull/58262)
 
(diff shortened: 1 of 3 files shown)


jtibshirani (approved):
Looks good to me. Could you explain what the outcome of the bug would look like (in case others run into this problem and want to know if it fixes their symptoms?)

keegancsmith (approved):
changelog entry and backport?

stefanhengl:
> Could you explain what the outcome of the bug would look like (in case others run into this problem and want to know if it fixes their symptoms?

If users define file filters without defining `MaxFileSizeBytes`, all files are skipped during an embedding job. The job will be marked as completed, but no files will have been embedded.

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ All notable changes to Sourcegraph are documented in this file.
 - Fixed a bug where typing in the GraphQL editor in the Site Admin API console could cause the cursor to jump to the start of the editor. [#57862](https://github.com/sourcegraph/sourcegraph/pull/57862)
 - The blame column no longer ignores whitespace-only changes by default. [#58134](https://github.com/sourcegraph/sourcegraph/pull/58134)
 - Long lines now wrap correctly in the diff view. [#58138](https://github.com/sourcegraph/sourcegraph/pull/58138)
+- Defining file filters for embeddings jobs no longer causes all files to be skipped if `MaxFileSizeBytes` isn't defined. [#58262](https://github.com/sourcegraph/sourcegraph/pull/58262)
```
keegancsmith, line 37: given you are backporting needs to be moved to 5.2.x unreleased sections?

stefanhengl: I wasn't sure how to handle this TBH. I thought it was preferable to have a clean merge and not assume the back port here. Probably overthinking it. I will take care that eventually things will make sense.

PR #58272, opened 2023-11-10
File tree: get rid of single child logic

camdencheek:
This PR removes the ability to collapse single-child directories. This feature has been pretty unstable and has been the source of quite a few bugs lately. Additionally, it's surprisingly expensive to implement, requiring fetching ton of extra data (the _entire_ recursive git tree) from gitserver, which adds significant latency to interactions with the file tree. Between the complexity, bugginess, and fundamental perf issues, I don't think it's worth keeping around. Especially since it's pretty rare for directories to contain only a single subdirectory. 

Before:
CleanShot 2023-11-14 at 12 16 06@2x

After:
CleanShot 2023-11-10 at 13 11 50@2x

There is a lot of very complex backend code that I will clean up as a followup. 

## Test plan

Tested that single-file directories now get rendered consistently.

### DIFF
--- a/client/shared/src/backend/repo.ts
+++ b/client/shared/src/backend/repo.ts
@@ -74,7 +74,7 @@ export const fetchTreeEntries = memoizeObservable(
                 fragment TreeFields on GitTree {
                     isRoot
                     url
-                    entries(first: $first, recursiveSingleChild: true) {
+                    entries(first: $first) {
                         ...TreeEntryFields
                     }
                 }
@@ -87,7 +87,6 @@ export const fetchTreeEntries = memoizeObservable(
                         url
                         commit
                     }
-                    isSingleChild
                 }
             `,
             variables: args,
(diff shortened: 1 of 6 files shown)


client/shared/src/backend/repo.ts
```diff
@@ -74,7 +74,7 @@ export const fetchTreeEntries = memoizeObservable(
                 fragment TreeFields on GitTree {
                     isRoot
                     url
-                    entries(first: $first, recursiveSingleChild: true) {
```
eseliger, line 77: can we remove it from the API as well?

eseliger (approved):

client/shared/src/backend/repo.ts
camdencheek: Yep, planning to do that as a followup. I don't expect this is used by any consumers other than Sourcegraph

PR #58278, opened 2023-11-11
Remove ensureRevision

eseliger:
## Preview 🤩
[Preview Link](https://docs.sourcegraph.com/@es/no-ensure-revision)


cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -3883,10 +3883,6 @@ type Repository implements Node & GenericSearchResultInterface {
         The head of the diff ("new" or "right-hand side"), or "HEAD" if not specified.
         """
         head: String
-        """
-        Attempt to fetch missing revisions from remote if they are not found
-        """
-        fetchMissing: Boolean = true
```
eseliger, line 3889: Not used by us, but defaulted to true. This'll mean that we no longer call ensure revision here.

PR #58279, opened 2023-11-12
Refactor HTTP cli package

eseliger:


internal/embeddings/embed/embed.go
```diff
@@ -23,13 +23,18 @@ import (
 )
 
 func NewEmbeddingsClient(config *conftypes.EmbeddingsConfig) (client.EmbeddingsClient, error) {
+	cli, err := httpcli.NewExternalClientFactory().Doer()
+	if err != nil {
+		return nil, err
+	}
```
eseliger, line 29: maybe httpcli.ExternalDoer

internal/search/backend/zoekt.go
```diff
@@ -14,21 +14,7 @@ import (
...
-// We don't use the normal factory for internal requests because we disable
-// retries. Currently our retry framework copies the full body on every
-// request, this is prohibitive when zoekt generates a large query.
-//
-// Once our retry framework supports the use of Request.GetBody we can switch
-// back to the normal internal request factory.
-var zoektHTTPClient, _ = httpcli.NewFactory(
-	httpcli.NewMiddleware(
-		httpcli.ContextErrorMiddleware,
-	),
```
eseliger, line 26: Middlewares aren't actually used when using Client over Doer. After patching zoekt to allow anything that fulfills the Doer interface, we can pass our full doer in here, and actually get the ContextErrorMiddleware.

Also, retries have been reimplemented in the httpcli package and now support Request.GetBody, so I believe there's no need to have a separate factory here any longer.

internal/httpcli/client.go
```diff
@@ -89,52 +80,29 @@ var redisCache = rcache.NewWithTTL("http", 604800)
...
 //
 // Responses load from cache will have the 'X-From-Cache' header set.
-var CachedTransportOpt = NewCachedTransportOpt(redisCache, true)
-
-// ExternalClientFactory is a httpcli.Factory with common options
-// and middleware pre-set for communicating with external services.
-// WARN: Clients from this factory cache entire responses for etag matching. Do not
-// use them for one-off requests if possible, and definitely not for larger payloads,
-// like downloading arbitrarily sized files! See UncachedExternalClientFactory instead.
-var ExternalClientFactory = NewExternalClientFactory()
```
eseliger, line 99: No standard factory anymore, create your own if you need to. It didn't add a bunch over calling NewExternalClientFactory instead, other than it being a public variable that was "abused" in tests.

internal/httpcli/client.go
```diff
@@ -89,52 +80,29 @@ var redisCache = rcache.NewWithTTL("http", 604800)
...
-
-// UncachedExternalClientFactory is a httpcli.Factory with common options
-// and middleware pre-set for communicating with external services, but with caching
-// responses disabled.
-var UncachedExternalClientFactory = newExternalClientFactory(false)
```
eseliger, line 104: no need for uncached anymore, the default behavior is now to NOT cache, which is what we want in the majority of cases.

internal/httpcli/client.go
```diff
@@ -145,51 +113,19 @@ func newExternalClientFactory(cache bool, middleware ...Middleware) *Factory {
...
-// a convenience for existing uses of http.DefaultClient.
-// WARN: This client caches entire responses for etag matching. Do not use it for
-// one-off requests if possible, and definitely not for larger payloads, like
-// downloading arbitrarily sized files! See UncachedExternalClient instead.
-var ExternalClient, _ = ExternalClientFactory.Client()
-
-// UncachedExternalClient returns a shared client for external communication. This is
-// a convenience for existing uses of http.DefaultClient.
-// WARN: This client does not cache responses. To cache responses see ExternalClient instead.
-var UncachedExternalClient, _ = UncachedExternalClientFactory.Client()
```
eseliger, line 181: Using Clients instead of Doers have a footgun: They don't support middlewares. So in most cases, we don't actually want to use it. I've cleaned up incorrect usages.
If a client is really necessary (because we pass it to a library that expects the *http.Client), then get it explicitly from the Factory.

internal/httpcli/retry.go
```diff
@@ -0,0 +1,393 @@
...
+}
+
+// Everything below this line is heavily inspired by https://github.com/PuerkitoBio/rehttp
+// but modified and simplified for Sourcegraph needs such as:
+// - Not supporting old Go versions for simplicity
+// - Adding support for GetBody on the Request to reduce memory overhead
+// The BSD 3-clause license (https://opensource.org/licenses/BSD-3-Clause).
+
+// Copyright (c) 2015-2016, Martin Angers
+// All rights reserved.
```
bobheadxi, line 252: Should we just put this in a separate internal package, eg `internal/httpcli/internal/rehttp /rehttp.go`?

internal/httpcli/transport.go
```diff
@@ -26,6 +26,13 @@ func unwrapAll(transport WrappedTransport) *http.RoundTripper {
 	return wrapped
 }
 
+func WrapTransport(transport, base http.RoundTripper) WrappedTransport {
```
bobheadxi, line 29: IIRC `transport` must call `base` in `transport`'s implementation of `http.RoundTripper` right? `WrappedTransport` itself doesn't do it for you, which I realize may be a bit confusing to implementors. Might be worth documenting especially if we make this exported

eseliger: It should make some use of it, yes. I can add a doc string!

internal/httpcli/client.go
bobheadxi: > They don't support middlewares

No middlewares, but _does_ support Opts right?

bobheadxi: I actually only just noticed that there is a docstring with this exact disclaimer on `(*Factory).Client()` - maybe we can make it more prominent, and encourage callers to use `Doer` in the docstring?

bobheadxi: Also, are we sure that we can't extend `http.Client`? Our Middleware type looks like:

```go
type Doer interface {
	Do(*http.Request) (*http.Response, error)
}

type Middleware func(Doer) Doer
```

`http.Client.Transport` accepts a `RoundTripper` that looks like: 

```
type RoundTripper interface {
	RoundTrip(*http.Request) (*http.Response, error)
}
```

It seems like we can have an adapter that coverts `Doer` into `RoundTripper` and wrap the client transport in it?

eseliger: We could convert the middlewares into Opts, they're really not much else than a simplified transport, no? That way we could, I believe

bobheadxi: Yeah, I think so - `httpcli.Doer` is still the simpler interface for internal usage, but some libraries require you to provide a `http.Client`, so we should probably try to close the disparity, especially because the behaviour disparity is surprising

eseliger: Yep, that feels useful to me. I'll play with that (and if it bloats the PR even more I might split it out into a stacked one 😬)

bobheadxi: Circling back given the thread in https://github.com/sourcegraph/sourcegraph/pull/58016#discussion_r1397367483 , I think it's important we keep (and prefer) `Doer` and `Middleware`: because `Doer` middleware seems intended to be applied _before_ `http.Client`. i.e.

```
Doer-s -> *http.Client.Do() -> http.RoundTripper
  ^                                  ^
  |                                  | 
  |                              modified by Opts
modified by Middleware
```

See https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/internal/httpcli/client.go?L227-259

`Doer` implementations _do not_ need to avoid mutating Request in the same way `RoundTripper` implementations do, because `net/http` internals don't have access to the request at the point `Doer` is used.

If anything, we should probably Do (haha) away with `Opt` instead, and construct `http.Client` by:

```
*http.Client.Do() -> (request.Clone() -> Doer-s) -> http.RoundTripper
                                 ^
                             in http.RoundTripper wrapper
```

bobheadxi: I think with `Doer` and `Middleware`, we don't need to clone on every request, as the ask to Clone seems specific to `net/http` - i.e. within the `Middleware` and `Doer` ecosystem, we don't assume that requests must be cloned for modification

bobheadxi: Avoiding RoundTripper-wrapping and only using `Doer` ensures that any changes we make are only applied (in general) before the Request enters `net/http`, and after the Response leaves `net/http`, so we don't have to worry about `net/http` internals

PR #58280, opened 2023-11-12
Chore: clean up dead code

camdencheek:
I was playing around with the soon-to-be-stabilized [deadcode](https://pkg.go.dev/golang.org/x/tools/internal/cmd/deadcode) command, and wanted to see how it worked on a large codebase like `sourcegraph/sourcegraph`.

This PR is the result of me running `deadcode -test -filter 'github.com/sourcegraph/sourcegraph/.*internal/.*' ./...` and manually going through the results. I only included internal packages because those cannot be imported from external projects.

I only deleted things I had high confidence were unintentionally left unused. There are quite a lot of additional results that I ignored because I was either not confident that they weren't still useful or they were written for the sake of having a consistent API. Feel free to take another pass and take a peek at the code you're more familiar with 🙂 

## Test plan

Everything builds. Everything deleted is dead code, confirmed by `gopls` and run through `git log -S` to ensure that it _should_ be unused.


cmd/gitserver/internal/cacert/cacert.go
```diff
@@ -21,16 +21,3 @@ var System = syncx.OnceValues(func() ([][]byte, error) {
...
-	return &CertPool{}
-}
-
-func (c *CertPool) AppendCertsFromPEM(data []byte) {
-	c.certs = append(c.certs, data)
-}
-
-func (c *CertPool) len() int {
-	return len(c.certs)
-}
```
keegancsmith, line 36: safe to delete this code but note next time update.bash is run in this directory it will add it back.

keegancsmith (approved):
lgtm

PR #58289, opened 2023-11-13
Cody: Update Azure OpenAI provider to use azure go client and support additional auth methods 

chwarwick:
Converts the Azure OpenAI provider to use Azure's go client for the purpose of supporting additional auth methods beyond the API Key.  When the Access token is not specified AzureDefaultCredentials are used which are tried in the following order:

1. EnvironmentCredential
2. WorkloadIdentityCredential
3. ManagedIdentityCredential
4. AzureCLICredential - useful for local development authenticate via `az login` 

resolves https://github.com/sourcegraph/sourcegraph/issues/58301

## Test plan
Update tests
Tested chat & autocomplete with both API Key and default credentials
Verified auth errors after `az logout` 
Embedded multiple repos
Checked embeddings search results used with chat request


internal/completions/client/client.go
```diff
@@ -37,7 +37,11 @@ func getBasic(endpoint string, provider conftypes.CompletionsProviderName, acces
 	case conftypes.CompletionsProviderNameOpenAI:
 		return openai.NewClient(httpcli.ExternalDoer, endpoint, accessToken), nil
 	case conftypes.CompletionsProviderNameAzureOpenAI:
-		return azureopenai.NewClient(httpcli.ExternalDoer, endpoint, accessToken), nil
+		azureAPIClient, err := azureopenai.GetAzureAPIClient(endpoint, accessToken)
+		if err != nil {
+			return nil, err
+		}
+		return azureopenai.NewClient(azureAPIClient)
```
slimsag, line 44: I think we should avoid complexity in this function and match the old pattern used here so all the switch cases are single-line

slimsag (approved):
Looks great!

Issue #58064, opened 2023-11-01
Account requests: Approving request errors when email already exists


gabtorre:
- **Sourcegraph version:**  5.2.1
- **Platform information:** 

#### Steps to reproduce:

1. Go to the admin panel and find the list of pending account requests.
2. Try to approve a pending account request for a user that is already registered and using Sourcegraph.

#### Expected behavior:
The account request should be approved without errors, and the user's request status should be updated to reflect that they have been approved.

#### Actual behavior:
When approving the account request, the admin sees the following error message: "failed to create user: cannot create user: err_email_exists." This indicates that the system is trying to create a new user account, even though the user is already registered and using Sourcegraph. The system should recognize that the user is already registered and simply update their status to approved.

If you would like immediate help on this, please email support@sourcegraph.com (you can still create the issue, but there are no [SLAs](https://about.sourcegraph.com/support/) on issues like there are for support requests).

PR #58293, opened 2023-11-13
Handle Existing Verified Email on Access Request Approval

gabtorre:
Fixes https://github.com/sourcegraph/sourcegraph/issues/58064

This PR introduces a new workflow for approving access requests involving an existing verified email. Previously, admins encountered an error when they attempted to approve an access request for a user whose email was already verified. 

- The approval now invokes accessRequestCreateUser to handle scenarios where the user might already exist in the system.
-  The access request is marked as approved without creating a duplicate user.
- Rather than displaying a link to a new account, the system will display the user's email with the approved message.

Before:

https://github.com/sourcegraph/sourcegraph/assets/69164745/5dfd4809-754a-448e-a8ef-02b9616d141e

After:

https://github.com/sourcegraph/sourcegraph/assets/69164745/6ced2f7e-2ebc-46b6-b5cd-f0d4edcf70f9

## Test plan
locally tested

### DIFF
--- a/cmd/frontend/graphqlbackend/access_requests.go
+++ b/cmd/frontend/graphqlbackend/access_requests.go
@@ -175,3 +175,29 @@ func unmarshalAccessRequestID(id graphql.ID) (userID int32, err error) {
 	err = relay.UnmarshalSpec(id, &userID)
 	return
 }
+
+func (r *schemaResolver) AccessRequestCreateUser(ctx context.Context, args *struct {
+	Username      string
+	Email         *string
+	VerifiedEmail *bool
+}) (*createUserResult, error) {
+
+	newResult, err := r.CreateUser(ctx, &struct {
+		Username      string
+		Email         *string
+		VerifiedEmail *bool
+	}{
+		Username:      args.Username,
+		Email:         args.Email,
+		VerifiedEmail: args.VerifiedEmail,
+	})
+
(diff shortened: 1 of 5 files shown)


cmd/frontend/graphqlbackend/access_requests.go
```diff
@@ -175,3 +175,29 @@ func unmarshalAccessRequestID(id graphql.ID) (userID int32, err error) {
...
 	return
 }
+
+func (r *schemaResolver) AccessRequestCreateUser(ctx context.Context, args *struct {
+	Username      string
+	Email         *string
+	VerifiedEmail *bool
+}) (*createUserResult, error) {
+
+	newUser, err := r.CreateUser(ctx, &struct {
```
eseliger, line 185: we must add verification that the requesting user is site-admin here, otherwise anyone can create a new user on the instance.

eseliger (changes requested):
Left a comment on the original ticket, I wonder if we could instead change how this works and never even have invalid records on this page.
Right now, it's not really possible to distinguish "the user already existed and my button click didn't do anything" and "the user was added to the instance".

At the very least we'll have to make sure that only site-admins can create users.

gabtorre:
I made the requested changes in this PR https://github.com/sourcegraph/sourcegraph/pull/59113

PR #58294, opened 2023-11-14
Add secondary telemetry using Telemetry Gateway to web

nathan-downs:
## Test plan

run sg start
verify logged events in event_logs
verify logged events in bigquery


nathan-downs:
So I'm sure this PR is going to raise some eyebrows because of the scope of changes, but there's really no way around it here. Once you start touching props in one location is cascades out and before you know it, 107 files.

On the plus side, we do have fairly robust tests in place and this is just adding secondary telemetry.

PR #58312, opened 2023-11-14
Code Navigation: clean up prev next buttons

jasonhawkharris:
Better looking previous and next buttons for blob view search. Now more closely resembles safari's find feature.
 
Before: 
Screenshot 2023-11-14 at 3 06 47 PM

After:
Screenshot 2023-11-14 at 3 07 16 PM

## Test plan
Manual testing in light and dark mode.

### DIFF
--- a/client/web/src/repo/blob/codemirror/search.module.scss
+++ b/client/web/src/repo/blob/codemirror/search.module.scss
@@ -1,0 +1,10 @@
+.bgroup-left {
+    border-top-right-radius: 0%;
+    border-bottom-right-radius: 0%;
+}
+
+.bgroup-right {
+    border-top-left-radius: 0%;
+    border-bottom-left-radius: 0%;
+    border-left: none;
+}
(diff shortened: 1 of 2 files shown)


client/web/src/repo/blob/codemirror/search.module.scss
```diff
@@ -0,0 +1,10 @@
+.bgroup-left {
+    border-top-right-radius: 0%;
```
vovakulikov, line 2: I thought that our stylist should catch this but just FYI (I don't have problems with it) with zero values you can omit units so `border-top-right-radius: 0;`

vovakulikov (approved):
Thanks!

client/web/src/repo/blob/codemirror/search.module.scss
jasonhawkharris: Good to know!

PR #58317, opened 2023-11-14
telemetry: allow env-var allowlisting of sensitive metadata export

bobheadxi:
In case we need more metadata from Cloud instances or a particular customer, an env var override could come in handy. It must be proactively configured.

This change also fixes a misspelling in the file name.

## Test plan

unit test on the parser

### DIFF
--- a/internal/telemetry/sensitivemetadataallowlist/sensitivemetadataallowlist.go
+++ b/internal/telemetry/sensitivemetadataallowlist/sensitivemetadataallowlist.go
@@ -2,24 +2,36 @@ package sensitivemetadataallowlist
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/sourcegraph/sourcegraph/cmd/frontend/envvar"
 	"github.com/sourcegraph/sourcegraph/lib/errors"
 
+	"github.com/sourcegraph/sourcegraph/internal/env"
 	"github.com/sourcegraph/sourcegraph/internal/telemetry"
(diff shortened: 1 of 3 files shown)


internal/telemetry/sensitivemetadataallowlist/sensitivemetadataallowlist.go
```diff
@@ -2,24 +2,36 @@ package sensitivemetadataallowlist
...
 	"github.com/sourcegraph/sourcegraph/cmd/frontend/envvar"
 	"github.com/sourcegraph/sourcegraph/lib/errors"
 
+	"github.com/sourcegraph/sourcegraph/internal/env"
 	"github.com/sourcegraph/sourcegraph/internal/telemetry"
 	telemetrygatewayv1 "github.com/sourcegraph/sourcegraph/internal/telemetrygateway/v1"
 )
 
+var rawAdditionalAllowedEventTypes = env.Get("SRC_TELEMETRY_SENSITIVEMETADATA_ADDITIONAL_ALLOWED_EVENT_TYPES", "",
+	"Additional event types to include in sensitivemetadataallowlist.AllowedEventTypes, in comma-separated '${feature}::${action}' format.")
```
bobheadxi, line 16: Used `::` because it's not a pattern we suggest in the new events naming scheme

unknwon: I like `::` anyway. 😁

unknwon (approved):

nathan-downs (approved):

PR #58320, opened 2023-11-15
GraphQL: remove recursiveSingleChild option from backend

camdencheek:
This is the backend half of [this PR](https://github.com/sourcegraph/sourcegraph/pull/58272). It removes the `recursiveSingleChild` logic from the tree resolver. All consumers of this field have been removed from the client already, and it's unlikely this would be something that customers would use (or even be able to use correctly) so I think it's safe to make what is technically a breaking change. 

## Test plan

Poked around in the file tree and everything looked fine. All consumers of this API have already been removed.

### DIFF
--- a/cmd/frontend/graphqlbackend/git_tree.go
+++ b/cmd/frontend/graphqlbackend/git_tree.go
@@ -22,16 +22,6 @@ func (r *GitTreeEntryResolver) IsRoot() bool {
 type gitTreeEntryConnectionArgs struct {
 	graphqlutil.ConnectionArgs
 	Recursive bool
-	// If recurseSingleChild is true, and the requested path only has a single child
-	// that is a directory, we recurse into that directory.
-	// Only respected when Recursive is false.
-	// Example:
-	// FS structure:
-	// /folder/subfolder/file
-	// Request for entries of /folder
-	// We don't return only /folder/subfolder, and instead also travserse into subfolder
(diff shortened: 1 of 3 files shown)


cmd/frontend/graphqlbackend/git_tree.go
```diff
@@ -114,15 +78,13 @@ func (r *GitTreeEntryResolver) entries(ctx context.Context, args *gitTreeEntryCo
...
-	seen := 0
 	for _, entry := range entries {
-		if seen == maxResolvers {
+		if len(resolvers) >= maxResolvers {
 			break
 		}
 
 		// Apply any additional filtering
 		if filter == nil || filter(entry) {
-			seen++
```
camdencheek, line 125: unrelated, just removing a redundant variable that confused me at first glance.

cmd/frontend/graphqlbackend/git_tree_test.go
```diff
@@ -370,55 +370,6 @@ func TestGitTree_Entries(t *testing.T) {
 		}, entries)
 	})
 
-	t.Run("RecursiveSingleChild", func(t *testing.T) {
```
eseliger, line 373: now that we don't need to test this anymore I think there's some files that we no longer mock out, but we can leave that for future selfs.

eseliger (approved):

PR #58326, opened 2023-11-15
Don't disable features when license disabled

### LINKED ISSUE #58200: Sourcegraph features shouldn't get disabled when a license expires

### LINKED ISSUE #58201: Sign out all user sessions when a license is expired

### LINKED ISSUE #58202: Prevent user sign-in when a license is expired

### LINKED ISSUE #58399: Only allow admin sign-in when license is expired

pjlast:
Closes #58200
Closes #58201 
Closes #58202 
Closes #58399 

I decided to bundle all of these changes together since:
1. They are all loosely connected
2. Having some of them without the other doesn't make a lot of sense and could cause an instance to be permanently inaccessible
3. The overall change count is quite small

After this PR is merged, this is how Sourcegraph would behave when a license is expired:
1. Sourcegraph would continue to function as it always did. Repo syncing, cloning, permissions syncing etc. all continues as normal.
2. All existing regular users (not site admin users) get signed out of Sourcegraph
3. Regular users will no longer be able to sign in to Sourcegraph. They will receive an error that states sign-in is disabled because the Sourcegraph license is no longer valid.
4. Site admins will still be able to sign in. This allows them to update the license key.
5. Site admins will not be able to turn other users into site admins, or create new users.
6. Regular user access tokens will not function while the license is expired

## Test plan

Added some tests for the expected behaviour around user sessions and access tokens.


cmd/frontend/internal/auth/oauth/session.go
```diff
@@ -123,7 +124,14 @@ func SessionIssuer(logger log.Logger, db database.DB, s SessionIssuerHelper, ses
 		}
 
 		// Since we obtained a valid user from the OAuth token, we consider the GitHub login successful at this point
-		ctx = actor.WithActor(ctx, actr)
+		ctx, err = session.SetActorFromUser(ctx, w, r, user, expiryDuration)
+		if err != nil { // TODO: test session expiration
```
eseliger, line 128: is this TODO still up to date? :D

internal/session/session_test.go
```diff
@@ -475,3 +478,80 @@ func TestOldUserSessionSucceeds(t *testing.T) {
...
+		WantActor bool
+	}{
+		"admin user is allowed": {
+			UID:       adminUID,
+			WantActor: true,
+		},
+		"regular user is not allowed": {
+			UID:       regularUID,
+			WantActor: false,
+		},
```
eseliger, line 521: could we assert here that the cookie was also removed?

internal/database/access_tokens.go
```diff
@@ -239,6 +239,31 @@ INSERT INTO access_tokens(subject_user_id, scopes, value_sha256, note, creator_u
...
 
+var adminOnlyAccessTokenQuery string = `
+UPDATE access_tokens t SET last_used_at=now()
+WHERE t.id IN (
+	SELECT t2.id FROM access_tokens t2
+	JOIN users subject_user ON t2.subject_user_id=subject_user.id AND subject_user.deleted_at IS NULL
+	JOIN users creator_user ON t2.creator_user_id=creator_user.id AND creator_user.deleted_at IS NULL
+	WHERE t2.value_sha256=$1 AND t2.deleted_at IS NULL AND
+	$2 = ANY (t2.scopes)
+	AND subject_user.site_admin AND creator_user.site_admin
```
eseliger, line 250: do we really need to assert here that creator_user is a site-admin?

internal/database/access_tokens.go
```diff
@@ -249,19 +274,19 @@ func (s *accessTokenStore) Lookup(ctx context.Context, token, requiredScope stri
...
 
+	info, err := licensing.GetConfiguredProductLicenseInfo()
+	if err != nil {
+		return 0, err
+	}
+
+	query := regularAccessTokenQuery
+	if info.IsExpired() {
+		query = adminOnlyAccessTokenQuery
+	}
```
eseliger, line 285: any chance we could use our "query builder" pattern here and avoid some duplication?

Also, from a code organization perspective this contains a tad too much business logic for my taste for a package called database 😬 
Could we make it a flag `allowOnlyAdmin bool` (or reconcile into a struct, so many options now), and from where we call it we figure out if we want to set this to true instead?
Not a blocking comment.

internal/database/access_tokens.go
```diff
@@ -249,19 +274,19 @@ func (s *accessTokenStore) Lookup(ctx context.Context, token, requiredScope stri
...
-	JOIN users creator_user ON t2.creator_user_id=creator_user.id AND creator_user.deleted_at IS NULL
-	WHERE t2.value_sha256=$1 AND t2.deleted_at IS NULL AND
-	$2 = ANY (t2.scopes)
-)
-RETURNING t.subject_user_id
-`,
+		query,
 		tokenHash, requiredScope,
 	).Scan(&subjectUserID); err != nil {
 		if err == sql.ErrNoRows {
```
eseliger, line 292: wonder if, for easier debugging etc, we should add a special error here that indicates "was found but you didn't pay" 🤔

eseliger (approved):
LGTM, left a few small inline comments. Great work love the cleanup :)

PR #58328, opened 2023-11-15
repo: Show last commit information in file list

### LINKED ISSUE #57827: Tree view: more useful recent activity

fkling:
Closes #57827 

This commit changes the files panel to show the last commit message and date that changed the respective file.

The commits panel is removed to avoid confusion and unnecessarily duplicating information. I originally planned to add the history toggle button to the "actions" bar for tree pages, so that users can still get the same information via the history panel.
However the whole panel logic is currently part of the blob page and making that work fro the tree page that would require some refactoring. This is better be done in a separate PR, but maybe we also find out that we don't need this functionality.

The history data is loaded separately because it's generally slower to load than just the file list. While the commit data is loading the file panel will appear as a table with a single column. Once the data is available the other column headers and data appear. Without a loading state this felt better than having empty columns.

The GraphQL query for fetching the history data specifies repo and commit IDs to cache this data client side.

## Test plan

- Open repository root page
- Open tree page

### DIFF
--- a/client/web/src/repo/tree/TreePagePanels.tsx
+++ b/client/web/src/repo/tree/TreePagePanels.tsx
@@ -1,10 +1,12 @@
-import React, { type FC, useCallback, useRef, useState, useEffect } from 'react'
+import React, { type FC, useRef, useState, useEffect, useMemo } from 'react'
 
-import { mdiFileDocumentOutline, mdiFolderOutline, mdiMenuDown, mdiMenuUp } from '@mdi/js'
+import { mdiFileDocumentOutline, mdiFolderOutline } from '@mdi/js'
 import classNames from 'classnames'
 
+import { Timestamp } from '@sourcegraph/branded/src/components/Timestamp'
 import { NoopEditor } from '@sourcegraph/cody-shared/dist/editor'
 import { basename } from '@sourcegraph/common'
+import { gql } from '@sourcegraph/http-client'
 import type { TreeFields } from '@sourcegraph/shared/src/graphql-operations'
 import {
     Card,
@@ -14,19 +16,42 @@ import {
     Link,
     LinkOrSpan,
     LoadingSpinner,
-    ParentSize,
-    StackedMeter,
-    Tooltip,
     useElementObscuredArea,
 } from '@sourcegraph/wildcard'
 
 import { FileContentEditor } from '../../cody/components/FileContentEditor'
 import { useCodySidebar } from '../../cody/sidebar/Provider'
-import type { BlobFileFields } from '../../graphql-operations'
+import type { BlobFileFields, GitCommitFields, TreeHistoryFields } from '../../graphql-operations'
 import { fetchBlob } from '../blob/backend'
 import { RenderedFile } from '../blob/RenderedFile'
+import { CommitMessageWithLinks } from '../commit/CommitMessageWithLinks'
 
 import styles from './TreePagePanels.module.scss'
+
+export const treeHistoryFragment = gql`
+    fragment TreeHistoryFields on TreeEntry {
+        path
+        history(first: 1) {
+            nodes {
+                commit {
+                    id
+                    canonicalURL
+                    subject
+                    author {
+                        date
+                    }
+                    committer {
+                        date
+                    }
+                    externalURLs {
+                        url
+                        serviceKind
+                    }
+                }
+            }
+        }
+    }
+`
 
 interface ReadmePreviewCardProps {
     entry: TreeFields['entries'][number]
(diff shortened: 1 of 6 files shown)


client/web/src/repo/tree/TreePagePanels.tsx
```diff
@@ -123,246 +148,80 @@ export interface DiffStat {
...
+            <tbody>
+                {entries.map(entry => (
+                    <tr key={entry.name}>
+                        <td className={styles.fileName}>
                             <LinkOrSpan
                                 to={entry.url}
                                 className={classNames(
                                     'test-page-file-decorable',
-                                    styles.treeEntry,
                                     entry.isDirectory && 'font-weight-bold',
```
taiyab, line 186: @fkling Let's switch this to `font-weight-medium` (for the folder/directory label).

taiyab (commented):
LGTM!

Once the toolbar issue gets resolved, this works before we come back to redesign this whole thing.

Great work @fkling!

client/web/src/repo/tree/TreePagePanels.tsx
fkling: Do you expect to still be a visual difference between folder and file text? I'm using a font that only providers "normal" and "bold", so folder and file entries look very similar now, but maybe I'm not the target audience 😆 :



GH seems to solve this by using a more visually distinctive folder icon. We could try switching to the filled folder icon?

Or maybe it's just something I have to get used to.

client/web/src/repo/tree/TreePagePanels.tsx
```diff
@@ -123,246 +148,80 @@ export interface DiffStat {
...
+                                        <CommitMessageWithLinks
+                                            to={fileHistoryByPath[entry.path].canonicalURL}
+                                            message={fileHistoryByPath[entry.path].subject}
+                                            className="text-muted"
+                                            externalURLs={fileHistoryByPath[entry.path].externalURLs}
                                         />
-                                        {
-                                            // In case of single child expansion, we need to get the name relative to
-                                            // the start of the directory (to include subdirectories)
-                                        }
```
camdencheek, line 311: oops, I missed a spot. Thanks!

client/web/src/repo/tree/TreePagePanels.tsx
```diff
@@ -123,246 +148,80 @@ export interface DiffStat {
...
-                                            )}
-                                        </ParentSize>
-                                    </div>
-                                </Tooltip>
-                            )}
-                        </div>
-                    </div>
+                                </td>
+                                <td className={classNames(styles.commitDate, 'text-muted')}>
+                                    <Timestamp noAbout={true} date={getCommitDate(fileHistoryByPath[entry.path])} />
```
camdencheek, line 214: Can we maybe set `noAgo={true}`? The `ago` on every line takes up some valuable horizontal space for no useful information since a commit should never be in the future. We could maybe rename the colume "last commit age" or something if "last commit date" doesn't make sense anymore.

CleanShot 2023-11-15 at 10 05 41@2x
CleanShot 2023-11-15 at 10 06 40@2x

cc @taiyab

client/web/src/repo/tree/TreePagePanels.tsx
```diff
@@ -123,246 +148,80 @@ export interface DiffStat {
...
+                                />
+                                {entry.name}
+                                {entry.isDirectory && '/'}
+                            </LinkOrSpan>
+                        </td>
+                        {fileHistoryByPath[entry.path] && (
+                            <>
+                                <td className={styles.commitMessage}>
+                                    <span title={fileHistoryByPath[entry.path].subject} data-testid="git-commit-message-with-links">
+                                        <CommitMessageWithLinks
```
camdencheek, line 205: Probably unrelated to this PR, but can we make this scale better to large screens? Now that the horizontal space is useful to display the commit message, it would be nice if the width of the component wasn't so limited.

CleanShot 2023-11-15 at 10 09 31@2x

camdencheek (approved):
This looks great! So much more useful.

client/web/src/repo/tree/TreePagePanels.tsx
taiyab: The folder one definitely needs to be bolder, but medium and not bold.

fkling: @taiyab This is what I see when I set it to medium. As I said, the font I use doesn't support different weights. I think that's more common on Linux.

client/web/src/repo/tree/TreePagePanels.tsx
taiyab: Yep, sounds good.

client/web/src/repo/tree/TreePagePanels.tsx
taiyab: Depends on the max container size we're using for that internal bit, which I think is 1200px right now (we should test going up to 1280px).

At minimum though, we can center it.

client/web/src/repo/tree/TreePagePanels.tsx
taiyab: @fkling Ah, OK. If it can't render medium, then fallback to bold is fine.

In a future iteration, we'll make the directory icon more differentiated so you can scan that list easier.

client/web/src/repo/tree/TreePagePanels.tsx
fkling: We can do that in a separate PR.

PR #58330, opened 2023-11-15
Don't remove the URL number / line selection when user clicks again

rafax:
Don't remove the line selection (and URL fragment) when a user clicks on a line again - Slack discussion, other Code Editor / browsing systems (Github / Gitlab / Google Code Search) don't remove it.
## Test plan

- tested locally

### DIFF
--- a/client/web/src/repo/blob/codemirror/linenumbers.ts
+++ b/client/web/src/repo/blob/codemirror/linenumbers.ts
@@ -432,7 +432,3 @@ export function shouldScrollIntoView(view: EditorView, range: SelectedLineRange)
         to.top <= view.scrollDOM.scrollTop
     )
 }
-
-function isSingleLine(range: SelectedLineRange): boolean {
-    return !!range && (!range.endLine || range.line === range.endLine)
-}
--- a/client/web/src/repo/blob/codemirror/linenumbers.ts
+++ b/client/web/src/repo/blob/codemirror/linenumbers.ts
@@ -335,11 +335,8 @@ export function selectableLineNumbers(config: SelectableLineNumbersConfig): Exte
                     }
 
                     const line = view.state.doc.lineAt(block.from).number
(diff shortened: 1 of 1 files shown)


fkling:
Could you verify that this doesn't add multiple history entries to the same line when clicking the selected line multiple times? If it does then we need to update the field logic in line 97 to not update itself if the effect value is the same as the field value (this should be done regardless so I could also just do a follow up PR).

rafax:
> Could you verify that this doesn't add multiple history entries to the same line when clicking the selected line multiple times? If it does then we need to update the field logic in line 97 to not update itself if the effect value is the same as the field value (this should be done regardless so I could also just do a follow up PR).

It doesn't add multiple items (see this [loom](https://www.loom.com/share/c288b0bc3cad41dcaab2de13ac428372?sid=5a9abae5-41b9-46c7-8fc3-41bd211f0651)).

fkling:
> It doesn't add multiple items (see this loom).

Great, thank you for checking! In the loom it looks like the first time you clicked on the line it didn't select it... did you just not hit the line number properly or is it actually an issue that clicking the first time doesn't select the line?

rafax:
> > It doesn't add multiple items (see this loom).
> 
> Great, thank you for checking! In the loom it looks like the first time you clicked on the line it didn't select it... did you just not hit the line number properly or is it actually an issue that clicking the first time doesn't select the line?

First time selection works correctly, I think it's an issue of moving focus from the "back/history" popup to browser viewport - first click moved focus viewport, next one selected the line. In subsequent navigations, I click some other part of viewport, and line gets selected without issues. In this [Loom](https://www.loom.com/share/f2b9aeb1b7d5408982ebe838f2644fe8), you can see that line selection is "inactive" when the popup is open, but becomes active if I dismiss the popup with Esc / clicking viewport.

fkling (approved):

PR #58331, opened 2023-11-15
client: Minor cleanup for search-based code intel

varungandhi-src:
The separation of the logic into different functions makes it clearer
what the order of searches is. It also makes it clearer that for some
reason, we're only using the locals information from the SCIP
Document for 'Find references', and not for 'Go to definition'.
Using the SCIP Document for for 'Go to definition' too could avoid
a network request.

## Test plan

TODO:
- [x] Check final state with `sg start web-standalone`

### DIFF
--- a/client/web/src/enterprise/codeintel/searchBased.ts
+++ b/client/web/src/enterprise/codeintel/searchBased.ts
@@ -110,11 +110,11 @@ function makeRepositoryPattern(repo: string): string {
 /** The time in ms to delay between unindexed search request and the fallback indexed search request. */
 const DEFAULT_UNINDEXED_SEARCH_TIMEOUT_MS = 5000
 
+export type RepoFilter = 'current-repo' | 'all-other-repos'
+
 /**
  * Invoke the given search function by modifying the query with a term that will
- * only look in the current git tree by appending a repo filter with the repo name
- * and the current commit or, if `negateRepoFilter` is set, outside of current git
- * tree.
+ * look in certain git trees (controlled by `repoFilter`).
  *
  * This is likely to timeout on large repos or organizations with monorepos if the
  * current commit is not an indexed commit. Instead of waiting for a timeout, we
@@ -125,7 +125,6 @@ const DEFAULT_UNINDEXED_SEARCH_TIMEOUT_MS = 5000
  *
  * @param search The search function.
  * @param args The arguments to the search function.
- * @param negateRepoFilter Whether to look only inside or outside the given repo.
  */
 export function searchWithFallback<
     P extends {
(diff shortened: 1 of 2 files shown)


client/web/src/enterprise/codeintel/searchBased.ts
```diff
@@ -18,7 +18,7 @@ export function definitionQuery({
 }: {
     /** The search token text. */
     searchToken: string
-    /** The path to file **/
+    /** The path to file */
```
varungandhi-src, line 21: Bunch of this was just done automatically by `pnpm run lint:js:changed --fix`

client/web/src/enterprise/codeintel/useSearchBasedCodeIntel.ts
```diff
@@ -273,29 +265,16 @@ export async function searchBasedDefinitions({
...
-            negateRepoFilter,
-            getSetting
+            { repo, isFork, isArchived, commit, path, fileContent, filterDefinitions, queryTerms },
+            repoFilter,
+            options.getSetting
         )
 
     // Perform a search in the current git tree
-    const sameRepoDefinitions = doSearch(false)
+    const sameRepoDefinitions = doSearch('current-repo')
```
keynmol, line 277: The introduction of this repoFilter is making everything much clearer in this codebase 👍

varungandhi-src: Actually, I should just delete the comment now that the code actually explains what it is doing instead of hiding it behind a boolean. 😮‍💨

client/web/src/enterprise/codeintel/searchBased.ts
varungandhi-src: Re-edited the patch; these are fixes for warnings.

client/web/src/enterprise/codeintel/useSearchBasedCodeIntel.ts
```diff
@@ -119,149 +120,139 @@ export const useSearchBasedCodeIntel = (options: UseSearchBasedCodeIntelOptions)
...
-    commit,
-    searchToken,
-    path,
-    position,
-    fileContent,
-    spec,
-    getSetting,
-    filter,
-}: UseSearchBasedCodeIntelOptions): Promise<Location[]> {
+function searchBasedReferencesViaSCIPLocals(options: UseSearchBasedCodeIntelOptions): Location[] | undefined {
```
fkling, line 123: Just for my understanding: This is "search based" because, although using SCIP, we don't actually have any connections between the occurrences/symbols. We are just looking up occurrences by name (or something like that)?

fkling (approved):

client/web/src/enterprise/codeintel/useSearchBasedCodeIntel.ts
varungandhi-src: Right now, "search-based" is the catch-all term for anything that is not precise (i.e. coming from the compiler). In this case, the SCIP Document that is attached to the blob view came from the syntax highlighter, which is able to add information about local symbols using purely tree-sitter for some languages.

That's why this is "search based".

SCIP as a protocol does not dictate what the source of SCIP data is. We have multiple producers, such as the syntax highlighter, compilers etc.

PR #58343, opened 2023-11-15
Replace default httpclient for cody requests 

chwarwick:
Switched the completions & embeddings clients to use an uncached client, as the requests should be unique and caching would not be helpful.
## Test plan
CI 
Chatted with Cody web successfully, embedded a repo successfully


bobheadxi (approved):
Nice 😁

rafax (commented):
Nice :)

PR #58346, opened 2023-11-15
sec: Make GraphQL cost limits configurable

evict:
This makes it easier for customer site-admin to change the GraphQL limits. It adds a `RateLimits` function to fetch the structure and makes sure we have default values.

## Test plan
CI tests, test locally.

### DIFF
--- a/cmd/frontend/graphqlbackend/graphqlbackend.go
+++ b/cmd/frontend/graphqlbackend/graphqlbackend.go
@@ -633,7 +633,7 @@ func NewSchema(
 	opts := []graphql.SchemaOpt{
 		graphql.Tracer(newRequestTracer(log.Scoped("GraphQL"), db)),
 		graphql.UseStringDescriptions(),
-		graphql.MaxDepth(maxDepth),
+		graphql.MaxDepth(conf.RateLimits().GraphQLMaxDepth),
 	}
 	opts = append(opts, graphqlOpts...)
 	return graphql.ParseSchema(
--- a/internal/conf/computed.go
+++ b/internal/conf/computed.go
@@ -461,29 +461,26 @@ func PasswordPolicyEnabled() bool {
 }
 
 func RateLimits() schema.RateLimits {
-	maxAliases := 500
-	maxFieldCount := 500_000
-	maxDepth := 30
-
-	rl := Get().RateLimits
-	if rl != nil {
-		if rl.GraphQLMaxAliases <= 0 {
-			rl.GraphQLMaxAliases = maxAliases
-		}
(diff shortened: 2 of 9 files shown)


schema/site.schema.json
```diff
@@ -49,6 +49,27 @@
       "type": "boolean",
       "default": false
     },
+    "RateLimits": {
```
eseliger, line 52: properties in the schema are usually lowercase

```suggestion
    "rateLimits": {
```

cmd/frontend/graphqlbackend/graphqlbackend.go
```diff
@@ -633,7 +633,7 @@ func NewSchema(
 	opts := []graphql.SchemaOpt{
 		graphql.Tracer(newRequestTracer(log.Scoped("GraphQL"), db)),
 		graphql.UseStringDescriptions(),
-		graphql.MaxDepth(maxDepth),
+		graphql.MaxDepth(conf.RateLimits().GraphQLMaxDepth),
```
eseliger, line 636: Hmm damn, this'll require a restart to take effect.. That makes rollouts much harder for multiple on-prem customers unfortunately.

I looked at the library and unfortunately the validator code lives in `internal` so we can't easily take it out of there.
Maybe a library change could help work around that, to add MaxDepthFunc that allows to return a new value for each validation pass (request).

internal/conf/computed.go
```diff
@@ -460,6 +460,32 @@ func PasswordPolicyEnabled() bool {
...
+		}
+		return *rl
+	}
+
+	return schema.RateLimits{
+		GraphQLMaxAliases:    maxAliases,
+		GraphQLMaxFieldCount: maxFieldCount,
+		GraphQLMaxDepth:      maxDepth,
+	}
+}
```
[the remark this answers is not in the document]
evict: TIL, thank you!

cmd/frontend/graphqlbackend/graphqlbackend.go
evict: I don't think we can make that on time for the release, what do you think?

eseliger: that's okay but can we make sure to follow up on that?

eseliger (approved):

cmd/frontend/graphqlbackend/graphqlbackend.go
evict: Will do !

mohammadualam (approved):

PR #58347, opened 2023-11-15
Code Navigation: fix double line highlight in Firefox

jasonhawkharris:
Fixes a bug that caused double line highlighting when using Code Search in Firefox. 

## Test plan
1. Manually test in firefox at different zoom levels
2. Manually test in chrome and safari at different zoom levels

### DIFF
--- a/client/web/src/repo/blob/codemirror/utils.ts
+++ b/client/web/src/repo/blob/codemirror/utils.ts
@@ -263,7 +263,3 @@ export function isRegularEvent(event: MouseEvent | KeyboardEvent): boolean {
         !event.ctrlKey
     )
 }
-
-export function browserIsFirefox(): boolean {
-    return navigator.userAgent.includes('Firefox')
-}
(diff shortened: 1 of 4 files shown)


client/web/src/repo/blob/codemirror/utils.ts
```diff
@@ -263,3 +263,7 @@ export function isRegularEvent(event: MouseEvent | KeyboardEvent): boolean {
         !event.ctrlKey
     )
 }
+
+export function browserIsFirefox(): boolean {
```
vovakulikov, line 267: I think we already have this helper, it's called `isFirefox` you can find it in `@sourcegraph/common` package

camdencheek:
@vovakulikov I think you need to have line wrapping enabled in firefox to reproduce.

jasonhawkharris:
@vovakulikov you'll need to have line-wrapping enabled and be using firefox in order to replicate the behavior.

fkling (approved):
Good find!

PR #58353, opened 2023-11-15
Code Navigation: add close button to blobview search

jasonhawkharris:
Add a close button to the blobview search component. 

Inactive: 
Screenshot 2023-11-15 at 2 44 13 PM

hover:
Screenshot 2023-11-15 at 2 44 33 PM

## Test plan
Manual testing in light and dark mode.

### DIFF
--- a/client/web/src/repo/blob/codemirror/search.module.scss
+++ b/client/web/src/repo/blob/codemirror/search.module.scss
@@ -9,12 +9,12 @@
     border-left: none;
 }
 
-.hovered {
+.close-button {
     border-radius: 100%;
     padding: 0.2rem;
 }
 
-.hovered:hover {
+.close-button:hover {
     background-color: var(--subtle-bg);
 }
 
--- a/client/web/src/repo/blob/codemirror/search.module.scss
+++ b/client/web/src/repo/blob/codemirror/search.module.scss
@@ -8,3 +8,16 @@
     border-bottom-left-radius: 0;
     border-left: none;
 }
+
+.hovered {
+    border-radius: 100%;
+    padding: 0.2rem;
+}
+
+.hovered:hover {
+    background-color: var(--subtle-bg);
+}
+
+.x {
+    color: var(--text-muted);
+}
--- a/client/web/src/repo/blob/codemirror/search.tsx
+++ b/client/web/src/repo/blob/codemirror/search.tsx
@@ -3,8 +3,6 @@
  * UI.
  */
 
-import React from 'react'
-
 import {
     findNext,
     findPrevious,
@@ -285,7 +283,7 @@ class SearchPanel implements Panel {
                         {searchKeybinding}
                     </Label>
                     {searchKeybindingTooltip}
-                    <span className={classNames(styles.closeButton, 'ml-2')}>
+                    <span className={classNames(styles.closeButton, 'ml-4')}>
                         <Icon
                             className={classNames(styles.x)}
                             onClick={() => closeSearchPanel(this.view)}
--- a/client/web/src/repo/blob/codemirror/search.tsx
+++ b/client/web/src/repo/blob/codemirror/search.tsx
@@ -3,11 +3,14 @@
  * UI.
  */
 
+import React from 'react'
+
 import {
     findNext,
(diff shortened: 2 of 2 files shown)


client/web/src/repo/blob/codemirror/search.tsx
```diff
@@ -275,6 +285,16 @@ class SearchPanel implements Panel {
                         {searchKeybinding}
                     </Label>
                     {searchKeybindingTooltip}
+                    <span className={classNames(styles.hovered, 'ml-2')}>
```
camdencheek, line 288: did you decide these class names can't go on the icon? Still a little confused why we need this span

client/web/src/repo/blob/codemirror/search.tsx
```diff
@@ -275,6 +285,16 @@ class SearchPanel implements Panel {
...
+                        <Icon
+                            className={classNames(styles.x)}
+                            onClick={() => closeSearchPanel(this.view)}
+                            size="sm"
+                            svgPath={mdiClose}
+                            aria-hidden={true}
```
camdencheek, line 294: why is this `aria-hidden`? Wouldn't we want the accessibility framework to be able to interact with a close button?

client/web/src/repo/blob/codemirror/search.tsx
```diff
@@ -275,6 +285,16 @@ class SearchPanel implements Panel {
                         {searchKeybinding}
                     </Label>
                     {searchKeybindingTooltip}
+                    <span className={classNames(styles.hovered, 'ml-2')}>
+                        <Icon
+                            className={classNames(styles.x)}
```
camdencheek, line 290: for my education: why a class rather than an id? This should be unique on the page, right? Or in the world of CSS modules, are ids no longer in vogue?

client/web/src/repo/blob/codemirror/search.module.scss
```diff
@@ -8,3 +8,16 @@
     border-bottom-left-radius: 0;
     border-left: none;
 }
+
+.hovered {
```
camdencheek, line 12: nit: can we use a more descriptive class name than `.hovered`? It makes me think this is a class describing something that's actively being hovered over. Maybe `.close-button` or something.

client/web/src/repo/blob/codemirror/search.tsx
jasonhawkharris: There are some CSS issues that come up with using the className on the icon. Being able to style the span separately has its advantages here.

client/web/src/repo/blob/codemirror/search.tsx
jasonhawkharris: yup! my bad!

client/web/src/repo/blob/codemirror/search.tsx
jasonhawkharris: @vovakulikov do you have thoughts on this? I don't really know what the convention is these days.

client/web/src/repo/blob/codemirror/search.module.scss
jasonhawkharris: sure can!

camdencheek (approved):

client/web/src/repo/blob/codemirror/search.tsx
```diff
@@ -275,6 +285,16 @@ class SearchPanel implements Panel {
                         {searchKeybinding}
                     </Label>
                     {searchKeybindingTooltip}
+                    <span className={classNames(styles.closeButton, 'ml-2')}>
```
taiyab, line 288: Is `ml-2` the margin left on the whole close button?

If so, let's up this a little more. Say `ml-4`.

Thanks.

taiyab (commented):
LGTM!

client/web/src/repo/blob/codemirror/search.tsx
peterguy: Don't know if this applies here, but in general I've used classes for things that apply to, or can apply to multiple elements, and ids for things that apply to only that particular element.

Issue #58363, opened 2023-11-16
Clean up gitserver clonePipelineRoutine


natikgadzhi:
## Problem statement

There's a nice `TODO` item in `clonePipelineRoutine` that seems like an easy first issue: remove the pipeline's dependency on the upstream gitserver `server` instance. 

https://github.com/sourcegraph/sourcegraph/blob/5666861dfa305e9b12bea12f05763275c08a26c5/cmd/gitserver/internal/server.go#L353

## Things to learn/practice during the implementation

Before pushing up some code, I wanted to make sure I understand what I'm doing clearly. The least coupling we can do, the better! 

`clonePipelineRoutine` uses server for a few things: 

- [ ] Easy: `p.s.Logger` is the same thing as `p.logger`, so we can just replace that.
- [ ] `p.s.doClone` — the pipeline needs to invoke actually cloning the repository. 
- [ ] `p.s.acquireCloneLimiter` — the server managers rate limits on cloning
- [ ] `p.s.setLastNonFatal` records an error into the repos PG database.
- [ ] Related: `cloneJobConsumer` should acquire it's own lock (see https://github.com/sourcegraph/sourcegraph/pull/24892)

We _could_ just make an interface with those methods and pass the interface instead of the server, abstracting some complexity away, but I'm not so sure if that improves code clarity much.

```go
type GitServerCloner interface {
    AcquireCloneLimit(argType) returnType
    SetLastNonLethal(argType) returnType
    DoClone(argType) returnType
}
```

The core curious piece is that because [`server.cloneRepo`](https://github.com/sourcegraph/sourcegraph/blob/5666861dfa305e9b12bea12f05763275c08a26c5/cmd/gitserver/internal/server.go#L1325 can clone synchronously, without using the pipeline, we can't move all of the logic (rate limiters and doClone) into the pipeline.

## Suggested approach

Alternatively, we could make `clonePipelineRoutine` responsible for all actual clone execution stuff. To do that, we could give it access to a few things, and move the `doClone` part into it.

`Server` will be responsible for the HTTP API, and for putting clone jobs on the queue, or executing them inline. `clonePipeline` will listen to the queue as it does now, and expose `doClone` to clone in a blocking way.

To make the shift gradual, we can leave the `isClonable` and it's RPS limiter on the server, or move it along too.

```go
type clonePipelineRoutine struct {
    ...
    // setLastNonFatal, setStatus and friends
    db database.GitserverRepoStore
    
[...]

PR #58366, opened 2023-11-16
Fix: Clean up gitserver clonePipelineRoutine.

xonx4l:
issue-:#58363

Removed the pipeline's dependency on the upstream gitserver server instance.
Implemented the suggested approach for clonePipelineRoutine as moved the doClone part into it.

Replaced p.s.Logger with p.logger.

### DIFF
--- a/cmd/gitserver/internal/server.go
+++ b/cmd/gitserver/internal/server.go
@@ -352,14 +352,13 @@ func (s *Server) NewClonePipeline(logger log.Logger, cloneQueue *common.Queue[*c
 }
 
 type clonePipelineRoutine struct {
-	logger log.Logger
-
-	tasks chan *cloneTask
-	// TODO: Get rid of this dependency.
-	s      *Server
-	queue  *common.Queue[*cloneJob]
-	cancel context.CancelFunc
+	logger     log.Logger
+    tasks      chan *cloneTask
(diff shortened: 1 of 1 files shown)


jdorfman:
Thanks @xonx4l! Someone will take a look ASAP.

cmd/gitserver/internal/server.go
```diff
@@ -352,15 +352,14 @@ func (s *Server) NewClonePipeline(logger log.Logger, cloneQueue *common.Queue[*c
...
-	s      *Server
-	queue  *common.Queue[*cloneJob]
-	cancel context.CancelFunc
+	logger     log.Logger
+    tasks      chan *cloneTask
+    db         database.GitserverRepoStore
+    rpsLimiter // RPS limiter shared with the server	
 }
 
+func (p* clonePipelineroutine) doClone() ...
```
eseliger, line 361: Hi and thanks for the contribution! 

Seems like this method still needs to be implemented for this PR to work.
That said, we got some work planned for getting rid of the clone queue here entirely early next year hopefully, so I'll leave it up to you if you want to spend the time on it right now :)

burmudar:
As there has been no activity on this PR for 3 months and as @eseliger mentioned there is planned work around this area - I'm going to go ahead and close this.

PR #58368, opened 2023-11-16
syntax-highlighter: Delete Theme and CSS fields

varungandhi-src:
In the code, the [CSS field was always set to be true](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@main/-/blob/internal/highlight/highlight.go?L420), which meant that the Theme field would
be ignored in the query to syntect server. This patch removes both fields, keeping only
the `q.css == true` code path, and deleting all the mentions of themes.

## Test plan

Run `cargo run --bin syntect_server` followed by `SRC_SYNTECT_SERVER=http://127.0.0.1:8000 sg start enterprise-codeintel` and check highlighting is still working fine.

### DIFF
--- a/docker-images/syntax-highlighter/README.md
+++ b/docker-images/syntax-highlighter/README.md
@@ -40,7 +40,7 @@ See [API](./docs/api.md)
 
 ## Configuration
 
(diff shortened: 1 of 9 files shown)


tjdevries (approved):
Nice!

PR #58372, opened 2023-11-16
repos: refactor and clean-up the `Source.ListRepos` tests

mrnugget:
Here's some `stop-the-bleeding` for you.

I ran into these tests yesterday and really, _really_ struggled adding a new test for the `Exclude` functionality on the GitHub code host connection. Turns out that the test's `assert` method basically re-implements the implementation and is prone to false negatives and false positives.

The rest of the `testCases` also struck me as "someone added them here because this was easy" and then they've grown over the past few years.

So what I did here was to

- Split up that huge test function into separate smaller functions that can be run independently and also allow easier updating of VCR fixtures
- **Keep the VCR fixtures the same!**
- Make fixture names more visible
- Add simple but clear test assertions that assure that the `Source` yields the repos we expect and not "negative tests" where we want to check whether nothing we didn't expect was yielded.
- Remove unnecessary checks for `err` and replace with more solid ones
- Add little helper to list repos
- Tweak assertions and test setup in `Exclude` test to make it clearer what goes on.

## Test plan

- Ran these tests multiple times.

### DIFF
--- a/internal/repos/BUILD.bazel
+++ b/internal/repos/BUILD.bazel
@@ -194,7 +194,6 @@ go_test(
         "@com_github_google_go_cmp//cmp",
         "@com_github_google_go_cmp//cmp/cmpopts",
         "@com_github_google_uuid//:uuid",
-        "@com_github_grafana_regexp//:regexp",
         "@com_github_hexops_autogold_v2//:autogold",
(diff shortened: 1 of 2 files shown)


pjlast (approved):

eseliger (commented):
thank you!!

PR #58376, opened 2023-11-16
github: fetch & persist disk usage of repositories

mrnugget:
This changes the V3/V4 clients to populate the new `DiskUsageKibibytes` field on `github.Repository`.

I've tried to follow existing conventions wrt to naming and JSON tags, but not 100% sure here. Any direction appreciated.

It's part of a larger effort, but thought I'd split this out, because the testdata makes this always a lot to review.

### Kibibytes vs. Kilobytes

A few words on the kibibytes/kilobytes thing:

as far as I can tell (and JH is my witness) GitHub's API returns kibibytes (multiples of 1024) and not kilobytes (multiples of 1000).

Look at src-cli in the API: https://api.github.com/repos/sourcegraph/src-cli

That includes:

    "size": 21236

Now when you do a bare clone and get the size in _kilobytes_ you get this:

- `du -k src-cli.git` prints `22532 src-cli.git`
- `du -h src-cli.git` prints `22M src-cli.git`

Since GitHub's value is lower, it means its multiples are higher (1024 instead of 1000).

And if we convert their value we get a lower value too:

`echo '21236' | numfmt --to=iec --from-unit=1024` -> `21M`

### Accuracy of this value

I couldn't reproduce GitHub's size/diskUsage on my machines. I think it has to do with filesystem, bare clones, etc. But I think treating it as kibibytes is good enough for now.

## Test plan

- All the tests touched here. Updated them and ran them multiple times.
- Booted up fresh instance locally and made sure `diskUsage` is populated in `repo.metadata` column in database


internal/extsvc/github/common.go
```diff
@@ -1806,6 +1806,11 @@ type Repository struct {
 
 	// Parent is non-nil for forks and contains details of the parent repository.
 	Parent *ParentRepository `json:",omitempty"`
+
+	// DiskUsageKibibytes is, according to GitHub's docs, in kilobytes, but
+	// empirically it's in kibibytes (meaning: multiples of 1024 bytes, not
+	// 1000).
```
eseliger, line 1812: Just a note for future selfs, before the next sync this number will be 0 so maybe 0 should not count as a criterion for purging a repo 😬

eseliger (approved):

PR #58380, opened 2023-11-16
Instrument buffer length for BigQuery events

rafax:
[From time to time](https://sourcegraph.slack.com/archives/C05ABRRGB0B/p1700063219480789), our buffer for sending to BigQuery is full - this PR adds an otel Gauge to report on the size of the buffer (to verify whether this is a systematic issue, or one-off occurence). 
## Test plan

- locally tested by slowing down consumers


cmd/cody-gateway/internal/events/buffered.go
```diff
@@ -133,6 +147,7 @@ func (l *BufferedLogger) Start() {
 					sgtrace.Logger(event.spanCtx, l.log).
 						Error("failed to log buffered event", log.Error(err))
 				}
+				time.Sleep(10000 * time.Millisecond)
```
eseliger, line 150: That change seems unintentional?

eseliger (approved):

cmd/cody-gateway/internal/events/buffered.go
```diff
@@ -76,6 +81,15 @@ func NewBufferedLogger(logger log.Logger, handler Logger, bufferSize, workerCoun
 		bufferClosed: &atomic.Bool{},
 		flushedC:     make(chan struct{}),
 	}
+	_, err := meter.Int64ObservableGauge("cody-gateway.buffered_logger_buffer_size",
+		otelmetric.WithDescription("number of events in buffered logger buffer"), otelmetric.WithInt64Callback(func(ctx context.Context, io otelmetric.Int64Observer) error {
+			io.Observe(int64(len(res.bufferC)))
+			return nil
```
bobheadxi, line 87: Minor nit - `io` is commonly the import alias for the stdlib package

```suggestion
		otelmetric.WithDescription("number of events in buffered logger buffer"), otelmetric.WithInt64Callback(func(ctx context.Context, observer otelmetric.Int64Observer) error {
			observer.Observe(int64(len(res.bufferC)))
			return nil
```

bobheadxi (approved):
nice, LGTM 😁

cmd/cody-gateway/internal/events/buffered.go
rafax: Yeah, that was the "slowing down consumer" part that I shouldn't have pushed.

rafax: Fixed, thx for spotting this.

unknwon (approved):
Would I be in trouble if I don't approve this PR? 😁

PR #58381, opened 2023-11-16
web: navigate to file@commit instead of file@branch

stefanhengl:
With this change we always jump to blob@commit instead of blob@head if a user clicks on a search result. This avoids inconsistencies in line numbers if indexing lags behind repo syncing.

We keep the old logic, for displaying commit SHAs or branches on the search results page.

As a consequence
- file URLs will contain the commit hash (see updated tests)
- we will always display the commit hash (instead of the branch) at the top of the blob view, even if commit = HEAD.

image

While touching the code, I also refactored `getRevision`. I think the code is more readable now.

Test plan:
- manual testing
  - for a couple of searches, I checked the commit returned by the Stream API and verified that it matches the commit ID of the blob we display.  
- updated test


client/shared/src/search/stream.ts
```diff
@@ -579,21 +579,18 @@ export function getRepositoryUrl(repository: string, branches?: string[]): strin
...
 export function getRevision(branches?: string[], version?: string): string {
-    let revision = ''
-    if (branches) {
-        const branch = branches[0]
-        if (branch !== '') {
-            revision = branch
-        }
-    } else if (version) {
-        revision = version
+    if (branches && branches.length > 0) {
```
keegancsmith, line 582: I ain't know TS programmer but I think this can be shorted with the elvis operator:
```suggestion
    if (branches?.length > 0) {
```

client/shared/src/search/stream.ts
```diff
@@ -579,21 +579,18 @@ export function getRepositoryUrl(repository: string, branches?: string[]): strin
...
+    if (version) {
+        return version
+    }
+    return ''
 }
 
 export function getFileMatchUrl(fileMatch: ContentMatch | SymbolMatch | PathMatch): string {
-    const revision = getRevision(fileMatch.branches, fileMatch.commit)
+    const revision =
+        fileMatch.commit ?? (fileMatch.branches && fileMatch.branches.length > 0 ? fileMatch.branches[0] : '')
```
keegancsmith, line 593: you should document why you don't use getRevision and maybe add a docstring to getRevision saying it prefers branches to version.

Also elvis comes to the party again (I think)

```typescript
const revision = fileMatch.commit ?? fileMatch.branches?.[0]
```

keegancsmith (approved):
This is def changelog worthy.

client/shared/src/search/stream.ts
stefanhengl: not a TS programmer either, but I believe typescript doesn't let use compare "undefined" to a number literal, which would happen if branches is undefined.

stefanhengl:
rebased to pull in latest CHANGELOG changes.

jtibshirani:
I noticed some confusing behavior on S2 after this was merged. The search indexes are fully up-to-date, but clicking on a search result brings me to a specific commit SHA. I think this SHA actually represents HEAD (a commit only 2 hours ago), but our blob view shows the latest commit is from 3 days ago 🤔 

Screenshot 2023-11-20 at 1 01 35 PM

jtibshirani:
@stefanhengl I noticed some confusing behavior on S2 after this was merged. The search indexes are fully up-to-date, but clicking on a search result brings me to a specific commit SHA. I think this SHA actually represents HEAD (a commit only 2 hours ago), but our blob view claims this commit is from 3 days ago 🤔 

Screenshot 2023-11-20 at 1 01 35 PM

stefanhengl:
That's expected, but admittedly confusing. With the change, we always show the SHA, even if it represents HEAD. The blob view shows the date (3 days ago) of the last commit that changed this file. The commit message is shown next to it.

stefanhengl:
Suggestion: We add a little prefix "committed" or "authored" to the date, depending on whether we show the `committer.date` (default) or `author.date` (fallback). WDYT?

jtibshirani:
I see, thanks for the explanation! If I was confused, it's likely we'll have some confused users too! Could we avoid showing the hash if it's up-to-date? It feels best to maintain the previous "happy path" behavior that was clean and unsurprising.

PR #58384, opened 2023-11-16
Code Navigation: tokens highlight immediately on hover

jasonhawkharris:
Before, when a user hovered over a token with a tooltip, the immediate lack of highlighting created a state of uncertainty. No indication precedes the appearance of a tooltip, as they take several seconds to fully load upon the initial page load, and highlighting will not appear until they are fully loaded. This behavior contributed to a sluggish and somewhat frustrating user experience. 

Video of behavior before: 
https://github.com/sourcegraph/sourcegraph/assets/62355966/1824c361-2a9f-4762-a3fe-c51aa3057b71

Now, highlighting occurs instantly upon hover, signaling to the user that an interaction is imminent.

## Test plan
1. Test manually on various file types:
    - go
    - javascript
    - typescript
    - cpp
    - C
    - ruby
    - python
    - java
    - rust

### DIFF
--- a/client/branded/src/search-ui/input/codemirror/token-info.ts
+++ b/client/branded/src/search-ui/input/codemirror/token-info.ts
@@ -86,7 +86,7 @@ export function tokenInfo(): Extension {
                 effect.is(setHighlighedTokenPosition)
             )
             if (effect) {
-                position = effect?.value
+                position = effect.value
             }
             if (position !== null) {
                 // Mapping the position might not be necessary since we clear
--- a/client/wildcard/src/global-styles/highlight.scss
+++ b/client/wildcard/src/global-styles/highlight.scss
@@ -556,6 +556,21 @@
         color: var(--hl-pink);
     }
 
+    // The following hl-typed-* that have associated tooltips.
+    // This ensures that tokens highlight before tooltip is available.
+    .hl-typed-Identifier:hover,
+    .hl-typed-IdentifierParameter:hover,
+    .hl-typed-IdentifierAttribute:hover,
+    .hl-typed-IdentifierType:hover,
+    .hl-typed-IdentifierFunction:hover,
+    .hl-typed-IdentifierModule:hover,
+    .hl-typed-IdentifierFunctionDefinition:hover,
+    .hl-typed-IdentifierNamespace:hover,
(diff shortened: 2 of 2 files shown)


vovakulikov (approved):
I like the thought here about highlighting this beforehand, the only concern I have here - is it possible to have no code intel support for some files and still highlight them on hover, I suppose it's not the problem because if we don't have code intel that means that blob UI markup won't have these typed CSS classes and hence no hover effects but just double check. 

cc @fkling @jasonhawkharris

fkling:
@jasonhawkharris , @vovakulikov 

> Exactly. These classes will never be added to the tokens if code intel isn't present, so it's not an issue.

That's not quite right. These classes are added from syntax highlighting data. Code intel uses the same data to guess whether a token might have code intel data in the first place, but we have a lot of files that are syntax highlighted and don't have code intel data. [Markdown for example](https://sourcegraph.sourcegraph.com/github.com/golang/go/-/blob/README.md?view=code).

This commit seems to have already been pushed to S2 and you can see that it affects files without code intel (I'm hovering over the title): 


---

My suggestion is to revert this PR and add that functionality in CodeMirror that handles the hover logic. That would fix the "no codel intel available" problem because that's built into the API the hover extension uses to determine whether to show a tooltip or not. It'd also have two advantages:

- We are not duplicating token logic. The source of truth for "interactive tokens" is [this list](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@cb068f3c98d7115014ff6ae6860dea560942a7db/-/blob/client/web/src/repo/blob/codemirror/occurrence-utils.ts?L11-26), which is used to generate and add the CSS classes for syntax highlighting. If we duplicate the list in CSS we'd have to keep them in sync (and it's not obvious to do that because there are no comments that indicate to do so).
- We could potentially show a loading tooltip, just like we do for the selected token when you press space.

And another thing to consider: We purposefully wait a couple of milliseconds before we trigger the tooltip logic. This is so that the user can move their mouse over tokens without showing the tooltip immediately and potentially covering the position they wanted to reach. Maybe the token should only be highlighted when we actually trigger the tooltip logic, not before (as it is now)9

PR #58386, opened 2023-11-16
requestclient: add country-level geolocation data

bobheadxi:
This change adds country-level geolocation data to `requestclient` as discussed [here](https://sourcegraph.slack.com/archives/C01PQEVJ5A9/p1700078374822459?thread_ts=1694717740.084639&cid=C01PQEVJ5A9), provided as a 2- or 3-character ISO 3166-1 country code that seems pretty widely used. The inference strategy goes:

1. If we are behind Cloudflare, we trust [Cloudflare's originating country data](https://developers.cloudflare.com/fundamentals/reference/http-request-headers/#cf-ipcountry) and use it as-is
2. If a `Forwarded-For` exists, we look up the first `(*requestclient.Client).Forwarded-For` address in the embedded IP geolocation database
3. Otherwise, we look up `(*requestclient.Client).IP` in the embedded IP geolocation database

WE use [DB-IP.com](https://db-ip.com/)'s ["IP to Country Lite" database](https://db-ip.com/db/download/ip-to-country-lite) which includes only country-level data, is free, and allows redistribution with attribution (see docs included in PR). It clocks in at ~8MB, which seems manageable for embedding. Lookup performance looks okay with a trivial benchmark, and we cache the inferred country code in `*requestclient.Client` so that it only gets evaluated once:

```
goos: darwin
goarch: arm64
pkg: github.com/sourcegraph/sourcegraph/internal/requestclient/geolocation
BenchmarkInferCountryCode-10    	 2359490	       906.9 ns/op	      34 B/op	       3 allocs/op
```

This change also adds this new data to request client `LogFields`. https://github.com/sourcegraph/sourcegraph/pull/58388 adds this to telemetry events as well.

## Test plan

Unit tests


unknwon (approved):

bobheadxi:
* main
  * **PR #58386** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/58386?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #58388** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/58388?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/58386?utm_source=stack-comment).

internal/requestclient/geolocation/geolocation.go
```diff
@@ -0,0 +1,56 @@
...
+	ip2location "github.com/ip2location/ip2location-go/v9"
+
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+)
+
+//go:embed data/IP2LOCATION-LITE-DB1.IPV6.BIN/IP2LOCATION-LITE-DB1.IPV6.BIN
+var ip2locationDBBin []byte
+
+// locationsDB holds the ip2location database embedded at ip2locationDBBin.
+var locationsDB = func() *ip2location.DB {
```
keegancsmith, line 23: nit do this behind a sync.Once or our internal syncx.OnceFunc. We don't need this to run in the critical path of starting the execution of func main

internal/requestclient/geolocation/geolocation.go
```diff
@@ -0,0 +1,56 @@
...
+	// characters as the threshold to treat this as an error.
+	if len(code) > 3 {
+		return "", errors.Newf("IP database query failed: %s", code)
+	} else if len(code) == 0 {
+		return "", errors.New("no result found")
+	}
+	return code, nil
+}
+
+type noOpCloser struct{ *bytes.Reader }
```
keegancsmith, line 54: this is in the stdlib under io.NopCloser

internal/requestclient/client.go
```diff
@@ -12,14 +17,26 @@ type clientKey struct{}
...
+	// wafGeolocationCountryCode is a ISO 3166-1 alpha-2 country code for the
+	// request client as provided by a WAF (typically Cloudlfare) behind which
+	// Sourcegraph is hosted.
+	wafGeolocationCountryCode string
+
+	// countryCode and friends are lazily hydrated once by
+	// (*Client).OriginCountryCode().
+	countryCode      string
+	countryCodeError error
+	countryCodeOnce  sync.Once
```
keegancsmith, line 39: how often is LogFields called? Is it worth the extra complexity of using a once vs just computing it each time?

internal/requestclient/client.go
```diff
@@ -40,9 +57,53 @@ func (c *Client) LogFields() []log.Field {
 	if c == nil {
 		return []log.Field{log.String("requestClient", "<nil>")}
 	}
+
+	var ccField log.Field
+	if cc, err := c.OriginCountryCode(); err != nil {
+		ccField = log.String("requestClient.countryCode.error", err.Error())
```
mrnugget, line 63: Why not `log.NamedError`?

internal/requestclient/grpc_test.go
```diff
@@ -98,60 +101,27 @@ func TestPropagator(t *testing.T) {
...
 
 			resultCtx := propagator.InjectContext(requestCtx, md)
-			if diff := cmp.Diff(test.wantClient, FromContext(resultCtx)); diff != "" {
+			if diff := cmp.Diff(
+				test.wantClient,
+				requestclient.FromContext(resultCtx),
+				// Ignore unexported fields in Client - this test only tests
+				// the exported API surface (hence why we run this test in
+				// package requestclient_test)
+				cmpopts.IgnoreUnexported(requestclient.Client{}),
```
mrnugget, line 121: Hmm, wouldn't it be easier to name the 4 fields here vs. changing packing and renaming imports etc. and moving the test around?

internal/requestclient/geolocation/geolocation.go
bobheadxi: I couldn't use the io.NopCloser because the constructor this is used for expects at least an io.ReaderAt - will add a docstring!

internal/requestclient/client.go
bobheadxi: Eventually this will be called on all events as well (https://github.com/sourcegraph/sourcegraph/pull/58388)

dadlerj (commented):
I trust the devs to review :). Conceptually this looks great to me.

Just one question: how does this cc information added in the requestclient/client.go file actually get included in our telemetry payloads?

internal/requestclient/client.go
```diff
@@ -40,9 +57,53 @@ func (c *Client) LogFields() []log.Field {
...
+func (c *Client) OriginCountryCode() (string, error) {
+	c.countryCodeOnce.Do(func() {
+		c.countryCode, c.countryCodeError = inferOriginCountryCode(c)
+	})
+	return c.countryCode, c.countryCodeError
+}
+
+func inferOriginCountryCode(c *Client) (string, error) {
+	// If we have a trusted value already, use that directly.
+	if c.wafGeolocationCountryCode != "" {
```
dadlerj, line 87: Should we also check for != "XX" here (from the docs you linked below: https://developers.cloudflare.com/fundamentals/reference/http-request-headers/#cf-ipcountry)?

bobheadxi:
> Just one question: how does this cc information added in the requestclient/client.go file actually get included in our telemetry payloads?

@dadlerj  Added this in a separate PR, stacked on this one 😁 https://github.com/sourcegraph/sourcegraph/pull/58388

internal/requestclient/client.go
bobheadxi: Good catch, added a guard against this!

internal/requestclient/grpc_test.go
bobheadxi: Yeah, switched it back and made it an explicit comparison instead

internal/requestclient/geolocation/geolocation.go
bobheadxi: Updated to use `syncx.OnceValue`, which is super handy 😁

peterguy:
Not a blocking question, but what kind of plan should we have to update the database? Could probably handle it in releases or patches, but will probably need a prompt of some kind to remind us.
The [ip2location FAQ](https://www.ip2location.com/faqs) says we should update yearly, or maybe a bit more frequently: 
> Ownership of IP addresses change hands from time to time. Therefore, a small percentage of IP address blocks needs to be updated every year. Our database is updated monthly to make sure it is always accurate.

bobheadxi:
> Not a blocking question, but what kind of plan should we have to update the database? Could probably handle it in releases or patches, but will probably need a prompt of some kind to remind us.

Thanks - I added a mention of this in the `geolocation/data/README.md`. I'm not sure what we can do here except maybe set a reminder? I put one up in #team-data-analytics:

image

eseliger (approved):
code lgtm, left an inline question about a legal review

internal/requestclient/geolocation/data/README.md
```diff
@@ -0,0 +1,16 @@
+# Geolocation data
+
+This package contains a small geolocation database that gets embedded in the `internal/requestclient/geolocation` package.
+We currently use the "IP2Location LITE" `IP-COUNTRY` (`db1`) database, which only includes country-level data.
+IP2Location LITE allows for redistribution with the following acknowledgement on all "sites, advertising materials and documentation mentioning features or use of this database":
```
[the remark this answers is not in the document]
bobheadxi: Good idea - https://sourcegraph.slack.com/archives/C01GJFE26BX/p1700249357799719

bobheadxi:
I've updated this PR to use a database with a more generous license per legal feedback: https://sourcegraph.slack.com/archives/C01GJFE26BX/p1700249357799719

It's not the best, but should be fine for on-prem - in dotcom, we have Cloudflare headers we can depend on.

internal/requestclient/geolocation/data/README.md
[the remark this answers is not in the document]
bobheadxi: https://github.com/sourcegraph/sourcegraph/pull/58386#issuecomment-1823505642

bobheadxi:
https://github.com/sourcegraph/sourcegraph/pull/58515

PR #58396, opened 2023-11-17
build: Point to /search URL in help text

varungandhi-src:
Recently, the plain URL became the marketing site, so the previous message:

```
[web-standal...p] ===================================================
[web-standal...p]
[web-standal...p]         ✱ Sourcegraph is really ready now!
[web-standal...p]
[web-standal...p]      Click here: https://sourcegraph.test:3443
[web-standal...p]
[web-standal...p] ===================================================
```

Would take you to an empty page, potentially leading one to the incorrect
conclusion that something was wrong.

## Test plan

Run `SOURCEGRAPH_API_URL=https://sourcegraph.com sg start web-standalone`


valerybugakov (approved):
TY

PR #58397, opened 2023-11-17
vitest: Fix workspace config wrt client/web/

fkling:
It appears that the workspace configuration is somehow broken because it doesn't include any tests defined in `client/web/src/`. This can be tested by filtering for a test inside this subtree, for example

    pnpm exec vitest RepoRevisionSidebar

vitest will error saying it didn't find any tests. The list of projects it prints also doesn't include `web`.

I don't know what exactly it is that makes it exclude `client/web/` but with the simpler version in this PR `client/web/` will be included.

**NOTE:** Or is it intentional that `client/web/` is excluded? If so it's not obvious from the code.

I removed `client/web/dev/` from the list because it will be included by `client/web/` too and seems to "just work" even though it has it's own configuration file.

Before: 

After: 

## Test plan

`pnpm exec vitest RepoSidebarRevision` and `pnpm exec vitest` work.

### DIFF
--- a/vitest.workspace.ts
+++ b/vitest.workspace.ts
@@ -1,18 +1,16 @@
-import { readFileSync, existsSync } from 'fs'
+import { readFileSync } from 'fs'
 import path from 'path'
 
-import glob from 'glob'
 import { load } from 'js-yaml'
-import { defineWorkspace } from 'vitest/config'
 
 interface PnpmWorkspaceFile {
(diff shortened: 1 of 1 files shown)


burmudar (approved):

sqs (approved):
Weird behavior. Thanks for catching and fixing this.

PR #58398, opened 2023-11-17
github star/size exclusion: incorporate feedback from code review

mrnugget:
This is a follow-up to the review in https://github.com/sourcegraph/sourcegraph/pull/58377 and addresses all comments.

## Test plan

- Unit tests


umpox (approved):

mrnugget:
The original PR was about size/star exclusion, that's why I put it in this one, so it's clearer they're both related.

PR #58401, opened 2023-11-17
tests: add typestest.MakeExternalService helper and use it

mrnugget:
I noticed that we've started to copy&paste this

    types.ExternalService{
        Kind:   FOO
        Config: extsvc.NewUnencryptedConfig(MarshalJSON(t, conf)),
    }

into a lot of places, including the `MarshalJSON` and `NewUnencryptedConfig` calls.

Started to abstract these away and then noticed I can create a helper for the whole thing.

So here we are.

## Test plan

- existing tests


pjlast (approved):
I like that it removes some obscure function chaining of `NewUnencryptedConfig(MashalJSON...` 👍

burmudar (approved):

PR #58404, opened 2023-11-17
Route StarCoder traffic on the server to avoid downtime

philipp-spiess:
Two things going on here:

1. We remove the check for custom models to only work on dotcom. This works because the allowlisted strings are Cody Gateway string and we already have allowlisting on models as part of the license data. This is a req. for enabling StarCoder on non dot-com instances.
2. We add two new magic model identifier: `fireworks/starcoder-16b` and `fireworks/starcoder-7b`. When these are used, we let the SG instance decide which model to route the traffic to (based on the same feature flag we already use on the clients). 
    
    Why? This allows us to reroute the traffic effective immediately. The clients heavily cache the feature flag values to avoid network requests and currently only re-read these values when the instance is restarted. Having this knob on the server allows us to re-route without the client being aware of it. 

### With no feature flag

Screenshot 2023-11-17 at 16 04 53

### With feature flag off

Screenshot 2023-11-17 at 16 05 36

### With feature flag on

Screenshot 2023-11-17 at 16 05 47


rafax (approved):
LGTM to address the delay, we may want to use a separate (from feature-flag) for this in the future.

internal/completions/httpapi/codecompletion.go
```diff
@@ -36,14 +38,29 @@ func NewCodeCompletionsHandler(logger log.Logger, db database.DB) http.Handler {
...
-		return false
-	}
-
+func allowedCustomModel(ctx context.Context, model string) string {
 	switch model {
+	// These special model strings allow the server to choose the model. This allows us to instantly
+	// route traffic from Fireworks multi-tenant cluster to our single-tenant cluster and
+	// vice-versa, without the client having to know about it
+	case "fireworks/starcoder-16b",
+		"fireworks/starcoder-7b":
```
eseliger, line 47: If a customer instance has specifically said "claude only please" because that's the only service they approved, their code would now unexpectedly also be sent to fireworks, with no knob on the admin site they can tweak to prevent that. 
Is that okay? 

I don't have a strong opinion here, but I feel like customers might have.

rafax: Not sure I follow the Claude example - by default, an Enterprise customer would set the CompletionModel and that's what would get used on line 34, and this switch would only be used for a situation where the Cody extension has explicitly requested a "fireworks/starcoder-16b" model (which should not happen?)

eseliger: My understanding is that it's intentional that the cody extension requests models at-will where allowed to pick the best experience? 🤔 

Anyhow, can we be absolutely sure that we'll never fat finger in the client vs. adding some allow list to the completions config for "model overrides"?

rafax: I might be missing some context how model configuration works across Clients, Code Search and Cody Gateway - would keeping the "isDotCom check" help here? It might a topic for Monday, not Friday evening :) 

Longer term, I think we need to implement the "virtual model" lookup for StarCoder in Cody Gateway, so I'd be supportive of keeping the the DotCom check for now and putting a permanent solution in place during working hours.

eseliger: yep that'd help :) happy to chat Monday! also not meaning to block this PR if it's urgent, it won't go out in a release and we can follow up on this thread.

philipp-spiess: I removed the dotcom check mostly to unblock S2 and the demo instance. We haven't enabled fireworks for any other cody gateway deployment yet and this shouldn't affect the default?

My idea of "model configuration" is also that we will have an endpoint that we can call on the instance that will give us a list of allowed models before the client picks it. Right now this is not happening and we will default to not sending a model string _unless_ feature flags are enabled. If a custom instance enables these, they would still not be able to use starcoder because they won't have access to it via Cody Gateway.

Does that clear up the confusion or spark more? 😅

rafax: From my POV, if you need both a feature flag and a config change and the right licence tags, this looks safe, WDYT @eseliger ?

eseliger: yep this is fine for now - I'm wondering if we should add an allow list for enterprise for the GA though, to make this explicit and not depend on what a CE on our end configured in a mostly invisible system to the customer. :)

PR #58405, opened 2023-11-17
repo exclusion: switch to dynamic rule builder

mrnugget:
This is a follow-up to https://github.com/sourcegraph/sourcegraph/pull/58377 and changes how `exclude` works by AND'ing all conditions in a single rule together.

It does this mostly by refactoring the code and turning things upside down (inside out?):
- instead of each `[]exclude` entry in a code host connection config being turned into a separate condition, one of which had to be true
- each entry is turned into a `*rule` that can have multiple conditions and all of them have to be true for a repo to be excluded.

Best explained with an example.

Old code:

```go
var eb excludeBuilder
for _, r := range c.Exclude {
  eb.Exact(r.Name)
  eb.Exact(r.Id)
}
exclude, err := eb.Build()
if err != nil {
  return nil, err
}

// then:

exclude(r.Id) || exclude(r.Name) || exclude(r)
```

New code:

```go
var ex repoExcluder
for _, r := range c.Exclude {
  ex.AddRule().
    Exact(r.Name).
    Exact(r.Id)
}

// then:

ex.ShouldExclude(r.Id) || ex.ShouldExclude(r.Name) || ex.ShouldExclude(r)
```

Comparison not totally fair because stil need to add error handling for invalid patterns.

## Downsides

Currently it's a lot more function calls per exclusion-check... we could probably amortize the cost but I'm not sure yet what the nicest way is.

## Test plan

- unit tests, extended

### DIFF
--- a/internal/repos/exclude.go
+++ b/internal/repos/exclude.go
@@ -9,124 +9,127 @@ import (
 	"github.com/sourcegraph/sourcegraph/internal/bytesize"
 	"github.com/sourcegraph/sourcegraph/internal/extsvc/github"
 	"github.com/sourcegraph/sourcegraph/lib/errors"
-	"github.com/sourcegraph/sourcegraph/schema"
 )
 
 // excludeFunc takes either a generic object and returns true if the repo should be excluded. In
 // the case of repo sourcing it will take a repository name, ID, or the repo itself as input.
 type excludeFunc func(input any) bool
 
-// excludeBuilder builds an excludeFunc.
-type excludeBuilder struct {
-	exact    map[string]struct{}
-	patterns []*regexp.Regexp
-	generic  []excludeFunc
-	err      error
+// repoExcluder is made up of several rules that can be chained together to
+// exclude a repository.
+//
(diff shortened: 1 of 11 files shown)


internal/repos/exclude.go
```diff
@@ -163,7 +149,12 @@ func buildSizeConstraintsExcludeFn(constraint string) (gitHubExcludeFunc, error)
 		return nil, err
 	}
 
-	return func(r github.Repository) bool {
+	return func(input any) bool {
+		r, ok := input.(github.Repository)
+		if !ok {
+			return false
+		}
```
eseliger, line 156: wonder if it would be nice to encapsulate the excluder in a separate package, and then implement this special excludeFn in the github source file, I stumbled a bit over the fact that there's a github specific implementation in this otherwise quite generic file. Not needed for this PR.

mrnugget: Yeah. I keep thinking about this. I haven't solved the puzzle yet. I have an idea for all the "rules" to compile down to this, but I still need to play around with the ideas a bit.

PR #58415, opened 2023-11-17
Admin: improve search indexing icon

jtibshirani:
In #58061 we added 'Search indexing' to the repo dropdown menu, in place of
having two Embeddings items. However, I forgot to update the icon, so it was
using the old Embeddings icon. 

## Test plan

Manual testing


jtibshirani:
Here's what it looks like now. Very happy to change the icon if there's a better one, I couldn't find any precedent for what to use.
Screenshot 2023-11-17 at 9 51 30 AM

camdencheek (approved):

PR #58418, opened 2023-11-17
WIP: Automocking GraphQL responses

fkling:
tbd


fkling:
We now have our own mocking library in the Svelte app.

Issue #56848, opened 2023-09-20
msp: implement resource.sql


bobheadxi:
Eventually, we'll likely have a service (product subscriptions, accounts) that requires a real database. Cloud SQL has been used extensively with success in Sourcegraph Cloud - we should offer an integration to MSP.

Issue #56846, opened 2023-09-20
msp: runtime utilities


bobheadxi:
Right now we leverage a bunch of shared Sourcegraph code for service init, etc. Many of these are tailored for in-Sourcegraph-deployment usage - whereas managed services tend to be external and have different patterns (e.g. no site config/very rare dynamic config). Additionally, we use OpenTelemetry for metrics instead of native Prometheus, for ease of integration with push metrics to Cloud monitoring.

there's also a few "service contracts" that have come up in https://docs.google.com/document/d/1QINfjijEdiKvHVUjVvEaKeteXpEzob5VBgwL_Q0B870/edit#bookmark=id.g33lckk1i5h0 and https://github.com/sourcegraph/sourcegraph/pull/55905 (i.e. https://github.com/sourcegraph/sourcegraph/pull/55905/files#r1300848570)

We should create a set of service init and other tooling in `internal/managedservicesplatform` for services to use and share.

- [x] Goroutine management
- [x] Tracing/metrics initialization
- [x] Expose Redis/SQL/etc configuration
- [x] Sentry initializtion

bobheadxi:
All the initial wishlist items are in place, and this is available via `lib/managedservicesplatform/runtime` now. Marking as done!

PR #58420, opened 2023-11-17
dev/msp: add CloudSQL integration

bobheadxi:
This PR has 3 major changes:

1. Provisioning a Cloud SQL PostgreSQL instance: `resource.postgreSQL`
2. Updates to the MSP runtime for connecting to the database
3. Publishable `msp-example` for the `msp-testbed` deployment, for testing integrations and other MSP features, and adds an example Gorm setup

Closes https://github.com/sourcegraph/sourcegraph/issues/56848. For more context, see go/msp

### `resource.postgreSQL`

This change adds `resource.postgreSQL` to the spec for provisioning a Cloud SQL database. Example:

```yaml
  resources:
    postgreSQL:
      databases: ["primary"]
      cpu: 1
      memoryGB: 4
```

When configured, a Cloud SQL instance is provisioned, and the Cloud Run workload SA is granted access to it. The specified `databases` are also created. `cpu` and `memoryGB` are used to generate the custom VM to provision the instance with. A lot of the base configuration for the database instance are taken directly from what we use in Cloud: https://github.com/sourcegraph/controller/blob/9cac1b1aa0dde038280d424f4d4b6c34ffb22e36/internal/resource/sql/sql.go#L116, and is similar to the Redis setup, in that we need the private peering VPC and install additional certs on the Cloud Run instance.

### MSP runtime

`service.Contract` now has a helper, `GetPostgreSQLDB(database)`, for generating the appropriate configuration for creating a connection to a given database and returning a `sql.DB`, which most libraries have adapters for. The helper sets up `pgx.ConnConfig` that uses workload identity to connect directly to Cloud SQL, following the guide here: https://github.com/GoogleCloudPlatform/cloud-sql-go-connector?tab=readme-ov-file#automatic-iam-database-authentication

In local dev, `PGDSN` can be used to hardcode an appropriate value.

### `msp-example`

I've moved the MSP example from #56846 to the top-level `cmd/msp-example,` and set up some guidance for publishing it. This service now also includes a simple database connection and setup using Gorm, used by https://github.com/sourcegraph/accounts.sourcegraph.com, and can be used for testing, e.g. https://github.com/sourcegraph/managed-services/pull/87 

## Test plan

- [x] Deploy https://github.com/sourcegraph/managed-services/pull/87, which will connect to the database and set up some tables
- [x] Review diff: https://github.com/sourcegraph/managed-services/actions/runs/6951246277
- [x] Tear down database in https://github.com/sourcegraph/managed-services/pull/87, with caveat: requires state surgery. Should be rare enough (we might prefer to tear down environments entirely) - see docstrings in package `postgresqlroles`


dev/managedservicesplatform/spec/environment.go
```diff
@@ -299,6 +308,19 @@ type EnvironmentResourceRedisSpec struct {
...
+	// Databases to provision - required.
+	Databases []string `json:"databases"`
+	// Defaults to 1.
+	CPU *int `json:"cpu,omitempty"`
+	// Defaults to 1.
+	MemoryGB *int `json:"memoryGB,omitempty"`
+
+	// DisableDeletionProtection disables the Cloud SQL deletion protection that
+	// we enable by default. Only set to true if tearing down the database.
+	DisableDeletionProtection bool `json:"disableDeletionProtection,omitempty"`
```
michaellzc, line 321: i prefer to just set `deletion_protection=false` at all time to save some time during teardown. it doesn't actually prevent one from deleting the instance from UI

dev/managedservicesplatform/internal/resource/cloudsql/cloudsql.go
```diff
@@ -0,0 +1,176 @@
...
+				Name:  pointers.Ptr("cloudsql.iam_authentication"),
+				Value: pointers.Ptr("on"),
+			}},
+
+			// 🚨SECURITY🚨 SOC2/CI-79
+			// Production disks for managed instances are configured with daily snapshots and retention set at ninety days,
+			// so we do the same.
+			BackupConfiguration: &sqldatabaseinstance.SqlDatabaseInstanceSettingsBackupConfiguration{
+				Enabled:                     pointers.Ptr(true),
+				PointInTimeRecoveryEnabled:  pointers.Ptr(true),
```
michaellzc, line 81: PITR is going to increase a lot of storage consumption, and the recovery process is cumbersome (it always requires restoring to a new instance).

if we don't need it, consider disabling it to start.

dev/managedservicesplatform/internal/resource/cloudsql/cloudsql.go
```diff
@@ -0,0 +1,176 @@
...
+	instanceCACert := gsmsecret.New(scope, id.Group("ca-cert"), gsmsecret.Config{
+		ProjectID: config.ProjectID,
+		ID:        strings.ToUpper(id.DisplayName()) + "_CA_CERT",
+		Value:     *instance.ServerCaCert().Get(pointers.Float64(0)).Cert(),
+
+		// instanceCACert is required to connect to this instance, so do ensure
+		// database resources are all fully provisioned, we gate the availability
+		// of this secret on all resources being ready.
+		DependsOn: databaseResources,
+	})
```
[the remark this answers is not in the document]
bobheadxi: Struggling to figure this one out - it seems maybe via private IP you can't use the cloud-sql-go-connector? I only get errors so far:

```
initDB: gorm.Open: failed to connect to `host=/tmp user=msp-testbed-sa@msp-testbed-test-8141b6810cda.iam database=primary`: server error (FATAL: Cloud SQL IAM service account authentication failed for user "msp-testbed-sa@msp-testbed-test-8141b6810cda.iam" (SQLSTATE 28000))
```

https://cloud.google.com/sql/docs/postgres/connect-run#private-ip doesn't list the connector. Maybe the better way is to use public IP and the connector? or use a user directly via private? not sure if I can even use the service account user in that case - would I have to use a regular user/password user?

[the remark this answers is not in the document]
michaellzc: I think you're just missing this role `roles/cloudsql.instanceUser` for the GSA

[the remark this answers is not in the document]
bobheadxi: Nice, that worked! Now the Gorm migrator is failing on:

```
initDB: auto migrating table for *example.User: ERROR: permission denied for schema public (SQLSTATE 42501)
``

looking into it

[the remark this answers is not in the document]
Comment (michaellzc) [reply dev/managedservicesplatform/internal/resource/cloudsql/cloudsql.go]: yeah, that's expected. on cloud, we have to use the `postgresql` provider to fix up the permission for IAM user using the admin credential.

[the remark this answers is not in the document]
Comment (bobheadxi) [reply dev/managedservicesplatform/internal/resource/cloudsql/cloudsql.go]: Added the perms, but am now running into 

```
initDB: gorm.Open: failed to connect to `host=/tmp user=msp-testbed-sa@msp-testbed-test-8141b6810cda.iam database=primary`: dial error (Dial error: failed to dial (connection name = "msp-testbed-test-8141b6810cda:us-central1:postgresql-2b57"): dial tcp 10.57.0.3:3307: connect: connection timed out)
```

Funnily enough, an older revision was able to connect top and set up the database correctly, but I'm unable to figure out what is in that older revision... and now I can't get it right again 😅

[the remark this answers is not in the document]
Comment (bobheadxi) [reply dev/managedservicesplatform/internal/resource/cloudsql/cloudsql.go]: Ah shoot, turns out it's becasue I accidentally detached Cloud Run from the private network. 🤦

File: dev/managedservicesplatform/internal/resource/cloudsql/cloudsql.go [hunk 4]
```diff
@@ -0,0 +1,193 @@
...
+
+	// Additional configuration directly via Postgres
+	pgProvider := postgresql.NewPostgresqlProvider(scope, id.TerraformID("postgresql_provider"), &postgresql.PostgresqlProviderConfig{
+		Scheme:    pointers.Ptr("gcppostgres"),
+		Host:      instance.ConnectionName(),
+		Username:  adminUser.Name(),
+		Password:  adminPassword.Result(),
+		Port:      jsii.Number(5432),
+		Superuser: jsii.Bool(false),
+	})
```
michaellzc, line 179: There's a gotcha here.

last time we tried mixing `postgresql` provider and the db instance in the same module, it will always result in some weird error at the end of `terraform destroy`, tho I think all resources are actually teardown.

it's been a while tho, maybe you can test out destroy again.

bobheadxi: I think you're right, even tearing down the database won't work. Hmmm

michaellzc: yeah, that's why we have the `sql` and `sqlschema` stack to prevent chicken-and-egg problem

bobheadxi: Solved to a "good enough" state with some state surgery: https://github.com/sourcegraph/sourcegraph/pull/58420/commits/e3c8d607c37aad0f6d8f2cf77ed2760932501494

michaellzc (approved):

cmd/msp-example/internal/example/example.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/sourcegraph/sourcegraph/lib/managedservicesplatform/service"
+)
+
+type Config struct {
+	StatelessMode bool
+	Variable      string
+}
+
+func (c *Config) Load(env *service.Env) {
+	c.StatelessMode = env.GetBool("STATELESSMODE", "false", "if true, disable dependencies")
```
unknwon, line 24: nit

```suggestion
	c.StatelessMode = env.GetBool("STATELESS_MODE", "false", "if true, disable dependencies")
```

dev/managedservicesplatform/spec/environment.go
```diff
@@ -256,14 +256,20 @@ type EnvironmentJobScheduleSpec struct {
...
 	//
 	//  - REDIS_ENDPOINT
 	//
 	// Sourcegraph Redis libraries (i.e. internal/redispool) will automatically
 	// use the given configuration.
 	Redis *EnvironmentResourceRedisSpec `json:"redis,omitempty"`
+	// PostgreSQL, if provided, provisions a PostgreSQL database instance backed
+	// by CloudSQL.
+	//
+	// TODO
```
unknwon, line 271: Still a TODO?

unknwon (commented):
29/37

dev/managedservicesplatform/internal/resource/cloudsql/cloudsql.go
```diff
@@ -0,0 +1,194 @@
...
+	// Network to connect the created Cloud SQL instance to.
+	Network computenetwork.ComputeNetwork
+
+	DependsOn []cdktf.ITerraformDependable
+}
+
+func New(scope constructs.Construct, id resourceid.ID, config Config) (*Output, error) {
+	machineType := fmt.Sprintf("db-custom-%d-%d",
+		pointers.Deref(config.Spec.CPU, 1),
+		pointers.Deref(config.Spec.MemoryGB, 4)*1024)
```
unknwon, line 54: Does it mean there are only specific list of CPU/RAM combinations we can choose from, or it doesn't really matter? i.e. could a service to have 8 CPU and 2 GB RAM?

unknwon (approved):

dev/managedservicesplatform/internal/resource/cloudsql/cloudsql.go
bobheadxi: > Does it mean there are only specific list of CPU/RAM combinations we can choose from, or it doesn't really matter? i.e. could a service to have 8 CPU and 2 GB RAM?

It's pretty free-form, with the following restrictions:

image

I'll add some validation 😁

dev/managedservicesplatform/spec/environment.go
bobheadxi: Updated!

cmd/msp-example/README.md
```diff
@@ -0,0 +1,9 @@
+# msp-example
+
+This service is an example service testing and demonstrating how to build Managed Services Platform services: [go/msp](https://handbook.sourcegraph.com/departments/engineering/teams/core-services/managed-services/platform/)
+
+## Pushing a new version
+
+```sh
+bazel run //cmd/msp-example:candidate_push --config darwin-docker --stamp -- --tag insiders --repository us.gcr.io/sourcegraph-dev/msp-example
```
jhchabran, line 8: You probably want to precise that the `--config darwin-docker` is just there for mac users. 

You're stamping here, but you're not providing a `--workspace_status_command`, what are we stamping with exactly?

cmd/msp-example/image_test.yaml
```diff
@@ -0,0 +1,15 @@
+schemaVersion: "2.0.0"
+
+commandTests:
+  - name: "binary is runnable"
```
jhchabran, line 4: I'm glad you kept that, it's rather easy otherwise to ship binaries with the wrong arch 😅

cmd/msp-example/BUILD.bazel
```diff
@@ -0,0 +1,63 @@
...
+    user = "sourcegraph",
+)
+
+oci_tarball(
+    name = "image_tarball",
+    image = ":image",
+    repo_tags = ["msp-example:candidate"],
+)
+
+container_structure_test(
```
jhchabran, line 47: Just in case, FYI, were you to remove this, the CI would simply not build this code (assuming `bazel test //...`) because there are no other tests. 

You could add a [`build_test`](https://github.com/bazelbuild/bazel-skylib/blob/main/docs/build_test_doc.md) so you can iterate in dev on this, without having to rely on the `--config darwin-docker` (which is not very nice). 

This is just a nit, it's totally ok as it currently is :)

dev/managedservicesplatform/internal/resource/postgresqlroles/postgresqlroles.go
```diff
@@ -0,0 +1,58 @@
...
+//
+//	terraform state list | grep postgresql_grant_role | xargs terraform state rm
+//
+// This is because we cannot instantiate the provider when removing the
+// database, causing plans and applies to fail. We'll likely be stuck with the
+// workaround for a while, which is acceptable because CloudSQL-only teardowns
+// should be rare - we'll more likely be removing entire environments in general.
+//
+// TODO(@bobheadxi): Improve documentation around this teardown scenario.
+func New(scope constructs.Construct, id resourceid.ID, config Config) (*Output, error) {
```
jhchabran, line 38: `New(...)` functions are usually expected to not cause side effects, and reading the doc comment here, `Apply` or `Create` are much clearer alternatives I believe.

jhchabran: Ack, makes sense then :)

bobheadxi: So the actual things happening is done after the fact by Terraform. That said, the docstrings do all sound like something is happening immediately, so I might make a pass and update them

cmd/msp-example/README.md
bobheadxi: Not gonna lie, I wasn't 100% sure what `--stamp` does, but I guess it's related to setting versions and things 😁 

I will follow up to add `--workspace_status_command` - I think we should still have it to see what version is up

jhchabran: TL;DR stamping is the mechanism you can use to inject "status variables", such as the version, or anything you want to inject in your build. It can be either _stable_ or _unstable_ (if a stable var changes, it will trigger a target to be rebuilt, as opposed to the unstable ones, where it will be injected at build time, but won't trigger a rebuild: useful when you're stamping with the build time for example). 

The way you pass those vars is through the workspace_status_command, which is a shell scripts that produces those definitions.

PR #58421, opened 2023-11-17
Code Navigation: different icons for different file types

jasonhawkharris:
Previously, each file in the tree view featured a default icon, causing challenges for users trying to swiftly scan and discern the contents of the files.
Screenshot 2023-11-17 at 11 47 55 AM

Now, files exhibit distinctive icons corresponding to their file types. Furthermore, test files are easily identifiable through the inclusion of a superimposed indicator on the bottom right corner of the corresponding icon.

Video: 
https://github.com/sourcegraph/sourcegraph/assets/129280/87210e2d-e9c5-441e-b4c5-adf16b4da46a

## Test plan
Manual testing:
1) check that every file type has the correct icon associated with it when rendered in the tree.

Ships with unit tests for helper functions


client/web/src/repo/constants.ts
```diff
@@ -13,3 +59,113 @@ export enum CodeHostType {
     AZUREDEVOPS = 'azureDevOps',
     OTHER = 'other',
 }
+
+export enum FileExtension {
```
peterguy, line 63: Is this list comprehensive for the icons included in `react-icons`?

jasonhawkharris: Not quite. But I've got the most popular languages in there and even some unpopular ones. I made it easy to update as we see fit.

jasonhawkharris:
It is! @taiyab you cool with it?

client/web/src/repo/utils.tsx
```diff
@@ -45,3 +47,37 @@ export const stringToCodeHostType = (codeHostType: string): CodeHostType => {
...
     }
 }
+
+export interface FileInfo {
+    extension: FileExtension
+    isTest: boolean
+}
+
+export const getFileInfo = (file: string, isDirectory: boolean): FileInfo => {
+    const fileInfo = { extension: '' as FileExtension, isTest: false }
```
BolajiOlajide, line 57: I'm a bit concerned due to the fact we are typecasting potential incorrect values as `FileExtension`, an empty string isn't of type `FileExtension` and I feel we shouldn't be casting it here.

client/web/src/repo/utils.tsx
```diff
@@ -45,3 +47,37 @@ export const stringToCodeHostType = (codeHostType: string): CodeHostType => {
...
+    const f = file.split('.')
+    // Last item in 'f' is file extension string
+    fileInfo.extension = f.pop() as FileExtension
+    return fileInfo
+}
+
+export const containsTest = (file: string): boolean => {
+    const f = file.split('.')
+    const matchTest1 = '_test'
+    const matchTest2 = 'test_'
```
BolajiOlajide, line 73: Some javascript/typescript environments prefix their test files with `_spec`, so maybe we add a case for that too.

client/web/src/repo/utils.tsx
```diff
@@ -45,3 +47,37 @@ export const stringToCodeHostType = (codeHostType: string): CodeHostType => {
...
+    for (const i of f) {
+        if (i === 'test') {
+            return true
+        }
```
BolajiOlajide, line 77: I don't think we should have this check since it's potentially going to yield false positives.
For example. `vitest.js`, `testSetup.js` and `protest_controller.js` will return `true` even though they aren't actual test files

client/web/src/repo/utils.tsx
```diff
@@ -45,3 +47,37 @@ export const stringToCodeHostType = (codeHostType: string): CodeHostType => {
...
+    const fileInfo = { extension: '' as FileExtension, isTest: false }
+    fileInfo.isTest = isDirectory ? false : containsTest(file)
+
+    if (isDirectory) {
+        return fileInfo
+    }
+
+    const f = file.split('.')
+    // Last item in 'f' is file extension string
+    fileInfo.extension = f.pop() as FileExtension
```
BolajiOlajide, line 66: Is the assumption that a filename `Dockerfile` has no extension or that the extension is `Dockerfile`?

If the former is true, then this is potentially confusing. I'm also worried about typecasting here, we could use a switch case to return the actual `FileExtension` or undefined if the icon library doesn't support the extension.

BolajiOlajide, line 66: Proposing this change here

```suggestion
export const getFileInfo = (file: string, isDirectory: boolean): FileInfo => {
    const isTest = isDirectory ? false : containsTest(file)

    if (isDirectory) {
        return { isTest, extension: undefined }
    }

    const f = file.split('.')
    // Last item in 'f' is file extension string
    const extension = converTotFileExtensionEnum(f.pop())
    return { extension, isTest }
    
```

BolajiOlajide (approved):
Approving to unblock.

client/web/src/repo/utils.tsx
jasonhawkharris: This conditional covers any test that is formatted like so:
`file.test.js`

It will not return true if `i` === `vitest.js`, `testSetup.js` or `protest_controller.js`

The first thing we do in the function is split on the `.` character. Then this conditional checks if there is an item in the resulting array that simply says "test". 

That won't be the case for files like `vitest.js` because split at the `.` character, it will return `[vitest, js]`. Since there is no item in the array exactly equal to `test`, it will skip the conditional and return false unless the other conditionals interupt the flow.

Whereas `file.test.js` split at the `.` will return `[file, test, js]`. Since there is an item that exactly equals test in this array, it will return true.  

Does that make sense?

client/web/src/repo/utils.tsx
jasonhawkharris: Exactly. Dockerfile doesn't have a file extension. To associate it with a docker icon, we need to account for it which is why I've added it to the File Extensions object.

jasonhawkharris: Marking this as resolved even though I'm not taking the suggestion verbatim. However, I am going to change this so that we're not casting an empty string as an enum that doesn't exist.

client/web/src/repo/utils.tsx
BolajiOlajide: Oh! My bad - I glanced through the `file.split` and didn't take note of that.
Thanks for clarifying.

client/web/src/repo/utils.tsx
jasonhawkharris: I was using a switch case at first, but I switched it out for a Hashmap because it was quite a bit more performant. I'm adding a DEFAULT case to the enum. That should fix this.

BolajiOlajide: Got it.

client/web/src/repo/utils.tsx
jasonhawkharris: Done! Good idea.

client/web/src/repo/utils.tsx
jasonhawkharris: Fixing this!

client/web/src/repo/utils.tsx
```diff
@@ -45,3 +47,38 @@ export const stringToCodeHostType = (codeHostType: string): CodeHostType => {
...
+    const matchTest1 = '_test'
+    const matchTest2 = 'test_'
+    const matchTest3 = '_spec'
+    for (const i of f) {
+        if (i === 'test') {
+            return true
+        }
+        if (includes(i, matchTest1) || includes(i, matchTest2) || includes(i, matchTest3)) {
+            return true
+        }
```
fkling, line 81: There is no need to use lodash's `include` function here. You can do native substring test with `i.indexOf(matchText1) > -1`. However, this is one of these cases where regular expressions are useful. How the expression would look like depends on what you want to test. Translating the currently logic to a regular expression would look something like this:

```
for (const i of f) {
   if (/^test$|_test|test_|_spec/.test(i)) {
     return true
   }
}
```

But I don't think that that's a very good test. E.g. it would return `true` for a file with name `my_testimonials.txt`.

A better test IMO is to check whether either the segment is `test` or `spec` (you already do that for `test`) or starts/ends with `test_`/`spec_`/`_test`/`_spec`. For this we need to anchor the expression to the start (`^`) or the end (`$`) of the string.
I would even remove the explicit check for `_`. Maybe somebody uses `-`. You can check whether `test` is followed by a *word-boundary* character.

So the expression could be:

    ^(test|spec)(\b|_|$)|(\b|_)(test|spec)$

Which reads as *string starts with (test or spec) followed by (word-boundary, _ or the end of the string), or string contains (word-boundary or _) followed by (test or spec) at the end of the string*. We still need to check for `_` explicitly because it's not considered a word-boundary, but we will now take e.g. `-` and `+` into account too.

Having said all that, we need to acknowledge that naming conventions for test files vary and ideally we can somehow make this configurable.

Also I could imagine that false positive are a bad UX, i.e. a file is detected and displayed as a test but it isn't actually a test.

vovakulikov: A minor comment from me, not exactly related to Felix's comment above, but also not sure what we should do with test-related directory files. For example, __snapshots__ or fixture or something similar. Maybe we should run regexp check not only over the filename but over the whole file path to take into account directories.

vovakulikov (commented):
Tech review is ok, but there is an interesting message from Felix that I think we should resolve before merging, will test it manually today

client/web/src/repo/RepoRevisionSidebarFileTree.tsx
```diff
@@ -470,12 +474,25 @@ function renderNode({
                 }
             }}
         >
-            <Icon
-                svgPath={isBranch ? (isExpanded ? mdiFolderOpenOutline : mdiFolderOutline) : mdiFileDocumentOutline}
-                className={classNames('mr-1', styles.icon)}
-                aria-hidden={true}
-            />
-            {name}
+            <div className={classNames(styles.iconContainer)}>
```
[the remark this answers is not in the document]
jasonhawkharris: done!

vovakulikov:
Tested manually, @jasonhawkharris @taiyab gray dot - test indicator seems a bit off. I got it as some visual bug at the beginning. I think if we don't want to highlight tests-like files, this is fine; we can go without dot indicator but if we want to, I think we need to work with item's backgrounds without additional elements over the icon IMO

Screenshot 2023-11-28 at 12 13 19


I thought about it like something like this (just an idea first green-like colors that I found in our colors tokens) 

Screenshot 2023-11-28 at 12 23 06

jasonhawkharris:
@vovakulikov When you say it's a "bit off" do you mean literally? Like... the gray dot is offset from the icon it is on top of? If so, that was intentional. 

OR do you just mean you don't think idea is quite right.

jasonhawkharris:
> Tested manually, @jasonhawkharris @taiyab gray dot - test indicator seems a bit off. I got it as some visual bug at the beginning. I think if we don't want to highlight tests-like files, this is fine; we can go without dot indicator but if we want to, I think we need to work with item's backgrounds without additional elements over the icon IMO
> 
> Screenshot 2023-11-28 at 12 13 19
> I thought about it like something like this (just an idea first green-like colors that I found in our colors tokens)
> 
> Screenshot 2023-11-28 at 12 23 06

Thoughts on this @taiyab?

PR #58422, opened 2023-11-18
SSC: Rate limit based on the user's plan

vdavid:
Adds four new config items ({chat, code completions} x {Community, Pro} = 4).
Using `user.CodyProEnabledAt` to determine the user's plan.
Returns the matching limit and interval based on this.

The requirement for the interval is "from the Nth day of this month until the Nth day of the next month", which the current architecture doesn't support, so I'm just using a 30-day interval to indicate "1 month", and the Cody Gateway side will need to recognize this and apply the "Nth-to-Nth" logic.

## Test plan

Currently, it's feature-flagged with `cody-pro-dec-ga`, which is off, so we're good to test this later.

### DIFF
--- a/cmd/frontend/graphqlbackend/user.go
+++ b/cmd/frontend/graphqlbackend/user.go
@@ -175,7 +175,7 @@ func (r *UserResolver) CodyProEnabledAt(ctx context.Context) *gqlutil.DateTime {
 		return nil
 	}
 
-	if !featureflag.FromContext(ctx).GetBoolOr("cody-pro-dec-ga", false) {
+	if !featureflag.FromContext(ctx).GetBoolOr("cody-pro", false) {
 		return nil
 	}
 
@@ -329,7 +329,7 @@ func (r *schemaResolver) UpgradeToCodyPro(ctx context.Context, args *upgradeToCo
 		return nil, errors.New("this feature is only available on sourcegraph.com")
 	}
 
-	if !featureflag.FromContext(ctx).GetBoolOr("cody-pro-dec-ga", false) {
+	if !featureflag.FromContext(ctx).GetBoolOr("cody-pro", false) {
 		return nil, errors.New("this feature is not enabled")
 	}
 
--- a/cmd/frontend/graphqlbackend/user.go
+++ b/cmd/frontend/graphqlbackend/user.go
@@ -175,7 +175,7 @@ func (r *UserResolver) CodyProEnabledAt(ctx context.Context) *gqlutil.DateTime {
 		return nil
 	}
 
-	if !featureflag.FromContext(ctx).GetBoolOr("cody_pro_dec_ga", false) {
+	if !featureflag.FromContext(ctx).GetBoolOr("cody-pro-dec-ga", false) {
 		return nil
 	}
 
@@ -329,7 +329,7 @@ func (r *schemaResolver) UpgradeToCodyPro(ctx context.Context, args *upgradeToCo
 		return nil, errors.New("this feature is only available on sourcegraph.com")
 	}
 
-	if !featureflag.FromContext(ctx).GetBoolOr("cody_pro_dec_ga", false) {
+	if !featureflag.FromContext(ctx).GetBoolOr("cody-pro-dec-ga", false) {
 		return nil, errors.New("this feature is not enabled")
 	}
 
--- a/cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go
+++ b/cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	sgactor "github.com/sourcegraph/sourcegraph/internal/actor"
+	"github.com/sourcegraph/sourcegraph/internal/conf/conftypes"
 	"github.com/sourcegraph/sourcegraph/internal/featureflag"
 	"math"
 
@@ -194,8 +195,8 @@ func getCompletionsRateLimit(ctx context.Context, db database.DB, userID int32, 
 	var limit *int
 	var err error
 
+	// Apply overrides first.
 	source := graphqlbackend.CodyGatewayRateLimitSourceOverride
-
 	switch scope {
 	case types.CompletionsFeatureChat:
 		limit, err = db.Users().GetChatCompletionsQuota(ctx, userID)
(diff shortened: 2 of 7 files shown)


internal/conf/conftypes/consts.go
```diff
@@ -12,11 +12,15 @@ type CompletionsConfig struct {
...
-	PerUserCodeCompletionsDailyLimit int
+	AccessToken                                 string
+	Provider                                    CompletionsProviderName
+	Endpoint                                    string
+	PerUserDailyLimit                           int
+	PerUserCodeCompletionsDailyLimit            int
+	PerUserCommunityMonthlyLimit                int
+	PerUserCommunityCodeCompletionsMonthlyLimit int
+	PerUserProDailyLimit                        int
+	PerUserProCodeCompletionsDailyLimit         int
```
vdavid, line 23: New items

cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go
```diff
@@ -212,7 +269,7 @@ func getCompletionsRateLimit(ctx context.Context, db database.DB, userID int32,
 				limit = pointers.Ptr(cfg.PerUserCodeCompletionsDailyLimit)
 			}
 		default:
-			return licensing.CodyGatewayRateLimit{}, graphqlbackend.CodyGatewayRateLimitSourcePlan, errors.Newf("unknown scope: %s", scope)
+			return licensing.CodyGatewayRateLimit{}, graphqlbackend.CodyGatewayRateLimitSourcePlan, errors.Newf("unknown scope (dotcom limiting): %s", scope)
```
vdavid, line 272: To make this error text unique for easier debugging

cmd/frontend/graphqlbackend/user.go
```diff
@@ -175,7 +175,7 @@ func (r *UserResolver) CodyProEnabledAt(ctx context.Context) *gqlutil.DateTime {
 		return nil
 	}
 
-	if !featureflag.FromContext(ctx).GetBoolOr("cody_pro_dec_ga", false) {
+	if !featureflag.FromContext(ctx).GetBoolOr("cody-pro-dec-ga", false) {
```
vdavid, line 178: Renamed feature flag to match our common style

chrsmith: Minor nit: It may be preferable to just call this "cody-pro" or "cody-pro-v1". If (heaven forbid) that Dec date changes, then everyone will feel a little awkward 😬 .

chrsmith (commented):
A couple of minor nits. I'll defer to @thenamankumar for approval. I'll ping you on Slack for a few other meta questions.

cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go
```diff
@@ -198,6 +207,45 @@ func getCompletionsRateLimit(ctx context.Context, db database.DB, userID int32,
...
+		actor := sgactor.FromContext(ctx)
+		user, err := actor.User(ctx, db.Users())
+		if err != nil {
+			return licensing.CodyGatewayRateLimit{}, graphqlbackend.CodyGatewayRateLimitSourcePlan, err
+		}
+		isProUser := user.CodyProEnabledAt != nil
+
+		cfg := conf.GetCompletionsConfig(conf.Get().SiteConfig())
+		switch scope {
+		case types.CompletionsFeatureChat:
```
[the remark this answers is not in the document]
vdavid: Thanks, I actually craved to do this extraction, but coming mostly from a TypeScript world where a function longer than 10–15 lines is frowned upon, Go code often looks spaghetti to me 😅 so I was like, I'll be idiomatic with the rest, and I resisted the temptation to pull it out. But then maybe it's okay to make it a bit more organized.
Extracted it in https://github.com/sourcegraph/sourcegraph/pull/58422/commits/e5cb2107f58c4500fbcc82c76ea3087940c5823d

cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go
```diff
@@ -198,10 +208,27 @@ func getCompletionsRateLimit(ctx context.Context, db database.DB, userID int32,
...
 		return licensing.CodyGatewayRateLimit{}, graphqlbackend.CodyGatewayRateLimitSourcePlan, err
 	}
+
+	// If there's no override, check the self-serve limits.
+	cfg := conf.GetCompletionsConfig(conf.Get().SiteConfig())
+	intervalSeconds := oneDayInSeconds
+	if limit == nil && cfg != nil && featureflag.FromContext(ctx).GetBoolOr("cody-pro", false) {
+		source = graphqlbackend.CodyGatewayRateLimitSourcePlan
+		actor := sgactor.FromContext(ctx)
+		user, _ := actor.User(ctx, db.Users())
```
thenamankumar, line 218: You are doing `user, _` and in next line checking `if err != nil`. It should be `user, err`

cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go
```diff
@@ -198,10 +208,27 @@ func getCompletionsRateLimit(ctx context.Context, db database.DB, userID int32,
...
+		if err != nil {
+			return licensing.CodyGatewayRateLimit{}, graphqlbackend.CodyGatewayRateLimitSourcePlan, err
+		}
+		isProUser := user.CodyProEnabledAt != nil
+		intervalSeconds, limit, err = getSelfServeUsageLimits(scope, isProUser, *cfg)
+	}
+	if err != nil {
```
thenamankumar, line 225: This check should be inside the above block. on line 224.

cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go
```diff
@@ -198,10 +208,27 @@ func getCompletionsRateLimit(ctx context.Context, db database.DB, userID int32,
...
+		return licensing.CodyGatewayRateLimit{}, graphqlbackend.CodyGatewayRateLimitSourcePlan, err
+	}
+
+	// Otherwise, fall back to the global limit.
```
thenamankumar, line 229: Update this comment to say "fallback to pre-dec-ga global limits"

cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go
```diff
@@ -221,10 +248,36 @@ func getCompletionsRateLimit(ctx context.Context, db database.DB, userID int32,
...
+		IntervalSeconds: intervalSeconds, // Daily limit TODO(davejrt)
 	}, source, nil
 }
 
+func getSelfServeUsageLimits(scope types.CompletionsFeature, isProUser bool, cfg conftypes.CompletionsConfig) (int32, *int, error) {
+	switch scope {
+	case types.CompletionsFeatureChat:
+		if isProUser {
+			if cfg.PerUserProDailyLimit > 0 {
+				return oneDayInSeconds, pointers.Ptr(cfg.PerUserProDailyLimit), nil
```
thenamankumar, line 260: I would rename these to 
- `PerProUserChatDailyLimit`
- `PerProUserCodeCompletionsDailyLimit`
- `PerCommunityUserChatMonthlyLimit`
- `PerCommunityUserCodeCompletionsMonthlyLimit`

We may even get the interval from config, but lets leave that for now, as it will require a significant update to the overrides system.

cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go
```diff
@@ -185,8 +194,9 @@ func (r codyUserGatewayAccessResolver) EmbeddingsRateLimit(ctx context.Context)
 func getCompletionsRateLimit(ctx context.Context, db database.DB, userID int32, scope types.CompletionsFeature) (licensing.CodyGatewayRateLimit, graphqlbackend.CodyGatewayRateLimitSource, error) {
 	var limit *int
 	var err error
-	source := graphqlbackend.CodyGatewayRateLimitSourceOverride
 
+	// Apply overrides first.
+	source := graphqlbackend.CodyGatewayRateLimitSourceOverride
 	switch scope {
 	case types.CompletionsFeatureChat:
 		limit, err = db.Users().GetChatCompletionsQuota(ctx, userID)
```
thenamankumar, line 202: I would also take a look at where else `db.Users().GetChatCompletionsQuota` is being used, so that we can integrate the new logic there as well.

thenamankumar:
I remember there are some other APIs as well, where we will need to integrate this new logic. Do we have a list of TODOs in some Github issue to track all the pending items? 

One such API is for the codeCompletionsQuota on the user type and then default values are exposed by the site type in graphql API.

cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go
vdavid: Oh fck, nice catch! 😓

cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go
vdavid: An again 🤦

cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go
vdavid: @thenamankumar `GetChatCompletionsQuota` and `GetCodeCompletionsQuota` are only used in two places:
1. In User's [`CompletionsQuotaOverride`](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/cmd/frontend/graphqlbackend/user.go?L567) to get the override itself
2. In httpsapi/limiter.go's [`getConfiguredLimit`](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/internal/completions/httpapi/limiter.go?L166), which is used in dotcom's limiting feature itself

Neither seems like a good idea to add our additional logic to, so I think this is good as it is.

thenamankumar (approved):

PR #58436, opened 2023-11-20
changelog: document exclude rules and size/stars rules

mrnugget:
Changelog for mentioned PRs.
## Test plan

- N/A


eseliger (approved):
is this also somewhere in docs? I fear that this might get lost pretty quickly

mrnugget:
> is this also somewhere in docs? I fear that this might get lost pretty quickly

There's the configuration in the docs. Otherwise there's no page that explains `exclude` at all.

PR #58451, opened 2023-11-21
new: add claude-2.0 and claude-2.1 models

abeatrix:
CONTEXT: https://sourcegraph.slack.com/archives/C05AGQYD528/p1700547046350299 & https://sourcegraph.slack.com/archives/C05AGQYD528/p1700592241779049

Update to port all "anthropic/claude-2" users to use the correct pinned version of the chat model "anthropic/claude-2.0".
- Add support for `anthropic/claude-2.0` & `anthropic/claude-2.1`

## Test plan

allow list updated to use pinned version


internal/completions/httpapi/chat.go
```diff
@@ -26,6 +26,9 @@ func NewChatCompletionsStreamHandler(logger log.Logger, db database.DB) http.Han
 		"chat",
 		func(requestParams types.CodyCompletionRequestParameters, c *conftypes.CompletionsConfig) (string, error) {
 			if isAllowedCustomChatModel(requestParams.Model) {
+				if requestParams.Model == "anthropic/claude-2" {
+					return "anthropic/claude-2.0", nil
+				}
```
philipp-spiess, line 31: Not sure if this is the right approach, shouldn't we instead update the clients to pin the version? 🤔 Also this won't work if `anthropic/claude-2` is either the default model (as configured by the site setting) or if BYOK is used. I'll let @rafax or @eseliger have the final say here though 🙈

internal/completions/httpapi/chat.go
```diff
@@ -27,6 +27,10 @@ func NewChatCompletionsStreamHandler(logger log.Logger, db database.DB) http.Han
 		"chat",
 		func(_ context.Context, requestParams types.CodyCompletionRequestParameters, c *conftypes.CompletionsConfig) (string, error) {
 			if isAllowedCustomChatModel(requestParams.Model) {
+				// Use pinned version claude-2.0 instead of claude-2
+				if requestParams.Model == "anthropic/claude-2" {
```
chwarwick, line 31: Since we aren't currently pinning the version users are getting `claude-2.1` do we want to default back down to 2.0?

bobheadxi: Should we really be rewriting the request model in this manner?

abeatrix: @bobheadxi We have gotten a note from anthropic to let us know claude-2 is an unpinned version and recommend us to use `claude-2.0` instead, so I thought this would be the best way (uneducated gues) to port users on claude-2 to claude-2.0 without breaking the application for current user, but happy to remove it if that's not the preferred way.

eseliger: I think the best will be to [update our default choice in the conf/computed.go file](https://sourcegraph.com/-/editor?remote_url=github.com%2Fsourcegraph%2Fsourcegraph&branch=main&file=internal%2Fconf%2Fcomputed.go&editor=VSCode&version=2.2.16&start_row=594&start_col=2&end_row=594&end_col=2), that way customers will get the version we recommend them to use, but don't prescribe them that 2 actually is meant to mean 2.0. That could cause unexpected behavior when someone actually wants to use latest 2, instead of 2.0 IMO. 

Warning: Many people on slack seem to start telling people that we use claude 2.1 - should we even go back to 2.0 in that case? Are you guys confident 2.1 won't make it much worse?

Also, we should be adding it to the default allowed models list here:
- https://sourcegraph.com/-/editor?remote_url=github.com%2Fsourcegraph%2Fsourcegraph&branch=main&file=cmd%2Fcody-gateway%2Finternal%2Factor%2Fanonymous%2Fanonymous.go&editor=VSCode&version=2.2.16&start_row=39&start_col=9&end_row=39&end_col=9
- https://sourcegraph.com/-/editor?remote_url=github.com%2Fsourcegraph%2Fsourcegraph&branch=main&file=cmd%2Ffrontend%2Finternal%2Fdotcom%2Fproductsubscription%2Fcodygateway_dotcom_user.go&editor=VSCode&version=2.2.16&start_row=230&start_col=3&end_row=230&end_col=3
- https://sourcegraph.com/-/editor?remote_url=github.com%2Fsourcegraph%2Fsourcegraph&branch=main&file=internal%2Flicensing%2Fcodygateway.go&editor=VSCode&version=2.2.16&start_row=26&start_col=1&end_row=26&end_col=1

Abuse detection also seems to require an update here: https://sourcegraph.com/-/editor?remote_url=github.com%2Fsourcegraph%2Fsourcegraph&branch=main&file=cmd%2Fcody-gateway%2Finternal%2Fhttpapi%2Fcompletions%2Fanthropic.go&editor=VSCode&version=2.2.16&start_row=39&start_col=3&end_row=41&end_col=2 cc @rafax 

..And make sure that customers who have overrides also get the model, currently unfortunately easiest done from the db.

And the final note: We have to ship cody gateway changes first, deploy them, await a sync or expire the cache, and only afterwards start referencing the model otherwise clients will start to break until gateway starts accepting requests for 2.1. So we'll likely need to split this PR in two separate pieces :)

abeatrix: @eseliger i added 2.0 and 2.1 to the default allowed models list but didn't make any changes to the computed.go.file since that will affect customers, is that correct?

I am hoping to focus on making it work for DotCom users first, and then update the default model in computerd.go in a separate PR for customer like you suggested, does this plan sounds good to you?

eseliger: Gotcha, works for me but I still think we shouldn't be rewriting requests here, and instead use another default model or change the site config of dotcom instead - WDYT?

The point about having to ship the cody gateway allowlist changes first is still important though I believe, and if we don't change the defaults we'll have to allow the claude-2.1 model for dotcom specifically in license management as well.

bobheadxi (approved):

internal/completions/httpapi/chat.go
abeatrix: Sounds good! I've removed that and will work on a separate PR to update computed.go for enterprise users.

Thanks @bobheadxi for the review 🙇‍♀️

internal/completions/httpapi/chat.go
rafax: dotcom has no explicit override, so it uses `anthropic/claude-2` as the default (cc @thenamankumar @vdavid as owners of configuration for PLG).

internal/completions/httpapi/chat.go
rafax: @abeatrix @eseliger should we update the default model we use when user doesn't specify one explicitly (https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/internal/conf/computed.go?L635)?

abeatrix: @rafax Sorry, here is the separated PR for the change in computed.go https://github.com/sourcegraph/sourcegraph/pull/58490

PR #58454, opened 2023-11-21
search jobs: allow repo: predicates

stefanhengl:
We allow repo predicates for Search Jobs by removing the corresponding validation checks. At the same time we improve the checks against file predicates.

We now support the following **repo** predicates:
- .has.meta
- .has.file
- .has.content
- .has.path
- .has.commit.after
- .has.description

Notes:
- Why don't we allow file predicates? `file:has.content()` is translated to an AND query in our (query parsing) job layer. In Search Jobs we don't call this code path. A quick copy&paste hack let to Search Jobs breaking in unexpected ways. Hence I want to leave file predicates to a separate investigation and PR.
- The validation in the client and backend are not 100% identical which is a trade-off we accepted for the EAP. The plan is to remove the client validation soon and instead call out to the backend for validation. 

Test plan:
- updated unit tests for client and backend
- manual testing
  - I ran searches with various repo predicates and compared match counts of search jobs with the match counts we display in the Search UI. Repo predicates worked without any issue. This is to be expected because repo predicates are translated to `RepoListOptions` which fits in nicely with our design of Search Jobs.

## Preview 🤩
[Preview Link](https://docs.sourcegraph.com/@sh/search-jobs-enable-repo-predicates)

### DIFF
--- a/internal/search/job/jobutil/exhaustive_job.go
+++ b/internal/search/job/jobutil/exhaustive_job.go
@@ -23,8 +23,6 @@ type Exhaustive struct {
 //
 // It will return an error if the input query is not supported by Exhaustive.
 func NewExhaustive(inputs *search.Inputs) (Exhaustive, error) {
-	// TODO(keegan) a bunch of tests around this after branch cut pls
-
 	if inputs.Protocol != search.Exhaustive {
 		return Exhaustive{}, errors.New("only works for exhaustive search inputs")
 	}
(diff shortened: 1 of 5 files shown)


stefanhengl:
cc @MaedahBatool tagging you for the 1-line doc change.

internal/search/job/jobutil/exhaustive_job.go
```diff
@@ -46,15 +44,12 @@ func NewExhaustive(inputs *search.Inputs) (Exhaustive, error) {
 		return Exhaustive{}, errors.Errorf("expected a simple expression (no and/or/etc). Got %v", b.Pattern)
 	}
 
-	// no predicates
-	if inputs.Query.Exists(query.FieldRepoHasFile) {
-		return Exhaustive{}, errors.Errorf("repoHasFile is not supported")
```
keegancsmith, line 51: so we support this predicate now?

internal/search/job/jobutil/exhaustive_job_test.go
```diff
@@ -157,10 +157,11 @@ func TestNewExhaustive_negative(t *testing.T) {
 		// catch-all regex
 		{query: `type:file index:no r:.* .*`, isPatterntypeRegex: true},
 		{query: `type:file index:no r:repo .*`, isPatterntypeRegex: true},
-		// predicates
-		{query: `type:file index:no repohasfile:foo.bar content`},
-		{query: `type:file index:no file:has.content("content")`},
-		{query: `type:file index:no repo:has.path("src") content`},
```
keegancsmith, line 163: should you add some tests for things we do support now?

keegancsmith (approved):

keegancsmith:
backport?

internal/search/job/jobutil/exhaustive_job.go
stefanhengl: yes, we support `repo:has.file(...)`

PR #58460, opened 2023-11-21
gitserver tests: sort results before comparing

burmudar:
Had this test flake on me recently where the result was actually correct but the ordering was different. 

_Can't find the build where this happene_ 😢
## Test plan
Test locally + CI


eseliger (approved):
Doesn’t look like a gitserver test but LGTM 😬

PR #58461, opened 2023-11-21
code-search: properly display disabled bulk operations

### LINKED ISSUE #58447: Hide or disable "Detach changesets" from bulk changes when not on archived tab
- **Sourcegraph version:** https://sourcegraph.com/search

The option to detach changesets is present in the list of bulk actions, but it applies only to archived changesets, so it should be hidden or disabled elsewhere.

The documentation about [bulk operations on changesets](https://docs.sourcegraph.com/batch_changes/how-tos/bulk_operations_on_changesets#supported-types-of-bulk-operations) says that the purpose of detach is, "to remove them from the archived tab" so perhaps Detach doesn't apply unless they've been archived first. The documentation for [updating a batch change](https://docs.sourcegraph.com/batch_changes/how-tos/updating_a_batch_change#removing-changesets) also mentions Detach in the context of archived changesets.

Reference [this comment](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@87c2250b707e3e97b85f178786040576e01fe906/-/blob/client/web/src/enterprise/batches/detail/changesets/ChangesetSelectRow.tsx?L89) on the source code for the Detach operation.

[Internal discussion](https://sourcegraph.slack.com/archives/C05EA9KQUTA/p1700497616036959)

#### Steps to reproduce:

1. View a batch change.
2. Click the checkbox next to changeset(s) that are not archived
3. click the `Select action` button

#### Expected behavior:

the `Detach changesets` option is not active/available 

#### Actual behavior:

the `Detach changesets` option _is_ active, but clicking on it does nothing.
Screenshot 2023-11-20 at 13 55 17

BolajiOlajide:
Closes #58447

CleanShot 2023-11-21 at 14 36 05
CleanShot 2023-11-21 at 14 35 30

A UI regression was introduced some time ago that caused disabled buttons in the DropDownButton to not be styled differently. This PR fixes that by adding styles for disabled state for a MenuItem.

## Test plan

Manual testing

### DIFF
--- a/client/web/src/enterprise/batches/DropdownButton.module.scss
+++ b/client/web/src/enterprise/batches/DropdownButton.module.scss
@@ -12,8 +12,18 @@
 
 .menu-list-item {
     padding: 0.5rem 1rem;
+    cursor: pointer;
 
-    &[data-reach-menu-item]:active {
+    &[data-reach-menu-item][aria-disabled="false"]:active {
         --text-muted: var(--white);
+    }
+
+    &[data-reach-menu-item][aria-disabled="true"]:hover {
+        background-color: transparent;
+    }
+
+    &[data-reach-menu-item][aria-disabled="true"] {
+        cursor: not-allowed;
+        opacity: 0.7;
     }
 }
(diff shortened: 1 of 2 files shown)


morgangauth (approved):
Looks great - much clearer now, thank you!

peterguy (approved):
It'll be great to undo this regression

PR #58465, opened 2023-11-21
[bitbucket cloud] Support workspace access tokens

### LINKED ISSUE #58321: Support for auth tokens for bitbucket cloud
We currently only support username + appPassword.

A customer has asked for this.

/cc @mrnugget

pjlast:
Closes #58321 

This PR adds support for Bitbucket Cloud Premium Workspace access tokens.

A new code host config option is added: `accessToken`, and when set instead of `username` and `appPassword`, will be used to sync and clone repositories.

## Test plan

VCR test added.

## Preview 🤩
[Preview Link](https://docs.sourcegraph.com/@pjlast/58321-support-bitbucket-cloud-access-tokens)

### DIFF
--- a/doc/admin/external_service/bitbucket_cloud.md
+++ b/doc/admin/external_service/bitbucket_cloud.md
@@ -18,6 +18,49 @@ In addition, there is one more field for configuring which repositories are mirr
 - [`teams`](bitbucket_cloud.md#configuration)<br>A list of teams (workspaces) that the configured user has access to whose repositories should be synced.
 - [`exclude`](bitbucket_cloud.md#configuration)<br>A list of repositories to exclude, which takes precedence over the `teams` field.
 
+## Configuration options
+
+Bitbucket Cloud code host connections can be configured with either a username and app password combination, or with workspace access tokens.
+
+### Username and app password
+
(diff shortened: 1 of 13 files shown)


doc/admin/external_service/bitbucket_cloud.md
eseliger, line 1: <3 for docs

schema/bitbucket_cloud.schema.json
```diff
@@ -6,7 +6,7 @@
   "allowComments": true,
   "type": "object",
   "additionalProperties": false,
-  "required": ["url", "username", "appPassword"],
+  "required": ["url"],
```
eseliger, line 9: We could probably use a `oneOf` to make it require either username AND appPassword, or accessToken, but never none of the two, and never both.

If that takes less than 30 min, might be worth it for the improved UX of getting at least some validation that you're lacking some fields?

eseliger (approved):
Great work, so quick!

Can you make sure to start a backport so we can ship this next Wednesday for the customer? Thanks :)

schema/bitbucket_cloud.schema.json
pjlast: Added it! much better now I think 😬

PR #58467, opened 2023-11-21
Propagate retry-after for upstream 503

rafax:
Set `retry-after` HTTP header for upstream 503 status codes - some of our backends don't return 429 when we're exceeding our limits, and go straight to returning 503. 
We want to slow down client retries (coming from sourcegraph.com), and we can use the same mechanism as we do for 429 (that we already rewrite to 503). 
## Test plan

- tested locally
-- before this change, retry-after header wasn't set for upstream 503, resulting in multiple client (and Cody gateway retries)
-- after this change, retry-after is set for upstream 503 (same as for 429 that gets rewritten to 503)


cmd/cody-gateway/internal/httpapi/completions/upstream.go
```diff
@@ -352,6 +352,8 @@ func makeUpstreamHandler[ReqT UpstreamRequest](
 					log.Error(errors.New(resp.Status)), // real error needed for Sentry reporting
 					log.String("resp.headers", headers.String()))
 				resolvedStatusCode = http.StatusServiceUnavailable
+			}
+			if resolvedStatusCode == http.StatusServiceUnavailable {
```
unknwon, line 356: Hmm, I'm expecting to see

```go
if upstreamStatusCode == http.StatusTooManyRequests || resolvedStatusCode == http.StatusServiceUnavailable {
```

?

rafax: `upstreamStatusCode == http.StatusTooManyRequests` means that `resolvedStatusCode` is equal to `StatusServiceUnavailable` (see the if above, and the last line where we update the resolved status code)?

rafax: So, I think line 356 is equivalent to what you suggested (both if expressions are evaluated to the same value), I used the simpler version because it's easy to miss that the left side of || looks at upstream, while right one looks at resolved, but if you think what you suggested is more readable, happy to change.

unknwon: Ohhh, sorry i missed the variable name is different!

unknwon (approved):

cmd/cody-gateway/internal/httpapi/completions/upstream.go
bobheadxi: Doesn't this change remove `retry-after` propagation from 429 handling?

bobheadxi: I think we should have a separate block with the condition Joe describes, specifically to do `retry-after` propagation.

We can also make this broader by using the same condition used in `httpcli` for respecting `retry-after`:

https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/internal/httpcli/client.go?L610-614

```go
		if status == 0 ||
			(status >= 500 && status != http.StatusNotImplemented) ||
			status == http.StatusTooManyRequests {
```

rafax: > Doesn't this change remove retry-after propagation from 429 handling?
@bobheadxi on the last line of 429 handling (L354) we set  `resolvedStatusCode = http.StatusServiceUnavailable`, so I think this keeps the propagation for 429?

bobheadxi: Ahhh, I see what you mean earlier now

bobheadxi: Maybe:

```suggestion
			}

			// This handles upstream 429 responses as well, since they get
			// resolved to http.StatusServiceUnavailable.
			if resolvedStatusCode == http.StatusServiceUnavailable {
```

PR #58490, opened 2023-11-22
update: set default chat model to claude-2.0

abeatrix:
FOLLOW UP FROM https://github.com/sourcegraph/sourcegraph/pull/58451#discussion_r1401349931

This PR updates to set default chat model from claude 2 (unpinned version) to claude 2.0 (pinned).

## Test plan

Check existing tests

### DIFF
--- a/internal/conf/computed.go
+++ b/internal/conf/computed.go
@@ -592,7 +592,7 @@ func GetCompletionsConfig(siteConfig schema.SiteConfiguration) (c *conftypes.Com
 	if completionsConfig == nil {
 		completionsConfig = &schema.Completions{
 			Provider:        string(conftypes.CompletionsProviderNameSourcegraph),
-			ChatModel:       "anthropic/claude-2",
+			ChatModel:       "anthropic/claude-2.0",
 			FastChatModel:   "anthropic/claude-instant-1",
 			CompletionModel: "anthropic/claude-instant-1",
 		}
@@ -632,7 +632,7 @@ func GetCompletionsConfig(siteConfig schema.SiteConfiguration) (c *conftypes.Com
 
 		// Set a default chat model.
 		if completionsConfig.ChatModel == "" {
-			completionsConfig.ChatModel = "anthropic/claude-2"
+			completionsConfig.ChatModel = "anthropic/claude-2.0"
 		}
 
 		// Set a default fast chat model.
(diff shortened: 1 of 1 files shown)


internal/conf/computed.go
```diff
@@ -1119,11 +1119,14 @@ func anthropicDefaultMaxPromptTokens(model string) int {
...
 
 	}
-	if model == "claude-2" || model == "claude-v2" {
+	if model == "claude-2" || model == "claude-2.0" || model == "claude-v2" {
 		// TODO: Technically, v2 also uses a 100k window, but we should validate
 		// that returning 100k here is the right thing to do.
 		return 12_000
 	}
+	if model == "claude-2.1" {
+		return 200_000
```
rafax, line 1128: We kept the window small for Claude 2.0, should we similarly keep it small for 2.1? 

Also, our current limit enforced in Cody Gateway is 20_000 tokens.

rafax (approved):
One question about Claude-2.1 max prompt length, LGTM otherwise.

bobheadxi (approved):

bobheadxi:
Might want to backport if you want this available in 5.2.x

abeatrix:
> Might want to backport if you want this available in 5.2.x

Since we are focusing on dotCom users at the beginning i think we should be fine?

> And the final note: We have to ship cody gateway changes first, deploy them, await a sync or expire the cache, and only afterwards start referencing the model otherwise clients will start to break until gateway starts accepting requests for 2.1. So we'll likely need to split this PR in two separate pieces :)

Question: How do I know if this change has already been shipped or not 😅 ?

abeatrix:
@bobheadxi @eseliger @rafax I'm not sure if it's safe to merge this change consider Erik's comments, so I'll wait for your confirmation before I merge :) 

thank you for all your help on this change ❤️

rafax:
> > Might want to backport if you want this available in 5.2.x
> 
> Since we are focusing on dotCom users at the beginning i think we should be fine?
> 
> > And the final note: We have to ship cody gateway changes first, deploy them, await a sync or expire the cache, and only afterwards start referencing the model otherwise clients will start to break until gateway starts accepting requests for 2.1. So we'll likely need to split this PR in two separate pieces :)
> 
> Question: How do I know if this change has already been shipped or not 😅 ?

They will be shipped with this PR https://github.com/sourcegraph/infrastructure/pull/5634

abeatrix:
> > > Might want to backport if you want this available in 5.2.x
> > 
> > 
> > Since we are focusing on dotCom users at the beginning i think we should be fine?
> > > And the final note: We have to ship cody gateway changes first, deploy them, await a sync or expire the cache, and only afterwards start referencing the model otherwise clients will start to break until gateway starts accepting requests for 2.1. So we'll likely need to split this PR in two separate pieces :)
> > 
> > 
> > Question: How do I know if this change has already been shipped or not 😅 ?
> 
> They will be shipped with this PR https://github.com/sourcegraph/infrastructure/pull/5634

I don't have access to the repo to check its status. @kalanchan do you mind checking it for me please?

PR #58502, opened 2023-11-22
feat: add mistral-7b-instruct-4k chat model

abeatrix:
CONTEXT: https://sourcegraph.slack.com/archives/C05AGQYD528/p1700598853078489?thread_ts=1700598772.652229&cid=C05AGQYD528

This adds the "accounts/fireworks/models/mistral-7b-instruct-4k" chat model to the allowed list of custom chat models.

## Test plan

Add support for the mistral-7b-instruct-4k chat model to the allowed list for chat.

The same model is already allowed for completions.

### DIFF
--- a/internal/completions/httpapi/chat.go
+++ b/internal/completions/httpapi/chat.go
@@ -50,6 +50,7 @@ func isAllowedCustomChatModel(model string) bool {
 		"anthropic/claude-2.0",
 		"anthropic/claude-2.1",
 		"anthropic/claude-instant-1.2-cyan",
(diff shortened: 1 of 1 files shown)


umpox (approved):

philipp-spiess (changes requested):
chat is not implemented for fireworks yet so this will throw here: https://github.com/sourcegraph/sourcegraph/blob/main/internal/completions/client/fireworks/fireworks.go#L36 

we need to implement their /chat API and properly forward the payloads

abeatrix:
added by @philipp-spiess

PR #58503, opened 2023-11-22
code-search: remove redundant useState

BolajiOlajide:
We currently save the initial values for fetching the tree at a revision in state, however this seems redundant as we can memoize so that the state isn't recreated on every render.

## Test plan

Manual testing to ensure this doesn't introduce a regression.

### DIFF
--- a/client/web/src/repo/RepoRevisionSidebarFileTree.tsx
+++ b/client/web/src/repo/RepoRevisionSidebarFileTree.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useRef, useState } from 'react'
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 
 import { useApolloClient, gql as apolloGql } from '@apollo/client'
 import {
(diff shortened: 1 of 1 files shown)


BolajiOlajide:
Tagging @philipp-spiess here since this change was introduced by you.

philipp-spiess (approved):
I like to use useState if I need to 100% have the right reference everywhere because in the beginning of hooks the messaging was that `useMemo` cached objects might be released if React needs more memory but I think now both of those things are pretty much identical.

useMemo might be a bit more expensive even since in every render it needs to compare the dependencies array where as useState would just noop on every render.

BolajiOlajide:
> I like to use useState if I need to 100% have the right reference everywhere because in the beginning of hooks the messaging was that `useMemo` cached objects might be released if React needs more memory but I think now both of those things are pretty much identical.
> 
> useMemo might be a bit more expensive even since in every render it needs to compare the dependencies array where as useState would just noop on every render.

Good point. I'll close this out for now - I think we can leave it as is.

PR #56569, opened 2023-09-13
aspect workflows: add initial aspect workflow yaml

burmudar:
Initial aspect workflow config
## Test plan
[aspect workflow pipeline ](https://buildkite.com/sourcegraph/aspect-workflow-sourcegraph)picks this up

PR #58504, opened 2023-11-22
build(bazel): upgrade to lastest 1.x aspect-bazel-lib 1.38.1

gregmagolan:
Pre-factored out of https://github.com/sourcegraph/sourcegraph/pull/56569

## Test plan

CI


client/web/src/integration/BUILD.bazel
```diff
@@ -96,6 +96,7 @@ ts_project(
         "//:node_modules/mocha",
         "//:node_modules/puppeteer",
         "//:node_modules/ts-key-enum",
+        "//client/shared/src/testing",  # keep
```
gregmagolan, line 99: This is a missing dep that makes `//client/web/src/integration:integration-tests_binary` flaky since it runs out of the sandbox.

https://buildkite.com/sourcegraph/sourcegraph/builds/251945

gregmagolan: Moved this to https://github.com/sourcegraph/sourcegraph/pull/58507 instead

gregmagolan: Looks like https://github.com/sourcegraph/sourcegraph/pull/58507 needs to land first and then a rebase on top of main should green this one up as it hitting the flake.

burmudar: Landed #58507

burmudar (approved):

Issue #58360, opened 2023-11-15
Embeddings: fix slow scheduling queries


jtibshirani:
Once a minute, we query the `repo_embedding_jobs` table to check a repo needs to be re-embedded, or if there is already an up-to-date embeddings index. At some customer sites, we've seen this query add a ton of load to the DB.

Ideas for fixing:
* Run this query less frequently -- we can check this less frequently, as [default minimum interval](https://docs.sourcegraph.com/cody/core-concepts/embeddings#minimum-time-interval-between-automatically-scheduled-embeddings) between embedding jobs is 24 hours
* Add indexes to this table

PR #58510, opened 2023-11-22
Embeddings: fix low-hanging issues with scheduling job

jtibshirani:
As part of the embeddings policy framework, a worker periodically checks what
repos can be embedded. For every candidate repo, it queries the DB to see if
there's a new revision to embed. This runs every minute and becomes
increasingly expensive as the jobs table fills up with more entries over time.

This change makes small optimizations to improve this:
* Add an index to make selecting on `repo_id` and `revision` much faster
* Check the repos every 5 minutes instead of 1 minute. This shouldn't make a
huge difference in user experience, since by default embeddings jobs aren't
allowed to be scheduled within 24h of the last run

Addresses #58360

## Test plan

Covered by existing embeddings jobs tests and DB upgrade tests


jtibshirani:
I'm planning a follow-up that avoids making queries per every repo against the repo and jobs tables, which can be expensive when a large number of repos match the policy.

keegancsmith (approved):

PR #58513, opened 2023-11-22
Add Cody usage stats API

thenamankumar:
Based on event logs, confirmed the event names from:
1. Looker dashboard
2. Connecting prod VS Code extension with the local server.
3. Big Query

CleanShot 2023-11-23 at 03 57 22@2x

This might be slow on dotcom given a huge event_logs table size. Will test the speed on prod and will add a caching layer (which might not help in the first load) if required.

## Test plan

https://sourcegraph.test:3443/api/console#%7B%22query%22%3A%22%7B%5Cn%20%20currentUser%20%7B%5Cn%20%20%20%20createdAt%5Cn%20%20%20%20codyProEnabledAt%5Cn%20%20%20%20codyCurrentPeriodStartDate%5Cn%20%20%20%20codyCurrentPeriodEndDate%5Cn%20%20%20%20codyCurrentPeriodChatUsage%5Cn%20%20%20%20codyCurrentPeriodCodeUsage%5Cn%20%20%7D%5Cn%7D%5Cn%22%7D


cmd/frontend/graphqlbackend/user.go
```diff
@@ -190,6 +192,114 @@ func (r *UserResolver) CodyProEnabled(ctx context.Context) bool {
...
+	if !envvar.SourcegraphDotComMode() {
+		return nil, errors.New("this feature is only available on sourcegraph.com")
+	}
+
+	currentPeriodStartDate, err := r.CodyCurrentPeriodStartDate()
+	if err != nil {
+		return nil, err
+	}
+
+	query := sqlf.Sprintf(`WHERE user_id = %s AND timestamp >= %s AND timestamp <= NOW() AND (name LIKE '%%recipe%%' OR name LIKE '%%command%%' OR name LIKE '%%chat%%')`, r.user.ID, currentPeriodStartDate.Time)
```
vdavid, line 205: Did you check the indexes on these fields, maybe ran an EXPLAIN ANALYZE on the performance of this new query? Just want to make sure that we're not doing anything _really_ inefficient here.

cmd/frontend/graphqlbackend/user.go
```diff
@@ -190,6 +192,114 @@ func (r *UserResolver) CodyProEnabled(ctx context.Context) bool {
...
+	return &intCount, err
+}
+
+func (r *UserResolver) CodyCurrentPeriodStartDate() (*gqlutil.DateTime, error) {
+	if !envvar.SourcegraphDotComMode() {
+		return nil, errors.New("this feature is only available on sourcegraph.com")
+	}
+
+	// 🚨 SECURITY: Only the user and admins are allowed to access the user's
+	// settings, because they may contain secrets or other sensitive data.
```
vdavid, line 239: Nit to satisfy my spellchecker 😄 
```suggestion
	// settings because they may contain secrets or other sensitive data.
```

cmd/frontend/graphqlbackend/user.go
```diff
@@ -190,6 +192,114 @@ func (r *UserResolver) CodyProEnabled(ctx context.Context) bool {
...
+	if err := auth.CheckSiteAdminOrSameUserFromActor(r.actor, r.db, r.user.ID); err != nil {
+		return nil, err
+	}
+
+	subscriptionStartDate := r.user.CreatedAt
```
vdavid, line 244: This is a bit weird behavior for old users, but we need to start counting _somewhere_, and it's very low stakes affecting only Community users, so I agree that this is good enough

cmd/frontend/graphqlbackend/user.go
```diff
@@ -190,6 +192,114 @@ func (r *UserResolver) CodyProEnabled(ctx context.Context) bool {
...
+	}
+
+	daysInMonth := time.Date(endMonth.Year(), endMonth.Month()+1, 0, 0, 0, 0, 0, endMonth.Location()).Day()
+
+	if endDayOfTheMonth > daysInMonth {
+		endDayOfTheMonth = daysInMonth
+	}
+
+	return &gqlutil.DateTime{Time: time.Date(endMonth.Year(), endMonth.Month(), endDayOfTheMonth, 0, 0, 0, 0, endMonth.Location())}, nil
+}
```
vdavid, line 301: Not too happy that this is a verbatim copy of `CodyCurrentPeriodStartDate`, save for two lines.
Maybe create `func (r *UserResolver) codyCurrentPeriodStartOrEndDate(getEndDate bool) (*gqlutil.DateTime, error)` and pack all the logic in it, then just call it from the two public functions?

cmd/frontend/graphqlbackend/user.go
```diff
@@ -190,6 +192,114 @@ func (r *UserResolver) CodyProEnabled(ctx context.Context) bool {
...
+	query := sqlf.Sprintf(`WHERE user_id = %s AND timestamp >= %s AND timestamp <= NOW() AND name LIKE '%%completion:suggested%%'`, r.user.ID, currentPeriodStartDate.Time)
+
+	count, err := r.db.EventLogs().CountBySQL(ctx, query)
+
+	intCount := int32(count)
+
+	return &intCount, err
+}
+
+func (r *UserResolver) CodyCurrentPeriodStartDate() (*gqlutil.DateTime, error) {
```
vdavid, line 233: This function badly needs some tests to prove that it handles the edge cases well. Same for its "*EndDate" pair. Would you mind testing these two functions to make sure we got it right?

vdavid (commented):
Coming along nicely! I left some comments

vdavid:
I imagine the usage data won't be that helpful for clients without also knowing the limits.
Could we perhaps also send the limits so clients can actually display how much of the quota is remaining?

cmd/frontend/graphqlbackend/user.go
thenamankumar: its fine

cmd/frontend/graphqlbackend/user.go
thenamankumar: added tests

vdavid (approved):
Nice! With the tests in the mix, this looks quite sturdy! 💪

Issue #58112, opened 2023-11-06
zoekt: only one ctags process per build


keegancsmith:
I came across this while reading the zoekt sourcecode. Currently we create one ctags process per process. We call NewParser once at the creation of a builder. It then is shared amongst the shards via lockedParser. From what I have seen in local testing symbol parsing is our bottleneck which means we effectively never index concurrenctly.

This means some of our advice at reducing concurrency with the expectation that ctags is the problem is likely incorrect. I think before increasing concurrency here we likely need to instrument our ctags code such that we can monitor memory usage better. Additionally we need to make our use of ctags more robust.

Additionally I thought there may be issues in what happens if we hit a timeout. Luckily we correctly clean up and will restart the ctags parser on the next call. However we always have CTagsMustSucceed set so in practice we abort indexing at that point.

cc @sourcegraph/search-platform

keegancsmith:
We discussed this a bit in our sync yesterday. Given that disabling symbols has often unblocked issues at customers we are not sure if blanket enabling concurrency is safe here. So if we do add concurrency here we likely will do it carefully.

Meta: @jtibshirani I see you remove this from our project board. Is that intentional?

jtibshirani:
@keegancsmith sorry I missed your comment! I removed it from the board because I'm tracking it under the umbrella issue #58133 and wanted to declutter.

I opened a draft change here: https://github.com/sourcegraph/zoekt/pull/702. Although it's a bit risky changing this, it feels like the right move to me.
* CPU usage: parallelism is bounded by the available CPUs. With the change, we can have at most `2 * parallelism` ctags processes (one for universal-ctags, one for scip-ctags), but only `1 * parallelism` of them actively processing a document at one time. As of [#699](#699), we respect this limit even when `index_concurrency` is set (like for dot com).
* Memory: from [profiling on megarepo](https://github.com/sourcegraph/sourcegraph/issues/58133#issuecomment-1811257126), each process takes 20-100MB of memory. Client-side, parsing and converting tags is also not a big memory consumer.

> I think before increasing concurrency here we likely need to instrument our ctags code such that we can monitor memory usage better. Additionally we need to make our use of ctags more robust.

> We discussed this a bit in our sync yesterday. Given that disabling symbols has often unblocked issues at customers we are not sure if blanket enabling concurrency is safe here. So if we do add concurrency here we likely will do it carefully.

My best understanding is that disabling ctags helped not because ctags itself was causing problems, but rather other memory issues:
[...]

PR #58514, opened 2023-11-22
Search: add index shard concurrency to site config

jtibshirani:
In sourcegraph/zoekt#702, we updated indexserver to parse symbols in parallel
by spawning a new ctags process per shard. By default, indexing uses all
available CPUs to create shards in parallel, so now it will create many more
processes than before.

As a safeguard, we're exposing a site config setting to reduce the indexing
concurrency. It's not intended to be set by users, but will let us experiment
and make sure the defaults are solid.

As part of this change, I bumped the Zoekt dependency:

https://github.com/sourcegraph/zoekt/compare/0959170c1623...5c51794c39d0

- https://github.com/sourcegraph/zoekt/commit/d982320abe Indexing: configure CPU count through index options
- https://github.com/sourcegraph/zoekt/commit/5c51794c39 Indexing: use one ctags process per shard

Relates to #58112

## Test plan

New unit test, also existing coverage from proto round trip tests


schema/site.schema.json
```diff
@@ -16,6 +16,12 @@
       "group": "Search",
       "examples": [true]
     },
+    "search.index.shardConcurrency": {
+      "description": "The number of threads each indexserver should use to index shards. If not set, indexserver will use the number of available CPUs. This is exposed as a safeguard and should usually not require being set.",
+      "type": "integer",
+      "group": "Search",
+      "examples": [["10"]]
```
keegancsmith, line 23: shouldn't this be
```suggestion
      "examples": [10]
```

keegancsmith (approved):

schema/site.schema.json
jtibshirani: Oops, fixed.

PR #58518, opened 2023-11-23
msp/runtime: add OpenTelemetry, make submodule

bobheadxi:
This change:

1. Renames `lib/managedservicesplatform/service` to `lib/managedservicesplatform/runtime` - I've been referring to it as runtime, and we use the word "service" a lot already, so this makes it easier to refer to and is more descriptive.
2. Makes everything under `lib/managedservicesplatform` a separate submodule from `lib` - the MSP runtime contributes a lot of dependencies to `lib`, and I don't think we should keep bloating `lib` to add more functionality to MSP.
3. Adds OpenTelemetry tracing and metrics configuration similar to the setup used by Cody Gateway and Telemetry Gateway

~This also upgrades some of the OpenTelemetry packages, but there shouldn't be any major changes, other than the addition of `embedded.*`~ Gave up on that

Part of https://github.com/sourcegraph/sourcegraph/issues/56846

## Test plan

CI, `sg run msp-example`


third_party/com_github_grpc_ecosystem_grpc_gateway_v2/README.md
```diff
@@ -1,36 +0,0 @@
-## How to regenerate a patch
```
bobheadxi, line 1: FYI @sourcegraph/team-dev-infra I was having some trouble with this patch, and found that if I just removed it entirely it seems to work fine

jac (approved):

third_party/com_github_grpc_ecosystem_grpc_gateway_v2/README.md
bobheadxi: Nevermind, it's broken and I think this upgrade is making too many changes 😁

PR #58523, opened 2023-11-23
Remove ExpiredFeatures from licenses since they're now unused

pjlast:
Follow-up to #58326

Removes the now-unused `ExpiredFeatures`.

## Test plan

Unused code.


mrnugget (approved):

eseliger:
thanks!

PR #58538, opened 2023-11-23
sg/msp: add completions for services and environments

bobheadxi:
This change adds completions for services and environments on MSP commands. The original plan was to use a `-e` flag based on feedback from @jhchabran , but after playing around with it a bit I think posargs with completions is easier to implement and much more ergonomic: https://github.com/sourcegraph/sourcegraph/pull/58538#discussion_r1404598941

## Test plan

Relevant commands:

```
sg msp gen -all msp-testbed
sg msp tfc sync -all msp-testbed
```

Completions:

https://github.com/sourcegraph/sourcegraph/assets/23356519/1ddb7cb4-ec07-4fc7-a2b6-259e79028b8d


michaellzc (approved):

dev/sg/msp/repo/repo.go
```diff
@@ -53,6 +54,28 @@ func ListServices() ([]string, error) {
...
+	}
+	repoRoot, err := repositoryRoot(cwd)
+	if err != nil {
+		return nil
+	}
+	return completions.CompleteOptions(func() (options []string) {
+		services, _ := listServicesFromRoot(repoRoot)
+		return services
+	})
+}
```
michaellzc, line 77: can we support completion for `-env` flag as well? that will be nice

bobheadxi: Unfortunately, the library doesnt support it :(

michaellzc: is it because of the you can't have flag after posargs limitation from `urfave/cli`?

bobheadxi: I'm not sure, but the library doesnt seem to give us a way to detect whether we are in the middle of a flag or not

bobheadxi: But it _does_ give us the number or args when generating a completion, so my PR description is actually invalid - it's easier to do completions with multiple posargs, since we know the first is service, we can complete the service's environments for you

Also, after playing around with this a bit I think I prefer multiple posargs a bit more myself, since you must provide the env before the service (`sg msp generate -e $env $service`, which mentally feels strange

cc @jhchabran WDYT?

bobheadxi: I switched back to posargs and added completions for `sg msp generate $service $env` - IMO this is super nice 😁 let's keep it this way


https://github.com/sourcegraph/sourcegraph/assets/23356519/1ddb7cb4-ec07-4fc7-a2b6-259e79028b8d

Issue #56849, opened 2023-09-20
msp: implement resource.bigquery


bobheadxi:
This should mirror our Cody Gateway BigQuery setup, allowing us to provision a BigQuery table within the service's project. At least in Cody Gateway this is only used in development - most other services likely want to send events directly to the TelligentSourcegraph project.

PR #58541, opened 2023-11-23
dev/msp: add BigQuery support

bobheadxi:
This change adds support for provisioning a BigQuery dataset and tables for a service to write to, similar to how we set up Cody Gateway. For now, we provision only a single BigQuery table - the intent is to allow services to generate their own telemetry events, which should only have a single table.

For schema configuration, we ask that it be provided as a JSON file alongside the spec, i.e.

```
service.yaml
${mytable}.bigquerytable.json
```

Basic configuration just looks like this:

```yaml
environments:
- id: ...
  # ...
  resources:
    bigQueryDataset:
      tables:
        - "example"
```

Example diff: https://github.com/sourcegraph/managed-services/pull/116

Usage:

```go
		bq, err := contract.BigQuery.GetTableWriter(ctx, "example")
		if err != nil {
			return nil, errors.Wrap(err, "BigQuery.GetTableWriter")
		}
		if err := bq.Write(ctx, BigQueryEntry{
			Name:      "service.started",
			CreatedAt: time.Now(),
		}); err != nil {
			return nil, errors.Wrap(err, "bq.Write")
		}
```

For usage, we wrap `bigquery.Inserter` using a helper that guides users to implement `bigquery.ValueSaver`, which is what we do in Cody Gateway.

This change also removes the service-env-var prefix, which I've decided might be rather useless :) it's not currently used at all.

Closes https://github.com/sourcegraph/sourcegraph/issues/56849

## Test plan

- [x] Run the "test changes" workflow: https://github.com/sourcegraph/managed-services/actions/runs/7095313915 - no diff for existing services
- [ ] Provision a new testbed environment for now (James is using the only one we have right now)
- [ ] Verify events received
- [ ] Try to provision a dataset in telligentsourcegraph


unknwon (approved):

bobheadxi:
Going to merge for now to avoid conflicts with some other changes - https://github.com/sourcegraph/managed-services/actions/runs/7095313915 confirmed this doesn't affect existing services. There's no planned use case for this yet until Cody Gateway migration (as SSC will use some [separately-managed assets for now](https://sourcegraph.slack.com/archives/C05E2LHPQLX/p1701731218931429)), so not going to 100% invest in making sure this is perfectly functional just yet

PR #58545, opened 2023-11-24
hide "brain dot" from repo header menu by default

sqs:
Clean up and hide-by-default the SCIP debug menu (fka BrainDot). This is not useful to most users, so it is now hidden behind the `scip-debug` feature flag. You can reload the current page with `?feat=scip-debug` appended to the URL to enable this feature flag and see the SCIP debug menu.

Also remove indirection around BrainDot. This used to be only injected in the enterprise build, but now that is the only build, so we do not need "dependency injection" for it anymore.

## Test plan

CI


varungandhi-src:
@rrhyne's [design in Figma](https://www.figma.com/file/skTkU7PPWCvH2SAlP9GYtC/Search-Suite-Q4-UX-work?type=design&node-id=3-10576&mode=design&t=ZaW7vdTCEbJQyYcH-0) moves this setting behind a dropdown rather than behind a feature flag.

I think the dropdown option has more discoverability.

BolajiOlajide:
> @rrhyne's [design in Figma](https://www.figma.com/file/skTkU7PPWCvH2SAlP9GYtC/Search-Suite-Q4-UX-work?type=design&node-id=3-10576&mode=design&t=ZaW7vdTCEbJQyYcH-0) moves this setting behind a dropdown rather than behind a feature flag.
> 
> I think the dropdown option has more discoverability.

Yes, I'm reworking the actions discoverability in [#58122](https://github.com/sourcegraph/sourcegraph/pull/58122)

sqs:
@BolajiOlajide @varungandhi-src awesome, thanks for pointing that out. I'll close this PR as the other one looks great and much better than this.

PR #58547, opened 2023-11-24
SSC: Use the user's access token for Gateway on dotcom

vdavid:
The logic I introduced is:
- IF the `cody-pro` feature flag is enabled
  - AND we're on dotcom
  - AND the completions provider is Cody Gateway
- THEN pass the user's access token to Gateway instead of the default access token in the site config.
- ELSE just do what we did before this PR

## Test plan

I logged the access token on the Cody Gateway side; it was the correct one with the feature flag on and off, and Gateway gave me correct responses.

### DIFF
--- a/internal/completions/httpapi/handler.go
+++ b/internal/completions/httpapi/handler.go
@@ -77,7 +77,7 @@ func newCompletionsHandler(
 			Build()
 		defer done()
 
-		// Use the user's access token for Cody Gateway on dotcom if PLG is enabled
+		// Use the user's access token for Cody Gateway on dotcom if PLG is enabled.
 		accessToken := completionsConfig.AccessToken
 		if envvar.SourcegraphDotComMode() &&
 			featureflag.FromContext(ctx).GetBoolOr("cody-pro", false) &&
--- a/internal/completions/httpapi/handler.go
+++ b/internal/completions/httpapi/handler.go
@@ -79,9 +79,10 @@ func newCompletionsHandler(
 
 		// Use the user's access token for Cody Gateway on dotcom if PLG is enabled.
 		accessToken := completionsConfig.AccessToken
-		if envvar.SourcegraphDotComMode() &&
-			featureflag.FromContext(ctx).GetBoolOr("cody-pro", false) &&
-			completionsConfig.Provider == conftypes.CompletionsProviderNameSourcegraph {
+		isCodyProEnabled := featureflag.FromContext(ctx).GetBoolOr("cody-pro", false)
+		isDotcom := envvar.SourcegraphDotComMode()
+		isProviderCodyGateway := completionsConfig.Provider == conftypes.CompletionsProviderNameSourcegraph
+		if isCodyProEnabled && isDotcom && isProviderCodyGateway {
 			apiToken, _, err := authz.ParseAuthorizationHeader(r.Header.Get("Authorization"))
 			if err != nil {
 				http.Error(w, err.Error(), http.StatusUnauthorized)
--- a/internal/completions/httpapi/handler.go
+++ b/internal/completions/httpapi/handler.go
@@ -85,10 +85,12 @@ func newCompletionsHandler(
 			apiToken, _, err := authz.ParseAuthorizationHeader(r.Header.Get("Authorization"))
 			if err != nil {
 				http.Error(w, err.Error(), http.StatusUnauthorized)
+				return
 			}
 			accessToken, err = accesstoken.GenerateDotcomUserGatewayAccessToken(apiToken)
 			if err != nil {
 				http.Error(w, err.Error(), http.StatusUnauthorized)
+				return
 			}
 		}
 
--- a/internal/completions/httpapi/handler.go
+++ b/internal/completions/httpapi/handler.go
@@ -85,12 +85,14 @@ func newCompletionsHandler(
 		if isCodyProEnabled && isDotcom && isProviderCodyGateway {
 			apiToken, _, err := authz.ParseAuthorizationHeader(r.Header.Get("Authorization"))
 			if err != nil {
-				http.Error(w, err.Error(), http.StatusUnauthorized)
+				trace.Logger(ctx, logger).Info("Error parsing auth header", log.String("Authorization header", r.Header.Get("Authorization")), log.Error(err))
+				http.Error(w, "Error parsing auth header", http.StatusUnauthorized)
 				return
 			}
 			accessToken, err = accesstoken.GenerateDotcomUserGatewayAccessToken(apiToken)
 			if err != nil {
-				http.Error(w, err.Error(), http.StatusUnauthorized)
+				trace.Logger(ctx, logger).Info("Access token generation failed", log.String("API token", apiToken), log.Error(err))
+				http.Error(w, "Access token generation failed", http.StatusUnauthorized)
 				return
 			}
 		}
--- a/internal/completions/httpapi/handler.go
+++ b/internal/completions/httpapi/handler.go
@@ -4,6 +4,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/sourcegraph/sourcegraph/cmd/frontend/envvar"
+	"github.com/sourcegraph/sourcegraph/internal/accesstoken"
+	"github.com/sourcegraph/sourcegraph/internal/authz"
+	"github.com/sourcegraph/sourcegraph/internal/featureflag"
 	"net/http"
 	"strconv"
 	"time"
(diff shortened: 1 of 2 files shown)


internal/completions/httpapi/handler.go
```diff
@@ -73,12 +77,27 @@ func newCompletionsHandler(
...
 
+		// Use the user's access token for Cody Gateway on dotcom if PLG is enabled
+		accessToken := completionsConfig.AccessToken
+		if envvar.SourcegraphDotComMode() &&
+			featureflag.FromContext(ctx).GetBoolOr("cody-pro", false) &&
+			completionsConfig.Provider == conftypes.CompletionsProviderNameSourcegraph {
+			apiToken, _, err := authz.ParseAuthorizationHeader(r.Header.Get("Authorization"))
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusUnauthorized)
+			}
```
bobheadxi, line 88: Do users in the web app access this endpoint? Do they also use auth headers?

internal/completions/httpapi/handler.go
```diff
@@ -73,12 +77,27 @@ func newCompletionsHandler(
...
 		defer done()
 
+		// Use the user's access token for Cody Gateway on dotcom if PLG is enabled
+		accessToken := completionsConfig.AccessToken
+		if envvar.SourcegraphDotComMode() &&
+			featureflag.FromContext(ctx).GetBoolOr("cody-pro", false) &&
+			completionsConfig.Provider == conftypes.CompletionsProviderNameSourcegraph {
+			apiToken, _, err := authz.ParseAuthorizationHeader(r.Header.Get("Authorization"))
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusUnauthorized)
```
bobheadxi, line 87: Careful, if you `http.Error`, you must also return, otherwise execution will continue and you'll make invalid writes on `w`... I've been gotcha'd by this many times 😆 

```suggestion
				http.Error(w, err.Error(), http.StatusUnauthorized)
				return
```

vdavid: I was actually wondering today whether those writes would be blocked and then this would be safe and standard practice. Thanks, Robert, makes total sense!

internal/completions/httpapi/handler.go
vdavid: Do you mean Cody Web?
Cody Web doesn't use access tokens. I have a task to make it use them. To track this gap properly, I've just created and self-assigned an issue for this: https://github.com/sourcegraph/sourcegraph/issues/58572
Did you mean this or something else, @bobheadxi?

internal/completions/httpapi/handler.go
vdavid: Fixed in 187278c1aec0afbfc77b381cc9f074a4f4443773

internal/completions/httpapi/handler.go
bobheadxi: Something like that, yeah, but instead of making the web app use access tokens, if this is a web app access with no access token we should still have the actor context here right? We could make this query the DB for a token associated with the user

vdavid: Yes, that's what we'll do, but we can't just use a random token for two reasons:
1. We display the last usage of each access token on the admin UI. It'd be totally awkward as a user to see that, say, my 2-years-old-and-long-time-unused token was just used recently 😬 
2. The user might not have any generated tokens at all.

But we do need to pass a valid access token to Gateway, to let Gateway correctly identify the user making the request and make the call back to Dotcom to get the user's limits.

So for this reason we want to create an internal token in the background, and use that.

bobheadxi: > So for this reason we want to create an internal token in the background, and use that.

+1!

internal/completions/httpapi/handler.go
```diff
@@ -73,12 +77,29 @@ func newCompletionsHandler(
 			Build()
 		defer done()
 
+		// Use the user's access token for Cody Gateway on dotcom if PLG is enabled
```
chrsmith, line 80: Super-ulta pedantic nitpick. Please let me know if you never want to see PR suggestions like this in the future...

From Google's C++ style guide:

> Pay attention to punctuation, spelling, and grammar; it is easier to read well-written comments than badly written ones.
>
> Comments should be as readable as narrative text, with proper capitalization and punctuation. In many cases, complete sentences are more readable than sentence fragments. Shorter comments, such as comments at the end of a line of code, can sometimes be less formal, but you should be consistent with your style.
>
> Although it can be frustrating to have a code reviewer point out that you are using a comma when you should be using a semicolon, it is very important that source code maintain a high level of clarity and readability. Proper punctuation, spelling, and grammar help with that goal.
> https://google.github.io/styleguide/cppguide.html#Punctuation,_Spelling_and_Grammar

```suggestion
		// Use the user's access token for Cody Gateway on dotcom if PLG is enabled.
```

internal/completions/httpapi/handler.go
```diff
@@ -73,12 +77,29 @@ func newCompletionsHandler(
...
+		accessToken := completionsConfig.AccessToken
+		if envvar.SourcegraphDotComMode() &&
```
chrsmith, line 82: Having an `if`-statement with 3x different expressions makes this a bit more difficult to read. Similarly, the function is already kinda large so perhaps we would benefit from splitting this up into smaller functions? What do you think of something like:

```go
// Use the user's access token for Cody Gateway on dotcom if PLG is enabled.
if codyProEnabled(cox, envVar, completionsConfig.Provider) {
    ...
}
```

internal/completions/httpapi/handler.go
```diff
@@ -73,12 +77,29 @@ func newCompletionsHandler(
...
+			featureflag.FromContext(ctx).GetBoolOr("cody-pro", false) &&
+			completionsConfig.Provider == conftypes.CompletionsProviderNameSourcegraph {
+			apiToken, _, err := authz.ParseAuthorizationHeader(r.Header.Get("Authorization"))
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusUnauthorized)
+				return
+			}
+			accessToken, err = accesstoken.GenerateDotcomUserGatewayAccessToken(apiToken)
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusUnauthorized)
```
chrsmith, line 92: Writing the `error` directly to the HTTP response seems a little dangerous from a security perspective. e.g. we would be sharing information we would likely want to keep private. (Think stack traces and the like.)

I think we would want to log the `err.Error()` instead, and return a static string.

Also, since we are failing in _generating_ a new access token, wouldn't we want to return a 500 ISE instead?

```go
if err != nil {
    logger.Warn("error generating Dotcom Gateway access token", log.Error(err))
    http.Error(w, "Internal Server Error", http.StatusInternalServerError)
}
```

chrsmith (approved):
I'll admit I don't quite grok the details of how the Sourcegraph backend and what's going on here with the access tokens. But I'll trust you know what you are doing :)

Just a few minor nitpicks type things to consider.

internal/completions/httpapi/handler.go
vdavid: I agree with the sentiment, but we don't have this in our coding style guide (which we also don't quite have, btw. We have [this](https://handbook.sourcegraph.com/company-info-and-process/communication/content_guidelines/style_and_mechanics/) and Prettier + ES Lint rules, but no real coding style guide), and as a result, we have a bunch of comments which are sentence fragments. I prefer complete sentences that have a beginning and end, proper capitalization, and punctuation, but I also want to be idiomatic with the rest of the codebase. This often makes me wonder about stuff like periods at the end of comments.

chrsmith: In short, hopefully it isn't too much trouble to just click "Add suggestion to batch" and not worry about it :)

But as far as this not being a _requirement_ at Sourcegraph, thanks for calling that out. I'll try to be less vocal about this particular class of code review feedback. It's something I feel stronger than I ought to about, but at the same time am 100% clear in saying it generates "light but no heat". But in the spirit of making code as clear and high-quality as possible, IMHO, it's worth calling out. (But if you disagree or finding this type of PR feedback annoying, I can certainly not mention it, since I know it can be irksome.)

internal/completions/httpapi/handler.go
thenamankumar: +1

thenamankumar:
```golang
if codyGatewayReqAndDotComMode(ctx) {
  ...
}
```

thenamankumar (approved):

internal/completions/httpapi/handler.go
vdavid: I've tweaked it avoid outputting errors directly and log traces instead, but I kept the specific error message instead of outputting a generic "internal server error" to be idiomatic with the rest of the code.

Also, I kept the `StatusUnauthorized` status code because this branch of the code is run if an invalid code is given

internal/completions/httpapi/handler.go
vdavid: I've tried to improve the condition in https://github.com/sourcegraph/sourcegraph/pull/58547/commits/f61c81dc8df4cc77b3b80b22f09860304aa36edd – it's a bit different from what you proposed because I wasn't satisfied with any short name of the bundled condition, plus it was only used once, so I think with the fix, it's slightly better documented without complicating the code.

PR #58549, opened 2023-11-24
search jobs: unify query validation

stefanhengl:
We remove the query validation from the client, eliminating an inconsistency 
we had accepted for the EAP. Instead we call a new GraphQL endpoint that 
calls the same validation we use for job creation.

The validation in the backend is just as powerful, but we don't return a slice of errors or error types. 
This means we don't show more than 1 error at the time, which I think this is an acceptable trade-off
to keep the API small.

As a consequence we don't send the error types as telemetry anymore. However, 
we still keep track whether an error was shown or not.

Test plan:
- CI
- manual testing
  - I checked that the "create search job" button stays disabled while the validation request is in-flight
  - I checked that the error messages are well formatted.
 
Kapture 2023-11-24 at 15 49 17


cmd/frontend/internal/search/resolvers/resolver.go
```diff
@@ -27,6 +28,15 @@ type Resolver struct {
 	svc    *service.Service
 }
 
+func (r *Resolver) ValidateSearchJob(ctx context.Context, args *graphqlbackend.CreateSearchJobArgs) (*graphqlbackend.EmptyResponse, error) {
+	actor := actor.FromContext(ctx)
+	if !actor.IsAuthenticated() {
+		return nil, errors.New("search jobs can only be validated by an authenticated user")
+	}
```
keegancsmith, line 35: minor: in other svc functions we do this sort of business logic inside of the service.

cmd/frontend/graphqlbackend/search_jobs.graphql
```diff
@@ -31,6 +31,16 @@ extend type Mutation {
...
 extend type Query {
+    """
+    EXPERIMENTAL: Validate a search job.
+    """
+    validateSearchJob(
+        """
+        The query to validate.
+        """
+        query: String!
+    ): EmptyResponse
```
keegancsmith, line 42: no idea what I am talking about: do we treat graphql errors as first class citizens in other parts of our app? I would imagine you leave it for exceptional circumstances and rather model the response?

keegancsmith (approved):

cmd/frontend/internal/search/resolvers/resolver.go
stefanhengl: done

cmd/frontend/graphqlbackend/search_jobs.graphql
stefanhengl: Sometimes we model the error explicitly, but I only found instances where the error is a field of a larger response type. Generally, modelling an "error as data" makes sense if we want to return more context or possibly a list of validation errors. After implementing a version of this I found it adds quite a bit of complexity to the client code and I don't think this is worth it right now. I suggest to revisit this once we get feedback about the API from customers.

PR #58552, opened 2023-11-24
PLG: Add config items for interaction-based limits

vdavid:
- Renamed the existing four config items to make it more explicit that they pertain to LLM requests
- Added four new config items that limit interactions.

The new set of constants is not used yet.

## Test plan

Not much to test on this: the old ones are covered by CI, and the new constants are not used yet.


thenamankumar (approved):
Looks Good! 

To confirm we are not using `...InteractionLimit` in any rate limiting business logic yet. It will be used either once we add `interaction_id` logic or to implement client side gating.

PR #58553, opened 2023-11-24
dotcom: don't lazy add repos with `size >= 1GB && stars < 100`

mrnugget:
This adds a new restriction to the on-demand cloning mechanism that's used on sourcegraph.com: when a repository has `>= 1GB` in size and `stars < 100` and it's hosted on GitHub.com, then we don't add it and show an error message.

When a repository runs into this restriction, we show the following error message:

screenshot_2023-12-05_15 23 03@2x

## Test plan

- Ensure your `external-services-config.json` in `dev-private` contains an external service pointing to GitHub with `cloudDefault: true`:

```json
{
  "GITHUB": [
    {
      "url": "https://github.com",
      "token": "<TOKEN>",
      "cloudDefault": true,
      "repos": [ ]
    }
  ]
}
```
- Start sourcegraph in dotcom-mode: `sg start dotcom`
- Visit https://sourcegraph.test:3443/github.com/depop/documentation -- should *NOT* be cloned, too big and too few stars
- Visit https://sourcegraph.test:3443/github.com/torvalds/linux -- this should clone, since it has enough stars


client/web/src/repo/RepoContainerError.tsx
```diff
@@ -35,6 +38,20 @@ export const RepoContainerError: React.FunctionComponent<React.PropsWithChildren
...
+                icon={AlertIcon}
+                title={displayRepoName(repoName)}
+                body={
+                    <Text className="mt-4">
+                        Repository cannot be added on-demand: {(repoFetchError as RepoDeniedError).reason}.
+                    </Text>
+                }
+            />
+        )
+    }
```
eseliger, line 53: If you make `isRepoDeniedErrorLike` return `isRepoDeniedErrorLike(t: unknown) => t is RepoDeniedError` I think you might be able to get around the casting in l48.., but not a blocker

internal/repos/syncer.go
```diff
@@ -280,7 +282,8 @@ func (s *Syncer) SyncRepo(ctx context.Context, name api.RepoName, background boo
 		return s.syncRepo(ctx, codehost, name, repo)
 	})
 	if err != nil {
-		return nil, errors.Wrapf(err, "foreground.SyncRepo (shared=%v)", shared)
+		logger.Debug("foreground sync failed", log.Error(err), log.Bool("shared", shared))
```
eseliger, line 285: How about warn?

internal/repos/syncer.go
```diff
@@ -936,3 +943,27 @@ func syncErrorReason(err error) string {
...
+type ErrRepoDeniedTooBig struct{}
+
+func (e ErrRepoDeniedTooBig) Error() string {
+	// NOTE: This message is visible to users!
+	return "repository is larger than 1GB and has less than 100 stars"
+}
+
+func (e ErrRepoDeniedTooBig) IsRepoDenied() bool { return true }
+
+func repoCanBeAddedOnDemandDotcom(repo *types.Repo) error {
```
eseliger, line 961: small docstring might be nice

eseliger (approved):

PR #58556, opened 2023-11-24
Enable logging Fireworks requests for Self-Serve users

rafax:
Enable logging Fireworks auto-complete requests for Self-Serve users. 

This is required to create a realistic "synthetic" load testing request dataset, and has been [approved by Legal/Security ](https://sourcegraph.slack.com/archives/C01GJFE26BX/p1700498843047059) (it is compliant our Terms of Service for .com users). 

This feature is disabled by default, and will only be enabled for a limited amount of time (through a Cody Gateway infra PR).

## Test plan

- tested locally


eseliger (approved):

philipp-spiess (approved):

cmd/cody-gateway/internal/httpapi/completions/fireworks.go
```diff
@@ -55,7 +56,34 @@ func NewFireworksHandler(
 					body.N = 1
 				}
 			},
-			getRequestMetadata: func(body fireworksRequest) (model string, additionalMetadata map[string]any) {
+			getRequestMetadata: func(ctx context.Context, logger log.Logger, act *actor.Actor, body fireworksRequest) (model string, additionalMetadata map[string]any) {
+				// we checked this is a code completion request in validateRequest
+				// check that actor is a PLG user
+				if logSelfServeCodeCompletionRequests && act.IsDotComActor() {
+					// LogEvent is a channel send (not an external request), so should be ok here
```
bobheadxi, line 63: does `act.IsDotComActor` === actor is a PLG user?

cmd/cody-gateway/internal/httpapi/completions/fireworks.go
```diff
@@ -55,7 +56,34 @@ func NewFireworksHandler(
 					body.N = 1
 				}
 			},
-			getRequestMetadata: func(body fireworksRequest) (model string, additionalMetadata map[string]any) {
+			getRequestMetadata: func(ctx context.Context, logger log.Logger, act *actor.Actor, body fireworksRequest) (model string, additionalMetadata map[string]any) {
```
bobheadxi, line 59: Instead of expanding this signature and generating another event log, can we just return all the new data as part of `additionalMetadata`? It gets logged in https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@237d1f237ef79938e432387cdc75b437d1fd9f5e/-/blob/cmd/cody-gateway/internal/httpapi/completions/upstream.go?L286

bobheadxi (changes requested):
Let's keep `getRequestMetadata` without side effects, and avoid logging an additional event, and return the desired new metadata as part of the `additionalMetadata` return parameter of `getRequestMetadata`, which gets logged at the end of the request (whether it succeeded or not)

cmd/cody-gateway/internal/httpapi/completions/fireworks.go
rafax: For now - I think that's true, and even if there's a special case, we can log prompts for sourcegraph.com users.

bobheadxi: let's update the docstring to reflect that this covers all doctom users then, seems we have pretty liberal terms for dotcom users so it's a bit of a free-for-all anyway

cmd/cody-gateway/internal/httpapi/completions/fireworks.go
rafax: I thought about that, but my primary considerations were:

- this code is temporary, I expect it to be gone in 14 days (ideally sooner) - keeping it separate makes it easier to remove
- the data we want to capture will be removed from event_logs table, and it's "cleaner" to filter and delete by event name than by a subproperty in metadata field

With those 2 things in mind, would you still use `additionalMetadata`?

rafax:
> Let's keep `getRequestMetadata` without side effects, and avoid logging an additional event, and return the desired new metadata as part of the `additionalMetadata` return parameter of `getRequestMetadata`, which gets logged at the end of the request (whether it succeeded or not). You can add a flag field in `additionalMetadata` that can be used to query and generate your dataset

See my comment above - assuming we plan to remove this code in 14 days, and also remove the data captured from `event_logs` table (and lather remove it altogether), would you agree that using a separate logging event is the way to go?

cmd/cody-gateway/internal/httpapi/completions/fireworks.go
bobheadxi: > the data we want to capture will be removed from event_logs table, and it's "cleaner" to filter and delete by event name than by a subproperty in metadata field

Ah, I missed this requirement - fair enough, a separate event would be better for this use case. Thanks for clarifying!

bobheadxi: (I thought we would only be generating this data for N time, not that the data itself would need to be removed as well)

bobheadxi (approved):

bobheadxi:
LGTM to serve the use case of removing this data later 👍

rafax:
> > Let's keep `getRequestMetadata` without side effects, and avoid logging an additional event, and return the desired new metadata as part of the `additionalMetadata` return parameter of `getRequestMetadata`, which gets logged at the end of the request (whether it succeeded or not). You can add a flag field in `additionalMetadata` that can be used to query and generate your dataset
> 
> See my comment above - assuming we plan to remove this code in 14 days, and also remove the data captured from `event_logs` table (and lather remove it altogether), would you agree that using a separate logging event is the way to go?

This took more than 14 days, but it's now getting removed https://github.com/sourcegraph/sourcegraph/pull/60372

PR #58558, opened 2023-11-24
sec: Allow localdev to connect to localhost

evict:
For local development we require localhost, for instance for Cody Gateway. We now add a check to make sure we don't block localhost access from when `INSECURE_DEV` is set to `true`. 

## Test plan
Tested locally.


rafax (approved):

keegancsmith (commented):
using the TestExternalTransportOpt seems like in the future we will accidently break stuff and not notice until we are in prod. Can't we instead just use a different default deny list?

PR #58559, opened 2023-11-24
svelte: Add codeintel support to blob view

fkling:
Based on #58099 and #58258 this PR adds codeintel tooltips to the SvelteKit prototype.


camdencheek (approved):
Scanned through the code, but my review primarily consisted of clicking through the UI. And that looks great!

PR #58563, opened 2023-11-24
Improve GQL ratelimits

evict:
This will allow us to tweak our GraphQL rate limiting in more detail for overly large queries. We've added more restrictions for duplicate and unique field count. In addition to these restriction we now calculate the query cost before validating the query against the schema. The validation is quit heavy and could still lead to resource exhaustion.

See https://github.com/sourcegraph/security-issues/issues/375 for more context.

Changes: 
- Calculates QueryCost before validating the schema
- Extended QueryCost struct with new metrics
- Updated fragment cost tracking
- Added tracking of duplicate/unique fields
- Updated tests

## Test plan
CI tests, local tests.

### DIFF
--- a/cmd/frontend/graphqlbackend/rate_limit.go
+++ b/cmd/frontend/graphqlbackend/rate_limit.go
@@ -130,7 +130,10 @@ func EstimateQueryCost(query string, variables map[string]any) (totalCost *Query
 		}
 		totalCost.FieldCount += cost.FieldCount
 		totalCost.AliasCount += cost.AliasCount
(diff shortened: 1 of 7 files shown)


eseliger (approved):
This seems like a good workaround for the problem, as we discovered today, there's some things that we should probably optimize in the library so it doesn't do 18M checks in validation -- but let's keep that for a follow-up. 

Ultimately, I'd love if we get that fixed if we could simplify the number of options again, to make it easier to understand for admins but also easier to explain to users who ran into limits.

PR #58564, opened 2023-11-24
Openid state improvements

evict:
We longer store the state in the cookie, but we now use redis.

## Test plan
CI tests.


unknwon (approved):

unknwon:

PR #58567, opened 2023-11-24
requestclient: add SRC_USE_CLOUDFLARE_HEADERS to use CF headers

bobheadxi:
Our Sourcegraph Cloud instances are also behind Cloudflare WAF, which makes `Cf-` headers available to them as well, which could be useful for the 2 data points in `requestclient` that prefer `Cf-` headers in Sourcegraph.com: `ForwardedFor` and client geolocation (https://github.com/sourcegraph/sourcegraph/pull/58386).

This change adds `SRC_USE_CLOUDFLARE_HEADERS` to opt in to this mechanism outside of dotcom.

[Slack discussion](https://sourcegraph.slack.com/archives/C03LCPCT3SP/p1700848383697759)

## Test plan

CI


michaellzc (approved):

keegancsmith (approved):
This may be a security issue for self hosted instances? I don't know enough about how we use this data, but it may be the case that due to the reverse proxies in front of sourcegraph x-forwarded-for is not in fact spoofed. But it may be, so in that case this is all good. I would just double check with security, but otherwise this LGTM.

bobheadxi:
Good point - talked with @michaellzc about this, and I'm going to circle back to this and make it an opt-in mechanism still, and configure it explicitly in Cloud

bobheadxi:
Since branch cut is close, not going to get this landed before then - will come back to CI failures after branch cut

michaellzc (approved):

PR #58571, opened 2023-11-24
msp/runtime: add Sentry enablement

bobheadxi:
After this, we should be able to migrate existing MSP services like Pings and Telemetry Gateway to use the shared MSP runtime library.

## Test plan

CI


jac (approved):
LGTM pending fixing the CI issue for the unused `serviceName` param in `newContract`

PR #58583, opened 2023-11-27
gitserver: Convert MergeBase client call to server-side

eseliger:
This PR is a PoC for what it could look like to convert the client-side implementation of common git functionality into well-scoped server-side offered APIs, starting with a not very widely used one, MergeBase.

I did not implement an HTTP equivalent because only next-minor clients will use it, which means that gRPC will be on by default by then. The MergeBase endpoint was so rarely changed that I don't expect that to cost any velocity in backports.

## Test plan

CI passes, including integration tests. Ran a bunch of workflows locally and checked logs that no background processes error.


internal/gitserver/v1/gitserver.proto
```diff
@@ -208,11 +216,12 @@ message CreateCommitFromPatchBinaryResponse {
   // changelistid is the Perforce changelist id
   string changelist_id = 3;
 }
+
 message ExecRequest {
   string repo = 1;
   bytes ensure_revision = 2;
   repeated bytes args = 3;
-  bytes stdin = 4;
+  bytes stdin = 4 [deprecated = true];
```
eseliger, line 224: Already unused, trying to prevent a new usage before we remove it.

internal/gitserver/v1/gitserver.proto
```diff
@@ -68,6 +68,14 @@ service GitserverService {
   rpc PerforceGetChangelist(PerforceGetChangelistRequest) returns (PerforceGetChangelistResponse) {
     option idempotency_level = NO_SIDE_EFFECTS;
   }
+  // MergeBase returns the merge base commit sha for the specified revspecs.
+  // If no common merge base exists, an empty string is returned.
+  //
+  // If the given repo is not cloned, it will be enqueued for cloning and a NotFound
+  // error will be returned, with a NotFoundPayload in the details.
+  rpc MergeBase(MergeBaseRequest) returns (MergeBaseResponse) {
+    option idempotency_level = NO_SIDE_EFFECTS;
```
eseliger, line 77: Coming back to this, we do have a side-effect which is prioritizing a clone when not found. Not sure if we want to be that nitpicky?

cmd/gitserver/internal/vcssyncer/git.go
```diff
@@ -48,7 +49,7 @@ func (s *gitRepoSyncer) IsCloneable(ctx context.Context, repoName api.RepoName,
 	}
 
 	args := []string{"ls-remote", remoteURL.String(), "HEAD"}
-	ctx, cancel := context.WithTimeout(ctx, executil.ShortGitCommandTimeout(args))
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
eseliger, line 52: This is effectively what ShortGitCommandTimeout returned. Just more expressive. Want to get away from implicit ever-growing select-based behavior.

cmd/gitserver/internal/git/cli/mergebase.go
```diff
@@ -0,0 +1,40 @@
...
+	"io"
+	"os/exec"
+	"syscall"
+
+	"github.com/sourcegraph/sourcegraph/internal/api"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+)
+
+func (g *gitCLIBackend) MergeBase(ctx context.Context, baseRevspec, headRevspec string) (api.CommitID, error) {
+	cmd, cancel, err := g.gitCommand(ctx, "merge-base", "--", baseRevspec, headRevspec)
```
eseliger, line 15: gitCommand validates the arguments, thus I don't think we need addtl validation in this method?

cmd/gitserver/internal/git/cli/config.go
eseliger, line 1: moved into cli package to fulfill interface

cmd/gitserver/internal/git/cli/config_test.go
```diff
@@ -0,0 +1,88 @@
...
+			require.NoError(t, err)
+			// Unsetting twice should not error.
+			err = git.Config().Unset(ctx, testKey)
+			require.NoError(t, err)
+			val, err = git.Config().Get(ctx, testKey)
+			require.NoError(t, err)
+			require.Equal(t, "", val)
+		}
+
+		// Check that concurrent writes aren't a problem:
```
eseliger, line 63: That's something we see regularly in sentry: some config operations failing because concurrent access is bad.
Wonder if we should guard access with our own lockfile to it, or add retries or just use go-git instead of the CLI? 
Anyhow, a problem for future us 😬

cmd/frontend/graphqlbackend/repository_comparison.go
```diff
@@ -94,13 +94,16 @@ func NewRepositoryComparison(ctx context.Context, db database.DB, client gitserv
 
 	// Find the common merge-base for the diff. That's the revision the diff applies to,
 	// not the baseRevspec.
-	mergeBaseCommit, err := client.MergeBase(ctx, r.RepoName(), api.CommitID(baseRevspec), api.CommitID(headRevspec))
+	mergeBaseCommit, err := client.MergeBase(ctx, r.RepoName(), baseRevspec, headRevspec)
```
eseliger, line 97: I noticed that we lied about that it has to be a commit SHA in the API and that we actually also pass down other revspec strings, so fixed this interface.

cmd/frontend/graphqlbackend/repository_comparison.go
```diff
@@ -94,13 +94,16 @@ func NewRepositoryComparison(ctx context.Context, db database.DB, client gitserv
...
+		return nil, err
+	}
 
 	// If possible, use the merge-base as the base commit, as the diff will only be guaranteed to be
 	// applicable to the file from that revision.
-	commitString := strings.TrimSpace(string(mergeBaseCommit))
+	commitString := string(mergeBaseCommit)
 	rangeType := "..."
-	if err != nil {
+	if commitString == "" {
```
eseliger, line 106: separated the error case and the not found case by convention of "" meaning no result.

cmd/gitserver/internal/cleanup.go
eseliger, line 1: all the changes here are just to accommodate the new Backend model and use the newly exposed config API.

cmd/gitserver/internal/executil/executil.go
```diff
@@ -26,31 +25,6 @@ import (
...
-		return conf.GitLongCommandTimeout()
-
-	case "ls-remote":
-		return 30 * time.Second
-
-	default:
-		return time.Minute
-	}
-}
-
```
eseliger, line 53: I split this out - since we don't want to exec anymore we can make this determination at callsite, vs magically based on the string. 

the ls-remote case we had now hard-coded uses 30s, archive uses gitLongCommandTimeout in the handler, and the cli backend for other requests sets the timeout to 1 minute if no deadline is set.

cmd/gitserver/internal/git/cli/config.go
eseliger, line 1: These methods are just moved here and adapted to use g.gitCommand and g.runGitCommand with the new Backend pattern.

cmd/gitserver/internal/git/cli/config_test.go
eseliger, line 1: new tests!

cmd/gitserver/internal/git/cli/exec.go
```diff
@@ -1,17 +1,30 @@
...
 
-	"github.com/grafana/regexp"
-	"k8s.io/utils/strings/slices"
-
 	"github.com/sourcegraph/log"
+
+	"github.com/sourcegraph/sourcegraph/cmd/gitserver/internal/common"
 )
 
+func (g *gitCLIBackend) Exec(ctx context.Context, args ...string) (io.ReadCloser, error) {
```
eseliger, line 18: expected to be removed once we're done with the migration work

cmd/gitserver/internal/git/cli/mergebase.go
```diff
@@ -0,0 +1,40 @@
...
+		// Exit code 1 and empty output most likely means that no common merge-base was found.
+		var e *exec.ExitError
+		if errors.As(err, &e) && e.Sys().(syscall.WaitStatus).ExitStatus() == 1 {
+			if len(e.Stderr) == 0 {
+				return "", nil
+			}
+		}
+
+		return "", err
+	}
```
eseliger, line 37: this API works similarly to exec.Command.Start, where command start issues will occur in the err, but other runtime errors will occur in a later call - for example exec.Command.Wait, here depicted by `out` which is a readcloser. Thus, one of the errors it returns can be the exec.ExitError.

WDYT about this pattern? It gives us maximum control over how we consume from the out reader, while still capturing most of the logic to run the command in runGitCommand.

cmd/gitserver/internal/git/cli/object.go
```diff
@@ -12,29 +12,26 @@ import (
...
 	"github.com/sourcegraph/sourcegraph/internal/api"
 	"github.com/sourcegraph/sourcegraph/internal/gitserver/gitdomain"
 	"github.com/sourcegraph/sourcegraph/internal/trace"
 	"github.com/sourcegraph/sourcegraph/internal/wrexec"
 	"github.com/sourcegraph/sourcegraph/lib/errors"
 )
 
-func GetObject(ctx context.Context, rcf *wrexec.RecordingCommandFactory, reposDir string, repo api.RepoName, objectName string) (_ *gitdomain.GitObject, err error) {
-	return getObject(ctx, rcf, reposDir, getObjectType, revParse, repo, objectName)
+func (g *gitCLIBackend) GetObject(ctx context.Context, objectName string) (*gitdomain.GitObject, error) {
```
eseliger, line 22: nice benefit of using the CLI: At the point where you already have a backend constructed, the interface for methods is fairly straightforward, context and actual args to the call, but no longer need to pass various helpers like recording factory etc down here.

cmd/gitserver/internal/git/cli/util.go
eseliger, line 1: just moved this code into the cli package

cmd/gitserver/internal/git/config.go
eseliger, line 1: moved to the cli package, but not recognized as a rename.

cmd/gitserver/internal/git/iface.go
eseliger, line 1: main entrypoint for talking to git in the API layer, gives an abstraction between git logic and the API layer, and let's us swap out the backend later on by something more performant without needing to touch the API layer directly.

cmd/gitserver/internal/p4exec.go
eseliger, line 1: Moved to here because this is the only thing still using it, and we're going to remove this file after the 5.3 release.

cmd/gitserver/internal/patch.go
```diff
@@ -127,7 +128,7 @@ func (s *Server) createCommitFromPatch(ctx context.Context, req protocol.CreateC
 	// Temporary logging command wrapper
 	prefix := fmt.Sprintf("%d %s ", atomic.AddUint64(&patchID, 1), repo)
 	run := func(cmd *exec.Cmd, reason string) ([]byte, error) {
-		if !gitdomain.IsAllowedGitCmd(logger, cmd.Args[1:], repoDir) {
+		if !cli.IsAllowedGitCmd(logger, cmd.Args[1:], common.GitDir(tmpRepoDir)) {
```
eseliger, line 131: TODO for later: Once everything is implemented in the backend layer, this method should no longer be exported and used here.

cmd/gitserver/internal/server_grpc.go
```diff
@@ -109,7 +112,6 @@ func (gs *GRPCServer) Exec(req *proto.ExecRequest, ss proto.GitserverService_Exe
 	internalReq := protocol.ExecRequest{
 		Repo:      api.RepoName(req.GetRepo()),
 		Args:      byteSlicesToStrings(req.GetArgs()),
-		Stdin:     req.GetStdin(),
```
eseliger, line 112: wasn't used, removing it to simplify what the backend.Exec method has to support.

cmd/gitserver/internal/server_grpc.go
```diff
@@ -177,8 +179,16 @@ func (gs *GRPCServer) Archive(req *proto.ArchiveRequest, ss proto.GitserverServi
 		})
 	})
 
+	// This is a long time, but this never blocks a user request for this
+	// long. Even repos that are not that large can take a long time, for
+	// example a search over all repos in an organization may have several
+	// large repos. All of those repos will be competing for IO => we need
+	// a larger timeout.
+	ctx, cancel := context.WithTimeout(ss.Context(), conf.GitLongCommandTimeout())
+	defer cancel()
```
eseliger, line 188: Archive requests now set the timeout explicitly, so we don't need to do a wonky special-cased assertion on `args[0] == "archive"` in the exec code itself. The timeout is the same as before, so no logic change here.

internal/gitserver/client.go
```diff
@@ -308,8 +308,8 @@ type Client interface {
 	// ListBranches returns a list of all branches in the repository.
 	ListBranches(ctx context.Context, repo api.RepoName, opt BranchesOptions) ([]*gitdomain.Branch, error)
 
-	// MergeBase returns the merge base commit for the specified commits.
-	MergeBase(ctx context.Context, repo api.RepoName, a, b api.CommitID) (api.CommitID, error)
+	// MergeBase returns the merge base commit sha for the specified revspecs.
+	MergeBase(ctx context.Context, repo api.RepoName, base, head string) (api.CommitID, error)
```
eseliger, line 312: realized we sometimes pass revspecs here, so that's been false advertising through the type.

internal/gitserver/commands.go
```diff
@@ -1227,23 +1228,31 @@ func (c *clientImplementor) GetDefaultBranch(ctx context.Context, repo api.RepoN
...
+	res, err := client.MergeBase(ctx, &proto.MergeBaseRequest{
+		Repo: string(repo),
+		Base: base,
+		Head: head,
+	})
+	if err != nil {
+		return "", err
+	}
+
+	return api.CommitID(res.GetMergeBaseCommitSha()), nil
```
eseliger, line 1255: notice: Basically no logic here beyond mapping types and adding observability! We could technically even code-gen the client at this point 😬

internal/gitserver/commands_test.go
eseliger, line 1: nice thing: the gitserver client doesn't need crazy extensive tests anymore involving running the git CLI etc. It's very simple after we're done with the migrations. 

Test moved to CLI package.

internal/gitserver/v1/gitserver.proto
```diff
@@ -682,3 +691,16 @@ message PerforceUser {
   string username = 1;
   string email = 2;
 }
+
+// MergeBaseRequest is a request to find the merge base of two revspecs.
+message MergeBaseRequest {
+  // repo is the name of the repo to get the merge base for.
+  string repo = 1;
```
eseliger, line 698: q: I still hope that we'll soon be able to switch to ID-based repo targeting in gitserver, since name is unfortunately not unique. 
Anything we should do here (change the field name, change the field ordering,...) that would be a good preparation for that?

cmd/gitserver/internal/server.go
```diff
@@ -641,11 +642,6 @@ func (s *Server) ensureOperations() *operations {
 	return s.operations
 }
 
-var blockedCommandExecutedCounter = promauto.NewCounter(prometheus.CounterOpts{
```
eseliger, line 644: moved to cli package that does the meaty logic of exec now, for more encapsulation

cmd/gitserver/internal/server.go
```diff
@@ -665,34 +661,23 @@ type execStatus struct {
 // TODO(@camdencheek): once gRPC is the only consumer of this, do everything with errors
 // because gRPC can handle trailing errors on a stream.
 func (s *Server) exec(ctx context.Context, logger log.Logger, req *protocol.ExecRequest, userAgent string, w io.Writer) (execStatus, error) {
-	// 🚨 SECURITY: Ensure that only commands in the allowed list are executed.
-	// See https://github.com/sourcegraph/security-issues/issues/213.
```
eseliger, line 669: happening in the backend layer now.

cmd/gitserver/internal/git/cli/clibackend.go
```diff
@@ -0,0 +1,223 @@
...
+	// See https://github.com/sourcegraph/sourcegraph/issues/37872 for more
+	// context.
+	commitGraphCorruptionRegex = lazyregexp.NewPOSIX(`^fatal: commit-graph requires overflow generation data but has none`)
+)
+
+// stdErrIndicatesCorruption returns true if the provided stderr output from a git command indicates
+// that there might be repository corruption.
+func stdErrIndicatesCorruption(stderr string) bool {
+	return objectOrPackFileCorruptionRegex.MatchString(stderr) || commitGraphCorruptionRegex.MatchString(stderr)
+}
```
eseliger, line 223: Corruption checks are now happening here in the CLI package, because depending on the implementation, the detection logic might need adjustment, for example if we use go-git in the future.

Instead of just logging corruption and returning a exec failed error, we return a special error type now, so that the caller can decide what to do about it (either just forward the error, or trigger some additional logic, like the corruption logging to the database).
After this logic ran, all we need to fixup corruption in the next janitor run is already done, so all behavior bolted on top of what this package does feels fairly optional, thus ok to me to break the concerns up here at this line.

cmd/gitserver/internal/git/cli/clibackend.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+	n, err := l.W.Write(p)
+	l.N -= n
+	if l.N <= 0 {
+		// If we have written limit bytes, then we can include the discarded
+		// part of p in the count.
+		n = origLen
+	}
+	return n, err
+}
```
eseliger, line 173: This is a copy of what we already had in the server package, no code change here. Just moving it to where it is encapsulated best.

cmd/gitserver/internal/server.go
```diff
@@ -873,56 +851,15 @@ func (s *Server) setLastErrorNonFatal(ctx context.Context, name api.RepoName, er
...
-		reason := stderr
-		if err := s.DB.GitserverRepos().LogCorruption(ctx, repo, reason, s.Hostname); err != nil {
+func (s *Server) logIfCorrupt(ctx context.Context, repo api.RepoName, dir common.GitDir, err error) {
+	var corruptErr cli.ErrRepoCorrupted
+	if errors.As(err, &corruptErr) {
+		if err := s.DB.GitserverRepos().LogCorruption(ctx, repo, corruptErr.Stderr, s.Hostname); err != nil {
 			s.Logger.Warn("failed to log repo corruption", log.String("repo", string(repo)), log.Error(err))
 		}
 	}
 }
```
eseliger, line 861: Now based on detecting a well-known error type, so the string matching is more encapsulated in the backend.

cmd/gitserver/internal/server.go
```diff
@@ -873,56 +851,15 @@ func (s *Server) setLastErrorNonFatal(ctx context.Context, name api.RepoName, er
...
-	}
-
-	return true
-}
-
-// stdErrIndicatesCorruption returns true if the provided stderr output from a git command indicates
-// that there might be repository corruption.
-func stdErrIndicatesCorruption(stderr string) bool {
-	return objectOrPackFileCorruptionRegex.MatchString(stderr) || commitGraphCorruptionRegex.MatchString(stderr)
-}
```
eseliger, line 924: moved to backend

internal/gitserver/v1/gitserver.proto
eseliger: we don't use 1 now, reserve it for future use, and use anything else < 16 here. 

we also change the name to repo_name.

cmd/gitserver/internal/server_grpc.go
```diff
@@ -627,3 +637,62 @@ func byteSlicesToStrings(in [][]byte) []string {
...
+
+	// TODO: This should be included in requests where we do ensure the revision exists.
+	// gs.Server.ensureRevision(ctx, repoName, "THE REVISION", repoDir)
+
+	backend := gs.backendForRepo(ctx, repoName, repoDir)
+
+	sha, err := backend.MergeBase(ctx, req.GetBase(), req.GetHead())
+	if err != nil {
+		// TODO: Better error checking.
+		return nil, status.New(codes.Internal, err.Error()).Err()
```
eseliger, line 686: we could also just return `err` here and it would get codes.Unknown.

cmd/gitserver/internal/perforce/perforce.go
```diff
@@ -232,7 +232,7 @@ func headCommitSHA(ctx context.Context, dir common.GitDir) (string, error) {
 
 	output, err := cmd.Output()
 	if err != nil {
-		return "", &common.GitCommandError{Err: err, Output: string(output)}
+		return "", errors.Wrap(err, "failed to get head commit sha: "+string(output))
```
eseliger, line 235: we never used this special error type, so changing to a standard err here to hide something called CommandError in the clibackend package.

cmd/gitserver/internal/git/cli/clibackend.go
```diff
@@ -0,0 +1,217 @@
...
+
+const gitCommandTimeout = time.Minute
+
+func (g *gitCLIBackend) gitCommand(ctx context.Context, args ...string) (wrexec.Cmder, context.CancelFunc, error) {
+	var cancel context.CancelFunc = func() {}
+	// If no deadline is set, use the default git command timeout.
+	if _, ok := ctx.Deadline(); !ok {
+		ctx, cancel = context.WithTimeout(ctx, gitCommandTimeout)
+	}
```
eseliger, line 80: this is a slight change in behavior -- previously we would apply a timeout all the time. Now, we only _overwrite_ timeouts. I think this is ultimately the right thing, or even a check here that overwrites given timeouts if > 1min, but it might currently cause unexpected deadlines where the client propagates very long deadlines. 
WDYT - should we revert or change to >1min overwrites for now?

cmd/gitserver/internal/server_grpc.go
```diff
@@ -627,3 +636,56 @@ func byteSlicesToStrings(in [][]byte) []string {
...
+	}
+
+	// TODO: This should be included in requests where we do ensure the revision exists.
+	// gs.Server.ensureRevision(ctx, repoName, "THE REVISION", repoDir)
+
+	backend := gs.Server.GetBackendFunc(repoDir, repoName)
+
+	sha, err := backend.MergeBase(ctx, string(req.GetBase()), string(req.GetHead()))
+	if err != nil {
+		// TODO: Better error checking.
```
eseliger, line 684: I don't know what kinds of errors we would want to propagate here exactly, basically for every possible error we would need to return a well-typed error from the git package and remap it to gRPC details / codes here. 
WDYT - worth it? Keep as is? The exec endpoint didn't do anything more sophisticated here, but some client methods have some error code mapping IIRC - so maybe something that we would want to port when we migrate those.

cmd/gitserver/internal/server_grpc.go
```diff
@@ -627,3 +636,56 @@ func byteSlicesToStrings(in [][]byte) []string {
...
+			Repo:            req.GetRepoName(),
+		})
+		if err != nil {
+			return nil, err
+		}
+		return nil, s.Err()
+	}
+
+	// TODO: This should be included in requests where we do ensure the revision exists.
+	// gs.Server.ensureRevision(ctx, repoName, "THE REVISION", repoDir)
```
eseliger, line 678: this is just a note for ourselves when we migrate other endpoints that use this option - but left it here as a reminder that it might need to be added

cmd/gitserver/internal/server_grpc.go
```diff
@@ -627,3 +636,56 @@ func byteSlicesToStrings(in [][]byte) []string {
...
+
+	backend := gs.Server.GetBackendFunc(repoDir, repoName)
```
eseliger, line 680: once we migrated a few of these, we'll probably see how to nicely extract these operations into a helper method

cmd/gitserver/internal/server.go
```diff
@@ -816,31 +799,33 @@ func (s *Server) exec(ctx context.Context, logger log.Logger, req *protocol.Exec
...
+			exitStatus = commandFailedErr.ExitStatus
+			status = strconv.Itoa(exitStatus)
+			return execStatus{
+				ExitStatus: commandFailedErr.ExitStatus,
+				Stderr:     string(commandFailedErr.Stderr),
+				Err:        commandFailedErr.Unwrap(),
+			}, nil
+		}
+		return execStatus{}, execErr
+	}
```
eseliger, line 823: Don't love this massive spillover of logic here from the internal workings, but we'll be removing this endpoint, so I think that's fine for another bit.

cmd/gitserver/internal/server.go
```diff
@@ -665,34 +665,23 @@ type execStatus struct {
...
-	if !req.NoTimeout {
-		var cancel context.CancelFunc
-		ctx, cancel = context.WithTimeout(ctx, executil.ShortGitCommandTimeout(req.Args))
-		defer cancel()
-	}
+	// if !req.NoTimeout {
+	// 	var cancel context.CancelFunc
+	// 	ctx, cancel = context.WithTimeout(ctx, executil.ShortGitCommandTimeout(req.Args))
+	// 	defer cancel()
+	// }
```
eseliger, line 676: TODO: the NoTimeout flag seems dangerous. I don't feel super comfortable that we have it, as it causes very long-running git subprocesses. We could comment it back in, but that is currently overwritten by the logic that sets a deadline in the backend package, if none is set. We could also say "NoTimeout means 1h" for now?

cmd/gitserver/internal/git/cli/clibackend.go
pjlast: I think it's fine as is.

I'd rather question places in the code where the client propagates very long deadlines? I also think it'll cause some confusion if someone tries to lengthen a timeout to 2 min and then having it time out after 1 min.

cmd/gitserver/internal/git/cli/config_test.go
pjlast: Migrating to go-git is looking very alluring to me 😬

pjlast (approved):
One thing I would consider is renaming the `git/cli` package to `git/gitcli`. Reason being that, when the package is actually being used, the `git/` prefix doesn't mean anything, and everything will be called with `cli.` which feels a bit ambiguous, and a bit annoying once you need to import 2 `cli` packages.

PR #58588, opened 2023-11-27
Add datname property to SQL statement

airamare01:
This PR adds the datname row to make sure we exclude rows with empty database names and also database names assed in a parameter

## Test plan
Tested locally.

### DIFF
--- a/cmd/frontend/internal/cli/autoupgrade.go
+++ b/cmd/frontend/internal/cli/autoupgrade.go
@@ -327,7 +327,7 @@ func heartbeatLoop(logger log.Logger, db database.DB) (func(), error) {
 
 func checkForDisconnects(ctx context.Context, _ log.Logger, db database.DB) (remaining []string, err error) {
 	query := sqlf.Sprintf(`SELECT DISTINCT(application_name) FROM pg_stat_activity
-			WHERE application_name <> '' AND application_name <> %s AND application_name <> 'psql'`,
+	WHERE application_name <> '' AND application_name <> %s AND application_name <> 'psql' AND datname <> '' AND datname <> %s`,
 		appName)
 	store := basestore.NewWithHandle(db.Handle())
 	remaining, err = basestore.ScanStrings(store.Query(ctx, query))


cmd/frontend/internal/cli/autoupgrade.go
```diff
@@ -326,8 +326,8 @@ func heartbeatLoop(logger log.Logger, db database.DB) (func(), error) {
 }
 
 func checkForDisconnects(ctx context.Context, _ log.Logger, db database.DB) (remaining []string, err error) {
-	query := sqlf.Sprintf(`SELECT DISTINCT(application_name) FROM pg_stat_activity
-			WHERE application_name <> '' AND application_name <> %s AND application_name <> 'psql'`,
+	query := sqlf.Sprintf(`SELECT DISTINCT(application_name, datname) FROM pg_stat_activity
+	WHERE application_name <> '' AND application_name <> %s AND application_name <> 'psql' AND datname <> '' AND datname <> %s`,
```
DaedalusG, line 330: Hey @airamare01 I don't think this will run as is. For each string you are interpolating you need to pass in a variable to `Sprintf`, i.e. ... ``datname <> %s`, appName, <some variable>`

DaedalusG:
Is there an issue associated with this change, I could take a look at? The datname value to my understanding here should always be sg since all our dbs are named this way?

airamare01:
You're right @DaedalusG. For this draft I wanted to check if the change was logically sound but since you mentioned that each variable should be output to it's own `Sprintf` function I can make that change.

> Is there an issue associated with this change, I could take a look at? The datname value to my understanding here should always be sg since all our dbs are named this way?
No there's no open issue report opened for this but what this is trying to achieve is to make it so that Sourcegraph isn't checking for any connections on any db on the server other than sg and, optionally, psql. The reason for this change is that when the db server hosts multiple databases Sourcegraph checks for all open connections and not open connections to Sourcegraph. This attempts to fix that.

DaedalusG:
Thanks for opening this @airamare01 I think the approach you took here will work. I'd like to reproduce the issue and try this out before merging. Will take another look soon 👍🏼

PR #58591, opened 2023-11-27
SSC: Add "rate limits exceeded" feature flag to aid client testing

vdavid:
This PR creates the `rate-limits-exceeded-for-testing` feature flag and applies a limit=1 with interval=max(int32) for users that have the FF. This override takes precedence over all other types of limits, and applies both on dotcom and Cody Gateway.

## Test plan

@thenamankumar, please test this or let's test it together—again I'm not sure why, but my cURL requests to Gateway hang.

My test plan would be this:
- Enable feature flag
- Make sure the rate limit errors get triggered consistently. Test code completions, chat, and embeddings
- Disable feature flag
- Make sure rate limit errors don't get triggered

Note that Gateway-side caching might mess things up during this testing.

### DIFF
--- a/cmd/frontend/graphqlbackend/user.go
+++ b/cmd/frontend/graphqlbackend/user.go
@@ -482,6 +482,10 @@ func (r *schemaResolver) ChangeCodyPlan(ctx context.Context, args *changeCodyPla
 		return nil, errors.New("this feature is not enabled")
 	}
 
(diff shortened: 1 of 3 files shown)


thenamankumar:
Getting the following error when the ff is disabled: `Request failed: Sourcegraph Cody Gateway: unexpected status code 401: {"error":"token not from source: no source found for token"}`

These are the results after testing: 
CleanShot 2023-12-01 at 20 47 00@2x

thenamankumar (approved):

PR #58592, opened 2023-11-27
Code intel: remove introspection queries

camdencheek:
These introspection queries are leftover from a time when this code was in extensions and we didn't know what version of the server was running. Now that it is part of the bundled javascript, we can remove the GraphQL introspection since we can guarantee we've got the version as the server. 

## Test plan

CI and some clicking/hovering around


vovakulikov:
Not sure what happened with Percy snapshots but triggered re-run for failed integration tests (code insights flake I guess)

PR #58615, opened 2023-11-28
Chore: upgrade to go 1.21

camdencheek:
Upgrades to go 1.21 after updating Bazel's `rules_go` so it works with it now.

## Test plan

CI.


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/sourcegraph/sourcegraph
 
-go 1.19
+go 1.21
```
camdencheek, line 3: note: there are plenty of other `go.mod` files in the repo, but I think it's fine to update those lazily. This is also the first version that will start enforcing that the `go.mod` version matches the feature set, so it's possible we'll run into as-of-yet-unseen issues because of this

burmudar (approved):

PR #58640, opened 2023-11-29
Extensions: some prep work for removal

camdencheek:
This is some setup for getting rid of all the cruft from extensions. The first step is to make the browser extension use the codeintel packages directly rather than requiring codeintel extensions running in a web worker. This extracts and simplifies few functions to prepare for that.

## Test plan

Existing tests. Everything here should be semantics-preserving. I'll do more careful tests of the browser extension before making another release, but I'd like to move as fast as I can on the removal of the extensions cruft because it's causing pain. This should not break anything in the webapp because the webapp does not use extensions.

### DIFF
--- a/client/shared/src/api/contract.ts
+++ b/client/shared/src/api/contract.ts
@@ -31,6 +31,7 @@ export interface ScipParameters {
 
 // Extracted from FlatExtensionHostAPI so it can be implemented separately.
 // The goal is to unify this with the CodeIntelAPI in client/shared/src/codeintel/api.ts
+// TODO(camdencheek)
 export interface CodeIntelExtensionHostAPI {
     getHover: (parameters: TextDocumentPositionParameters) => ProxySubscribable<MaybeLoadingResult<HoverMerged | null>>
     getDocumentHighlights: (parameters: TextDocumentPositionParameters) => ProxySubscribable<DocumentHighlight[]>
--- a/client/shared/src/api/contract.ts
+++ b/client/shared/src/api/contract.ts
@@ -29,25 +29,9 @@ export interface ScipParameters {
     documentOccurrences: Occurrence[]
 }
 
-/**
- * This is exposed from the extension host thread to the main thread
- * e.g. for communicating  direction "main -> ext host"
- * Note this API object lives in the extension host thread
- */
-export interface FlatExtensionHostAPI {
-    /**
-     * Updates the settings exposed to extensions.
-     */
-    syncSettingsData: (data: Readonly<SettingsCascade<object>>) => void
-
-    // Workspace
-    addWorkspaceRoot: (root: clientType.WorkspaceRoot) => void
-    getWorkspaceRoots: () => ProxySubscribable<clientType.WorkspaceRoot[]>
-    removeWorkspaceRoot: (uri: string) => void
-
-    setSearchContext: (searchContext: string | undefined) => void
-
-    // Languages
+// Extracted from FlatExtensionHostAPI so it can be implemented separately.
+// The goal is to unify this with the CodeIntelAPI in client/shared/src/codeintel/api.ts
+export interface CodeIntelExtensionHostAPI {
     getHover: (parameters: TextDocumentPositionParameters) => ProxySubscribable<MaybeLoadingResult<HoverMerged | null>>
     getDocumentHighlights: (parameters: TextDocumentPositionParameters) => ProxySubscribable<DocumentHighlight[]>
     getDefinition: (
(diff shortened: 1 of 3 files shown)


client/shared/src/api/contract.ts
```diff
@@ -65,6 +49,25 @@ export interface FlatExtensionHostAPI {
...
 
     hasReferenceProvidersForDocument: (parameters: TextDocumentPositionParameters) => ProxySubscribable<boolean>
+}
+
+/**
+ * This is exposed from the extension host thread to the main thread
+ * e.g. for communicating  direction "main -> ext host"
+ * Note this API object lives in the extension host thread
+ */
+export interface FlatExtensionHostAPI extends CodeIntelExtensionHostAPI {
```
camdencheek, line 60: Weird diff, but this commit just pulls out the methods that correspond to the `CodeIntelAPI` into the named type `CodeIntelExtensionHostAPI`. Essentially just slightly unflattens the part we will still need so it can be separated from extensions.

client/shared/src/codeintel/api.ts
```diff
@@ -244,15 +264,7 @@ export function injectNewCodeintel(
...
 
-    const codeintelOverrides: Pick<
-        FlatExtensionHostAPI,
-        | 'getHover'
-        | 'getDocumentHighlights'
-        | 'getReferences'
-        | 'getDefinition'
-        | 'getLocations'
-        | 'hasReferenceProvidersForDocument'
-    > = {
```
camdencheek, line 255: This is now equivalent to the new named type `CodeIntelExtensionHostAPI`

client/shared/src/codeintel/api.ts
```diff
@@ -230,10 +230,30 @@ function newSettingsGetter(settingsCascade: SettingsCascade<Settings>): sourcegr
 // extensions, we monkey patch the old implementation with new implementations.
 // The benefit of monkey patching is that we can optionally disable if for
 // customers that choose to enable the legacy extensions.
+//
+// TODO(camdencheek): USE THIS to patch code intel into extensions
 export function injectNewCodeintel(
     old: FlatExtensionHostAPI,
     codeintelContext: sourcegraph.CodeIntelContext
 ): FlatExtensionHostAPI {
+    const codeintelOverrides = newCodeIntelExtensionHostAPI(codeintelContext)
```
camdencheek: extracted API creation into a new function so it can be used elsewhere. Also, I think it's a bit more readable.

client/shared/src/codeintel/api.ts
```diff
@@ -236,25 +236,12 @@ export function injectNewCodeintel(
...
-        get(target, prop) {
-            // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-explicit-any
-            const codeintelFunction = (codeintelOverrides as any)[prop]
-            if (codeintelFunction) {
-                return codeintelFunction
-            }
-            // eslint-disable-next-line prefer-rest-params
-            return Reflect.get(target, prop, ...arguments)
-        },
-    })
```
camdencheek: Okay. I'm _pretty_ sure that this is equivalent to `{...old, ...codeintelOverrides}`. It seems to just be intercepting every method call and forwarding it to the `codeintelOverrides` if that method exists there. Can someone check me on this?

fkling: As so often: it depends.

Object spread will evaluate the value of each property when it is created. This makes a difference for getters and setters. Example:

```js
let counter = 1
const obj = {get count() {return counter++}}
obj.count // 1
obj.count // 2

const copy = {...obj}
copy.count  // 3
copy.count  // 3 (`copy` literally has property count:3)

const proxy = new Proxy({}, {
  get(target, prop) { return obj[prop]; }
})
proxy.count // 4
proxy.count // 5
```

But if you know that `codeintelOverrides` only has "normal" data properties then yes.

camdencheek: Wow, that is sneaky. To rephrase for understanding, the "getter-ness" of a property is actually a characteristic of the object, not a characteristic of the property, so spreading into a new object loses that. 

In our case though, none of the properties of either `old` or `codeintelOverrides` are getters, just normal function properties (the object is constructed [here](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@0ac33e1d35accdc55f127aa8e278e08992f9e5ed/-/blob/client/shared/src/api/extension/extensionHostApi.ts?L41-41)).

fkling (approved):

PR #58643, opened 2023-11-29
telemetry: support float64 attributes

bobheadxi:
In Javascript, there is no way to easily determine if a number is an integer or a float, as JS numbers are [floats](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number). This makes `value: Int!` hard to use, because there is no way to ask callers to provide metadata in integer values only. To support clients (typically in JS/TS) well, we need to accept `value: Float!` - however, this is a breaking change:

```
In element #0: In field "parameters": In field "metadata": In element #1: In field "value": Expected type "Float", found 0. (line 3, column 28)
```

To make this change smoothly, we accept `value: JSONValue` in the GraphQL input and validate that it is a numeric value in the backend - this allows clients to provide both integers and floats.

For events export, the change is simpler - we add a new field `metadata = 5`, and mark the existing metadata field as deprecated (`legacy_metadata`). In proto, only the field index matters, so the rename is safe, and ensures there is no disruption for data going to BigQuery, since we can continue to use `metadata` as the JSON field and simply drop `legacy_metadata` entirely.

For existing clients, out-of-band telemetry clients (namely Cody) will need to be updated to round values to integers unless interacting with a more recent Sourcegraph version, e.g. https://github.com/sourcegraph/cody/pull/1913#discussion_r1408726932

## Test plan

- Unit tests validating GraphQL schema works as expected, including back-compat
- Unit tests on telemetry gateway field migration
- Eixsting tests

## Preview 🤩
[Preview Link](https://docs.sourcegraph.com/@telemetry-float-metadata)


nathan-downs (approved):
This approach looks good to me!

bobheadxi:
cc @BolajiOlajide as release captain - we'd appreciate if we could get this into the patch tomorrow 🙏  Because of proto changes, previous changes need to be merged in succession - this one is currently blocked on https://github.com/sourcegraph/sourcegraph/pull/58649 . I am signing off for the day, but once that is merged, tomorrow morning I'll get a backport up for this change first thing

bobheadxi:
Backport: https://github.com/sourcegraph/sourcegraph/pull/58660

PR #58666, opened 2023-11-29
Code Navigation: fix test indicator positioning

jasonhawkharris:
This PR fixes two bugs.
1) Before, the test indicator caused some strange positioning behavior with the actual file names. Now, it behaves as expected.
2) Before, folder names that happened to be named the same as a valid file extension (nix, js, ts, etc.) appeared with language icons instead of the directory icon. This has been fixed.  

### Test plan
Manual testing


camdencheek (approved):
LGTM

PR #58668, opened 2023-11-29
sg: remove duplicate watch target

camdencheek:
Looks like when we got rid of the enterprise dir, it left a duplicate entry in `watch` for symbols.

## Test plan
N/A


vdavid (approved):
STAMP!

PR #58252, opened 2023-11-10
consolidate enterprise/ contributions code in web app

sqs:
Previously, we defined sidebaritems, navitems, routes, etc., in both `enterprise/` and the non-`enterprise/` tree. This is a remnant of the non-enterprise build. This layer of indirection is no longer needed.

## Test plan

CI

PR #58671, opened 2023-11-30
[fix] Add repo settings paths back

pjlast:
Minor regression from #58252

The repo settings sidebar items have a special case for Enterprise, which we don't use anymore. The PR above removed the override, and the `Repo Permissions` and `Logs` options are no longer available from the repo settings page.

## Test plan

Verified locally.


eseliger (approved):
oh! I knew something was missing on that page but I talked myself into "this is part of the reconciliation of some of the pages we did recently" but no it's not!

Thanks!

PR #58679, opened 2023-11-30
Auto enroll user to Cody Pro

thenamankumar:
This PR, makes user automatically get enrolled into Cody Pro if `?pro=true`. This will be used by marketing page to redirect to after auth if the user clicks on Get Cody Pro. It also updates the copy for Cody web and Onboarding flow.

closes: https://github.com/sourcegraph/cody/issues/1908 
closes: https://github.com/sourcegraph/sourcegraph/issues/58616

## Test plan

- sign in with community user and open `/cody/manage?pro=true`, it should automatically enroll you into the pro tier.


vdavid (approved):
All parts look good to me!

PR #58684, opened 2023-11-30
Redact observability alert sensitiva data

andreeleuterio:
Observability alert secrets were not redacted in the site config UI, as are auth provider tokens and other secrets.

The affected values are:
1. Opsgenie api key
image

2. Slack URL
image

3. Webhook bearer token and password
image

4. Pagerduty integration key
image

## Test plan
Checked locally that the values are redacted in the UI and remain the same in the db.


internal/conf/validate.go
```diff
@@ -221,6 +221,7 @@ var siteConfigSecrets = []struct {
 	{readPath: `embeddings.accessToken`, editPaths: []string{"embeddings", "accessToken"}},
 	{readPath: `completions.accessToken`, editPaths: []string{"completions", "accessToken"}},
 	{readPath: `app.dotcomAuthToken`, editPaths: []string{"app", "dotcomAuthToken"}},
+	{readPath: `observability.alerts\.notifier\.apiKey`, editPaths: []string{"observability.alerts", "notifier.apiKey"}},
```
unknwon, line 224: Do we still need this?

andreeleuterio: good catch

internal/conf/validate.go
```diff
@@ -233,53 +234,76 @@ func UnredactSecrets(input string, raw conftypes.RawUnified) (string, error) {
...
 		}
 	}
-	unredactedSite, err := jsonc.Edit(input, newCfg.AuthProviders, "auth.providers")
+
+	unredactedSite, err := jsonc.Edit(input, newAuthProviderCfg.AuthProviders, "auth.providers")
 	if err != nil {
 		return input, errors.Wrap(err, `unredact "auth.providers"`)
 	}
 
+	var oldObservabilitySecrets []string
```
unknwon, line 285:
```suggestion
	var observabilitySecretsReplaceList []string
```

is more accurate IMO (it has both old and new 😂 )

internal/conf/validate.go
```diff
@@ -344,19 +368,23 @@ func RedactAndHashSecrets(raw conftypes.RawUnified) (empty conftypes.RawUnified,
...
-	if hashSecrets {
-		getRedactedSecret = func(secret string) string {
-			hash := hashutil.ToSHA256Bytes([]byte(secret))
-			digest := hex.EncodeToString(hash)
-			return "REDACTED-DATA-CHUNK" + "-" + digest[:10]
+	getRedactedSecret := func(secret string, hashSecrets ...bool) string {
+		if len(hashSecrets) == 0 || !hashSecrets[0] {
+			return redactedSecret
 		}
+		return redactHashString(secret)
```
unknwon, line 387: For other reviewers, arguably this approach is a superset of the static `REDACTED` literal approach, meaning we can get rid of JSON editing all together (still need reading), that can make code a lot simpler when doing "unredact".

internal/conf/validate.go
```diff
@@ -365,29 +393,54 @@ func redactConfSecrets(raw conftypes.RawUnified, hashSecrets bool) (empty confty
 	}
 
 	for _, ap := range cfg.AuthProviders {
-		if ap.Openidconnect != nil {
-			ap.Openidconnect.ClientSecret = getRedactedSecret(ap.Openidconnect.ClientSecret)
-		}
 		if ap.Github != nil {
-			ap.Github.ClientSecret = getRedactedSecret(ap.Github.ClientSecret)
+			ap.Github.ClientSecret = getRedactedSecret(ap.Github.ClientSecret, false)
```
unknwon, line 397: FYI you don't need to pass false explicitly, `getRedactedSecret` accepts a variable length in the end (i.e. no arg we treat as false),

internal/conf/validate.go
```diff
@@ -233,53 +234,76 @@ func UnredactSecrets(input string, raw conftypes.RawUnified) (string, error) {
...
+		if oa.Notifier.Webhook != nil && oa.Notifier.Webhook.BearerToken != "" {
+			oldObservabilitySecrets = append(oldObservabilitySecrets, redactHashString(oa.Notifier.Webhook.BearerToken), oa.Notifier.Webhook.BearerToken)
+		}
+		if oa.Notifier.Webhook != nil && oa.Notifier.Webhook.Password != "" {
+			oldObservabilitySecrets = append(oldObservabilitySecrets, redactHashString(oa.Notifier.Webhook.Password), oa.Notifier.Webhook.Password)
+		}
+	}
+	unredactedSite = strings.NewReplacer(oldObservabilitySecrets...).Replace(unredactedSite)
+
+	//************************
```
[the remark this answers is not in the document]
andreeleuterio: had a commit already 😄

internal/conf/validate.go
```diff
@@ -365,29 +390,54 @@ func redactConfSecrets(raw conftypes.RawUnified, hashSecrets bool) (empty confty
 	}
 
 	for _, ap := range cfg.AuthProviders {
-		if ap.Openidconnect != nil {
-			ap.Openidconnect.ClientSecret = getRedactedSecret(ap.Openidconnect.ClientSecret)
-		}
 		if ap.Github != nil {
-			ap.Github.ClientSecret = getRedactedSecret(ap.Github.ClientSecret)
+			ap.Github.ClientSecret = getRedactedSecret(ap.Github.ClientSecret, false)
```
unknwon, line 394:
```suggestion
			ap.Github.ClientSecret = getRedactedSecret(ap.Github.ClientSecret, hashSecrets)
```

unknwon (approved):

Issue #57932, opened 2023-10-26
Search Contexts with >1000 repos in JSON configuration cannot be created


morgangauth:
- **Sourcegraph version:** 5.1.9
- **Platform information:** observed on docker-compose and k8s deployments

#### Steps to reproduce:

1. Create or Edit a search context 
2. Select JSON configuration
3. Add >1000 repos
-- For ease, you can use this [context](https://cse-k8s.sgdev.org/contexts/%40morgangauth%2Fmany-repos/edit) with this json config: [searchContext1000repos.json](https://github.com/sourcegraph/sourcegraph/files/13180917/searchContext1000repos.json)
5. Click 'Test Configuration'
6. Comment out the last repo in the JSON and then test configuration again

#### Expected behavior:
scenario 1: After step 5, the search context should be successful. 
scenario 2: The error message states that repos cannot be found, so by commenting out the last repo (which does not match the error message), the error should persist if that repo actually cannot be found
*Note the expected behavior was confirmed to be working in v3.37*

#### Actual behavior:
1. After step 5, the search context returns an error that repositories cannot be found (the # of repositories not found will match the # of repos over 1000 listed in the search context)
![Screenshot 2023-10-25 at 3 58 44 PM](https://github.com/sourcegraph/sourcegraph/assets/113058716/199faa9e-82a6-4705-97e4-97202a408007)
2. After commenting out a repo which is NOT the same one listed in the error, the search context is able to be saved successfully indicating that the repo WAS able to be found, but the search context has an issue with finding repos over the limit of 1000. 
![Screenshot 2023-10-25 at 3 59 12 PM](https://github.com/sourcegraph/sourcegraph/assets/113058716/8ae03cf3-1762-40e9-992d-8ad2a7011f34)

If you would like immediate help on this, please email support@sourcegraph.com (you can still create the issue, but there are no [SLAs](https://about.sourcegraph.com/support/) on issues like there are for support requests).

morgangauth:
Occurring at https://github.com/sourcegraph/accounts/issues/7485

camdencheek:
[...]

PR #58685, opened 2023-11-30
alter function using apollo client to achieve pagination

jasonhawkharris:
This solves [#57932](https://github.com/sourcegraph/sourcegraph/issues/57932). 

A customer reported that search contexts with over a thousand repos were failing when trying to create them. 

The root cause is that the webapp was fetching repositories to validate them, but was not using pagination, and instead tried to get them all in one go. The backend limits the number of results in a page to 1000, even if the client requests more. Now, the web app pages through results.

## Test plan
Create search context exceeding the max number of results. 
Check network responses for the correct output.

### DIFF
--- a/client/web/src/enterprise/searchContexts/backend.ts
+++ b/client/web/src/enterprise/searchContexts/backend.ts
@@ -38,7 +38,7 @@ export async function fetchRepositoriesByNames(
         })
 
         if (result.error) {
-            throw new Error('failed to add repositories to paginated result')
+            throw new Error(result.error.message)
         }
 
         repos = repos.concat(result.data.repositories.nodes)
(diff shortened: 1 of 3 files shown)


client/web/src/enterprise/searchContexts/backend.ts
```diff
@@ -1,32 +1,51 @@
...
+            query,
+            variables: {
+                names,
+                first,
+                after,
+            },
+        })
+
+        if (result.error) {
+            throw new Error('failed to add repositories to paginated result')
```
camdencheek, line 41: should we just throw the message inside `result.error`?

camdencheek (approved):

morgangauth (approved):
LGTM - thanks!

PR #58690, opened 2023-11-30
Update VsCode.tsx

danielmarquespt:
Links the Open Marketplace button to the marketplace

## Test plan

Tested locally using `sg start dotcom`


vdavid (approved):
Yeah, this approach is more convenient. I trust your testing, @danielmarquespt, but in PRs like this, I normally appreciate a before-after screenshot pair to take a quick look :)

Issue #58609, opened 2023-11-28
Implement a transcript clustering tool


cbart:
This has been first prototyped in https://github.com/sourcegraph/sourcegraph/pull/58498, we need the following improvements:

- use embeddings and vector similarity for clustering
- allow easily to browse through the clusters so that we can try the tool with 10/20 clusters based on a transcript input

sharing with @rishabhmehrotra so that we have common working area for developing a content-based metric

PR #58691, opened 2023-11-30
Dev tool: python script for text clustering based on local embeddings

chwarwick:
This adds a script that can be used to cluster strings based on their locally generating embeddings and outputs the data.

resolves https://github.com/sourcegraph/sourcegraph/issues/58609

## Test plan
n/a dev script only 
manually tested


cbart (approved):
Thanks!

PR #58696, opened 2023-11-30
sg/msp: better templates on init

bobheadxi:
Currently the service templates generated by `sg msp init` are pretty bad - the output is alphabetically sorted, so pretty useless, and cant include docstrings.

This changes the generated templates to be handwritten, so that we can make them ✨ artisanal ✨ 

## Test plan

```
sg msp init -owner robert msp-example-service
sg msp init -kind job -owner robert msp-example-job
```


dev/sg/msp/repo/repo.go
```diff
@@ -15,7 +15,7 @@ import (
 // UseManagedServicesRepo is a cli.BeforeFunc that enforces that we are in the
 // sourcegraph/managed-services repository by setting the current working
 // directory.
-func UseManagedServicesRepo(c *cli.Context) error {
+func UseManagedServicesRepo(_ *cli.Context) error {
```
unknwon, line 18: fyi you can do

```suggestion
func UseManagedServicesRepo(*cli.Context) error {
```

unknwon (approved):

dev/sg/msp/repo/repo.go
bobheadxi: aha, right 😁

PR #58703, opened 2023-12-01
Recreate repo statistics every Sunday morning

mrnugget:
We've noticed over the past year that sometimes there can be drift in the `repo_statistics` and `gitserver_repo_statistics` tables. We still don't know why, our strongest hypothesis is that it's because someone hand-modifies the `repo` table.

In any case: it doesn't hurt to make the statistics self-repairing.

The only downside is that the worker requires an exclusive lock on the `repo` and `gitserver_repos` tables, which are some of our busiest tables.

That's why we only run it when the instance is presumably _not_ busy and we also use a 5min timeout.

## Test plan

- Unit tests
- Manually ran the worker every 1s to make sure it doesn't run

## Preview 🤩
[Preview Link](https://docs.sourcegraph.com/@mrn/recreate-repo-statistics)

### DIFF
--- a/cmd/worker/internal/repostatistics/resetter.go
+++ b/cmd/worker/internal/repostatistics/resetter.go
@@ -1,0 +1,84 @@
+package repostatistics
+
+import (
+	"context"
+	"time"
+
+	"github.com/sourcegraph/log"
+
+	"github.com/sourcegraph/sourcegraph/cmd/worker/job"
+	workerdb "github.com/sourcegraph/sourcegraph/cmd/worker/shared/init/db"
+	"github.com/sourcegraph/sourcegraph/internal/database"
+	"github.com/sourcegraph/sourcegraph/internal/env"
+	"github.com/sourcegraph/sourcegraph/internal/goroutine"
+	"github.com/sourcegraph/sourcegraph/internal/observation"
+)
+
+// resetter is a worker responsible for deleting all rows in
+// `repo_statistics`/`gitserver_repos_statistics` and recreating their value.
+//
+// This is done because sometimes the statistics might get out of sync
+// (possibly due to people modifying the `repo` table by hand)
+type resetter struct{}
+
+func NewResetter() job.Job {
+	return &resetter{}
+}
+
+func (j *resetter) Description() string {
+	return ""
+}
+
+func (j *resetter) Config() []env.Config {
+	return nil
+}
+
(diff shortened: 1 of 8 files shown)


cmd/worker/internal/repostatistics/resetter.go
```diff
@@ -0,0 +1,90 @@
...
+	h.logger.Info("Deleting and recreating repo statistics")
+
+	// We use a 5 minute timeout here to prevent the table lock from taking
+	// down the instance.
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
+	defer cancel()
+	return h.store.DeleteAndRecreateStatistics(ctx)
+}
+
+func (h *resetterHandler) HandleError(err error) {
```
eseliger, line 88: do we need to overwrite HandleError here? Wouldn't it log it anyways?

internal/database/repo_statistics.go
```diff
@@ -164,6 +165,76 @@ FROM gitserver_repos_statistics
...
 `
 
+func (s *repoStatisticsStore) DeleteAndRecreateStatistics(ctx context.Context) (err error) {
+	tx, err := s.Transact(ctx)
+	if err != nil {
+		return err
+	}
+	defer func() { err = tx.Done(err) }()
+
+	// First lock repo, since we have triggers that write repo and cause updates on
```
eseliger, line 175: I think this could deadlock - should we retry if that happens? 

Scenario:

```
tx1: SELECT repo_id FROM gitserver_repos FOR UPDATE // ok
tx2: LOCK repo IN EXCLUSIVE MODE // ok, no other locks, proceed
tx1: SELECT id FROM repo FOR UPDATE // waits for tx2
tx2: LOCK gitserver_repos IN EXCLUSIVE MODE // waits for tx1 which got a lock at the beginning
```

eseliger (approved):
let's see if this causes bigger trouble on S2 or dotcom - I'm a bit scared of these stop the world locks, but we can see what it means in practice. 

Should we initially run it on Monday afternoon instead, so no one has to get paged this weekend - should things go wrong?

internal/database/repo_statistics.go
mrnugget: Done. Please re-review.

eseliger (approved):

PR #58704, opened 2023-12-01
bazel+rbac: mention correct target in generated files

burmudar:
previously the mentioned target was the actual program that generated the code but pretty useless invoking on its own as a user. Updated it now to mention the **actual** target one should use.

## Test plan
* `bazel run //dev:write_all_generated`
* `bazel run //client/web/src/rbac:write_generated`
* `bazel run //internal/rbac:write_generated`


internal/rbac/gen/yamldata.go
```diff
@@ -10,13 +10,16 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+const (
+	GoGeneratedByTarget = "// Code generated by //internal/rbac:write_generated. DO NOT EDIT."
```
burmudar, line 14: It would be nicer to infer the executing target but there doesn't seem to be a nice why to do it 🤔

jhchabran (approved):

PR #58710, opened 2023-12-01
repo statistics: coalesce when compacting table

mrnugget:
This is an edge case, because we should always have `0` values in the table, but maybe sometimes we don't due to manual editing or something. So let's be on the safe side.

## Test plan

- Manual testing of both queries in SQL console
- Unit tests

### DIFF
--- a/internal/database/repo_statistics_test.go
+++ b/internal/database/repo_statistics_test.go
@@ -419,6 +419,15 @@ func TestRepoStatistics_Compaction(t *testing.T) {
 	ctx := context.Background()
 	s := &repoStatisticsStore{Store: basestore.NewWithHandle(db.Handle())}
 
+	// Empty table, should work, even though we should always have a row in
+	// there with 0 values.
+	if err := s.Exec(ctx, sqlf.Sprintf("DELETE FROM repo_statistics")); err != nil {
+		t.Fatalf("failed to query repo name: %s", err)
+	}
+	if err := s.CompactRepoStatistics(ctx); err != nil {
+		t.Fatalf("CompactRepoStatistics failed: %s", err)
+	}
+
 	shards := []string{
 		"shard-1",
 		"shard-2",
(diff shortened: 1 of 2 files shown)


internal/database/repo_statistics_test.go
eseliger, line 1: Curious, did these tests fail before the addition of coalesce?

internal/database/repo_statistics_test.go
```diff
@@ -497,6 +506,15 @@ func TestGitserverReposStatistics_Compaction(t *testing.T) {
 	ctx := context.Background()
 	s := &repoStatisticsStore{Store: basestore.NewWithHandle(db.Handle())}
 
+	// Empty table, should work, even though we should always have a row in
+	// there with 0 values.
+	if err := s.Exec(ctx, sqlf.Sprintf("DELETE FROM gitserver_repos_statistics;")); err != nil {
```
eseliger, line 511: Giganit: 

```suggestion
	if err := s.Exec(ctx, sqlf.Sprintf("DELETE FROM gitserver_repos_statistics")); err != nil {
```

internal/database/repo_statistics_test.go
```diff
@@ -419,6 +419,15 @@ func TestRepoStatistics_Compaction(t *testing.T) {
 	ctx := context.Background()
 	s := &repoStatisticsStore{Store: basestore.NewWithHandle(db.Handle())}
 
+	// Empty table, should work, even though we should always have a row in
+	// there with 0 values.
+	if err := s.Exec(ctx, sqlf.Sprintf("DELETE FROM repo_statistics;")); err != nil {
```
eseliger, line 424: teranit: 

```suggestion
	if err := s.Exec(ctx, sqlf.Sprintf("DELETE FROM repo_statistics")); err != nil {
```

eseliger (approved):

internal/database/repo_statistics_test.go
mrnugget: Yep, one of them did. This crashes:

```sql
delete from repo_statistics;

WITH deleted AS (
	DELETE FROM repo_statistics
	RETURNING
		total,
		soft_deleted,
		not_cloned,
		cloning,
		cloned,
		failed_fetch,
		corrupted
)
INSERT INTO repo_statistics (total, soft_deleted, not_cloned, cloning, cloned, failed_fetch, corrupted)
SELECT
	SUM(total),
	SUM(soft_deleted),
	SUM(not_cloned),
	SUM(cloning),
	SUM(cloned),
	SUM(failed_fetch),
	SUM(corrupted)
FROM deleted;
```

But only if you delete the single row that should be there due to the migration.

PR #58732, opened 2023-12-04
Allow refreshing usage limits for dotcom users

rafax:
This PR adds a `/v1/limits/refresh` API endpoint to Cody Gateway, allowing an authenticated user to trigger a refresh of their limit (bypassing the Cody Gateway cache), [context](https://sourcegraph.slack.com/archives/C05SZB829D0/p1701364254661389?thread_ts=1701187278.753019&cid=C05SZB829D0). This endpoint:

- doesn't accept any parameters or return any values
- blocks until the refresh is complete
- replaces a cached value with a newly read value 
-- in case of errors, the cached value is not updated

This is currently only implemented for sourcegraph.com users using a `dotcom-user` Source (i.e. not Product Subscriptions). 

As a [follow-up](https://github.com/sourcegraph/sourcegraph/issues/58734) we should make sure people don't use it to refresh the token "too often" (making sourcegraph.com do more work). 
## Test plan

- some unit tests + tested locally (this relies on integration of multiple system, and is not trivally unit testable)

### DIFF
--- a/cmd/cody-gateway/internal/actor/source.go
+++ b/cmd/cody-gateway/internal/actor/source.go
@@ -60,6 +60,12 @@ type SourceSyncer interface {
 	Sync(ctx context.Context) (int, error)
 }
 
+type SourceSingleSyncer interface {
+	Source
+	// SyncOne retrieves a single actor from this source and updates its cache.
+	SyncOne(ctx context.Context, token string) error
+}
+
 type Sources struct{ sources []Source }
 
 func NewSources(sources ...Source) *Sources {
@@ -144,6 +150,23 @@ func (s *Sources) SyncAll(ctx context.Context, logger log.Logger) error {
 
 	logger.Info("All sources synced")
 	return nil
+}
+
+// SyncOne immediately runs a sync on the source implementing SourceSingleSyncer that can sync for a given token.
+// Syncing is done sequentially, first error is returned - this mirrors the behaviour of Source.Get()
+//
+// By default, this is only used by "/v1/limits/refresh" endpoint.
+func (s *Sources) SyncOne(ctx context.Context, token string) error {
+	for _, src := range s.sources {
+		if src, ok := src.(SourceSingleSyncer); ok {
+			err := src.SyncOne(ctx, token)
+			if err != nil {
+				return errors.Wrapf(err, "failed to sync %s", src.Name())
+			}
+			return nil
+		}
+	}
+	return errors.Newf("no source found for token %v", token[:4])
 }
 
 // Worker is a goroutine.BackgroundRoutine that runs any SourceSyncer implementations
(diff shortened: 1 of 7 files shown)


vdavid (approved):
Thanks a lot, the code looks good, and this should be very helpful to us!

eseliger (approved):
> As a https://github.com/sourcegraph/sourcegraph/issues/58734 we should make sure people don't use it to refresh the token "too often" (making sourcegraph.com do more work).

Agreed on that, otherwise LGTM

cmd/cody-gateway/internal/actor/source.go
```diff
@@ -60,6 +60,12 @@ type SourceSyncer interface {
 	Sync(ctx context.Context) (int, error)
 }
 
+type SourceSingleSyncer interface {
+	Source
+	// SyncOne retrieves a single actor from this source and updates its cache.
+	SyncOne(ctx context.Context, token string) error
```
bobheadxi, line 66: nice 😁

bobheadxi: actually, looking at this again @rafax , we already have this exact interface:

```go
type SourceUpdater interface {
	Source
	// Update updates the given actor's state, though the implementation may
	// decide not to do so every time.
	Update(ctx context.Context, actor *Actor)
}
```

If you use the actor middleware as usual, you can extract the actor from context and just provide it to the updater implementation, or use the existing update helper:

https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/cmd/cody-gateway/internal/actor/actor.go?L108:1-120:2

This helper is currently already called when we run into a rate limit error:

https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/cmd/cody-gateway/internal/actor/limits.go?L148:1-152:3

The docstring indicates we don't have to update every single time, which is still a valuable trait to avoid some bug or other issue abusing the refresh

cmd/cody-gateway/internal/httpapi/featurelimiter/featurelimiter.go
```diff
@@ -209,6 +210,23 @@ func ListLimitsHandler(baseLogger log.Logger, eventLogger events.Logger, redisSt
...
+
+		token, err := authbearer.ExtractBearer(r.Header)
+		if err != nil {
+			response.JSONError(logger, w, http.StatusBadRequest, err)
+		}
+
+		err = sources.SyncOne(r.Context(), token)
+		if err != nil {
+			response.JSONError(logger, w, http.StatusBadRequest, err)
+		}
```
bobheadxi, line 226: This could just be:

```go

func RefreshLimitsHandler(baseLogger log.Logger, sources *actor.Sources) http.Handler {
		act := actor.FromContext(r.Context())
		act.Update(r.Context())
	})
}
```

using the existing helper

cmd/cody-gateway/internal/actor/source.go
bobheadxi: Was in this part of the code again so spun up a quick cleanup: https://github.com/sourcegraph/sourcegraph/pull/59102

PR #58735, opened 2023-12-04
sg generate: report duration for bazel step

mrnugget:
Previously it would always report 0s for Bazel, even though it takes way longer than that (and many more cores :P).

This change returns the report as-is which means the duration reporting is correct.

## Test plan

- Run `go run ./dev/sg generate` and see `✅ Target "bazel" done (116s)`


burmudar (approved):

burmudar:
Thanks!

PR #58748, opened 2023-12-04
Remove unused OAuth stateconfig

evict:
After recent session improvements, we no longer need statecookie helpers as we are now using the session store for the oauthstate. This PR:

- Removes StateConfig field from provider structs 
- Deletes getStateConfig helper method
- Removes unused state cookie constants

## Test plan
CI tests.

## Preview 🤩
[Preview Link](https://docs.sourcegraph.com/@vincent/oauth2-session-clean-up)


cmd/frontend/internal/auth/bitbucketcloudoauth/provider.go
```diff
@@ -89,8 +88,8 @@ func failureHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	ctx := r.Context()
-	encodedState, err := goauth2.StateFromContext(ctx)
+	var encodedState string
+	err := session.GetData(r, "oauthState", &encodedState)
```
unknwon, line 92: Hmm, this means... bitbucketcloudoauth was broken? 😂 (and we backported the changes to 5.3.4.. so that version is broken with bitbucketcloudoauth)

unknwon (approved):

PR #58759, opened 2023-12-04
Use clear map function from go 1.21

jtibshirani:
A tiny change to use the new `clear` function in go 1.21 instead of deleting
each entry manually. This was the only place I saw where we could use it.

## Test plan

Covered by existing tests


camdencheek (approved):
Nice!

unknwon (approved):

PR #58766, opened 2023-12-05
Improve cursor based pagination implementation

eseliger:
This PR makes more clear how cursors are meant to be used, namely, not an arbitrary string but instead using the actual values passed to the SQL prepared statement.

## Test plan

Added some additional tests, didn't break existing tests.

### DIFF
--- a/cmd/frontend/graphqlbackend/graphqlutil/connection_resolver.go
+++ b/cmd/frontend/graphqlbackend/graphqlutil/connection_resolver.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/sourcegraph/sourcegraph/internal/database"
 	"github.com/sourcegraph/sourcegraph/lib/errors"
+	"github.com/sourcegraph/sourcegraph/lib/pointers"
 )
 
 const DefaultMaxPageSize = 100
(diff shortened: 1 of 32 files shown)


internal/database/perms_store_test.go
```diff
@@ -3994,7 +3994,7 @@ func TestPermsStore_ListUserPermissions(t *testing.T) {
 			Name:   "TestPagination",
 			UserID: 555,
 			Args: &ListUserPermissionsArgs{
-				PaginationArgs: &PaginationArgs{First: pointers.Ptr(2), After: pointers.Ptr("'public_repo_5'"), OrderBy: OrderBy{{Field: "name"}}},
+				PaginationArgs: &PaginationArgs{First: pointers.Ptr(2), After: []any{"public_repo_5"}, OrderBy: OrderBy{{Field: "name"}}},
```
eseliger, line 3997: no more quoting needed in the After here is what the goodie of the PR is

cmd/frontend/graphqlbackend/graphqlutil/connection_resolver.go
```diff
@@ -20,13 +21,13 @@ type ConnectionResolver[N any] struct {
...
-	ComputeTotal(context.Context) (*int32, error)
+	ComputeTotal(context.Context) (int32, error)
 	// ComputeNodes returns the list of nodes based on the pagination args.
 	ComputeNodes(context.Context, *database.PaginationArgs) ([]N, error)
 	// MarshalCursor returns cursor for a node and is called for generating start and end cursors.
 	MarshalCursor(N, database.OrderBy) (*string, error)
-	// UnmarshalCursor returns node id from after/before cursor string.
-	UnmarshalCursor(string, database.OrderBy) (*string, error)
+	// UnmarshalCursor returns SQL values to be used in the query for fetching nodes based on cursor.
+	UnmarshalCursor(string, database.OrderBy) ([]any, error)
```
eseliger, line 30: The main change: We now return the values that are meant to be passed to SQL directly instead of enumerating them in a single string.

I tried hard to type this better, but some connections need to order by multiple columns with differing datatypes, so that was hard to fit into the current pattern without a bigger rewrite - at least I couldn't think of a simple-ish solution.

internal/insights/scheduler/backfill.go
```diff
@@ -379,11 +379,11 @@ func (s *BackfillStore) GetBackfillQueueInfo(ctx context.Context, args BackfillQ
 	// therefore we can't use the where clause it generates.  Below builds the correct where from the before or after
 	// from the cursor
 
-	if pagination.After != nil {
-		where = append(where, sqlf.Sprintf("isb.id > %s", *pagination.After))
+	if len(pagination.After) > 0 {
+		where = append(where, sqlf.Sprintf("isb.id > %s", pagination.After[0].(int32)))
```
eseliger, line 383: This is unfortunate and brittle. Will need to revisit the pattern more holistically.

evict (approved):

PR #58768, opened 2023-12-05
bazel: write docsite script to .bin

burmudar:
Tried removing the script with a trap in the sg command but you can't remove the script after it has started - even with a trap since sg gets killed before the trap is executed. So instead we put the script in `.bin` which is _technically the same as what other commands do ... the command just produces a script artifact.

## Test plan
Tested locally


sg.config.yaml
```diff
@@ -463,7 +463,13 @@ commands:
...
   docsite:
     description: Docsite instance serving the docs
-    cmd: bazel run --script_path=docsite_run.sh --noshow_progress --noshow_loading_progress //doc:serve && ./docsite_run.sh
+    env:
+      RUN_SCRIPT_NAME: .bin/bazel_run_docsite.sh
+    cmd: |
+      bazel run --script_path=${RUN_SCRIPT_NAME} --noshow_progress --noshow_loading_progress //doc:serve
+
+      ./${RUN_SCRIPT_NAME}
+
```
burmudar, line 472:
```suggestion
```

jhchabran (commented):
Can we put a comment explaining why we need to do this?

burmudar:
> Can we put a comment explaining why we need to do this?

Done

eseliger (approved):

PR #58776, opened 2023-12-05
emacs: hack for integrating src search into consult--grep

keegancsmith:
Lots of hardcoding/etc, but just for a demo.


keegancsmith:
abandonship

Issue #58773, opened 2023-12-05
Sidebar Symbols: broken navigation for symbols with identical names


bevzzz:
- **Sourcegraph version:** 5.2.4 ([current](https://sourcegraph.com/search))
- **Platform information:** n/a

In some cases symbol names are not unique within their parent scope.
For example:
-  Overloaded methods/functions (Java, C++, etc): [`StringUtils.join`](https://sourcegraph.com/github.com/stanfordnlp/CoreNLP/-/blob/src/edu/stanford/nlp/util/StringUtils.java?L440:25-440:29)
- Type conversion with blank identifier (Go): [`var _  Interface = (*Implementation)(nil)`](https://sourcegraph.com/github.com/uptrace/bun@8a4383505d7e954897b23811a412b9cdafaf41eb/-/blob/model_table_slice.go?L113:3-113:4)

#### Steps to reproduce:

1. Open one of the two links above
2. Try to navigate between different symbols using Symbol Sidebar

#### Expected behavior:

Clicking on different symbols in the Sidebar navigates to different sections of the code, selecting the correct line. 

#### Actual behavior:

All symbols with identical names navigate to the last occurrence of the symbol's name.

<details><summary>Click to expand</summary>

![recording](https://github.com/sourcegraph/sourcegraph/assets/53943884/bba9be7f-7bef-4a6d-8be0-92b73fecb038)

</details>

This happens because `hierarchyOf` uses `containerName + symbolName` as the unique identifier, which leaves `fullNameToSymbol` map with only the last occurrence of a symbol and its URL.

https://github.com/sourcegraph/sourcegraph/blob/ec506b663b1b19062bc2c6df23ec36a0b52c4043/client/web/src/repo/RepoRevisionSidebarSymbols.tsx#L142-L148

Issue #58392, opened 2023-11-17
Symbol sidebar selection not updated after selecting another line.


peterguy:
- **Sourcegraph version:** https://sourcegraph.com/search

When navigating around a file using the Symbol sidebar, the symbol selection does not update/deselect after clicking on a line number.

#### Steps to reproduce:

1. Navigate to [a file](https://sourcegraph.com/github.com/sourcegraph/cody@b2cb9f34ff5bd9e91169c47de96d4164f6bda93c/-/blob/lib/shared/src/sourcegraph-api/graphql/client.ts)
2. Click a symbol in the Symbols sidebar.
3. Click on a different line number to highlight that line.
4. Click on the same symbol as you did in step 2.

#### Expected behavior:

1. the currently-highlighted line is de-highlighted
2. the line where the symbol is defined is highlighted
3. the view is scrolled to the symbol's line
4. the URL is updated with the location of that symbol

#### Actual behavior:

Nothing happens. The symbol is behaving as if it is still selected.

#### explanatory video

https://github.com/sourcegraph/sourcegraph/assets/129280/04f44093-9072-424a-b582-7987f1a23308

camdencheek:
Since we are focusing on the svelte webapp, I'm going to close this since it doesn't seem to be a major UX bug and we can wait until the svelte rewrite is live.

bevzzz:
@camdencheek I just noticed the issue got closed. There's a PR I made a while back (December) that solves this issue and also brings a nice improvement to how the symbols are displayed.

The PR is #58780

PR #58780, opened 2023-12-05
Improvements to symbol sidebar navigation

bevzzz:
@peterguy, as discussed, here's my PR for #58392. It's a bit larger than I initially expected, but I think still well-contained for single PR. 

The required functionality was pretty much already there, provided by `selectableLines` extension, but it was just deactivated for `CodeMirrorBlob`. This happened because `navigateToLineOnAnyClick` prop controlled too many things: code intel, keyboard navigation, other extensions, and was only set to `true` in `ReferencePanel`.

To provide finer-grain control I introduced 2 additional props:
- `codeIntelAndSgExtensions` to toggle code intel (enabled in `CodeMirrorBlob` / disabled in `ReferencePanel`)
- `navigateToLineOnLineClick` to toggle navigation on line numbers (enabled in `CodeMirrorBlob` )
- `navigateToLineOnAnyClick`  now only toggles navigation on any line-click (enabled in `ReferencePanel`)

The three options allow handling these features more flexibly and transparently. 

While working on the PR, I also noticed #58773 (it started causing some weird behaviour once I've added line-to-symbol navigation), so I included a fix for that as well.

Finally, I changed the order in which the symbols are displayed -- it used to be alphanumeric, which meant that navigation was going all over the place. They now appear  in the sidebar panel in the same order in which they appear in the file.  This was a small change, but a nice improvement. WDYT?

## Test plan

- Updates to  `CodeMirrorBlob` and `navigateToLineOnAnyClick`: manually tested
- Different algorithm for organizing symbols in`hierarchyOf` + symbol sorting: added several test cases

Below is a screen recording showing the effect of both changes on [`StringUtils.java`](https://sourcegraph.com/github.com/stanfordnlp/CoreNLP/-/blob/src/edu/stanford/nlp/util/StringUtils.java?L440:25-440:29).

<details><summary>Click to expand</summary>

SidebarNavigation

</details>


bevzzz:
Added another small improvement that I thought about: scrolling back to the selected line when its no longer visible by clicking the symbol again.   
Previously we would only scroll when the URL changed, navigating to the _already selected symbol_ was not possible:

<details><summary>Before</summary>

before

</details>

Now, **if the line is not in the viewport**, clicking on its symbol will navigate back to it:

<details><summary>After</summary>

after

</details>

bevzzz:
Hi @fkling! 
I see you were assigned as a reviewer and I wanted to say that I'm really excited about my first contribution to Sourcegraph and that if I could somehow help you review it, HMU ✌️

PR #58782, opened 2023-12-05
svelte: Add Svelte prototype toggle button to global navbar

fkling:
To make the prototype easier to test this commit adds a button to the React global navbar that makes it easy to switch to the prototype (and back).
Links that lead out of the SvelteKit version are marked with a React logo. It's a bit noisy, but at this point it's important to me that people know what happend when they click on a link. Additionally a banner tells users that they are in the prototype.

In the React version the Svelte button will only be shown on pages that have a SvelteKit equivalent.
The button is hidden behind a feature flag. The plan is to turn on that feature flag on S2.

NOTE: Ideally (and eventually), if the prototype is enabled, navigating inside the React app would bring you back to the prototype if the page is supported by it. But links inside the React app are a bit inconsistent. So I though it's better if there is less magic: If you are inside the React app you stay there until you click the button in the global navbar. If you are inside the prototype then refreshing the page or following a link marked as external will bring you back to the React app.

https://github.com/sourcegraph/sourcegraph/assets/179026/a33442dc-c34c-48b3-9d5d-9093f285db6d

## Test plan

Manual testing.


peterguy (approved):
easier than a query string! LGTM

PR #58786, opened 2023-12-05
completions: add claude-instant-1.2 to all references of claude-instant-1.2-cyan

bobheadxi:
Adds a reference to `anthropic/claude-instant-1.2` at all places where `claude-instant-1.2-cyan` are referenced. It seems this was changed in https://github.com/sourcegraph/cody/pull/1884 and might break single-tenant usage: https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1701802444496609

Interestingly enough `claude-instant-1.2-cyan` was never added to the default models, so maybe I'm looking at the wrong thing, but I've added  `anthropic/claude-instant-1.2` to the Cody Gateway default models as well just in case.

## Test plan

n/a


eseliger (approved):

philipp-spiess (approved):
anthropic/claude-instant-1.2-cyan can be deleted now (it was only rolled out for a very brief time and is now the same as claude-instant-1.2). I think I only cleaned up the client but not the server yet.

Not sure why and when we started to use `anthropic/claude-instant-1.2` as an identifier though, maybe this is related to the new chat design cc @abeatrix. For code completions we have just not set a `model` in the past and then the default configured model was used which was `claude-instant-1` I suppose.

philipp-spiess:
`anthropic/claude-instant-1.2-cyan` should not be used anymore, it was the codename for a temporary improvement that we A/B tested for about a week and is now merged into `anthropic/claude-instant-1.2`. I think 1.2 is the right identifier we just need to add it to the backends, which is what this PR is doing :)

bobheadxi:
I think I'll keep the cyan model just to be safe for now. Overall, I feel like the way we allowlist models might need some revisiting - I already raised this with @rafax - but I'm erring towards being generous with the allowlists just in case 😁

PR #58789, opened 2023-12-05
Svelte Prototype: add file icons to svelte prototype file tree

jasonhawkharris:
Adds file icons for the sveltekit prototype.

Screenshot 2023-12-05 at 3 59 14 PM

## Test plan
Visual test/Manual test


client/web-sveltekit/src/lib/Icon.svelte
```diff
@@ -13,13 +13,22 @@
     interface $$Props extends SVGAttributes<SVGElement> {
         svgPath: string
         inline?: boolean
+        iconClass: string
```
fkling, line 16: Passing classes to components isn't that common in Svelte (it's also not well supported). E.g. Svelte will tell you now that all the color classes below are unused.

The class already supports changing the color of an icon either by inheriting `color:` from the parent or via the `--color` variable.

E.g. you could do something like this instead below:

```svelte
<Icon svgPath={svgPath ?? mdiFileCodeOutline} inline --color={iconColor} />
```

No need to add new classes or cryptic prop names.

client/web-sveltekit/src/lib/repo/fileIcons.ts
fkling, line 1: Would be nice if that could be shared between the React version on this. But afaik you are using different icons there?

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte
```diff
@@ -11,22 +11,55 @@
...
+            return {
+                extension: extension as FileExtension,
+                isTest: containsTest(file),
+            }
+        }
+
+        return {
+            extension: 'default' as FileExtension,
+            isTest: false,
         }
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte:62]: I'd move that to the fileIcons file. Doesn't look like you need a `isDirectory` parameter either.

client/web-sveltekit/src/lib/repo/fileIcons.ts
```diff
@@ -0,0 +1,194 @@
...
+export interface IconInfoSvelte {
+    svgPath: string
+    iconClass: string
+}
+
+export const FILE_ICONS: Map<FileExtension, IconInfoSvelte> = new Map([
+    [FileExtension.ASSEMBLY, { svgPath: mdiFileCode, iconClass: 'default-icon' }],
+    [FileExtension.BASH, { svgPath: mdiBash, iconClass: 'default-icon' }],
+    [FileExtension.BASIC, { svgPath: mdiFileCode, iconClass: 'default-icon' }],
+    [FileExtension.C, { svgPath: mdiLanguageC, iconClass: 'blue' }],
```
fkling, line 124: Continuing the comment above, you could either rename `iconClass` to `color` and each to something like `var(--blue)`, or create helper constants:

```
const BLUE = 'var(--blue)';

...

{..., color: BLUE}
```

client/web-sveltekit/src/lib/repo/fileIcons.ts
```diff
@@ -0,0 +1,194 @@
...
+    SVG = 'svg',
+    SWIFT = 'swift',
+    TEST = 'test',
+    TYPESCRIPT = 'ts',
+    TSX = 'tsx',
+    TEXT = 'txt',
+    YAML = 'yaml',
+    YML = 'yml',
+    ZIG = 'zig',
+}
```
fkling, line 113: Don't know if it makes sense to have an enum for this. Below you are casting an arbitrary string to this enum. It's possible that the string isn't actually part of this list.

A possible alternative:

- Define a list of strings containing known file extensions.
- Have two functions whith huge `switch` statements that each return the icon and the color for the provided file extension. You have a lot of repeated data here (especially the colors), so that would avoid some repetition.
- Use those functions to build a `Map<string, {icon: string, color: string}>` for fast access.

Of course it's also fine to hardcode the map.

I guess one advantage of the enum is that the typechecker helps you to ensure that you are creating an entry for each type... for that it might be useful, but then I'd abstract that fact away from the caller. I.e. the caller should be able to pass any string (which it currently does anyway by casting to an enum value, but the caller shouldn't have to do that).

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte
```diff
@@ -11,22 +11,55 @@
     import TreeView, { setTreeContext } from '$lib/TreeView.svelte'
     import { createForwardStore } from '$lib/utils'
     import { replaceRevisionInURL } from '$lib/web'
+    import { FILE_ICONS, FileExtension } from '$lib/repo/fileIcons'
+    import { containsTest } from '@sourcegraph/web/src/repo/utils'
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte:15]: For now I'm trying to avoid imports from other packages in Svelte components. If a component (or any module really) needs something from another package, the corresponding module in `lib/` (e.g. `lib/web.ts`) should re-export that value, and the other module should import it from there (e.g. `import {containsTest} form '$lib/web'`).
I did that to concentrate all "external" dependencies in as few files as possible, so that others, should they change any of these dependencies, don't have to deal with Svelte directly.

client/web-sveltekit/src/lib/repo/fileIcons.ts
jasonhawkharris: Wouldn't two huge switch statements make this slower than hard coding everything?

> I guess one advantage of the enum is that the typechecker helps you to ensure that you are creating an entry for each type

This was my thought process when I implemented it this way.

client/web-sveltekit/src/lib/FileIcon.svelte
```diff
@@ -0,0 +1,12 @@
+<script lang="ts">
+    interface $$Props {
+        value: string
+    }
+    export let value: string | undefined
+</script>
+
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/gh/devicons/devicon@v2.15.1/devicon.min.css" />
```
jasonhawkharris, line 8: This just seems like the absolute wrong way to do this. Would love your advice @fkling

fkling: Yeah, atm that would cause this link to be included for every icon, which is probably not too bad but not great either.

But that aside I don't actually want to use an "live" external dependency. I understand that the MDI icon set is limited, but for the prototype it's OK to just stick with that.

A solution for the future might be https://github.com/unplugin/unplugin-icons

client/web-sveltekit/src/lib/repo/fileIcons.ts
```diff
@@ -0,0 +1,620 @@
...
+    [FileExtension.SVELTE, { ...getIconSvelte(FileExtension.SVELTE) }],
+    [FileExtension.SVG, { ...getIconSvelte(FileExtension.SVG) }],
+    [FileExtension.SWIFT, { ...getIconSvelte(FileExtension.SWIFT) }],
+    [FileExtension.TYPESCRIPT, { ...getIconSvelte(FileExtension.TYPESCRIPT) }],
+    [FileExtension.TSX, { ...getIconSvelte(FileExtension.TSX) }],
+    [FileExtension.TEXT, { ...getIconSvelte(FileExtension.TEXT) }],
+    [FileExtension.YAML, { ...getIconSvelte(FileExtension.YAML) }],
+    [FileExtension.YML, { ...getIconSvelte(FileExtension.YML) }],
+    [FileExtension.ZIG, { ...getIconSvelte(FileExtension.ZIG) }],
+])
```
fkling, line 620: If you are still building the map manually then you are not gaining anything from the helper/builder functions. I had something like this in mind:

```js
export const FILE_ICONS = new Map(
  Object.values(FileExtension)
    .map(value => [value, getIconsSvelte(value)])
)
```

client/web-sveltekit/src/lib/repo/fileIcons.ts
```diff
@@ -0,0 +1,620 @@
...
+            extension: extension as FileExtension,
+            isTest: containsTest(file),
+        }
+    }
+
+    return {
+        extension: 'default' as FileExtension,
+        isTest: false,
+    }
+}
```
fkling, line 179: Calling `Object.values(FileExtension).includes(extension as FileExtension)` for every file is probably expensive. It's also confusing that there is a file `FileInfo` and a `IconSvelte` type and that the caller needs to use both `getFileInfo` and `FILE_ICON` to get the information it needs.

I'd suggest the following: Export a single function that returns a single value containing all the information the caller needs. Instead of having searching for the file extension in the array you can just access `FILE_ICON`.

```js
interface FileInfo {
  icon: ...
  isTest: boolean
}

export const getFileInfo = (file: string): FileInfo => {
    const extension = file.split('.').at(-1)
    const icon = FILE_ICONS[extension]

    if (icon) {
        return {
            icon,
            isTest: containsTest(file),
        }
    }

    return {
        icon: // handle default
        isTest: false,
    }
}
```

client/web-sveltekit/src/lib/repo/fileIcons.ts
```diff
@@ -0,0 +1,620 @@
...
+
+const getIconSvelte = (extension: FileExtension): IconSvelte => {
+    return {
+        info: getIconAttributes(extension),
+        color: getColor(extension),
+    }
+}
+
+const getIconAttributes = (extension: string): IconInfo => {
```
fkling, line 199: This should be typed as `FileExtension` to get type checker support.

```suggestion
const getIconAttributes = (extension: FileExtension): IconInfo => {
```

client/web-sveltekit/src/lib/repo/fileIcons.ts
```diff
@@ -0,0 +1,620 @@
...
+            return {
+                value: mdiFileCode,
+                kind: Kind.MDI,
+            }
+        }
+    }
+}
+
+// only relevant for MDI icons
+const getColor = (extension: string): string => {
```
fkling, line 507: This should be typed as `FileExtension` to get type checker support.

```suggestion
const getColor = (extension: FileExtension): string => {
```

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte
```diff
@@ -11,22 +11,33 @@
...
     export let selectedPath: string
     export let revision: string
 
+    interface IconPath {
+        info?: {
+            value?: string | undefined
+            kind: 'mdi' | 'devicon'
+        }
+        color?: string
+    }
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte:27]: Import the type from the `fileIcons` instead.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte
```diff
@@ -11,22 +11,33 @@
...
-            return mdiFolderArrowUpOutline
+            return { info: { value: mdiFolderArrowUpOutline, kind: 'mdi' } }
         }
         if (entry.isDirectory) {
-            return open ? mdiFolderOpenOutline : mdiFolderOutline
+            return { info: { value: open ? mdiFolderOpenOutline : mdiFolderOutline, kind: 'mdi' } }
         }
-        return mdiFileCodeOutline
+        const fileInfo = getFileInfo(entry.name)
+        return { ...FILE_ICONS.get(fileInfo.extension) }
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte:40]: You don't need to copy the objects everywhere. If you want to disallow mutation you can mark properties as `readonly` (TS will complain if you write to such a property).

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte
```diff
@@ -124,8 +136,19 @@
...
+                    <span class="icons">
+                        {#if iconProps.info?.kind === 'mdi'}
+                            <Icon
+                                svgPath={iconProps?.info.value ?? mdiFileCodeOutline}
+                                inline
+                                color={iconProps.color ?? ''}
+                            />
+                        {:else if iconProps.info?.kind === 'devicon'}
+                            <FileIcon value={iconProps?.info?.value ?? ''} />
+                        {/if}
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte:150]: As I mentioned above, I'd avoid using an external source for now.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte
```diff
@@ -124,8 +136,19 @@
...
                     data-go-up={isRoot ? true : undefined}
+                    class="icons"
                 >
-                    <Icon svgPath={getIconPath(entry, expanded)} inline />
+                    <span class="icons">
+                        {#if iconProps.info?.kind === 'mdi'}
+                            <Icon
+                                svgPath={iconProps?.info.value ?? mdiFileCodeOutline}
+                                inline
+                                color={iconProps.color ?? ''}
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte:146]: You don't need to add a new prop. You should be able to just do `--color={iconProps.color ?? ''}`. It's a [style prop](https://svelte.dev/docs/component-directives#style-props).

jasonhawkharris:
There is a better way to do this, and this has ballooned.

PR #54092, opened 2023-06-23
grpc: add interceptors that convert raw context.[Cancelled/DeadlineExceeded] errors into status errors and vice versa

ggilmore:
This PR adds interceptors that convert _raw_ `context.Cancelled/DeadlineExceeded` errors into [status.Cancelled / status.DeadlineExceeded](https://grpc.github.io/grpc/core/md_doc_statuscodes.html) errors and vice versa.

For **clients**, this greatly simplifies error handling as we shouldn't need to sprinkle checks like "if status.Code(err)== Cancelled: return context.Cancelled" all over the place any more.

For **servers**, this is more of a defensive mechanism. These interceptors _only_ operate on errors that don't already have a status code associated with them, since it can't assume when it is or isn't appropriate to "override" a specific status code that a server author chose for any given method. (I am willing to revisit this after the code freeze if we find this restriction too annoying). 

Therefore, it's still important for server author to check for context cancellation errors where appropriate if they're already returning a specific status code. For example, in this RPC method:

```go 
func (gs *GRPCServer) ListGitolite(ctx context.Context, req *proto.ListGitoliteRequest) (*proto.ListGitoliteResponse, error) {
	host := req.GetGitoliteHost()
	repos, err := defaultGitolite.listRepos(ctx, host)
	if err != nil {
		return nil, status.Error(codes.Internal, err.Error())
	}

	// ...
}
```

The interceptor won't operate since the server author is already providing a specific status code. In this scenario, the explicit check is still needed:

```go
func (gs *GRPCServer) ListGitolite(ctx context.Context, req *proto.ListGitoliteRequest) (*proto.ListGitoliteResponse, error) {
	host := req.GetGitoliteHost()
	repos, err := defaultGitolite.listRepos(ctx, host)

	// Check for context errors before returning a status.Error()
	if ctxErr := ctx.Err(); ctxErr != nil {
		return nil, status.FromContextError(ctxErr).Err()
	}

	if err != nil {
		return nil, status.Error(codes.Internal, err.Error())
	}

       // ...
}
```

## Test plan

Unit test suite

PR #58790, opened 2023-12-05
grpc: make example package for how to use grpc RPC types

ggilmore:
This is my attempt at making an example gRPC package that tries to demonstrates gRPC basics including gRPC best practices, all the RPC types, error handling etc. for the engineering team.

Quoting from internal/grpc/example/doc.go:

```go
// Package example contains didactic examples for the basic usage / implementation of gRPC including:
//
// - All the basic Protobuf types (e.g. primitives, enums, messages, one-ofs, etc.)
// - All the basic RPC types (e.g. unary, server streaming, client streaming, bidirectional streaming)
// - Error handling (e.g. gRPC errors, wrapping status errors, etc.)
// - Implementing a gRPC server (with proper separation of concerns)
// - Implementing a gRPC client
// - Some known footguns (non-utf8 strings, huge messages, etc.)
// - Some Sourcegraph specific helper packages and patterns (grpc/defaults, grpc/streamio, etc.)
//
//	The service is a simple weather-reporting service that allows you to query the current weather, send updated weather
//	location for a given sensor, receive severe weather alerts and more.
//
//	The examples are organized into the following directories:
//	- weather/v1: contains the protobuf definitions for the weather service
//	- server/: contains the implementation of the gRPC server
//	- client/: contains the implementation of the gRPC client
//
//
//	When going through this example for the first time, it is recommended to:
//
//	1. Read the protobuf definitions in weather/v1/weather.proto to get a sense of the service.
//	2. Run the server and client examples in server/ and client/ (via server/run-server.sh and client/run-client.sh) respectively to see the service in action.
//	3. Read the implementation of the server and client to get a sense of how things are implemented, and follow the explanatory comments in the code.
package example
```

## Test plan

This is an example package that doesn't get compiled into our main application.

You can run the examples by first starting the server with the provide `run-server.sh` script, and the client with `run-client.sh`

Here a record of what the output of both commands should look like:

[asciicast](https://asciinema.org/a/ETly3qO5MSHvrG3Za3rZI4UDA)


internal/grpc/example/weather/v1/weather.proto
```diff
@@ -0,0 +1,55 @@
...
+  // Client Streaming RPC: Send weather data from sensors.
+  rpc UploadWeatherData(stream SensorData) returns (UploadStatus) {}
+
+  // Bidirectional Streaming RPC: Real-time weather updates.
+  rpc RealTimeWeather(stream LocationUpdate) returns (stream WeatherResponse) {}
+}
+
+// Request and response messages.
+message LocationRequest {
+  string location = 1;
```
eseliger, line 24: maybe we could add something to the example about "non UTF-8 string data has to be bytes"?

internal/grpc/example/weather/v1/weather.proto
eseliger, line 1: Maybe we can add one additional endpoint that is deprecated?

internal/grpc/example/server/main.go
```diff
@@ -0,0 +1,171 @@
...
+		// - https://godoc.org/google.golang.org/grpc/codes
+		// - https://cloud.google.com/apis/design/errors (intended for Google developers, but generally applicable advice)
+		err := status.Errorf(codes.InvalidArgument, "We don't go to Ravenholm.")
+		return nil, err
+	}
+
+	// Now that we know that we have a "valid" location, we can safely return the response.
+	response := &pb.WeatherResponse{
+		Description: "It's always Sunny!",
+		Temperature: 25.0,
```
eseliger, line 54: idea: we should make temperature a nested object.
That would both show that you can nest objects, and we could add a Unit field that's an enum, for Fahrenheit and Celsius (and Kelvin maybe?)

internal/grpc/example/server/main.go
```diff
@@ -0,0 +1,171 @@
...
+	logger logger.Logger
+
+	// All gRPC services should embed the Unimplemented*Server structs to ensure forwards compatibility (if the service is
+	// compiled against a newer version of the proto file, the server will still have default implementations of any new
+	// RPCs).
+	pb.UnimplementedWeatherServiceServer
+}
+
+// GetCurrentWeather is a Unary RPC (single request, single response) that returns the current weather for the requested location.
+func (s *weatherServer) GetCurrentWeather(ctx context.Context, req *pb.LocationRequest) (*pb.WeatherResponse, error) {
```
eseliger, line 28: What's your take on keeping the proto local to the server implementation, and not polluting other API layers with it? 

Since the proto is the transport layer representation, I'm inclined to say it shouldn't bleed out of the gRPC server layer.

So for example:


```
 _____________ 
| gRPC Client |
 -------------
       |
       | &pbv1.Type{}
       \/
 _____________ 
| gRPC Server |
 -------------
       |
       | &service.Type{}
       \/
 _____________ 
|   Service   |
 -------------
       |
       | &database.Type{}
       \/
 _____________ 
|   DB Store  |
 -------------
```

Maybe we could showcase that a little bit as well, by letting the weatherServer call the weatherService, without any pb type involved. If you agree here!

internal/grpc/example/server/main.go
```diff
@@ -0,0 +1,171 @@
...
+			return status.FromContextError(ctx.Err()).Err()
+
+		case <-time.After(2 * time.Second): // Wait 2 seconds between the fake alerts we're sending.
+
+			// Send a message to the client
+			err := stream.Send(&pb.AlertResponse{
+				Alert: "Severe weather alert!",
+			})
+			if err != nil {
+				return err
```
eseliger, line 82: we explicitly don't need to wrap errors from Send in codes, correct? Should we document that?

internal/grpc/example/server/main.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+
+		data, err := stream.Recv()
+		if errors.Is(err, io.EOF) {
+			// io.EOF is a sentinel value that indicates that the client has explicitly closed its end of the stream, which signals the end of the RPC.
+
+			// We can use SendAndClose to send a final message to the client and close our end of the stream.
+			return stream.SendAndClose(&pb.UploadStatus{
+				Message: "Data received successfully",
+			})
```
eseliger, line 106: Is it legitimate to just call Close as well? In case there's not really a message other than "no error" we want to send.
If that's true, maybe we could add 
```suggestion
			return stream.SendAndClose(&pb.UploadStatus{
				Message: "Data received successfully",
			})
			// Or: return stream.Close() to simply close the connection without triggering an error on the client side.
```

internal/grpc/example/server/main.go
```diff
@@ -0,0 +1,171 @@
...
+		if errors.Is(err, io.EOF) { // The client has closed its end of the stream, so we can close our end as well.
+			return nil
+		}
+		if err != nil {
+			return err
+		}
+
+		description := "Cloudy"
+		location := locUpdate.GetLocation()
+		if location == "Philadelphia" {
```
eseliger, line 140: biased!

internal/grpc/example/client/main.go
```diff
@@ -0,0 +1,131 @@
...
+
+	client := pb.NewWeatherServiceClient(conn)
+
+	// Unary RPC: Normal case - get weather for a specific location
+
+	weather, err := client.GetCurrentWeather(context.Background(), &pb.LocationRequest{Location: "New York"})
+	if err != nil {
+		log.Fatalf("Could not get weather: %v", err)
+	}
+	log.Printf("Weather in NYC - description: %s, temp: %f", weather.GetDescription(), weather.GetTemperature())
```
eseliger, line 36: should we add a comment about using getters on the client side, too? Might be overlooked if it's just in the server.

internal/grpc/example/client/main.go
```diff
@@ -0,0 +1,131 @@
...
+
+	// Unary RPC: Error case - get weather for a specific location (that doesn't exist for didactic purposes)
+	weather, err = client.GetCurrentWeather(context.Background(), &pb.LocationRequest{Location: "Ravenholm"})
+	fmt.Println(err)
```
eseliger, line 40: is that a leftover?

internal/grpc/example/client/main.go
```diff
@@ -0,0 +1,131 @@
...
+		err := biStream.Send(&pb.LocationUpdate{
+			Location: "Location " + strconv.Itoa(i),
+		})
+		if err != nil {
+			log.Fatalf("Error while sending location update: %v", err)
+		}
+		time.Sleep(2 * time.Second)
+	}
+
+	biStream.CloseSend()
```
eseliger, line 130: Does this need error handling?

eseliger (approved):

internal/grpc/example/server/main.go
```diff
@@ -0,0 +1,257 @@
...
+	return stream.SendAndClose(&pb.UploadWeatherPhotoResponse{
+		Message: "Screenshot received successfully.",
+	})
+}
+
+// You can run the server implementation via:
+//
+// ```
+// cd internal/grpc/example/server
+// go run main.goSRC_DEVELOPMENT=true SRC_LOG_LEVEL=dbug go run .
```
eseliger, line 236: this comment looks wrong

internal/grpc/example/server/conversion.go
```diff
@@ -0,0 +1,188 @@
+// A good practice is to maintain a good separation between the gRPC-specific service implementation
+// (which only handles any gRCP specifics like error handling) and the "internal" service which handles the actual
+// business logic.
+//
+// This makes it easier to identify dependencies and test the business logic without having to worry about transport-specific
+// details.
+//
+// One strategy to achieve this is to define a "conversion" layer that converts between the gRPC types and the internal types (this file).
+// You can use fuzz tests to ensure that the conversion layer is correct even if the underlying types change (see conversion_test.go).
```
eseliger, line 9: thanks for this!

internal/grpc/example/client/main.go
```diff
@@ -0,0 +1,303 @@
...
+	"github.com/sourcegraph/sourcegraph/internal/grpc/streamio"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+//go:embed sun.png
+var sunDrawingImage []byte
+
+func main() {
```
eseliger, line 24: I wonder if we should've moved the different usage-examples into separate functions that we call from `main`, but we can leave that for next time.

internal/grpc/example/client/sun.png
eseliger, line 1: does this file need any attribution?

eseliger (approved):

internal/grpc/example/server/main.go
ggilmore: It's technically correct, but not needed anymore since I created the scripts.

internal/grpc/example/client/sun.png
ggilmore: Added

internal/grpc/example/server/main.go
eseliger: Really the missing space between main.go and the env var? What magic syntax is that 😬

ggilmore: 🤦🏾

camdencheek (approved):
Thank you for this! Looks like a great resource for getting up to speed.

internal/grpc/example/server/main.go
```diff
@@ -0,0 +1,252 @@
...
+	err := s.service.SubscribeWeatherAlerts(ctx, req.GetRegion(), callback)
+	if err != nil {
+		if ctx.Err() != nil {
+			// status.FromContextError is a convenience function that converts a context error
+			// to a Status object (with code codes.Canceled or codes.DeadlineExceeded).
+			//
+			// Returning a proper status (instead of a nil error) here makes it clearer to the caller / service-wide observability tools that look at
+			// response codes what exactly happened with this RPC call.
+			return status.FromContextError(ctx.Err()).Err()
+		}
```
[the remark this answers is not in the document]
ggilmore: @camdencheek 

I don't think we're always consistent in our main codebase to always propagate context errors back _up_ the call chain (hence the need for packages like https://github.com/sourcegraph/sourcegraph/pull/54092).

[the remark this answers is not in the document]
camdencheek: aah, got it. Sounds good then, feel free to ignore

PR #58792, opened 2023-12-06
SSC: Hook up exempted users with cody-pro FF

vdavid:
We have a few test users whom we want to make Cody PLG users automatically. This PR does that.

## Test plan

Tested this:
- ✅ Created a new user through OpenID Connect. If the user was on `MinimumExternalAccountAgeExemptList` then they got `cody-pro` as a feature flag.
- ✅ Otherwise, they didn't get the feature flag.
- ✅ If the `MinimumExternalAccountAgeExemptList` config item doesn't exist, it still works

### DIFF
--- a/cmd/frontend/internal/auth/oauth/session.go
+++ b/cmd/frontend/internal/auth/oauth/session.go
@@ -138,7 +138,7 @@ func SessionIssuer(logger log.Logger, db database.DB, s SessionIssuerHelper, ses
 		if newUserCreated {
 			dc := conf.Get().Dotcom
 			verifiedEmails, err := db.UserEmails().ListByUser(ctx, database.UserEmailsListOptions{UserID: user.ID, OnlyVerified: true})
-			if dc != nil && err != nil && dc.MinimumExternalAccountAge > 0 {
+			if dc != nil && err != nil {
 				exempted := false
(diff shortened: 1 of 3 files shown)


cmd/frontend/internal/auth/oauth/session.go
```diff
@@ -132,6 +134,33 @@ func SessionIssuer(logger log.Logger, db database.DB, s SessionIssuerHelper, ses
 			return
 		}
 
+		// If the user is new, check if they should be exempted from the minimum external account age
```
unknwon, line 137: This comment is not accurate, right? From what I understand, we are "abusing" the `MinimumExternalAccountAgeExemptList` as a way to automatically enable the feature flag, (which is fine... but) we need to say that instead 😂

cmd/frontend/internal/auth/oauth/session.go
```diff
@@ -123,7 +123,7 @@ func SessionIssuer(logger log.Logger, db database.DB, s SessionIssuerHelper, ses
 			return
 		}
 
-		// Since we obtained a valid user from the OAuth token, we consider the GitHub login successful at this point
+		// Since we obtained a valid user from the OAuth token, we consider the login successful at this point
```
vdavid, line 126: Stumbled upon this mistake

thenamankumar (approved):

Issue #57337, opened 2023-10-04
Initial zoekt indexing does not report errors and stays processing


eseliger:
I recently broke my indexer by setting CTAGS_COMMAND to something zoekt didn't like. All I get is 3/31 repos indexed in the status message, and on the repo page I get 

![image](https://github.com/sourcegraph/sourcegraph/assets/19534377/26a0d9f4-6b44-415d-93ff-746ded66ca9d)

Two things:

- It wasn't actually attempting to index when I looked at the page, which feels misleading
- It doesn't give any indication that there's a problem. For simpler debugging without reading server logs, it would be great to get some more information there

/cc @sourcegraph/search-platform

jtibshirani:
I opened #58793 to try to address this as much as possible:
* Don't show a loading spinner and falsely claim that a repo is indexing when it's not :)
* If an index is out of date and it's been a long time since the last successful run, show a warning. Unfortunately this doesn't address the case where the initial indexing job for a repo fails.

 Because of our indexing job architecture, it's really tricky know when an indexing attempt has failed. We could try to have Zoekt tell Sourcegraph about failed indexing attempts, but it would be messy and didn't seem worth the complexity.

PR #58793, opened 2023-12-06
Admin: avoid false impression that Zoekt is indexing

jtibshirani:
Previously, when an index is not up-to-date, we'd always claim that we're
actively indexing the repo. This is misleading, because indexing indexing could
be stuck or failing.

Now, we remove the loading spinner and just say the repo is "queued for
indexing". If the index is out-of-date and it's been more than 8 hours since
the last successful job, then we also show a warning tooltip.

Addresses #57337

## Test plan

Lots of manual testing.

### DIFF
--- a/client/web/src/repo/settings/RepoSettingsIndexPage.tsx
+++ b/client/web/src/repo/settings/RepoSettingsIndexPage.tsx
@@ -1,6 +1,6 @@
 import React from 'react'
 
-import { mdiCheckCircle } from '@mdi/js'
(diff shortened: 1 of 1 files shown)


jtibshirani:
Here's what it looks like now when we've fallen behind on indexing. (Note that I pushed a new commit to clarify this tooltip message.)

Screenshot 2023-12-05 at 3 20 03 PM

keegancsmith (approved):

jtibshirani:
Merging even though `buildkite/aspect-workflow-sourcegraph` is failing, as it's not critical and a bit flaky.

PR #58795, opened 2023-12-06
[Backport 5.2] Updating an code host config maintains the correct unrestricted…

pjlast:
Fixes an issue where the logic that determines whether or not a code host connection is unrestricted differs between the code host connection creation and update code.

## Test plan

Tests updated

cc @sourcegraph/release-guild


internal/database/external_services.go
```diff
@@ -1652,11 +1653,8 @@ WHERE EXISTS(
 	return v && exists, nil
 }
 
-// recalculateFields updates the value of the external service fields that are
-// calculated depending on the external service configuration, namely
-// `Unrestricted` and `HasWebhooks`.
-func (e *externalServiceStore) recalculateFields(es *types.ExternalService, rawConfig string) error {
-	es.Unrestricted = !envvar.SourcegraphDotComMode() && !gjson.Get(rawConfig, "authorization").Exists()
+func calcUnrestricted(config string) bool {
```
keegancsmith, line 1656: I know this is a backport, so treat this as a driveby followup. But the mutliple checks on dotcom make this a little tricky to read. Seems like you could do something like "if dotcom return true" and then the rest of the logic may also be easier to read.

keegancsmith (approved):

keegancsmith:
CI is not pleased.

internal/database/external_services.go
pjlast: Yeah gotcha, thanks! I'll make a follow-up!

pjlast:
@keegancsmith I see in the test I added I depended on a new field that wasn't backported. Updated the test, so 🤞

Issue #58572, opened 2023-11-24
Update Cody Web to use access tokens when talking to Gateway


vdavid:

vdavid:
Also: Cody Web is broken due to the recent rate-limiting changes passing access tokens to Cody Gateway. Only for users with feature flag enabled.

vdavid:
Also there is an unused GraphQL API for completions for which we need to update rate limits.
It doesn't have the access token in context, so maybe we can write a function to use at both places (Cody Web and this), using the internal access tokens.

vdavid:
Once this is fixed, respond [here](https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1701823375135379?thread_ts=1701802259.445399&cid=C04MSD3DP5L)

PR #58800, opened 2023-12-06
SSC: Fix Cody Web

vdavid:
- Closes https://github.com/sourcegraph/sourcegraph/issues/58572

Our plan to use internal access tokens worked!
I found no way to get the SHA256-hashed token value from the DB, so I created a new function to get it.
I tested the new function to make sure it works. 

## Test plan

Tested it in two ways:
1. With Cody Web locally: I could repro the problem before this change, and not after. 🎉 
2. In the API console with the query
```
query CodyGatewayAccess {
  completions(input: {messages: [{speaker:HUMAN, text:"hi\n\nAssistant:"}], temperature:1,maxTokensToSample: 2, topK: 1, topP: 1}) {}
}
```

Btw, it's an additional win that the API console works again!

### DIFF
--- a/internal/accesstoken/personal_access_token.go
+++ b/internal/accesstoken/personal_access_token.go
@@ -11,7 +11,7 @@ import (
 	"github.com/sourcegraph/sourcegraph/lib/errors"
 )
 
-// personalAccessTokenPrefix is the token prefix for Sourcegraph personal access tokens. Its purpose
+// PersonalAccessTokenPrefix is the token prefix for Sourcegraph personal access tokens. Its purpose
 // is to make it easier to identify that a given string (in a file, document, etc.) is a secret
 // Sourcegraph personal access token (vs. some arbitrary high-entropy hex-encoded value).
 const PersonalAccessTokenPrefix = "sgp_"
@@ -21,7 +21,7 @@ const InstanceIdentifierHmacKey = "instance_identifier_hmac_key" // Public as we
 
 var personalAccessTokenRegex = lazyregexp.New("^(?:sgp_|sgph_)?(?:[a-fA-F0-9]{16}_|local_)?([a-fA-F0-9]{40})$")
 
-// ParseAccessToken parses a personal access token to remove prefixes and extract the <token> that is stored in the database
+// ParsePersonalAccessToken parses a personal access token to remove prefixes and extract the <token> that is stored in the database
 // Personal access tokens can take several forms:
 //   - <token>
 //   - sgp_<token>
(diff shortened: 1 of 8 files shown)


internal/accesstoken/personal_access_token.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/sourcegraph/sourcegraph/lib/errors"
 )
 
-// personalAccessTokenPrefix is the token prefix for Sourcegraph personal access tokens. Its purpose
+// PersonalAccessTokenPrefix is the token prefix for Sourcegraph personal access tokens. Its purpose
```
vdavid, line 14: These two changes in this file are admittedly unrelated, but these two typos really bothered me for a long time

internal/completions/httpapi/handler.go
```diff
@@ -88,15 +91,30 @@ func newCompletionsHandler(
...
-				http.Error(w, "Error parsing auth header", http.StatusUnauthorized)
-				return
-			}
-			accessToken, err = accesstoken.GenerateDotcomUserGatewayAccessToken(apiToken)
-			if err != nil {
-				trace.Logger(ctx, logger).Info("Access token generation failed", log.String("API token", apiToken), log.Error(err))
-				http.Error(w, "Access token generation failed", http.StatusUnauthorized)
-				return
+				// No actor either, so we fail.
+				if r.Header.Get("Authorization") != "" || sgactor.FromContext(ctx) == nil {
```
thenamankumar, line 95: We should check if header exists even before calling `authz.ParseAuthorizationHeader`

vdavid: No, I've tested this, and it's not the case.
- If the header doesn't exist, `r.Header.Get("Authorization")` will be ""
- `authz.ParseAuthorizationHeader` will call `parseHTTPCredentials("")`
- `parseHTTPCredentials` does
  ```
   parts := strings.SplitN("", " ", 2)
   scheme = parts[0]
   if len(parts) == 1 { return scheme, "", nil, nil }
   ```
Which will return `"", "", nil, nil`
- `ParseAuthorizationHeader` gets back an unknown schema ("") and will throw an error
- we check for `if err != nil` (which is `true`) and then `r.Header.Get("Authorization") != ""` (which is `false`) so the behavior will be correct.

Also demoing this here in a [2-min Loom](https://www.loom.com/share/d7a511e0de78490dafbdaf58758aef7d)
But to avoid confusion, I've added a comment about this here: https://github.com/sourcegraph/sourcegraph/pull/58800/commits/7e6326090a94eacccbf305bc1e597e1bccbfe011

thenamankumar (approved):

PR #58807, opened 2023-12-06
svelte: Add proper page titles

fkling:
This commit updates pages to set a reasonable page title. They are similar to what we do in the current web app.

## Test plan

Manual testing.


client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/-/tags/+page.svelte
```diff
@@ -14,6 +14,10 @@
     $: total = $tags?.totalCount
 </script>
 
+<svelte:head>
```
Comment (vovakulikov) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/-/tags/+page.svelte:17]: We probably could extract this part into separate reusable component but maybe it's react system says in me.

Comment (fkling) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/-/tags/+page.svelte:17]: Maybe... it seemed simple enough to just do it this way.

Issue #23251, opened 2021-07-26
Scope @sourcegraph/code-host-integration styles to prevent conflict with native stylesheet


souldzin:
# Description

The `npm` package used for integrations [like in GitLab](https://gitlab.com/gitlab-org/gitlab/-/blob/6d4aa2d86c51e57e40ad5345fccc1b4d486f6b3b/package.json#L67) includes a `css/style.bundle.css`. This stylesheet is automatically injected into the HTML by the Sourcegraph `integration.bundle.js`.

It looks like all the selectors in [the Sourcegraph stylesheet](https://unpkg.com/browse/@sourcegraph/code-host-integration@0.0.59/css/style.bundle.css) are globally scoped and include some pretty generic names like `.file-header`. It's just a matter of time before a code host's stylesheet and this one clash.

# What's the worst case?

If the styles conflict, the Sourcegraph styles will take precedence which could break the code host's UI.

# Proposal

Maybe we can scope these integration styles under a specific element? Something like:

```patch
diff --git a/client/browser/src/app.scss b/client/browser/src/app.scss
index a466d5f8d0..68f5dd5142 100644
--- a/client/browser/src/app.scss
+++ b/client/browser/src/app.scss
@@ -1,3 +1,4 @@
+.sourcegraph-container {
 @import '../../branded/src/global-styles/colors';
 
 // Bootstrap configuration before Bootstrap is imported
@@ -59,3 +60,4 @@ $theme-colors-light: (
 @import './shared/code-hosts/shared/HoverOverlay.scss';
 @import './shared/code-hosts/shared/ViewOnSourcegraphButton';
 @import './shared/code-hosts/shared/nativeTooltips';
+}

```

souldzin:
@Joelkw and/or @felixfbecker, could you please check out this issue? It looks like we haven't run into this at GitLab, but it seems like it's just a matter of time 😅

What do you think?

j-erasmus:
@Joelkw @felixfbecker any update on this? We recently started using Highlight.js in GitLab for highlighting source code and it looks like the stylesheet mentioned in this issue causes a conflict with the default themes from Highlight.js ([issue](https://gitlab.com/gitlab-org/gitlab/-/issues/365167)).

| before | after(conflict) |
| ------ | ------ |
[...]

PR #58809, opened 2023-12-06
Use document element for class theme in code hosts integrations

vovakulikov:
Fixes https://github.com/sourcegraph/sourcegraph/issues/23251

I think this problem has been with us since the very beginning of native code integration and browser extension development. This may became even worse since we switched from Webpack to esbuild where we include all possible code host styles in one bundle (meaning that we still load GitHub-related styles even if we're in GitLab integration). 

Overall, the current state is 
- In this PR we use just a hot fix way and use the most top level HTML element to assign css tokens and colors (this gives us ability to not native styles and keep cascade specificity lower than native styles)
- Possible we could use here `@layer` cascade at-rule, but at the moment our CSS token system structure just ins't ready for this change and this could lead to some regressions in the main application. 
- Ideally, we should assign CSS tokens only on nodes where we actually render something Sourcegraph specific, like code-intel tooltip or "open in Sourcegraph" button. But again, this would require to change our global css tokens in our wildcard, so we skip this option for now but should take this option as a main one if we would return to our browser extension development in future.
- We still do have `.hljs` css global classes in the bundle, but since Gitlab increase cascade specificity on their side this is kind of fixed but this is where we could use `@layers` approach IMO)

## Test plan
- Check gitlab with enabled browser extension in different themes, check that code editor and general UI has native gitlab styles.

## Note after merging this
- Don't forget to run publish script as soon as this change is merged.

### DIFF
--- a/client/browser/src/shared/code-hosts/shared/codeHost.tsx
+++ b/client/browser/src/shared/code-hosts/shared/codeHost.tsx
@@ -727,8 +727,8 @@ export async function handleCodeHost({
     // Handle theming
     subscriptions.add(
         (codeHost.isLightTheme ?? of(true)).subscribe(isLightTheme => {
-            document.body.classList.toggle('theme-light', isLightTheme)
-            document.body.classList.toggle('theme-dark', !isLightTheme)
+            document.documentElement.classList.toggle('theme-light', isLightTheme)
+            document.documentElement.classList.toggle('theme-dark', !isLightTheme)
         })
     )
 


souldzin:
Thanks for purring this iterative fix together @vovakulikov!

fkling (approved):

taras-yemets (approved):
Thanks for the fix and the detailed description!

andr3:
+1 Thanks for fixing this @vovakulikov! 🙏

PR #58813, opened 2023-12-06
oidc: store access token into auth_data

unknwon:
We haven't been storing token info into the `user_external_accounts.auth_data` column, which might be an oversight. We need to store the token info for SAMS (and Cody Gateway later on), so need to fix it now.

## Test plan

CI and tested e2e (local credentials):

CleanShot 2023-12-06 at 17 52 21@2x

### DIFF
--- a/cmd/frontend/internal/auth/openidconnect/user.go
+++ b/cmd/frontend/internal/auth/openidconnect/user.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 
 	"github.com/coreos/go-oidc"
+	"golang.org/x/oauth2"
 
 	"github.com/sourcegraph/sourcegraph/cmd/frontend/auth"
 	"github.com/sourcegraph/sourcegraph/cmd/frontend/hubspot"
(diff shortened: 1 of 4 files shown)


cmd/frontend/internal/auth/openidconnect/user.go
```diff
@@ -26,7 +27,7 @@ type ExternalAccountData struct {
 // getOrCreateUser gets or creates a user account based on the OpenID Connect token. It returns the
 // authenticated actor if successful; otherwise it returns a friendly error message (safeErrMsg)
 // that is safe to display to users, and a non-nil err with lower-level error details.
-func getOrCreateUser(ctx context.Context, db database.DB, p *Provider, idToken *oidc.IDToken, userInfo *oidc.UserInfo, claims *userClaims, usernamePrefix, anonymousUserID, firstSourceURL, lastSourceURL string) (newUserCreated bool, _ *actor.Actor, safeErrMsg string, err error) {
+func getOrCreateUser(ctx context.Context, db database.DB, p *Provider, token *oauth2.Token, idToken *oidc.IDToken, userInfo *oidc.UserInfo, claims *userClaims, usernamePrefix, anonymousUserID, firstSourceURL, lastSourceURL string) (newUserCreated bool, _ *actor.Actor, safeErrMsg string, err error) {
```
bobheadxi, line 30: that's a _lot_ of args and return values 😆

bobheadxi (approved):
nice :)

bobheadxi:
Should this get a security review as well?

unknwon:
Oh dang, auto-merged 😂 requested review!

cmd/frontend/internal/auth/openidconnect/user.go
unknwon: hahahaha!

Issue #56876, opened 2023-09-20
msp: monitoring module POC


bobheadxi:
Monitoring is pretty important and we currently have none of that. We should have similar provisioning of notification channels, etc as Cody Gateway, and maybe even automate the clickops steps here: https://handbook.sourcegraph.com/departments/engineering/teams/cody/cody-gateway/#deployment

We should have a set of default alerts - at the very least, achieve parity with Cody Gateway's more generic alerts - as well as the ability to configure custom ones as a stretch goal.

It would also be very nice to be able to automatically provision various Sentry resources for error reporting

PR #58816, opened 2023-12-07
msp: add monitoring stack

jac:
Add monitoring stack to MSP

Includes alerts based on those defined in cody gateway [alerts_cloudrun.tf](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/infrastructure/-/blob/cody-gateway/modules/monitoring/alerts_cloudrun.tf)

Closes https://github.com/sourcegraph/sourcegraph/issues/56876

## Test plan

- Deployed to a msp-testbed GCP project observed alert policies in place
- eyeballed the cdktf matches alerts in [alerts_cloudrun.tf](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/infrastructure/-/blob/cody-gateway/modules/monitoring/alerts_cloudrun.tf)

### DIFF
--- a/dev/managedservicesplatform/internal/resource/monitoringalertpolicy/monitoringalertpolicy.go
+++ b/dev/managedservicesplatform/internal/resource/monitoringalertpolicy/monitoringalertpolicy.go
@@ -211,8 +211,8 @@ func buildFilter(config *Config) string {
 		filters = append(filters, fmt.Sprintf(`%s = "%s"`, key, val))
 	}
 
-	// Sort to ensure stable output for testing
-	// This code runs so infreuqently that sort performance is not a concern and there are only a couple elements
+	// Sort to ensure stable output for testing, because
+	// config.ThresholdAggregation.Filters is a map.
 	sort.Strings(filters)
 
 	if config.ServiceKind == spec.ServiceKindService {
--- a/dev/managedservicesplatform/internal/resource/monitoringalertpolicy/monitoringalertpolicy.go
+++ b/dev/managedservicesplatform/internal/resource/monitoringalertpolicy/monitoringalertpolicy.go
@@ -82,7 +82,7 @@ type ThresholdAggregation struct {
 	Duration     string
 }
 
-// ResponseCodeMetric for alerting when the numer of a certain response code exceeds a threshold
+// ResponseCodeMetric for alerting when the number of a certain response code exceeds a threshold
 //
 // Must specify either `Code` (e.g. 404) or `CodeClass` (e.g. 4xx)
 //
--- a/dev/managedservicesplatform/internal/resource/monitoringalertpolicy/monitoringalertpolicy.go
+++ b/dev/managedservicesplatform/internal/resource/monitoringalertpolicy/monitoringalertpolicy.go
@@ -274,11 +274,12 @@ func responseCodeMetric(scope constructs.Construct, id resourceid.ID, config *Co
 func responseCodeBuilder(config *Config) string {
 	var builder strings.Builder
 
-	builder.WriteString("fetch cloud_run_revision\n")
-	builder.WriteString("| metric 'run.googleapis.com/request_count'\n")
-	builder.WriteString("| group_by 15s, [value_request_count_aggregate: aggregate(value.request_count)]\n")
-	builder.WriteString("| every 15s\n")
-	builder.WriteString("| {\n")
+	builder.WriteString(`fetch cloud_run_revision
+| metric 'run.googleapis.com/request_count'
+| group_by 15s, [value_request_count_aggregate: aggregate(value.request_count)]
+| every 15s
+| {
+`)
 	if config.ResponseCodeMetric.CodeClass != nil {
 		builder.WriteString("  group_by [metric.response_code, metric.response_code_class],\n")
 	} else {
@@ -295,11 +296,12 @@ func responseCodeBuilder(config *Config) string {
 			builder.WriteString(fmt.Sprintf("  | filter (metric.response_code != '%s')\n", code))
 		}
 	}
-	builder.WriteString("; group_by [],\n")
-	builder.WriteString("  [value_request_count_aggregate_aggregate: aggregate(value_request_count_aggregate)]\n")
-	builder.WriteString("}\n")
-	builder.WriteString("| join\n")
-	builder.WriteString("| value [response_code_ratio: val(0) / val(1)]\n")
+	builder.WriteString(`; group_by [],
+  [value_request_count_aggregate_aggregate: aggregate(value_request_count_aggregate)]
+}
+| join
+| value [response_code_ratio: val(0) / val(1)]
+`)
 	builder.WriteString(fmt.Sprintf("| condition gt(val(), %s)\n", strconv.FormatFloat(config.ResponseCodeMetric.Ratio, 'f', -1, 64)))
 	return builder.String()
 }
--- a/dev/managedservicesplatform/internal/resource/monitoringalertpolicy/monitoringalertpolicy.go
+++ b/dev/managedservicesplatform/internal/resource/monitoringalertpolicy/monitoringalertpolicy.go
@@ -12,7 +12,6 @@ import (
 	"github.com/sourcegraph/managed-services-platform-cdktf/gen/google/monitoringalertpolicy"
 
 	"github.com/sourcegraph/sourcegraph/dev/managedservicesplatform/internal/resourceid"
-	"github.com/sourcegraph/sourcegraph/dev/managedservicesplatform/spec"
 	"github.com/sourcegraph/sourcegraph/lib/pointers"
 )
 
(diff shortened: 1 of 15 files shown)


dev/managedservicesplatform/internal/resource/monitoringalertpolicy/monitoringalertpolicy.go
```diff
@@ -0,0 +1,299 @@
...
+			},
+			AlertStrategy: &monitoringalertpolicy.MonitoringAlertPolicyAlertStrategy{
+				AutoClose: pointers.Ptr("604800s"),
+			},
+		})
+	return &Output{}, nil
+}
+
+// filter creates the Filter string for a ThresholdAggregation monitoring alert policy
+func filter(config *Config) string {
```
bobheadxi, line 208: nit - we use `filters` as a variable name, let's make this a little more explicit:

```suggestion
// buildFilter creates the Filter string for a ThresholdAggregation monitoring alert policy
func buildFilter(config *Config) string {
```

dev/managedservicesplatform/internal/resource/monitoringalertpolicy/monitoringalertpolicy.go
```diff
@@ -0,0 +1,299 @@
...
+		filters = append(filters, fmt.Sprintf(`%s = "%s"`, key, val))
+	}
+
+	// Sort to ensure stable output for testing
+	// This code runs so infreuqently that sort performance is not a concern and there are only a couple elements
+	sort.Strings(filters)
+
+	if config.ServiceKind == spec.ServiceKindService {
+		filters = append(filters, `resource.type = "cloud_run_revision"`)
+		filters = append(filters, fmt.Sprintf(`resource.labels.service_name = "%s"`, config.ServiceName))
```
bobheadxi, line 220: nit - might be a bit easier to read all in one, given we add these together:

```suggestion
		filters = append(filters,
			`resource.type = "cloud_run_revision"`)
			fmt.Sprintf(`resource.labels.service_name = "%s"`, config.ServiceName))
```

same for the next block :)

dev/managedservicesplatform/internal/resource/monitoringalertpolicy/monitoringalertpolicy.go
```diff
@@ -0,0 +1,299 @@
...
+// filter creates the Filter string for a ThresholdAggregation monitoring alert policy
+func filter(config *Config) string {
+	filters := make([]string, 0)
+	for key, val := range config.ThresholdAggregation.Filters {
+		filters = append(filters, fmt.Sprintf(`%s = "%s"`, key, val))
+	}
+
+	// Sort to ensure stable output for testing
+	// This code runs so infreuqently that sort performance is not a concern and there are only a couple elements
+	sort.Strings(filters)
```
bobheadxi, line 216: Might want to clarify we d o this specifically for `onfig.ThresholdAggregation.Filters` (I was wondering why we don't do it later, after the ServiceKind checks):

```suggestion
	// Sort to ensure stable output for testing, because
	// config.ThresholdAggregation.Filters is a map.
	sort.Strings(filters)
```

Also, in general feel free to not justify performance things if the ergonomics are obviously better (stable output), in most cases there is nothing to worry about. If anything I think the inverse is expected: there should be justification for non-ergonomic code in the name of performance :)

dev/managedservicesplatform/internal/resource/monitoringalertpolicy/monitoringalertpolicy.go
```diff
@@ -0,0 +1,299 @@
...
+
+// threshholdAggregation defines a monitoring alert policy based on a single metric threshold
+func thresholdAggregation(scope constructs.Construct, id resourceid.ID, config *Config) (*Output, error) {
+	// Set some defaults
+	if config.ServiceKind == spec.ServiceKindService {
+		// For Cloud Run Services we need to group by the revision_name
+		config.ThresholdAggregation.GroupByField = "resource.label.revision_name"
+	} else if config.ServiceKind == spec.ServiceKindJob {
+		// No default for this
+	} else {
```
bobheadxi, line 151: Surprised by this but poke around and indeed, doesn't seem like we get revisions as labels for jobs :o

bobheadxi: Might want to update docstring to indicate as such

dev/managedservicesplatform/internal/stack/monitoring/monitoring.go
```diff
@@ -0,0 +1,228 @@
...
+func NewStack(stacks *stack.Set, vars Variables) (*CrossStackOutput, error) {
+	stack, _, err := stacks.New(StackName, googleprovider.With(vars.ProjectID))
+	if err != nil {
+		return nil, err
+	}
+
+	id := resourceid.New("monitoring")
+	err = commonAlerts(stack, id, vars)
+	if err != nil {
+		return nil, err
```
bobheadxi, line 72: nit - do an error wrap somewhere, either here (e.g. `errors.Wrap(err, "common alerts")`), within `commonAlerts`, or even both. You'll thank yourself later, since even 1 wrap will give you a lot of context if anything comes up

dev/managedservicesplatform/managedservicesplatform.go
```diff
@@ -119,6 +121,20 @@ func (r *Renderer) RenderEnvironment(
 		return nil, errors.Wrap(err, "failed to create cloudrun stack")
 	}
 
+	// We only create the instance_count alert when MaxCount > 5 so default to 0 if unspecified
+	maxCount := 0
+	if env.Instances.Scaling != nil {
+		maxCount = pointers.Deref(env.Instances.Scaling.MaxCount, 0)
+	}
```
bobheadxi, line 128: nit - let's make `MaxCount` an `*int`, and just handle this at the callsite

dev/managedservicesplatform/managedservicesplatform.go
```diff
@@ -53,6 +54,7 @@ func (r *Renderer) RenderEnvironment(
 	svc spec.ServiceSpec,
 	build spec.BuildSpec,
 	env spec.EnvironmentSpec,
+	mntrng spec.MonitoringSpec,
```
bobheadxi, line 57: classic Go conundrum, sometimes a word is just so good it becomes the package, the var, and we're left with hard names 😆 

IMO when there's a conflict and an abbreviation isn't obvious, it might be better to just use a longer name like `monitoringSpec`

bobheadxi (commented):
Initial review 😁

dev/managedservicesplatform/managedservicesplatform.go
jac: Ah yeah good catch. I meant to go back and change it because I hated what I named it

dev/managedservicesplatform/spec/monitoring.go
```diff
@@ -0,0 +1,86 @@
...
+	if r.CodeClass != nil {
+		if !codeClassPattern.MatchString(*r.CodeClass) {
+			errs = append(errs, errors.New("responseCodeRatios[].codeClass must match the format NXX (e.g. 4xx, 5xx)"))
+		}
+	}
+
+	if r.Duration != nil {
+		duration, err := time.ParseDuration(*r.Duration)
+		if err != nil {
+			errs = append(errs, errors.New("responseCodeRatios[].duration must be in the format of XXs"))
```
bobheadxi, line 79:
```suggestion
			errs = append(errs, errors.Wrap(err, "responseCodeRatios[].duration must be in the format of XXs"))
```

Always useful to include the original error, if we have one 😁

dev/managedservicesplatform/internal/stack/monitoring/monitoring.go
```diff
@@ -0,0 +1,228 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func responseCodeMetrics(stack cdktf.TerraformStack, id resourceid.ID, vars Variables) error {
+	for _, config := range vars.Monitoring.Alerts.ResponseCodeRatios {
+
+		_, err := monitoringalertpolicy.New(stack, id, &monitoringalertpolicy.Config{
```
bobheadxi, line 208: Every resource needs a unique ID - `monitoringalertpolicy` happens to only create a single resource that we give a unique sub-ID but I think it's good still to add an ID on top, e.g.

```suggestion
		_, err := monitoringalertpolicy.New(stack, id.Group(...), &monitoringalertpolicy.Config{
```

to ensure it's scoped

dev/managedservicesplatform/internal/resource/monitoringalertpolicy/monitoringalertpolicy.go
```diff
@@ -0,0 +1,305 @@
...
+		builder.WriteString(fmt.Sprintf("  | filter (metric.response_code_class = '%s')\n", *config.ResponseCodeMetric.CodeClass))
+	}
+	if config.ResponseCodeMetric.ExcludeCodes != nil && len(config.ResponseCodeMetric.ExcludeCodes) > 0 {
+		for _, code := range config.ResponseCodeMetric.ExcludeCodes {
+			builder.WriteString(fmt.Sprintf("  | filter (metric.response_code != '%s')\n", code))
+		}
+	}
+	builder.WriteString("; group_by [],\n")
+	builder.WriteString("  [value_request_count_aggregate_aggregate: aggregate(value_request_count_aggregate)]\n")
+	builder.WriteString("}\n")
```
bobheadxi, line 300: a chain like this might be easier to read in a single block with backtick string literals, instead of being written over multiple calls

bobheadxi (approved):
Only nits, overall LGTM 😁 Sorry about the delay on the review, I forgot to check back!

PR #58820, opened 2023-12-07
Fix top_p type for Fireworks

rafax:
Top_p is a number (i.e. a float) in Fireworks API spec, but we used int32 to represent it. This didn't throw errors because we converted it here, but resulted in all requests using a value of 0 (instead of what was passed).
## Test plan

- tested locally
- tests should pass


philipp-spiess (approved):

philipp-spiess:
Removing my approval as I’m checking wether changing `top_p` causes a perf regression

philipp-spiess (approved):
After some API requests to fireworks, it seems that this doesn't make much of a difference if `temperature` is also set (which we do). Let's clean this up. We know how to revert in case of issues. 👍

unknwon (approved):

PR #58822, opened 2023-12-07
Rg/fix response flushing

rafax:
Implement response flushing in StatusHeaderRecorder - until now, we embedded a ResponseWriter, but didn't implement Flusher, so our own code or [stdlib](https://tip.golang.org/doc/go1.20#http_responsecontroller) couldn't flush responses.
## Test plan

- unit tests
- tested locally


eseliger (approved):

bobheadxi:
Thanks for the fix!! This was a major TIL

PR #58824, opened 2023-12-07
sg: add `doctor` command to generate a diagnostics report

burmudar:
`sg doctor` runs the diagnostics defined in `sg-doctor.yaml`:
```yaml
diagnostics:
  general:
    - name: current env
      cmd: env
    - name: current path
      cmd: pwd
    - name: current branch
      cmd: git rev-parse --abbrev-ref HEAD
    - cmd: show current commit
      name: git rev-parse HEAD
    - name: os kernel
      cmd: uname -s
  bazel:
    - name: architecture
      cmd: file -L $(which bazel)
    - name: where is bazel
      cmd: which bazel
    - name: bazel info
      cmd: bazel info
  go:
    - name: architecture
      cmd: file -L $(which go)
    - name: where is go
      cmd: which go
    - name: version of go
      cmd: go version
  node:
    - name: architecture
      cmd: file -L $(which node)
    - name: where is node
      cmd: which node
    - name: version of node
      cmd: node --version
    - name: version of pnpm
      cmd: pnpm --version
    - name: where is pnpm
      cmd: which pnpm

```

which produces the following report [sg-doctor-report-2023-12-07-110158.md](https://github.com/sourcegraph/sourcegraph/files/13599892/sg-doctor-report-2023-12-07-110158.md)

I'm not quite convinced at the console output yet so still working on that:
```
go run ./dev/sg doctor
🥼 Gathering diagnostics
💊 Running general diagnostics

✅ Diagnostic "current env" complete
✅ Diagnostic "current path" complete
✅ Diagnostic "current branch" complete
✅ Diagnostic "git rev-parse HEAD" complete
✅ Diagnostic "os kernel" complete

💊 Running bazel diagnostics

✅ Diagnostic "architecture" complete
✅ Diagnostic "where is bazel" complete
✅ Diagnostic "bazel info" complete

💊 Running go diagnostics
✅ Diagnostic "architecture" complete
✅ Diagnostic "where is go" complete
✅ Diagnostic "version of go" complete

💊 Running node diagnostics
✅ Diagnostic "architecture" complete
✅ Diagnostic "where is node" complete
✅ Diagnostic "version of node" complete
✅ Diagnostic "version of pnpm" complete
✅ Diagnostic "where is pnpm" complete

💉 Gathering of diagnostics complete!
📋 Diagnostic report written to sg-doctor-report-2023-12-07-110158.md
```
## Test plan
Tested locally

### DIFF
--- a/dev/sg/sg_doctor.go
+++ b/dev/sg/sg_doctor.go
@@ -1,13 +1,23 @@
 package main
 
 import (
+	"context"
+	"fmt"
+	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/urfave/cli/v2"
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
+	"gopkg.in/yaml.v3"
 
 	"github.com/sourcegraph/sourcegraph/dev/sg/internal/category"
+	"github.com/sourcegraph/sourcegraph/dev/sg/internal/run"
+	"github.com/sourcegraph/sourcegraph/dev/sg/internal/std"
 	"github.com/sourcegraph/sourcegraph/dev/sg/root"
 	"github.com/sourcegraph/sourcegraph/lib/errors"
+	"github.com/sourcegraph/sourcegraph/lib/output"
 )
 
 var doctorCommand = &cli.Command{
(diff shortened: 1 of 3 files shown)


burmudar:
> My two cents on the console output: might it be useful to also include the command results in the output? I know if I was ever in a situation where I was running this, it'd likely be because I wanted the answer to one or more of these questions.

I've reworked it now ... by default it will render the markdown in the termianl so you'll see the command it ran, the output and the error (if any). Below is an example

Screenshot 2023-12-11 at 15 22 34

dev/sg/sg_doctor.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/sourcegraph/sourcegraph/dev/sg/internal/run"
+	"github.com/sourcegraph/sourcegraph/dev/sg/internal/std"
+	"github.com/sourcegraph/sourcegraph/dev/sg/root"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+	"github.com/sourcegraph/sourcegraph/lib/output"
+)
+
+var doctorCommand = &cli.Command{
+	Name:  "doctor",
+	Usage: "performs various diagnostics and generates a report",
```
jhchabran, line 26:
```suggestion
	Usage: "performs diagnostics of the local environment and prints out a report",
```

dev/sg/sg_doctor.go
```diff
@@ -0,0 +1,182 @@
...
+	further diagnosis.
+	`,
+	Category: category.Util,
+	Flags: []cli.Flag{
+		&cli.PathFlag{
+			Name:    "outputFile",
+			Aliases: []string{"o"},
+			Usage:   "write the report to a file with this name",
+		},
+		&cli.BoolFlag{
```
burmudar, line 39: not used - to be removed

dev/sg/sg_doctor.go
```diff
@@ -0,0 +1,182 @@
...
+
+	The output of the commands are stored in a report, which can then be given to a dev-infra team memeber for
+	further diagnosis.
+	`,
+	Category: category.Util,
+	Flags: []cli.Flag{
+		&cli.PathFlag{
+			Name:    "outputFile",
+			Aliases: []string{"o"},
+			Usage:   "write the report to a file with this name",
```
jhchabran, line 37: not sure why we'd need this

dev/sg/sg_doctor.go
```diff
@@ -0,0 +1,182 @@
...
+		r[group] = append(v, result)
+	}
+}
+
+func runDoctorDiagnostics(cmd *cli.Context) error {
+	repoRoot, err := root.RepositoryRoot()
+	if err != nil {
+		return err
+	}
+	diagnosticsPath := filepath.Join(repoRoot, "sg-doctor.yaml")
```
jhchabran, line 83: I feel like the `-` in there is a bit unusual, why not `sg.doctor.yaml` instead to match more the existing convention?

dev/sg/sg_doctor.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+func (d *diagnosticRunner) Run(ctx context.Context) DiagnosticReport {
+	env := os.Environ()
+	report := make(DiagnosticReport)
+
+	for group, diagnostics := range d.diagnostics.Diagnostic {
+		d.reporter.WriteLine(output.Emojif("💊", "Running %s diagnostics", group))
+		for _, diagnostic := range diagnostics {
+			out, err := run.BashInRoot(ctx, diagnostic.Cmd, env)
```
jhchabran, line 132: We might want to be careful here, if the user is not running bash, but let's say fish.

dev/sg/sg_doctor.go
```diff
@@ -0,0 +1,182 @@
...
+			outputSection := fmt.Sprintf("Output: \n```\n%s\n```\n", item.Output)
+			errSection := fmt.Sprintf("Error: \n```\n%v\n```\n", item.Err)
+			fmt.Fprintf(&sb, "### %s\n\n%s\n\n%s\n%s", titleCaser.String(item.Diagnostic.Name), cmdLine, outputSection, errSection)
+		}
+	}
+
+	return sb.String()
+}
+
+func loadDiagnostics(path string) (*Diagnostics, error) {
```
jhchabran, line 167:
```suggestion
func readDiagnosticDefinitions(path string) (*Diagnostics, error) {
```

dev/sg/sg_doctor.go
burmudar: Agreed - moved. I was inbetween minds whether we would need this

dev/sg/sg_doctor.go
burmudar: Yeah. I'll have to look since some commands I think are bash builtins 🤔

burmudar: Also it makes a bit tricky since you might think "I'll split and extract the main program and just pass the args - letting go handle finding the program executable", but args can have spaces in them so splitting accurate becomes a problem.

A better approach might be to:
(a): Detect default shell
(b): For the top 3 shells execute `${SHELL} -c` - I'll have to see if `fish` supports this as well.

I'll do the above in follow up work

jhchabran: Let's merge it as it is, see if it helps us dealing with the support and then we'll see WDYT?

jhchabran (approved):

dev/sg/sg_doctor.go
burmudar: Agreed.

PR #58825, opened 2023-12-07
nix: empty out before writing to it .bazelrc-nix

keegancsmith:
Currently on non-nixos hosts we will append to .bazelrc-nix everytime the shell hook runs. This leads to the file growing quite large, I am unsure on the impact on bazel. This was an accidental regression when we removed bazel from the nix darwin environment.

Test Plan: ran nix develop and I didn't have duplicated lines.


burmudar (approved):
Ah I encountered this yesterday!

Issue #58624, opened 2023-11-28
(P2) Revise Cody permissions system to use RBAC instead of feature-flag and inclusion-list only


slimsag:
Today Cody permissions system allows for enabling Cody only for specific users ([docs](https://docs.sourcegraph.com/cody/overview/enable-cody-enterprise#enable-cody-only-for-some-users)) but we want to be able to express the inverse (enable Cody excluding specific users) too.

PR #58831, opened 2023-12-07
cody: replace access-control feature flag with RBAC

### LINKED ISSUE #58623: Revise Cody permissions system to not be a ‘feature flag’ docs but rather site config or user management page
Today Cody's permissions controls are managed via a feature flag: https://docs.sourcegraph.com/cody/overview/enable-cody-enterprise#enable-cody-only-for-some-users

This should be a site config or user management page instead.

### LINKED ISSUE #58627: **Investigate** feasibility of non-explicit-list user access controls
**Investigate** the feasibility of non-explicit-list access controls; i.e. instead of admins explicitly listing usernames to allow/deny access to Cody, derive this information from IdP, Okta, RBAC, cost host user groups, or other sources of user group management.

Note: the goal of this issue is to figure out what we could implement in a timely manor, which of these would overlap well with customers, and to have more in-depth conversations about this feature with product (Ryan Phillips) & customers (ideally we find a single customer to build this for first) to identify which reasonable solutions we can implement.

See https://github.com/sourcegraph/sourcegraph/issues/58628 for the tracking issue for **implementing** this behavior, if we decide the juice is worth the squeeze.

slimsag:
This change replaces the old feature-flag which controls users who have access to Cody, with an RBAC-based implementation instead.

After this change, `cody.restrictUsersFeatureFlag` is deprecated. For Sourcegraph instances that still have it set, it will still be respected it until the site admin removes it from their config. A warning is displayed to let admins know this needs to be done:

image

Once `cody.restrictUsersFeatureFlag` is not set in the site config, the new RBAC system is used instead: site admins can manage user roles, and assign/remove `cody:access` to users or groups of users via the [access control](https://docs.sourcegraph.com/admin/access_control) web UI.

By default, all users have access (admins can change this default via the access control web UI.)

The new Cody RBAC system defaults to _enabled_ unless `cody.restrictUsersFeatureFlag` is set in the site config. It can be disabled by setting the new `cody.permissions` site config option to `false`, in which case Sourcegraph behaves as it did before this change.

Fixes #58623
Helps (`fixes?`) #58624
Resolves #58627

## Remaining TODOs before this PR can be merged

* [x] Some Go tests need updating / CI needs to pass
* [x] [These docs](https://docs.sourcegraph.com/cody/overview/enable-cody-enterprise#enable-cody-only-for-some-users) need to be updated to tell site admins on new versions what to do, tell site admins on old versions how they can use the old docs, and tell site admins upgrading what to do (remove the old config, and then use the RBAC system going forward.) with the exact Sourcegraph versions mentioned.

## Test plan

* Created a non-admin user, signed into VS Code' Cody via that user, confirmed I have access by default.
* Using the site admin account, change normal users'


cmd/frontend/graphqlbackend/site.go
```diff
@@ -627,7 +627,7 @@ func (r *siteResolver) RequiresVerifiedEmailForCody(ctx context.Context) bool {
 	return !isAdmin
 }
 
-func (r *siteResolver) IsCodyEnabled(ctx context.Context) bool { return cody.IsCodyEnabled(ctx) }
+func (r *siteResolver) IsCodyEnabled(ctx context.Context) bool { return cody.IsCodyEnabled(ctx, r.db) }
```
chwarwick, line 630: Looks like it's always been this way but feels strange to me that the site resolver is returning if Cody is enabled for the current user, not is Cody enabled for the site.

cmd/frontend/internal/completions/resolvers/resolver.go
```diff
@@ -33,7 +33,7 @@ func NewCompletionsResolver(db database.DB, logger log.Logger) graphqlbackend.Co
 }
 
 func (c *completionsResolver) Completions(ctx context.Context, args graphqlbackend.CompletionsArgs) (_ string, err error) {
-	if isEnabled := cody.IsCodyEnabled(ctx); !isEnabled {
+	if isEnabled := cody.IsCodyEnabled(ctx, c.db); !isEnabled {
 		return "", errors.New("cody experimental feature flag is not enabled for current user")
```
chwarwick, line 37: Clean up this error since I believe it will be displayed to the user.  
```suggestion
		return "", errors.New("cody is not enabled for current user")
```

cmd/frontend/graphqlbackend/site.go
slimsag: Yeah, it is very strange. It's also weird that we have Cody access control enforced in the feature-flag portion of `internal/cody`.

I haven't set out to 'correct' this in this PR with the hopes of keeping it from increasing in scope/risk

slimsag:
> If i'm following correctly the RBAC permissions are only used if the config value is set to enable permissions, which means the default case is that all users have access (same as today 👍 )
>
> But what happens in the RBAC UI? Does it look like you are configuring Cody access for groups/all users but because it's not enabled in the config it isn't considered?

That is correct; however if the RBAC permissions are not being respected then the warning would display at the top of the page indicating you're using the old permission system:

image

The one edge-case here would be if a site admin _explicitly disables the new RBAC permissions model_ (`cody.permissions` set to `false`). Eventually I would like to remove the ability for a site admin to do that at all (in +1 release, so many months have passed and everyone has had a chance to upgrade.)

slimsag:
@chwarwick or @eseliger would either of you be up for a quick review? I'd love to land this soon as it's blocking some other work :)

doc/cody/overview/enable-cody-enterprise.md
```diff
@@ -203,7 +203,36 @@ To turn Cody off:
 
 ## Enable Cody only for some users
 
-To enable Cody only for some users, for example, when rolling out a Cody POC, follow all the steps mentioned in [Enabling Cody on your Sourcegraph instance](#enable-cody-on-your-sourcegraph-instance). Then, use the feature flag `cody` to turn Cody on selectively for some users. To do so:
+To enable Cody only for some users, for example, when rolling out a Cody POC, follow all the steps mentioned in [Enabling Cody on your Sourcegraph instance](#enable-cody-on-your-sourcegraph-instance). Then, do the following:
```
chwarwick, line 206: Docs changes need to go into the new docs repo.  I think having them here will fail the build?

internal/cody/feature_flag_test.go
```diff
@@ -129,22 +172,134 @@ func TestIsCodyEnabled(t *testing.T) {
...
+		})
+		ctx := context.Background()
+		ctx = actor.WithActor(ctx, &actor.Actor{UID: 1})
+		db := mockDB(defaultUserPerms)
+		if !IsCodyEnabled(ctx, db) {
+			t.Error("Expected IsCodyEnabled to return true for authenticated actor")
+		}
+	})
+
+	t.Run("RBAC, Enabled cody, but not completions", func(t *testing.T) {
```
chwarwick, line 234: What does but not completions mean? This test looks the same as the one above.

chwarwick (approved):
Nice, this looks correct to me. See note about new docs process.

eseliger (approved):
LGTM, might warrant a changelog entry to make people aware prior to the upgrade that they need to migrate their access settings :)

internal/cody/feature_flag_test.go
slimsag: Good catch, I mirror'd the existing good test suite but for the case of RBAC - and this ended up being redundant. Removed.

PR #58840, opened 2023-12-08
Expanding logging to siteadmin actions

mohammadualam:
To expand our audit logging capabilities, we have to add more actions, especially siteadmin ones within our even logging setup.

The actions that we have identified to be of particular importance are being logged with this change so that they can be collected for analysis

## Test plan

Tested locally and currently running on k8s


evict:
Can we create a helper in our security events code like they did here: https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/cmd/frontend/internal/app/sign_out.go?L81:73-81:90?

internal/database/security_event_logs.go
```diff
@@ -216,3 +249,18 @@ func (s *securityEventLogsStore) LogEventList(ctx context.Context, events []*Sec
 		}
 	}
 }
+
+func (s *securityEventLogsStore) LogSecurityEvent(ctx context.Context, eventName SecurityEventName, url string, userID uint32, anonymousUserID string, source string, arguments []byte) {
```
mohammadualam, line 253: @evict Something like this? Re:your feedback on making it a function.

evict, line 253:
```suggestion
func LogSecurityEvent(ctx context.Context, eventName SecurityEventName, url string, userID uint32, anonymousUserID string, source string, arguments []byte) {
```

internal/database/security_event_logs.go
```diff
@@ -101,6 +132,8 @@ type SecurityEventLogsStore interface {
 	LogEvent(ctx context.Context, e *SecurityEvent)
 	// Bulk "LogEvent" action.
 	LogEventList(ctx context.Context, events []*SecurityEvent)
+
+	LogSecurityEvent(ctx context.Context, eventName SecurityEventName, url string, userID uint32, anonymousUserID string, source string, arguments []byte)
```
evict, line 136:
```suggestion
```

internal/database/security_event_logs.go
evict: Almost! Just as a function as good enough, and then replace the previous logging with that function. 👍

mohammadualam: Yup, i am updating those now, just wanted to see if i got your feedback correct

cmd/frontend/internal/dotcom/productsubscription/subscriptions_db.go
```diff
@@ -277,6 +297,14 @@ func (s dbSubscriptions) Update(ctx context.Context, id string, update dbSubscri
 	if nrows == 0 {
 		return errSubscriptionNotFound
 	}
+	if featureflag.FromContext(ctx).GetBoolOr("auditlog_expansion", false) {
+
+		argsJSON, _ := json.Marshal(id)
+		// Log an event when a subscription is updated
+		if err := database.LogSecurityEvent(ctx, database.SecurityEventNameDotComSubscriptionUpdated, "", uint32(actor.FromContext(ctx).UID), "", "BACKEND", argsJSON, s.db.SecurityEventLogs()); err != nil {
```
evict, line 304: You are passing the `s.db.SecurityEventLogs()`. Why don't you update the function and do that there? Same for the marshalling of the json, you can do that in `LogSecurityEvent` as well.

evict (changes requested):

client/web/src/featureFlags/featureFlags.ts
```diff
@@ -37,6 +37,7 @@ export const FEATURE_FLAGS = [
     'signup-survey-enabled',
     'cody-pro',
     'opencodegraph',
+    'auditlog_expansion',
```
evict, line 40: nit: other names have `-` instead of `_`. What about naming it 'enhanced-auditlog' or 'extra-audit-logs'. It's temporary right?

cmd/frontend/internal/dotcom/productsubscription/subscriptions_db.go
mohammadualam: i will work on doing the marshalling within the function. 
Regarding the s.db.securityeventlog() passing, i had my function setup this way 
`func (s *securityEventLogsStore) LogSecurityEvent(ctx context.Context, eventName SecurityEventName, url string, userID uint32, anonymousUserID string, source string, arguments []byte) {......}` 

do you mean i should go back to that definition or are you referring to something else?

evict: > do you mean i should go back to that definition or are you referring to something else?
I am sorry, yes. 🤦  I didn't realize you'd then have to pass that `db` reference around. So your previous method is the cleanest one. I thought it would be nice to instantiate everything in the exported function, but we need the `db` handle from the request to actually touch the database.

mohammadualam: reverted back to the original version.

internal/database/security_event_logs.go
```diff
@@ -216,3 +249,24 @@ func (s *securityEventLogsStore) LogEventList(ctx context.Context, events []*Sec
 		}
 	}
 }
+
+func (s *securityEventLogsStore) LogSecurityEvent(ctx context.Context, eventName SecurityEventName, url string, userID uint32, anonymousUserID string, source string, arguments any) error {
+	argsJSON, err := json.Marshal(arguments)
+	if err != nil {
+		return errors.New("error marshalling arguments")
+
```
pjlast, line 257:
```suggestion
		return errors.Wrap(err, "error marshalling arguments")
```

Might be worth seeing why the marshalling failed in the error

pjlast (approved):
Seems good to me pending @evict 's approval if he's happy after his requested changes

internal/database/security_event_logs_test.go
```diff
@@ -140,3 +142,25 @@ func assertEventField(t *testing.T, field map[string]any) {
...
+
+		// Assert event was logged
+	})
+
+	t.Run("invalid arguments", func(t *testing.T) {
+		err := db.SecurityEventLogs().LogSecurityEvent(ctx, SecurityEventAccessTokenCreated, "http://sourcegraph.com", 123, "AnonymousUserID", "source", make(chan int))
+		require.Error(t, err)
+
+		// Assert error from marshalling arguments
+	})
```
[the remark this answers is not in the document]
mohammadualam: They were supposed to be removed.

evict (approved):
Looks good! Nice work.

cmd/frontend/graphqlbackend/org.go
```diff
@@ -48,12 +50,28 @@ func (r *schemaResolver) Organization(ctx context.Context, args struct{ Name str
 		if err := hasAccess(); err != nil {
 			// site admin can access org ID
 			if auth.CheckCurrentUserIsSiteAdmin(ctx, r.db) == nil {
+				if featureflag.FromContext(ctx).GetBoolOr("auditlog-expansion", false) {
```
eseliger, line 53: I'm curious, what's so secret about orgs that we would log access to them? 
Not saying it's not important, just curious :)

PR #58847, opened 2023-12-08
svelte: Add pagination to commit page

fkling:
That was another roller coaster ride. This PR does two things (yes, I should have done that separately):

- Add pagination to the commits page
- Move all GraphQL codegen for the SvelteKit prototype inside the package itself, and enable support for `.gql` files.

## Pagination

The commits page was only showing the first 20 commits. The current React version shows a button at the bottom to load more commits, but that's not very user friendly. I implemented some form of pagination that works with the current forward-only page information by making assumptions about the cursor. I think that's fine for the prototype.

## GraphQL

Foolish me wanted to explore a "better" approach for working with GraphQL queries, letting components define their data dependencies via fragments and compose them by higher up components.
This PR demonstrates how this could work for the commits page:
- The commits page defines its queries in `page.gql`, which the loader imports and executes.
- The page imports fragments for the components it renders.

I added more graphql codgen logic to make working with GraphQL in general, and `.gql` files specifically, easier. The end result is that a module can import a query from a `.gql` file, pass it to Apollo's `query` method and get full type support without having to import variable and result types explicitly. Example:

```ts
import { CommitsQuery } from './page.gql'

//...
graphqlClient.query({query: CommitsQuery, variables: {...})
```

where `page.gql` is defined as

```gql
#import "$lib/Commit.gql"

fragment Commits on GitCommitConnection {
    nodes {
        ...Commit
    }
    pageInfo {
        hasNextPage
        endCursor
    }
}

query CommitsQuery($repo: ID!, $revspec: String!, $first: Int, $afterCursor: String) {
    node(id: $repo) {
        __typename
        ... on Repository {
            id
            commit(rev: $revspec) {
                id
                ancestors(first: $first, afterCursor: $afterCursor) {
                    ...Commits
                }
            }
        }
    }
}
```

I tried various combinations of plugins from https://the-guild.dev/graphql/codegen/plugins and ran into different kinds of issues. I eventually settled on the solution used in `vite.config.ts`.

## pnpm and bazel woes

The short story is that I ran into various issues that all somehow related to how the codgen plugins are dynamically imported. Even if specified in `client/web-sveltekit/src/package.json`, they would have been loaded from the workspace root, due how pnpm works.
Initially it seemed to work fine with pnpm alone, but broke in Bazel. Luckily later I encountered the same issue in pnpm alone, which made it easier to fix.
The solution was to declare the plugins as peer dependencies of `@graphql-codegen/cli` so that [pnpm would change the `node_modules` structure ](https://pnpm.io/how-peers-are-resolved)which makes the plugins (and auxiliary packages such as `graphql`) all resolve to the same version.

## Test plan

`pnpm dev`, `pnpm build`, `bazel build //client/web-sveltekit:web-sveltekit` and manual testing of the commits page.


camdencheek (approved):
Cool stuff!

Issue #58799, opened 2023-12-06
Prototype search language simplifications


keegancsmith:
We discussed simplifying the query language at our offsite. This issue is to track the work in actually working out the details on that as well as a prototype so we can gain experience on the change.

/cc @sourcegraph/search-platform

keegancsmith:
Notes from our meeting related to query language

If we could design the query language from scratch, what would it look like?

Repository filters:
* Exact match with glob patterns
   * Match sourcegraph repo: `sourcegraph/sourcegraph`
   * Match sourcegraph org: `sourcegraph/*`

File filters:
* Exact match with glob patterns
* Question: should `*` match across directories, or should we include `**`?
   * Extended glob with `**` might be a bit of a footgun … consider `*/vendor/*`
* Maybe fall back to regex for more complex use cases

Language filters
* Also support extension names, like `lang:md`

Query contents:
* Match literally
* Break on whitespace and AND tokens
* If you want to match exactly, wrap in quotes
* If you want to do regex, use `/…/`
* Quotes must be escaped – this may be annoying when copying and pasting in things like `throw new Exception("too many terms")`
* Keyword search: we actually make no guarantee about whether we use AND or OR, or stem …
   * If you explicitly use AND and OR, then we don’t stem
   * Question: What about single term queries? To avoid stemming these, do they need quotes?

keegancsmith:
I've been thinking of this in how we can translate our new queries into our old queries. The main reason is that is the easiest way to prototype this... write some code which transforms the queries in the frontend then everything else doesn't need to know about this.

Exact match with glob patterns. Only support `*`, and it matches across path components. This allows us to translate to regex.

```
repo:sourcegraph
repo:^sourcegraph$

repo:github.com/*
repo:^github\.com/.*$

repo:*/sourcegraph
repo:^.*/sourcegraph$
```

File has some logic as repo. We need some extra handling for the transformation since we transparently add in the `/` to ensure we are matching on absolute paths.

TODO For more complicated use cases we need a way to fallback to regex. This is a bit confusing due to `/PATTERN/` being the usual way to indicate regex.

```
file:README.md
file:^/README\.md$

file:client/README.md
file:^/client/README\.md$

file:*/Dockerfile
file:^.*/Dockerfile$

file:*.go
file:^.*\.go$

file:src/*
file:^/src/.*$
```

[...]

PR #58849, opened 2023-12-08
search: hack in new search language

keegancsmith:
We want to experiment with a new query language we cooked up in discussions last week. This is the first attempt of putting those ideas into code. Because this is intentionally experimental, this is done in a way where it is minimally invasive and simple: we convert queries into our older syntax just before we send them off to the server.

This is all hidden behind a feature flag "search-simple" which probably only works if you set it as an override in the client (and not via the normal feature flags). This is fine since we are experimenting.

This only implements the file and repo filters so far. Will follow up with more advanced things soon.

Test Plan: pnpm it and playing around with it via "sg start web-standalone" and adding "feat=search-simple" to the url.

image

Part of https://github.com/sourcegraph/sourcegraph/issues/58799


client/shared/src/search/searchSimple.test.ts
```diff
@@ -0,0 +1,41 @@
+import { describe, expect, test } from 'vitest'
+
+import { hacksGobQueriesToRegex } from './searchSimple'
+
+describe('hacksGobQueriesToRegex', () => {
+    test('gob corner cases', () => {
+        expect(hacksGobQueriesToRegex('')).toEqual('')
+
+        // TODO should this match nothing? Right the filter is a noop
```
fkling, line 9: ?

```suggestion
        // TODO should this match nothing? Right now the filter is a noop
```

fkling (approved):
Accepting, in the spirit of "moving fast"

jtibshirani (commented):
+1 to building a quick PoC, and a sort of translation approach makes sense to me. However, I'm wondering if we can tie this together with our existing work for `search-new-keyword`? Maybe the translation could happen as part of query parsing in frontend? At least it'd be nice to use the same feature flag to make it easy to enable it and get feedback.

keegancsmith:
FYI I updated this PR to use our query tokenizer. Was easier than I expected :) I imagine this might be easier to support than I imagined.

> However, I'm wondering if we can tie this together with our existing work for `search-new-keyword`?

Yes that is the state I am aiming for. Right now it is so hacky/experimental I'd rather not break that. IE this exists to play with the idea for us, not the rest of sg.

It feels like query language changes can rapidly be prototyped in a way where as a team we feel comfortable with what it becomes, then we decide the rest of the company should try it out.

In particular I wanna test it more and make sure things don't break too badly with it on in the UX. IE if autocomplete and query highlighting are broken I wanna make sure doesn't distract from our keyword experiment.

> Maybe the translation could happen as part of query parsing in frontend?

Yes, that is likely the final place for this to happen.

> At least it'd be nice to use the same feature flag to make it easy to enable it and get feedback.

I didn't wanna do it here because I was worried about too many other random things breaking.

keegancsmith:
@fkling mind taking another quick pass at the code and critique it from a TS style perspective? Keen to learn.

client/shared/src/search/searchSimple.ts
```diff
@@ -0,0 +1,88 @@
...
+                    case 'file':
+                    case 'f': {
+                        return gobToRegex(token.value.value)
+                    }
+
+                    default: {
+                        return token.value.value
+                    }
+                }
+            })()
```
fkling, line 51: Unusual for me to see, but I like the idea (I wish there were more kinds of expressions in JS). In general I avoid introducing unnecessary function calls but it's not a problem either.

fkling (approved):
Good idea to use the scanner (I should have suggested it). I've been working with all that code to improve the query input and it's nice to see it used more.

Overall there isn't any cray TS stuff going on, looks all good to me!

jtibshirani (approved):
Ah now I understand it's an extra-temporary feature flag :) Works for me.

PR #58850, opened 2023-12-08
database: simplify test helper

mrnugget:
We bumped into this earlier. Turns out this helper was a "would be nice" approach that didn't pan out. Instead of changing all of the tests to follow this approach, I decided to simplify and reduce the things that aren't needed because it makes maintenance easier.

## Test plan

NA


evict (approved):

eseliger (approved):
lgtm if ci is happy

PR #58851, opened 2023-12-08
embeddings: Return a 429 code when getting 429 from upstream

rafax:
Return 429 when upstream returns 429 status code (instead of returning 500, which can be retried immediately).

## Test plan

- tested locally with API client and Cody Chat
-
Screenshot 2023-12-08 at 14 44 02


keegancsmith (approved):

bobheadxi:
Minor nit - it would be helpful to title this in more detail, I saw this from a thread about embeddings status codes and came to make a comment about how Cody Gateway intentionally does not return upstream status codes, but turns out this is a different status code in our embeddings service 😛

rafax:
> Minor nit - it would be helpful to title this in more detail, I saw this from a thread about embeddings status codes and came to make a comment about how Cody Gateway intentionally does not return upstream status codes, but turns out this is a different status code in our embeddings service 😛

Updated - is that better?

bobheadxi (approved):
thanks for addressing title nit 😁

PR #58854, opened 2023-12-08
SSC: Handle returnTo params on /cody/manage

vdavid:
- Closes https://github.com/sourcegraph/cody/issues/2054

It was a small change, after all.

## Test plan

Watch me test it in this [3-min Loom](https://www.loom.com/share/66611ce7d82043669ef06d597876784f?sid=6be701b9-23ac-4a6e-b20d-e5d47255aeda)

I tested:
- If `/cody/manage` gets a `returnTo` parameter, it correctly goes through the onboarding process, then redirects to the desired page
- If it doesn't get the new param, it won't do a redirect

### DIFF
--- a/client/web/src/cody/onboarding/CodyOnboarding.tsx
+++ b/client/web/src/cody/onboarding/CodyOnboarding.tsx
@@ -114,9 +114,11 @@ export function CodyOnboarding({
 
     const navigate = useNavigate()
 
-    if (completed && !showEditorStep && returnToURL) {
-        navigate(returnToURL)
-    }
+    useEffect(() => {
+        if (completed && !showEditorStep && returnToURL) {
+            navigate(returnToURL)
+        }
+    }, [completed, returnToURL, navigate, showEditorStep])
 
     if (!showEditorStep && (completed || step === -1 || step > 1)) {
         return null
--- a/client/web/src/cody/onboarding/CodyOnboarding.tsx
+++ b/client/web/src/cody/onboarding/CodyOnboarding.tsx
@@ -115,10 +115,14 @@ export function CodyOnboarding({
     const navigate = useNavigate()
 
     useEffect(() => {
-        if (completed && !showEditorStep && returnToURL) {
+        if (completed && returnToURL) {
             navigate(returnToURL)
         }
-    }, [completed, returnToURL, navigate, showEditorStep])
+    }, [completed, returnToURL, navigate])
+
+    if (completed && returnToURL) {
+        return null
+    }
 
     if (!showEditorStep && (completed || step === -1 || step > 1)) {
         return null
--- a/client/web/src/cody/onboarding/CodyOnboarding.tsx
+++ b/client/web/src/cody/onboarding/CodyOnboarding.tsx
@@ -1,10 +1,11 @@
-import React, { useState, useEffect, useCallback } from 'react'
+import React, { useCallback, useEffect, useState } from 'react'
 
 import classNames from 'classnames'
+import { useNavigate } from 'react-router-dom'
 
 import { useTemporarySetting } from '@sourcegraph/shared/src/settings/temporary'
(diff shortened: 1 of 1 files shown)


client/web/src/cody/onboarding/CodyOnboarding.tsx
```diff
@@ -101,14 +102,21 @@ export function CodyOnboarding({
     authenticatedUser: AuthenticatedUser | null
 }): JSX.Element | null {
     const [showEditorStep, setShowEditorStep] = useState(false)
-    const [completed = true, setOnboardingCompleted] = useTemporarySetting('cody.onboarding.completed', false)
+    const [completed = false, setOnboardingCompleted] = useTemporarySetting('cody.onboarding.completed', false)
```
vdavid, line 105: Not sure why this was `true` but the page worked well in my test (see the attached demo) so it should be fine to change it?

client/web/src/cody/onboarding/CodyOnboarding.tsx
```diff
@@ -101,14 +102,21 @@ export function CodyOnboarding({
...
 
     const onNext = (): void => setOnboardingStep(currentsStep => (currentsStep || 0) + 1)
 
     const parameters = useSearchParameters()
     const enrollPro = parameters.get('pro') === 'true'
+    const returnToURL = parameters.get('returnTo')
+
+    const navigate = useNavigate()
+
+    if (completed && !showEditorStep && returnToURL) {
```
vdavid, line 117: It was a bit counter-intuitive to me why `completed` becomes `true` before actually completing the flow. Is this an artifact of rapid iterative development (that is, that the editor steps came later than the original design) @thenamankumar, or is this intentional?

client/web/src/cody/onboarding/CodyOnboarding.tsx
thenamankumar: If you make completed = false, then the modal will always first open while the settings are still loading, and then it hides, which flashes the screen.

client/web/src/cody/onboarding/CodyOnboarding.tsx
thenamankumar: It is because, as part of the recent request, we do not want to preserve the editor step state. It should only appear the first time. If the user has submitted the work/personal form, then we need to consider it completed and on next re-load not show the editor step.

Also, because we do not need to show the editor step in case we need to redirect back to the editor, so the condition should rather just be:

`completed && returnToURL`

thenamankumar (changes requested):
There are some logic changes proposed in the comments above. 

Also I don't not see where we are making `/post-sign-up` redirect to `/cody/manage`.

If you will be doing that in a separate PR, please take care that the new redirection should be feature flagged using the same `cody-pro` ff.

client/web/src/cody/onboarding/CodyOnboarding.tsx
thenamankumar: The code should look like:

```javascript
useEffect(() => {
  if (completed &&  returnToURL) {
        navigate(returnToURL)
  }
}, [completed, returnToURL, navigate])

if (completed &&  returnToURL) {
        return null;
}
```

client/web/src/cody/onboarding/CodyOnboarding.tsx
thenamankumar: Yeah, sounds good to me now.

vdavid:
> Also I don't not see where we are making /post-sign-up redirect to /cody/manage.
> If you will be doing that in a separate PR, please take care that the new redirection should be feature flagged using the same cody-pro ff.

Oh, yeah, almost forgot that. Thanks for pointing it out. I think that will be in the `sourcegraph/cody` repo somewhere. I'll take a look.

PR #58856, opened 2023-12-08
internal/database: remove test that takes 10s and is not worth it

mrnugget:
This test is the slowest one in the whole package and it's not worth it.

## Test plan

- NA

### DIFF
--- a/internal/database/BUILD.bazel
+++ b/internal/database/BUILD.bazel
@@ -276,7 +276,6 @@ go_test(
         "//internal/actor",
         "//internal/api",
         "//internal/authz",
-        "//internal/collections",
         "//internal/conf",
         "//internal/database/basestore",
         "//internal/database/batch",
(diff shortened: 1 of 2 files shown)


eseliger (approved):

unknwon:
Fly-by: what's the argument/statement of "not worth it"? We did run into a bug that on-prem customer had so many repos that made the PostgreSQL throw the 65535 limit exceeded error (which we fixed it at the time).

mrnugget:
> Fly-by: what's the argument/statement of "not worth it"? We did run into a bug that on-prem customer had so many repos that made the PostgreSQL throw the 65535 limit exceeded error (which we fixed it at the time).

It's a tradeoff. On the one hand: test takes 10s (!!!) to run. That's an eternity in computer time. I have to run that every time I execute the database tests. On the other hand: code can be confirmed to be correct without tests too, it has been tested, it works, and I think the risk that this behaviour will be reintroduced without this test is low.

Issue #56704, opened 2023-09-15
cody-gateway/fireworks: collect streaming completions usage data


bobheadxi:
This was recently added by the fireworks team: https://sourcegraph.slack.com/archives/C0514L18QHG/p1694821490267199?thread_ts=1694617539.506169&cid=C0514L18QHG

https://readme.fireworks.ai/docs/openai-compatibility#token-usage-for-streaming-responses

/cc @unknwon @eseliger @bobheadxi

PR #53510, opened 2023-06-14
completions/http: fix broken stream error reporting

bobheadxi:

PR #58859, opened 2023-12-08
cody-gateway: For streaming requests, flush response after every write

rafax:
For streaming requests, we don't want to buffer responses, but instead send the data to the client as soon as possible - see [Slack](https://sourcegraph.slack.com/archives/C05AGQYD528/p1701791953787829) and this issue https://github.com/sourcegraph/cody/issues/2180. This PR detects streaming requests (from all providers) and automatically flushes the response after every write.

Improvement can be observed using this https://sourcegraph.slack.com/archives/C05AGQYD528/p1702287275092429?thread_ts=1701791953.787829&cid=C05AGQYD528.
From local tests, this cuts time-to-first-event from ~1s to ~300ms (Fireworks Starcoder, measured client side). 
## Test plan

- tested locally
- unit tests

### DIFF
--- a/cmd/cody-gateway/internal/httpapi/completions/fireworks.go
+++ b/cmd/cody-gateway/internal/httpapi/completions/fireworks.go
@@ -32,6 +32,7 @@ func NewFireworksHandler(
 	allowedModels []string,
 	logSelfServeCodeCompletionRequests bool,
 	disableSingleTenant bool,
+	autoFlushStreamingResponses bool,
 ) http.Handler {
 	return makeUpstreamHandler(
 		baseLogger,
@@ -165,6 +166,7 @@ func NewFireworksHandler(
 		// Setting to a valuer higher than SRC_HTTP_CLI_EXTERNAL_RETRY_AFTER_MAX_DURATION to not
 		// do any retries
 		30, // seconds
+		autoFlushStreamingResponses,
 	)
 }
 
--- a/cmd/cody-gateway/internal/httpapi/completions/fireworks.go
+++ b/cmd/cody-gateway/internal/httpapi/completions/fireworks.go
@@ -6,9 +6,6 @@ import (
 	"encoding/json"
 	"github.com/sourcegraph/log"
 	"github.com/sourcegraph/sourcegraph/cmd/cody-gateway/internal/actor"
-	"io"
-	"net/http"
-
 	"github.com/sourcegraph/sourcegraph/cmd/cody-gateway/internal/events"
 	"github.com/sourcegraph/sourcegraph/cmd/cody-gateway/internal/limiter"
 	"github.com/sourcegraph/sourcegraph/cmd/cody-gateway/internal/notify"
@@ -16,7 +13,10 @@ import (
 	"github.com/sourcegraph/sourcegraph/internal/completions/client/fireworks"
 	"github.com/sourcegraph/sourcegraph/internal/conf/conftypes"
 	"github.com/sourcegraph/sourcegraph/internal/httpcli"
+	streamhttp "github.com/sourcegraph/sourcegraph/internal/search/streaming/http"
 	"github.com/sourcegraph/sourcegraph/lib/errors"
+	"io"
+	"net/http"
 )
 
 const fireworksAPIURL = "https://api.fireworks.ai/inference/v1/completions"
@@ -101,6 +101,35 @@ func NewFireworksHandler(
 			transformRequest: func(r *http.Request) {
 				r.Header.Set("Content-Type", "application/json")
 				r.Header.Set("Authorization", "Bearer "+accessToken)
+			},
+			forwardResponse: func(ctx context.Context, w http.ResponseWriter, body io.Reader, logger log.Logger, request fireworksRequest) error {
+				if !request.Stream {
+					return defaultForwardResponse[fireworksRequest](ctx, w, body, logger, request)
+				}
+				eventWriter, err := streamhttp.NewWriter(w)
+				if err != nil {
+					return err
+				}
+
+				dec := fireworks.NewDecoder(body)
+
+				for dec.Scan() {
+					if ctx.Err() != nil && errors.Is(ctx.Err(), context.Canceled) {
+						return err
+					}
+
+					data := dec.Data()
+
+					err := eventWriter.EventBytes("", data)
+
+					if err != nil {
+						logger.Error("failed to write event", log.Error(err))
+					}
+				}
+				defer func() {
+					_ = eventWriter.EventBytes("", []byte("[DONE]"))
+				}()
+				return nil
 			},
 			parseResponseAndUsage: func(logger log.Logger, reqBody fireworksRequest, r io.Reader) (promptUsage, completionUsage usageStats) {
 				// First, extract prompt usage details from the request.
--- a/cmd/cody-gateway/internal/httpapi/completions/fireworks.go
+++ b/cmd/cody-gateway/internal/httpapi/completions/fireworks.go
@@ -4,7 +4,11 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"io"
+	"net/http"
+
 	"github.com/sourcegraph/log"
+
 	"github.com/sourcegraph/sourcegraph/cmd/cody-gateway/internal/actor"
 	"github.com/sourcegraph/sourcegraph/cmd/cody-gateway/internal/events"
 	"github.com/sourcegraph/sourcegraph/cmd/cody-gateway/internal/limiter"
@@ -14,8 +18,6 @@ import (
 	"github.com/sourcegraph/sourcegraph/internal/conf/conftypes"
 	"github.com/sourcegraph/sourcegraph/internal/httpcli"
 	"github.com/sourcegraph/sourcegraph/lib/errors"
-	"io"
-	"net/http"
 )
 
 const fireworksAPIURL = "https://api.fireworks.ai/inference/v1/completions"
--- a/cmd/cody-gateway/internal/httpapi/completions/fireworks.go
+++ b/cmd/cody-gateway/internal/httpapi/completions/fireworks.go
@@ -13,7 +13,6 @@ import (
 	"github.com/sourcegraph/sourcegraph/internal/completions/client/fireworks"
 	"github.com/sourcegraph/sourcegraph/internal/conf/conftypes"
 	"github.com/sourcegraph/sourcegraph/internal/httpcli"
-	streamhttp "github.com/sourcegraph/sourcegraph/internal/search/streaming/http"
 	"github.com/sourcegraph/sourcegraph/lib/errors"
 	"io"
 	"net/http"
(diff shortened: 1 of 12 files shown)


cmd/cody-gateway/internal/httpapi/completions/fireworks.go
```diff
@@ -6,17 +6,17 @@ import (
...
 	"github.com/sourcegraph/sourcegraph/cmd/cody-gateway/internal/notify"
 	"github.com/sourcegraph/sourcegraph/internal/codygateway"
 	"github.com/sourcegraph/sourcegraph/internal/completions/client/fireworks"
 	"github.com/sourcegraph/sourcegraph/internal/conf/conftypes"
 	"github.com/sourcegraph/sourcegraph/internal/httpcli"
+	streamhttp "github.com/sourcegraph/sourcegraph/internal/search/streaming/http"
 	"github.com/sourcegraph/sourcegraph/lib/errors"
+	"io"
+	"net/http"
 )
```
eseliger, line 20: import ordering seems off

cmd/cody-gateway/internal/httpapi/completions/fireworks.go
```diff
@@ -102,6 +102,35 @@ func NewFireworksHandler(
...
+				if err != nil {
+					return err
+				}
+
+				dec := fireworks.NewDecoder(body)
+
+				for dec.Scan() {
+					if ctx.Err() != nil && errors.Is(ctx.Err(), context.Canceled) {
+						return err
+					}
```
eseliger, line 119: is there a reason to specifically only return for canceled, and not for example deadlineexceeded or so?

cmd/cody-gateway/internal/httpapi/completions/fireworks.go
```diff
@@ -102,6 +102,35 @@ func NewFireworksHandler(
 				r.Header.Set("Content-Type", "application/json")
 				r.Header.Set("Authorization", "Bearer "+accessToken)
 			},
+			forwardResponse: func(ctx context.Context, w http.ResponseWriter, body io.Reader, logger log.Logger, request fireworksRequest) error {
+				if !request.Stream {
+					return defaultForwardResponse[fireworksRequest](ctx, w, body, logger, request)
+				}
+				eventWriter, err := streamhttp.NewWriter(w)
```
eseliger, line 109: I think it would be good do add a comment here why we do this -- it doesn't seem super obvious to me.

Also, is this not a problem for anthropic or openai?

cmd/cody-gateway/internal/httpapi/completions/openai.go
```diff
@@ -69,6 +69,7 @@ func NewOpenAIHandler(
 					r.Header.Set("OpenAI-Organization", orgID)
 				}
 			},
+			forwardResponse: defaultForwardResponse[openaiRequest],
```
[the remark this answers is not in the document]
rafax: I will move to interfaces after this ships (this is blocking Dec GA improvements).

For the current implementation, I biased toward being "explicit" (as a reader, I'd be surprised that all other functions are required)

cmd/cody-gateway/internal/httpapi/completions/fireworks.go
bobheadxi, line 109: Using `streamhttp` is tricky within a large abstraction like the upstream handlers here - once you create a writer, you cannot interact with the `ResponseWriter` anymore.  I left a docstring on `streamhttp` a while back:

https://github.com/sourcegraph/sourcegraph/blob/54d01435661c4f72d3c9918bcbc2ac066848016e/internal/search/streaming/http/writer.go#L26-L35

Some more details here, and a confirmation from Camden: https://github.com/sourcegraph/sourcegraph/pull/53510

eseliger (approved):

cmd/cody-gateway/internal/httpapi/completions/fireworks.go
bobheadxi: In this case, on L378 of `cmd/cody-gateway/internal/httpapi/completions/upstream.go` 
 (I should've commented there instead 😆 ) you have an `if err != nil` check that could write to the ResponseWriter which is not allowed. Looks like that's the only instance so far, but we need to make sure no future additions add any writes from that point on

bobheadxi: At a second look I'm not sure I follow either - when other providers initiate a stream, Cody Gateway forwards that directly back to the caller and it "just works". Here it sounds like we read decode the stream and send the same stream back to the client again, with an extra `[DONE]` event - do we need to do that?

We should be able to just forward the Fireworks streaming responses as-is, as we do the other providers today - if there's an issue with their API they've been very responsive in the past (eg #56704 ) so maybe they can just fix any issues on their end, instead of us rewriting their stream?

bobheadxi: I think https://pkg.go.dev/io#CopyBuffer might let you control the "chunking" by providing a buffer you've allocated with the desired size. The default buffer is very large: `32 * 1024` https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/io/io.go;l=418-427

bobheadxi: I think for code completions which will probably be less than 120 characters per line, maybe we can use a buffer of 64, or even 32? e.g.

```go
io.CopyBuffer(w, respBody, make([]byte, 64))
```

cmd/cody-gateway/internal/httpapi/completions/fireworks.go
rafax: No, that's code that I copied from existing uses of Decoder, but you're right that it needs adjusting for CG.

cmd/cody-gateway/internal/httpapi/completions/fireworks.go
rafax: > Also, is this not a problem for anthropic or openai?
It is far less noticable for providers with higher latency, but I intend to fix all of them. Fixing this for Fireworks is motivated by unblocking the "adaptive multi-line completion" feature that is planned for launch on December GA.

rafax: > At a second look I'm not sure I follow either - when other providers initiate a stream, Cody Gateway forwards that directly back to the caller and it "just works". Here it sounds like we read+decode the stream and write the same stream back to the client again, with an extra [DONE] event - do we need to do that?

Cody Gateway forwards the response, but only after an initial buffering period, and the http.Server buffer is not configurable - we need to call Flush ourselves. An alternative would be to copy in small chunks and flush after every single one, but this has other problems (ex. force flushing small chunks introduces more network overhead, and the client can'd do anything with an incomplete JSON object)

rafax: > Looking at the linked Slack thread it looks like an issue with io.Copy not having the properties we want. I think a better fix might be to replace io.Copy on the io.Reader/io.Writer level, rather than building the stream ourselves

Fair, we could try to reimplement io.Copy to do "adaptive flushing", I chose to use an existing Fireworks-specific SSE parser to optimize for "time-to-fix" and correctness, once I try to support this for other providers, I will see if there's a generalized way of doing this (or, I will rewrite all streaming endpoints in gRPC 😛  )

bobheadxi: IMO an adaptive flush is probably less error prone, ensuring we pass the provider response back as-is without any funny business. With this approach, `io.CopyBuffer` will be our buffering mechanism to balance network writes with expedience - we could make the buffer size configurable for each provider via env var as well

bobheadxi: Out of curiosity, I took a look at what happens in https://pkg.go.dev/net/http/httputil#ReverseProxy and it seems they have special handling specific to streams to forgo buffering: https://github.com/golang/go/commit/5440bfc2ea8c0a4c78d5161605659c07ea10e37a

Maybe you know this already 😛 but based on this, it looks _possible_ that any `io.Read` with a sufficiently large buffer will immediately return as soon as the upstream server flushes with an "entire message" so to speak. So `io.Copy` that flushes on every `Write` should be sufficient to send entire messages back to the client as soon as the upstream server sends it to us, and we don't need to worry about partial responses or excessive writes 🤔

cmd/cody-gateway/internal/httpapi/completions/openai.go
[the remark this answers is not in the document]
rafax: We'll move to the interface, and I reworked the code to use io.Copy with different writers.

cmd/cody-gateway/internal/httpapi/completions/fireworks.go
rafax: No longer using `streamhttp`

cmd/cody-gateway/internal/httpapi/completions/fireworks.go
rafax: Yeah, weird, fixed now.

cmd/cody-gateway/internal/httpapi/completions/fireworks.go
rafax: Comments added + code re-worked.

rafax: I went with "non-adaptive flushing" for streaming requests - basically, flush after every write (like reverse proxy with negative FlushInterval). From testing locally, this is sufficient, but I added a metric on dotcom to track improvements.

client/web-sveltekit/src/lib/icons/BatchChanges.svelte
```diff
@@ -0,0 +1,20 @@
+<svg viewBox="0 -3 38 38" width="20" height="20" fill="currentColor" xmlns="http://www.w3.org/2000/svg">
```
valerybugakov, line 1: Unexpected changes; leftovers from rebasing the branch?

client/web-sveltekit/src/lib/icons/BatchChanges.svelte
philipp-spiess, line 1: There are a bunch of svelte files in this PR which seem out of place

cmd/cody-gateway/internal/response/response.go
```diff
@@ -101,3 +104,31 @@ func (e HTTPStatusCodeError) HTTPStatusCode() int { return e.status }
...
+	flusher http.Flusher
+}
+
+var _ io.Writer = autoFlusher{}
+
+// NewAutoFlushingWriter returns an io.Writer that will flush after every Write()
+// call. It will return an error if the response writer doesn't implement the Flusher interface.
+func NewAutoFlushingWriter(w http.ResponseWriter) (io.Writer, error) {
+	if flusher, ok := w.(http.Flusher); !ok {
+		return nil, errors.Newf("can't flush response using", w)
```
valerybugakov, line 119: Would it be helpful to say here why we can't flush?

client/web-sveltekit/src/lib/icons/BatchChanges.svelte
philipp-spiess: Oh it seems that there are some commits in this branch which aren't on `main`. Maybe GitHub is lagging. The Cody Gateway change itself looks very reasonable!

valerybugakov (approved):

client/web-sveltekit/src/lib/icons/BatchChanges.svelte
rafax: For a moment, it looked like local `main` rebase was ahead of remote `main` rebase, but should look fine now.

rafax:
@bobheadxi @eseliger I will merge this now to test in a non-local environment (code is behind a "feature flag"), happy to address any post-review feedback

cmd/cody-gateway/internal/httpapi/completions/upstream.go
```diff
@@ -372,8 +372,18 @@ func makeUpstreamHandler[ReqT UpstreamRequest](
...
-			_, _ = io.Copy(w, respBody)
+			// if this is a streaming request, we want to flush ourselves instead of leaving that to the http.Server
+			// (so events are sent to the client as soon as possible)
+			var responseWriter io.Writer = w
+			if autoFlushStreamingResponses && body.ShouldStream() {
+				if fw, err := response.NewAutoFlushingWriter(w); err == nil {
+					responseWriter = fw
+				} else {
+					// We can't stream the response, but it's better to write it without streaming that fail, so we just log the error
+					logger.Error("failed to create auto-flushing writer", log.Error(err))
```
bobheadxi, line 383: I might add a `log.String("writerType", fmt.Sprintf("%T", w)) here as well

cmd/cody-gateway/internal/httpapi/completions/upstream.go
```diff
@@ -75,6 +75,7 @@ type upstreamHandlerMethods[ReqT UpstreamRequest] struct {
 
 type UpstreamRequest interface {
 	GetModel() string
+	ShouldStream() bool
```
bobheadxi, line 78: nice 😁

cmd/cody-gateway/internal/httpapi/completions/fireworks.go
bobheadxi: LGTM overall, thank you!!!

PR #58861, opened 2023-12-08
Sveltekit Prototype: add file icons to svelte prototype

jasonhawkharris:
This adds file icons to the tree view in the svelte kit prototype.

## Test plan
Manual, visual testing

### DIFF
--- a/client/web-sveltekit/src/lib/fileIcons.ts
+++ b/client/web-sveltekit/src/lib/fileIcons.ts
@@ -44,7 +44,7 @@ const YELLOW = 'var(--yellow)'
 const RED = 'var(--red)'
 const GREEN = 'var(--green)'
 const CYAN = 'var(--blue)'
-const DEFAULT_ICON = 'var(--gray-05)'
+const GRAY = 'var(--gray-05)'
 
 enum FileExtension {
     ASSEMBLY = 'asm',
@@ -61,7 +61,6 @@ enum FileExtension {
     CSHARP = 'cs',
     CSS = 'css',
     DART = 'dart',
-    DEFAULT = 'default',
     DOCKERFILE = 'Dockerfile',
     DOCKERIGNORE = 'dockerignore',
     FORTRAN_F = 'f',
@@ -135,7 +134,7 @@ interface FileInfo {
     isTest: boolean
 }
 
-const getColor = (extension: FileExtension) => {
+function getColor(extension: FileExtension) {
     switch (extension) {
         case FileExtension.DOCKERFILE:
         case FileExtension.DOCKERIGNORE:
@@ -175,7 +174,7 @@ const getColor = (extension: FileExtension) => {
             return CYAN
         }
         default: {
-            return DEFAULT_ICON
+            return GRAY
         }
     }
 }
@@ -184,8 +183,7 @@ const getColor = (extension: FileExtension) => {
 // Most programming language Material Design icons will be
 // deprecated very soon, and will not be included in the next release.
 // Additionally, it doesn't have all the icons we need anyway.
-const getIconAttributes = (extension: FileExtension): IconInfo => {
-    const color = getColor(extension)
+function getIcon(extension: FileExtension): string {
     switch (extension) {
         // TODO: Ideally, we would switch to a new icon library
         // so we don't have to default to a general icon
@@ -198,7 +196,6 @@ const getIconAttributes = (extension: FileExtension): IconInfo => {
         case FileExtension.CLOJURE_EDN:
         case FileExtension.CLOJURE_CLJS:
         case FileExtension.DART:
-        case FileExtension.DEFAULT:
         case FileExtension.FORTRAN_F:
         case FileExtension.FORTRAN_FOR:
         case FileExtension.FORTRAN_FTN:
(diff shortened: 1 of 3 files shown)


client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte
```diff
@@ -125,7 +123,12 @@
                     tabindex={-1}
                     data-go-up={isRoot ? true : undefined}
                 >
-                    <Icon svgPath={getIconPath(entry, expanded)} inline />
+                    {#if entry.isDirectory}
+                        <Icon svgPath={getIconPath(entry, expanded)} inline />
+                    {:else}
+                        {@const fileInfo = getFileInfo(entry.name)}
+                        <Icon svgPath={fileInfo.icon.svgPath} inline --color={fileInfo.icon.color} />
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte:130]: Tiny nit: I think you could do this if you want to make it more compact (I like shorthand props 🤷🏻 )

```suggestion
                        {@const {icon: {svgPath, color}} = getFileInfo(entry.name)}
                        <Icon {svgPath} inline --color={color} />
```

client/web-sveltekit/src/lib/fileIcons.ts
```diff
@@ -0,0 +1,358 @@
...
+        case FileExtension.YML: {
+            return { svgPath: mdiCog, color }
+        }
+        default: {
+            return { svgPath: mdiConsole, color }
+        }
+    }
+}
+
+export const getExtension = (file: string): FileExtension => {
```
fkling, line 336: Doesn't look like this needs to be exported.

IMO this should return `string` because you don't actually know whether the file extension is in `FileExtension` but you are giving the (false) impression that it is.

The `FileExtension` enum is only really useful internally to ensure that you are processing all defined extensions.

client/web-sveltekit/src/lib/fileIcons.ts
```diff
@@ -0,0 +1,358 @@
...
+    TOML = 'toml',
+    TYPESCRIPT = 'ts',
+    TSX = 'tsx',
+    TEXT = 'txt',
+    YAML = 'yaml',
+    YML = 'yml',
+    ZIG = 'zig',
+}
+
+export const FILE_ICONS = new Map(
```
fkling, line 127: Looks like this doesn't need to be exported.

client/web-sveltekit/src/lib/fileIcons.ts
fkling, line 1: This file still contains code paths that are not clear/obvious. For example: I noticed in line 338 that the following is returned as the default icon

```js
        default: {
            return { svgPath: mdiConsole, color }
        }
```

which didn't seem like a good choice for a default icon to me. But it appears that this case is never reached, because `getExtension` returns `'default'` for an unknown extension and the following icon is returned for `'default'`
 `return { svgPath: mdiFileCodeOutline, color }`.

For the same reason the code in line 363 is never executed. `if (icon)` will never be `false` (unless I overlooked something).

---

My suggestion would be:

- Remove `default` from  `FileExtension`. It's not an actual file extension anyway.
- `getExtension` should return a `string` value and it can just be `return file.split('.').at(-1)` (or just do that in `getFileInfo`, no need for a separate function). You are using that value only to look up something in `FILE_ICON`. If the extension is not known, `icon` will be `null` and the fallback case in line 363 is executed.
- Either do not have a `default:` case in the `getIconAttribute` function, choose a different default icon or add a comment that explains why this icon is the `default:` case.
- Change `FILE_ICON` to be a `Map<string, IconInfo>` (if it isn't already, maybe that's what TypeScript infers anyways).


What this essentially means is:

- `getFileInfo` returns the configured icon/color for a known file extension, or a default icon/color for an unknown file extension.
- `getIconAttribute` and `getColor` return the configuration for known file extensions. The `default:` case in each function then means something like "for all known but not explicitly configured extensions, use this value".

client/web-sveltekit/src/lib/fileIcons.ts
```diff
@@ -0,0 +1,372 @@
...
+        case FileExtension.JULIA:
+        case FileExtension.OCAML:
+        case FileExtension.PERL:
+        case FileExtension.PERL_PM:
+        case FileExtension.SCALA:
+        case FileExtension.SVELTE:
+        case FileExtension.SQL:
+        case FileExtension.TEST:
+        case FileExtension.ZIG: {
+            return { svgPath: mdiFileCodeOutline, color }
```
fkling, line 218: To avoid repetition I'd suggest that this function only return the icon path, i.e. `return mdiFileCodeOutline`, and you build the final `{svgPath, color}` below where you create the map.

I understand that this is probably a leftover from using multiple icon libraries. So maybe at some point we need to return more complex information, not just the icon path.
But I assume the color will most likely be independent of the icon library, so we could pull that out.

client/web-sveltekit/src/lib/fileIcons.ts
```diff
@@ -0,0 +1,372 @@
...
+    return {
+        icon: {
+            svgPath: mdiFileCodeOutline,
+            color: DEFAULT_ICON,
+        },
+        isTest: false,
+    }
+}
+
+const FILE_ICONS = new Map(Object.values(FileExtension).map(extension => [extension, getIconAttributes(extension)]))
```
fkling, line 372: If you change `getIconAttributes` to only return the icon path (and e.g. rename it to `getIcon`) then this could be

```js
...map(extension => [extension, {svgPath: getIcon(extension), color: getColor(extension)}]
```

client/web-sveltekit/src/lib/fileIcons.ts
```diff
@@ -0,0 +1,372 @@
...
+    svgPath: string
+    color: string
+}
+
+interface FileInfo {
+    icon: IconInfo
+    isTest: boolean
+}
+
+const getColor = (extension: FileExtension) => {
```
fkling, line 138: Nit: I know we use arrow functions a lot in this repo, but there are at least two reasons why I prefer function declarations when possible: 
- It avoids the error you've mentioned in slack. Because function declarations are _hoisted_ you have calls to that function before its definition (i.e. it doesn't matter where the function definition is located in the file).
- It makes it slightly clearer that this is a function. When scanning the code from left to right, `const getColor` doesn't tell you anything about what `getColor` is, but `function getColor` does.

Especially the first point makes function declarations more useful at the top level IMO, but it's up to you.

client/web-sveltekit/src/lib/fileIcons.ts
```diff
@@ -0,0 +1,372 @@
...
+
+import { containsTest } from '$lib/web'
+
+const BLUE = 'var(--blue)'
+const PINK = 'var(--pink)'
+const YELLOW = 'var(--yellow)'
+const RED = 'var(--red)'
+const GREEN = 'var(--green)'
+const CYAN = 'var(--blue)'
+const DEFAULT_ICON = 'var(--gray-05)'
```
fkling, line 47: I'd rename this to `GRAY`. No need to encode the semantics in the variable name.

client/web-sveltekit/src/lib/fileIcons.ts
```diff
@@ -0,0 +1,366 @@
...
+        }
+        default: {
+            return mdiFileCodeOutline
+        }
+    }
+}
+
+export function getFileInfo(file: string): FileInfo {
+    const extension = file.split('.').at(-1)
+    const icon = FILE_ICONS.get(extension as FileExtension)
```
fkling, line 343: Do you have to cast `extension`? If yes, you could probably declare the map as `const FILE_ICONS: Map<string, ...> = ...`.

fkling (approved):
Thank you for iterating on this... how do you feel about the end result?
FWIW, for visual changes it's nice to see a screenshot or video of the change in the PR description.

PR #58865, opened 2023-12-08
svelte: Show commit info next to tree entries on tree page

fkling:
This commit builds on #58847 (which enables a new way of working with GraphQL). I initially also changed the loader to query for the list of files directly instead of using `fetchTreeEntries`, but the same function is used by the file tree sidebar, which, thanks to caching, ensures that we have to fetch this data only once. I'd still like to refactor this at some point and make it so that data dependencies are declared next to the components that need them, but that should be done in a separate PR/commit.

This commit also changes the prop of the Timestamp component, because props that default to `false` are easier to work with. This also uncovered an issue with the test setup.

## Test plan

Manual testing

### DIFF
--- a/client/web-sveltekit/src/lib/Timestamp.svelte
+++ b/client/web-sveltekit/src/lib/Timestamp.svelte
@@ -24,20 +24,22 @@
     export let date: Date | string
 
     /** Use exact timestamps (i.e. omit "less than", "about", etc.) */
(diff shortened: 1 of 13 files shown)


vovakulikov (approved):

camdencheek (approved):
Nice! Looks great

PR #58869, opened 2023-12-08
Embeddings: fail job immediately if rate limited exceeded

jtibshirani:
Usually, during an embeddings job we allow 10% of embedding requests to fail,
simply skipping over failed chunks. If a customer has hit their rate limits,
this means we might continually send a huge number of embedding requests that
we know will immediately fail. With this change, we immediately fail a job if
the rate limit is exceeded.

It also increases the amount of time between attempting to run a job to 15
minutes. This won't make a big difference to user experience, since by default
embeddings jobs aren't allowed to be scheduled within 24h of the last run. But
it helps prevent jobs from continuously being scheduled then failing.

This change is unlikely to have a user-facing impact, but just helps cut down
on noise in logs and excessive requests to Cody Gateway.

## Test plan

Added new unit test

## Preview 🤩
[Preview Link](https://docs.sourcegraph.com/@jtibs/embeddings)

### DIFF
--- a/internal/embeddings/embed/embed.go
+++ b/internal/embeddings/embed/embed.go
@@ -259,22 +259,25 @@ func embedFiles(
 		}
 
 		batchEmbeddings, err := embeddingsClient.GetDocumentEmbeddings(ctx, batchChunks)
-		if err != nil && !excludeChunksOnError {
-			return nil, errors.Wrap(err, "error while getting embeddings")
-		} else if err != nil {
-			// To avoid failing large jobs on a flaky API, just mark all files
-			// as failed and continue. This means we may have some missing
-			// files, but they will be logged as such below and some embeddings
-			// are better than no embeddings.
-			logger.Warn("error while getting embeddings", log.Error(err))
-			failed := make([]int, len(batchChunks))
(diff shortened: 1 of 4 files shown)


internal/embeddings/embed/embed.go
```diff
@@ -259,22 +259,25 @@ func embedFiles(
...
-				Dimensions: dimensions,
+		if err != nil {
+			if !excludeChunksOnError || errors.Is(err, &client.RateLimitExceededError{}) {
+				// Fail immediately if we hit a rate limit, so we don't continually retry and fail on every chunk.
+				return nil, errors.Wrap(err, "error while getting embeddings")
+			} else {
+				// To avoid failing large jobs on a flaky API, just mark all files
+				// as failed and continue. This means we may have some missing
+				// files, but they will be logged as such below and some embeddings
+				// are better than no embeddings.
```
keegancsmith, line 270: do we know what errors we would get back in this case? Feels like it would be more robust to allow list the errors we have this behaviour with rather than assuming it is the flaky API. Either way this is an improvement so rather land this PR than directly address this comment. Also fine to say we don't know what the errors are :)

keegancsmith (approved):

PR #58877, opened 2023-12-11
frontend/auth: support ip allowlist

michaellzc:
Context: https://sourcegraph.slack.com/archives/C03LCPCT3SP/p1702082239371689

This PR implemented IP allowlist filtering at the application level. This only works when:

- we only filter on connected client IP addessr (this is usually the load balancer IP instead of true user IP)
- the admin is 100% confident the configured user IP header is reliable. 

For example, the Sourcegraph instance only permits connections from the load balancer (or Cloudflare proxy), and we trust upstream to configure the correct user IP in the header.

Open question:

- this will most certainly break inter-pod-comm when enabled. the user must adds the pod/svc cidr to the list as well.

### Will this accidentally lock admin out if the wrong IP is provided?

Yes. However, we will always permit `127.0.0.1` access for break glass.

### I thought we can spoof the headers.

Yes. Hence the admin needs to make sure the header is coming from a trusted source, e.g., restrict connection at the load balancer level.

## Test plan

Below is some basic cases, more can be found from added unit tests

Edit `../dev-private/enterprise/dev/site-config.json`

```json
  "auth.allowedIpAddress": {
    "enabled": false,
    "userIpAddress": [
      "1.1.1.1",
    ],
  },
```

```sh
sg start
```

You should be access frontend.

Edit `../dev-private/enterprise/dev/site-config.json`

```json
  "auth.allowedIpAddress": {
    "enabled": true,
    "userIpAddress": [
      "1.1.1.1",
    ],
  },
```

You should be able to access frontend

Edit `../dev-private/enterprise/dev/site-config.json`

```json
  "auth.allowedIpAddress": {
    "enabled": true,
    "userIpAddress": [
      "127.0.0.1",
    ],
  },
```

You should be able to access frontend

Edit `../dev-private/enterprise/dev/site-config.json`

```json
  "auth.allowedIpAddress": {
    "enabled": true,
    "userIpAddress": [
      "127.0.0.1",
    ],
    "userIpRequestHeaders": ["CF-Connecting-IP"],
  },
```

You should see error message unable to retrieve user IP.

Then, try `curl -H "CF-Connecting-IP: 127.0.0.1" https://sourcegraph.test:3443`, it should work again

Edit site config

```
  "auth.allowedIpAddress": {
    "enabled": true,
    "userIpAddress": [
      "185.170.224.233",
      // "127.0.0.1",
    ],
    "errorMessageTemplate": "Please reach out to your Sourcegraph instance admin at admin@acme.com to unblock your IP address: {{.UserIP}}.",
    // "userIpRequestHeaders": ["CF-Connecting-IP"],
  },
```

CleanShot 2023-12-11 at 09 18 38


michaellzc:
* main
  * **PR #58877** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/58877?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/58877?utm_source=stack-comment).

cmd/frontend/internal/cli/http.go
```diff
@@ -100,14 +106,14 @@ func newExternalHTTPHandler(
...
-	sm.Handle("/", secureHeadersMiddleware(appHandler, crossOriginPolicyNever))
+	sm.Handle("/.api/", ipAllowlistMiddleware.Handle(secureHeadersMiddleware(apiHandler, crossOriginPolicyAPI)))
+	sm.Handle("/.executors/", ipAllowlistMiddleware.Handle(secureHeadersMiddleware(executorProxyHandler, crossOriginPolicyNever)))
+	sm.Handle("/", ipAllowlistMiddleware.Handle(secureHeadersMiddleware(appHandler, crossOriginPolicyNever)))
 	const urlPathPrefix = "/.assets"
 	// The asset handler should be wrapped into a middleware that enables cross-origin requests
 	// to allow the loading of the Phabricator native extension assets.
 	assetHandler := assetsutil.NewAssetHandler(sm)
-	sm.Handle(urlPathPrefix+"/", http.StripPrefix(urlPathPrefix, secureHeadersMiddleware(assetHandler, crossOriginPolicyAssets)))
+	sm.Handle(urlPathPrefix+"/", ipAllowlistMiddleware.Handle(http.StripPrefix(urlPathPrefix, secureHeadersMiddleware(assetHandler, crossOriginPolicyAssets))))
```
michaellzc, line 116: this wraps the ip allowlist at the outest layer, so it's always run first before even attempting anything else

cmd/frontend/auth/ipallowlist/checker/checker.go
```diff
@@ -0,0 +1,242 @@
...
+	//  - Cloud needs to configure IP allowlist that are managed by Cloud infra, and we don't user to override it, e.g., gke pod cidr, executors ip range.
+	stableIpAllowlist = func() []string {
+		if allowlist := os.Getenv("AUTH_ALLOWED_IP_ADDRESS_STABLE_IP_ALLOWLIST"); allowlist != "" {
+			addr := strings.Split(allowlist, ",")
+			for i := range addr {
+				addr[i] = strings.TrimSpace(addr[i])
+			}
+			return addr
+		}
+		return nil
```
evict, line 35: You could use a `map[string]struct{}`, then you don't have to iterate over all the list for existence checks.

cmd/frontend/auth/ipallowlist/checker/checker.go
```diff
@@ -0,0 +1,242 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "parse remote addr")
+	}
+	for _, header := range c.userHeaders {
+		ip := req.Header.Get(header)
+		ips := strings.Split(ip, ",")
+		if len(ips) > 0 {
+			userIp = strings.TrimSpace(ips[0])
+			break
+		}
```
evict, line 144: This can be spoofed if the using the `X-Forwarded-For` header with a Google Loadbalancer: https://cloud.google.com/load-balancing/docs/https/#x-forwarded-for_header.  [The sample applies for](https://developers.cloudflare.com/fundamentals/reference/http-request-headers/#x-forwarded-for) CF as well.

evict (changes requested):
I like the idea, but there are some concerns around IP spoofing with some of the supplied headers. We could add a trusted proxy/lb field, so we know after which IP we can trust the client IP? I also would love to see some tests for the new behaviour.

cmd/frontend/auth/ipallowlist/checker/checker.go
michaellzc: yeah, that's a known limitation as I documented in the code and user-facing docs.

the admin is responsible for ensuring the header are trusted. For example:

user -> cloudflare proxy -> application load balancer -> sourcegraph frontend svc

the admin needs to ensure the ALB only accepts connection from cloudflare edge servers, and use [CF-Connecting-IP](https://developers.cloudflare.com/fundamentals/reference/http-request-headers/#cf-connecting-ip) to infer the real user IP.

cmd/frontend/auth/ipallowlist/checker/checker.go
```diff
@@ -0,0 +1,269 @@
...
+	enabled atomic.Bool
+
+	authorizedIPs      []net.IP
+	authorizedRanges   []net.IPNet
+	authorizedSourceMu sync.RWMutex
+
+	userHeaders   []string
+	userHeadersMu sync.RWMutex
+
+	errorMessageTmpl *template.Template
```
keegancsmith, line 54: You could simplify a lot of the logic you have here around synchronization by flipping how you think of this from a bunch of little fields, to something that is computed from state (eg config, envvars)  and you just do a single atomic load to get the latest version of it.

For example right now you don't have any protection for reads on errorMessageTmpl even though it can be written to concurrently.

A useful primitive we have for this is the conf.Cached function.

michaellzc:
> I like the idea, but there are some concerns around IP spoofing with some of the supplied headers. We could add a trusted proxy/lb field, so we know after which IP we can trust the client IP? I also would love to see some tests for the new behaviour.

added some tests.

we can add `remoteAddr[]` for addition check when `userHeaders` is in used

keegancsmith (commented):
Are there any other systems which make requests into sourcegraph? The only thing that comes to mind is webhooks. Maybe worth mentioning in the docs.

Additionally all the code you are adding is only used by cmd/frontend. Feels like it should all live under frontend/internal.

I just did a quick pass by review which is why I didn't approve, can do a deeper review if ya like.

cmd/frontend/auth/ipallowlist/checker/checker.go
michaellzc: nice, thank you. the code is much better now.

doc/admin/auth/index.md
```diff
@@ -532,6 +533,42 @@ Some proxies add a prefix to the username header value. For example, Google IAP
 }
 ```
 
+## IP Allow List
```
bobheadxi, line 536: nit - I think allowlist is generally one word? you use it as such in most of your docstrings as well

cmd/frontend/internal/auth/ipallowlist/checker/checker.go
```diff
@@ -0,0 +1,308 @@
...
+
+	c := &Checker{
+		logger: logger,
+	}
+
+	if cfg := getConfig(); cfg != nil {
+		if cfg.err != nil {
+			return nil, errors.Wrap(cfg.err, "parse config")
+		}
+	}
```
bobheadxi, line 97: are we sure we want to error out at startup? It looks like `Checker` is mostly stateless, and resolves the most recent configuration from the computed `configResolver` every time we authorize a request, so if an error occurs it'll show up in logs there. Preventing startup on a misconfiguration in site config seems difficult to deal with

cmd/frontend/internal/auth/ipallowlist/checker/checker.go
```diff
@@ -0,0 +1,308 @@
...
+	authorizedUserRanges   []net.IPNet
+	userHeaders            []string
+	errorMessageTmpl       *template.Template
+
+	// err indicates error occurred while parsing the configuration.
+	err error
+}
+
+// New returns a new Checker. You must call Start() on the returned Checker
+// to subscribe to configuration changes in the background.
```
bobheadxi, line 80: I think there's no longer a `Start()`:

```suggestion
// New returns a new Checker.
```

schema/site.schema.json
```diff
@@ -1136,6 +1136,59 @@
...
+          "examples": ["100.100.100.0/25", "23.34.56.21"]
+        },
+        "trustedClientIpAddress": {
+          "description": "List of trusted client IP addresses that will bypass user IP address check. If empty, nothing can be bypass. This is useful to support access from trusted internal services. It will always permit connection from `127.0.0.1`. You must include the IP range allocated for the Sourcegraph deployment services to allow inter-serivice communication, e.g., kubernetes pod ip range.",
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "examples": ["100.100.100.0/25", "23.34.56.21"]
+        },
```
michaellzc, line 1163: added the concept of `trustedClientIpAddress`. if `Remote.Addr` matches one of these, any further check will be bypass.

cc @evict

michaellzc:
> Are there any other systems which make requests into sourcegraph? The only thing that comes to mind is webhooks. Maybe worth mentioning in the docs.
> 
> Additionally all the code you are adding is only used by cmd/frontend. Feels like it should all live under frontend/internal.
> 
> I just did a quick pass by review which is why I didn't approve, can do a deeper review if ya like.

thank you. it should be good for a proper review now.

cmd/frontend/internal/cli/http.go
```diff
@@ -122,6 +128,8 @@ func newExternalHTTPHandler(
 	h = internalauth.ForbidAllRequestsMiddleware(h)
 	h = tracepkg.HTTPMiddleware(logger, h, conf.DefaultClient())
 	h = instrumentation.HTTPMiddleware("external", h)
+	// 🚨 SECURITY: ip allowlist must be the first middleware to run to avoid doing unnecessary things
+	h = ipAllowlistMiddleware.Handle(h)
```
michaellzc, line 132: @evict this ensures we always run ip allowlist first before touching anything else

evict (approved):
Nice work and a nice feature to have! 🚀

filiphaftek (approved):
Very nice feature @michaellzc ! My only concern is how to highlight, that enabling this feature without properly setting `trustedClientIpAddress` could break instance - would this make sense to enforce having `trustedClientIpAddress` even if empty? So admin is aware that their disabled this feature on purpose.

schema/schema.go
```diff
@@ -106,6 +106,22 @@ type AuthAccessTokens struct {
...
+	ErrorMessageTemplate string `json:"errorMessageTemplate,omitempty"`
+	// TrustedClientIpAddress description: List of trusted client IP addresses that will bypass user IP address check. If empty, nothing can be bypass. This is useful to support access from trusted internal services. It will always permit connection from `127.0.0.1`. You must include the IP range allocated for the Sourcegraph deployment services to allow inter-serivice communication, e.g., kubernetes pod ip range.
+	TrustedClientIpAddress []string `json:"trustedClientIpAddress,omitempty"`
+	// UserIpAddress description: List of user IP addresses to allow. If empty, all IP addresses are allowed.
+	UserIpAddress []string `json:"userIpAddress,omitempty"`
+	// UserIpRequestHeaders description: An optional list of case-insensitive request header names to use for resolving the callers user IP address. You must ensure that the header is coming from a trusted source. If the header contains multiple IP addresses, the right-most is used. If no IP is found from provided headers, the connected client IP address is used.
+	UserIpRequestHeaders []string `json:"userIpRequestHeaders,omitempty"`
```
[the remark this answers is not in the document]
michaellzc: ClientIp only cares about connect (like TCP, so no HTTP header), UserIp could be either client IP or header from a trusted header.

So, fs matches TrustedClientIpAddress, userIp check will be bypass

michaellzc:
> Very nice feature @michaellzc ! My only concern is how to highlight, that enabling this feature without properly setting `trustedClientIpAddress` could break instance - would this make sense to enforce having `trustedClientIpAddress` even if empty? So admin is aware that their disabled this feature on purpose.

I am hesitant to do so, cuz it may not be needed depending on the deployment type. I think our docs should cover this case. Also, they can always access the instance from localhost.

daxmc99:
> This will break inter-pod communication
Looking at https://stackoverflow.com/questions/61847788/inquiring-pod-and-service-subnets-from-inside-kubernetes-cluster it also appears there is no easy solution here. 

That is my biggest concern for end-users. A site admin is not necessarily a DevOps or network engineer. Not blocking but something where it would be nice to have some tooltip inform the cluster of current IP address of the pod so they consider ensuring local comms still work. 

Lastly, we could request the full list of endpoints in our namespace and automatically add them as whitelisted (might be too much for this but a possibility) 
We have done this [before](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/internal/endpoint/k8s.go?L37)

michaellzc:
> That is my biggest concern for end-users. A site admin is not necessarily a DevOps or network engineer. Not blocking but something where it would be nice to have some tooltip inform the user of current IP address of the pod so they consider ensuring local comms still work.

local comm will always work since `127.0.0.1` is added to trusted client IP, which bypass any further check.

> Looking at https://stackoverflow.com/questions/61847788/inquiring-pod-and-service-subnets-from-inside-kubernetes-cluster it also appears there is no easy solution here.

> Lastly, we could request the full list of endpoints in our namespace and automatically add them as whitelisted (might be too much for this but a possibility) We have done this [before](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/internal/endpoint/k8s.go?L37)

yeah, this will require pulling all endpoints (they are ephemeral), and update the list on the fly. that feels like a lot of work, and customers may not like RBAC.

Also, this approach only works for k8s. If customer is on docker-compose, we're out of luck.

pjlast (approved):

evict:
> > That is my biggest concern for end-users. A site admin is not necessarily a DevOps or network engineer. Not blocking but something where it would be nice to have some tooltip inform the user of current IP address of the pod so they consider ensuring local comms still work.
> 
> local comm will always work since `127.0.0.1` is added to trusted client IP, which bypass any further check.
> 
> > Looking at https://stackoverflow.com/questions/61847788/inquiring-pod-and-service-subnets-from-inside-kubernetes-cluster it also appears there is no easy solution here.
> 
> > Lastly, we could request the full list of endpoints in our namespace and automatically add them as whitelisted (might be too much for this but a possibility) We have done this [before](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/internal/endpoint/k8s.go?L37)
> 
> yeah, this will require pulling all endpoints (they are ephemeral), and update the list on the fly. that feels like a lot of work, and customers may not like RBAC.
> 
> Also, this approach only works for k8s. If customer is on docker-compose, we're out of luck. unless we run another privileged service that has access to the docker daemon, and query from it.
> 
> My take is: a niche use case, and only Cloud will use it. Generally customers are better off just deploying in a private network or do filtering at the LB.



> > That is my biggest concern for end-users. A site admin is not necessarily a DevOps or network engineer. Not blocking but something where it would be nice to have some tooltip inform the user of current IP address of the pod so they consider ensuring local comms still work.
> 
> local comm will always work since `127.0.0.1` is added to trusted client IP, which bypass any further check.
> 
> > Looking at https://stackoverflow.com/questions/61847788/inquiring-pod-and-service-subnets-from-inside-kubernetes-cluster it also appears there is no easy solution here.
> 
> > Lastly, we could request the full list of endpoints in our namespace and automatically add them as whitelisted (might be too much for this but a possibility) We have done this [before](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/internal/endpoint/k8s.go?L37)
> 
> yeah, this will require pulling all endpoints (they are ephemeral), and update the list on the fly. that feels like a lot of work, and customers may not like RBAC.
> 
> Also, this approach only works for k8s. If customer is on docker-compose, we're out of luck. unless we run another privileged service that has access to the docker daemon, and query from it.
> 
> My take is: a niche use case, and only Cloud will use it. Generally customers are better off just deploying in a private network or do filtering at the LB.

@michaellzc you can also opt to use [Gitea's hostmatcher](https://pkg.go.dev/code.gitea.io/gitea@v1.18.0/modules/hostmatcher) for the matching logic. You can then easily allow all private ranges using  `MatchBuiltinPrivate` as a hostlist in an allowlis. It's quite flexible and we're already [using it for external connections](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@cc03635fdd41d5a829e81e7b50ac6492299bb604/-/blob/internal/httpcli/client.go?L443-449).

keegancsmith:
> My take is: a niche use case, and only Cloud will use it. Generally customers are better off just deploying in a private network or do filtering at the LB.

Why can't cloud do the same?

cmd/frontend/internal/auth/ipallowlist/checker/checker.go
```diff
@@ -0,0 +1,364 @@
...
+}
+
+func isAuthorized(cfg config, req *http.Request) error {
+	clientIp, userIp, err := getIP(req, cfg.userHeaders)
+	if err != nil {
+		return errors.Wrap(err, "get user ip")
+	}
+
+	// trusted client ip is allowed to bypass further checks
+	if ok := containsIP(clientIp, cfg.trustedClientIps, cfg.trustedClientRanges); ok {
```
keegancsmith, line 155: very minor: I think its clearer to not use the indirection of the ok variable here, since you can more easily "read" what the conditional is saying. eg if containsIP(... and below if !containsIP(...

cmd/frontend/internal/auth/ipallowlist/checker/checker.go
```diff
@@ -0,0 +1,364 @@
...
+	if tmpl == nil {
+		return errors.Newf(msg)
+	}
+
+	var buf bytes.Buffer
+	if err := tmpl.Execute(&buf, unauthorizedErrorContext{
+		Error:  msg,
+		UserIP: ip.String(),
+	}); err != nil {
+		return errors.Wrap(err, "execute error message template")
```
keegancsmith, line 181: if we end up in this situation, its useful to also include the original `msg`.

cmd/frontend/internal/auth/ipallowlist/checker/checker.go
```diff
@@ -0,0 +1,364 @@
...
+		return renderUnAuthorizedErrorMessage(userIp, cfg.errorMessageTmpl)
+	}
+
+	return nil
+}
+
+func renderUnAuthorizedErrorMessage(ip net.IP, tmpl *template.Template) error {
+	msg := fmt.Sprintf("You are not allowed to access this Sourcegraph instance: %q", ip.String())
+	if tmpl == nil {
+		return errors.Newf(msg)
```
keegancsmith, line 173: Newf should only have format strings as the first argument, if msg contains format strings in it you will end up with a formatting error in your string. Same applies below when you pass in buf.String(). I think you can just do errors.New

cmd/frontend/internal/auth/ipallowlist/checker/checker.go
```diff
@@ -0,0 +1,364 @@
...
+	clientIps, clientIpRanges, err := parseIPs(append(cfg.ClientIpAddress, stableClientIpAllowlist...))
+	if err != nil {
+		errs = errors.Append(errs, err)
+	}
+
+	// 🚨 SECURITY: we always allow localhost to access the instance, otherwise liveness prob will fail
+	trustedClientIps, trustedClientIpRanges, err := parseIPs(concactSlices(cfg.TrustedClientIpAddress, stableTrustedClientIpAllowlist, defaultTrustedClientIpAllowlist))
+	if err != nil {
+		errs = errors.Append(errs, err)
+	}
```
keegancsmith, line 298: if we fail to parse TrustedClientIpAddress then we will end up with a config without localhost. You should append in localhost independently of parsing user input.

cmd/frontend/internal/auth/ipallowlist/checker/checker.go
```diff
@@ -0,0 +1,364 @@
...
+			return addr
+		}
+		return nil
+	}()
+	// userIpRequestHeaders overrides .auth.allowedIpAddress.userIpRequestHeaders from site config.
+	//
+	// Use case:
+	// 	- Cloud needs to manager this header to ensure the user ip is reading from a trusted header.
+	userIpRequestHeaders = func() []string {
+		if v := os.Getenv("AUTH_ALLOWED_IP_ADDRESS_USER_IP_REQUEST_HEADERS"); v != "" {
```
keegancsmith, line 62: you should make it clearer in the name that this is an override given it behaves differently to the rest.

andreeleuterio:
@keegancsmith Cloud was doing this on the Cloudflare level but a recent Cloudflare migration pulled the rug from under us and we can no longer use terraform to manage these allowlists. They changed their APIs so that we can no longer manage multiple Cloudflare rules in separate tf namespaces. The GCP LB was also restrictive in its options and did not allow for us to have as many CIDR ranges as we needed in the rules.

cmd/frontend/internal/auth/ipallowlist/checker/checker.go
```diff
@@ -0,0 +1,364 @@
...
+	}()
+	defaultTrustedClientIpAllowlist = []string{"127.0.0.1"}
+
+	// configResolver is a cached resolver for the current configuration.
+	// If empty, implies it is disabled.
+	//
+	// It will attempt to parse all config with best effort, and
+	// attach a multi-error in the struct field `err` if any error occurred.
+	// The caller should decide what to do with the error.
+	configResolver = conf.Cached(getConfig)
```
keegancsmith, line 78: If we only expect cloud to use this, will they ever configure it via the site config? Or will they only use environment variables?

If the latter, then I would consider only exposing this as envvars and removing the logic around site configuration. That would simplify this a bunch and avoids another feature we need to support that customers may use.

cmd/frontend/internal/auth/ipallowlist/checker/checker.go
keegancsmith, line 155: there is coupling here which may bite us down the road. If trustedClient* is empty this will return true, and we will never do the other checks. The coupling is that it will never be empty due to use always having localhost. But that feels like the sort of change we might accidently make.

You could adjust the signature of the function to explicitly encode the empty return. Or maybe adjust the name, since this isn't really containsIP, but containsIPOrEmpty

cmd/frontend/internal/auth/ipallowlist/ipallowlist.go
```diff
@@ -0,0 +1,64 @@
...
+		if err := m.checker.IsAuthorized(r); err != nil {
+			w.WriteHeader(http.StatusForbidden)
+			data := &pageError{
+				StatusCode: http.StatusForbidden,
+				StatusText: http.StatusText(http.StatusForbidden),
+				Error:      err.Error(),
+			}
+			var buf bytes.Buffer
+			if err := errorHTMLTmpl.Execute(&buf, data); err != nil {
+				w.Write([]byte(err.Error()))
```
keegancsmith, line 48: you need to return here

cmd/frontend/internal/auth/ipallowlist/checker/checker.go
```diff
@@ -0,0 +1,364 @@
...
+}
+
+// New returns a new Checker.
+func New(logger log.Logger) (*Checker, error) {
+	logger = logger.Scoped("ipAllowlistChecker")
+
+	cfg := conf.Get()
+	if cfg == nil {
+		return nil, errors.New("no config available")
+	}
```
keegancsmith, line 113: unnecessary to check this. We never check if this is nil, it is implied it is always set.

cmd/frontend/internal/auth/ipallowlist/checker/checker.go
```diff
@@ -0,0 +1,364 @@
...
+
+	if cfg := getConfig(); cfg != nil {
+		if cfg.err != nil {
+			logger.Error("site config 'auth.allowedIpAddress' contains error, please resolve it.", log.Error(cfg.err))
+		}
+	}
```
keegancsmith, line 119: is it worth failing the startup of frontend if misconfigured? It looks like the decision was made to instead just log. I think you can likely update New to not return an error, which simplifies handling at the callsites for it and ipallowlist.

keegancsmith (approved):
LGTM. A bunch of mostly minor feedback, but some may make you delete a bunch of code :)

keegancsmith:
> @keegancsmith Cloud was doing this on the Cloudflare level but a recent Cloudflare migration pulled the rug from under us and we can no longer use terraform to manage these allowlists. They changed their APIs so that we can no longer manage multiple Cloudflare rules in separate tf namespaces. The GCP LB was also restrictive in its options and did not allow for us to have as many CIDR ranges as we needed in the rules.

To play the devils advocate then, why not configure in the kubernetes ingress? I assuming that is likely just nginx under the hood :)

michaellzc:
> > @keegancsmith Cloud was doing this on the Cloudflare level but a recent Cloudflare migration pulled the rug from under us and we can no longer use terraform to manage these allowlists. They changed their APIs so that we can no longer manage multiple Cloudflare rules in separate tf namespaces. The GCP LB was also restrictive in its options and did not allow for us to have as many CIDR ranges as we needed in the rules.
> 
> To play the devils advocate then, why not configure in the kubernetes ingress? I assuming that is likely just nginx under the hood :)

we don't run our own ingress (e.g., nginx controller) on Cloud, but use GCP external application load balancer. The filtering rule syntax available (Cloud Armor, GCP's WAF solution) for GCP ALB does not work with our current setup.

Below is the lifecycle of a user accessing frontend on a Cloud instance:

users -> Cloudflare proxy -> Cloud Armor -> GCP ALB -> frontend

Cloud Armor is configured to prevent origin bypass attacks, i.e., only accept connections from Cloudflare proxy. 

When IP allowlist is enabled, we create rules in Cloudflare that are evaluated in the Cloudflare proxy. Like what @andreeleuterio described, this path no longer works.

> Cloudflare proxy adds `CF-Connecting-IP` header to all proxied requests to reflect real user IP.

Cloud Armor rule syntax is impossible to filter on connection IP and user IP (from headers) simultaneously. This could change if we move the WAF capability to Cloud Armor from Cloudflare, but we are not dropping it anytime soon.

users -> Cloud Armor -> GCP ALB -> frontend

cmd/frontend/internal/auth/ipallowlist/checker/checker.go
```diff
@@ -0,0 +1,364 @@
...
+// Checker implements IP allowlist checking and subscription to configuration changes.
+type Checker struct {
+	logger log.Logger
+}
+
+// config is the computed internal config from site config.
+// use configResolver() to get the current config.
+type config struct {
+	authorizedClientIps    []net.IP
+	authorizedClientRanges []net.IPNet
```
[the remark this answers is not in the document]
michaellzc: for ipv4, we can just append `/32`. I am not sure if ipv6 works the same tho, without digging into another rabbit hole, I'll keep it the way it is.

cmd/frontend/internal/auth/ipallowlist/checker/checker.go
michaellzc: on Cloud, Cloud will set some stable IP allowlist via env var to ensure internal comm still works (this will prevent user from accidentally removing internal IP addresses from site config)

then customers will manage their own IP allowlist from the site config.

cmd/frontend/internal/auth/ipallowlist/checker/checker.go
michaellzc: I decided to offload this check to the caller, and explicty check if empty. more code but less gotchat haha

cmd/frontend/internal/auth/ipallowlist/checker/checker.go
michaellzc: ended up returnning a multi error

cmd/frontend/internal/auth/ipallowlist/checker/checker.go
michaellzc: fixed

cmd/frontend/internal/auth/ipallowlist/checker/checker.go
```diff
@@ -0,0 +1,364 @@
...
+	"github.com/sourcegraph/sourcegraph/internal/conf/conftypes"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+	"github.com/sourcegraph/sourcegraph/schema"
+)
+
+var (
+	// stableUserIpAllowlist will be merged with the user ip allowlist from site config.
+	stableUserIpAllowlist = func() []string {
+		if allowlist := os.Getenv("AUTH_ALLOWED_IP_ADDRESS_STABLE_USER_IP_ALLOWLIST"); allowlist != "" {
+			addr := strings.Split(allowlist, ",")
```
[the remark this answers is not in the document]
michaellzc: whitespace separated feels kind of weird to me, maybe just I've never used it before.

took a glance of the usage in` sourcegra/sourcegraph`, it seems like we have a mix use of whitesapce and comma separated config as a string, but the majority are still comma separated.

how about adding a new helper to `internal/env`?

```go
func GetSliceFunc[V any](name string, defaultValues []V, description string, parse func(v string) (V, error)) ([]V, error)
```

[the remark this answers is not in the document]
keegancsmith: Only issue is how will you handle errors here when this is running at init time? The env package normally calls these helpers Must*, so panics if it runs into an error.

Alternatively you can just have a simple unexported helper in this package which does what you want and isn't as generic:

```go
func getEnvCommaSeparated(name string) []string {
  vals := strings.Split(os.Getenv(name), ",")
  filtered := vals[:0]
  for _, s := range vals {
    s = strings.TrimSpace(s)
    if s != "" {
      filtered = append(filtered, s)
    }
  }
  return filtered
}
```

cmd/frontend/internal/auth/ipallowlist/checker/checker.go
michaellzc: fixed

cmd/frontend/internal/auth/ipallowlist/checker/checker.go
```diff
@@ -0,0 +1,364 @@
...
+	"github.com/sourcegraph/sourcegraph/internal/conf"
+	"github.com/sourcegraph/sourcegraph/internal/conf/conftypes"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+	"github.com/sourcegraph/sourcegraph/schema"
+)
+
+var (
+	// stableUserIpAllowlist will be merged with the user ip allowlist from site config.
+	stableUserIpAllowlist = func() []string {
+		if allowlist := os.Getenv("AUTH_ALLOWED_IP_ADDRESS_STABLE_USER_IP_ALLOWLIST"); allowlist != "" {
```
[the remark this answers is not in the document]
michaellzc: fixed, move to config builder method

PR #47748, opened 2023-02-16
codeintel: Expose inference limit errors in GraphQL

efritz:
Pulled from https://github.com/sourcegraph/sourcegraph/pull/47287. This PR captures and exposes errors of type `inference.LimitError` as a non-critical error. This will be useful for dashboards in the future, where we don't want to crash on over-limit (but we do want to display to the user the reason).

## Test plan

Existing CI.

PR #58879, opened 2023-12-11
Surfaces errors in autoindex configuration

### LINKED ISSUE #58453: Nil pointer exception on index configuration page
https://sourcegraph.com/github.com/vitejs/vite/-/code-graph/index-configuration

Cannot tell what is specific to that repo, have not been able to reproduce it
with other repos on Sourcegraph.com. I have not yet attempted to reproduce it locally.

Reported by @Strum355 in Slack: https://sourcegraph.slack.com/archives/CHXHX7XAS/p1699625070804179

kritzcreek:
Fixes #58453

## Test plan

Add the `vitejs/vite` repo to your local sg instance, and view the autoindex configuration page.

After this PR you should see the real error: 
`Error fetching index configuration: Inference limit: requested content for more than 100 (100) files`

instead of a nil pointer exception.

## Implementation notes

I added two commits, but the first is completely overridden by the second. In 67ed1c33e3c589b0a067a5f08be05116eb98a180 I tried to 'interpret' what the code was trying to do in the first place, and return a 'succesful' result even if the evaluation ran into a limit error. That produces very confusing UX though, because now it looks like everything is fine on the configuration page, but there are no build jobs. I think it's better to surface the error in the UI here.

This change leaves things in a bit of an awkward spot. The assignment to `limitErr` is dead code. I don't think the special handling for limit errors make sense here, and it doesn't work anyway (because somewhere on the code path it's not treated different from other errors). I tried ripping it out, but my go-foo isn't quite there yet. I could either open an issue for the follow-up cleanup, or we could try to pair on it @varungandhi-src WDYT?

### DIFF
--- a/internal/codeintel/autoindexing/transport/graphql/root_resolver_configuration_repository.go
+++ b/internal/codeintel/autoindexing/transport/graphql/root_resolver_configuration_repository.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"errors"
 
 	"github.com/graph-gophers/graphql-go"
 	"go.opentelemetry.io/otel/attribute"
@@ -14,6 +13,7 @@ import (
 	sharedresolvers "github.com/sourcegraph/sourcegraph/internal/codeintel/shared/resolvers"
 	"github.com/sourcegraph/sourcegraph/internal/observation"
 	"github.com/sourcegraph/sourcegraph/lib/codeintel/autoindex/config"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
 	"github.com/sourcegraph/sourcegraph/lib/pointers"
 )
 
--- a/internal/codeintel/autoindexing/transport/graphql/root_resolver_configuration_repository.go
+++ b/internal/codeintel/autoindexing/transport/graphql/root_resolver_configuration_repository.go
@@ -114,9 +114,20 @@ func (r *indexConfigurationResolver) InferredConfiguration(ctx context.Context) 
 		}
 	}
 
-	marshaled, err := config.MarshalJSON(config.IndexConfiguration{IndexJobs: result.IndexJobs})
+	var marshaled []byte
 	if err != nil {
-		return nil, err
+		emptyJobs := []config.IndexJob{}
+		json, err := config.MarshalJSON(config.IndexConfiguration{IndexJobs: emptyJobs})
+		if err != nil {
+			return nil, err
+		}
+		marshaled = json
+	} else {
+		json, err := config.MarshalJSON(config.IndexConfiguration{IndexJobs: result.IndexJobs})
+		if err != nil {
+			return nil, err
+		}
+		marshaled = json
 	}
 
 	var indented bytes.Buffer
--- a/internal/codeintel/autoindexing/transport/graphql/root_resolver_configuration_repository.go
+++ b/internal/codeintel/autoindexing/transport/graphql/root_resolver_configuration_repository.go
@@ -8,12 +8,10 @@ import (
 	"github.com/graph-gophers/graphql-go"
 	"go.opentelemetry.io/otel/attribute"
 
-	"github.com/sourcegraph/sourcegraph/internal/codeintel/autoindexing/internal/inference"
 	resolverstubs "github.com/sourcegraph/sourcegraph/internal/codeintel/resolvers"
 	sharedresolvers "github.com/sourcegraph/sourcegraph/internal/codeintel/shared/resolvers"
 	"github.com/sourcegraph/sourcegraph/internal/observation"
 	"github.com/sourcegraph/sourcegraph/lib/codeintel/autoindex/config"
-	"github.com/sourcegraph/sourcegraph/lib/errors"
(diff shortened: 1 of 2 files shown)


internal/codeintel/autoindexing/transport/graphql/root_resolver_configuration_repository.go
```diff
@@ -107,11 +105,7 @@ func (r *indexConfigurationResolver) InferredConfiguration(ctx context.Context)
 	var limitErr error
 	result, err := r.autoindexSvc.InferIndexConfiguration(ctx, r.repositoryID, "", "", true)
 	if err != nil {
-		if errors.As(err, &inference.LimitError{}) {
-			limitErr = err
-		} else {
-			return nil, err
-		}
+		return nil, err
```
varungandhi-src, line 108: This error was meant to be exposed in the GraphQL API as per https://github.com/sourcegraph/sourcegraph/pull/47748 where it was introduced. Removing this assignment here would stop that from working.

kritzcreek: Yes, exactly. That's what I was describing in the PR description.

With the code as is it never made it to the GraphQL API, as 

[result, err := r.autoindexSvc.InferIndexConfiguration(ctx, r.repositoryID, "", "", true)](https://github.com/sourcegraph/sourcegraph/pull/58879/files#diff-3b17be7913c30a2cdff0dac0b827d578f6a1db18acaa783cf6be82e2dc8cb47cL108)

always returns `nil` if `err` is set, even when it is set to LimitErr.

Repeating from the PR description: l wrote 67ed1c33e3c589b0a067a5f08be05116eb98a180 to show a change that would actually let limitErr make it to the GraphQL API, but the UI doesn't seem to handle that case at all and just produce a bad UX.

varungandhi-src: I think the problem needs to be fixed in the frontend where we should pattern match against this error and show a more useful message.

kritzcreek: I don't think the frontend can do much better than show the error as is with this PR:

image

If we actually wanted to make the error message more useful we'd need to include what files the user tried to read that led to this error.

varungandhi-src: See these two GraphQL queries:
- https://sourcegraph.com/github.com/sourcegraph/sourcegraph@ea79dcfb62e9c60bf062dfbcd2573a81dc442c43/-/blob/client/web/src/enterprise/codeintel/configuration/backend.ts?L62-82
- https://sourcegraph.com/github.com/sourcegraph/sourcegraph@ea79dcfb62e9c60bf062dfbcd2573a81dc442c43/-/blob/client/web/src/enterprise/codeintel/configuration/backend.ts?L84-102

They're accessing the `configuration` and `parsedConfiguration` fields, but not checking the `limitError` field.

I'm not 100% sure if the idiomatic thing (wrt GraphQL) is to return a non-error with the limit error inside the payload... That seems off, but that's what the end of this function is doing currently. I think we should change that to return `limitErr` instead of `nil`.

internal/codeintel/autoindexing/transport/graphql/root_resolver_configuration_repository.go
```diff
@@ -104,29 +104,25 @@ func (r *indexConfigurationResolver) Configuration(ctx context.Context) (_ *stri
...
 	result, err := r.autoindexSvc.InferIndexConfiguration(ctx, r.repositoryID, "", "", true)
+	resolver := &inferredConfigurationResolver{siteAdminChecker: r.siteAdminChecker}
 	if err != nil {
 		if errors.As(err, &inference.LimitError{}) {
-			limitErr = err
-		} else {
-			return nil, err
+			resolver.limitErr = err
 		}
+		return resolver, err
```
varungandhi-src, line 113: @kritzcreek what do you think about this fix instead? The main difference is that we're also setting the `limitErr` field as defined in the API contract.

internal/codeintel/autoindexing/transport/graphql/root_resolver_configuration_repository.go
varungandhi-src: I've filed https://github.com/sourcegraph/sourcegraph/issues/58885 for improving the UX here.

internal/codeintel/autoindexing/transport/graphql/root_resolver_configuration_repository.go
kritzcreek: I see what you're saying now. This would return partial data _and_ the error from the GraphQL request, right?

varungandhi-src: Yeah, I think we should at least attempt to respect the API contract. Whether the API is a good one/one we want to tweak/deprecate is a separate question.

varungandhi-src (approved):
Can you cross-check that my suggested fix still works? If so, we can merge the PR.

PR #58886, opened 2023-12-11
Construct telemetryRecorder within existing eventLogger infrastructure

nathan-downs:
## Test plan

Ensure existing test continue to pass.

run sg start
verify locally logged events in event_logs

### DIFF
--- a/client/web/src/tracking/eventLogger.ts
+++ b/client/web/src/tracking/eventLogger.ts
@@ -5,6 +5,7 @@ import * as uuid from 'uuid'
 import { isErrorLike, isFirefox, logger } from '@sourcegraph/common'
 import type { SharedEventLogger } from '@sourcegraph/shared/src/api/sharedEventLogger'
 import { EventClient } from '@sourcegraph/shared/src/graphql-operations'
+import { noOpTelemetryRecorder } from '@sourcegraph/shared/src/telemetry'
 import type { TelemetryService } from '@sourcegraph/shared/src/telemetry/telemetryService'
 import type { UTMMarker } from '@sourcegraph/shared/src/tracking/utm'
 
@@ -221,6 +222,8 @@ export class EventLogger implements TelemetryService, SharedEventLogger {
  * src/telemetry instead.
  */
 export const eventLogger = new EventLogger()
+
+export const telemetryRecorder = noOpTelemetryRecorder
 
 /**
  * Log events associated with URL query string parameters, and remove those parameters as necessary
--- a/client/web/src/site-admin/UserManagement/index.tsx
+++ b/client/web/src/site-admin/UserManagement/index.tsx
@@ -6,7 +6,7 @@ import { useQuery } from '@sourcegraph/http-client'
 import { H1, Card, Text, Icon, Button, Link, Alert, LoadingSpinner, AnchorLink } from '@sourcegraph/wildcard'
 
 import type { UsersManagementSummaryResult, UsersManagementSummaryVariables } from '../../graphql-operations'
-import { eventLogger } from '../../tracking/eventLogger'
+import { eventLogger, telemetryRecorder } from '../../tracking/eventLogger'
 import { checkRequestAccessAllowed } from '../../util/checkRequestAccessAllowed'
 import { ValueLegendList, type ValueLegendListProps } from '../analytics/components/ValueLegendList'
 
@@ -30,6 +30,7 @@ export const UsersManagement: React.FunctionComponent<UsersManagementProps> = ({
 }) => {
     useEffect(() => {
         eventLogger.logPageView('UsersManagement')
+        telemetryRecorder.recordEvent('UserManagement', 'viewed')
     }, [])
 
     const { data, error, loading, refetch } = useQuery<UsersManagementSummaryResult, UsersManagementSummaryVariables>(


client/web/src/tracking/eventLogger.ts
```diff
@@ -222,6 +223,8 @@ export class EventLogger implements TelemetryService, SharedEventLogger {
  */
 export const eventLogger = new EventLogger()
 
+export const telemetryRecorder = noOpTelemetryRecorder
```
bobheadxi, line 226: I think we don't need this? It's also deceptively no-op

nathan-downs: Removing the definition here breaks a bunch of Bazel web tests, so it needs some kind of definition here or re-pointing to a more appropriate location.

bobheadxi: But this is not used right? It never changes to a non-noop recorder. If it is used, the reference should be removed and replaced with `noOpTelemetryRecorder` directly, otherwise it _looks_ like a recorder that works when in fact it never works

nathan-downs: You are absolutely correct, replaced the noop with a reference to the object from context.

client/web/src/storm/pages/SearchPage/TryCodySignUpCtaSection.tsx
```diff
@@ -69,9 +69,12 @@ const MeetCodySVG: React.FC = () => (
 export const TryCodySignUpCtaSection: React.FC<TelemetryProps & { className?: string }> = ({
     className,
     telemetryService,
+    telemetryRecorder,
 }) => {
-    const onSignUpClick = (): void =>
+    const onSignUpClick = (): void => {
         telemetryService.log(EventName.CODY_SIGNUP, { type: 'ComHome' }, { type: 'ComHome' })
+        telemetryRecorder.recordEvent(EventName.CODY_SIGNUP, 'clicked')
```
vovakulikov, line 76: Don't we need 'ComHome' property anymore?

client/web/src/nav/UserNavItem.tsx
```diff
@@ -90,10 +91,12 @@ export const UserNavItem: FC<UserNavItemProps> = props => {
 
     const onExperimentalQueryInputChange = useCallback(
         (enabled: boolean) => {
-            telemetryService.log(`SearchInputToggle${enabled ? 'On' : 'Off'}`)
+            // telemetryService.log(`SearchInputToggle${enabled ? 'On' : 'Off'}`)
```
vovakulikov, line 94: Should we still have telemetryService call here?

vovakulikov (commented):
I may missed something but did leave a few comments about some particular events

client/web/src/cody/useCodyChat.tsx
```diff
@@ -174,8 +181,9 @@ export const useCodyChat = ({
                 return
             }
             eventLogger.log(eventLabel, { transcriptId: transcript.id, ...eventProperties })
+            telemetryRecorder.recordEvent(EventName[eventLabel as keyof typeof EventName], 'viewed')
```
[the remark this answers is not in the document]
nathan-downs: This is a little kludgey, but passing `eventLabel` directly violates the type safety check. 
Screenshot 2023-12-12 at 2 14 19 PM

The logic behind this is:
1) User passes a string eventLabel that may or may not contain sensitive data
2) We ensure it exists in the enum object via lookup
3) We can't directly access the elements in the enum with a string, so we have to convert the string literal to a key and then use that as a lookup
Screenshot 2023-12-12 at 2 18 10 PM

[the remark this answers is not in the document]
bobheadxi: Can this just be changed to

```
    const logTranscriptEvent = useCallback(
        (eventLabel: EventName, eventProperties?: { [key: string]: any }) => {
```

That should solve the unknown string issue?

bobheadxi (commented):
FYI I think eventlogger and V1 events calls _must_ be removed entirely before this gets approved for merge. As it stands right now, this PR will double the event volume on an instance (both V1 and V2 events are written to the instance DB), which is not acceptable for on-prem and Cloud

client/web/src/storm/pages/SearchPage/TryCodySignUpCtaSection.tsx
nathan-downs: I can't imagine we do, this is the only event using that name constant, so 'ComHome' isn't adding anything.

client/web/src/nav/UserNavItem.tsx
nathan-downs: Good catch, commented out from testing.

client/vscode/src/webview/search-panel/SearchResultsView.tsx
```diff
@@ -96,6 +96,7 @@ export const SearchResultsView: React.FunctionComponent<React.PropsWithChildren<
     const onExpandAllResultsToggle = useCallback(() => {
         setAllExpanded(oldValue => !oldValue)
         platformContext.telemetryService.log(allExpanded ? 'allResultsExpanded' : 'allResultsCollapsed')
+        platformContext.telemetryRecorder.recordEvent('allResults', `${allExpanded}? 'expanded' : 'collapsed'`)
```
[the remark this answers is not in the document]
nathan-downs: Definitely more simple than what I was doing 😄

client/web/src/cody/useCodyChat.tsx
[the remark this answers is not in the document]
bobheadxi: Also it seems none of the `logTranscriptEvent` callsites provides `eventProperties` - I think you should remove it entirely for now and make it V2-compatible on a need-to-have basis. `transcriptID` should be placed in `privateMetadata`, _or_ `interactionID` https://sourcegraph.com/github.com/sourcegraph/telemetry/-/blob/src/index.ts?L203:1-210:26 though that requires an upgrade of `@sourcegraph/telemetry`

With `eventProperties` removed, this callback only serves to add transcript ID. Universally appending metadata for a particular part of a product is a first-class consideration in `@sourcegraph/telemetry` with `processors`. The right way to do this would be to procure your own `recorder` instance with a  processor that adds the desired metadata, instead of defining helpers that make every callsite talk to telemetry differently:

https://sourcegraph.com/github.com/sourcegraph/telemetry/-/blob/src/index.ts?L162-169

This way, all callsites always talk to a uniform `recorder` interface, and not a bunch of bespoke helpers 😁

PR #58898, opened 2023-12-11
Add HubSpot cookie logging to all relevant syncuser calls

dadlerj:
I just assigned a couple of people who have recently edited these files. Joe and Petri, hope one of you can take a quick scan! 

While I'd like to have this live _by_ the release, this change would only affect dotcom, so there's no need to get it into the RC.

## Test plan

* Low impact change
* SG runs locally
* CI
* Only affects dotcom


cmd/frontend/internal/app/ui/handlers.go
```diff
@@ -574,26 +574,33 @@ func servePingFromSelfHosted(w http.ResponseWriter, r *http.Request) error {
...
-		AnonymousUserID: anonymousUserId,
-		FirstSourceURL:  firstSourceURL,
-		LastSourceURL:   lastSourceURL,
-		HasAgreedToToS:  tosAccepted == "true",
+		IsServerAdmin:          true,
+		AnonymousUserID:        anonymousUserId,
+		FirstSourceURL:         getCookie("sourcegraphSourceUrl"),
+		LastSourceURL:          getCookie("sourcegraphRecentSourceUrl"),
+		OriginalReferrer:       getCookie("originalReferrer"),
+		LastReferrer:           getCookie("Sg_referrer"),
```
akalia25, line 593: should be lowercase`sg_referrer` based on cookie name i'm seeing

akalia25, line 593: for `LastReferrer` should we not be using the cookie `session_referrer`? As that is managed/maintained by us?

cmd/frontend/internal/app/ui/handlers.go
```diff
@@ -574,26 +574,33 @@ func servePingFromSelfHosted(w http.ResponseWriter, r *http.Request) error {
...
+		SignupSessionSourceURL: getCookie("sourcegraphSignupSourceUrl"),
+		SignupSessionReferrer:  getCookie("sourcegraphSignupReferrer"),
```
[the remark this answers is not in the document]
dadlerj: See discussion here: https://docs.google.com/document/d/115B-1AFo09ar3U9M7h3ZrRhxqPuUv7uL0A2ngqr0tI4/edit?disco=AAABB4Sxfrg

internal/auth/userpasswd/handlers.go
```diff
@@ -131,7 +131,31 @@ func handleSignUp(logger log.Logger, db database.DB, eventRecorder *telemetry.Ev
...
-		go hubspotutil.SyncUser(creds.Email, hubspotutil.SignupEventID, &hubspot.ContactProperties{AnonymousUserID: creds.AnonymousUserID, FirstSourceURL: creds.FirstSourceURL, LastSourceURL: creds.LastSourceURL, DatabaseID: usr.ID})
+		getCookie := func(name string) string {
+			c, err := r.Cookie(name)
+			if err != nil || c == nil {
+				return ""
+			}
+			return c.Value
+		}
+
+		go hubspotutil.SyncUser(creds.Email, hubspotutil.SignupEventID, &hubspot.ContactProperties{
```
unknwon, line 142: If you want to go one step further, the `hubspotutil.SyncUser` can take `*http.Request` as the third argument so we don't need to do this heavy struct initialization in multiple places.

unknwon (approved):
Nothing blocking to me

internal/auth/userpasswd/handlers.go
dadlerj: Great call. There are several other callers of `SyncUser`, so I'd probably need to create a separate `SyncUserFromRequest`. In the interest of getting this live, I'm going to go ahead and merge as-is, and I'll follow up on that at a later date.

PR #58902, opened 2023-12-11
Search: fix CSV export with special values

camdencheek:
Per [RFC 4180](https://datatracker.ietf.org/doc/html/rfc4180), CSV fields with commas, quotes, or newlines need to be escaped by quoting the field and replacing double quotes (`"`) with doubled double quotes (`""`). This applies RFC 4180-conformant escaping to every cell in the CSV if it contains one of those special characters. 

## Test plan

I created a test commit with commas and imported it into Google Sheets to ensure it was parsed correctly.

CleanShot 2023-12-11 at 14 09 37@2x


peterguy (approved):
Most excellent. Referencing the CSV RFC is a great touch.

PR #58909, opened 2023-12-12
msp/iam: provison secret access for external secrets

bobheadxi:
The Cloud Run secret ref allows for a secret in the format `projects/{project}/secrets/{secretName}`: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloud_run_v2_service#nested_secret_key_ref - this can be provided directly via the `secretEnv` references.

This change also automatically generates IAM bindings for the specific secrets that gets referenced. This makes migrations, using shared secrets, etc easier as you no longer need to manually copy everything into the service project. Been thinking about this ever since we first tried to migrate the devx-support bot, which used a lot of secrets.

## Test plan

https://github.com/sourcegraph/managed-services/actions/runs/7176248721

Not used yet - will test live in a follow-up.


dev/managedservicesplatform/internal/stack/iam/iam.go
```diff
@@ -161,3 +187,46 @@ func extractImageGoogleProject(image string) *string {
...
 }
+
+type externalSecret struct {
+	key string // original configuration key, lowercased
+
+	projectID string
+	secretID  string
+}
+
+func extractExternalSecrets(secrets map[string]string) ([]externalSecret, error) {
```
jac, line 198: Should this also handle the case where a user specifies a secret with a version e.g. `projects/12345/secrets/secret-name/versions/1`. At the moment the code will return `secret-name/versions/1` as the secretID instead of `secret-name`

bobheadxi: Good idea, updated!

jac (approved):
LGTM!

PR #58913, opened 2023-12-12
feat: add neovim to cody redirects, instead of re-using jetbrains

tjdevries:
## Test plan

- [ ] Test with neovim plugin to make sure the redirect is working (just re-uses the same logic from Jetbrains, so should be alright)


client/web/src/user/settings/accessTokens/UserSettingsCreateAccessTokenCallbackPage.tsx
```diff
@@ -164,7 +176,7 @@ export const UserSettingsCreateAccessTokenCallbackPage: React.FC<Props> = ({
         }
 
         // SECURITY: If the request is coming from JetBrains, verify if the port is valid
```
olafurpg, line 178: Minor
```suggestion
        // SECURITY: If the request is going to redirect to localhost (for JetBrains/Neovim), verify that the port number is at least a valid number
```

olafurpg (approved):

PR #58920, opened 2023-12-12
database: speed up UpdateRepoSizes

mrnugget:
Noticed as part of #wg-cheaper-sourcegraph that the calls to UpdateRepoSizes on sourcegraph.com would never, ever return. Instead the goroutine would get stuck, somehow connection between client and DB would get lost, DB would think client is still there, and the whole thing might have locked rows, and it all became a big disaster.

Digging into this I found that even locally, I could never get the test to update, say, 100k repos with the given batch size. It would time out after 10min.

So what I was to

1. Remove the transaction and the locking -- we update two fields anywhere, that aren't updated by anything else (and in the case of `updated_at` it doesn't matter)
2. Reduce the batch size drastically
3. Add a test to play around with large repo sizes
4. Add a logger

When running the test with 20k repos, I noticed the following:

    Batch size 100: test takes 35s
    Batch size 1000: test takes 64s

Then, with 5k repos, I played with the batch size some more and found:

    Batch size 100: one batch takes ~50ms to update.
    Batch size 500: one batch takes ~650ms to update.

So let's go with 100.

On sourcegraph.com, we right now have 770k repos per shard. That means we end up with 7.700 batches. That should take ~6-7min.

That's still better than one continuous query that never finishes.

## Test plan

- Unit test

### DIFF
--- a/internal/database/gitserver_repos.go
+++ b/internal/database/gitserver_repos.go
@@ -11,9 +11,10 @@ import (
 	"github.com/keegancsmith/sqlf"
 	"github.com/lib/pq"
 
+	"github.com/sourcegraph/log"
+
 	"github.com/sourcegraph/sourcegraph/internal/api"
 	"github.com/sourcegraph/sourcegraph/internal/database/basestore"
-	"github.com/sourcegraph/sourcegraph/internal/database/batch"
 	"github.com/sourcegraph/sourcegraph/internal/database/dbutil"
 	"github.com/sourcegraph/sourcegraph/internal/ratelimit"
 	"github.com/sourcegraph/sourcegraph/internal/types"
@@ -65,7 +66,7 @@ type GitserverRepoStore interface {
 	// counting repos with an associated cloud_default external service.
 	TotalErroredCloudDefaultRepos(ctx context.Context) (int, error)
 	// UpdateRepoSizes sets repo sizes according to input map. Key is repoID, value is repo_size_bytes.
-	UpdateRepoSizes(ctx context.Context, shardID string, repos map[api.RepoName]int64) (int, error)
+	UpdateRepoSizes(ctx context.Context, logger log.Logger, shardID string, repos map[api.RepoName]int64) (int, error)
 	// SetCloningProgress updates a piece of text description from how cloning proceeds.
 	SetCloningProgress(context.Context, api.RepoName, string) error
 	// GetLastSyncOutput returns the last stored output from a repo sync (clone or fetch), or ok: false if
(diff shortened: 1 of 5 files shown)


internal/database/gitserver_repos.go
```diff
@@ -651,44 +652,50 @@ WHERE repo_id = (SELECT id FROM repo WHERE name = %s)
...
-	tx, err := s.Store.Transact(ctx)
-	if err != nil {
-		return 0, err
-	}
-	defer func() { err = tx.Done(err) }()
-
+func (s *gitserverRepoStore) updateRepoSizesWithBatchSize(ctx context.Context, logger log.Logger, repos map[api.RepoName]int64, batchSize int) (updated int, err error) {
 	queries := make([]*sqlf.Query, batchSize)
 
+	logger.Info("Updating repository sizes", log.Int("count", len(repos)), log.Int("batchSize", batchSize))
```
keegancsmith, line 666: should this be info or debug?

worth logging batchSize? its always a constant.

internal/database/gitserver_repos.go
```diff
@@ -651,44 +652,50 @@ WHERE repo_id = (SELECT id FROM repo WHERE name = %s)
...
 			rowsAffected, err := res.RowsAffected()
 			if err != nil {
-				return 0, err
+				logger.Info("Failed to read updated rows", log.Error(err))
+				return updatedRows, err
 			}
 			updatedRows += int(rowsAffected)
 
+			duration := time.Since(start)
+			logger.Info("Done", log.Duration("duration", duration), log.Int("rowsAffected", int(rowsAffected)))
```
keegancsmith, line 697: Done message will be confusing without more context. Maybe this logger needs to be a scoped logger?

internal/database/gitserver_repos_test.go
```diff
@@ -1121,6 +1121,44 @@ func TestGitserverUpdateRepoSizes(t *testing.T) {
 	}
 }
 
+func TestGitserverUpdateRepoSizes_LargeAmountOfRepos(t *testing.T) {
```
keegancsmith, line 1124: how long does this test take? Looks more lef a Benchmark test than a correctness test?

keegancsmith (approved):

internal/database/gitserver_repos.go
mrnugget: I want it in Info because I want to see it on sourcegraph.com, where we don't log debug.

Want to log `batchSize` because maybe it gets tweaked with deployments and that makes it easier to compare logs.

internal/database/gitserver_repos_test.go
mrnugget: 5sec on my machine. I think that's still fine. But yeah, we could also convert to Benchmark.

internal/database/gitserver_repos.go
mrnugget: Logger is already scoped to the cleanup. Do you mean scoping to this particular function?

internal/database/gitserver_repos.go
mrnugget: It's also only logged once per janitor run which happens... once per day? So I think it's fine.

internal/database/gitserver_repos.go
mrnugget: Added more scoping.

PR #58922, opened 2023-12-12
update vitest to 1.0.4

keegancsmith:
I was having an issue where vitest was reporting the wrong line numbers on test failure in watch mode. This fixed it.

Test Plan: pnpm it and then make a test fail. See that it reports the correct line.

### DIFF
--- a/package.json
+++ b/package.json
@@ -262,7 +262,7 @@
     "utc-version": "^2.0.2",
     "vite": "^4.1.4",
     "vite-plugin-turbosnap": "^1.0.3",
-    "vitest": "1.0.0-beta.4",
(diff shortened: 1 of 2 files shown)


keegancsmith:
I sent this PR too soon, actually testing now.

keegancsmith:
I went down a bit of a rabbit hole, and for some reason by updating vitest, the VisibilitySensor component no longer works in tests. It seems to still work when I use sg start, just in vitest. It complains with:

```
Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object
```

I learnt a bit, but did not resolve it. I give up.

PR #58924, opened 2023-12-12
New implementation for tree-sitter based local detection

kritzcreek:
## Description

A re-worked implementation of locals detection via tree-sitter queries.
The main difference to the previous implementation is that we now support lexical scoping and "targeted" hoisting.

### Lexical scoping

Let's look at a Go example to show the improved behaviour for lexical scoping:

```go
package main

var my_var = 1
func main() {
    my_var = 3 + my_var

    var my_var = 4
    my_var = 3 + my_var
}
```

The new implementation produces this diff when compared to the previous implementation:

```diff
  package main

  var my_var = 1
//    ^^^^^^ definition local 1
  func main() {
//     ^^^^ definition local 2
-      my_var = 3 + my_var
-//    ^^^^^^ reference local 3
-//                 ^^^^^^ reference local 3
+      my_var = 3 + my_var
+//    ^^^^^^ reference local 1
+//                 ^^^^^^ reference local 1

      var my_var = 4
//        ^^^^^^ definition local 3
      my_var = 3 + my_var
//    ^^^^^^ reference local 3
//                 ^^^^^^ reference local 3
```

Previously we'd indiscriminately lift all definitions to the top of their enclosing scope.
But Go's scoping rules for variables are purely lexical, so a definition is only in scope _after_ it was declared.
This means the first two references to `my_var` in this example have to reference the global definition.

Other languages that'll benefit from this include Perl, Rust, and OCaml (lol)

### Targeted hoisting

As we don't have local queries for JavaScript and Python yet I'll have to explain what this means without a diff in output.

Let's take this JavaScript example:
```js
function my_fun() {
  let x = 10;
  { var y = x };
  console.log(`${x} and ${y}`);
};
```

`var` is famously hoisted to the function level in JavaScript, and so both of the references in `console.log` should be resolved in this example.
Previously we'd only lift definitions to the closest enclosing scope, but with the new implementation we can specify that `var` bindings should be lifted to the closest enclosing scope of kind `function`.

A similar situtation exists for Python, where all local variables only exist at the function scope, and as such will need to be lifted as well.

## Test plan

My current best means of evaluating the new implementation against the previous one is to compare the test snapshots.
While definitions are renumbered, all references should still point to the "same" renumbered definition.

The new implementation also commits snapshots of the scope tree it generated during its execution. This should help us spot regressions in the future.

## Quick and dirty benchmarking

After a bit of profiling and fixing one egregious performance problem we are now slightly faster than the previous implementation for both Go and Java. Peak memory usage on the sourcegraph/sourcegraph repo also goes down from 150mb to 100mb. One thing of note is that the new implementation is significantly (2x) faster than the previous one on the largest Go file (3MB) in the sourcegraph repo.

<details>
<summary>Details</summary>

https://github.com/sourcegraph/sourcegraph

```
======
Previous
======
Measuring parsing
Slowest files:
mocks_temp (3319kb): 628.590667ms
mocks_test (704kb): 65.774708ms
mocks_test (430kb): 40.798334ms
mocks_test (440kb): 40.719875ms
mocks_test (427kb): 39.004792ms
gitserver.pb (247kb): 30.492917ms
mocks_test (303kb): 26.187458ms
mocks_temp (274kb): 25.060458ms
mocks_test (229kb): 19.002125ms
perms_store_test (129kb): 15.6135ms
Done 4.889227667s
======
New
======
Measuring parsing
Slowest files:
mocks_temp (3319kb): 350.291541ms
mocks_test (704kb): 54.395292ms
mocks_test (440kb): 35.303709ms
mocks_test (427kb): 35.22075ms
mocks_test (430kb): 34.949958ms
gitserver.pb (247kb): 28.288625ms
mocks_test (303kb): 23.554041ms
mocks_temp (274kb): 21.877375ms
mocks_test (229kb): 17.744042ms
perms_store_test (129kb): 15.049542ms
Done 4.491829541s

```

https://github.com/spring-projects/spring-framework:

```
======
Previous
======
Measuring parsing
Slowest files:
SpelCompilationCoverageTests (233kb): 23.238959ms
ServletAnnotationControllerHandlerMethodTests (170kb): 15.837125ms
AutowiredAnnotationBeanPostProcessorTests (163kb): 15.81025ms
DefaultListableBeanFactoryTests (132kb): 13.253625ms
MergedAnnotationsTests (144kb): 11.031292ms
DataBinderTests (92kb): 9.773125ms
ClassReader (172kb): 9.5185ms
SpelReproTests (79kb): 7.625125ms
AbstractPropertyAccessorTests (71kb): 7.518709ms
ResolvableTypeTests (70kb): 6.893208ms
Done 3.4509975s
======
New
======
Measuring parsing
Slowest files:
SpelCompilationCoverageTests (233kb): 23.672458ms
AutowiredAnnotationBeanPostProcessorTests (163kb): 15.413667ms
ServletAnnotationControllerHandlerMethodTests (170kb): 14.913167ms
DefaultListableBeanFactoryTests (132kb): 12.959834ms
MergedAnnotationsTests (144kb): 10.826292ms
DataBinderTests (92kb): 9.529833ms
ClassReader (172kb): 9.451625ms
AbstractPropertyAccessorTests (71kb): 7.398167ms
SpelReproTests (79kb): 7.373459ms
ResolvableTypeTests (70kb): 6.700875ms
Done 3.368848875s
```

https://github.com/aludnam/MATLAB:

```
======
Previous
======
Measuring parsing
Slowest files:
mtimesx_test_ssspeed (420kb): 21.280625ms
mtimesx_test_sdspeed (393kb): 19.575ms
mtimesx_test_dsspeed (392kb): 19.428042ms
mtimesx_test_ddspeed (121kb): 17.436875ms
mtimesx_test_ssequal (359kb): 16.542208ms
mtimesx_test_sdequal (354kb): 16.2475ms
mtimesx_test_dsequal (354kb): 15.889459ms
mtimesx_test_ddequal (94kb): 12.933917ms
gui_cp (80kb): 9.220958ms
nmf_smart_alg (23kb): 5.752375ms
Done 934.253625ms
======
New
======
Measuring parsing
Slowest files:
mtimesx_test_ssspeed (420kb): 22.3835ms
mtimesx_test_sdspeed (393kb): 20.55ms
mtimesx_test_dsspeed (392kb): 20.250375ms
mtimesx_test_ddspeed (121kb): 18.290083ms
mtimesx_test_ssequal (359kb): 17.372833ms
mtimesx_test_sdequal (354kb): 16.939709ms
mtimesx_test_dsequal (354kb): 16.5ms
mtimesx_test_ddequal (94kb): 13.711ms
gui_cp (80kb): 8.552125ms
exportfig (33kb): 3.653959ms
Done 921.589833ms
```

</details>


docker-images/syntax-highlighter/crates/scip-syntax/README.md
```diff
@@ -0,0 +1,110 @@
+# scip-locals
```
varungandhi-src, line 1:
```suggestion
# scip-syntax
```

docker-images/syntax-highlighter/crates/scip-syntax/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Globals (To be documented)
+## Symbols (To be documented)
```
varungandhi-src, line 4: If this is a TODO for the future, it'd be nice to file an issue and link it here.

docker-images/syntax-highlighter/crates/scip-syntax/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+`src/locals.rs` implements an evaluator for our tree-sitter query based DSL to label local definitions and references for various programming languages using purely syntactic information. This data is used to enable fast and lightweight file-local navigation in the blob view.
+
+Queries describing the local binding structure of various programming languages are maintained in `queries/*/scip-locals.scm`.
+
+### The Query DSL
+
+For specifying how to resolve local bindings we use the [tree-sitter query language] and a set of custom captures and properties. The three main concepts are _scopes_, _definitions_, and _references_.
+
+### Scopes
```
varungandhi-src, line 16: nit: The heading structure is a bit off, either this should be a level 4 heading and below should be level 5, or the one above it should be level 2.

docker-images/syntax-highlighter/crates/scip-syntax/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+Scopes are specified by labeling a capture as a `@scope[.kind]`. The optional scope kind can be used to hoist definitions to scopes of that kind. There is an implicit top-level scope that is of kind `"global"`
```
varungandhi-src, line 18: Please specify if `.kind` can be any string or if it needs to be one of a selected set of strings.

docker-images/syntax-highlighter/crates/scip-syntax/README.md
```diff
@@ -0,0 +1,110 @@
...
+```js
+print(my_var) // Will not be resolved
+
+let my_var = 10
+print(my_var) // Will be resolved
+```
+
+#### Hoisting
+
+If you want a definition to be _hoisted_ instead, you can specify what kind of scope it should be hoisted to.
```
varungandhi-src, line 48:
```suggestion
If you want a definition to be _hoisted_ to the start of a scope instead, you can specify the kind of the nearest enclosing scope it should be hoisted to.
```

docker-images/syntax-highlighter/crates/scip-syntax/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+```scm
+(function_definition
+ (identifier) @definition
+ #set! "hoist" "function")
+```
+
+It'll then be visible from the closest parent scope with the given kind. If no parent with the given kind is found, the definition will be visible in the global scope.
```
varungandhi-src, line 56:
```suggestion
The definition will be visible to the nearest enclosing scope with kind `function`. If no such enclosing scope is found, the definition will be visible in the global scope.
```

docker-images/syntax-highlighter/crates/scip-syntax/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+```js
+my_func(10) // Will be resolved
```
varungandhi-src, line 59:
```suggestion
my_func(10) // Will be resolved as `my_func` will be visible at global scope
```

docker-images/syntax-highlighter/crates/scip-syntax/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+```scm
+(function_definition
+ (identifier) @definition
+ #set! "hoist" "function")
+```
+
+It'll then be visible from the closest parent scope with the given kind. If no parent with the given kind is found, the definition will be visible in the global scope.
+
+```js
```
varungandhi-src, line 58: This example is a bit confusing as it is only talking about the final fallback case, and not the more common case related to hoisting to the start of a scope like Python.

docker-images/syntax-highlighter/crates/scip-syntax/README.md
```diff
@@ -0,0 +1,110 @@
...
+my_func(10) // Will be resolved
+
+function my_func(x) {
+  print(x)
+}
+```
+
+#### First assignment is Definition
+
+Certain languages (eg. Python or Matlab) do not have special syntactic forms for introducing variables.
```
varungandhi-src, line 68:
```suggestion
Certain languages (Python, MATLAB etc.) do not have special syntactic forms for introducing variables.
```

docker-images/syntax-highlighter/crates/scip-syntax/README.md
```diff
@@ -0,0 +1,110 @@
...
+Instead the first assignment of a variable is considered to be its definition, and all further ones are references.
+To support this in our DSL you can can mark a `@definition` as a 'def_ref'.
```
varungandhi-src, line 70: Very very minor nit, but it's a little bit jarring to see some paragraphs with all sentences on one line, and some paragraphs with one sentence per line. Please stick to a single style within a file, doesn't matter which one.

(I personally prefer https://sembr.org/, but we don't use it much, so you're not required to use it.)

docker-images/syntax-highlighter/crates/scip-syntax/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+```scm
+(assignment
+ (identifier) @definition
+ #set! "def_ref")
+```
+
+If you also specify a hoist level, only existing assignments that match the current hoist-level will be considered when deciding between definition and reference.
+As an example here's how Python's hoisted ref_defs would get resolved.
```
varungandhi-src, line 79:
```suggestion
As an example, here's how local variables in Python functions could be handled.
```

docker-images/syntax-highlighter/crates/scip-syntax/design/scoping.md
```diff
@@ -0,0 +1,167 @@
+# Scoping of different programming languages
```
varungandhi-src, line 1: 1. We don't have enough docs to warrant a multi-level structure, and putting this under `design` doesn't match existing conventions in our other codebases where docs usually go under a `doc` or `docs` folder. Instead, you can move this under `docs/scoping.md` and add a brief entry in the crate's README or in `docs/README.md` describing this file.
2. It'd be nice to have all docs be transitively linked from the crate's README. Whether that is directly linked or via a `docs/README.md` is up to you.

docker-images/syntax-highlighter/crates/scip-syntax/README.md
varungandhi-src: "hoisting" is defined in the scoping.md file but it is used here. It'd be nice to have a small section in the Design.md file with just `Hoisting`, give a couple of examples, and link that section's heading from here.

docker-images/syntax-highlighter/crates/scip-syntax/README.md
kritzcreek: This was more aspirational... I'm definitely not getting to it this in this PR, and I don't see a reason why this needs to be documented any more urgent than any other part of our code. I've decided to remove these headers.

docker-images/syntax-highlighter/crates/scip-syntax/README.md
kritzcreek: Thank you for spotting this! I generally try to use one sentence per line, because it makes line-based diffs so much better.

Unfortunately Github doesn't render these linebreaks as spaces, but I prefer good diffs over aesthetics in this case.

docker-images/syntax-highlighter/crates/scip-syntax/README.md
kritzcreek: I've made the example a little more complex to cover both cases.

docker-images/syntax-highlighter/crates/scip-syntax/README.md
kritzcreek: Done

docker-images/syntax-highlighter/crates/scip-syntax/README.md
kritzcreek: For now scope kinds are arbitrary strings. I think we might want to move to using scip terminology for these (if that exists), and check these when reading in the locals query.

I've added a definition for hoisting in the scoping document and linked to it.

kritzcreek:
Thank you very much for all the comments and suggestions on the documentation. I've taken a pass over all of them, and moved some things around. PTAL

kritzcreek:
I've done some performance testing against a couple large-ish Matlab projects on GitHub and the new implementation is slightly faster for all of them. I added a run against the largest repo I could find to the initial description.

docker-images/syntax-highlighter/crates/scip-syntax/queries/go/scip-locals.scm
```diff
@@ -77,10 +74,10 @@
...
+;; ;; TODO: These may not make much sense to have for locals... {{{
+;; ((package_identifier) @reference
+;;   (set! reference.kind "namespace"))
 
-(package_clause
-   (package_identifier) @definition.namespace)
-;; }}}
+;; (package_clause
+;;    (package_identifier) @definition.namespace)
+;; ;; }}}
```
varungandhi-src, line 83: nit: Delete these commented out queries? I can't use the GitHub suggestions feature to suggest a deletion, but I would refrain from leaving in commented code except for debugging stuff.

docker-images/syntax-highlighter/crates/scip-syntax/src/bin/scip-local-nav.rs
```diff
@@ -41,7 +41,7 @@ fn parse_files(config: &LocalConfiguration, root: &Path, dir: &Path) -> Vec<Docu
             .parse(contents.as_bytes(), None)
             .expect("to parse the tree");
 
-        let occs = parse_tree(config, &tree, contents.as_bytes()).expect("to get occurrences");
+        let occs = parse_tree(config, &tree, contents.as_bytes());
```
varungandhi-src, line 44: Question: Unrelated to this PR, but for languages where syntax highlighting is already using Tree-sitter, we'll end up parsing the file twice, right? It'd be nice to avoid that as we expand support for SCIP locals to more languages.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
```diff
@@ -1,504 +1,721 @@
...
-        for (idx, child) in self.children.iter().enumerate() {
-            if child.lvalues.is_empty() {
-                empty_children.push(idx);
-            }
-        }
+// Can we make these usize's NonZeroUsize to make this enum smaller?
+#[derive(Debug, Clone, Copy)]
+enum DefRef {
+    PreviousDefinition(usize),
+    NewDefinition(usize),
```
varungandhi-src, line 223: Is this `usize` because of `id: usize`? Maybe introduce a newtype so that can be used instead? If the IDs are just monotonic, then we can replace the innards with NonZeroU32 easily.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
```diff
@@ -1,504 +1,721 @@
...
+struct LocalResolver<'a> {
+    arena: Arena<Scope<'a>>,
+    interner: StringInterner,
+
+    source_bytes: &'a [u8],
+    definition_id_supply: usize,
+    // This is a hack to not record references that overlap with
+    // definitions. Ideally we'd fix our queries to prevent these
+    // overlaps.
+    definition_start_bytes: HashSet<usize>,
```
varungandhi-src, line 236: Maybe rename to `definition_start_positions` or `definition_start_byte_offsets`? The `bytes` is a bit confusing, because I'd expect the data type to have some actual bytes/contents from the start of the definition based on the name of the field.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
```diff
@@ -1,504 +1,721 @@
...
+            source_bytes,
+            definition_id_supply: 0,
+            definition_start_bytes: HashSet::new(),
+            occurrences: vec![],
         }
+    }
 
-        self.children
-            .sort_by_key(|s| (s.range.start_line, s.range.end_line, s.range.start_col));
+    fn _start_byte(&self, id: ScopeRef<'a>) -> usize {
```
varungandhi-src, line 252: Any specific reason why this is underscored when other non-pub methods aren't?

varungandhi-src (commented):
Need to head out for a bit, I'll do a third pass later today or on Monday.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
```diff
@@ -1,504 +1,721 @@
...
+type ScopeRef<'a> = Id<Scope<'a>>;
+
+/// We use string_interner to intern variable names
+type Name = DefaultSymbol;
+
+#[derive(Debug)]
+struct Scope<'a> {
+    /// For a query that captures a "@scope.function" this will
+    /// contain the string "function"
+    ty: String,
```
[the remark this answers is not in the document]
kritzcreek: Thanks, well spotted!

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
kritzcreek: These used to create warnings because they were only used in debugging code (print_scope). Now that they're always used this isn't necessary anymore. Well spotted!

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
```diff
@@ -1,504 +1,721 @@
...
+                    continue;
+                };
+                if capture_name.starts_with("scope") {
+                    let ty = capture_name.strip_prefix("scope.").unwrap_or(capture_name);
+                    scopes.push(ScopeCapture {
+                        ty,
+                        node: capture.node,
+                    })
+                } else if capture_name.starts_with("definition") {
+                    let is_def_ref = properties.iter().any(|p| p.key == "def_ref".into());
```
varungandhi-src, line 532: Here, we run the risk of having old properties stick around in the queries, like `reassignment_behavior`. It would be good to validate properties against known allowlists for the various capture types. In production, we can skip the validation step and ignore any unknown properties.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
```diff
@@ -1,504 +1,721 @@
...
+                    let mut hoist = None;
+                    if let Some(prop) = properties.iter().find(|p| p.key == "hoist".into()) {
```
varungandhi-src, line 534: Prefer `p.key.as_ref() == "mystring"` over comparing after `into()`, it's not 100% obvious that the compiler will be able to optimize away the malloc call.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
```diff
@@ -1,504 +1,721 @@
...
     }
 
-    fn rec_into_occurrences(
-        &self,
-        id: &mut usize,
-        occurrences: &mut Vec<Occurrence>,
-        declarations_above: &mut Vec<HashMap<&'a str, usize>>,
+    fn add_definition(
+        &mut self,
+        id: ScopeRef<'a>,
```
varungandhi-src, line 266: Given that there are multiple IDs floating around, it would be nice to consistently name this `scope_id: ScopeId<'a>` instead. Also, as a suffix, I would say that 'Ref' is preferable for pointer-like types, whereas 'Id' is preferable for indexes.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
kritzcreek: Now that definition ids are u32, this enum is just pointer sized on 64bit systems. so no need for NonZero

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
```diff
@@ -1,504 +1,721 @@
...
+    // TODO: (perf) we could also remember how many definitions
+    // precede us in the parent, for efficient slicing when searching
+    // up the tree
+    parent: Option<ScopeRef<'a>>,
+
+    /// Definitions that have been hoisted to the top of this scope
+    // TODO: (perf) for hoisted definitions the lexicographical order
+    // shouldn't matter anymore, so we might want to turn this into a
+    // HashMap for faster lookups
+    hoisted_definitions: Vec<Definition<'a>>,
```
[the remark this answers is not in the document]
kritzcreek: I put it off until later because I wasn't sure how namespacing was going to show up as part of the HashMap key here, but I'm sure it won't be hard to figure out when the time comes.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
```diff
@@ -1,504 +1,721 @@
...
-        }
+    fn get_scope(&self, id: ScopeRef<'a>) -> &Scope<'a> {
+        self.arena.get(id).unwrap()
     }
 
-    pub fn display_scopes(&self) {
-        self.rec_display_scopes(0);
+    fn get_scope_mut(&mut self, id: ScopeRef<'a>) -> &mut Scope<'a> {
+        self.arena.get_mut(id).unwrap()
     }
```
varungandhi-src, line 357: Thoughts on if it would make sense to replace `get_scope` and `get_scope_mut` with `Index` and `IndexMut` impls to make the code a bit more compact? 🤔 This seems like a pretty common operation, and in essence, we do have an indexing operation.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
kritzcreek: Thanks, that totally makes sense.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
```diff
@@ -1,504 +1,721 @@
...
+            self._print_scope(w, *child, depth + 2)
+        }
+    }
 
-        let mut scope = None;
-        let mut lvalue = None;
-        let mut reference = None;
-        let mut scope_modifier = None;
-        let mut reassignment_behavior = None;
+    fn mk_name(&mut self, s: &str) -> Name {
```
varungandhi-src, line 451: IME, it's not very common in Rust code to abbreviate `make` to `mk`.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
```diff
@@ -1,504 +1,721 @@
...
+            };
+            (def_id, definition)
+        };
+
+        let is_new_definition = match hoist {
+            Some(hoist_scope) => {
+                let mut target_scope = id;
+                // If we don't find any matching scope we hoist all
+                // the way to the top_scope
+                for ancestor in self.ancestors(id) {
```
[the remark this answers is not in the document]
varungandhi-src: Given that we're only iterating over the ancestors in one place, it might be less code overall to replace this with a dedicated function without the need for a separate `Iterator` impl. But it's not a big deal either ways, your call.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
```diff
@@ -1,504 +1,721 @@
...
-    /// ^ local 1
-    /// a = 10
-    /// ^ local 1 (reference)
-    OldestIsDefinition,
-}
+    fn parent(&self, id: ScopeRef<'a>) -> ScopeRef<'a> {
+        self.get_scope(id)
+            .parent
+            .expect("Tried to get the root node's parent")
+    }
```
varungandhi-src, line 370: Given that this function is only used in two places, I think it would be better to inline the body at both call-sites so that it is clearer at the call-site why a parent will always exist. Otherwise, based on the function's name, it's not obvious that it may panic internally.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
kritzcreek: I called them `start_bytes`, because they're obtained by calling `node.start_byte()` on tree-sitter's Node type. `position` is something different in the tree-sitter domain (includes line/column). I can go with `byte_offsets`, but just wanted to confirm you still think that's a better name given the `node.start_byte()` call.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
```diff
@@ -1,504 +1,721 @@
...
+    node: Node<'a>,
+}
 
-                self.children[idx].insert_reference(reference);
+/// Created by LocalResolver::ancestors()
+#[derive(Debug)]
+struct Ancestors<'arena, 'a> {
+    arena: &'arena Arena<Scope<'a>>,
+    current_scope: ScopeRef<'a>,
+}
```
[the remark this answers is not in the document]
kritzcreek: I wrote it like this, because the Rust std library does something similar for its Iterator structs. For example: https://doc.rust-lang.org/std/iter/struct.Filter.html, do you still want me to get rid of it?

I'll add a comment about the arena (it's for the whole tree).

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
```diff
@@ -1,504 +1,721 @@
...
+    /// enclosing scope, or to hoist them to the closest matching
+    /// scope.
+    fn build_tree(&mut self, top_scope: ScopeRef<'a>, captures: Captures<'a>) {
+        let Captures {
+            mut scopes,
+            mut definitions,
+            mut references,
+        } = captures;
+        // In order to do a pre-order traversal we need to sort scopes and definitions
+        // TODO: (perf) Do a pass to check if they're already sorted first?
```
varungandhi-src, line 569:
```suggestion
    /// Build a tree of scopes for pre-order traversal by sorting scopes, definitions
    /// and references. Definitions and references are added or hoisted to the
    /// closest enclosing scope as appropriate.
    fn build_tree(&mut self, top_scope: ScopeRef<'a>, captures: Captures<'a>) {
        let Captures {
            mut scopes,
            mut definitions,
            mut references,
        } = captures;
```

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
```diff
@@ -1,504 +1,721 @@
...
+        assert!(
+            definitions_iter.next().is_none(),
+            "Should've entered all definitions into the tree"
+        );
     }
 
-    while let Some(m) = lvalues.pop() {
-        root.insert_lvalue(m);
+    /// Walks up the scope tree and tries to find the definition for a given name
+    fn find_def(
```
varungandhi-src, line 640: nit: Since this is fallible and returns an `Option` on failure, I think it would be more idiomatic to have a `try_` prefix.

```suggestion
    fn try_find_def(
```

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
```diff
@@ -1,504 +1,721 @@
...
+    ) -> Option<&Definition<'a>> {
+        let mut current_scope = scope;
+        loop {
+            let scope = self.get_scope(current_scope);
+            if let Some(def) = scope.find_def(name, start_byte) {
+                break Some(def);
+            } else if let Some(parent_scope) = scope.parent {
+                current_scope = parent_scope
+            } else {
+                break None;
```
varungandhi-src, line 654:
```suggestion
                return Some(def);
            } else if let Some(parent_scope) = scope.parent {
                current_scope = parent_scope
            } else {
                return None;
```

As a defensive measure, I would also add a loop limit with an `eprintln!` if the limit is hit, in case there is a bug elsewhere that causes an infinite loop.

varungandhi-src (commented):
Leaving some more comments as time permits, will take a deeper look at some of the more complex logic later.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
```diff
@@ -1,504 +1,721 @@
...
+        let mk_def = |this: &mut Self| {
+            this.definition_id_supply += 1;
+            let def_id = this.definition_id_supply;
+            let definition = Definition {
+                id: def_id,
+                name,
+                node,
+            };
+            (def_id, definition)
+        };
```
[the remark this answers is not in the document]
kritzcreek: Yeah this is a bit tricky. I've refactored this a couple times. I'll describe what this is doing and maybe you can come up with a nicer way.

All of the current queries do something like this:

```scm
(var_definition (identifier) @definition)
(func_definition (identifier) @definition)
;; ... list all ways to define names
(identifier) @reference
```

This means for every definition capture you'll also get a reference capture that overlaps it, which is why we're recording these start_bytes in the first place. Any reference that overlaps with a definition should be discarded.

But now with def_refs we end up converting these definitions into references. How do we make sure we still only consider the reference once?

To understand the logic here, we also need the logic at the place where `definition_start_bytes` is actually used:
```rust
let def_id = if let Some(resolved) = resolves_to {
    *resolved
} else if self.definition_start_bytes.contains(&node.start_byte()) {
    // See the comment on LocalResolver.definition_start_bytes
    continue;
} else if let Some(def) = self.find_def(scope_ref, *name, node.start_byte()) {
    def.id
} else {
    continue;
};
```

Basically if the `resolves_to` field is set to Some we don't check for definition_start_bytes. That's how we make sure to only consider these 'duplicate' references once, and that's why we still add the start_byte to definition_start_bytes here.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
kritzcreek: `Ref` is still left over from when I started out with typed_arena rather than id_arena, where these actually were references :D I'll change the name to ScopeId and rename parameter bindings.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
kritzcreek: What would you implement `Index` and `IndexMut` on? `LocalResolver`?

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
kritzcreek: `vec![].iter().find()` also returns an Option, right? I'd agree if there _also_ was a non-fallible find function around here.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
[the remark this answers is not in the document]
varungandhi-src: > I wrote it like this, because the Rust std library does something similar for its Iterator structs

Right, but in our case, we're not publishing this as a library, it's likely being consumed in the editor, so it seems redundant. The Rust standard library in contrast probably has more people reading the comments on docs.rs compared to those looking at the source code, and you can't do 'Find references' on docs.rs

I don't think it's a big deal either way, your call.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
varungandhi-src: Yeah, I think we should use `byte_offset`. I'd argue that the Tree-sitter method `start_byte` is not named well, because `byte` as the suffix kinda' implies a return type of `u8`. If it were named `bytes_start` that would also make more sense than `start_byte` IMO, or perhaps even just a `bytes_range()` returning a `std::ops::Range`. In our field names, we should try to not repeat the same mistake.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
varungandhi-src: Yep

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
varungandhi-src: Fair point. Funnily enough, I did find a separate `try_find` on Iterator here: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.try_find

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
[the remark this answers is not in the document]
varungandhi-src: Hmm, maybe it should be renamed to `skip_references_offsets` or `skip_references_at_offsets` then? So only pure references with matching offsets would be ignored, but def-refs at that offset wouldn't be skipped.

Also, for clarity at the point of use, instead of unpacking the fields of the loop variable:

```
            for Reference {
                name,
                node,
                resolves_to,
            } in scope.references.iter()
```

If the code read like this:

```
                let def_id = if let Some(resolved) = ref.resolves_to {
                    // skip_references_at_offsets only applies to pure references,
                    // but this is a def-ref, see the comment on LocalResolver.skip_references_at_offsets
                    *resolved
                } else if self.skip_references_at_offsets.contains(&ref.node.start_byte()) {
                    continue;
                } else if let Some(def) = self.find_def(scope_ref, *ref.name, ref.node.start_byte()) {
                    def.id
                } else {
                    continue;
                };
```

It'd be a bit clearer.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
```diff
@@ -1,504 +1,734 @@
...
+                } else {
+                    continue;
+                };
+
+                let symbol = format_symbol(def_id.as_local_symbol());
+                ref_occurrences.push(scip::types::Occurrence {
+                    range: node.to_scip_range(),
+                    symbol: symbol.clone(),
+                    ..Default::default()
+                });
```
varungandhi-src, line 697:
```suggestion
                ref_occurrences.push(scip::types::Occurrence {
                    range: node.to_scip_range(),
                    symbol: format_symbol(def_id.as_local_symbol()),
                    ..Default::default()
                });
```

nit: Prefer move over `clone()`.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
```diff
@@ -1,504 +1,734 @@
...
+        self.occurrences.extend(ref_occurrences);
     }
 
-    let occs = root.into_occurrences(capacity);
+    // The entry point to locals resolution
+    fn process(
+        mut self,
+        config: &'a LocalConfiguration,
+        tree: &'a tree_sitter::Tree,
+        test_writer: Option<&mut impl Write>,
```
varungandhi-src: I've changed this to `dyn Write` because it's a testing-only code path, and using `impl Write` will unnecessarily create multiple monomorphizations of this and the `print_scope` function if we end up calling it with two different type parameters.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
kritzcreek: I did that in 5b4535c5156ccc1869d748c42d08938dd8c81d50. TBH I don't think it's better or worse, it's just different. Your call

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
kritzcreek: Is there any chance we'll end up using these queries for different purposes, with different sets of query captures?

That's mostly why I didn't do that validation after removing the old code.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
```diff
@@ -1,504 +1,721 @@
...
-use crate::languages::LocalConfiguration;
+// Missing features at this point
+// a) Namespacing
+//
+// The simplest thing I can think of right now is to use
+// `@definition.namespace` and `@reference.namespace`. Because
+// most of the locals queries just declare all `(identifier)` as
+// references we'll probably make it so `@reference` with no
+// namespace matches definitions in any namespace and
+// `@definition` matches any `@reference.namespace`
```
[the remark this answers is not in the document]
kritzcreek: I've thought about this a bit. I can't come up with an example that negatively impacts locals for the languages we support right now. I'm pretty sure it'll come up when we do TypeScript, but until then I don't really have a good example I could put into an issue.

docker-images/syntax-highlighter/crates/scip-syntax/src/bin/scip-local-nav.rs
kritzcreek: Maybe this is just bad naming `parse_tree` here accepts the parsed syntax tree from tree sitter as an argument. I don't think we're re-parsing with tree-sitter anywhere.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
```diff
@@ -1,504 +1,721 @@
...
+// C = 10..12
+// B.cmp(C) = Less
+// Because C is contained within B we want to make sure to visit B first.
+fn compare_range(a: Range<usize>, b: Range<usize>) -> Ordering {
+    // TODO: This assert fails for Java, ideally we'd fix the query to
+    // not report duplicate scopes
+    // assert!(
+    //     result != Ordering::Equal,
+    //     "Two scopes must never span the exact same range: {a:?}"
+    // );
```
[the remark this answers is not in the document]
kritzcreek: I fixed the java query to no longer trigger this issue and turned the check back on. Ideally it wouldn't be a panic though. How should we communicate errors that should be logged/investigated that don't require us to abort?

docker-images/syntax-highlighter/crates/scip-syntax/src/bin/scip-perf.rs
```diff
@@ -1,25 +1,48 @@
...
-    pub filepath: String,
+    pub file_path: String,
+    pub file_size: usize,
     pub duration: std::time::Duration,
 }
 
-fn parse_files(dir: &Path) -> Vec<ParseTiming> {
-    let config = scip_syntax::languages::get_local_configuration(BundledParser::Go).unwrap();
-    let extension = "go";
+fn parse_files(dir: &Path, language: Language) -> Vec<ParseTiming> {
```
varungandhi-src, line 31: Given that this `enum Language` is essentially only created and destroyed in one place, I think it's fine to accept a string here, use [`get_parser`](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/docker-images/syntax-highlighter/crates/scip-treesitter-languages/src/parsers.rs?L60:12-60:22) to convert the string to a `BundledParser::Language`, and compare that to a standard array/set constant which contains the list for which we support locals & parameters. (Maybe add a wrapper macro around the `create_locals_configuration!` calls that initializes a `HashSet` and calls out to `create_locals_configuration!` for the listed languages?)

We should avoid creating more places that need updating whenever we add support to a new language.

docker-images/syntax-highlighter/crates/scip-syntax/src/bin/scip-local-nav.rs
varungandhi-src: Ah! 🤦🏽 In that case, please rename this function to not have `parse_` as it's not actually converting bytes/string -> a tree structure.

docker-images/syntax-highlighter/crates/scip-syntax/src/bin/scip-perf.rs
```diff
@@ -58,11 +91,16 @@ fn measure_parsing() {
...
     let root = Path::new(&args.root_dir);
 
-    let mut timings = parse_files(root);
+    let mut timings = parse_files(root, args.language);
     timings.sort_by(|a, b| a.duration.cmp(&b.duration));
     println!("Slowest files:");
     for timing in timings.iter().rev().take(10) {
-        println!("{}: {:?}", timing.filepath, timing.duration);
+        println!(
+            "{} ({}kb): {:?} ",
```
varungandhi-src, line 99:
```suggestion
            "{} ({}kb): {:?}",
```

nit: extraneous space at end before newline

varungandhi-src (approved):

docker-images/syntax-highlighter/crates/scip-syntax/src/bin/scip-perf.rs
kritzcreek: I'm using an enum here so `--help` works out of the box. I can take a look at implementing the necessary traits for some auto-populated structure to make that work when I do some more benchmarking for excluding globals from these results.

varungandhi-src: Given that this isn't going to be used a lot, I think it's fine if `--help` doesn't list all the languages and there is just a comment in the code saying "look at this constant" or something.

But it's not a big deal, your call.

kritzcreek: I've decided to fix this in the future. I'll be using this code to benchmark the new locals queries for Scala and Kotlin, so that'll be a good time.

docker-images/syntax-highlighter/crates/scip-syntax/src/locals.rs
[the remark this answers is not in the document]
kritzcreek: I've changed this to a `debug_assert` so it doesn't cause panics in production.

PR #58927, opened 2023-12-12
search: expose GitHub/GitLab topics in UI

stefanhengl:
Currently, we sync topics from GitHub and GitLab and it is possible to
filter repositories by their topics with the `repo:has.topic` filter.

However, unlike user-added metadata, the synced topics are not visible in the UI,
neither for repo matches nor on the repo tree page. This means it is impossible 
for users to figure out which topics they can filter by.

With this PR we display topics, such as "language" or "golang", in addition to 
metadata (EG `fruit:banana`) for repo matches and on the repo tree page (see screenshots).

**Search results**
Screenshot 2023-12-15 at 12 44 54

**Repository tree page (right panel)**
Screenshot 2023-12-15 at 12 45 03

## Test plan

Note: Our search language distinguishes between user provided metadata (has.meta) and automatically synced metadata (has.topic). 

- manual testing
  - I checked that clicking on a "topic" badge adds a "has.topic" filter, while clicking on a "metadata" badge add a "has.meta" filter
  - Adding and deleting metadata works like before
  - Topics and metadata are sorted alphabetically
  - Results without metadata or topics are displayed correctly
- added new unit test

### DIFF
--- a/client/branded/src/search-ui/components/RepoSearchResult.tsx
+++ b/client/branded/src/search-ui/components/RepoSearchResult.tsx
@@ -75,17 +75,16 @@ export const RepoSearchResult: React.FunctionComponent<RepoSearchResultProps> = 
 
     const showExtraInfo = result.archived || result.fork || result.private
 
-    const metadataTags = !metadata
-        ? []
-        : Object.entries(metadata).map(([key, value]) =>
-              metadataToTag({ key, value }, queryState, false, buildSearchURLQueryFromQueryState)
-          )
+    const tags = [
+        ...(metadata
+            ? Object.entries(metadata).map(([key, value]) =>
+                  metadataToTag({ key, value }, queryState, false, buildSearchURLQueryFromQueryState)
+              )
+            : []),
+        ...(topics ? topics.map(topic => topicToTag(topic, queryState, false, buildSearchURLQueryFromQueryState)) : []),
+    ]
 
-    const topicTags = !topics
-        ? []
-        : topics.map(topic => topicToTag(topic, queryState, false, buildSearchURLQueryFromQueryState))
-
-    const showRepoMetadata = enableRepositoryMetadata && (metadataTags.length > 0 || topicTags.length > 0)
+    const showRepoMetadata = enableRepositoryMetadata && tags.length > 0
 
     return (
         <ResultContainer
@@ -147,7 +146,7 @@ export const RepoSearchResult: React.FunctionComponent<RepoSearchResultProps> = 
                     )}
                     {showRepoMetadata && (
                         <div className="d-flex">
-                            <TagList tags={[...metadataTags, ...topicTags]} />
+                            <TagList tags={tags} />
                         </div>
                     )}
                 </div>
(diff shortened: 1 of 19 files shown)


internal/database/repos.go
```diff
@@ -361,12 +364,22 @@ func (s *repoStore) Metadata(ctx context.Context, ids ...api.RepoID) (_ []*types
...
 		}
 
 		r.KeyValuePairs = kvps.kvps
+
+		metadata, ok, err := unmarshalMetadata(s.logger, typ, metadataBytes)
+		if err != nil {
+			return err
+		} else if ok {
+			r.Topics = GetTopics(metadata)
+		}
```
camdencheek, line 381: FYI, there is a virtual column on the `repo` table called `topics` that you can just query for this info

stefanhengl: Oops. I didn't see that. That makes things a bit tidier.

client/branded/src/search-ui/components/RepoSearchResult.tsx
```diff
@@ -73,9 +73,20 @@ export const RepoSearchResult: React.FunctionComponent<RepoSearchResultProps> =
         }
     }, [result, repositoryMatches, repoNameElement, description, descriptionMatches, repoDescriptionElement, repoName])
 
-    const showRepoMetadata = enableRepositoryMetadata && !!metadata
     const showExtraInfo = result.archived || result.fork || result.private
 
+    const metadataTags = !metadata
```
camdencheek, line 78: I don't think we want to assert with `!` here since we're doing a type-safe check with `... ? ... : ...`, right? Same with topics below?

client/web/src/repo/tree/TreePageContent.tsx
```diff
@@ -82,15 +82,28 @@ const ExtraInfoSection: React.FC<{
...
+        [repo.topics, queryState]
+    )
+
     return (
         <Card className={className}>
             <ExtraInfoSectionItem>
                 <ExtraInfoSectionItemHeader title="Description" tooltip="Synchronized from the code host" />
                 {repo.description && <Text>{repo.description}</Text>}
             </ExtraInfoSectionItem>
+            <ExtraInfoSectionItem>
```
camdencheek, line 103: Should this section only be rendered if the repo has topics? In particular, it seems odd to show a section for this, say, on Bitbucket.

cmd/frontend/graphqlbackend/graphqlbackend_test.go
```diff
@@ -216,7 +216,7 @@ func TestResolverTo(t *testing.T) {
 		typ := reflect.TypeOf(r)
 		t.Run(typ.Name(), func(t *testing.T) {
 			for i := 0; i < typ.NumMethod(); i++ {
-				if name := typ.Method(i).Name; strings.HasPrefix(name, "To") {
+				if name := typ.Method(i).Name; strings.HasPrefix(name, "To") && name != "Topics" {
```
camdencheek, line 219: That's painful 😄

camdencheek (approved):
This looks great, thank you!

client/web/src/repo/tree/TreePageContent.tsx
stefanhengl: > Should this section only be rendered if the repo has topics? 

Done

Issue #25438, opened 2021-09-28
sg: Build in support for downloading and using binaries


mrnugget:
Multiple commands in `sg.config.yaml` follow the same pattern:

1. Download a binary based on environment variables
2. Unzip the download
3. Move it into `.bin`

Example: `caddy`.

https://github.com/sourcegraph/sourcegraph/blob/548ec915b34e79d44334155a2efb721e8a7ba5b9/sg.config.yaml#L256-L274

We could build that functionality right into `sg` and reduce the amount of shell code in `sg.config.yaml` (and provide better output/debugging for the install process).

For example, we could have something like this:

```yaml
  caddy:
    ignoreStdout: true
    ignoreStderr: true
    cmd: .bin/caddy run --watch --config=dev/Caddyfile
    installBinary:
      url: https://github.com/caddyserver/caddy/releases/download/v${CADDY_VERSION}/caddy_${CADDY_VERSION}_${SG_OS}_amd64
      binaryName: caddy
    env:
      CADDY_VERSION: 2.3.0
```

PR #58928, opened 2023-12-12
fix(sg): start Caddy admin server before `caddy trust`

bevzzz:
This MR addresses #58926.  
The changes ensure the Caddy admin server is running before calling another `caddy [command]`.

I also replaced `exec` with `eval`, as that allowed using `trap` while staying in the current shell session.

<details><summary>Before</summary>

Commit [`57166ea80caa7e2eb7d9c8134b35c47997a0835c`](https://github.com/bevzzz/sourcegraph/commit/57166ea80caa7e2eb7d9c8134b35c47997a0835c)

before

</details>

<details><summary>After</summary>

after

</details>

## Test plan

Manually, since most of the logic is still in `caddy.sh` and, as far as I can tell, the script is not used in production.

1. Configure the "initial" state for anyone running `sg setup` for the first time:
- Run  `caddy untrust` to remove the root certificate from your local trust store
- Run ` caddy stop` to make sure the  Caddy admin server is not running (`curl localhost:2019` should exit with `curl: (7) Failed to connect`)

2. Run `sg setup` and wait for Caddy check to fail, then see available fixes `[8]`
3. Opt for automatic resolution `[1]`

The fix should succeed (unlike in the **Before** screen recording).

P.S.: following the [`sg` vision](https://docs.sourcegraph.com/dev/background-information/sg/vision#principles), I also looked into replacing the `caddy.sh` script with a Go equivalent, but did not find a stable public API in the `caddyserver/caddy` package for some commands (e.g. [`caddy trust`](https://sourcegraph.com/github.com/caddyserver/caddy/-/blob/modules/caddypki/command.go?L102) is not exposed and copying  a lot of what looks like internal logic did not feel quite right atm).   
Additionally, there's this ongoing discussion in #25438 about using binaries in `sg`, which might result in an altogether different solution.

### DIFF
--- a/dev/caddy.sh
+++ b/dev/caddy.sh
@@ -29,4 +29,6 @@ chmod +x "${target}"
 
 popd >/dev/null
 
-exec "${target}" "$@"
+trap '${target} stop' EXIT SIGINT
+eval "${target} start"
+"${target}" "$@"


burmudar (approved):
Thanks @bevzzz!

PR #58930, opened 2023-12-12
Fix Cody Usage Metrics SQL Query

thenamankumar:
IDEs including VS Code and JetBrains, are sending double the number of event logs, with similar event names but different sources. This PR updates the usage stats logic to count intercations based on event logs only from Cody agent. It also now covers event logs from cody web. 

## Test plan

use this query: https://sourcegraph.com/api/console#%7B%22query%22%3A%22%7B%5Cn%20%20currentUser%20%7B%5Cn%20%20%20%20id%5Cn%20%20%20%20codyCurrentPeriodChatUsage%5Cn%20%20%20%20codyCurrentPeriodChatLimit%5Cn%20%20%20%20eventLogs(first%3A%2050)%20%7B%5Cn%20%20%20%20%20%20nodes%20%7B%5Cn%20%20%20%20%20%20%20%20name%5Cn%20%20%20%20%20%20%20%20source%5Cn%20%20%20%20%20%20%20%20user%20%7B%5Cn%20%20%20%20%20%20%20%20%20%20id%5Cn%20%20%20%20%20%20%20%20%7D%5Cn%20%20%20%20%20%20%20%20timestamp%5Cn%20%20%20%20%20%20%7D%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%7D

### DIFF
--- a/cmd/frontend/graphqlbackend/user.go
+++ b/cmd/frontend/graphqlbackend/user.go
@@ -241,7 +241,40 @@ func (r *UserResolver) CodyCurrentPeriodChatUsage(ctx context.Context) (int32, e
 		return 0, err
 	}
 
-	query := sqlf.Sprintf(`WHERE user_id = %s AND timestamp >= %s AND timestamp <= NOW() AND (name LIKE '%%recipe%%' OR name LIKE '%%command%%' OR name LIKE '%%chat%%')`, r.user.ID, currentPeriodStartDate.Time)
+	query := sqlf.Sprintf(`
+		WHERE
+			user_id = %s
+			AND timestamp >= %s
+			AND timestamp <= NOW()
+			AND name NOT LIKE '%%clicked%%'
+			AND name NOT LIKE '%%resetChat%%'
+			AND name NOT LIKE '%%menu:opened%%'
+			AND name NOT LIKE 'cody.%%'
+			AND (
+					(
+						SOURCE = 'IDEEXTENSION'
+						AND (
+							name LIKE '%%recipe%%'
+							OR name LIKE '%%command%%'
+							OR name LIKE '%%chat:executed%%'
+							OR name LIKE '%%chat-question:executed%%'
+							OR name LIKE '%%inline-chat%%'
+							OR name LIKE '%%inline-assist%%'
+							OR name LIKE '%%chat:submitted%%'
+						)
+					)
+
+					OR
+
+					(
+						SOURCE = 'WEB'
+						AND (
+							name = 'web:codyChat:submit'
+							OR name LIKE 'web:codyChat:recipe:%%'
+						)
+					)
+			)
+	`, r.user.ID, currentPeriodStartDate.Time)
 
 	count, err := r.db.EventLogs().CountBySQL(ctx, query)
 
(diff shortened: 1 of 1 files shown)


client/web/src/repo/blob/codemirror/tooltips/LoadingTooltip.ts
```diff
@@ -12,3 +12,12 @@ export class LoadingTooltip implements Tooltip {
...
 }
+
+export const loadingTooltip = (pos: number, end: number) => new LoadingTooltip(pos, end)
+
+export const isLoadingTooltip = (tooltip: Tooltip): tooltip is LoadingTooltip => tooltip instanceof LoadingTooltip
+
+// write a function to add 2 numbers
+function add(a: number, b: number): number {
+    return a + b
+}
```
vdavid, line 23:
```suggestion
```

vdavid (approved):

kelsey-brown:
This will work for suggestions, but the telemetry around chats and commands is more nuanced. For example, I just ran this query against our raw events table and found that it returns the following events, which would not AFAIK count towards the rate limit, would they? 

- CodyVSCodeExtension:command:resetChat:executed
- CodyVSCodeExtension:command:menu:opened
- CodyVSCodeExtension:chatTitleButton:clicked
- CodyVSCodeExtension:chatResponse:noCode
- CodyVSCodeExtension:chatResponse:hasCode
- CodyVSCodeExtension:chatReset:executed
- CodyVSCodeExtension:chatPredictions:used
- CodyVSCodeExtension:chatPanelsManger:activated

thenamankumar:
Thanks for the testing and helping over the zoom call.. The rate limits are not set up based on the event logs as discussed. 

I have come up with this [query](https://redash.sgdev.org/queries/249/source), which only includes the intended events without leaving out any. 

Updated the PR as well.

kelsey-brown:
It looks like that's still including some chat events that we don't want (e.g. the hasCode events) and the command:menu event. [This](https://redash.sgdev.org/queries/249/source) is a bit closer. I have to jump into an interview but I'll revisit after.

thenamankumar:
> Thanks you two.
> 
> Do we want to include `clicked` events, or just `executed` and `called` events? My understanding is that `clicked` doesn't actually correlate to any actual usage (aka LLM requests).

It should only be executed & called events.

chenkc805:
I agree Naman, that's what I think too but I'm seeing `clicked` events in @kelsey-brown 's query: 
Screenshot 2023-12-12 at 9 51 29 AM

kelsey-brown:
I removed click events from the query. This [updated query](https://redash.sgdev.org/queries/251/source) also removes v2 telemetry events and a couple command events that aren't really commands but weren't excluded in the query as its written here. 

@thenamankumar what about commands that have a `called` event and an `executed` event? Should both be included? For example: 

- CodyVSCodeExtension:command:explain:called
- CodyVSCodeExtension:command:explain:executed

thenamankumar:
Thanks @kelsey-brown I have updated the PR with your latest query and it seems good.

We do not need to count both `CodyVSCodeExtension:command:explain:called` and `CodyVSCodeExtension:command:explain:executed`.

For the sake of saving time as the current query is definitely wrong, I'll merge this change now and we can address any further feedback separately.

PR #58931, opened 2023-12-12
Gitlab Native integration: Fix code element target selector

vovakulikov:
Thanks to @taras-yemets this PR fixes code target selector since Gitlab changed their markup about code blocks recently. 

## Test plan
- Check that code intel tooltip shows up in Gitlab code view

### DIFF
--- a/client/browser/src/shared/code-hosts/gitlab/domFunctions.ts
+++ b/client/browser/src/shared/code-hosts/gitlab/domFunctions.ts
@@ -9,7 +9,9 @@ const getSingleFileCodeElementFromLineNumber = (
 ): HTMLElement | null => codeView.querySelector<HTMLElement>(`#LC${line}`)
 
 export const singleFileDOMFunctions: DOMFunctions = {
-    getCodeElementFromTarget: target => target.closest('div.line'),
+    // We have to support div-like line markup and span-like line markup since
+    // different GitLab versions have different code layout markup.
+    getCodeElementFromTarget: target => target.closest('div.line, span.line'),
     getLineNumberFromCodeElement: codeElement => {
         const line = codeElement.id.replace(/^LC/, '')
         return parseInt(line, 10)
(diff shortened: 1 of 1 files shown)


taras-yemets (changes requested):
I just noticed that some versions of GitLab still have "old" markup. Check out [our GitLab instance](https://gitlab.sgdev.org/warren/about/-/blob/main/cypress/support/e2e.ts), for example. 
We need to support both versions.

vovakulikov:
@taras-yemets I added universal selector for div and span likes lines. How do you test extensions on our GitLab instance that you mentioned, do you just run your private local instance? I tried default sourcegraph.com but it doesn't support custom cloud instances. S2 and k8s also failed to connect

taras-yemets (approved):

vovakulikov:
@taras-yemets Yeah, I got this, but I think it still doesn't work or I did something wrong. I signed in on GitLab instance and run local browser extension but got this message about integration part 

>Sourcegraph code host integration: stopped initialization since gitlab.sgdev.org is not a supported code host when Sourcegraph URL is https://sourcegraph.com.
 List of supported code hosts on https://sourcegraph.com: github.com, gitlab.com
 

Screenshot 2023-12-12 at 15 00 16

vovakulikov:
I'm going ahead and try to merge this and release the new version of extension since I think that this change doesn't break anything.

taras-yemets:
> How do you test extensions on our GitLab instance that you mentioned, do you just run your private local instance? I tried default sourcegraph.com but it doesn't support custom cloud instances. S2 and k8s also failed to connect

Sorry, @vovakulikov, I misunderstood your message at first. To test it on https://gitlab.sgdev.org/, we need to add a corresponding code host connection to the Sourcegraph instance. I remember there was one on the k8s instance some time ago, and I also added one to my local instance.

PR #58932, opened 2023-12-12
SSC: Handle model permissions correctly

vdavid:
Problem: I forgot to update one of the model lists in https://github.com/sourcegraph/sourcegraph/pull/58817. As a result, we got mistakenly limited with some of the models.

Note: I know it's not great to have these lists duplicated (see the changes to see what I'm talking about), but I wanted to minimize this change, so just left a reminder to don't forget to update the other list when making changes.

## Test plan

Tested it with a Free and a Pro user, and it returned the right models.

### DIFF
--- a/cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go
+++ b/cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go
@@ -273,7 +273,7 @@ func getCompletionsRateLimit(ctx context.Context, db database.DB, userID int32, 
 		}
 		isProUser := user.CodyProEnabledAt != nil
 		// Update the allowed models based on the user's plan.
-		models = allowedModels(scope,isCodyProEnabled, isProUser)
+		models = allowedModels(scope, isCodyProEnabled, isProUser)
 		intervalSeconds, limit, err = getSelfServeUsageLimits(scope, isProUser, *cfg)
 		if err != nil {
 			return licensing.CodyGatewayRateLimit{}, graphqlbackend.CodyGatewayRateLimitSourcePlan, err
--- a/cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go
+++ b/cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go
@@ -235,13 +235,12 @@ func getCompletionsRateLimit(ctx context.Context, db database.DB, userID int32, 
 	var err error
 
 	isCodyProEnabled := featureflag.FromContext(ctx).GetBoolOr("cody-pro", false)
-	// For Cody PLG users, allow all models a Pro user can get for now.
-	models := allowedModels(scope, isCodyProEnabled, true)
 
 	// Apply override for testing if set.
 	if featureflag.FromContext(ctx).GetBoolOr("rate-limits-exceeded-for-testing", false) {
 		return licensing.CodyGatewayRateLimit{
-			AllowedModels:   models,
+			// For this special tester user, just allow all models a Pro user can get.
+			AllowedModels:   allowedModels(scope, true, true),
 			Limit:           1,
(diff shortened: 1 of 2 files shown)


cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go
```diff
@@ -228,13 +229,16 @@ func getEmbeddingsRateLimit(ctx context.Context, db database.DB) (licensing.Cody
...
 
+// getCompletionsRateLimit returns a rate limit for the given user and feature.
 func getCompletionsRateLimit(ctx context.Context, db database.DB, userID int32, scope types.CompletionsFeature) (licensing.CodyGatewayRateLimit, graphqlbackend.CodyGatewayRateLimitSource, error) {
 	var limit *int
 	var err error
 
-	models := allowedModels(scope)
+	isCodyProEnabled := featureflag.FromContext(ctx).GetBoolOr("cody-pro", false)
+	// For Cody PLG users, allow all models a Pro user can get for now.
+	models := allowedModels(scope, isCodyProEnabled, true)
```
rafax, line 239: Why do you pass `true` as the last param - shouldn't this depend on something like `user.isCodyPro` ?

rafax: If this is only used for `rate-limits-exceeded-for-testing`, then I'd move it to that `if` (so it's clear that it is only used in that single exceptional situation.

rafax (approved):

rafax (changes requested):
Claude-v1 missing

cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go
```diff
@@ -328,10 +334,27 @@ func getSelfServeUsageLimits(scope types.CompletionsFeature, isProUser bool, cfg
...
+			return []string{"anthropic/claude-v1", "anthropic/claude-2", "anthropic/claude-2.0", "anthropic/claude-2.1", "anthropic/claude-instant-v1", "anthropic/claude-instant-1"}
+		}
+
+		// When updating the below lists, make sure you also update `isAllowedCustomChatModel` in `chat.go`
+
+		if !isProUser {
+			return []string{"anthropic/claude-2.0"}
+		}
+
+		return []string{
```
rafax, line 350: Should "anthropic/claude-v1" be included here?

vdavid: No because this is chat only and we don't use that model for that.

cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go
vdavid: Thanks, updated the code for clarity, and re-tested it!

cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go
rafax: Ok, looks like [usage](https://redash.sgdev.org/queries/248/source#500) has died down in November, SGTM.

rafax: I'd keep the two lists consistent (to the point where we may want to use the same variable for both), but up to you.

rafax (approved):

cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go
vdavid: We'll remove the old list when we remove the `cody-pro` flag in a few weeks anyway, so I'll err on the side of speed in this.

PR #58938, opened 2023-12-12
Copy changes and new video

danielmarquespt:
Links to new videos with better animation

Changes copy in the Dashboard and success modal

## Test plan

Tested local in Light and Dark mode

### DIFF
--- a/client/web/src/cody/subscription/UpgradeToProModal.tsx
+++ b/client/web/src/cody/subscription/UpgradeToProModal.tsx
@@ -28,11 +28,11 @@ export function UpgradeToProModal({
             {data?.changeCodyPlan?.codyProEnabled ? (
                 <div className="d-flex flex-column justify-content-between align-items-center mby-4 py-4">
                     <CodyColorIcon width={40} height={40} className="mb-4" />
-                    <H2>Pro Membership: Confirmed!</H2>
-                    <Text>Welcome to the exclusive zone of coding</Text>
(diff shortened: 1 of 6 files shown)


client/web/src/cody/subscription/UpgradeToProModal.tsx
```diff
@@ -28,11 +28,11 @@ export function UpgradeToProModal({
             {data?.changeCodyPlan?.codyProEnabled ? (
                 <div className="d-flex flex-column justify-content-between align-items-center mby-4 py-4">
                     <CodyColorIcon width={40} height={40} className="mb-4" />
-                    <H2>Pro Membership: Confirmed!</H2>
-                    <Text>Welcome to the exclusive zone of coding</Text>
+                    <H2>Upgraded to Cody Pro 🎉</H2>
+                    <Text>You now have unlimited autocomplete suggestions, chat messages and commands.</Text>
```
chenkc805, line 32: Good catch, this copy is way better (although the old copy was kind of cute and growing on me) 😆

chenkc805 (approved):

client/web/src/cody/subscription/UpgradeToProModal.tsx
danielmarquespt: The old copy made more sense with the Badge, because you were unlocking a new "zone" so you are a VIP kinda of thing.

We can go back to it with the fancy Badge

PR #58942, opened 2023-12-12
actor: always preserve anonymousUID, even on authenticated actors

bobheadxi:
In the backend, the anonymous user ID is discarded if there is an authenticated actor. There is some importance to preserving both to follow the path of an unauthenticated user to an authenticated one - this change updates cookie, HTTP, and gRPC actor propagation to always look for an anonymous user ID and preserve it, even if there is some other form of actor already.

I also did a review of references to `(Actor).AnonymousUID` and there doesn't seem to be any paths that override authenticated user behaviour with `AnonymousUID` - callsites typically check for `UID` first. I've added a docstring that indicates this should always be done.

## Test plan

Existing unit tests and some new cases covering authenticated with anonymous ID scenarios.


nathan-downs (approved):
lgtm

PR #58943, opened 2023-12-12
Search: respect the `search.contextLines` setting in the backend

camdencheek:
Currently, we do not respect the `search.contextLines` setting in the backend. Specifically, the search backend currently always yields results with zero lines of context. 

One effect of this is in order to display matches with context lines, any client needs to make a followup request to fetch the full contents of the file. This makes the UI feel laggy because results filter in slowly as you scroll. This is exacerbated by the fact that we load the highlighted code, and highlighting can be unpredictable, sometimes taking a couple seconds to return.

We already stream the matched chunk back to the client, so this just updates the backend so that the streamed results include the number of context lines a user requested. Zoekt already supports this, so it was just a matter of taking advantage of that setting and updating searcher to do the same.

The tradeoff is that we're streaming more content through the backend. I do not expect this to cause significant instability, but I will monitor when it is deployed. Ideally, there would be a way to signal to our backends to stop streaming back content once we've hit the display limit. I considered an alternative where we could fetch from gitserver as we stream back, but that's a lot of extra load on gitserver when we already have the information we need. 

Of note here is that, with the work @jtibshirani is doing to add and/or support to searcher, we will rarely need to collect the result set in memory (since we no longer need to join in `frontend`), which should help with memory spikes.

## Test plan

Added a bunch of unit tests for the line mucking stuff for adding context lines. Added a few searcher tests. This is extracted from another branch where I manually tested the feature end-to-end with the web client changes as well.


keegancsmith:
> One effect of this is in order to display matches with context lines, any client needs to make a followup request to fetch the full contents of the file. This makes the UI feel laggy because results filter in slowly as you scroll. This is exacerbated by the fact that we load the highlighted code, and highlighting can be unpredictable, sometimes taking a couple seconds to return.

Don't we need the full file to do highlighting? If not this change makes sense. Otherwise this change doesn't really make a difference does it?

> The tradeoff is that we're streaming more content through the backend. I do not expect this to cause significant instability, but I will monitor when it is deployed. 

This is my main concern. Especially since there are a lot of uses of our search backend which doesn't even care about the previews (insights, most API uses, search jobs, once we hit display limit), or if they are used don't use the context lines (src search).

However, I do like the idea of optimizing for our main use case of interactive querying.

Zoekt already has a display limit, I wonder if we are correctly respecting it / using it properly. This does seem like something we should correctly use to avoid concerns. Additionally, an option to disable previews would likely also help a bunch of use cases.

> ## Test plan
> Added a bunch of unit tests for the line mucking stuff for adding context lines. Added a few searcher tests. This is extracted from another branch where I manually tested the feature end-to-end with the web client changes as well.

My concerns are mostly around automated uses where result sets may be larger but never displayed just have aggregated statistics.

cmd/searcher/internal/search/zoekt_search.go
```diff
@@ -77,6 +77,7 @@ func zoektSearch(ctx context.Context, logger log.Logger, client zoekt.Streamer,
 
 	searchOpts := (&search.ZoektParameters{
 		FileMatchLimit: args.FileMatchLimit,
+		// TODO: numContextLines
```
keegancsmith, line 80: don't we need this since hybrid searcher is the most usual code path on searcher?

cmd/searcher/protocol/searcher.go
```diff
@@ -52,6 +52,8 @@ type Request struct {
 	// Whether the revision to be searched is indexed or unindexed. This matters for
 	// structural search because it will query Zoekt for indexed structural search.
 	Indexed bool
+
+	NumContextLines int32
```
keegancsmith, line 56: docs?

keegancsmith (approved):
Do you have this working e2e doing less work? This PR seems to make us do more work, but it isn't actually used and we still speak to gitserver.

Maybe you can hide behind a feature flag setting NumContextLines to the user settting, and instead use 0 if not in the feature flag. This would also be useful operationally / can be the same feature flag for using preview instead of gitserver.

camdencheek:
> Don't we need the full file to do highlighting?

Yes, we do. However, the web client makes a separate GraphQL request which fetches the full file content from `gitserver`, sends it to `syntax-highlighter`, and returns the highlighted HTML to the browser. This can take quite a long time (seconds), especially for large files.

We don't need the full file to display the plaintext match while we're waiting for highlighting. We just need the chunk content with context lines. That's the goal of this: include the context lines so we can show something while we wait for highlighting. This allows us to render the results immediately as they're streamed rather than rendering an empty box. 

> Zoekt already has a display limit

Does it work the same way the `frontend` display limit does? Reading the doc string, it sounds like zoekt's definition of "display limit" is just "truncate the results after sorting" which is not quite what we're looking for since we still need the matched file names and stuff for aggregations. Maybe we could return richer aggregated stats from Zoekt so we can still get the progress updates without streaming any of the results? Just spitballing.

> Do you have this working e2e doing less work? 

I have a draft [here](https://github.com/sourcegraph/sourcegraph/pull/58979) that takes advantage of this. I split the change up into frontend/backend because I expect the reviewers will be quite different 🙁 

> Maybe you can hide behind a feature flag setting NumContextLines to the user settting

Yeah, I can add this. The reason I hadn't is that, with the web client changes, turning the setting off will just show results with zero lines of context now since we the client now trusts the backend to return the right chunk ranges. I briefly looked at trying to implement the client changes in a way that maintained the old way of doing things (just use the match ranges and reconstruct the chunks in the client), but the client code is quite tangled and I ended rewriting/removing large chunks to simplify.

I'll add the flag anyways so we can compare perf.

> My concerns are mostly around automated uses where result sets may be larger but never displayed just have aggregated statistics.

I'll make sure that we set zero context lines for insights. We do still need the content for insights for capture group extraction, but context lines will always just be bloat. What other uses are you thinking? I'm thinking we should add `contextLines` to the streaming API so it can be controlled by the client for things like `src`

camdencheek:
For reference, [here](https://sourcegraph.sourcegraph.com/search?q=context:global+file:.json%24+search_results&patternType=newStandardRC1&sm=1) is a search where this problem is _very_ visible. The files we're returning matches for are large, so it takes quite a while for the highlighting to return, and it usually times out for the largest file. This is a case where, even though the search itself is ~instant, the user doesn't experience that.

camdencheek:
Okay, I addressed some of the feedback in the latest commit. 
- We can now set context lines dynamically with the `cm` query parameter on the stream endpoint. This allows clients like `src` (or maybe even the web client) to request custom amounts of context.
- Consumers of the search client in the cluster can now decide whether they want context. For our computer-oriented clients like compute, insights, search jobs, and code monitors, I've set the number of context lines to zero so we're not wasting memory during aggregation.
- We can toggle the default number of context lines globally with the setting `search.contextLines`.

keegancsmith:
> We don't need the full file to display the plaintext match while we're waiting for highlighting. We just need the chunk content with context lines. 

Aah nice.

> Reading the doc string, it sounds like zoekt's definition of "display limit" is just "truncate the results after sorting" which is not quite what we're looking for since we still need the matched file names and stuff for aggregations. Maybe we could return richer aggregated stats from Zoekt so we can still get the progress updates without streaming any of the results? Just spitballing.

Yes that is exactly what we need to do. Would also result in a lot less work for zoekt. This is what I had in mind when I was hinting that :)

> > My concerns are mostly around automated uses where result sets may be larger but never displayed just have aggregated statistics.
> 
> I'll make sure that we set zero context lines for insights. We do still need the content for insights for capture group extraction, but context lines will always just be bloat. What other uses are you thinking? I'm thinking we should add `contextLines` to the streaming API so it can be controlled by the client for things like `src`

I think that is maybe what we should do? It is an API input rather than something we read from user settings. But for now don't let that block you. Your answers to everything else now makes this make a lot of sense in my head.

camdencheek:
> I think that is maybe what we should do?

Done 🙂 The streaming API can now take the number of context lines as a parameter, and the search planner accepts that as a parameter as well, so I set it to zero lines in non-human-facing contexts.

PR #58944, opened 2023-12-12
graphqlbackend/telemetry: allow client-provided timestamp in events

bobheadxi:
Generating the timestamp serverside can cause client-generated events to appear out-of-order, and in general will have a gap between when an event actually occurred and the timestamp that gets persisted.

This change only fixes it for V2 telemetry - the V1 `logEvent` mutation doesn't accept a timestamp, and while the V1 `logEvents` GraphQL schema does accept a timestamp, [the resolver input implementation does not use it](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/cmd/frontend/graphqlbackend/user_usage_stats.go?L85:1-109:2).

I will update https://github.com/sourcegraph/telemetry after this PR, and add handling to VSCode.

## Test plan

Unit tests

### DIFF
--- a/cmd/frontend/graphqlbackend/telemetry.go
+++ b/cmd/frontend/graphqlbackend/telemetry.go
@@ -2,6 +2,8 @@ package graphqlbackend
 
 import (
 	"context"
+
+	"github.com/sourcegraph/sourcegraph/internal/gqlutil"
 )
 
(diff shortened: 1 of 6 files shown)


nathan-downs (approved):
Thank you for catching this and saving us data headaches later!

PR #58953, opened 2023-12-13
[Cody] fix limit reached UI

thenamankumar:
closes: https://github.com/sourcegraph/cody/issues/2310
closes: https://github.com/sourcegraph/cody/issues/1931
closes: https://github.com/sourcegraph/cody/issues/2230
closes: https://github.com/sourcegraph/cody/issues/2309
closes: https://github.com/sourcegraph/cody/issues/2276

## Test plan

CleanShot 2023-12-13 at 12 03 25@2x

### DIFF
--- a/internal/database/users.go
+++ b/internal/database/users.go
@@ -647,11 +647,8 @@ func (u *userStore) ChangeCodyPlan(ctx context.Context, id int32, pro bool) (err
 			return userNotFoundErr{args: []any{id}}
 		}
 
-		if pro {
-			return errors.New("user is already on Cody Pro plan")
-		}
-
-		return errors.New("user is already on Cody Community plan")
+		// Intentionally not returning an error if the user is already pro/commiunity. This makes the mutation idempotent.
+		return nil
 	}
 
 	return nil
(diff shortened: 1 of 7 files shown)


internal/database/users.go
```diff
@@ -647,11 +647,8 @@ func (u *userStore) ChangeCodyPlan(ctx context.Context, id int32, pro bool) (err
 			return userNotFoundErr{args: []any{id}}
 		}
 
-		if pro {
-			return errors.New("user is already on Cody Pro plan")
-		}
-
-		return errors.New("user is already on Cody Community plan")
+		// Intentionally not returning an error if the user is already pro/commiunity. This makes the mutation idempotent.
```
vdavid, line 650:
```suggestion
		// Intentionally not returning an error if the user is already Pro/Free. This makes the mutation idempotent.
```

vdavid (approved):
Went through it, it all LGTM.

internal/database/users.go
thenamankumar: For the sake of not waiting for another 25 mins for tests to run, ignoring this for now.

Issue #58548, opened 2023-11-24
Error 'could not find token' when doing Find references


varungandhi-src:
https://sourcegraph.com/github.com/TadasBaltrusaitis/OpenFace/-/blob/matlab_version/CCNF/sparsity_grid.m?L18:54-18:70#tab=references

![](https://github.com/sourcegraph/sourcegraph/assets/93103176/ba85e498-26d9-473c-97b3-4a433a4677aa)

varungandhi-src:
The problem is (again) language detection. The code looks like this:

```
    const languageId = getModeFromPath(props.token.filePath)
    const spec = findLanguageSpec(languageId)
    const tokenResult =
        spec &&
        findSearchToken(...)
```

https://sourcegraph.com/github.com/sourcegraph/sourcegraph@cc03635fdd41d5a829e81e7b50ac6492299bb604/-/blob/client/web/src/codeintel/ReferencesPanel.tsx?L226:5

1. We end up getting an incorrect `languageId` of `objectivec`.
2. `findLanguageSpec` fails for `objectivec`, returning `undefined` for `spec`.
3. `tokenResult` is `undefined`, so we end up hitting the "Could not find token" error.

Issue #56376, opened 2023-09-06
☂️ Centralize language detection


varungandhi-src:
- [ ] https://github.com/sourcegraph/sourcegraph/issues/7841
- [ ] Remove TypeScript code paths doing language detection based on file extension alone
- [ ] Remove Squirrel code paths doing language detection based on file extension alone
- [ ] Provide centralized Go package which handles language detection, file extension mappings etc.

peterguy:
I made this sam comment on #56374 but it may be more appropriate here:

I’ve done some work around using file content for language detection, caching and persisting the result using the database and redis. The implementation is currently lazy, but the framework can be leveraged by background jobs to pre-calculate and persist language detection. See https://github.com/sourcegraph/sourcegraph/tree/pbg/determine-file-type-using-content

varungandhi-src:
Above, the first bullet point that I mentioned says "Provide centralized Go package". I wonder if instead of a Go baseline, we should start with Rust. Specifically, we can wrap and/or fork the [hyperpolyglot](https://crates.io/crates/hyperpolyglot) crate. Not only is it faster than enry in benchmarks, utilizing Rust would allow us to consume the same library from both Rust (syntax-highlighter etc.), Go (via CGo) and if necessary TypeScript (via WASM).

Right now, we do need a Rust library for batch indexing/language detection, and that's not coupled to the network directly, we're just operating on a folder of files.

For situations where that's not the case, especially when we're handling a single file, I think it'd be better to propagate the language type information consistently instead of re-computing it on-the-fly, because using different settings/sets of heuristics in different places may lead to bugs that would be hard to track down.

PR #58954, opened 2023-12-13
client: Avoid complex tokenization in ref panel code

varungandhi-src:
Previously, we relied on detecting the language from file paths,
then using various regexes associated with the language to identify
token boundaries. However, the code mirror blob view always provides
a full token range, which can be used directly, instead of attempting
to recompute the token boundaries.

For older URLs, we fallback to simple identifiers, which should
work for the vast majority of languages and identifiers.

We cannot yet remove the language detection here because the file
extensions associated with the language are later used for search-based
code navigation.

This patch also makes the language spec optional for search-based
code intel, as we do not have a solution to #56376 which would
guarantee that we always have a language available. If a language
is not available, search-based code intel falls back to searching
other files with the same extension as a best effort guess.

Fixes https://github.com/sourcegraph/sourcegraph/issues/58548

## Test plan

Locally tested for MATLAB code. The ref panel shows up correctly,
unlike the error earlier in #58548

CleanShot 2023-12-13 at 15 25 17@2x

### DIFF
--- a/client/shared/src/languages.ts
+++ b/client/shared/src/languages.ts
@@ -3,9 +3,22 @@ import { basename } from 'path'
 /** The LSP mode used for plain text files and all other unrecognized files. */
 const PLAINTEXT_MODE = 'plaintext'
 
+// FIXME(id: language-detection)
+//
+// For correctness, we should do language detection on the server
+// and use the inferred language on the client instead of incorrectly
+// doing language detection on the client using incomplete information
+// like file extensions.
+//
+// For example, MATLAB and Objective-C both use the '.m' file extension.
+//
+// See https://github.com/sourcegraph/sourcegraph/issues/56376
+
 /**
  * getModeFromPath returns the LSP mode for the provided file path. If the file path does not correspond to any
  * known mode, 'plaintext' is returned.
+ *
+ * DO NOT introduce new usages of this API. See FIXME(id: language-detection).
  */
 export function getModeFromPath(path: string): string {
     const fileName = basename(path)
--- a/client/shared/src/languages.ts
+++ b/client/shared/src/languages.ts
@@ -18,7 +18,7 @@ const PLAINTEXT_MODE = 'plaintext'
  * getModeFromPath returns the LSP mode for the provided file path. If the file path does not correspond to any
  * known mode, 'plaintext' is returned.
  *
- * DO NOT introduce new usages of this API. See FIXME(id: language-detection).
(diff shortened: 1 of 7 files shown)


varungandhi-src:
There is an off-by-one error for the end of the span.



Also, weirdly enough, we're not picking up results in Mathematica files but only in MATLAB and Objective-C files. 🤔

client/shared/src/languages.ts
```diff
@@ -3,9 +3,22 @@ import { basename } from 'path'
...
+//
+// For example, MATLAB and Objective-C both use the '.m' file extension.
+//
+// See https://github.com/sourcegraph/sourcegraph/issues/56376
+
 /**
  * getModeFromPath returns the LSP mode for the provided file path. If the file path does not correspond to any
  * known mode, 'plaintext' is returned.
+ *
+ * DO NOT introduce new usages of this API. See FIXME(id: language-detection).
```
fkling, line 21: You could mark this function as `@deprecated`. Editors usually pick this up.

client/web/src/codeintel/ReferencesPanel.tsx
```diff
@@ -99,9 +99,14 @@ interface State {
     }
 }
 
+interface OneBasedPosition {
+    line: number
+    character: number
+}
```
fkling, line 105: I've also run into quite a few issues with 0-based vs 1-based positions. I think it's useful to have separate types to express intent, but just note that the type checker won't prevent you from passing a `ZeroBasedPostition` as a `OneBasedPosition`, because TS uses structural equivalence (or whatever it is called).

fkling (approved):

client/web/src/codeintel/ReferencesPanel.tsx
varungandhi-src: Do you think I should use classes here? I'd be happy to introduce new vocabulary types for Positions and Ranges in a central place that can be reused elsewhere.

client/shared/src/languages.ts
varungandhi-src: Good point, I didn't consider that! Done.

client/web/src/codeintel/ReferencesPanel.tsx
varungandhi-src: OK, let me attempt to do that in a follow-up PR.

varungandhi-src:
Fixed the off-by-one issue

CleanShot 2023-12-14 at 19 16 09@2x

PR #58957, opened 2023-12-13
[Cody] disable dotcom level rate limitting for Cody GA

thenamankumar:
Disable dotcom level user-based rate limiting for Cody GA. We need to only rely on Cody gateway which performs user-level rate limiting now. 

## Test plan

.


vdavid (approved):
We did this paired, it's a small change, solves the problem, and it looks safe to me.

PR #58962, opened 2023-12-13
client: Get language(s) from server for search-based code intel

varungandhi-src:
Before this PR:
For search-based code intel queries, the client code attempts to
infer the language from the file path, maps those to a list of
file extensions, and then performs queries for search-based code intel.

After this PR:
The backend provides a list of potential languages for the blob.
The client just reuses that list to construct a `lang:` filter,
without futzing around with file extensions.

In the future, once language detection takes content into
account correctly, search-based code intel will automatically
also benefit from it.

## Test plan

Test that search-based code intel still works.


cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -5811,6 +5811,10 @@ interface File2 {
         """
         endLine: Int
     ): HighlightedFile!
+    """
+    The language(s) determined based on the file's contents, paths etc.
+    """
```
kritzcreek, line 5816: I'd maybe adjust this comment to clarify that multiple languages here means 'ambigious' rather than multiple languages detected in the same file.

fkling (commented):
It seems a bit weird to me to "attach" languages to `CodeMirrorBlob`, only to have some external code to extract it (in `ReferencePanel.tsx`).
In a way CodeMirrorBlob is just a messenger.

I understand that atm `ReferencePanel` has no access to this data because it's loaded inside `SideBlob`. Maybe that's an indication that the data should be fetched higher up in the component tree, not in `SideBlob`.

It would be OK for me if `languages` is going to be used inside CodeMirrorBlob extensions in the future.

varungandhi-src:
The main thing I wanted to change was to have an extra `languages` be tracked in `BlobPage.tsx` so that the ref panel has access to the languages setting for the _main_ blob view. The SideBlob change came about purely because I was following the type errors.

client/web/src/enterprise/codeintel/searchBased.ts
```diff
@@ -58,27 +57,20 @@ export function referencesQuery({
...
- * include list, an empty file term will be returned.
- *
- * @param textDocument The current text document.
- * @param includelist The file extensions for the current language.
- */
-function fileExtensionTerm(path: string, includelist: string[]): string {
+
+function languageFilter(languages: string[], path: string): string[] {
+    if (languages.length > 0) {
+        return [languages.map(language => `lang:${language}`).join(' OR ')]
```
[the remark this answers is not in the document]
varungandhi-src: Thanks, fixed

client/web/src/codeintel/ReferencesPanel.tsx
```diff
@@ -255,6 +253,9 @@ const SearchTokenFindingReferencesList: React.FunctionComponent<
         )
     }
 
+    const blobView = BlobAPI.getBlobEditView()
+    const languages = blobView?.state.facet(blobPropsFacet).blobInfo.languages ?? []
```
fkling, line 257: I still wished this could be done differently, but I don't want to block you on that.

fkling (approved):
OK, that makes sense, thank you!

client/web/src/codeintel/ReferencesPanel.tsx
varungandhi-src: If you have alternate suggestions, I'm open to it. I know that facet is described as a hack, but I didn't have an alternate approach, that's why I went with this.

fkling: > If you have alternate suggestions, I'm open to it.

Nothing atm besides having `ReferencesPanel` fetch the data and pass it to `SideBlob`. But that might be ugly too.

varungandhi-src: OK, I'll try it in a bit and see how that goes. I agree for the side blob, it would make sense for the ref panel to have the data (e.g. in case it wants to use language icons for files)

varungandhi-src: I looked at this code again, and I think the current solution is actually a better fit, because the side blob is fetching the blob contents anyways, so it makes sense for it to retrieve the language at the same time in the same request. 

If we do at the level of the ref panel, it would complicate things quite a bit more. We'd have to essentially get the language when doing `fetchHighlightedFileRanges`, then propagate that language "backwards" via `setActiveLocation` (or a separate callback) all the way from `CollapsibleLocationGroup` to `ReferencesList` so that we can pass it to `SideBlob`.

PR #58979, opened 2023-12-14
Search Web: render results immediately using streamed content

camdencheek:
This PR builds on the change in https://github.com/sourcegraph/sourcegraph/pull/58943 to take advantage of the newly-complete chunks being streamed back from the backend.

This PR is fairly large. Mostly because there is a ton of code for chunking, limiting, and ranking results that is just completely irrelevant now since we can just depend on what we send from the backend.

Before:

https://github.com/sourcegraph/sourcegraph/assets/12631702/4d56e476-1772-4d74-aa07-96e7bc07ba9b

After:

https://github.com/sourcegraph/sourcegraph/assets/12631702/4433304d-2635-4e1a-8b74-bbee5e7a451f

# Test plan

Updated the tests, deleted a bunch of obsolete tests for removed functionality, and lots of manual testing.


client/shared/src/components/ranking/ZoektRanking.ts
camdencheek, line 1: This whole file can go away because we can just use the chunk sizes returned from the backend. This whole file is a bunch of (historically buggy) logic to add context lines, merge chunks, and update ranking appropriately. None of that is necessary if the returned chunks already respect the context lines setting.

client/shared/src/components/ranking/LineRanking.ts
camdencheek, line 1: This file can also go away, replaced by a simple "sort by line" function in the file that uses it. Similar to `ZoektRanking.ts`, this whole file is rendered obsolete when we already have good chunks coming from the backend.

client/shared/src/components/ranking/PerFileResultRanking.ts
```diff
@@ -44,7 +23,12 @@ export interface MatchItem {
  * Describes a single group of matches.
  */
 export interface MatchGroup {
```
camdencheek, line 25: TODO: move the types out of this file to something that makes more sense now that we don't do any ranking in this file.

client/branded/src/search-ui/components/CodeExcerpt.tsx
```diff
@@ -60,30 +44,38 @@ export interface HighlightRange {
...
-
 /**
  * A code excerpt that displays syntax highlighting and match range highlighting.
  */
 export const CodeExcerpt: React.FunctionComponent<Props> = ({
-    blobLines,
-    fetchHighlightedFileRangeLines,
-    fetchPlainTextFileRangeLines,
+    plaintextLines,
+    highlightedLines,
```
camdencheek, line 52: I refactored this component to do a lot less work. It no longer does any fetching. Instead, it just takes a set of plaintext lines and an optional set of HTML-highlighted lines, and renders the excerpt appropriately. It will be re-rendered once highlightedLines are updated, but that is handled outside the component.

client/branded/src/search-ui/components/CodeExcerpt.tsx
```diff
@@ -149,45 +118,11 @@ export const CodeExcerpt: React.FunctionComponent<Props> = ({
...
-                                </tr>
-                            ))}
-                        </tbody>
-                    </table>
-                )}
-            </Code>
-        </VisibilitySensor>
+        <Code data-testid="code-excerpt" onCopy={onCopy} className={classNames(styles.codeExcerpt, className)}>
+            <div ref={updateTableContainerElementReference}>{table}</div>
+        </Code>
```
camdencheek, line 126: The important changes here:
1) Visibility sensor is moved up to the component that fetches the highlighted lines
2) We no longer render a highlighting error since we will always have a plaintext version to fall back to

client/web/src/codeintel/ReferencesPanel.tsx
```diff
@@ -877,94 +866,101 @@ const CollapsibleLocationGroup: React.FunctionComponent<
...
-                                        event.ctrlKey ||
-                                        event.metaKey ||
-                                        event.shiftKey
-                                    ) {
-                                        return
-                                    }
+                    <VisibilitySensor
+                        onChange={(visible: boolean) => visible && onVisible()}
+                        partialVisibility={true}
+                        offset={{ bottom: -500 }}
```
vovakulikov, line 872: Interesting value, I mean it was like this before this PR just curios why)?

vovakulikov (approved):
Tested it manually. Nothing to add really. Thanks for the change

client/web/src/codeintel/ReferencesPanel.tsx
camdencheek: I'm think this is saying "count it as visible if it's within 500 pixels of the bottom of the page". It essentially works as a small amount of preload.

client/branded/src/search-ui/components/FileContentSearchResult.tsx
```diff
@@ -113,62 +116,58 @@ export const FileContentSearchResult: React.FunctionComponent<React.PropsWithChi
...
-            const contextLinesSetting =
-                isSettingsValid(settingsCascade) && settingsCascade.final?.['search.contextLines']
-
-            if (typeof contextLinesSetting === 'number' && contextLinesSetting >= 0) {
-                return contextLinesSetting
-            }
+    const contextLines = useMemo(() => {
+        const settings = settingsCascade.final
+        if (!isErrorLike(settings)) {
+            return settings?.['search.contextLines'] || 1
```
fkling, line 122: I believe this can be

```
const settings = useSettings();
const contextLines = settings?.['search.contextLines'] ?? 1
```

At the very least `... || 1` should be `... ?? 1`, otherwise users won't be able to set context lines to `0`.
The previous code also verified the type of the value, but I don't know if we really need that.

fkling (approved):
I like red PRs. Thank you for updating the SvelteKit version too!

PR #58980, opened 2023-12-14
Embeddings: avoid constantly rerunning job if it failed

jtibshirani:
The embeddings policy framework attempts to rerun a repo job even if a previous
run failed at the exact same revision. This means that when a job failed, for
example because of rate limits or a problematic file, it would immediately be
rescheduled and fail again. This can be expensive and noisy.

Now, the policy framework does **not** rerun failed jobs unless the revision
changes. An admin can always kick off a job manually if they want to rerun a
job at the revision. This reduces noise and feels like a better trade-off.

## Test plan

Modified unit tests


stefanhengl (approved):
LGTM!

jtibshirani:
Thanks for the review. I'll backport this as part of a round of embeddings fixes.

PR #58987, opened 2023-12-14
client: Update list of languages for lang: filter

varungandhi-src:
Updated per the script in the supporting comment.

```
/**
 * A list of all supported languages, ranking the popular languages higher by
 * including them first.
 * The whole list was extracted from
 * https://github.com/go-enry/go-enry/blob/master/data/alias.go via
 *
   curl "https://raw.githubusercontent.com/go-enry/go-enry/$(grep go-enry/go-enry go.mod|cut -d ' ' -f 2)/data/alias.go" \
   | awk -F : '/^var LanguageByAliasMap/ { MATCH=1; next } /^}/ { exit } MATCH==1 { gsub(/^ +/, "", $2); print $2 }' \
   | sort | uniq
 *
 */
```

## Test plan

n/a


fkling (approved):
I'm glad the script still works 😅

PR #58990, opened 2023-12-14
frontend: remove app resolvers

mrnugget:
This removes the App resolvers from `frontend`. That removes `//cmd/frontend/internal/app/resolvers` from the list of packages required to build `frontend`, which means it should remove references to `servegit`, which requires CGO, which makes our build slower.

## Test command

This times an _incremental rebuild_ of `frontend`:

```bash
git restore cmd/frontend/main.go && \
    rm -rf ~/frontend && \
    rm -rf ~/frontend2 && \
    go build -o ~/frontend github.com/sourcegraph/sourcegraph/cmd/frontend && \
    echo "\n\n func foo() int { return 99 }" >> cmd/frontend/main.go && \
    echo "\n\n\n" && \
    time go build -o ~/frontend2 github.com/sourcegraph/sourcegraph/cmd/frontend &&
    git restore cmd/frontend/main.go
```

## Before

```
  16.55s user 3.95s system 118% cpu 17.257 total
```

## After

```
9.99s user 2.93s system 133% cpu 9.711 total
```

## Test plan

- N/A

## Preview 🤩
[Preview Link](https://docs.sourcegraph.com/@mrn/remove-app-resolverss)


daxmc99:
Might also be good to remove these [refs](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/internal/conf/deploy/deploytype.go?L17:2-17:5#tab=references)

sqs:
The build perf improvement here seems great! Hope someone picks this up.

eseliger:
@daxmc99 @sqs see here https://github.com/sourcegraph/sourcegraph/pull/59115 :)

PR #58996, opened 2023-12-14
Bypass the access token auth handler for license checks

pjlast:
#57160 introduced a regression where requests to `/.api/license/check` all started failing with `401 Unauthorized`. License checks have been failing since 4 October. We never noticed because we disabled any consequences of license checks failing, but we would like these checks to succeed so that we can use the events produced by the checks to detect when licenses are used by multiple instances.

This PR adds a bypass to this handler, similar to the bypass for `/.api/scim/v2`, so that the license check handler checks the authentication instead.

## Test plan

Added unit test

### DIFF
--- a/cmd/frontend/internal/httpapi/auth.go
+++ b/cmd/frontend/internal/httpapi/auth.go
@@ -38,6 +38,13 @@ func AccessTokenAuthMiddleware(db database.DB, baseLogger log.Logger, next http.
 			return
 		}
 
+		// The license check handler uses a Bearer token and request body which
+		// is checked in `productsubscription/license_check_handler.go`
+		if strings.HasPrefix(r.URL.Path, "/.api/license/check") {
+			next.ServeHTTP(w, r)
+			return
+		}
+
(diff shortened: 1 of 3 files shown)


cmd/frontend/internal/httpapi/auth.go
```diff
@@ -38,6 +38,13 @@ func AccessTokenAuthMiddleware(db database.DB, baseLogger log.Logger, next http.
 			return
 		}
 
+		// The license check handler uses a Bearer token and request body which
+		// is checked in `productsubscription/license_check_handler.go`
+		if strings.HasPrefix(r.URL.Path, "/.api/license/check") {
+			next.ServeHTTP(w, r)
+			return
+		}
+
```
evict, line 47: This is Dotcom specific right? Do we need to add a check there?

pjlast: Yeah good idea, thanks

pjlast: Updated and added a unit test as well

eseliger (approved):

PR #58997, opened 2023-12-14
Fix modal onboarding styles

vovakulikov:
This PR fixes scope issue in Cody onboarding modal styles. Prior to this fix onboarding styles overrides global modal styles which affects all modals in the app. This PR just tries to fix the scope issue and apply onboarding styles only on onboarding related UI

Screenshot 2023-12-14 at 11 36 28

## Test plan
- Check that Cody onboarding modal still looks correct
- Check that other modals are not affected by these onboarding styles.


danielmarquespt:
Thanks @vovakulikov! I thought I that using css in the module was enough to keep everything scoped!

vovakulikov:
@danielmarquespt no worries. 

> I thought I that using css in the module was enough to keep everything scoped!

It's correct but only for local-like selectors, like CSS classes for instance. All global like selectors like data attributes (such as data-modal-overlay selector) will be applied for all elements with this data attribute. CSS modules don't change them. Same thing for id selectors.

PR #59004, opened 2023-12-14
svelte: Link individual search results to file page

fkling:
This makes individual search results clickable, like in the React app. Unlike the React app selecting text from search results is not supported. This behavior brings in a lot of complexity and I'd rather spend the time looking for a better solution in general (e.g. maybe text selection in search results isn't necessary anymore if we have a quick preview panel).

Note: I moved the logic from `FileMatchChildren.svelte` directly into `FileContentSearchResult.svelte`. It seemed unnecessary to have this intermediate component.

## Test plan

Open search results page and click on a result.


client/web-sveltekit/src/routes/search/FileContentSearchResult.svelte
```diff
@@ -84,8 +93,20 @@
...
-        <FileMatchChildren {result} grouped={expanded ? expandedMatchGroups.grouped : collapsedMatchGroups.grouped} />
+    <div bind:this={root} class="matches">
+        {#each matchesToShow as group}
+            <div class="code">
+                <a href={getMatchURL(group.startLine + 1, group.endLine)}>
+                    <CodeExcerpt
+                        startLine={group.startLine}
+                        endLine={group.endLine}
+                        fetchHighlightedFileRangeLines={async (...args) =>
+                            group.blobLines ? group.blobLines : fetchHighlightedFileMatchLineRanges(...args)}
```
camdencheek, line 104: Async rather than RxJS makes this sooo much nicer

client/web-sveltekit/src/routes/search/FileContentSearchResult.svelte
```diff
@@ -59,19 +64,23 @@
...
+        )
+        return `${fileURL}?${searchParams}`
+    }
+
+    async function fetchHighlightedFileMatchLineRanges(startLine: number, endLine: number) {
+        const highlightedGroups = await fetchFileRangeMatches({ result, ranges: matchRanges })
+        return highlightedGroups[
+            matchesToShow.findIndex(group => group.startLine === startLine && group.endLine === endLine)
+        ]
     }
```
camdencheek, line 84: Won't this make a separate fetch for each `CodeExcerpt`? Are we just relying heavily on caching here?

camdencheek (approved):

client/web-sveltekit/src/routes/search/FileContentSearchResult.svelte
fkling: You are asking the same question that I had when I looked at the original code. Yes, we just rely on caching (I think the original code does it a bit differently but still).
If I remember correctly I wanted to do it differently but due to "lazy fetching" something along these lines made sense so I kept the overall approach, for the time being.

Note: This is just copied from the deleted file.

fkling: I should add a comment at some point, but I also expect this to change one way or the other.

PR #59012, opened 2023-12-14
Apply model list for users with overrides

rafax:
For users with quota overrides, we should still use their Cody-pro limits
## Test plan

- tested locally
- curl correctly refreshes limits


vdavid (approved):
Oh shite, what an oversight for this case. Sorry for this and thanks for the fix, Rafal!

PR #59034, opened 2023-12-15
graphqlbackend: use v2 telemetry for Sourcegraph Operator

bobheadxi:
Quick migration since I spotted a lot of these events on my user. I don't think the name of the event is used anywhere else, so these are probably only used for reference today

Also did a bit of cleanup on the slow request logging below to make sure it captures the trace of the request.

## Test plan

CI

### DIFF
--- a/cmd/frontend/graphqlbackend/graphqlbackend.go
+++ b/cmd/frontend/graphqlbackend/graphqlbackend.go
@@ -31,9 +31,11 @@ import (
 	"github.com/sourcegraph/sourcegraph/internal/errcode"
 	"github.com/sourcegraph/sourcegraph/internal/gitserver"
 	"github.com/sourcegraph/sourcegraph/internal/repoupdater"
+	"github.com/sourcegraph/sourcegraph/internal/telemetry"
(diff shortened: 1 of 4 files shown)


cmd/frontend/graphqlbackend/graphqlbackend.go
```diff
@@ -77,66 +79,42 @@ func (t *requestTracer) TraceQuery(ctx context.Context, queryString string, oper
...
-			t.logger.Error(
-				"failed to marshal JSON for event log argument",
-				log.String("eventName", eventName),
-				log.Error(err),
-			)
-		}
-
-		// NOTE: It is important to propagate the correct context that carries the
-		// information of the actor, especially whether the actor is a Sourcegraph
-		// operator or not.
```
unknwon, line 102: I think this NOTE is still valid?

unknwon (approved):

cmd/frontend/graphqlbackend/graphqlbackend.go
bobheadxi: I think it's redundant in the context of recording telemetry, since all callsites should require this

bobheadxi:
Closing to change this to audit logs instead

PR #59035, opened 2023-12-15
Code Navigation: bug fixes and optimization; file icons

jasonhawkharris:
Extract tests from utils.test.ts and move them to their own file that corresponds with the rest of the `fileIcons` functions. 

## Test plan
Ships with passing unit tests


client/web/src/repo/fileIcons.ts
```diff
@@ -132,12 +132,12 @@ interface IconInfo {
...
 }
 
-export interface FileInfo {
+interface FileInfo {
     extension: FileExtension
     isTest: boolean
 }
 
-export const getFileInfo = (file: string, isDirectory: boolean): FileInfo => {
+export function getFileInfo(file: string, isDirectory: boolean): FileInfo {
```
camdencheek, line 140: Is `export function` generally considered better style?

camdencheek (approved):

client/web/src/repo/fileIcons.ts
jasonhawkharris: Felix made a good comment [here](https://github.com/sourcegraph/sourcegraph/pull/58861#discussion_r1426492824) about using function syntax vs. arrow functions. I tend to agree with his ideas about it.

camdencheek: Interesting! I didn't know that's how that works. Thanks for sharing

PR #59039, opened 2023-12-15
svelte: Add repo/language/path filters to search results sidebar

fkling:
This commit approximates another part of Taiyab's design, namely the language, repo, path filters. As discussed in our huddle today we want to explore having filters be a separate query parameter instead of modifying the users input. This commit enables this feature. It takes the filters that are currently returned by the streaming API and generates links for them that update the search filter query parameter.

A couple of things to note:

- It's currently not possible to include multiple filters of the same type.
- Clicking the filter again will remove it.
- Because the filters are implemented as links we get data preloading on hover, which results in a very smooth experience for queries that are fast.
- For now the "type filter" is still implemented by modifying the query input.
- Section collapsing has been removed

Please keep in mind that this doesn't aim to be a faithful implementation of the design, nor is it the final behavior. It's a version that easy to achieve with the current resources available to get things going and iterate.

https://github.com/sourcegraph/sourcegraph/assets/179026/44d2bf54-ee4b-4006-8539-106a723a7048

## Test plan

Manual testing.


peterguy (approved):
nifty behavior! Something to consider is instead of removing the other filters from the list, leave them in, and toggle the selected one off when another one is clicked.

fkling:
@peterguy Yes, we have discussed various solutions to this. See [this thread](https://sourcegraph.slack.com/archives/C05EA9KQUTA/p1702652537109209). I believe someone is going to write this down in a more formal way. This is just the first iteration.

client/web-sveltekit/src/routes/search/filters.ts
camdencheek, line 1: Looks like this file is a duplicate of what's in `src/routes/search/utils.ts`?

client/web-sveltekit/src/routes/search/SearchResults.svelte
```diff
@@ -150,16 +139,34 @@
...
+                {/if}
             </div>
         {/if}
+        <a class="section help" href="/help/code_search/reference/queries" target="_blank">
+            <span class="icon">
+                <Icon svgPath={mdiBookOpenVariant} inline />
+            </span>
+            <div>
+                <h4>Need more advanced filters?</h4>
+                <span>Explore the query syntx docs</span>
```
camdencheek, line 167:
```suggestion
                <span>Explore the query syntax docs</span>
```

client/web-sveltekit/src/routes/search/SearchResults.svelte
```diff
@@ -48,11 +50,11 @@
...
-    $: filters = $stream?.filters
-    $: langFilters =
-        filters
-            ?.filter(filter => filter.kind === 'lang')
-            .map((filter): SidebarFilter => ({ ...filter, runImmediately: true })) ?? []
+    $: filters = groupFilters($stream?.filters)
+    $: langFilters = filters.lang
+    $: repoFilters = filters.repo
+    $: fileFilters = filters.file
+    $: hasFilters = langFilters.length > 0 || repoFilters.length > 0 || fileFilters.length > 0
```
camdencheek, line 57: Style question: why make so many different pieces of state rather than just using using `filters.lang`, etc. below?

camdencheek (approved):

client/web-sveltekit/src/routes/search/filters.ts
fkling: Oops, I forgot to remove that, thank you!

client/web-sveltekit/src/routes/search/SearchResults.svelte
fkling: I guess it doesn't make a difference here. Theoretically, if you use `filters.lang` in the template below, Svelte would rerender that part whenever `filters` changes, not just when `filters.lang` changes. However in this specific case, `filters.lang` will always be a new array instance when `filters` is computed.

Issue #59038, opened 2023-12-15
Searcher: optimize AND/ OR patterns


jtibshirani:
Currently, if an unindexed search contains a pattern like `test AND server`, then the search jobs framework executes it by
1. Creating two separate searcher jobs, one for the pattern `test`, and another for `server`
2. Streaming the results back and taking a conjunction

This can be quite inefficient and use a lot of memory within the frontend service.

We're planning an update to the query language to interpret spaces as an AND between terms by default, as opposed to treating them literally. So this will result in many more unindexed AND searches. We should update searcher to just handle AND/ OR patterns directly, making the execution much more efficient.

/cc @sourcegraph/search-platform

jtibshirani:
### Implementation plan

Initial refactors and bug fixes:
- https://github.com/sourcegraph/sourcegraph/pull/58950
- https://github.com/sourcegraph/sourcegraph/pull/59040
- https://github.com/sourcegraph/sourcegraph/pull/59306
- https://github.com/sourcegraph/sourcegraph/pull/59331
- https://github.com/sourcegraph/sourcegraph/pull/59288

Main implementation:
- [ ] Add support for 'and' and 'or' matchers in the searcher service
- [ ] Update jobs code to pass whole query tree to searcher

jtibshirani:
I ran the benchmarks for the main searcher logic before and after these changes. Documenting that I didn't see any regression:

<details>
  <summary>Before</summary>

```
BenchmarkSearchRegex_rare_fixed-10                   	      67	  15303353 ns/op	15542221 B/op	     617 allocs/op
BenchmarkSearchRegex_large_fixed_casesensitive
BenchmarkSearchRegex_large_fixed_casesensitive-10    	     228	   5127657 ns/op	   11425 B/op	     159 allocs/op
BenchmarkSearchRegex_large_re_dotstar
BenchmarkSearchRegex_large_re_dotstar-10             	       3	 507240514 ns/op	628485160 B/op	 6098749 allocs/op
BenchmarkSearchRegex_large_re_common
BenchmarkSearchRegex_large_re_common-10              	     120	   9848522 ns/op	 7489471 B/op	   68727 allocs/op
BenchmarkSearchRegex_large_re_anchor
BenchmarkSearchRegex_large_re_anchor-10              	      10	 122277004 ns/op	 4909972 B/op	   50084 allocs/op
BenchmarkSearchRegex_large_capture_group
[...]

PR #59040, opened 2023-12-15
Searcher: pull out 'matcher' interface

jtibshirani:
Currently searcher can only handle patterns with a single atom. We'd like to update searcher to handle AND/ OR patterns directly.

This refactor takes us a step towards being able to natively handle AND/ OR patterns by refactoring the way we match files. It pulls out a `matcher` interface, which for now just has a single implementation (`regexMatcher`), but eventually can represent nested matchers combined through AND/ OR. Even without this goal, I think the refactor stands on its own and helps clean up the logic.

Specific changes:
* Lots of renames to make things clearer, like `readerGrep` -> `regexMatcher`
* Define `matcher` interface, which finds matches given file contents, and related logic into new file
* Pull `pathMatcher` out of `matcher`, since it's really its own top-level concept
* Move all file content and buffer management to `search_regex`. This lets each `matcher` object become threadsafe, and avoids the need for copying them for every goroutine.

Relates to #59038

## Test plan

Refactor-only change, covered by existing unit tests (I checked and the coverage is quite good!) Also ran benchmarks to ensure we don't regress performance.


jtibshirani:
Note to reviewers: we have more pressing things to work on for the query language updates, but I was already in the middle of this refactor and wanted to checkpoint. So no rush on reviewing.

I ran the benchmarks to check there was no perf regression. I couldn't detect any, and in the numbers were a bit better (although I don't see why this would be, and suspect it's just random variance). Here's an example of a large-ish repo with a literal search.

**Existing code**
```
BenchmarkSearchRegex_large_fixed
BenchmarkSearchRegex_large_fixed-10    	     156	   7780120 ns/op	15592069 B/op	     202 allocs/op

BenchmarkSearchRegex_large_fixed
BenchmarkSearchRegex_large_fixed-10    	     157	   8066124 ns/op	15559601 B/op	     203 allocs/op

BenchmarkSearchRegex_large_fixed
BenchmarkSearchRegex_large_fixed-10    	     163	   8069660 ns/op	15577939 B/op	     203 allocs/op
```

**After refactor**
```
BenchmarkSearchRegex_large_fixed
BenchmarkSearchRegex_large_fixed-10    	     150	   7789635 ns/op	15466177 B/op	     200 allocs/op

BenchmarkSearchRegex_large_fixed
BenchmarkSearchRegex_large_fixed-10    	     158	   7423843 ns/op	15651497 B/op	     205 allocs/op

BenchmarkSearchRegex_large_fixed
BenchmarkSearchRegex_large_fixed-10    	     142	   7591836 ns/op	15636801 B/op	     202 allocs/op
```

cmd/searcher/internal/search/matcher.go
```diff
@@ -0,0 +1,215 @@
...
+	"github.com/sourcegraph/sourcegraph/cmd/searcher/protocol"
+	"github.com/sourcegraph/sourcegraph/internal/search/casetransform"
+	"github.com/sourcegraph/sourcegraph/internal/trace"
+
+	zoektquery "github.com/sourcegraph/zoekt/query"
+)
+
+type matcher interface {
+	// AddAttributes adds attributes to the trace
+	AddAttributes(tr trace.Trace)
```
jtibshirani, line 21: Some of these methods are a bit awkward, but I'd refine this when we add more implementations

cmd/searcher/internal/search/matcher.go
```diff
@@ -0,0 +1,215 @@
...
+
+	// MatchesAllContent returns whether the pattern will match all content
+	MatchesAllContent() bool
+
+	// IgnoreCase returns whether matches will ignore case
+	IgnoreCase() bool
+
+	// MatchesPath returns whether the pattern matches the given file path
+	MatchesPath(path string) bool
```
camdencheek, line 30: After pulling out `pathMatcher`, does this method still need to exist?

camdencheek (approved):
LGTM!

cmd/searcher/internal/search/matcher.go
jtibshirani: So this matches the *pattern* against the file path for searches that target file names. In contrast, `pathMatcher` represents the explicit file filters in the query. For maybe I can rename this `MatchesString` to avoid confusion.

In a follow-up, it'd be nice to refactor this to fold this path matching into `pathMatcher` when appropriate.

camdencheek: Ah, got it, thanks for explaining!

PR #59043, opened 2023-12-15
sg msp: apply workspace on creation, optionally on update

bobheadxi:
This change adds the following to `sg msp tfc sync`:

- `-apply-created`, default `true`: automatically apply newly created workspaces
- `-apply-updated`, default `false`: automatically apply updated workspaces

This makes for a nicer experience if following the VCS path: after merge, run `sg msp tfc sync`, and your workspaces automatically get runs started.

I'm not 100% sure this will work very well on the first run, where dependent workspaces haven't run yet, but this is handy for e.g. rolling out new stacks and creating runs locally.

## Test plan

```
$ sg msp tfc sync -all -apply-updated msp-testbed 
[msp-testbed] Synchronized Terraform Cloud workspaces for environment "test"

                                                                                                                      
  • msp-msp-testbed-test-project: https://app.terraform.io/app/sourcegraph/workspaces/msp-msp-testbed-test-project (updated,
  running)                                                                                                            
  • msp-msp-testbed-test-iam: https://app.terraform.io/app/sourcegraph/workspaces/msp-msp-testbed-test-iam (updated, running)
  • msp-msp-testbed-test-cloudrun: https://app.terraform.io/app/sourcegraph/workspaces/msp-msp-testbed-test-cloudrun (updated,
  running)                                                                                                            
  • msp-msp-testbed-test-monitoring: https://app.terraform.io/app/sourcegraph/workspaces/msp-msp-testbed-test-monitoring
  (updated, running)                                                                                                  

```


michaellzc:
on create is probably going to be awkward, both `project` and `run` workspaces will be applied at the same time, and `run` will 100% fail because the former is not ready yet

michaellzc (approved):

bobheadxi:
> consider using this in the future? https://registry.terraform.io/providers/hashicorp/tfe/latest/docs/resources/workspace_run if you want to full tfc-managed apply experience

wow TIL! let me look into this instead

bobheadxi:
With `wait_for_run` it looks like I should use that in conjunction with some run trigger, like this PR?

bobheadxi:
TIL `QueueAllRuns` exists:

> Whether to queue all runs. Unless this is set to true, runs triggered by a webhook will not be queued until at least one run is manually queued.

In between that and the TFE resource thing mentioned, I think there is a better solution to this. Closing this PR

Issue #58815, opened 2023-12-06
☂️ Search: simpler query language inspired by "keyword search"


jtibshirani:
We plan to bring code search closer to the "keyword search" style that users are used to, where queries are broken up into individual terms that can match broadly across the file name, contents, and symbols. This issue tracks our first round of work on improving the query language, which we aim to merge next release as 'beta'.

**Example query**: `repo:sourcegraph/sourcegraph auth provider`
* `auth provider` matches files that contain both the strings `auth` and `provider` in any order. Before, it only matched files containing the exact string `auth provider`
* There are no changes to filters in this release, so `repo:sourcegraph/sourcegraph` works as usual

### Core changes
- [x] #58799
- [x] Fold prototype into "search keyword" search mode @stefanhengl 
  - [x] Spaces interpreted as ANDs
    - [x] https://github.com/sourcegraph/sourcegraph/pull/59164
  - [x] #59057
  - [x] Support regex patterns using `/.../`
  - [x] https://github.com/sourcegraph/sourcegraph/pull/59686
- [x] Solidify the UX @jtibshirani 
  - https://github.com/sourcegraph/sourcegraph/pull/59707
  - https://github.com/sourcegraph/sourcegraph/pull/59779
  - https://github.com/sourcegraph/sourcegraph/pull/59840
  - https://github.com/sourcegraph/sourcegraph/pull/59915
- [x] https://github.com/sourcegraph/sourcegraph/pull/60047
- [x] Address issue with file match limit, where all top results may be from a single file @keegancsmith

### Usability / performance
- [x] Test that Zoekt can handle AND queries with many clauses @keegancsmith 
- [x] https://github.com/sourcegraph/sourcegraph/issues/59038 @jtibshirani 
- [x] Debug [why new search mode is much slower on S2](https://sourcegraph.slack.com/archives/C05R619V4F8/p1704484790035859) @keegancsmith 
- [x] Promote results with an exact match, so users can still easily copy/ paste in snippets @keegancsmith 
  - https://github.com/sourcegraph/zoekt/pull/728
  - https://github.com/sourcegraph/sourcegraph/pull/59940

### Testing + polish
- [x] https://github.com/sourcegraph/docs/pull/55
- [x] Telemetry @jtibshirani 
  - https://github.com/sourcegraph/sourcegraph/pull/59097
  - https://github.com/sourcegraph/sourcegraph/pull/59993
  - https://github.com/sourcegraph/sourcegraph/pull/60022
[...]

PR #59057, opened 2023-12-18
search: support "exact match" pattern using "..."

stefanhengl:
Relates to #58815

With this change, a quoted pattern, like `"foo bar"`, is interpreted literally IE spaces are interpreted as spaces instead of as logical `AND`. Quotes that should be matched literally have to be escaped.

Example:

To search for the text `foo "bar"`, where `bar` is surrounded by quotes, the query is either `"foo \"bar\""` or, if we use single quotes, `'foo "bar"'`.

Note: This change only affects our keyword search prototype.

Test plan:
- updated unit test


jtibshirani (commented):
This change turned out really cleanly!

I'm not totally convinced we should be supporting the single quote version. I haven't seen this in any other search query language ... I've only seen it in the context of string syntax in certain programming languages. I don't feel strongly though. What do you think?

Also adding @camdencheek as a reviewer so he can weigh in on these types of changes.

internal/search/query/labels.go
```diff
@@ -19,6 +19,7 @@ const (
 	// than canonical form (r: instead of repo:)
 	IsAlias
 	Standard
+	QuotesAsLiterals
```
camdencheek, line 22: Is this actually added to any nodes? In all the added tests, I don't see this show up anywhere. It seems like this is being used as an option rather than a label, which is a little confusing to me. That said...seems like we might be doing that already with `Standard` 😬

camdencheek (approved):

camdencheek:
No strong opinion personally on single/double quotes. It is pretty nice to be able to just switch to the other type of quote if the literal I'm looking for contains quotes.

internal/search/query/labels.go
stefanhengl: >  It seems like this is being used as an option rather than a label, which is a little confusing to me. That said...seems like we might be doing that already with Standard 

You are correct, it's used as an option, not as a label. I modelled it after Standard. I can see that this might be confusing. Given that Standard follows the same approach and given that this change here was straightforward by treating the label as option, I wonder whether it is worth it to introduce dedicated parser options to distinguish between the two concepts.

stefanhengl:
I think the single quotes are a nice touch, but not essential. I usually would try if single quotes work if escaping double quotes gets out of hand. If you don't mind, I suggest we leave it in for now. We still have some time to get a feeling for this.

PR #59080, opened 2023-12-18
search: support glob syntax for repo and file filters

stefanhengl:
Relates to https://github.com/sourcegraph/sourcegraph/issues/58815
Ported directly from https://github.com/sourcegraph/sourcegraph/pull/58849/

We add support for glob syntax to file and repo filters.

Notes:
- `f:` matches nothing. I think this is less surprising than our [current behavior](https://sourcegraph.com/search?q=context:global+f:&patternType=standard&sm=1)
- `*` matches any sequence of characters, including `/`
- No other special characters are supported

Test plan:
- new unit tests

### DIFF
--- a/internal/search/query/parser.go
+++ b/internal/search/query/parser.go
@@ -881,7 +881,7 @@ func (p *parser) ParseParameter(label labels) (Parameter, bool, error) {
 		return Parameter{}, false, err
 	}
 
-	if label.IsSet(GlobFilters) {
+	if label.IsSet(GlobFilters) && !parsedLabels.IsSet(IsPredicate) {
 		switch field {
 		case "r", "repo", "f", "file":
 			value = globToRegex(value)
--- a/internal/search/query/parser.go
+++ b/internal/search/query/parser.go
@@ -896,10 +896,11 @@ func (p *parser) ParseParameter(label labels) (Parameter, bool, error) {
 	}, true, nil
 }
 
-func globToRegex(gob string) string {
+var regexSpecialCharacter = regexp.MustCompile(`[\$\(\)\*\+\.\?\[\\\]\^\{\|\}]`)
+
+func globToRegex(glob string) string {
 	// First, we escape all the regex special characters.
-	re := regexp.MustCompile(`[\$\(\)\*\+\.\?\[\\\]\^\{\|\}]`)
-	r := re.ReplaceAllStringFunc(gob, func(match string) string {
+	r := regexSpecialCharacter.ReplaceAllStringFunc(glob, func(match string) string {
 		if match == "*" {
 			return ".*"
 		}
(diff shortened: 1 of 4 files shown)


stefanhengl:
@camdencheek I know, I am doubling down on the "options" pattern. If you think I should go down a different route, let me know. This should be the last of these kind of changes though ;-)

internal/search/query/parser.go
```diff
@@ -866,26 +868,65 @@ func (p *parser) ParsePattern(label labels) Pattern {
...
 	return Parameter{
 		Field:      field,
 		Value:      value,
 		Negated:    negated,
-		Annotation: Annotation{Range: newRange(start, p.pos), Labels: labels},
+		Annotation: Annotation{Range: newRange(start, p.pos), Labels: parsedLabels},
 	}, true, nil
 }
 
+func globToRegex(gob string) string {
```
camdencheek, line 899: standardize on "glob"?

```suggestion
func globToRegex(glob string) string {
```

internal/search/query/parser.go
```diff
@@ -866,26 +868,65 @@ func (p *parser) ParsePattern(label labels) Pattern {
...
+	// First, we escape all the regex special characters.
+	re := regexp.MustCompile(`[\$\(\)\*\+\.\?\[\\\]\^\{\|\}]`)
```
camdencheek, line 901: We should probably move this out of the function so we're not recompiling a regex for every glob

internal/search/query/parser.go
```diff
@@ -866,26 +868,65 @@ func (p *parser) ParsePattern(label labels) Pattern {
...
+	if r == ".*" {
+		return r
+	}
+
+	// Adjust the regex to account for implicit .* on either end.
+	if strings.HasPrefix(r, ".*") {
+		r = strings.TrimPrefix(r, ".*")
+	} else {
+		r = "^" + r
+	}
```
camdencheek, line 919: Am I understanding this right that, now, `file:test.json` will not match `subdirectory/test.json` and the filter would need to be `file:*/test.json`?

camdencheek (approved):
> I am doubling down on the "options" pattern

I'll try not to be a squeaky wheel 🙂 This is fine with me for now. Would you add a comment to the "option" labels for the sake future code spelunkers (and future me, when I inevitably get confused by this in the future)?

jtibshirani (approved):
Nice tests!

internal/search/query/parser.go
jtibshirani: Yes indeed! I added an example to the meta issue to help clarify the new semantics: https://github.com/sourcegraph/sourcegraph/issues/58815. We can add more examples as other questions come up.

We felt it was least surprising for filters to match the entire path (for both repo and file filters). This is a simpler mental model, removes the need for regex anchoring, and matches many other search inputs (like the file filter in JetBrains, repo filter in GitHub, etc.)

camdencheek: Cool! Particularly with globs, it'll be nice to not need anchoring

PR #59085, opened 2023-12-18
ci: enable chromatic again and update chromatic to latest

burmudar:
- Update chromatic from v7 to v10 - which seems to be the reason why it was previously failing
- re-enable chromatic

## Test plan
CI


client/web/src/auth.ts
```diff
@@ -7,6 +7,7 @@ import type { CurrentAuthStateResult } from '@sourcegraph/shared/src/graphql-ope
 
 import { requestGraphQL } from './backend/graphql'
 
+// REMOVE ME
```
burmudar, line 10:
```suggestion
```

PR #59091, opened 2023-12-18
svelte: Fix setting theme from temporary settings

fkling:
#59086 didn't work properly. Turns out the theme values stored in temporary setting are different from what the prototype expected. It only accidentally worked when I tested it.

This commit introduces a helper method that accepts an arbitrary string value and updates the theme state accordingly, instead of using an arbitrary string value as theme state (ideally TS would have complained about this before but it looks like it didn't).

## Test plan

Go to dotcom, change theme to the different values (system, light, dark). Refresh local dev build (proxies to dotcom) after every change and verify that it applies the set theme.


camdencheek:
> ideally TS would have complained about this before but it looks like it didn't

That sounds concerning. Is type checking not running in CI?

camdencheek (approved):

fkling:
@camdencheek The SvelteKit prototype is not typechecked but I didn't even get type warnings in my editor about it.

PR #59094, opened 2023-12-18
Make sure we test gRPC in CI suites

eseliger:
Some of the suites weren't yet using gRPC - this PR turns it on so we gain some more confidence before we start dropping HTTP entirely.

## Test plan

See if CI passes.


ggilmore (approved):
Thanks for catching this!

PR #59098, opened 2023-12-19
actions: create reusable workflow for sentry reports

burmudar:
This adds a reusable workflow which when called will send an event to sentry.

### There are some perculiarities
* A reusable workflow that _needs_ values and doesn't have defaults **must** define `needs: job_name` to depend on a job, otherwise it will be skipped
* Initially it seemed like all "needs" job would need to fail for a dependent job to run but I've tested and it is not the case

### Why didn't you use actions?
Well I did see commit https://github.com/sourcegraph/sourcegraph/pull/59098/commits/f55379865e6a1508de7abbaa04cc3b032194021f - ultimately it didn't bring much to the table. Initially I wanted to have it on every step, but you have to pass in all the variables, the action cannot pick up anything from *your* workflow run

## Test plan
reran buildchecker and universal-ctags workflows manually


.github/actions/report-failure/action.yml
```diff
@@ -0,0 +1,51 @@
+name: 'Report failure if action failed'
```
burmudar, line 1: To be removed

.github/workflows/buildchecker.yml
```diff
@@ -19,24 +19,22 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v2
         with: { go-version: '1.21' }
-      - uses: actions/setup-node@v4
-      - run: ./dev/buildchecker/run-check.sh
+      - run: exit 1 #./dev/buildchecker/run-check.sh
```
jhchabran, line 22: test thing I suppose :)

.github/workflows/buildchecker.yml
```diff
@@ -19,24 +19,22 @@ jobs:
...
-        env:
-          SENTRY_DSN: ${{ secrets.SENTRY_DEV_INFRA_DSN }}
-          SENTRY_ORG: sourcegraph
-          SENTRY_PROJECT: dev-infra
+
+  report_failure:
+    needs: run
+    if: ${{ failure() }}
+    uses: sourcegraph/sourcegraph/.github/workflows/report-job-failure.yml@main
+    with:
```
jhchabran, line 34: It seems that we can provide default inputs, see https://github.com/microsoft/action-python/blob/main/action.yml#L78-L80 but I don't know in which context they'll be evaluated. 

Worth checking I think, because right now the reusable variant takes as many lines as the original one. It's still better because you can't forget to update one, but still it's annoying, which I'm sure you agree with too.

jhchabran: If the context gets evaluated in the wrong place, could you we infer some of those values at runtime perhaps?

.github/workflows/buildchecker.yml
```diff
@@ -19,24 +19,22 @@ jobs:
...
-            -t "workflow-run:$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID"
-        env:
-          SENTRY_DSN: ${{ secrets.SENTRY_DEV_INFRA_DSN }}
-          SENTRY_ORG: sourcegraph
-          SENTRY_PROJECT: dev-infra
+
+  report_failure:
+    needs: run
+    if: ${{ failure() }}
+    uses: sourcegraph/sourcegraph/.github/workflows/report-job-failure.yml@main
```
jhchabran, line 33: It's pretty cool that this guarantees we're using the latest version of the step, this could come handy for others like pr-auditor for example.

.github/workflows/buildchecker.yml
burmudar: Good call. Just found the documenting piece which states that, the `github` context of the reusable workflow gets populated by the calling workflow ... so all the values can be inferred.

I opted to keep the inputs block as it makes it a bit more obvious this is the case imo. If you disagree I am happy to remove it :)

.github/workflows/buildchecker.yml
burmudar: Fixed 🤦🏼

jhchabran (approved):
Assuming the QA was correct, in which case it's pretty nice now 😊 

Pre-approving so you don't have to wait another review after you fix the branch names in the action ref.

jhchabran:
Uh, why do we have this 🤔

burmudar:
> Uh, why do we have this 🤔
> 
> CleanShot 2023-12-20 at 14 11 04@2x

It's because buildchecker deploys to the "autosheriff" environment. So each time I QA'd buildchecker, and it failed we get that error. Once I fix buildchecker to not just fail it should return to normal

PR #59101, opened 2023-12-19
Svelte: highlight expanded lines

camdencheek:
When clicking the "expand results" button, the expanded lines were not syntax highlighted for the svelte prototype. This happened the ranges we were fetching for highlights are pre-calculated from the collapsed groups. Instead, we want to fetch ranges for every group so when we expand, we have the highlighted result already. The first commit does that.

The second commit just moves the `matchRanges` calculation into the callback because that's the only place it's used and it doesn't need to be a separate piece of state.

Third commit is just the output of `pnpm format` in the `client/web-sveltekit` directory.

## Test plan

Manually tested that expanding the match groups shows highlighted code.

### DIFF
--- a/client/web-sveltekit/src/lib/react-interop.tsx
+++ b/client/web-sveltekit/src/lib/react-interop.tsx
@@ -190,7 +190,7 @@ function createURL(path: string | { pathname?: string; search?: string; hash?: s
     return url
 }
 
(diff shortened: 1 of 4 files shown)


fkling (approved):
You'll have to run `sg lint --fix format` in the workspace root because the package formatting options are different than the root ones. I'll have to fix that.

PR #59102, opened 2023-12-19
cody-gateway: remove SyncOnce, use existing SourceUpdater

bobheadxi:
Follow-up on https://github.com/sourcegraph/sourcegraph/pull/58732#discussion_r1414348506 as I was reading the code again this morning after looking at some discussions - this change removes the SyncOnce interface and uses our existing SourceUpdater.

There is some more nuance here than I remember, however: `Actor.Update` is also called when a user hits their rate limit. This could be a good thing (poll for update if a user is rate-limited) - if we increase a user's rate limit it will get updated quickly. Because of the above  if the actor was recently updated we no-op and return an error. This could be useful in the general refresh endpoint use case as well, to avoid something repeatedly hitting dotcom in quick succession.

## Test plan

Updated tests assert same behaviour as existing implementation

### DIFF
--- a/cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
+++ b/cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
@@ -10,6 +10,7 @@ import (
 	"github.com/Khan/genqlient/graphql"
 	graphqltypes "github.com/graph-gophers/graphql-go"
 	"github.com/graph-gophers/graphql-go/relay"
+
 	"github.com/sourcegraph/sourcegraph/internal/accesstoken"
 
 	"github.com/gregjones/httpcache"
@@ -42,7 +43,7 @@ type Source struct {
 	concurrencyConfig codygateway.ActorConcurrencyLimitConfig
 }
 
-var _ actor.SourceSingleSyncer = &Source{}
+var _ actor.SourceUpdater = &Source{}
 
 func NewSource(logger log.Logger, cache httpcache.Cache, dotComClient graphql.Client, concurrencyConfig codygateway.ActorConcurrencyLimitConfig) *Source {
 	return &Source{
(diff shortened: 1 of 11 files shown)


cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
```diff
@@ -59,8 +60,15 @@ func (s *Source) Get(ctx context.Context, token string) (*actor.Actor, error) {
 	return s.get(ctx, token, false)
 }
 
-func (s *Source) SyncOne(ctx context.Context, token string) error {
-	_, err := s.get(ctx, token, true)
+func (s *Source) Update(ctx context.Context, act *actor.Actor) error {
+	// Update is also called when a user's rate limit is hit, so we don't want
+	// to update every time. We use a shorter interval than the default in
+	// this case.
+	if act.LastUpdated != nil && time.Since(*act.LastUpdated) < 5*time.Minute {
```
rafax, line 67: nit: can you make this a package-level var (along with other timeouts)

cmd/cody-gateway/internal/actor/source.go
```diff
@@ -45,11 +44,19 @@ type Source interface {
 	Get(ctx context.Context, token string) (*Actor, error)
 }
 
+type ErrActorRecentlyUpdated struct{}
+
+func (e ErrActorRecentlyUpdated) Error() string {
+	return "actor already updated recently - try again later"
```
rafax, line 50: Say when to try?

rafax (approved):
Thank you for this!

cmd/cody-gateway/internal/httpapi/featurelimiter/featurelimiter.go
```diff
@@ -213,17 +213,18 @@ func ListLimitsHandler(baseLogger log.Logger, redisStore limiter.RedisStore) htt
 func RefreshLimitsHandler(baseLogger log.Logger, sources *actor.Sources) http.Handler {
```
rafax, line 213: You no longer need this param?

rafax:
```suggestion
func RefreshLimitsHandler(baseLogger log.Logger) http.Handler {
```

bobheadxi:
@rafax review comments addressed - thanks !

PR #59112, opened 2023-12-19
cody-gateway/anthropic: flag requests using feature instead of model

bobheadxi:
This change makes it so that all chat completion requests are eligible for flagging, now that there is an expanded set of models users can choose from for chat: https://sourcegraph.slack.com/archives/C06062P5TS5/p1703027281422839?thread_ts=1703025059.833749&cid=C06062P5TS5

## Test plan

Unit tests


rafax (approved):
Seeing that this is Anthropic-only, LGTM

rafax:
> Seeing that this is Anthropic-only, LGTM

But, this PR got pretty out-of-date, I'll apply the same change manually.

PR #59115, opened 2023-12-20
Remove App from codebase

eseliger:
Cody no longer needs it and it is obsolete now!

Since App added a non-insignificant amount of new concepts and alternative code paths, I decided to take some time and remove it from out codebase.
This PR removes ~21k lines of code. If we ever want parts of single binary (app), the redis kv alternatives, or the release pipeline for a native mac app back, we can look back at this PR and revert parts of it, but maintaining 21k lines of code and many code paths for which I had to delete a surprisingly small amount of tests justifies this move for me very well.

Technically, to some extent SG App and Cody App both still existed in the codebase, but we don't distribute either of them anymore, so IMO we shouldn't keep this weight in our code.

So.. here we go.

This should not affect any of the existing deployments, we only remove functionality that was special-cased for app.

## Test plan

CI. 

## Preview 🤩
[Preview Link](https://docs.sourcegraph.com/@es/remove-app)


client/web/src/LegacyLayout.tsx
```diff
@@ -79,14 +76,13 @@ export const LegacyLayout: FC<LegacyLayoutProps> = props => {
...
     const [wasSetupWizardSkipped] = useLocalStorage('setup.skipped', false)
-    const [wasAppSetupFinished] = useLocalStorage('app.setup.finished', false)
 
     const showDeveloperDialog =
         useDeveloperSettings(state => state.showDialog) &&
         (process.env.NODE_ENV === 'development' || isSourcegraphDev(props.authenticatedUser))
     const { fuzzyFinder } = useExperimentalFeatures(features => ({
         // enable fuzzy finder by default unless it's explicitly disabled in settings, or it's the Cody app
-        fuzzyFinder: features.fuzzyFinder ?? !props.isCodyApp,
+        fuzzyFinder: features.fuzzyFinder ?? false,
```
eseliger, line 85: todo

internal/session/session.go
```diff
@@ -113,20 +113,13 @@ func NewRedisStore(secureCookie func() bool) sessions.Store {
...
-	if pool, ok := redispool.Store.Pool(); ok {
-		rstore, err := redistore.NewRediStoreWithPool(pool, []byte(sessionCookieKey))
-		if err != nil {
-			waitForRedis(rstore)
-		}
-		store = rstore
-		options = rstore.Options
-	} else {
-		// Redis is not available, we fallback to storing state in cookies.
-		// TODO(keegan) ask why we can't just always use this.
```
eseliger, line 125: cc Keegan in case you still want to ask

fkling:
🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥🟥

internal/updatecheck/build.go
```diff
@@ -1,111 +1,17 @@
...
 		Version: *semver.New(version),
 	}
 }
-
-type Notification struct {
-	Key     string
-	Message string
-}
-
-func getNotifications(clientVersionString string) []Notification {
```
chwarwick, line 34: I think we need to retain this to display notifications to existing app users, regarding the shutting down of app.

schema/site.schema.json
```diff
@@ -2053,36 +2053,6 @@
             "type": "string"
           }
         },
-        "app.notifications": {
```
chwarwick, line 2056: I think we need to keep this for dotcom for notification purposes to existing app users. To stop using app.

internal/session/session.go
```diff
@@ -113,20 +113,13 @@ func NewRedisStore(secureCookie func() bool) sessions.Store {
...
-		cstore := sessions.NewCookieStore([]byte(sessionCookieKey))
-		store = cstore
-		options = cstore.Options
+	pool := redispool.Store.Pool()
+	rstore, err := redistore.NewRediStoreWithPool(pool, []byte(sessionCookieKey))
```
chwarwick, line 117: Do we not need the fall back to the cookie store anymore?

chwarwick (commented):
Thanks for doing this,  I don't have a good understanding on the cache changes I assume that App was the only thing using the DB for this.  I think we need to keep the dotcom parts that send make notifications available to app users.  Otherwise I think it's all good.

chwarwick:
@vovakulikov Since you will have a better idea about any ui things that might have overlap.

schema/site.schema.json
chwarwick: Well then I guess it can be removed.

unknwon (approved):
PR title LGTM!

unknwon (approved):
PR title LGTM!

daxmc99 (approved):
LGTM 🔥 
Maybe the largest removal of code yet?

chwarwick (approved):

internal/session/session.go
eseliger: no - this was app specific where we didn't have redis 👍

PR #59095, opened 2023-12-19
grpc: create stub retry utilities

ggilmore:
This PR adds a basic configuration for enabling retries with gRPC  for certain RPC types. 

The description for `defaults.RetryPolicy` is probably the most important thing to read:

```go
// RetryPolicy is the default retry policy for internal GRPC requests.
//
// The retry policy will trigger on Unavailable and ResourceExhausted status errors, and will retry up to 20 times using an
// exponential backoff policy with a maximum duration of 3s in between retries.
//
// Only Unary (1:1) and ServerStreaming (1:N) requests are retried. All other types of requests will immediately
// return an Unimplemented status error. It's up to the caller to manually retry these requests.
//
// These defaults can be overridden with the following environment variables:
// - SRC_GRPC_RETRY_DELAY_BASE: Base retry delay duration for internal GRPC requests
// - SRC_GRPC_RETRY_MAX_ATTEMPTS: Max retry attempts for internal GRPC requests
// - SRC_GRPC_RETRY_MAX_DURATION: Max retry duration for internal GRPC requests
var RetryPolicy = []grpc.CallOption{
	retry.WithCodes(codes.Unavailable, codes.ResourceExhausted),
	// Together with the default options, the maximum delay will behave like this:
	// Retry# Delay
	// 1	0.05s
	// 2	0.1s
	// 3	0.2s
	// 4	0.4s
	// 5	0.8s
	// 6	1.6s
	// 7	3.0s
	// 8	3.0s
	// ...
	// 20	3.0s
	retry.WithMax(uint(internalRetryMaxAttempts)),
	retry.WithBackoff(fullJitter(internalRetryDelayBase, internalRetryMaxDuration)),
}
```

This is off by default for all services (since this logic doesn't work with RPCS or might not be desirable as the default behavior if you don't know whether or not your method is idempotent).

 The upstack PRs selectively enable this logic for appropriate RPCs (see those PRs for the exact semantics). 

## Test plan

CI

PR #59130, opened 2023-12-20
grpc: repo-updater: add support for automatic retries for all methods (all are idempotent)

ggilmore:
This PR adds support for automatic retries in the repo-updater grpc client. 

---

I have gone through the repo-updater protobuf file and marked all the methods I thought were idempotent (we can't inspect this using the go protobuf packages, but I thought this was nice for documentation). 

I then wrapped the basic repo-updater grpc client with an "automaticRetryClient" that uses the default retry policy that was defined in https://github.com/sourcegraph/sourcegraph/pull/59095. See that PR for more details. 

## Test plan

CI

PR #59111, opened 2023-12-19
grpc: searcher: add support for automatic retries

ggilmore:
This PR adds support for automatic retries in the searcher grpc client. 

---

I have gone through the searcher protobuf file and marked all the methods I thought were idempotent (we can't inspect this using the go protobuf packages, but I thought this was nice for documentation). 

I then wrapped the basic searcher grpc client with an "automaticRetryClient" that uses the default retry policy that was defined in https://github.com/sourcegraph/sourcegraph/pull/59095. See that PR for more details. 

Note that for ServerStreaming methods like Search, the retry logic will only automatically retry if we haven't received any messages back from the server yet. 

After we receive a single message, we can't know whether or not the callers has consumed the message yet (e.x: started presenting the data in the WebUI from `Search`) and can tolerate receiving old messages, duplicated messages, etc.

 If we get an error after this point, we'll fail the RPC immediately and bubble up the underlying error to the caller. Only the caller would know the semantics of how it's consuming the stream to know how to proceed. 

## Test plan

CI

PR #59110, opened 2023-12-19
grpc: symbols: add support for automatic retries

ggilmore:
This PR adds support for automatic retries in the symbols grpc client. 

I have gone through the symbols protobuf file and marked all the methods I thought were idempotent (we can't inspect this using the go protobuf packages, but I thought this was nice for documentation). 

I then wrapped the basic symbols grpc client with an "automaticRetryClient" that uses the default retry policy that was defined in https://github.com/sourcegraph/sourcegraph/pull/59095. See that PR for more details. 

Note that for ServerStreaming methods like LocalCodeIntel and SymbolInfo, the retry logic will only automatically retry if we haven't received any messages back from the server yet. 

After we receive a single message, we can't know whether or not the callers has consumed the message yet (e.x: started aggregating the symbols from `LocalCodeIntel` ) and can tolerate receiving old messages, duplicated messages, etc. If we get an error after this point, we'll fail the RPC immediately and bubble up the underlying error to the caller. Only the caller would know the semantics of how it's consuming the stream to know how to proceed. 

## Test plan

CI

PR #59133, opened 2023-12-20
grpc: searcher: zoekt-webserver support automatic retries

ggilmore:
This PR adds support for automatic retries in the `zoekt-webserver` grpc client that `searcher` uses.

---

I wrapped the basic zoekt-webserver grpc client with an "automaticRetryClient" that uses the default retry policy that was defined in https://github.com/sourcegraph/sourcegraph/pull/59095. See that PR for more details. All the methods don't have any side effects, so they're all capable of being retried. 

Note that for ServerStreaming methods like StreamSearch and List, the retry logic will only automatically retry if we haven't received any messages back from the server yet. 

After we receive a single message, we can't know whether or not the caller has consumed the message yet (e.x: started consuming the search results from `StreamSearch` and displaying them in the WebUI) and can tolerate receiving old messages, duplicated messages, etc. If we get an error after this point, we'll fail the RPC immediately and bubble up the underlying error to the caller. Only the caller would know the semantics of how it's consuming the stream to know how to proceed. 

## Test plan

CI

### DIFF
--- a/internal/search/backend/BUILD.bazel
+++ b/internal/search/backend/BUILD.bazel
@@ -13,6 +13,7 @@ go_library(
         "index_options.go",
         "indexers.go",
         "metered_searcher.go",
+        "retry.go",
         "zoekt.go",
     ],
     importpath = "github.com/sourcegraph/sourcegraph/internal/search/backend",
--- a/internal/search/backend/retry.go
+++ b/internal/search/backend/retry.go
@@ -1,0 +1,37 @@
+package backend
+
+import (
+	"context"
+
+	"github.com/sourcegraph/sourcegraph/internal/grpc/defaults"
+	proto "github.com/sourcegraph/zoekt/grpc/protos/zoekt/webserver/v1"
+	"google.golang.org/grpc"
+)
+
+// automaticRetryClient is a convenience wrapper around a base proto.WebserverServiceClient that automatically retries
+// idempotent ("safe") methods in accordance to the policy defined in internal/grpc/defaults.RetryPolicy.
+//
+// Read the implementation of this type for more details are automatically retried (and why).
+//
+// Callers are free to override the default retry behavior by proving their own grpc.CallOptions when invoking the RPC.
+// (example: providing retry.WithMax(0) will disable retries even when invoking StreamSearch - which is idempotent).
+type automaticRetryClient struct {
+	base proto.WebserverServiceClient
+}
+
+func (a *automaticRetryClient) Search(ctx context.Context, in *proto.SearchRequest, opts ...grpc.CallOption) (*proto.SearchResponse, error) {
+	opts = append(defaults.RetryPolicy, opts...)
+	return a.base.Search(ctx, in, opts...)
+}
+
+func (a *automaticRetryClient) StreamSearch(ctx context.Context, in *proto.StreamSearchRequest, opts ...grpc.CallOption) (proto.WebserverService_StreamSearchClient, error) {
+	opts = append(defaults.RetryPolicy, opts...)
+	return a.base.StreamSearch(ctx, in, opts...)
+}
+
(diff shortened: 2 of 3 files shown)


ggilmore:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/59133?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `main`
  * **PR #59095** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/59095?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #59107** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/59107?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #59110** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/59110?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #59111** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/59111?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
          * **PR #59130** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/59130?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
            * **PR #59133** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/59133?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/59133?utm_source=stack-comment).

ggilmore:
@sourcegraph/search-platform @sourcegraph/code-search 

https://github.com/sourcegraph/sourcegraph/blob/6f3bf5179bb1e9c1c610c620362ee13ca7ef78ac/internal/search/backend/zoekt.go#L17-L31

I know we haven't used a retriable HTTP client for zoekt-webserver in the past due to this comment, but is this still an going concern with this approach? 

The gRPC retry logic holds a reference to the original message sent from the client in case it needs to send a retry. However:

1) This is a reference, not a copy
2) We already have other gRPC interceptors (like our [internalerror](https://github.com/sourcegraph/sourcegraph/tree/12-20-grpc_searcher_zoekt-webserver_support_automatic_retries/internal/grpc/internalerrs) logic) that hold on to a reference to the message for similar reasons
3) gRPC needs to hold on the entire message in memory anyway before it sends it due to the nature of the protocol

eseliger (approved):
should we also make a PR to zoekt to add the idempotency fields to the proto there?

ggilmore:
### Merge activity

* **Dec 22, 6:01 PM**: [Graphite](https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/59133) rebased this pull request as part of a merge.

keegancsmith:
@ggilmore

> https://github.com/sourcegraph/sourcegraph/blob/6f3bf5179bb1e9c1c610c620362ee13ca7ef78ac/internal/search/backend/zoekt.go#L17-L31
> 
> I know we haven't used a retriable HTTP client for zoekt-webserver in the past due to this comment, but is this still an going concern with this approach?

We handle network errors in our application code for zoekt. I think our concerns around large requests are not longer as much of a concern (that mostly doesn't happen anymore). But what I would worry about is requests being slow due to a down zoekt host. Right now our aggregator marks the search as incomplete and rather returns quickly. By having transparent retry logic we will end up with poor behaviour.

To be honest, this might not be so bad given we now use something called a flush timer in our aggregator...

At the end of the day, I am a bit concerned at introducing retries automatically. It feels like the sort of thing that should be decided per client. The default policy of retry sounds fine, but in the zoekt case I don't this we want it.

PR #59139, opened 2023-12-20
Guard repo code-intel page on Dotcom

BolajiOlajide:
[Slack Thread](https://sourcegraph.slack.com/archives/C07KZF47K/p1703102588008859)

## Test plan

* Start your sourcegraph instance with Dotcom configuration. `sg start dotcom`
* Navigate to the code intel dashboard of any repo: `https://sourcegraph.test:3443/github.com/sourcegraph/sourcegraph/-/code-graph/dashboard`
* You should get a 404 page.

### Screenshots

* Before adding the guard on the client side and added the backend guard.
CleanShot 2023-12-20 at 22 09 57

* With both guards (Client & Backend)
CleanShot 2023-12-20 at 22 10 36

### DIFF
--- a/client/web/src/enterprise/codeintel/repo/RepositoryCodeIntelArea.tsx
+++ b/client/web/src/enterprise/codeintel/repo/RepositoryCodeIntelArea.tsx
@@ -68,7 +68,7 @@ export const codeIntelAreaRoutes: readonly CodeIntelAreaRoute[] = [
     {
         path: '/dashboard',
         render: props => <RepoDashboardPage {...props} />,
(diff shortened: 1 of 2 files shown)


evict:
Nice! Do you also know where to make the resolver return the site-admin error message? Because technically I could still call the API with GraphQL.

BolajiOlajide:
> Nice! Do you also know where to make the resolver return the site-admin error message? Because technically I could still call the API with GraphQL.

Good catch. I'll add that in a few.

evict (approved):
Thanks @BolajiOlajide! That was 🚀

PR #46202, opened 2023-01-06
sg: Fix incorrect matching of comments in `sg lint -fix go`

sanderginn:
- Fixes unexpected behaviour described in https://github.com/sourcegraph/sourcegraph/pull/46151

For some reason, the `--not` flag used in `git grep` was not acknowledged. To achieve the intended behaviour, this PR introduces a negative lookahead instead and drops the `--not` flag.  
This also reverts the now obsolete `(CI:LOCAL_HOST_OK)` comments.

## Test plan
`sg lint -fix go` output without this change:
```console
❌ Check for localhost usage
   exit status 1
   
   --- no localhost guard
   
   Error: Found instances of "localhost":
     cmd/server/shared/postgres.go:                     // Set *PGHOST to default to 127.0.0.1, NOT localhost, as localhost does not correctly resolve in some environments
     internal/txemail/txemail.go:       // for it and always echoes "localhost" which makes it unusable.
   
   We generally prefer to use "127.0.0.1" instead of "localhost", because
   the Go DNS resolver fails to resolve "localhost" correctly in some
   situations (see https://github.com/sourcegraph/issues/issues/34 and
   https://github.com/sourcegraph/sourcegraph/issues/9129).
   
   If your usage of "localhost" is valid, then either
   1) add the comment "CI:LOCALHOST_OK" to the line where "localhost" occurs, or
   2) add an exclusion clause in the "git grep" command in  no-localhost-guard.sh
   
   ^^^ +++
```   

`sg lint -fix go` output with this change:
```console
✅ 1. go (24s)
✅ 2. format (47s)

👌 Everything looks good! Happy hacking!
```

PR #59144, opened 2023-12-21
bazel: add no-localhost-guard lint to nogo

Strum355:
Also fixes all the violations that snuck past due to https://github.com/sourcegraph/sourcegraph/pull/46202 confusingly breaking the `git grep` call :thinking: 

## Test plan

Tested with various combinations of [the following line](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@main/-/blob/internal/conf/computed.go?L300)


jamesmcnamara (approved):
LGTM

PR #59158, opened 2023-12-21
svelte: Add more repo navigation entries and make navigation somewhat response

### LINKED ISSUE #59149: Complete repository navigation

fkling:
Closes #59149

This PR adds more entries to the repo nav bar, all leading to the current web app. I basically copied the same list of links we are showing on the repo page in the current app plus some additional entries (Commits and Contributors; I assume those have been removed because the repo page shows/showed a panel for them? idk)

The navbar felt really crowded, so I put the additional items into a menu. This menu will also hold the navbar items if there is not enough space. We need to make a decision about which items we want to show.

I know the "selected entry" style in the menu is not ideal. While I think in general Svelte's way of handling style is better (it prevents issues with style ordering, which we have right now), it also makes it more difficult to build somewhat styleable component libraries. I'll have to spend more time investigating this. 

The includes some changes to the global navbar to make it work better on narrower screens (I originally wanted to use the same overflow approach, but the global navbar is setup differently, so I couldn't apply it as is).

https://github.com/sourcegraph/sourcegraph/assets/179026/e982a4c7-d5b5-456e-abbc-ffc51fea0446

## Test plan

Manual testing.

### DIFF
--- a/client/web-sveltekit/src/lib/dom.ts
+++ b/client/web-sveltekit/src/lib/dom.ts
@@ -8,7 +8,7 @@ import { highlightNode } from '$lib/common'
  * Returns a unique ID to be used with accessible elements.
  * Generates stable IDs in tests.
  */
-export function uniqueID(prefix = '') {
+export function uniqueID(prefix = ''): string {
     if (process.env.VITEST) {
         return `test-${prefix}-123`
     }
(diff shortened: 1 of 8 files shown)


client/web-sveltekit/src/lib/dom.ts
```diff
@@ -127,3 +127,38 @@ export const restrictToViewport: Action<HTMLElement, { offset?: number }> = (nod
         },
     }
 }
+
+/**
+ * An action to compute the number of elements that fit inside the container.
+ * Only works for static element lists (i.e. the first render should .
```
camdencheek, line 133: the first render should...?

client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte
```diff
@@ -1,98 +1,146 @@
...
-        <ul class="menu">
+<nav>
+    <h1><a href="/{repoName}"><Icon svgPath={mdiSourceRepository} inline /> {displayRepoName}</a></h1>
+    <!--
+        TODO: Add back revision
+        {#if revisionLabel}
+            @ <span class="button">{revisionLabel}</span>
+        {/if}
+        -->
+    <ul use:computeFit on:fit={event => (visibleNavEntries = event.detail.itemCount)}>
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte:72]: This is a neat example of composable behavior

camdencheek (approved):

client/web-sveltekit/src/lib/dom.ts
fkling: I got distracted ;)

client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte
Comment (fkling) [reply client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte:72]: Yeah, it's not a solution that can apply to every situation but it's an acceptable complexity tradeoff IMO.

PR #59164, opened 2023-12-21
search: respect patternType when navigating to search results

stefanhengl:
When a user clicks on a search result we navigate to the file and set the repo and file filters in the query input. However, this was hardcoded to use regex syntax. With this change we respect the patternType.

## Test plan
CI

### DIFF
--- a/client/web/src/search/index.ts
+++ b/client/web/src/search/index.ts
@@ -148,8 +148,22 @@ export function repoFilterForRepoRevision(repoName: string, revision?: string): 
     return `${escapeSpaces(`^${escapeRegExp(repoName)}$${revision ? `@${abbreviateOID(revision)}` : ''}`)}`
 }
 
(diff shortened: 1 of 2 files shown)


client/web/src/search/index.ts
```diff
@@ -148,10 +148,24 @@ export function repoFilterForRepoRevision(repoName: string, revision?: string):
     return `${escapeSpaces(`^${escapeRegExp(repoName)}$${revision ? `@${abbreviateOID(revision)}` : ''}`)}`
 }
 
-export function searchQueryForRepoRevision(repoName: string, revision?: string): string {
+export function searchQueryForRepoRevision(
```
fkling, line 151: For consistency you could rename this to `repoFilterForRepoREvision`.

fkling (approved):

PR #59173, opened 2023-12-21
Code Navigation: Fix C++ file icons and add .cc and .cxx to verified extensions

jasonhawkharris:
This commit adds two C++ file extensions to the file icon flow.

## Test plan
Ships with passing tests


camdencheek (approved):

client/web/src/repo/fileIcons.ts
```diff
@@ -61,7 +61,9 @@ export enum FileExtension {
     CLOJURE_CLJC = 'cljc',
     CLOJURE_EDN = 'edn',
     COFFEE = 'coffee',
-    CPP = 'cc',
+    CPP = 'cpp',
+    CPP_CC = 'cc',
+    CPP_CXX = 'cxx',
```
varungandhi-src, line 66: This list is still incomplete. C++ header files have extensions `.hpp`, `.h` (which is also used by C header files) and `.hxx`. I don't know if this is going to a case sensitive check or not, but sometimes uppercase extensions like `.CPP` extension is also used, especially for code developed on case-insensitive file systems.

PR #59176, opened 2023-12-21
Remove experimental indexRepositoryName for rust packages

eseliger:
Discussed this with Olaf and Thorsten before:
The way this repo is cloned on gitserver is using the API in an illegal way in that it assumes that repos without an entry in the repo table can just be cloned and will stay.

However, gitserver eventually deletes those repos again if there's no DB match for them.

This is blocking making improvements to tracking repositories on disk because it would mean no entry in the DB is found, and never worked reliably for on-prem instances due to the purge worker eliminating untracked repos like this one.

## Test plan

Very self-contained change. Relying on CI.


olafurpg (approved):
LGTM 👍🏻 Thank you for following up on this!

PR #59178, opened 2023-12-21
Search Results: Add new filters panel

vovakulikov:
## What have been done in this PR
This PR contains changes about filter panel UI. Most of these changes were implemented by design spec [here](https://www.figma.com/file/EPgl0qVbqGSumvf52brhda/Improving-search-results?type=design&node-id=431-55452&mode=design&t=JVOoFT0yVIjLTXz3-0). In a nutshell 
- New filter panel UI 
- Context-free filters (like first section with type filters)
- Context-based filters (like repositories, commits authors and symbol kinds)
- Improvements about caching system so changing filters won't trigger query twice

https://github.com/sourcegraph/sourcegraph/assets/18492575/42f4c101-702c-4340-ad21-ab24c3c878e3

## Follow-ups
- Static filters, at the moment filters like "Symbols kind", "Commit date" and "Authors" are generated on the client and therefore they don't have count info and it also possible that some of them may produce zero state. This is because currently we don't have them in our Search API. 
- Polishing designs details such icons and badges (in this PR I was focusing more on how filters work rather how they look, polishing will be done in a follow up PR to simplify tech and design review)
- Mobile layout for the search results page with new filters UI
- Removing feature flags (some flags about search results page will go away very soon)

## Feature flags setup

This feature eventually will be enabled by default but at the moment it's behind a feature flag `newSearchResultFiltersPanel`, also make sure that your other feature flag `newSearchResultsUI` is enabled too since new filters UI is added only to the most recent version of the search results page (new search results page feature flag will be removed in a follow up PR) 

## Test plan
- Check that you can see and interact with new filters panel if you have enabled `newSearchResultFiltersPanel: true` feature flag
- Check that the old panel got no regressions


client/branded/src/search-ui/results/filters/components/filters-doc-footer/FiltersDocFooter.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+import { Icon, Link, H4 } from '@sourcegraph/wildcard'
+
+import styles from './FiltersDocFooter.module.scss'
+
+interface FiltersDocFooterProps {
+    className?: string
+}
+
+export const FiltersDocFooter: FC<FiltersDocFooterProps> = ({ className }) => (
```
jasonhawkharris, line 14: It would be nice to change the positioning of this so that it is fixed to the bottom of the filter panel, regardless of scroll position. If the search produces so many results that it hides this, it might be frustrating for a user desiring to be more granular in their code search usage.

client/branded/src/search-ui/results/filters/NewSearchFilters.tsx
```diff
@@ -0,0 +1,162 @@
...
+import { useFilterQuery } from './hooks'
+import { COMMIT_DATE_FILTERS, SearchFilterType, SYMBOL_KIND_FILTERS } from './types'
+import { generateAuthorFilters } from './utils'
+
+import styles from './NewSearchFilters.module.scss'
+
+interface NewSearchFiltersProps {
+    query: string
+    results: SearchMatch[] | undefined
+    filters?: ResultFilter[]
```
jasonhawkharris, line 34: Is there a time when this could be undefined/null?

client/branded/src/search-ui/results/filters/components/dynamic-filter/SearchDynamicFilter.tsx
```diff
@@ -0,0 +1,279 @@
...
+                <Input
+                    variant="small"
+                    value={searchTerm}
+                    placeholder={`Filter ${filterTypes.join(',')}`}
+                    onChange={event => setSearchTerm(event.target.value)}
+                />
+            )}
+
+            <ul className={styles.list}>
+                {filtersToShow.map(filter => {
```
jasonhawkharris, line 175: readability nit: could this be extracted to its own function?

client/branded/src/search-ui/results/filters/components/filter-type-list/FilterTypeList.tsx
```diff
@@ -0,0 +1,157 @@
...
+            return 'symbol'
+        }
+        case SearchFilterType.Commits: {
+            return 'commit'
+        }
+        case SearchFilterType.Diffs: {
+            return 'diff'
+        }
+
+        default: {
```
jasonhawkharris, line 153: Do we need the default case here? Would we ever pass anything other than repo, path, symbol, commit or diff into `toSearchSyntaxTypeFilter`?

client/browser/src/shared/code-hosts/gitlab/scrape.ts
```diff
@@ -54,6 +54,7 @@ export const getPageKindFromPathName = (owner: string, projectName: string, path
  */
 export const getGitlabRepoURL = (): string => {
     const projectLink = document.querySelector<HTMLAnchorElement>('.context-header a, .shortcuts-project')
+    console.log('link', projectLink)
```
jasonhawkharris, line 57: I'm assuming this was for debugging and can be deleted?

client/web/src/repo/tree/TreePagePanels.tsx
```diff
@@ -25,7 +27,7 @@ import type { BlobFileFields, TreeHistoryFields } from '../../graphql-operations
 import { fetchBlob } from '../blob/backend'
 import { RenderedFile } from '../blob/RenderedFile'
 import { CommitMessageWithLinks } from '../commit/CommitMessageWithLinks'
-import { FILE_ICONS, FileExtension, getFileInfo } from '../fileIcons'
+import { getFileInfo } from '../fileIcons'
```
jasonhawkharris, line 30: The icon logic may need to be reworked when [this PR](https://github.com/sourcegraph/sourcegraph/pull/59234) is merged. You may want to wait to merge this until that one is merged, as it will probably change the invocations. For example, `getFileInfo` as a function won't even exist any more once it is merged.

client/wildcard/src/components/Icon/language-icon/LanguageIcon.module.scss
```diff
@@ -0,0 +1,52 @@
+.icon {
```
jasonhawkharris, line 1: See comment above regarding the file-icons PR. This might create some nasty merge conflicts with [this](https://github.com/sourcegraph/sourcegraph/pull/59234).

schema/settings.schema.json
```diff
@@ -13,6 +13,14 @@
       "type": "object",
       "additionalProperties": true,
       "properties": {
+        "newSearchResultFilters": {
```
jasonhawkharris, line 16: May be good to note that the filters themselves haven't changed in terms of what's available, just the way in which we filter. 

```suggestion
        "newSearchResultFiltersPanel": {
          "description": "Enables new experimental search results filters panel"
```

jasonhawkharris (commented):
LGTM overall. A few nits, a few suggestions. 

I'm going to stop short of approving because in terms of language Icons, a lot of the logic may need to change to accommodate [this PR](https://github.com/sourcegraph/sourcegraph/pull/59234) when it is merged, which will be soon. 

I think this is a vast improvement on the product as a whole. Killer job getting it out so fast!

client/branded/src/search-ui/results/filters/types.ts
```diff
@@ -0,0 +1,74 @@
...
+        SearchResultFilters.ByCommitDate,
+        SearchResultFilters.ArchivedAndForked,
+    ],
+    [SearchFilterType.Diffs]: [
+        SearchResultFilters.ByDiffType,
+        SearchResultFilters.ByAuthor,
+        SearchResultFilters.ByRepository,
+        SearchResultFilters.ArchivedAndForked,
+    ],
+}
```
fkling, line 60: I can't find where these are used... are they still needed?

client/web/src/repo/tree/TreePagePanels.tsx
jasonhawkharris: Specifically, some language detection improvements changed quite a few aspects of how the file icons are populated. 

cc: @varungandhi-src

client/browser/src/shared/code-hosts/gitlab/scrape.ts
fkling, line 57: Delete

```suggestion
```

client/shared/src/search/query/transformer.ts
```diff
@@ -21,7 +21,7 @@ export function omitFilter(query: string, filter: Filter): string {
     return `${query.slice(0, start).trimEnd()} ${query.slice(end).trimStart()}`.trim()
 }
 
-const succeedScan = (query: string): Token[] => {
+export const succeedScan = (query: string): Token[] => {
```
fkling, line 24: If we want to expose this function, it would make more sense to me to move it to `scanner.ts`.

client/shared/src/search/stream.ts
```diff
@@ -270,7 +270,7 @@ export interface Filter {
     label: string
     count: number
     limitHit: boolean
-    kind: 'file' | 'repo' | 'lang' | 'utility'
+    kind: 'file' | 'repo' | 'lang' | 'utility' | 'select' | 'after' | 'before' | 'author'
```
fkling, line 273: I find it questionable to extend the server side filter kind with client side values (that's what's happening here, isn't it?). If there is no other way I'd at least move these into a separate line and a comment about it.

client/web/src/search/results/SearchResultsCacheProvider.tsx
```diff
@@ -68,14 +91,15 @@ export function useCachedSearchResults(
             // See: https://github.com/ReactiveX/rxjs/issues/5732
             return merge(stream.pipe(throttleTime(500)), stream.pipe(last())).pipe(
                 tap(results => {
-                    cachedResults.current = { results, query, options }
+                    const previosCache = cachedResults.current?.cache ?? {}
+                    cachedResults.current = { query, options, cache: { ...previosCache, [filterQuery]: results } }
```
fkling, line 95:
```suggestion
                    const previousCache = cachedResults.current?.cache ?? {}
                    cachedResults.current = { query, options, cache: { ...previousCache, [filterQuery]: results } }
```

fkling (commented):
I'll do another pass tomorrow.
Note that the file icon stuff is being changed in https://github.com/sourcegraph/sourcegraph/pull/59234 .

client/branded/src/search-ui/results/filters/NewSearchFilters.tsx
vovakulikov: Yes, when we just started observe results from the backend (right after user submitted their search) default value for filters is undefined in the stream search API.

client/branded/src/search-ui/results/filters/components/filter-type-list/FilterTypeList.tsx
vovakulikov: Yeah because of Code type which is in my opinion isn't `type:file` but `type:file OR type:commit OR type:diffs`, since this filter type doesn't have one single types we don't assign any type filter to this section but rather just clear search query from all type filters.

client/web/src/repo/tree/TreePagePanels.tsx
vovakulikov: I left a comment about this in the Varun's PR, will merge his and then I adjust this one. Thanks for the heads up!

client/branded/src/search-ui/results/filters/NewSearchFilters.module.scss
```diff
@@ -0,0 +1,45 @@
...
+
+        &::-webkit-scrollbar-track {
+            background-color: var(--oc-gray-7);
+        }
+
+        @-moz-document url-prefix('') {
+            scrollbar-width: thin;
+            scrollbar-color: var(--oc-gray-6);
+        }
+    }
```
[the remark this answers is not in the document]
vovakulikov: @fkling this is just hacks due to needs to set different group of rules in different browsers, to my understanding flat list of rules might cause problems because maybe at the moment these rules don't overlap but someday they will. 

I was thinking about improvements here anyway since I think we don't have a good solution at the moment in our codebase for unified scrollbars, everyone does this differently, I took this snippet from some place in Code Insights because I used it before. Ideally, we should have either separate component or CSS mixin for this kind of things. I will try to do it shortly after this PR.

client/shared/src/search/stream.ts
vovakulikov: It makes sense yeah, I created a separate interface for new filter panel components which extends the backend one. When backend supports new type of filters we could remove it from one place.

client/branded/src/search-ui/results/filters/components/filters-doc-footer/FiltersDocFooter.tsx
vovakulikov: @jasonhawkharris my logic here is if we have a lot of filters content we probably should want to render as much them as possible to give more context and filtering options to the user. 

The filter footer section in my opinion is kind of helper here when you picked some filters but this produced a zero result state (when this footer panel will be visible and placed at the bottom of the filters panel) 

cc @taiyab

vovakulikov:
@jasonhawkharris @fkling I resolved most of the comments and adjusted by the new lang icons approach, take a look whenever you have a minute, thanks

fkling (approved):
Did you merge in the other PR? Or why are there so many file icon related changes now? Regardless, you should rebase on top of https://github.com/sourcegraph/sourcegraph/pull/59268 to "fix" the Timestamp test.

Accepting to unblock.

client/branded/src/search-ui/results/filters/NewSearchFilters.tsx
```diff
@@ -0,0 +1,162 @@
...
+                    filters={SYMBOL_KIND_FILTERS}
+                    exclusive={true}
+                    staticFilters={true}
+                    filterQuery={filterQuery}
+                    renderItem={symbolFilter}
+                    onFilterQueryChange={setFilterQuery}
+                />
+            )}
+
+            {(type === SearchFilterType.Commits || type === SearchFilterType.Diffs) && (
```
jasonhawkharris, line 97: So, if I'm reading this correctly, if this expression evaluates to true, this means that 'By author' and 'By commit date' will render above the other dynamic filters like 'By language', 'by repositories' and 'by file'. 

IMO, 'By author' and 'By commit date', being less common use cases, should be positioned below these other more common filter types.

client/branded/src/search-ui/results/filters/components/filters-doc-footer/FiltersDocFooter.tsx
jasonhawkharris: @taiyab I'm curious about your thoughts here.

jasonhawkharris (approved):
Nice! Killer job. Very excited about this.

client/branded/src/search-ui/results/filters/NewSearchFilters.tsx
vovakulikov: By author and by commit date are more specific when you select type:commit or type:diff filters IMO. But to be honest I don't have any preference here just got it order from the Figma design.

vovakulikov:
> Did you merge in the other PR? Or why are there so many file icon related changes now? 

I rebased the main and fixed the language icon component, the problem with new approach was that it newly added langs icon logic wasn't straightforward if you want to reuse the language icons. So my idea was to create a common LanguageIcon component where you can pass just lang name (extension optionally) and render lang icon. But to do this I had to refactor lang icon logic and also resoled common → wildcard → web dependencies. So my point is I had to do some changes about lang icons to reuse them in filters UI. 

I rebased onto the main with timestamp fix so I think this should be ok now.

client/branded/src/search-ui/results/filters/components/filters-doc-footer/FiltersDocFooter.tsx
taiyab: @jasonhawkharris Basically what @vovakulikov said. The footer section is almost a "tried everything else and not got what you want, OK go here" type prompt. It doesn't need to be so prominently in someone's face, and it's likely someone will scroll to the bottom in their attempt to find whatever other options they're looking for.

client/branded/src/search-ui/results/filters/NewSearchFilters.tsx
taiyab: @jasonhawkharris You defo have a point, but I also didn't want the list to look the same across different types.

It should be immediately obvious you can filter different things across different types, along with the fact that `By author` and `By commit date` are more specific to that type.

taiyab:
Great job on this @vovakulikov!

PR #59179, opened 2023-12-21
Code Navigation: add 'ago' back to timestamps

jasonhawkharris:
Adds 'ago' back to timestamps in the 'last updated' column of the TreePanel. 

## Test plan
Manual testing/CI passing

### DIFF
--- a/client/web/src/repo/tree/TreePagePanels.tsx
+++ b/client/web/src/repo/tree/TreePagePanels.tsx
@@ -227,11 +227,7 @@ export const FilesCard: FC<FilePanelProps> = ({ entries, historyEntries, classNa
                                         </span>
                                     </td>
                                     <td className={classNames(styles.commitDate, 'text-muted')}>
-                                        <Timestamp
-                                            noAbout={true}
-                                            noAgo={true}
-                                            date={getCommitDate(fileHistoryByPath[entry.path])}
-                                        />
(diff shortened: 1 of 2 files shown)


client/branded/src/components/Timestamp/Timestamp.tsx
```diff
@@ -80,7 +80,7 @@ export const Timestamp: React.FunctionComponent<React.PropsWithChildren<Timestam
 
     return (
         <Tooltip content={preferAbsolute ? label : tooltip}>
-            <span className="timestamp">{preferAbsolute ? tooltip : label}</span>
+            <span className="timestamp">{preferAbsolute ? tooltip : label} {noAgo ? "ago" : ""}</span>
```
fkling, line 83: The "ago" part is already added by the library we are using (see `calculateLabel`). Instead you want to remove this prop: https://github.com/sourcegraph/sourcegraph/blob/9ce3f1858ea2b325501eeb7031bb6b8b6756286e/client/web/src/repo/tree/TreePagePanels.tsx#L232

fkling (changes requested):

fkling (approved):

eseliger (approved):

camdencheek (approved):
Maybe change the copy on "Last commit date" to "Last modified" since we aren't actually listing dates (and it's shorter)?

PR #59185, opened 2023-12-21
svelte: Add search query input to repo pages

### LINKED ISSUE #59162: Make search input available on repo pages somehow

fkling:
Closes #59162

This PR adds the search query input to the repo page header. I used a similar solution as GitHub, which renders a button that looks a bit like an input and opens a dialog with the real input.
The input is prepopulated with a `repo:` filter for the current input. It has a tab/focus trap and closes on escape.

Before settling on this solution I tried various other approaches:

- Directly render the input in the header and make it expand on focus. That caused strange text selection because mousdown would expand the input and mouseup would happen at a different place in the input, causing text selection.
- Using the `<dialog>` element. It has some default styles/behavior that interfered with what I wanted to do.
- I wanted to use the newly added `openFocus` option in `createDialog` to disable melt's autofocus behavior, but that didn't work.

Happy to iterate on the design of the button, or on the behavior in general.

I decided against introducing keyboard shortcuts in this PR, I think that warrants it's own PR.

https://github.com/sourcegraph/sourcegraph/assets/179026/79f72d15-30ee-41f0-be13-6f96f1a1bdab

## Test plan

- Click search button opens dialog with search input, focused, with current repo filter and cursor at the end
- Tab/shift-tab keeps focus inside the dialog
- Escape closes the dialog, preserving the entered search query
- Enter submits the query and loads the search results page

### DIFF
--- a/client/web-sveltekit/package.json
+++ b/client/web-sveltekit/package.json
@@ -63,7 +63,7 @@
   },
   "type": "module",
   "dependencies": {
-    "@melt-ui/svelte": "^0.34.3",
+    "@melt-ui/svelte": "^0.66.2",
     "@popperjs/core": "^2.11.8",
     "@remix-run/router": "~1.3.3",
     "@sourcegraph/branded": "workspace:*",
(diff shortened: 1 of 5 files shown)


camdencheek (approved):
Nice!

PR #59114, opened 2023-12-20
SG Linter to prevent editing the `doc/` directory

jamesmcnamara:
Adds a linter to `sg lint` which will fail if any manual changes commited to the `doc/` folder. Should allow auto generated docs but not manual edits. Those should be migrated to https://github.com/sourcegraph/docs

## Test plan
Adding as a CI check

## Preview 🤩
[Preview Link](https://docs.sourcegraph.com/@jsm/block-docs)

PR #59189, opened 2023-12-21
ci: permit changes to doc/cloud/*

michaellzc:
In https://github.com/sourcegraph/sourcegraph/pull/59114, we started blocking all changes to docs v1.

After discussing with @MaedahBatool @eshamow, we decided to permit changes to Cloud docs until docs v1 can be sunset to avoid showing outdated content to Cloud customers.

## Test plan

CI should pass

### DIFF
--- a/doc/cloud/index.md
+++ b/doc/cloud/index.md
@@ -110,7 +110,7 @@ If you have specific requirements for the region, please reach out to your Sourc
 
 More details about the locations and data storage can be found in [our handbook](https://handbook.sourcegraph.com/departments/cloud/technical-docs/multi-region/)
 
-### Private Connectivit (test remote later)
+### Private Connectivity
 
 Sourcegraph Cloud can connect to resources that are publically accessible or protected by IP-based firewall rules out-of-the-box. Sourcegraph can provide static IP addresses for customers to add to their firewall allowlist. Please let your account team know.
 
--- a/doc/cloud/index.md
+++ b/doc/cloud/index.md
@@ -110,7 +110,7 @@ If you have specific requirements for the region, please reach out to your Sourc
 
 More details about the locations and data storage can be found in [our handbook](https://handbook.sourcegraph.com/departments/cloud/technical-docs/multi-region/)
 
-### Private Connectivity
+### Private Connectivit (test remote later)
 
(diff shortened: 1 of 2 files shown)


michaellzc:
* `main`
  * **PR #59189** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/59189?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/59189?utm_source=stack-comment).

doc/cloud/index.md
```diff
@@ -110,7 +110,7 @@ If you have specific requirements for the region, please reach out to your Sourc
 
 More details about the locations and data storage can be found in [our handbook](https://handbook.sourcegraph.com/departments/cloud/technical-docs/multi-region/)
 
-### Private Connectivity
+### Private Connectivit (test remote later)
```
eseliger, line 113: is this change intentional?

eseliger (approved):

doc/cloud/index.md
michaellzc: yeah going to revert later after pasing the build as test plan

jamesmcnamara (approved):

jamesmcnamara (changes requested):
Sorry my github UI is currently not loading very effectively so I can't leave this inline.

Currently the doubly-nested continue is causing the `continue` to have no effect. I think you'll need to wrap it in a function i.e.

```go
func isDiffAllowed(filename string) bool {
  for _, allowed := range allowDocChangesPrefixAllowlist {
    if strings.HasPrefix(allowed, filename) {
      return true
    }
  }
  return false
}
```

And then later:
```go
for filename := range diff {
  if !isDiffAllowed(filename) {
    diffset[filename] = struct{}{}
  }
}
```

jamesmcnamara (approved):

doc/cloud/index.md
michaellzc, line 113:
```suggestion
### Private Connectivity
```

PR #59190, opened 2023-12-21
msp: add 'sg msp logs'

bobheadxi:
Jump to a tidy view of service logs (assuming `sourcegraph/log`) in browser quickly.

## Test plan

```
sg msp logs sams dev
```

image


unknwon (approved):
nice

PR #59197, opened 2023-12-22
Fix logic for when to show add access token button

eseliger:
The previous logic didn't account for the allow-site-admin state so we never displayed the button in that case. This PR fixes that.

I want to reconsider these three options in the future and see if we couldn't replace it with RBAC instead, but for now this seems like the most straightforward way to improvement.

## Test plan

Manually tested all the states.

### DIFF
--- a/client/web/src/site-admin/SiteAdminTokensPage.tsx
+++ b/client/web/src/site-admin/SiteAdminTokensPage.tsx
@@ -4,7 +4,7 @@ import { mdiPlus } from '@mdi/js'
 import { Subject } from 'rxjs'
 
 import type { TelemetryProps } from '@sourcegraph/shared/src/telemetry/telemetryService'
-import { Button, ButtonLink, Container, Icon, PageHeader } from '@sourcegraph/wildcard'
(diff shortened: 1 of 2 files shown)


client/web/src/site-admin/SiteAdminTokensPage.tsx
```diff
@@ -51,18 +51,15 @@ export const SiteAdminTokensPage: React.FunctionComponent<React.PropsWithChildre
                                 className="ml-2"
                                 to={`${authenticatedUser.settingsURL!}/tokens/new`}
                             >
-                                <Icon aria-hidden={true} svgPath={mdiPlus} /> Generate access token
```
ggilmore, line 54: intentional?

ggilmore (approved):

client/web/src/site-admin/SiteAdminTokensPage.tsx
eseliger: The wording change? somewhat, yeah I wanted to slightly shorten it to look better, but admittedly it doesn't have anything to do with the purpose of the PR 😬

PR #59209, opened 2023-12-22
Consolidate GitBlame client method and improve performance

eseliger:
This PR makes three main changes:

- Removes the non-streaming implementation of BlameFile that was mostly a duplicative implementation that we need to maintain
- Eliminates the need to buffer all of stdout in memory for parsing
- Reduces overhead of the SearchResultsResolver.blameFileMatch method, we now only parse and store in memory max 1 entry
- Adds caching to authors in the endpoint so we don't need to look up a user multiple times if they own multiple hunks

## Test plan

Streaming blame has been enabled for a long time on multiple instances and hasn't revealed problems. Also, CI should catch any further issues.

## Preview 🤩
[Preview Link](https://docs.sourcegraph.com/@es/blame-consolidation)


fkling:
Are there any noticeable performance improvements?

eseliger:
> Are there any noticeable performance improvements?

Likely, although I didn't take benchmarks. For certain that'll require less memory, but I also believe that by not parsing more hunks than we need in the search code path, that should be faster

camdencheek (approved):
Nice! Thank you!

jhchabran (approved):
Thanks! Happy to see that this thing has been working well for all this time 😊 

Good catch caching the users 🙏

jhchabran (changes requested):
Spotted a few missing things: 

- doc/dev/background-information/gitserver-api.md still mentions `BlameFile` 

TODO currently editing

jhchabran (approved):

PR #59214, opened 2023-12-22
Add slack webhook notifications on license creation

dadlerj:
Dotcom only, not urgent for release.

Adds slack notification to license creation process to provide oversight to internal TS and deal desk teams

## Test plan

Tested locally, see test posts in https://sourcegraph.slack.com/archives/C06BC899NJX/p1703262143062429


schema/schema.go
```diff
@@ -676,6 +676,8 @@ type Dotcom struct {
 	SlackLicenseAnomallyWebhook string `json:"slackLicenseAnomallyWebhook,omitempty"`
 	// SlackLicenseExpirationWebhook description: Slack webhook for upcoming license expiration notifications.
 	SlackLicenseExpirationWebhook string `json:"slackLicenseExpirationWebhook,omitempty"`
+	// SlackLicenseCreationWebhook description: Slack webhook for when a license key is created.
+	SlackLicenseCreationWebhook string `json:"slackLicenseCreationWebhook,omitempty"`
```
camdencheek, line 680: I do not see an accompanying change in `schema/site.schema.json`. The `schema/schema.go` file should not be edited directly since it's a generated file, so the edits should be made to `site.schema.json` then `sg generate` can be run to update this file.

cmd/frontend/internal/dotcom/productsubscription/licenses_db.go
```diff
@@ -94,6 +113,44 @@ func (s dbLicenses) Create(ctx context.Context, subscriptionID, licenseKey strin
...
+			*info.SalesforceOpportunityID,
+			*info.SalesforceOpportunityID,
+		),
+	})
+	if err != nil {
+		logger.Error("error sending Slack message", log.Error(err))
+		return id, nil
+	}
+
 	return id, nil
```
camdencheek, line 154: Can we move all the added code to a helper function just so it's more clear that any failures do not cause the whole operation to fail?

schema/schema.go
dadlerj: Ah thank you!

camdencheek (approved):
LGTM!

PR #59215, opened 2023-12-22
bazel: migrate go:generate for staticcheck

Strum355:
Closes https://github.com/sourcegraph/sourcegraph/issues/54833

## Test plan

`bazel run //dev/linters/staticcheck:write_generated_bazelfile` and `cd dev/linters/staticcheck && go generate` as well as adding staticchecks to be ignored


Strum355:
@jhchabran @burmudar Is there a catch-all target for generate that we should add this one to?

dev/linters/staticcheck/staticcheck.go
```diff
@@ -1,4 +1,5 @@
-//go:generate go run ./cmd/gen.go BUILD.bazel
+//go:generate go run ./cmd/gen.go targets.bzl
+//go:generate buildifier -mode=fix targets.bzl analyzers.bzl
```
Strum355, line 2: Unfortunately dont have buildifier in buildkite, without it we have an extra newline :roll_eyes:

jamesmcnamara (approved):
I'll rubber stamp this in case it's blocked, but I can't purport to really get the ramifications of this yet.

dev/linters/staticcheck/staticcheck.go
Strum355: Solved without buildifier

burmudar:
> @jhchabran @burmudar Is there a catch-all target for generate that we should add this one to?

Yup. //dev:write_all_generated

Issue #58631, opened 2023-11-28
(P2) Allow admins to configure in site config whether clients can use chat/autocomplete/commands as features.


slimsag:
Allow site admins to put in place restrictions on which Cody features may be used client-side in VS Code, etc.

i.e. the ability to globally turn on/off the following for all users (or better, for groups of users, if easily implementable):

* Chat
* Autocomplete
* Commands
* TODO: ask Ryan Phillips if any other functionality is missing here.

This will require e.g. adding a new site config option, and propagating that to VS Code clients to have them decide which features users can use based on the Sourcegraph instance.

It may be a good starter task for someone looking to dive in to both codebases.

ayatofrench:
Hello!

I am new to the Sourcegraph repo and very interested in contributing! I would like to tackle this issue and its related issue https://github.com/sourcegraph/sourcegraph/issues/58630. However, I saw that this was on a roadmap and just wanted to make sure that this was open to external contributors.

Thank you!

cbart:
Taking this off the kanban board. I'll retriage once I'll work on cleaning up https://github.com/sourcegraph/sourcegraph/issues/58375

Issue #2559, opened 2019-03-04
Unclear or missing documentation for adding TLS to Sourcegraph language server extensions


KattMingMing:
After deploying a Sourcegraph instance and following the documentation for the Typescript language server, it was unclear how to add TLS to the language server.

The documentation can be found here: https://sourcegraph.com/extensions/sourcegraph/typescript

There's minimal information TLS in the documentation:

> TLS is optional but recommended for production deployments. It is used if TLS_KEY and TLS_CERT environment variables are set.

A couple questions (cc/ @chrismwendt and @felixfbecker ):
- Is the mention of TLS specific to the Kubernetes deployment?
    - If yes, why is it not necessary when using Docker?
- Do all language server extensions need to have TLS configured?
- Do we have other documentation for using Let's Encrypt on our code intel extensions?

---

### Tasks

- [] Documentation for TypeScript, Go and Python language servers.
- [] Screencast for setting up the Go language server with SSL.

chrismwendt:
Our docs are lacking in this area. I think it's in part because I haven't set up any language server in the same way a customer would (behind an auth proxy), let alone documented it.

If it's ok for the communication between the auth proxy and language server to be unencrypted, then I don't think it's necessary to set up TLS for the language server.

> - Is the mention of TLS specific to the Kubernetes deployment?

No, it applies to Docker, too.

> - Do all language server extensions need to have TLS configured?

I think it will work without TLS, but it's insecure without it, which is why I made the "Secure deployment" section stand out more https://sourcegraph.com/extensions/sourcegraph/go

> - Do we have other documentation for using Let's Encrypt on our code intel extensions?

Not that I know of.

ryan-blunden:
I'll take ownership of this as part of improving the docs for 3.2

KattMingMing:
Open PR for fixing Typescript docs: https://github.com/sourcegraph/sourcegraph-typescript/pull/122

ryan-blunden:
Fixed by https://github.com/sourcegraph/sourcegraph-typescript/pull/122

PR #59229, opened 2023-12-25
Adding Config options for Siteadmins to configure whether clients can use Chat/autocomplete/commands

arafatkatze:
Main Issue https://github.com/sourcegraph/sourcegraph/issues/58631
Related PR https://github.com/sourcegraph/dev-private/pull/82

## Test plan
Run in Dev mode and test the GQL by tweaking the site admin config.

### DIFF
--- a/cmd/frontend/graphqlbackend/site.go
+++ b/cmd/frontend/graphqlbackend/site.go
@@ -3,7 +3,6 @@ package graphqlbackend
 import (
 	"bytes"
 	"context"
-	"fmt"
 	"os"
 	"strconv"
 	"strings"
@@ -592,12 +591,14 @@ func (r *siteResolver) CodyConfigFeatures(ctx context.Context) *codyConfigFeatur
 	if c == nil {
 		return nil
 	}
-	fmt.Println("This is a reading")
-	fmt.Println(c)
 	return &codyConfigFeaturesResolver{config: c}
 }
 
 func (c *codyConfigFeaturesResolver) Chat() bool { return c.config.Chat }
+
+func (c *codyConfigFeaturesResolver) AutoComplete() bool { return c.config.AutoComplete }
+
+func (c *codyConfigFeaturesResolver) Commands() bool { return c.config.Commands }
 
 type codyLLMConfigurationResolver struct {
 	config *conftypes.CompletionsConfig
--- a/cmd/frontend/graphqlbackend/site.go
+++ b/cmd/frontend/graphqlbackend/site.go
@@ -587,23 +587,24 @@ func (r *siteResolver) CodyLLMConfiguration(ctx context.Context) *codyLLMConfigu
 	return &codyLLMConfigurationResolver{config: c}
 }
 
-func (r *siteResolver) CodyConfigFeatures(ctx context.Context) *codyLLMConfigurationResolver {
-	mySite := conf.Get().SiteConfig()
-	c := conf.GetCompletionsConfig(mySite)
+func (r *siteResolver) CodyConfigFeatures(ctx context.Context) *codyConfigFeaturesResolver {
+	c := conf.GetConfigFeatures(conf.Get().SiteConfig())
 	if c == nil {
 		return nil
 	}
-
-	wholeConf := conf.GetConfigFeatures(mySite)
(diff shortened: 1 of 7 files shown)


internal/conf/computed.go
```diff
@@ -793,6 +793,31 @@ func GetCompletionsConfig(siteConfig schema.SiteConfiguration) (c *conftypes.Com
 	return computedConfig
 }
 
+func GetConfigFeatures(siteConfig schema.SiteConfiguration) (c *conftypes.ConfigFeatures) {
+	// If cody is disabled, don't use any of the other features.
+	if !codyEnabled(siteConfig) {
+		return nil
+	}
+	configFeatures := siteConfig.ConfigFeatures
+	// If no completions configuration is set at all, but cody is enabled, assume a default configuration
```
cbart, line 802: I think you meant features configuration not completions configuration.

internal/conf/computed.go
```diff
@@ -793,6 +793,31 @@ func GetCompletionsConfig(siteConfig schema.SiteConfiguration) (c *conftypes.Com
...
+	// where all the features are enabled this is to handle edge cases where no config is set etc
+	if configFeatures == nil {
+		defaultConfig := &conftypes.ConfigFeatures{
+			Chat:         true,
+			AutoComplete: true,
+			Commands:     true,
+		}
+		return defaultConfig
```
cbart, line 810:
```suggestion
		return &conftypes.ConfigFeatures{
			Chat:         true,
			AutoComplete: true,
			Commands:     true,
		}
```

nit

cmd/frontend/graphqlbackend/site.go
cbart, line 1: Can you add a unit test where you change site config, query graphql and see the changes reflected?

arafatkatze: Added some testing: in a standard way to test these conftypes. Hopefully this is okay.

[Another example](https://github.com/sourcegraph/sourcegraph/blob/994286815e94d74deacf00819c4e8e9e5bf7eb40/internal/conf/computed_test.go#L317)

cbart (commented):
I just saw that I did not click submit in review! Apologies for the lag.

slimsag (approved):
This looks good to me as-is. It immediately lets us test / work with the client-side changes in #2559 so feel free to merge this.

As [discussed in Slack](https://sourcegraph.slack.com/archives/C05JDP433DL/p1704408478471069), I will send a follow-up PR soon (in the next day or so) to replace this approach with a RBAC-based approach instead which has a nicer UI and other benefits.

For now, though, this site config approach is great - so I say go ahead :)

schema/site.schema.json
```diff
@@ -2866,6 +2866,29 @@
           "perUserDailyLimit": 100
         }
       ]
+    },
+    "configFeatures": {
```
eseliger, line 2870: drive-by comment: 

The name `configFeatures` isn't very expressive, could we name this something like `cody.features` or so, and what's the relationship between this setting and `completions.enabled` and `cody.enabled` for example?

arafatkatze: @eseliger There are other places with deprecated placeholders for enabling/disabling completions/autocomplete and I could have probably used them but I wanted as simple and easy place to just make a clean GQL query to be able to test things.

I do agree that this is not full+ best naming and also misses some good tests like @cbart  suggested but merging this now just as a placeholder for my testing and we shall remove it on the coming the RBAC stuff.

internal/conf/computed_test.go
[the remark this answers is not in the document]
arafatkatze: @cbart Great suggestions Cezary I went and tried to implement them and leant a few things on the way but then I learnt for Stephen that we are going to replace all of this anyway for RBAC stuff so I am merging this now just for testing. Hopefully thats okay with you?

schema/site.schema.json
slimsag: Ack! I plan to remove this entirely and replace it with RBAC instead anyway, this is just helpful in the interim for testing the client-side changes primarily - so doesn't need to be perfect. I'll cc you on the new approach PR

PR #59231, opened 2023-12-26
release: cleanup sg release subcommand

BolajiOlajide:
I'm preparing to add some release automation to the sg tooling - this PR moves some of the cve checker functionality into the `dev/sg/internal` package so that the `sg_release.go` file doesn't get bulky.

## Test plan

* Existing CI tests should pass
* No new functionality was added so everything should work as they were.


cmd/server/shared/shared.go
```diff
@@ -290,7 +290,7 @@ func runMigrator() {
 	log.Println("Migrated postgres schemas.")
 }
 
-func shouldPostgresReindex() (shouldReindex bool) {
+func shouldPostgresReindex() bool {
```
BolajiOlajide, line 293: Snuck this one in while investigation why migrations are automatically executed in dev environment.

mohammadualam (approved):

PR #59232, opened 2023-12-26
client: Deprecate getFileInfo API

varungandhi-src:
See FIXME(id: language-detection) in the code for context,
and the linked issues therein.

## Test plan

n/a


varungandhi-src:
Closing this for now, see https://github.com/sourcegraph/sourcegraph/pull/59234

PR #59233, opened 2023-12-26
code-search: fix for regression introduced by actions discoverability

BolajiOlajide:
I did a bunch of testing on S2 and Dotcom, and I noticed several bugs introduced by #58122 - this PR takes an attempt to fix them.

The lack of a `max-height` attribute on the container for the context menu causes the list to grow abnormally long. 

CleanShot 2023-12-26 at 13 32 37@2x

Also, in a case where there are no items in the context menu (e.g the revision comparison page), the context menu renders nothing - we should hide it in this case instead of showing an empty context menu.

CleanShot 2023-12-26 at 13 44 01@2x

## Test plan

Navigating to a repository root page that has been indexed and contains a lot of visible indexes, the context menu shouldn't be abnormally long.

CleanShot 2023-12-26 at 13 41 18@2x

The radio button for indices shouldn't be hidden when the context menu is open.

### DIFF
--- a/client/web/src/enterprise/codeintel/dashboard/components/BrainDot.module.scss
+++ b/client/web/src/enterprise/codeintel/dashboard/components/BrainDot.module.scss
@@ -1,36 +1,9 @@
-.dropdown-menu {
-    min-width: 20rem;
+.container {
+    max-height: 15rem;
+    overflow-y: auto;
+    margin-left: 0.5rem;
 }
 
-.braindot {
-    position: relative;
-    margin-right: 0.625rem;
-    padding: 0.25rem;
-}
-
-.braindot::after {
-    content: '';
-    position: absolute;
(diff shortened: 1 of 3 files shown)


vovakulikov (approved):
Changes are ok, but I just tested the action width wrap layout on different screens and noticed that since wrapping is working based on fixed media query the whole layout looks off on the repository page on the middle size screens. Nothing that we can do about it now, but it tells me that we should prioritise work about actions refactoring before the 6.0 release. 

cc @rrhyne or @sourcegraph/team-graph  do you know anything about code intel things here, does anyone use it except us?

varungandhi-src:
@vovakulikov Apart from us, the debug view is used by (1) customers which develop their own precise SCIP indexers and (2) customers + us when sharing troubleshooting information related to precise code nav issues.

PR #59234, opened 2023-12-26
client: Obtain languages from server for file tree icons

varungandhi-src:
To determine the icon to show for a file, we need to at least
know the language. The code previously relied on the
getFileInfo API which did this using file extensions, leading to
incorrect results for header files with the '.h' extension,
which could have either C, C++ or Objective-C code.

Instead, we now get the language from the server,
where the go-enry library can use the file contents
(retrieved only if the filepath is insufficient)
to return a more suitable result.

Consequently, we can remove the getFileInfo API altogether.

This patch also updates the language detection logic in the backend
to work around some counter-intuitive behavior in go-enry.

Updates both the React and Svelte versions of the code.

Other changes in this patch:
- I've added icons for a bunch of languages which were missing.
- I've centralized the icon lists (and the bulk of the related logic) for
   both React and Svelte in one place.

Supersedes https://github.com/sourcegraph/sourcegraph/pull/59232, makes partial progress towards https://github.com/sourcegraph/sourcegraph/issues/56377.

## Test plan

TODO:
- [x] Test that new icons are showing up (see screenshot 1 for Mathematica)
- [x] Test that C vs C++ detection is working better (see screenshots 1, 2 and 3)
        (Requires the backend changes in this PR, so won't work with `pnpm dev` using SvelteKit.)
- [x] Test that icons are working for common repos
- [x] Test that the Svelte version also continues working (see screenshot 4)


client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte
```diff
@@ -125,11 +125,15 @@
...
+                            <Icon svgPath={icon.svg.path} inline --color={icon.svg.color} />
+                        {:else}
+                            <Icon svgPath={DEFAULT_FILE_ICON.path} inline --color={DEFAULT_FILE_ICON.color} />
+                        {/if}
                     {:else}
-                        {@const {
-                            icon: { svgPath, color },
-                        } = getFileInfo(entry.name)}
-                        <Icon {svgPath} inline --color={color} />
+                        <Icon svgPath={DEFAULT_FILE_ICON.path} inline --color={DEFAULT_FILE_ICON.color} />
```
Comment (varungandhi-src) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte:136]: There is a bit of duplication for the default case which seems like it'd be easily avoided in normal TypeScript code with a local mutable variable and a "if this is unset, only then set it" dance, but it's unclear how to do that with Svelte.

client/web/src/repo/tree/TreePagePanels.tsx
```diff
@@ -148,7 +148,7 @@ export interface DiffStat {
 }
 
 export interface FilePanelProps {
-    entries: Pick<TreeFields['entries'][number], 'name' | 'url' | 'isDirectory' | 'path'>[]
+    entries: TreeFields['entries']
```
varungandhi-src, line 151: I didn't know how to do a richer type-level dance around "if it's this case of the union, then pick these fields, otherwise pick those fields", so I've just removed the `Pick`, as the `languages` field is only exposed by the `GitBlob` case of the union.

lib/codeintel/languages/languages.go
```diff
@@ -22,62 +21,93 @@ func NormalizeLanguage(filetype string) string {
...
+		break
+	case 1:
+		return newLangs, nil
+	default:
+		langs = newLangs
+	}
+	if getContent == nil {
+		return langs, nil
+	}
+	content, err := getContent()
```
cbart, line 71: This is likely expensive in terms of latency. Do we have data showing it won't kill gitserver for some repos where file extensions are not determining language?

varungandhi-src: I'd be happy to measure the latency impact if you have a specific suggestion on how I should conduct this measurement to be somewhat realistic.

In my local testing, I haven't noticed any slowdown in loading the sidebar when testing against repos like `postgres/postgres` and `llvm/llvm-project` which have C and C++ code respectively.

I could potentially add artificial latency of X milliseconds to simulate the latency of gitserver in production, and maybe make a screen recording with thousands of header files in a single folder as an extreme case.

Would that be acceptable?

jasonhawkharris (commented):
Is there a world in which we could keep the logic on the client side? I understand the main objection in terms of languages/header files with the same file extensions, but is this more the rule or the exception? If it's the exception, couldn't we keep this on the client side and account for the edge case (i.e. c languages and matlab)?

varungandhi-src:
> I'm just a little concerned that this solution is complicating a process that could be and was much simpler, and that we might be able to account for those edge/corner cases differently. Completely open to being wrong about this!

The reason it was much simpler was because it wasn't taking into account the inherent complexity of language detection. 😅

For example, go-enry contains a probabilistic detector, including this single 10MB data file generated from analyzing the samples in Linguist data. https://github.com/go-enry/go-enry/blob/master/data/frequencies.go

So we cannot simply compile go-enry to WebAssembly or similar and ship that.

For fixing the `.h` issue, the file contents need to be available. That is fundamentally unavoidable. Given that, the two options would roughly be:
1. Get the file contents client-side and do client-side detection
2. Get the file contents on the server and do server-side detection

Option 1 requires more network traffic from server to client, is likely to be slower when running on the client (cannot be done in parallel + CPU power not in our control) and would require maintaining extra language detection code (we already have go-enry on the server).

jasonhawkharris:
@varungandhi-src That makes sense. Thanks for the explanation. 

Approving to unblock, but I have one more question. I'm still learning the codebase/database so this might be possible, but is there any way we could pre-calculate the language of every file when it's synced and add it as a field in the database so that we could access it directly when determining the icon?

Alternatively, could we determine language based on the context of the folder the file in question exists in? For example, if there is a single `.h` file in a folder and everything else is cpp, c or .m... wouldn't that be enough to accurately detect the language without needing to do so on the backend?

jasonhawkharris (approved):

lib/codeintel/languages/languages.go
varungandhi-src: I'll attempt to do some load testing tomorrow so that we can get a sense of numbers.

Also, introducing an escape hatch to turn this off seems like a reasonable option, thanks for the suggestion. I will look into implementing that tomorrow.

varungandhi-src:
> Approving to unblock, but I have one more question. I'm still learning the codebase/database so this might not be possible, but is there any way we could pre-calculate the language of a file when it's synced/updated and add it as a field in the database so that we could access it directly when determining the icon?

In theory, we could do that. The problem with such an approach is that it essentially introduces a cache, and cache invalidation is one of the 3 hard problems in computer science. Specifically, if you update the language detection code, then there is the risk that the languages for some files have changed -- do we then perform a read-modify-write on all rows of such a table (not necessarily cheap for huge C++ monorepos, need to ensure that this happens consistently) or do we put with stale reads (leading to issues where you update the language detection code but you still see outdated information until you update the table via some other process)? Neither of those is appealing.

The other problem with this approach would be high disk usage for database tables. In a way, git is a database specialized to storing the history of source code. Storing the language for every ambiguous file at every commit would get very expensive very quickly and would be very wasteful because most files at most commits won't be visited.

The disk usage problem can be worked around -- only store the languages for the latest commit on the default branch (`main` or `master`) -- and this is what Zoekt does (well, slightly different, because Zoekt stores a single language, not a list). However, this means that you still need to maintain the fallback path for gitserver anyways to get the languages list for any arbitrary commit.

More generally, if there is a performance issue, I think we should prioritize understanding the core issue (e.g. in this case, there is a lower bound enforced because there's a new git process being spawned + it attempting to read something from disk) and fixing that before attempting to use a cache. For example, this case, if it turns out that the process spawning overhead due to the `git` CLI invocations increase the CPU usage of gitserver by a large factor, that might be fixable by using `libgit2` so that a file can be retrieved using a few simple function calls rather than by using a separate `git` CLI invocation.

lib/codeintel/languages/languages.go
varungandhi-src: @cbart I've introduced a new experimental setting which can be configured to override the fetching of file contents.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte
Comment (varungandhi-src) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte]: Thanks for the suggestion, I've updated the code with that. 👍🏽

lib/codeintel/languages/languages.go
varungandhi-src: @cbart if you're satisfied with the benchmarks above, could you approve this PR? It'd be nice to get this merged as it will require some knock on changes in a different PR (https://github.com/sourcegraph/sourcegraph/pull/59178#discussion_r1437722153), but I don't want to merge this if there are outstanding concerns around performance.

vovakulikov:
@varungandhi-src Let me know about your plans for merging this PR, since I touched icon logic on the client [here](https://github.com/sourcegraph/sourcegraph/pull/59178) I'm kind of blocked by this PRs (I don't block but there will be some problems with rebasing) and I think this PR should go firs and then I will adjust my.

varungandhi-src:
OK, let's merge this for now, we can change the default value of the experimental setting if needed.

lib/codeintel/languages/languages.go
camdencheek: I'm actually pretty concerned by those numbers. Fetching the file contents for every interaction with the file sidebar is a lot of extra load on gitserver. Additionally, I expect in a this to be much worse in a non-local environent. 

Distribution of folder size is not a great proxy for the added latency because, when loading a file, we recursively load every expanded directory it's contained in so we can show the fully expanded file tree. 

I don't think a small increase in correctness of a visual aid is going to be worth the increase in cost here, especially since the file tree view has historically been a huge issue for monorepos. However, moving the file type detection to the backend is probably still a positive change so we can unify how we're determining file type and improve it moving forward (maybe by precalculating it).

cbart: @varungandhi-src apologies for not circling back. I'm glad you did not wait for my approval. I think @camdencheek's feedback is valuable. I'm sure there's a good path forward.

varungandhi-src: @camdencheek addressing your feedback:

> Fetching the file contents for every interaction with the file sidebar is a lot of extra load on gitserver

It's not for every interaction, it's only for the cases when the file extension isn't enough. So it should only trigger for `.h` files in the most common case. Most other file extensions should be unaffected.

> Distribution of folder size is not a great proxy for the added latency because, when loading a file, we recursively load every expanded directory it's contained in so we can show the fully expanded file tree.

In practice, the typical structure of headers is not deeply nested, even in large projects. Usually, there may be 1-2 levels of nesting even in larger projects for the `include/` directories. Usually, the directory structure ends up having only directories and build files (with maybe a few headers) in the internal nodes, with most of the headers being in the left nodes.

The 4ms/file latency roughly lines up with process spawning overhead on macOS. @eseliger mentioned there is [planned work on gitserver](https://sourcegraph.slack.com/archives/C05EMJM2SLR/p1704126731706369?thread_ts=1703827325.504129&cid=C05EMJM2SLR) to make certain git operations faster by avoiding exec-ing a new `git` process. I think this should address the issue for the tail of the distribution.

> I don't think a small increase in correctness of a visual aid is going to be worth the increase in cost here

I don't think we should ship this feature if it's going to be incorrect for C and C++, given those are among the top 10 languages we have. For a personal point of view, I'm sometimes in the position of recording demos or doing some live demo related to C++ code, and it's not a good look if Sourcegraph does not nail something as basic (from a customer's point-of-view, something they're used to in their IDE) as understanding what the language for a file is.

varungandhi-src: I just did some rough local benchmarks with `go-git` with the following script to get a sense for an order of magnitude.

<details>
<summary>

```go
package main

import (
	"fmt"
	"github.com/go-git/go-git/v5"
	"github.com/montanaflynn/stats"
	"io"
	"os"
	"strings"
	"time"
)

func main() {
	const dir = "../../cxx/llvm-headers"
	dirEntries, err := os.ReadDir(dir)
	if err != nil {
		panic(err)
	}

	timings := make([]time.Duration, 0, len(dirEntries))
	bufs := make([][]byte, 0, len(dirEntries))

	for _, dirEntry := range dirEntries {
		fileName := dirEntry.Name()
		if dirEntry.IsDir() || !strings.HasSuffix(fileName, ".h") {
			continue
		}
		start := time.Now()

		repo, err := git.PlainOpen(dir)
		if err != nil {
			panic(err)
		}
		worktree, err := repo.Worktree()
		if err != nil {
			panic(err)
		}
		file, err := worktree.Filesystem.Open(fileName)
		if err != nil {
			panic(err)
		}
		buf, err := io.ReadAll(file)
		if err != nil {
			panic(err)
		}
		bufs = append(bufs, buf)

		timings = append(timings, time.Now().Sub(start))
	}

	timingsMicros := []float64{}
	for _, timing := range timings {
		timingsMicros = append(timingsMicros,
			float64(timing)*0.001)
	}

	t := stats.Float64Data(timingsMicros)
	fmt.Printf("Percentile timings (micros): 50 - %.1f, 95 - %.1f, 99 - %.1f\n",
		percentile(t, 90.0), percentile(t, 95.0), percentile(t, 99.0))
}

func percentile(t stats.Float64Data, p float64) float64 {
	v, err := t.Percentile(p)
	if err != nil {
		panic(err)
	}
	return v
}
```

</summary>
</details>

With a warm disk cache, the 99% percentile timing is less than 100us (compare with ~4ms) for the whole process of opening a worktree and reading a file to the end, even when using go-git which is purportedly slower than libgit2. So I think if we're able to replace the git exec with go-git or a faster solution as planned, this should not be a problem even for extreme cases with hundreds or thousands of headers in the same folder.

camdencheek: Thank you for the detailed response!

> it's only for the cases when the file extension isn't enough.

Oh, I misunderstood that. That potentially makes a huge difference here. Thanks for making doing that 🙂 I do think it's still worth adding a metric (even just a temporary one without an associated grafana chart) to get an idea of the proportion of files this triggers on. I was under the impression there are quite a few file types that enry considers undecidable from filename. 

> Is your concern that the network overhead will be significant in a non-local set up?

Yes, that's mostly my concern. Macbooks have very low latency disks and local networking can make a significant difference. `sg` doesn't run in docker by default, so I wasn't expecting your benchmarking to be through the docker VM, but you might be right that it would effectively cancel out much local advantage.

> avoiding exec-ing a new git process

For sure! I'm very excited for that work 🙂 That doesn't remove the additional network request per file though. And I find git operations like this are petty sensitive to whether (and how) a repo is `gc`-ed and whether the content you need is in a packfile (which adds a few extra disk roundtrips), so I don't expect dropping the `exec` to be a panacea. 

---

In any case, you've convinced me to not be too worried by this. I'll make a followup PR with a metric tracking how frequently `enry` decides it needs file content, but between the "only fetch when it's ambiguous" and "we'll be minimizing execs in the future" I'm feeling better about this as a reasonable solution.

camdencheek: > So it should only trigger for .h files in the most common case. Most other file extensions should be unaffected.

> I was under the impression there are quite a few file types that enry considers undecidable from filename.

So, I added the metric locally, and was surprised to find that, just from clicking around in my local instance, almost half the language resolutions required fetching the content. A selection of examples:
```
titan.jpeg
command_list.md
.mailmap
queries.txt
pnpm-workspace.yaml
vitest.shared.ts
CODENOTIFY
.bazelignore
```

The ones that really stood out to me are that `.md`, `.yaml`, and `.ts` files all require fetching the full contents. We also end up fetching for binary files like `jpeg`, which seems particularly nasty since those can get quite large.

camdencheek: For the file tree, rather than fetching contents when unsure, maybe we should just use a generic icon if we're not confident the language is correct?

varungandhi-src: 1. For those cases, I would consider the `.md`, `.yaml` and `.ts` ones as just bugs because the other languages are quite obscure. I haven't looked at the other file extensions.
2. For binary files, there was a fast path in the frontend which avoids it for common binary files, so maybe there's a bug in the extension checking logic there. This seems fixable.

For fixing 1, we already have the `languages` package as a wrapper around working around odd behavior in `enry` and providing better APIs, and as a step towards #56376. I think we should add a wrapper for `enry`'s `GetLanguagesByExtension` similar to the wrappers we have for `GetLanguages` and `GetLanguagesByShebang` to work around it returning obscure languages in some cases.

I can take a stab at both of these during the next working day for me.

camdencheek: > there was a fast path in the frontend 

Are you talking about [this](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@56554758e89dd7ae90ee7f16125859278350576f/-/blob/lib/codeintel/languages/languages.go?L78-80)? Because that seems to only trigger after fetching the content

PR #59236, opened 2023-12-26
code-search: hide ownership pages on dotcom and refactor rbac permissions

BolajiOlajide:
On Dotcom, we display the ownership panel in the Repository Page, however users are taken to a blank page when they click on the `Show more` link.

CleanShot 2023-12-26 at 17 23 05

In an attempt to fix this, I noticed (when in `!dotcom` environment) that we display a `Add Owner` button when a user doesn't have the required permission to add an owner. Selecting an owner and trying to save results in the user greeted with an error, `You don't have the OWNERSHIP#ASSIGN permission`. We can avoid this by checking the permission of the authenticated user.

This further led to another rabbit hole which resulted in refactoring how we generate permissions for rbac so they are strongly typed.

## Test plan

* Existing CI checks should pass

* The ownership panel should not displayed on Dotcom

CleanShot 2023-12-26 at 17 30 15@2x


client/web/src/util/permission.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const doesUserHavePermission = (
+    user: Pick<AuthenticatedUser, 'permissions'> | null,
+    permissionToCheckFor: RbacPermission
+): boolean => {
+    if (user === null) {
+        return false
+    }
+
+    return findIndex(user.permissions.nodes, permission => permission.displayName === permissionToCheckFor) !== -1
```
camdencheek, line 14: Why lodash here? Couldn't we just use builtins?

```suggestion
    return user.permissions.nodes.some(permission => permission.displayName === permissionToCheckFor)
```

internal/rbac/gen/yamldata.go
```diff
@@ -126,11 +126,15 @@ func loadSchema(filename string) (*schema, error) {
 
 func generateTSConstants(output io.Writer, permissions []permissionNamespace) {
 	fmt.Fprintln(output, TSGeneratedByTarget)
-	for _, permission := range permissions {
+	var permissionNames = make([]string, len(permissions))
```
camdencheek, line 129: nit: use `:=` for combo declaration/assignment

```suggestion
	permissionNames := make([]string, len(permissions))
```

camdencheek (approved):
Nice! Always happy to see some more RBAC land 😄

PR #59242, opened 2023-12-27
client: Delete unused code involving getModeForPath

varungandhi-src:
The blob view now has access to all the potential languages.
See: https://github.com/sourcegraph/sourcegraph/pull/58962

So we can remove this call to `getModeForPath` as the return
value is not used.

## Test plan

Type-checking in CI should be enough


jasonhawkharris (approved):
Nice. Deleting code rules.

PR #59243, opened 2023-12-27
Add logic to auto disable google signups from non-gmail domain for 24hrs if too many signups from same domain.

thenamankumar:
1. Add logic to auto-disable Google signups from non-Gmail domains during a 24-hour window if too many (> 30) signups from the same domain.
2. Check if is email domain is banned based on SRC_EMAIL_DOMAIN_DENY_LIST for social logins as well. Currently only used for user password signup.

## Test plan
.

### DIFF
--- a/cmd/frontend/auth/user.go
+++ b/cmd/frontend/auth/user.go
@@ -12,6 +12,7 @@ import (
 
 	sglog "github.com/sourcegraph/log"
 
+	"github.com/sourcegraph/sourcegraph/cmd/frontend/envvar"
 	sgactor "github.com/sourcegraph/sourcegraph/internal/actor"
 	"github.com/sourcegraph/sourcegraph/internal/auth"
 	"github.com/sourcegraph/sourcegraph/internal/authz"
@@ -21,6 +22,7 @@ import (
 	"github.com/sourcegraph/sourcegraph/internal/errcode"
 	"github.com/sourcegraph/sourcegraph/internal/extsvc"
 	"github.com/sourcegraph/sourcegraph/internal/featureflag"
+	"github.com/sourcegraph/sourcegraph/internal/security"
 	"github.com/sourcegraph/sourcegraph/internal/usagestats"
 	"github.com/sourcegraph/sourcegraph/lib/errors"
 )
@@ -127,6 +129,48 @@ func GetAndSaveUser(ctx context.Context, db database.DB, op GetAndSaveUserOp) (n
 
 		act := &sgactor.Actor{
 			SourcegraphOperator: acct.AccountSpec.ServiceType == auth.SourcegraphOperatorProviderType,
+		}
+
+		if envvar.SourcegraphDotComMode() {
+			domain, _ := security.ParseEmailDomain(op.UserProps.Email)
+
+			banned, err := security.IsEmailBanned(op.UserProps.Email)
+			if err != nil {
+				return 0, false, false, "could not determine if email domain is banned", err
+			}
+
+			if banned {
+				recorder.Record(ctx, telemetry.FeaturesSignUpBlockedBannedDomain, telemetry.ActionFailed, &telemetry.EventParameters{
+					PrivateMetadata: map[string]any{
+						"serviceType": acct.AccountSpec.ServiceType,
+						"serviceId":   acct.AccountSpec.ServiceID,
+						"emailDomain": domain,
+					},
+				})
+
+				return 0, false, false, "this email address is not allowed to register", errors.New("email domain banned")
+			}
+
+			//if acct.AccountSpec.ServiceID == "https://accounts.google.com" && domain != "gmail.com" {
+			if true {
+				banned, err := security.IsEmailBlockedDueToTooManySignups(op.UserProps.Email)
+
+				if err != nil {
+					return 0, false, false, "could not determine if email domain is banned", err
+				}
+
+				if banned {
+					recorder.Record(ctx, telemetry.FeaturesSignUpBlockedTooManySignups, telemetry.ActionFailed, &telemetry.EventParameters{
+						PrivateMetadata: map[string]any{
+							"serviceType": acct.AccountSpec.ServiceType,
+							"serviceId":   acct.AccountSpec.ServiceID,
+							"emailDomain": domain,
+						},
+					})
+
+					return 0, false, false, "this email address is not allowed to register", errors.New("Email not allowed to register")
+				}
+			}
 		}
 
 		// Fourth and finally, create a new user account and return it.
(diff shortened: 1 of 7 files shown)


cmd/frontend/auth/user.go
```diff
@@ -129,6 +137,69 @@ func GetAndSaveUser(ctx context.Context, db database.DB, op GetAndSaveUserOp) (n
...
+			}
+
+			//if acct.AccountSpec.ServiceID == "https://accounts.google.com" && domain != "gmail.com" {
+			if true {
+				key := redisScopeKey + domain
+
+				value := redisStore.Get(key)
+
+				if value.IsNil() {
+					redisStore.SetEx(key, 24*60*60, 1)
```
evict, line 175:
```suggestion
					redisStore.SetEx(key, 24 * time.Hour, 1)
```

cmd/frontend/auth/user.go
```diff
@@ -129,6 +137,69 @@ func GetAndSaveUser(ctx context.Context, db database.DB, op GetAndSaveUserOp) (n
...
+							Source:         "BACKEND",
+						},
+					)
+					if err != nil {
+						logger.Error("failed to log event", sglog.String("eventName", "SignupBlocked:TooManySignups"), sglog.Error(err))
+					}
+					return 0, false, false, "this email address is not allowed to register", errors.New("Email not allowed to register")
+				} else {
+					emailsRegisteredInLast24Hours, err = redisStore.Incr(key)
+				}
```
evict, line 199: Can we move this code into a separate function? You could put this in the security package for instance, make it so that we could call it like `IsSignUpAllowedForDomain(serviceID, domain)`? Then we can easily re-use this for other sign-ups.

evict (commented):
If this is specifically for Google Workspace, can we move the checks to here: https://sourcegraph.com/github.com/sourcegraph/sourcegraph@1c1e795c2777444182be17769ae075b9623f35a5/-/blob/cmd/frontend/internal/auth/openidconnect/middleware.go?L313-318? 

We can then also add a field to the site-config so we don't have to hardcode the account auth provider and the limts:

```
  "auth.providers": [
    {
      "type": "openidconnect",
      "issuer": "https://accounts.google.com", // All Google Workspace domains use this issuer URI.
      "clientID": "my-client-id",
      "clientSecret": "my-client-secret",
      "requireEmailDomain": "example.com"
      "maxDomainSignUps": 30,
    }
```

cmd/frontend/auth/user.go
thenamankumar: Have moved this to a separate function inside security package.

thenamankumar:
> If this is specifically for Google Workspace, can we move the checks to here

As this is experimental and may extend to other signup options based on abuse behaviour in future, I have not moved the logic. 

I have also not made the limit controlled by the site config for now. Recently, it has been tricky to update the configuration and then wait for the cache to refresh, making fighting with abuse even more difficult. 

If we decide to keep this logic for longer, will integrate it with the site config. This is highly unlikely though as the new SAMS system is under works and this abuse prevention logic will move there.

internal/security/security.go
```diff
@@ -50,32 +51,76 @@ func ensureBannedEmailDomainsLoaded() error {
...
+	if err != nil {
+		return false, err
+	}
 
-	if len(parts) < 2 {
+	key := redisScopeKey + domain
+	value := redisStore.Get(key)
+
+	if value.IsNil() {
+		redisStore.SetEx(key, 24*60*60, 1)
```
evict, line 113: Can we handle the error for `SetEx`? And you can use `int(24 * time.Hour)` here for readability as well.

cmd/frontend/auth/user.go
```diff
@@ -129,6 +131,70 @@ func GetAndSaveUser(ctx context.Context, db database.DB, op GetAndSaveUserOp) (n
...
+						Argument:       args,
+						PublicArgument: args,
+						Source:         "BACKEND",
+					},
+				)
+				if err != nil {
+					logger.Error("failed to log event", sglog.String("eventName", "SignupBlocked:BannedDomain"), sglog.Error(err))
+				}
+				return 0, false, false, "this email address is not allowed to register", errors.New("email domain banned")
+			}
```
evict, line 163: This logic is repeated in the if-statement below as well. Can you create a function instead?

evict (commented):
Did you manage to test the changes? If so, can you provide that inside the test plan ?

schema/site.schema.json
```diff
@@ -1121,6 +1121,11 @@
         }
       ]
     },
+    "auth.dailyEmailDomainSignupLimit": {
+      "type": "integer",
+      "default": 0,
```
chenkc805, line 1126: Should this be 30?

chenkc805:
Thanks Naman. What's the error message we show to users if their domain gets blocked for 24 hours from too many signups?

chenkc805 (changes requested):

schema/site.schema.json
thenamankumar: No by default lets keep this 0 (disabled) in the code. I'll create a separate PR in separate repo to update the prod site config and set it to 30. 

Also just to confirm with you, 30 sounds good?

chenkc805: Gotcha. Yeah 30 sounds good.

chenkc805: And confirming, 0 won't automatically ban all non-gmail domains right? Because if I follow the logic:
- Add logic to auto-disable Google signups from non-Gmail domains during a 24-hour window if too many (> N) signups from the same domain.
  - If N = 0, then all non-gmail domains will be banned. 

If I'm following, should this be a very high number instead, like 10000 as a default?

thenamankumar: N = 0 means disabled. This check won't be performed. It's taken care of in the code.

chenkc805:
> Thanks Naman. What's the error message we show to users if their domain gets blocked for 24 hours from too many signups?

@thenamankumar friendly bump on this ^^

thenamankumar:
> > Thanks Naman. What's the error message we show to users if their domain gets blocked for 24 hours from too many signups?
> 
> @thenamankumar friendly bump on this ^^

"this email address is not allowed to register"

thenamankumar:
> Did you manage to test the changes? If so, can you provide that inside the test plan ?

I did test the non-google auth flows and they seem to work fine. Google Oauth2.0 [doesn't allow](https://developers.google.com/identity/protocols/oauth2/policies#secure-response-handling) local Oauth connections. 

Except the one if condition everything else is well tested.  `if acct.AccountSpec.ServiceID == "https://accounts.google.com"`

Verified from the prod DB that the serviced will be `"https://accounts.google.com"` only.

chenkc805:
Can we change the message to "There seem to be too many signups occurring today. Please try again at a later time." 

I know this may be giving away some information but I'm hoping that the actual blocking mechanism is annoying enough and that this copy is vague enough for a bad actor but kind of helpful for real users. 

Once that's changed, I'll approve.

thenamankumar:
Error message updated @chenkc805

pjlast (approved):

vdavid:
This is merged but the test plan is still empty. I think we need test plans for compliance reasons, plus also, as a team member it's helpful for me to know how much we tested this.

vdavid:
Forgot to ping @thenamankumar with my last message

PR #59245, opened 2023-12-27
Using new function for security even logs

mohammadualam:
Since the audit log expansion, there is a more streamlined way of logging security events. The changes here are porting over the code to the new way of invoking a log event. Logs stay the same. 

## Test plan
Tested locally


evict (approved):
Nice work! 🙏

PR #59246, opened 2023-12-28
cody-gateway: include operation name as query param

bobheadxi:
Part of my brief investigation looking at dotcom instability. It seems we have an internal standard where GraphQL requests can be "named" using the raw query param, e.g. `/.api/graphql?Foobar` - this is reflected in `slow http request` logs and metrics.

This change updates the Cody Gateway GraphQL client to do the same. This works nicely because all our GraphQL operations have names for use with the GraphQL generator.

## Test plan

Unit test


unknwon (approved):

rafax (approved):
Nice :)

PR #59247, opened 2023-12-28
bzl: Add cody* node_modules to .bazelignore

varungandhi-src:
I'm not sure when exactly this broke, but the absence of these ignores
caused my type-checking invocation to stop working:

```
bash -c 'bazel test `bazel query "attr(\"name\", \".*_typecheck_test\", //...)"`'
```

as `bazel query` would try to read files deep inside `node_modules`

## Test plan

n/a


varungandhi-src:
Oops, looks like I have stale `node_modules`

Issue #59262, opened 2024-01-01
Back-end requests guardrails search via gateway


cbart:

cbart:
https://sourcegraph.slack.com/archives/C05497E9MDW/p1705057527416409 asking how to get a product subscription token for gateway dev.

cbart:
works now!

```
➜  sourcegraph git:(main) time curl -H "Authorization: bearer $ATTRIBUTION_TOKEN" -d '{"snippet":"sourcegraph.Location(new VRR", "limit": 2}' https://cody-gateway.sgdev.org/v1/attribution
{"Repositories":null,"TotalCount":0,"LimitHit":false}
curl -H "Authorization: bearer $ATTRIBUTION_TOKEN" -d    0.01s user 0.01s system 3% cpu 0.663 total
➜  sourcegraph git:(main) time curl -H "Authorization: bearer $ATTRIBUTION_TOKEN" -d '{"snippet":"sourcegraph.Location(new URL", "limit": 2}' https://cody-gateway.sgdev.org/v1/attribution
{"Repositories":[{"Name":"github.com/sourcegraph/sourcegraph"},{"Name":"npm/sourcegraph/basic-code-intel"}],"TotalCount":2,"LimitHit":true}
curl -H "Authorization: bearer $ATTRIBUTION_TOKEN" -d    0.01s user 0.01s system 4% cpu 0.449 total
```

PR #59263, opened 2024-01-01
Gateway attribution handler

cbart:
Part of https://github.com/sourcegraph/sourcegraph/issues/59262

Misses issuing the graphQL request to dotcom (follow-up PR coming)

## Test plan

Component tests at the entry point of gateway middle-ware/routing (dotcom GraphQL API mocked out):

- Successful scenario - runs attribution search (TODO misses some elements)
- Failure scenario - only authorizes product subscriptions.


cmd/cody-gateway/internal/httpapi/attribution/handler.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// NewHandler creates a REST handler for attribution search.
+// graphql.Client can be nil which disables the search.
+func NewHandler(client graphql.Client, baseLogger log.Logger) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		a := actor.FromContext(r.Context())
+		if got, want := a.GetSource(), codygateway.ActorSourceProductSubscription; got != want {
+			w.WriteHeader(http.StatusUnauthorized)
+			fmt.Fprintln(w, "only available for enterprise product subscriptions")
```
bobheadxi, line 48: We have a shared helper in Cody Gateway to render the error in a structured manner and generate debug/error logs based on status code:

```suggestion
            response.JSONError(logger, w, errors.New("only available for enterprise product subscriptions")
```

cmd/cody-gateway/shared/main.go
```diff
@@ -171,7 +173,9 @@ func Main(ctx context.Context, obctx *observation.Context, ready service.ReadyFu
 			FireworksDisableSingleTenant:                config.Fireworks.DisableSingleTenant,
 			EmbeddingsAllowedModels:                     config.AllowedEmbeddingsModels,
 			AutoFlushStreamingResponses:                 config.AutoFlushStreamingResponses,
-		})
+			EnableAttributionSearch:                     os.Getenv("CODY_GATEWAY_ENABLE_ATTRIBUTION_SEARCH") == "true",
```
bobheadxi, line 176: nit - add this to config struct alongside the other config options

cmd/cody-gateway/internal/httpapi/attribution/handler.go
```diff
@@ -0,0 +1,75 @@
...
+
+// Repository represents matching of search content against a repository.
+type Repository struct {
+	// Name of the repo on dotcom. Like github.com/sourcegraph/sourcegraph.
+	Name string
+}
+
+// NewHandler creates a REST handler for attribution search.
+// graphql.Client can be nil which disables the search.
+func NewHandler(client graphql.Client, baseLogger log.Logger) http.Handler {
```
chwarwick, line 43: For search typically we default to using the streaming search endpoint over graphql.

cmd/cody-gateway/shared/main.go
chwarwick: ➕

cmd/cody-gateway/internal/httpapi/handler.go
```diff
@@ -230,6 +234,18 @@ func NewHandler(
 			otelhttp.WithPublicEndpoint(),
 		),
 	)
+
+	var attributionClient graphql.Client
+	if config.EnableAttributionSearch {
+		attributionClient = dotcomClient
+	}
+	v1router.Path("/attribution").Methods(http.MethodPost).Handler(
+		instrumentation.HTTPMiddleware("v1.attribution",
```
chwarwick, line 243: Will we be rate limiting this feature per subscription id?  If we do there is possible conflict with chat/autocomplete rate limits, but feels like a need.

cmd/cody-gateway/internal/httpapi/handler.go
```diff
@@ -230,6 +234,18 @@ func NewHandler(
...
+			authr.Middleware(
+				attribution.NewHandler(attributionClient, logger),
+			),
+		),
```
chwarwick, line 247:
```suggestion
		),
		otelhttp.WithPublicEndpoint(),
```
Like the other handlers, admittedly I'm not 100% sure on this, but appears to be related to linking traces to the original sg requests

cmd/cody-gateway/shared/main.go
chwarwick: ➕

cmd/cody-gateway/internal/httpapi/attribution/handler.go
bobheadxi: If the attribution endpoint itself isn't streaming, I think the GraphQL endpoint might be easier to integrate against - IIRC holding the streaming endpoint right is a bit tricky (from my time building the Sourcegraph extension for Raycast)

cmd/cody-gateway/internal/httpapi/handler.go
bobheadxi: It shouldn't conflict, as the feature limiting is handled within the individual handlers (e.g. `makeUpstreamHandler`). I don't think this needs a rate limit immediately, since it just calls search and we don't rate-limit search (I think), but maybe we have a billing component to this feature?

cmd/cody-gateway/internal/httpapi/handler.go
bobheadxi: `otelhttp.WithPublicEndpoint` is the right configuration for all Cody Gateway endpoints 👍

cmd/cody-gateway/internal/httpapi/handler.go
chwarwick: I meant a conflict in that you could run out of attribution requests but still have available autocomplete/chat requests in which case attribution would no longer work. 

We do rate limit anonymous search to dotcom, this won't be anonymous I assume it will eventually use the service account.

arafatkatze: My understand on conversation with Cezary is that the search attribution calls are not going to have rate limits(the requests are routed in a way that the upstream doesn't have rate limits coz if user account is used for search query then rate limits are too strict to do anything).

Also, there aren't self re-inforcing rate limits either in the sense that, If the Chat/autocompletions were rate limited then that doesn't mean that this will be rate limited as a subset. The check of rate limiting happens in a separate route so if two async calls are made then one is rated limited and the call to the search attribution is not.

cmd/cody-gateway/shared/main.go
unknwon: I think this is not a nit 😂

cmd/cody-gateway/internal/httpapi/attribution/handler.go
chwarwick: Probably not an issue since I think the use case is not for an exhaustive search but just several results.

cmd/cody-gateway/internal/httpapi/handler.go
chwarwick: If we aren't rate limiting search on dotcom and this remains only available to enterprise customers then it feels ok to omit for now.  

@arafatkatze  In general we need to be careful, we can't assume that public endpoints will be consumed only by our clients and for the intended purpose.

arafatkatze: > If we aren't rate limiting search on dotcom and this remains only available to enterprise customers then it feels ok to omit for now.

Okay makes sense.

>  In general we need to be careful, we can't assume that public endpoints will be consumed only by our clients and for the intended purpose.

Fair, but it is also true that public endpoints are rate limited for user accounts etc. In this case we use a special token/account(I am kinda hazy on the details), thats free from rate limits(Thats what cezary said to me). 

Am I correctly interpreting this?

cmd/cody-gateway/shared/main.go
cbart: Done!

cbart:
Please take another look!

cmd/cody-gateway/internal/httpapi/handler.go
chwarwick: Yes the other endpoints on the gateway are rate limited for users and subscriptions but this one as currently constructed isn't.

Since these requests will not be associated with a dotcom user, there will need to be a service account used to issue the actual searches.

cmd/cody-gateway/internal/httpapi/attribution/handler.go
chwarwick: Thought about this some more, should this be a streaming endpoint? It will be in the hotpath of features so time to first data is likely important. 

Also I can't recall any new use of the gql search endpoint since I've been here.  

We shouldn't need to hold a reference to client, I assume it would work similar to the streaming endpoints for anthropic/openai.  Looks like this reference is purely for testing purposes.  

Also isn't something I consider blocking.

chwarwick (approved):
Considering this is off by default I don't see any issues here now.  I would consider switching the endpoint to stream results back (even if you keep the gql search for now).

cbart:
Thanks @chwarwick! Let's talk more. I don't really understand the benefits of streaming approach if we really care about a coupld of results (which we have a chance to present in the small UI surface)

PR #59266, opened 2024-01-02
Minor refactoring and remove redundant checks

pjlast:
Thought it might be easier to do feedback like this :)

### DIFF
--- a/cmd/frontend/auth/user.go
+++ b/cmd/frontend/auth/user.go
@@ -179,7 +179,7 @@ func GetAndSaveUser(ctx context.Context, db database.DB, op GetAndSaveUserOp) (n
 		}
 
 		if envvar.SourcegraphDotComMode() {
-			err = checkIfEmailDomainIsBanned(ctx, recorder, op, acct)
+			reason, err := checkIfEmailDomainIsBanned(ctx, recorder, op, acct)
 			if err != nil {
 				return 0, false, false, reason, err
 			}
--- a/cmd/frontend/auth/user.go
+++ b/cmd/frontend/auth/user.go
@@ -37,6 +37,53 @@ type GetAndSaveUserOp struct {
 	LookUpByUsername    bool
 }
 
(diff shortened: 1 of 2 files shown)


cmd/frontend/auth/user.go
```diff
@@ -132,43 +179,9 @@ func GetAndSaveUser(ctx context.Context, db database.DB, op GetAndSaveUserOp) (n
 		}
 
 		if envvar.SourcegraphDotComMode() {
-			domain, _ := security.ParseEmailDomain(op.UserProps.Email)
-
-			banned, err := security.IsEmailBanned(op.UserProps.Email)
+			err = checkIfEmailDomainIsBanned(ctx, recorder, op, acct)
```
pjlast, line 182: I factored out this bit to its own function so it can be reasoned about on its own. This GetUserAndSave function is a pain enough as it is

internal/security/security.go
```diff
@@ -51,22 +51,17 @@ func ensureBannedEmailDomainsLoaded() error {
 	return bannedEmailDomainsErr
 }
 
+// Retrieves the domain name from an email address.
+// e.g. "user@example.com" -> "example.com"
+// The email address must be valid, or else an error is returned.
 func ParseEmailDomain(email string) (string, error) {
```
pjlast, line 57: I removed some redundant checks here that seem to be already covered by `mail.ParseAddress` .

cmd/frontend/auth/user.go
```diff
@@ -393,6 +406,3 @@ func addCodyProForTestUsers(ctx context.Context, db database.DB, userID int32, l
 		}
 	}
 }
-
-func logBannedDomainEvent(ctx context.Context, db database.DB, logger sglog.Logger, reason string, domain string, acct *extsvc.Account) {
-}
```
pjlast, line 398: This seems unused

thenamankumar (approved):

PR #59267, opened 2024-01-02
Adjust license anomalies checks

pjlast:
This PR reworks the way we do our license checks a bit with the following goals:

- Have license checks be less noisy
- Ensure license alerts are actionable

Originally, once a site ID was assigned to a license key, it would be permanently assigned and any pings from a new license key would trigger alerts. However, there are valid reasons under which a customer's site ID could change, such as them redeploying Sourcegraph.

In cases where customers are using the same license key on their dev and prod instances, the only way to fix this would be for someone to issue **two** new license keys for the customer: one for each instance. This is because the license key would be claimed by the dev instance, and then be rendered useless to any other instance without some SQL magic.

**AFTER THIS PR**

We now assign the site ID from the latest ping to the license key and send an alert in our Slack channel whenever a site ID associated with a license key changes. This solves two issues:

- Customer can redeploy Sourcegraph and use the same license key. We will only receive one alert when this happens, but this is perfectly reasonable.
- If we repeatedly receive site ID change alerts for the same customer, it means the license key is being shared between two instances. This issue can be resolved by simply  issuing a (possibly more limited) dev license key to customers, after which the alerts will stop and the prod instance will claim the correct license key instead.

We also have a bit more detail about who the license key belongs to to make it easier to follow up with.

## Test plan

Unit tests adjusted

### DIFF
--- a/cmd/frontend/internal/dotcom/productsubscription/license_check_handler.go
+++ b/cmd/frontend/internal/dotcom/productsubscription/license_check_handler.go
@@ -6,16 +6,15 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/google/uuid"
-	"github.com/keegancsmith/sqlf"
 	"github.com/sourcegraph/log"
 
 	"github.com/sourcegraph/sourcegraph/internal/authz"
 	"github.com/sourcegraph/sourcegraph/internal/conf"
 	"github.com/sourcegraph/sourcegraph/internal/database"
-	"github.com/sourcegraph/sourcegraph/internal/database/basestore"
 	"github.com/sourcegraph/sourcegraph/internal/licensing"
 	"github.com/sourcegraph/sourcegraph/internal/slack"
 )
@@ -35,15 +34,12 @@ var (
 	EventNameAssigned = "license.check.api.assigned"
 )
 
-// 🚨 SECURITY: we store the license check access token instead of the license key itself.
-func logEvent(ctx context.Context, db database.DB, name string, siteID string, accessToken string) {
+func logEvent(ctx context.Context, db database.DB, name string, siteID string) {
 	logger := log.Scoped("LicenseCheckHandler logEvent")
 	eArg, err := json.Marshal(struct {
-		SiteID      string `json:"site_id,omitempty"`
-		AccessToken string `json:"access_token,omitempty"`
+		SiteID string `json:"site_id,omitempty"`
 	}{
-		SiteID:      siteID,
-		AccessToken: accessToken,
+		SiteID: siteID,
 	})
 	if err != nil {
 		logger.Warn("error marshalling json body", log.Error(err))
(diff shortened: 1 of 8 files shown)


cmd/frontend/internal/dotcom/productsubscription/licenses_db.go
```diff
@@ -348,8 +347,7 @@ ORDER BY created_at DESC
 	}
 
 	if featureflag.FromContext(ctx).GetBoolOr("auditlog-expansion", false) {
-
-		//Log an event when liscenses list is viewed in Dotcom
+		// Log an event when liscenses list is viewed in Dotcom
```
burmudar, line 350:
```suggestion
		// Log an event when license list is viewed in Dotcom
```

internal/types/types.go
```diff
@@ -866,6 +866,16 @@ type User struct {
 	CodyProEnabledAt      *time.Time
 }
 
+// Returns a name for the user. If the user has a display name,
```
burmudar, line 869:
```suggestion
// Name returns a name for the user. If the user has a display name,
```
gif

cmd/frontend/internal/dotcom/productsubscription/license_check_handler.go
```diff
@@ -79,14 +93,38 @@ func sendSlackMessage(logger log.Logger, license *dbLicense, siteID string) {
...
+			logger.Warn("could not find user for subscription", log.String("subscriptionID", dbSubscription.ID), log.Error(err))
+		} else {
+			customerName = user.Name()
+		}
+	}
+
+	return customerName
+}
+
+// Creates a new license check handler that uses the provided database.
```
burmudar, line 124:
```suggestion
// NewLicenseCheckHandler creates a new license check handler that uses the provided database.
```

cmd/frontend/internal/dotcom/productsubscription/license_check_handler.go
```diff
@@ -79,14 +93,38 @@ func sendSlackMessage(logger log.Logger, license *dbLicense, siteID string) {
...
 }
 
+func getCustomerNameFromLicense(ctx context.Context, logger log.Logger, db database.DB, license *dbLicense) string {
+	// Best effort fetch of customer name for slack message
+	customerName := "could not load customer name"
+
+	subscriptionsStore := dbSubscriptions{db: db}
+	dbSubscription, err := subscriptionsStore.GetByID(ctx, license.ProductSubscriptionID)
+	if err != nil {
+		logger.Warn("could not find subscription for license", log.String("licenseID", license.ID), log.Error(err))
```
burmudar, line 111: Should this have `license.ProductSubscriptionID` as well?

cmd/frontend/internal/dotcom/productsubscription/license_check_handler.go
```diff
@@ -79,14 +93,38 @@ func sendSlackMessage(logger log.Logger, license *dbLicense, siteID string) {
...
+	} else {
+		user, err := db.Users().GetByID(ctx, dbSubscription.UserID)
+		if err != nil {
+			logger.Warn("could not find user for subscription", log.String("subscriptionID", dbSubscription.ID), log.Error(err))
```
burmudar, line 115: Does it make sense to have the userID here too?

cmd/frontend/internal/dotcom/productsubscription/license_check_handler.go
```diff
@@ -161,19 +211,18 @@ func NewLicenseCheckHandler(db database.DB) http.Handler {
 					Reason:  ReasonLicenseIsAlreadyInUseMsg,
 				},
 			})
-			sendSlackMessage(logger, license, siteID)
-			return
-		}
 
-		if license.SiteID == nil {
+			oldSiteID := *license.SiteID
 			if err := lStore.AssignSiteID(r.Context(), license.ID, siteID); err != nil {
```
burmudar, line 216: This might just be semantics but here we tell the store to assign the siteID, but only afterwards do we make the license struct consistent. Should `AssignSiteID` return the updated license instead? Then you don't need to do line 220? It seems like an easy thing to miss

PR #59281, opened 2024-01-02
fix: replace 'replace' with 'edit'

abeatrix:
Close https://github.com/sourcegraph/cody/issues/2186

There is no `replace` mode in custom command. The correct mode should be `edit`: https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/lib/shared/src/chat/prompts/index.ts?L81

## Test plan

This is a doc update


MaedahBatool (approved):
LGTM 👍

abeatrix:
@MaedahBatool Woops I just found out the docs has been moved to https://github.com/sourcegraph/docs 
Will create a PR in that repo and request your review again, sorry about that

PR #59283, opened 2024-01-02
Code Navigation: Add aggregated counts for various result types

jasonhawkharris:
This PR adds previously unavailable aggregate counts for certain filter types. The intention is to use the new counts to enhance the dynamic filters present in the new filters panel. 

- Symbols are now counted by symbol type
- Commit authors are now counted by author
- Commit dates are now counted by date, using common use cases

## Test plan
Ships with unit tests


internal/search/streaming/search_filters.go
```diff
@@ -5,19 +5,97 @@ import (
...
+	ONE_YEAR_AGO     = "1 year ago"
+	THREE_MONTHS_AGO = "3 months ago"
+	TWO_MONTHS_AGO   = "2 months ago"
+	ONE_MONTH_AGO    = "1 month ago"
+	TWO_WEEKS_AGO    = "2 weeks ago"
+	ONE_WEEK_AGO     = "1 week ago"
+	TODAY            = "today"
+)
+
+type DateFilterInfo struct {
```
camdencheek, line 32: We should probably unexport this struct

internal/search/streaming/search_filters.go
```diff
@@ -5,19 +5,97 @@ import (
...
+const (
+	AFTER            = "after"
+	BEFORE           = "before"
+	ONE_YEAR_AGO     = "1 year ago"
+	THREE_MONTHS_AGO = "3 months ago"
+	TWO_MONTHS_AGO   = "2 months ago"
+	ONE_MONTH_AGO    = "1 month ago"
+	TWO_WEEKS_AGO    = "2 weeks ago"
+	ONE_WEEK_AGO     = "1 week ago"
+	TODAY            = "today"
```
camdencheek, line 29: I think this is probably too many intervals. In order to keep the filters panel concise (and fit everything in a single screenful), we'll probably want to limit this to 4-5 items

internal/search/streaming/search_filters.go
```diff
@@ -107,6 +185,36 @@ func (s *SearchFilters) Update(event SearchEvent) {
...
+		for _, sym := range symbols {
+			selectKind := result.ToSelectKind[strings.ToLower(sym.Symbol.Kind)]
+			filter := fmt.Sprintf(`select:symbol.%s`, selectKind)
+			s.filters.Add(filter, selectKind, 1, limitHit, "symbol type")
+		}
+	}
+
+	addCommitAuthorFilter := func(commit gitdomain.Commit) {
+		author := fmt.Sprintf(`author:%s`, commit.Author.Email)
+		filter := fmt.Sprintf(`type:commit %s`, author)
```
camdencheek, line 198: similar to the reasons we don't need `type:symbol` above, we shouldn't need `type:commit` here

internal/search/streaming/search_filters.go
```diff
@@ -107,6 +185,36 @@ func (s *SearchFilters) Update(event SearchEvent) {
...
+		var filter string
+
+		if commit.Committer != nil {
+			cd = commit.Committer.Date
+		} else {
+			cd = commit.Author.Date
+		}
+
+		df = determineTimeframe(cd)
+		filter = fmt.Sprintf("%s:%s", df.Timeframe, df.Value)
```
camdencheek, line 214: No need to pre-declare variables. The `:=` operator declares and initializes a variable at the same time, which helps mitigate bugs from using zero-valued variables. 

```suggestion
		var cd time.Time
		if commit.Committer != nil {
			cd = commit.Committer.Date
		} else {
			cd = commit.Author.Date
		}

		df := determineTimeframe(cd)
		filter := fmt.Sprintf("%s:%s", df.Timeframe, df.Value)
```

internal/search/streaming/search_filters.go
```diff
@@ -5,19 +5,97 @@ import (
...
+	now := time.Now()
+
+	switch {
+	case date.After(now.Add(-25 * time.Hour)):
+		return DateFilterInfo{
+			Timeframe: AFTER,
+			Value:     TODAY,
+			Label:     "today",
+		}
+	case date.After(now.Add(-8 * 24 * time.Hour)):
```
camdencheek, line 48: Why are we adding an extra day to the week?

internal/search/streaming/search_filters.go
```diff
@@ -5,19 +5,97 @@ import (
...
+		}
+	case date.Before(now.Add(-367 * 24 * time.Hour)):
+		return DateFilterInfo{
+			Timeframe: BEFORE,
+			Value:     ONE_YEAR_AGO,
+			Label:     "before one year ago",
+		}
+
+	default:
+		return DateFilterInfo{}
```
camdencheek, line 92: I think we can merge these cases, right? If the time wasn't after a year ago, then it was before a year ago and we can just put that in a default case rather than returning the empty struct.

internal/search/streaming/search_filters.go
jasonhawkharris: This is what I determined we needed to do since `date.After()` uses the `>` operator to compare the two time values, [not the `>=` operator](https://sourcegraph.com/github.com/golang/go@6db1102605f227093ea95538f0fe9e46022ad7ea/-/blob/src/time/time.go?L256). Adding 1 to each value is necessary. For example, if the value was set at 7, we would only account for 6 days. 8 makes sure we account for 7. Definitely open to a better way of doing this. If we decide to keep it this way, I'll add a comment to the code explaining why.

internal/search/streaming/search_filters.go
jasonhawkharris: I ended up going with just three values because it looks like that's what Vova has already configured to appear in the UI.

`after:"yesterday"`
`before:"1 week ago"`
`before:"1 month ago"`

internal/search/streaming/search_filters.go
camdencheek: I don't actually think it's necessary because `time.Time` goes down to nanosecond precision. So, `date.After(now.Add(-24 * time.Hour))` would include timestamps up to 23h59m59s999ms999ns old. So you are not incorrect that it's a `>=` operator, but there is an approximately one nanosecond window where that would matter 😄

client/web/src/search/results/StreamingSearchResults.tsx
```diff
@@ -34,15 +34,15 @@ import { useStreamingSearchPings } from './useStreamingSearchPings'
...
-        ExtensionsControllerProps {
+    Pick<SearchContextProps, 'selectedSearchContextSpec' | 'searchContextsEnabled'>,
+    SettingsCascadeProps,
+    PlatformContextProps,
+    TelemetryProps,
+    CodeInsightsProps,
+    SearchAggregationProps,
+    CodeMonitoringProps,
+    OwnConfigProps,
+    ExtensionsControllerProps {
```
camdencheek, line 45: unrelated formatting change snuck in? I suspect CI will complain

internal/search/streaming/search_filters.go
```diff
@@ -107,6 +158,33 @@ func (s *SearchFilters) Update(event SearchEvent) {
...
+
+	addCommitAuthorFilter := func(commit gitdomain.Commit) {
+		filter := fmt.Sprintf(`author:%s`, commit.Author.Email)
+		s.filters.Add(filter, commit.Author.Name, 1, false, "author")
+	}
+
+	addCommitDateFilter := func(commit gitdomain.Commit) {
+		var cd time.Time
+
+		if commit.Committer != nil {
```
camdencheek, line 177: super tiny style nit: when variables are declared then initialized in an `if` branch, I find it easier to read if it's grouped with the initializer block.
```suggestion
		var cd time.Time
		if commit.Committer != nil {
```

camdencheek (approved):

jasonhawkharris:
This PR adds previously unavailable aggregate counts for certain filter types. The intention is to use the new counts to enhance the dynamic filters present in the new filters panel.

Symbols are now counted by symbol type (i.e. function, enum, variable, etc...)
Commit authors are now have a count for each author
Commit dates now have a count for each date use case

PR #59284, opened 2024-01-02
Allow last_used_at to be semi-stale

chrsmith:
It appears that we update a token's `last_used_at` value every time `AccessTokenStore.Lookup` is called. However, this causes far more writes than are useful.

This PR updates `Lookup()` to just have a read-only query, and only update the token's `last_used_at` if it is been greater than five minutes. The tradeoff being the common case of token access will be much faster, but once every five minutes -- or first lookup in a while -- will require 2x database queries. And moving forward the `last_used_at` value may be up to 5m stale.

## Test Plan

Added unit tests for the new functionality.

### DIFF
--- a/cmd/frontend/internal/dotcom/productsubscription/tokens_db.go
+++ b/cmd/frontend/internal/dotcom/productsubscription/tokens_db.go
@@ -2,10 +2,14 @@ package productsubscription
 
 import (
 	"context"
+	"database/sql"
 	"encoding/hex"
+	"errors"
 	"strings"
+	"time"
 
 	"github.com/keegancsmith/sqlf"
+
 	"github.com/sourcegraph/sourcegraph/internal/accesstoken"
 
 	"github.com/sourcegraph/sourcegraph/internal/database"
@@ -83,33 +87,56 @@ func (e dotcomUserNotFoundError) NotFound() bool {
 	return true
 }
 
-// LookupDotcomUserIDByAccessToken returns the userID
-// corresponding to a token, trimming token prefixes if there are any.
+// LookupDotcomUserIDByAccessToken returns the userID corresponding to the given token.
+// Requires the token has the DotcomUserGatewayAccessTokenPrefix, otherwise fails.
 func (t dbTokens) LookupDotcomUserIDByAccessToken(ctx context.Context, token string) (int, error) {
 	if !strings.HasPrefix(token, accesstoken.DotcomUserGatewayAccessTokenPrefix) {
 		return 0, dotcomUserNotFoundError{reason: "invalid token with unknown prefix"}
 	}
-	decoded, err := hex.DecodeString(strings.TrimPrefix(token, accesstoken.DotcomUserGatewayAccessTokenPrefix))
+	rawToken := strings.TrimPrefix(token, accesstoken.DotcomUserGatewayAccessTokenPrefix)
+	decoded, err := hex.DecodeString(rawToken)
 	if err != nil {
 		return 0, dotcomUserNotFoundError{reason: "invalid token encoding"}
 	}
 
+	// Query the token's id, subject_user_id, and last_used_at.
 	query := sqlf.Sprintf(`
-UPDATE access_tokens t SET last_used_at=now()
-WHERE t.id IN (
-	SELECT t2.id FROM access_tokens t2
-	JOIN users subject_user ON t2.subject_user_id=subject_user.id AND subject_user.deleted_at IS NULL
-	JOIN users creator_user ON t2.creator_user_id=creator_user.id AND creator_user.deleted_at IS NULL
-	WHERE digest(value_sha256, 'sha256')=%s AND t2.deleted_at IS NULL
-)
-RETURNING t.subject_user_id`,
-		decoded,
+	SELECT t.id, t.subject_user_id, t.last_used_at
+	FROM access_tokens t
+	WHERE t.id IN (
+		SELECT t2.id
+		FROM access_tokens t2
+		JOIN users subject_user ON t2.subject_user_id=subject_user.id AND subject_user.deleted_at IS NULL
+		JOIN users creator_user ON t2.creator_user_id=creator_user.id AND creator_user.deleted_at IS NULL
+		WHERE digest(value_sha256, 'sha256')=%s AND t2.deleted_at IS NULL
+	)`,
+		decoded)
+
+	var (
+		tokenID    int64
+		subjectID  int
+		lastUsedAt time.Time
 	)
-	userID, found, err := basestore.ScanFirstInt(t.store.Query(ctx, query))
+	row := t.store.QueryRow(ctx, query)
+	err = row.Scan(&tokenID, &subjectID, &lastUsedAt)
 	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return 0, dotcomUserNotFoundError{reason: "no associated token"}
+		}
 		return 0, err
-	} else if !found {
-		return 0, dotcomUserNotFoundError{reason: "no associated token"}
 	}
-	return userID, nil
+
+	// If the token hasn't been used recently, update the last_used_at value
+	// so indicate it is still in-use.
+	if time.Since(lastUsedAt) > database.MaxAccessTokenLastUsedAtAge {
+		// We ignore the error on updating the token, since hopefully we can just
+		// update the last used at time successfully the next time the token gets used.
+		updateQuery := sqlf.Sprintf(
+			`UPDATE access_tokens t SET last_used_at=now()
+			WHERE t.id=%d AND t.deleted_at IS NULL`,
+			tokenID)
+		_ = t.store.Exec(ctx, updateQuery)
+	}
+
+	return subjectID, nil
 }
--- a/cmd/frontend/internal/dotcom/productsubscription/tokens_db.go
+++ b/cmd/frontend/internal/dotcom/productsubscription/tokens_db.go
@@ -4,13 +4,13 @@ import (
 	"context"
 	"database/sql"
 	"encoding/hex"
-	"errors"
 	"strings"
 	"time"
 
 	"github.com/keegancsmith/sqlf"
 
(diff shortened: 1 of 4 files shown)


bobheadxi:
> I don't have the sg tool setup and/or sourcegraph/sourcegraph running locally. So I have no idea if this works as expected.

I recommend the setup, it's quick and easy 😁 Guide is here: https://docs.sourcegraph.com/dev/setup/quickstart 

`sg` will also be home of MSP tools to e.g. connect to SAMS/SSC service database, and more: https://sourcegraph.slack.com/archives/C063HAXQFT2/p1703110306038849

internal/database/access_tokens.go
```diff
@@ -283,26 +283,40 @@ type TokenLookupOpts struct {
 	RequiredScope string
 }
 
-func (o TokenLookupOpts) ToQuery() string {
+// Returns a query to upload the token's LastUsedAt column to "now". The returned query
+// requires two parameters: the token's hash and required scope.
+func (o TokenLookupOpts) ToUpdateLastUsedQuery() string {
```
bobheadxi, line 288: Since an update always follows a get query, maybe the "get" query can return `access_tokens.id` and we can directly update the access token's `last_used_at` column without querying over multiple indexes: https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/internal/database/schema.md?L23-27

internal/database/access_tokens.go
```diff
@@ -316,16 +330,33 @@ func (s *accessTokenStore) Lookup(ctx context.Context, token string, opts TokenL
 		return 0, errors.Wrap(err, "AccessTokens.Lookup")
 	}
 
-	if err := s.Handle().QueryRowContext(ctx,
+	var (
+		subjectId  string
```
unknwon, line 334:
```suggestion
		subjectID  string
```

😁

unknwon (approved):
🚀

internal/database/access_tokens.go
bobheadxi: We could probably use `value_sha256` directly as well to determine the row to update, since the "get" query will assert the other requirements hold first

thenamankumar: Does it need to be publically accessible? Why not keep it internal as `toUpdateLastUsedQuery`?

+1 to @bobheadxi idea of using `value_sha256` directly as the get query already asserts all the other parameters.

thenamankumar (approved):
Approving to unblock. concerns:

 https://github.com/sourcegraph/sourcegraph/pull/59284/files/030517a6c352db098d4abb7f9a6adc3c584bf48d#r1441567664
 
 https://github.com/sourcegraph/sourcegraph/pull/59284#issuecomment-1877061577

thenamankumar:
@chrsmith we need to do the same debounce thing for this other function, which performs an access token lookup and updates last_used_at. https://sourcegraph.com/github.com/sourcegraph/sourcegraph@f6fadab93438045d507984127371530aa4c9b213/-/blob/cmd/frontend/internal/dotcom/productsubscription/tokens_db.go?L88

I think this is the original culprit, as, after Dec GA, Gateway calls dotcom to fetch limits for each user separately using the access token. Earlier, the user-level rate limit used to happen only in dotcom.

internal/database/access_tokens.go
chrsmith: Done.

I had the `toGetQuery()` return the token's `id` and not the `value_sha256`. Both have an index associated with them, but it seems like just updating a token by its primary key seems better. Is there a reason why we would want to have the UPDATE query identify a token by its `value_sha256` column instead?

bobheadxi: Using `id` LGTM as well, `value_sha256` would have been the "lazy" way 😄

internal/database/access_tokens.go
```diff
@@ -126,7 +126,7 @@ type AccessTokenStore interface {
 	//
 	// The token prefix "sgp_", if present, is stripped.
 	//
-	// Calling Lookup also updates the access token's last-used-at date.
+	// Calling Lookup also updates the access token's last-used-at date as applicable.
```
bobheadxi, line 129: Should we mention the internal `maxTokenLastUsedAtAge` const here so that readers can easily drill down to where the condition is defined?

bobheadxi:
Thanks for working on this - I don't want to keep piling on asks, but I think we might want to update `testAccessTokens_Lookup` in `internal/database/access_tokens_test.go` (not 100% sure why the tests are set up like this but that looks like where this stuff is defined) to have a case that asserts that the "update last_used_at" path works as expected. The easiest way might be to have a new `TokenLookupOpts` field that lets you set the current time for testing, or force an update?

internal/database/access_tokens.go
chrsmith: Yeah. @thenamankumar pointed out another place where we need to do the same thing, so I made `MaxAccessTokenLastUsedAtAge` a constant exported from the `database` package. (And then referenced in `dotcom/productsubscription`.)

chrsmith:
I have written some tests, which (😢) found some bugs. And as soon as I can trick the pre-commit hook into letting me update `access_tokens_test.go` we'll be in business...

cmd/frontend/internal/dotcom/productsubscription/tokens_db.go
```diff
@@ -83,33 +87,56 @@ func (e dotcomUserNotFoundError) NotFound() bool {
...
+	// If the token hasn't been used recently, update the last_used_at value
+	// so indicate it is still in-use.
+	if lastUsedAt == nil || time.Since(*lastUsedAt) > database.MaxAccessTokenLastUsedAtAge {
+		// We ignore the error on updating the token, since hopefully we can just
+		// update the last used at time successfully the next time the token gets used.
+		updateQuery := sqlf.Sprintf(
+			`UPDATE access_tokens t SET last_used_at=now()
+			WHERE t.id=%d AND t.deleted_at IS NULL`,
+			tokenID)
+		_ = t.store.Exec(ctx, updateQuery)
```
bobheadxi, line 138: reasoning makes sense, but maybe we should do a warning log at least?

bobheadxi:
> I have written some tests, which (😢) found some bugs.

😌 whew!

cmd/frontend/internal/dotcom/productsubscription/tokens_db.go
chrsmith: 100% agree! The reason I'm not doing that though is I don't know how 😅 .

The function doesn't take a `log.Logger` but does have a `context.Context`. Should I instantiate a new logger instance? That seemed a bit much, but I wasn't sure if there was another way to write to the log without wiring through the logger instance to all callers.

bobheadxi: Ah, I think the right way would be to wire the logger instance to the `dbTokens` struct, i.e. at 

```go
type dbTokens struct {
	store *basestore.Store
}

func newDBTokens(db database.DB) dbTokens {
	return dbTokens{store: basestore.NewWithHandle(db.Handle())}
}
```

bobheadxi: however, I think we should be safe to ignore this for now to at least roll this out and monitor if it improves things :)

internal/database/access_tokens.go
```diff
@@ -306,17 +320,32 @@ func (s *accessTokenStore) Lookup(ctx context.Context, token string, opts TokenL
...
 		return 0, err
 	}
-	return subjectUserID, nil
+
+	if lastUsedAt == nil || time.Until(*lastUsedAt) < -MaxAccessTokenLastUsedAtAge {
+		s.logger.Debug("Updating access token's last_used_at value.")
+		_, err := s.Handle().ExecContext(ctx, opts.toUpdateLastUsedQuery(), tokenID)
+		if err != nil {
+			s.logger.Warn("Error trying to update token's last_used_at value.", log.Error(err))
+		}
```
bobheadxi, line 345:
```suggestion
        logger := s.logger.With(log.Int64("tokenID", tokenID), log.Int32("subjectID", subjectID)) 
		_, err := s.Handle().ExecContext(ctx, opts.toUpdateLastUsedQuery(), tokenID)
		if err != nil {
			logger.Warn("error trying to update token's last_used_at value", log.Error(err))
		} else {
		    logger.Debug("updated access token's last_used_at value")
		}
```

bobheadxi: https://docs.sourcegraph.com/dev/how-to/add_logging#fields-sub-loggers

cmd/frontend/internal/dotcom/productsubscription/tokens_db_test.go
```diff
@@ -58,4 +59,74 @@ func TestLookupProductSubscriptionIDByAccessToken(t *testing.T) {
...
+		// Create a new access token.
+		subject, err := db.Users().Create(ctx, database.NewUser{
+			Email:                 "u1@example.com",
+			Username:              "u1",
+			Password:              "p1",
+			EmailVerificationCode: "c1",
+		})
+		if err != nil {
+			t.Fatal(err)
+		}
```
bobheadxi, line 73: you can use `require.NoError(...)` and other `testify/assert`/`testify/require` helpers if you want (though no need to rewrite to do that, it's just a bit more concise in tests)

chrsmith: Yeah, I'm a big fan of that. But since the file wasn't using it already I didn't want to do a big refactoring. Or write inconsistent tests. But yes, 💯 .

bobheadxi (approved):
Thank you for working on this 😁😁

internal/database/access_tokens.go
thenamankumar: 👍

thenamankumar:
Thanks for the fix @chrsmith !

bobheadxi:
Manual backport created in https://github.com/sourcegraph/sourcegraph/pull/59553

PR #59288, opened 2024-01-03
Search: remove TextPatternInfo.IsWordMatch

jtibshirani:
Some small clean-ups I noticed while getting up to speed on `TextPatternInfo`:
* Remove `IsWordMatch`, since this is always `false`
* Correct comments

## Test plan

Refactor-only change covered by existing tests


internal/searcher/v1/searcher.proto
```diff
@@ -142,6 +142,7 @@ message PatternInfo {
   bool is_structural = 4;
 
   // is_word_match if true will only match the pattern at word boundaries.
+  // DEPRECATED: this field is no longer used
```
jtibshirani, line 145: Maybe I could just remove this?

camdencheek: Yep! It can be removed. For proto definitions, standard practice is to `reserve` the field number so it's not reused in the future.

camdencheek (approved):

PR #59292, opened 2024-01-03
productsubscription: fix panic on license creation

bobheadxi:
Fixes a panic when rendering a notification introduced in https://github.com/sourcegraph/sourcegraph/pull/59214 when an optional field is referenced. This change dereferences those fields safely as "unknown" and adds a test over zero values of various inputs to do a simple sanity-check that it doesn't blow up.

I also spotted a stray `log.Error` without a logger - fixed that while I was in here.

## Test plan

Unit test panicked without change, no longer panics with change

### DIFF
--- a/cmd/frontend/internal/dotcom/productsubscription/licenses_db.go
+++ b/cmd/frontend/internal/dotcom/productsubscription/licenses_db.go
@@ -20,7 +20,10 @@ import (
 	"github.com/sourcegraph/sourcegraph/internal/hashutil"
 	"github.com/sourcegraph/sourcegraph/internal/license"
 	"github.com/sourcegraph/sourcegraph/internal/slack"
+	"github.com/sourcegraph/sourcegraph/internal/trace"
+	"github.com/sourcegraph/sourcegraph/internal/types"
 	"github.com/sourcegraph/sourcegraph/lib/errors"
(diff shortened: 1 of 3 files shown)


cmd/frontend/internal/dotcom/productsubscription/licenses_db.go
```diff
@@ -95,69 +102,75 @@ func (s dbLicenses) Create(ctx context.Context, subscriptionID, licenseKey strin
 
 		// Log an event when a license is created in DotCom
 		if err := s.db.SecurityEventLogs().LogSecurityEvent(ctx, database.SecurityEventNameDotComLicenseCreated, "", uint32(actor.FromContext(ctx).UID), "", "BACKEND", nil); err != nil {
-			log.Error(err)
+			logger.Error("LogSecurityEvent", log.Error(err))
```
bobheadxi, line 105: @mohammadualam `log.Error` from package `sourcegraph/log` only creates a log field, not a log entry - for a log entry you need a `logger`. Can you check if you might have made this mistake elsewhere as well? 🙏 thank you!

camdencheek (approved):
thank you for the fix!

dadlerj (approved):
Thanks!

PR #59302, opened 2024-01-03
fixing error handling

mohammadualam:
The error returned from the security event log was not handled appropriately in a few places. this PR fixes those issues.

Thanks to @bobheadxi for catching this.

## Test plan

Locally tested


bobheadxi (approved):
Thanks!

PR #59306, opened 2024-01-03
Searcher: fix issues with negated patterns

jtibshirani:
This refactors the searcher logic to handle negated patterns within `matcher`, instead of the main file processing loop. This will be really helpful when adding support for 'and' and 'or' matchers to searcher, since now negation is handled at the level of a query atom.

The refactor uncovered two bugs in the negation logic, which are now fixed:
* When the pattern was negated, in some cases we ignored file include/ exclude patterns
* Hybrid search was totally ignoring pattern negation

Relates to #59038

## Test plan

New unit tests to cover the bug fixes

### DIFF
--- a/cmd/searcher/internal/search/matcher.go
+++ b/cmd/searcher/internal/search/matcher.go
@@ -26,12 +26,11 @@ type matcher interface {
 	// IgnoreCase returns whether matches will ignore case
 	IgnoreCase() bool
 
-	// MatchesString returns whether the pattern matches the given string
+	// MatchesString returns whether the string matches
 	MatchesString(s string) bool
 
-	// MatchesFile returns a LineMatch for each line that matches rm in reader.
-	// LimitHit is true if some matches may not have been included in the result.
-	MatchesFile(fileBuf []byte, limit int) [][]int
+	// MatchesFile returns whether the file matches, plus a LineMatch for each line that matches
+	MatchesFile(fileBuf []byte, limit int) (match bool, matches [][]int)
 
 	// ToZoektQuery returns a zoekt query representing the same rules as as this matcher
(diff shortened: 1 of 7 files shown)


cmd/searcher/internal/search/search_regex.go
```diff
@@ -87,7 +85,7 @@ func regexSearch(
 		// Fast path for only matching file paths (or with a nil pattern, which matches all files,
 		// so is effectively matching only on file paths).
 		for _, f := range files {
-			if match := pm.Matches(f.Name) && m.MatchesString(f.Name); match == !isPatternNegated {
+			if match := pm.Matches(f.Name) && m.MatchesString(f.Name); match {
```
jtibshirani, line 88: This logic was incorrect before -- `pm.Matches(...)` checks whether the include/ exclude patterns are satisfied. These are a separate component of the query from the pattern, and should not be affected by pattern negation.

cmd/searcher/internal/search/matcher.go
```diff
@@ -192,24 +206,36 @@ func (rm *regexMatcher) ToZoektQuery(matchContent bool, matchPath bool) (zoektqu
...
 		), nil
 	}
-	return &zoektquery.Regexp{
-		Regexp:        re,
-		Content:       matchContent,
-		FileName:      matchPath,
-		CaseSensitive: !rm.ignoreCase,
-	}, nil
+
+	return rm.negateIfNeeded(
```
jtibshirani, line 227: Before, we forgot to apply negation when delegating to Zoekt for hybrid search.

stefanhengl: Nice catch!

stefanhengl (approved):

PR #59311, opened 2024-01-03
productsubscription: stabilize license notification test

bobheadxi:
Previously this test depended on `time.Until` - instead, we inject a stable `time.Time` as `now` and evaluate the difference based on that.

## Test plan

Autogold assertion shows the expected time diff (3 days) - this should be stable now


mohammadualam (approved):

mohammadualam (approved):

Strum355 (approved):
🙏

PR #59312, opened 2024-01-03
Svelte: add topic badges to repo search results

camdencheek:
We already had metadata badges. This just adds additional badges for topics to match the behavior of the react webapp. I think we can upgrade repo metadata out of experimental as well and show it by default, but I'll do that in a followup PR.

CleanShot 2024-01-03 at 15 26 52@2x

## Test plan

Manually tested that the links work as expected.


fkling (approved):
Nice!

fkling:
I don't think it closes the linked issue though. That was about showing this information on the repository page itself.

camdencheek:
> That was about showing this information on the repository page itself.

Oh, gotcha! Removed the link 👍

PR #59314, opened 2024-01-04
msp: base startup latency alert on probe, add docstring and labels

bobheadxi:
This change makes the startup latency alert based on `3 retries x startup probe latency`, such that the alert fires if a container fails to start up based on the startup probe. This accounts for the default Cloud Run startup probe, which is 240 seconds.

In particular, this is currently firing for the `cloud-ops` service, which is expected to take along time to start up. It uses no startup probe so it should get the 240s threshold.

## Test plan

`cloud-ops` service gets very high threshold:

image


dev/managedservicesplatform/spec/environment.go
```diff
@@ -335,12 +340,24 @@ type EnvironmentServiceStartupProbeSpec struct {
 	// Defaults to 1 second.
 	Timeout *int `yaml:"timeout,omitempty"`
 	// Interval configures the interval, in seconds, at which to
```
daxmc99, line 342:
```suggestion
	// Interval configures the frequency, in seconds, at which to
```

daxmc99 (approved):

PR #59315, opened 2024-01-04
msp: supress all opsgenie alerts for now

bobheadxi:
We may have jumped the gun here - let's dogfood the alerts for a while with just Slack, and add Opsgenie further down the line. We may want to allow services to opt out of Opsgenie entirely with a mechanism other than Category as well.

## Test plan

n/a

### DIFF
--- a/dev/managedservicesplatform/internal/stack/monitoring/monitoring.go
+++ b/dev/managedservicesplatform/internal/stack/monitoring/monitoring.go
@@ -118,11 +118,12 @@ func NewStack(stacks *stack.Set, vars Variables) (*CrossStackOutput, error) {
 	var channels []monitoringnotificationchannel.MonitoringNotificationChannel
 
 	// Configure opsgenie channels
-	var opsgenieAlerts bool
-	switch vars.EnvironmentCategory {
-	case spec.EnvironmentCategoryInternal, spec.EnvironmentCategoryExternal:
-		opsgenieAlerts = true
-	}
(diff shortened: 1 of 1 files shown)


dev/managedservicesplatform/internal/stack/monitoring/monitoring.go
```diff
@@ -149,7 +150,8 @@ func NewStack(stacks *stack.Set, vars Variables) (*CrossStackOutput, error) {
 				// Supress all notifications if opsgenieAlerts is disabled -
 				// this allows us to see the alerts, but not necessarily get
 				// paged by it.
-				SuppressNotifications: pointers.Ptr(!opsgenieAlerts),
+				// TODO: Enable after we dogfood the alerts for a while.
+				SuppressNotifications: pointers.Ptr(false),
```
jac, line 154: Shouldn't this be true?
([docs](https://registry.terraform.io/providers/opsgenie/opsgenie/latest/docs/resources/api_integration#suppress_notifications))
```suggestion
				SuppressNotifications: pointers.Ptr(true),
```

jac (changes requested):

dev/managedservicesplatform/internal/stack/monitoring/monitoring.go
bobheadxi: aha!

jac (approved):

Issue #57791, opened 2023-10-23
Stop emitting local symbols for non-local entities


varungandhi-src:
Example:

https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/docker-images/syntax-highlighter/crates/scip-syntax/src/snapshots/scip_syntax__locals__test__can_do_go.snap

Here, top-level functions are marked as locals.

We should go through the other snapshots too to fix this.

The frontend has special code paths for local symbols when a SCIP document is available (https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/client/web/src/enterprise/codeintel/useSearchBasedCodeIntel.ts?L142:47, https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/client/shared/src/codeintel/api.ts?L293:55); that will give the wrong behavior when combining with this.

Some more issues flagged here: https://github.com/sourcegraph/sourcegraph/pull/57664/files/478dc1d87108d53252e8060d6ad2d48fe98e0716

- [x] Go https://github.com/sourcegraph/sourcegraph/pull/59320
- [x] Java https://github.com/sourcegraph/sourcegraph/pull/59376
- [x] Perl https://github.com/sourcegraph/sourcegraph/pull/59465
- [x] MATLAB https://github.com/sourcegraph/sourcegraph/pull/59415

PR #59320, opened 2024-01-04
Improves locals queries for Go

kritzcreek:
Fixes the go specific part of #57791

Stops recording top-level functions and var bindings as locals with our Go queries.

To allow skipping top-level var bindings I added the `skip` feature we talked about yesterday to our locals DSL.

## Test plan

Tested locally and in our test snapshots


varungandhi-src:
> Fixes the go specific part of ...

I've added a checklist to the issue so that we can track what languages need updating. I've also linked this PR from the Go item. For future PRs, it'd be helpful to check things off that checklist and link the PRs similarly. (Mentioning this in case I forget to update that issue myself)

docker-images/syntax-highlighter/crates/scip-syntax/src/snapshots/scip_syntax__locals__test__can_do_functions.snap
```diff
@@ -11,33 +11,30 @@ expression: dumped
   )
   
   func Something() {
-//     ^^^^^^^^^ definition local 2
```
varungandhi-src, line 14: Please update this test to make sure that we're not emitting local symbols for other non-local entities like:
- consts
- types
- field names in types
- methods in interface definitions
- references to "inherited" interfaces in interface definitions: `type B interface {}; type A interface { B }`
- type lists in interface definitions like `interface { int | string }`
- method definitions (this test currently only has free functions)

I'm not 100% sure if that covers everything.

docker-images/syntax-highlighter/crates/scip-syntax/src/snapshots/scip_syntax__locals__test__can_do_go.snap
```diff
@@ -4,26 +4,41 @@ expression: dumped
 ---
   package main
   
+  // Not actually local
+  var local = 10
+  
   func main() {
```
varungandhi-src, line 10: It seems like the 3 different test files are essentially checking the behavior of locals, just for different scoping structures/patterns, right? Thoughts on merging them? It seems a bit hard to see what cases are covered/what cases are missing when the tests are split across different files but are conceptually testing the same thing.

docker-images/syntax-highlighter/crates/scip-syntax/queries/go/scip-locals.scm
```diff
@@ -2,23 +2,37 @@
...
- name: ((identifier) @definition.function
-        (#set! "hoist" "function")
-        (#set! "scope" "global")))
+;; This syntax is only allowed to define top-level functions,
+;; which we consider as non-locals, so we don't want to track these
+;; here.
+;;
+;; (function_declaration
+;;  name: ((identifier) @definition.function)
+;;  (#set! "hoist" "function"))
```
varungandhi-src, line 35: There seems to be some similarity between `#set! "scope" "global"` and `@definition.skip` in that both are saying "this is a global entity, don't treat it as a local". So I'm not super clear on whether we need both... Do we?

I wonder if the `#set! "scope"` functionality can be removed entirely? With this removal, only the Java grammar will be using `#set! "scope" "parent"`. 🤔

docker-images/syntax-highlighter/crates/scip-syntax/src/snapshots/scip_syntax__locals__test__can_do_go.snap
```diff
@@ -4,26 +4,41 @@ expression: dumped
...
+    fmt.Println(local, "is negative")
+//              ^^^^^ reference local 5
+   } else if local < 10 {
+//           ^^^^^ reference local 5
+    fmt.Println(local, "has 1 digit")
+//              ^^^^^ reference local 5
+   } else {
+    fmt.Println(local, "has multiple digits")
+//              ^^^^^ reference local 5
+   }
```
varungandhi-src, line 40: I think we're missing test cases for `switch` and `for`? They have similar affordances. E.g. this is allowed:

```
	switch s := 0; s + 1 {
	case s + 2:
		_ = s
	default:
		_ = s
	}
    // s cannot be used here
```

`for` also allows defining local variables, e.g. `for i := 0; i < 3; i++`.

Please add them if they're missing.

varungandhi-src (changes requested):
Directionally, this looks correct, but please add more test cases.

docker-images/syntax-highlighter/crates/scip-syntax/docs/locals-query-dsl.md
```diff
@@ -96,6 +96,30 @@ def f():
...
+
+```go
+// Will be skipped
+var top_level = 10
+
+func main() {
+    // Will be recorded as a local
+    var local = 10
+}
+```
```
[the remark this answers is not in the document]
kritzcreek: There's no decision from my side yet. I've decided to go with this approach for now because it was easy to implement, which didn't seem worthy of writing down.
My plan was to go with it for a bit and see if it satisfies the immediate cases we know about.

docker-images/syntax-highlighter/crates/scip-syntax/queries/go/scip-locals.scm
```diff
@@ -2,23 +2,37 @@
...
+;;     ..
+;; }
 (if_statement) @scope
 (for_statement) @scope
 (block) @scope
 
+;; We want to skip global definitions as they're non-local.
+(source_file
+ (var_declaration
+  (var_spec name: (identifier) @definition.skip)))
```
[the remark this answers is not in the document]
kritzcreek: I don't think anything in here handles const definitions yet. I can look into adding those.

docker-images/syntax-highlighter/crates/scip-syntax/queries/go/scip-locals.scm
kritzcreek: There is no `#set! "scope"` functionality in the old or the new implementation. I'm not sure why this was here, I just didn't remove it.
If we're sure `scip-locals.scm` is exclusive to the locals.rs implementation I can get rid of these attributes.

docker-images/syntax-highlighter/crates/scip-syntax/src/snapshots/scip_syntax__locals__test__can_do_functions-2.snap
```diff
@@ -3,8 +3,6 @@ source: crates/scip-syntax/src/locals.rs
 expression: scope_tree
 ---
 scope global (0, 0)-(27, 0)
-  hoisted_def Something (7, 5)-(7, 14)
```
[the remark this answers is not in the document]
kritzcreek: I gave this a try before, but for some reason insta made this difficult. I'll try again and report back.

docker-images/syntax-highlighter/crates/scip-syntax/src/snapshots/scip_syntax__locals__test__can_do_go.snap
kritzcreek: Yeah I thought this was a bit silly as well. One thing that's a bit of a shame when merging all of them, is that a single line addition in the test file causes the entire snapshot to light up in the diff, because all line numbers change.

I'll merge the test cases for now, and we can decide if we want a multi-level structure at some point (e.g. `testdata/go/if.go`)

docker-images/syntax-highlighter/crates/scip-syntax/src/snapshots/scip_syntax__locals__test__can_do_functions-2.snap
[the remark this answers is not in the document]
kritzcreek: Okay, I remember now. Because it's macro nonsense we need to repeat the test names multiple times, but it's probably still better than bad output names. I added a commit for explicitly naming these tests.
Once I've merged the different go tests into one file, we can think about the naming scheme.

docker-images/syntax-highlighter/crates/scip-syntax/src/snapshots/scip_syntax__locals__test__go_occurrences.snap
```diff
@@ -0,0 +1,118 @@
...
+   "fmt"
+  )
+  
+  // Not actually local
+  var local = 10
+  
+  func main() {
+   local = 20
+   local := 5
+// ^^^^^ definition local 1
```
varungandhi-src, line 17: We've covered the different definition forms, but I just noticed that the various reference forms are not covered. Maybe it's worthwhile to make sure we're emitting references for them as expected?

- Assignment `a = 3` (we have plenty of `:=` but only assignments to globals from a quick glance)
- Slice/map assignment `x[y] = 3`
- Assignment after dereference: `*x = 3`

There's also the range syntax `x[a:b]` but I don't think that's allowed as an lvalue.

varungandhi-src (approved):
There is one minor comment where I thought of some more tests which I didn't realize were missing before, but LGTM otherwise.

PR #59321, opened 2024-01-04
Add rate limit information to RateLimited events

rafax:
Add information about limit that was hit and retry_after period.

## Test plan

- tested locally

### DIFF
--- a/cmd/cody-gateway/internal/events/events.go
+++ b/cmd/cody-gateway/internal/events/events.go
@@ -155,3 +155,12 @@ func (l *stdoutLogger) LogEvent(spanCtx context.Context, event Event) error {
 	)
 	return nil
 }
+
+func MergeMaps(dst map[string]any, srcs ...map[string]any) map[string]any {
+	for _, src := range srcs {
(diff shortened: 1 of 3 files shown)


cmd/cody-gateway/internal/events/events.go
```diff
@@ -155,3 +155,12 @@ func (l *stdoutLogger) LogEvent(spanCtx context.Context, event Event) error {
 	)
 	return nil
 }
+
+func MergeMaps(dst map[string]any, srcs ...map[string]any) map[string]any {
```
unknwon, line 159: nit: add docstring states the `dst` is mutated in-place

unknwon (approved):

cmd/cody-gateway/internal/events/events.go
rafax: Added here https://github.com/sourcegraph/sourcegraph/pull/59322/files#diff-2a10b7c9b7e1aa08fa780f1a0ceb468509418746c6fe712a5292dd24753307d1

PR #59325, opened 2024-01-04
svelte: Fix `svelte-check`, lint and TS errors

fkling:
This PR addresses all issues reported by `svelte-check` related to the packages own files. `svelte-check` reports many more errors but they all related to files in other packages (e.g. `client/web`). I haven't figured out a way yet to have `svelte-check`/TypeScript ignore these.

What was done:

- Remove React interop code (wasn't used)
- Move all stories to Svelte component format and co-located them with the tested component.
- Fixed TypeScript types
- Fixed eslint issues
- Migrated some GraphQL queries to `.gql` files (I _think_ those had some errors/warnings associated with them; I'll migrate more in an upcoming commit)

What hasn't worked:

- Upgrading `svelte-check`. I mean, it works but then it reports issues with extending `$$Props` in some Svelte components, which is strange because that's the way to do it. Idk if it's related to our TS setup.
- Migrating to `eslint-plugin-svelte`. This reports many more errors, including not being able to resolve aliased paths like `$lib/...`. This should be done eventually but requires more time to get the config right.
- Making use of TypeScript project references. I probably didn't get the configuration right. I wanted to speed up `pnpm check` but it resulted in more errors about not being able to find types.

What's still problematic:

- eslint produces a bunch of warnings regarding accessing properties on `any` types. It looks like the eslint setup doesn't take the types generated from SvelteKit into account. Will have to investigate this.

## Test plan

`cd client/web-sveltekit && pnpm check` only produces errors for files outside the package.


camdencheek (approved):
Nice!!! I was just stating to look at tackling some of these. My editor has been very unhappy 😄

PR #59330, opened 2024-01-04
Reset Cody Gateway usage cache automatically during limits sync on interval & TTL mismatch

thenamankumar:
- Closes https://github.com/sourcegraph/sourcegraph/issues/59418
- Closes https://github.com/sourcegraph/cody/issues/2475

## tldr;

1. The value of Cody usage counter is not reset to 0 when the user changes their plan. 
2. The TTL of Cody usage counter if not updated from 30 days to 1 day when the user upgrades from Free to Pro, or the other way around. 

## Slack
https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1704375768295309?thread_ts=1704370864.536539&cid=C04MSD3DP5L

## Background
Cody Gateway maintains 2 caches:
1. limits config with
    -  24hrs TTL (controlled by `CODY_GATEWAY_SOURCES_CACHE_TTL`). 
    - But is is considered stale in the code if it is older than 15 minutes and thus we re-fetch every 15 mins. 
2. usage counter with 
    - 30 days TTL for free users (monthly rate limits)
    - [SHOULD BE, BUT NEVER UPDATED] 1-day TTL for pro users (daily rate limits)

Right now, when the user upgrades, we request Cody Gateway to refresh the limits config in its cache. But nothing is implemented to reset the usage counter or change its TLL, which is the BUG. 

## Solution

When Cody Gateway fetches limit config from dotcom source, whether on
1. cache miss
2. stale cache older than 15 mins
3. trigger by refresh config API request

it checks if  `newConfig.Interval != oldConfig.interval` and deletes the usage counter cache.

We will also update `CODY_GATEWAY_SOURCES_CACHE_TTL` in the config from 24hours to 31 or 62 days so that the limits config is persisted in the cache for longer and covers the complete billing cycle. 

### FIRST
There is also a logic in the Limiter which checks the following:
`if usage_cache.TTL > rate_limit.Interval` then update the cache's TTL = interval. 

The above logic is only a safeguard from corrupt usage cache or handle the edge cases during any kind of code change and migrations. 

This logic only worked on successful requests and not when the user is already hitting the rate limits. Which defeats the purpose of having it in the first place. 

As part of this PR, I have moved this check, so it is performed even when the request it already hitting rate limits.

### SECOND (TODO, will be done as part of a separate PR)
We will block users to downgrade on the same day they upgrade to prevent abuse until the payments are integrated. 

## Test plan
1. Follow this guide to setup Gateway Locally https://docs.sourcegraph.com/dev/how-to/cody_gateway
2. Additionally, add `CODY_GATEWAY_SOURCES_CACHE_TTL: 720h` (30 days) to `sg.config.overwrite.yaml` under `commands.cody-gateway.env`. This will store the actor config in the cache for 30 days. CURRENTLY 24h.
3. Go to `cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go:40` and edit `defaultRefreshInterval` = 0. 
4. Run `sg start dotcom`
5. Make sure `cody-pro` feature flag is turned on
6. Connect a redis client to localhost:6379
7. Visit `/cody/manage` and verify that you are on Cody Free plan, if not, downgrade.
8. Visit Cody Web at `/cody/chat` and make a chat requests
9. Check for `v2:dotcom-users:v2:<user_id>` key in Redis: it stores the limits config in JSON.
      - The key should have a TTL in the neighbourhood of 25,92,000 seconds. (30 days as we updated the `CODY_GATEWAY_SOURCES_CACHE_TTL`).
      - Both `rateLimits.chat_completions.interval` and `rateLimits.code_completions.interval` should be exactly `2592000000000000` (30 days as the user is on the Free tier).
10. Check for`rate_limit:chat_completions:<user_id>` key in Redis: it stores the LLM requests usage counter.
      - The key should have a TTL in the neighbourhood of 25,92,000 seconds. (30 days as the user is on the Free tier).
      - The value should increment with each new chat message. 
11. Visit `/cody/manage` and upgrade to Cody Pro.
12.  Check for `v2:dotcom-users:v2:<user_id>` key in Redis:
      - Both `rateLimits.chat_completions.interval` and `rateLimits.code_completions.interval` should now be `86400000000000` (1 day as the user is on the Pro tier).
      - The TTL should still be in the neighourhood of 30 days. As controlled by the config.
13.  Check for`rate_limit:chat_completions:<user_id>` key in Redis.
      - The key be DELETED.
14. Visit `/cody/chat` and make a request.
15. Check for`rate_limit:chat_completions:<user_id>` key in Redis.
      - The key should have a TTL in the neighbourhood of 86,400 seconds. (1 day as the user is on the Pro tier).
      - The value should reset and will be 1 or 2 based on the number of chats you made after upgrading. 
16. Repeat the steps from 11-14 after downgrading back to Free. Accordingly check for the TLL and config numbers.

### DIFF
--- a/cmd/cody-gateway/internal/actor/actor.go
+++ b/cmd/cody-gateway/internal/actor/actor.go
@@ -148,37 +148,37 @@ func WithActor(ctx context.Context, a *Actor) context.Context {
 	return context.WithValue(ctx, actorKey, a)
 }
 
-func (a *Actor) Limiter(
-	logger log.Logger,
+// featurePrefix is the prefix used by redis store for the given
+// feature because we need to rate limit by feature.
+func featurePrefix(feature codygateway.Feature) string {
+	return fmt.Sprintf("%s:", feature)
+}
+
+// baseLimiterAndLimit returns the base limiter for the given feature,
+// and the rate limits for the actor.
+func (a *Actor) baseLimiterAndLimit(
 	redis limiter.RedisStore,
 	feature codygateway.Feature,
 	rateLimitNotifier notify.RateLimitNotifier,
-) (limiter.Limiter, bool) {
+) (*limiter.StaticLimiter, RateLimit, bool) {
 	if a == nil {
 		// Not logged in, no limit applicable.
-		return nil, false
+		return nil, RateLimit{}, false
 	}
+
 	limit, ok := a.RateLimits[feature]
 	if !ok {
-		return nil, false
+		return nil, RateLimit{}, false
 	}
-
 	if !limit.IsValid() {
 		// No valid limit, cannot provide limiter.
-		return nil, false
+		return nil, RateLimit{}, false
 	}
 
-	// The redis store has to use a prefix for the given feature because we need to
-	// rate limit by feature.
-	featurePrefix := fmt.Sprintf("%s:", feature)
-
-	// baseLimiter is the core Limiter that naively applies the specified
-	// rate limits. This will get wrapped in various other layers of limiter
-	// behaviour.
 	baseLimiter := limiter.StaticLimiter{
 		LimiterName: "actor.Limiter",
 		Identifier:  a.ID,
-		Redis:       limiter.NewPrefixRedisStore(featurePrefix, redis),
+		Redis:       limiter.NewPrefixRedisStore(featurePrefix(feature), redis),
 		Limit:       limit.Limit,
 		Interval:    limit.Interval,
 		// Only update rate limit TTL if the actor has been updated recently.
@@ -189,18 +189,46 @@ func (a *Actor) Limiter(
 		},
 	}
 
+	return &baseLimiter, limit, true
+}
+
+// BaseLimiter is the core Limiter that naively applies the specified
+// rate limits. This will get wrapped in various other layers of limiter
+// behaviour. Do not use this directly to apply rate limits, use Limiter instead.
+func (a *Actor) BaseLimiter(
+	redis limiter.RedisStore,
+	feature codygateway.Feature,
+	rateLimitNotifier notify.RateLimitNotifier,
+) (*limiter.StaticLimiter, bool) {
+	baseLimiter, _, ok := a.baseLimiterAndLimit(redis, feature, rateLimitNotifier)
+
+	return baseLimiter, ok
+}
+
+// Limiter is the main limiter used to apply rate limits to requests. It
+// applies the base rate limit, and wraps it in layers of concurrency
+// and update-on-error rate limiters.
+func (a *Actor) Limiter(
+	logger log.Logger,
+	redis limiter.RedisStore,
+	feature codygateway.Feature,
+	rateLimitNotifier notify.RateLimitNotifier,
+) (limiter.Limiter, bool) {
+	baseLimiter, limit, ok := a.baseLimiterAndLimit(redis, feature, rateLimitNotifier)
+	if !ok {
+		return nil, false
+	}
+
 	return &concurrencyLimiter{
 		logger:             logger.Scoped("concurrency"),
 		actor:              a,
 		feature:            feature,
-		redis:              limiter.NewPrefixRedisStore(fmt.Sprintf("concurrent:%s", featurePrefix), redis),
+		redis:              limiter.NewPrefixRedisStore(fmt.Sprintf("concurrent:%s", featurePrefix(feature)), redis),
 		concurrentRequests: limit.ConcurrentRequests,
 		concurrentInterval: limit.ConcurrentRequestsInterval,
-
 		nextLimiter: updateOnErrorLimiter{
-			logger: logger.Scoped("updateOnError"),
-			actor:  a,
-
+			logger:      logger.Scoped("updateOnError"),
+			actor:       a,
 			nextLimiter: baseLimiter,
 		},
 		nowFunc: time.Now,
--- a/cmd/cody-gateway/internal/actor/actor.go
+++ b/cmd/cody-gateway/internal/actor/actor.go
@@ -62,6 +62,14 @@ func (a *Actor) GetSource() codygateway.ActorSource {
 	return codygateway.ActorSource(a.Source.Name())
 }
 
+// IsEmpty returns true if the actor is empty, i.e. has no ID.
+// An empty actor is saved in the cache on fetch req failure,
+// so that we aren't constantly hitting the dotcom API. Check
+// the implementation of `Source.fetchAndCache`.
+func (a *Actor) IsEmpty() bool {
+	return a == nil || a.ID == ""
+}
+
 func (a *Actor) IsDotComActor() bool {
 	// Corresponds to sourcegraph.com subscription ID, or using a dotcom access token
 	return a != nil && (a.GetSource() == codygateway.ActorSourceProductSubscription && a.ID == "d3d2b638-d0a2-4539-a099-b36860b09819") || a.GetSource() == codygateway.ActorSourceDotcomUser
--- a/cmd/cody-gateway/internal/actor/actor.go
+++ b/cmd/cody-gateway/internal/actor/actor.go
@@ -62,7 +62,7 @@ func (a *Actor) GetSource() codygateway.ActorSource {
 	return codygateway.ActorSource(a.Source.Name())
 }
 
-// IsEmpty returns true if the actor is empty, i.e. has no ID.
+// IsEmpty returns true if the actor is empty, e.g. has no ID.
 // An empty actor is saved in the cache on fetch req failure,
 // so that we aren't constantly hitting the dotcom API. Check
 // the implementation of `Source.fetchAndCache`.
@@ -222,24 +222,37 @@ func (a *Actor) Limiter(
 	feature codygateway.Feature,
 	rateLimitNotifier notify.RateLimitNotifier,
 ) (limiter.Limiter, bool) {
+	// Start with the base limiter. This limiter enforces rate limits provided
+	// the actor's Source, and each Source implementation owns ensuring that
+	// Cody Gateway has an up-to-date view of the appropriate rate limits for
+	// a particular actor based on the application (e.g. users from enterprise product
+	// subscriptions, Self-Serve-Cody's tiers, etc.
 	baseLimiter, limit, ok := a.baseLimiterAndLimit(redis, feature, rateLimitNotifier)
 	if !ok {
 		return nil, false
 	}
 
+	// Wrap the base limiter with updateOnErrorLimiter, to call Actor.Update to
+	// refresh the actor cache on any errors.
+	updateOnErrLimiter := updateOnErrorLimiter{
+		logger:      logger.Scoped("updateOnError"),
+		actor:       a,
+		nextLimiter: baseLimiter,
+	}
+
+	// Finally return a concurrency limiter, to ensure that a user cannot have too many
+	// requests in-flight at a time. This is generally a percentage of the rate limit
+	// assigned to an Actor by its Source - see RateLimit for more details.
+	concurrentStorePrefix := fmt.Sprintf("concurrent:%s", featurePrefix(feature))
 	return &concurrencyLimiter{
 		logger:             logger.Scoped("concurrency"),
 		actor:              a,
 		feature:            feature,
-		redis:              limiter.NewPrefixRedisStore(fmt.Sprintf("concurrent:%s", featurePrefix(feature)), redis),
+		redis:              limiter.NewPrefixRedisStore(concurrentStorePrefix, redis),
 		concurrentRequests: limit.ConcurrentRequests,
 		concurrentInterval: limit.ConcurrentRequestsInterval,
-		nextLimiter: updateOnErrorLimiter{
-			logger:      logger.Scoped("updateOnError"),
-			actor:       a,
-			nextLimiter: baseLimiter,
-		},
-		nowFunc: time.Now,
+		nextLimiter:        updateOnErrLimiter,
+		nowFunc:            time.Now,
 	}, true
 }
 
--- a/cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
+++ b/cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
@@ -3,6 +3,7 @@ package dotcomuser
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"strconv"
 	"strings"
 	"time"
@@ -19,6 +20,8 @@ import (
 
 	"github.com/sourcegraph/sourcegraph/cmd/cody-gateway/internal/actor"
 	"github.com/sourcegraph/sourcegraph/cmd/cody-gateway/internal/dotcom"
+	"github.com/sourcegraph/sourcegraph/cmd/cody-gateway/internal/limiter"
+	"github.com/sourcegraph/sourcegraph/cmd/cody-gateway/internal/notify"
 	"github.com/sourcegraph/sourcegraph/internal/codygateway"
 	"github.com/sourcegraph/sourcegraph/internal/trace"
 	"github.com/sourcegraph/sourcegraph/lib/errors"
@@ -37,7 +40,7 @@ var (
 	// defaultRefreshInterval is used for updates, which is also called when a
 	// user's rate limit is hit, so we don't want to update every time. We use
 	// a shorter interval than the default in this case.
-	defaultRefreshInterval = 5 * time.Minute
+	defaultRefreshInterval = 0 * time.Minute
 )
 
 type Source struct {
@@ -45,16 +48,20 @@ type Source struct {
 	cache             httpcache.Cache // cache is expected to be something with automatic TTL
 	dotcom            graphql.Client
 	concurrencyConfig codygateway.ActorConcurrencyLimitConfig
+	rs                limiter.RedisStore
+	rateLimitNotifier notify.RateLimitNotifier
 }
 
 var _ actor.SourceUpdater = &Source{}
 
-func NewSource(logger log.Logger, cache httpcache.Cache, dotComClient graphql.Client, concurrencyConfig codygateway.ActorConcurrencyLimitConfig) *Source {
+func NewSource(logger log.Logger, cache httpcache.Cache, dotComClient graphql.Client, concurrencyConfig codygateway.ActorConcurrencyLimitConfig, rs limiter.RedisStore, rateLimitNotifier notify.RateLimitNotifier) *Source {
 	return &Source{
 		log:               logger.Scoped("dotcomuser"),
 		cache:             cache,
 		dotcom:            dotComClient,
 		concurrencyConfig: concurrencyConfig,
+		rs:                rs,
+		rateLimitNotifier: rateLimitNotifier,
 	}
 }
 
@@ -76,11 +83,16 @@ func (s *Source) Update(ctx context.Context, act *actor.Actor) error {
 }
 
 // fetchAndCache fetches the dotcom user data for the given user token and caches it
-func (s *Source) fetchAndCache(ctx context.Context, token string) (*actor.Actor, error) {
+func (s *Source) fetchAndCache(ctx context.Context, token string, oldAct *actor.Actor) (*actor.Actor, error) {
 	var act *actor.Actor
 	resp, checkErr := s.checkAccessToken(ctx, token)
 	if checkErr != nil {
-		// Generate a stateless actor so that we aren't constantly hitting the dotcom API
+		// if oldAct is present in the cache, even though it is stale,
+		// return it as a fallback if we fail to hit the dotcom API.
+		if oldAct != nil && oldAct.ID != "" {
+			return oldAct, nil
+		}
+		// Generate a stateless actor and save in the cache so that we aren't constantly hitting the dotcom API
 		act = newActor(s, token, dotcom.DotcomUserState{}, s.concurrencyConfig)
 	} else {
 		act = newActor(s, token,
@@ -91,6 +103,27 @@ func (s *Source) fetchAndCache(ctx context.Context, token string) (*actor.Actor,
 		s.log.Error("failed to marshal actor",
 			log.Error(err))
 	} else {
+		for _, feature := range codygateway.AllFeatures {
+			rl, ok := act.RateLimits[feature]
+			if !ok {
+				continue
+			}
+
+			// reset the usage cache if:
+			// 1. old actor is not present in the cache
+			// 2. new rl interval is different from old rl interval
+			if oldAct == nil || oldAct.ID == "" || rl.Interval != oldAct.RateLimits[feature].Interval {
+				// get the base limiter for the feature which implements the core rate limiting based on the config
+				l, ok := act.BaseLimiter(s.rs, feature, s.rateLimitNotifier)
+				if !ok {
+					return nil, errors.Wrap(err, fmt.Sprintf("failed to create base limiter for feature: %s and actor id: %s", feature, l.Identifier))
+				}
+				if err := l.ResetUsage(); err != nil {
+					return nil, errors.Wrap(err, fmt.Sprintf("failed to reset usage cache for feature: %s and actor id: %s", feature, l.Identifier))
+				}
+			}
+		}
+
 		s.cache.Set(token, data)
 		// Also save to the key based on the actor ID.
 		s.cache.Set(act.ID, data)
@@ -133,13 +166,11 @@ func (s *Source) get(ctx context.Context, token string, bypassCache bool) (*acto
 	if len(token) != tokenLength {
 		return nil, errors.New("invalid token format")
 	}
-	// force fetch of rate-limit data from upstream
-	if bypassCache {
-		return s.fetchAndCache(ctx, token)
-	}
+	// NOTE(naman): figure out a way to cache the actor in the limiter
+	// based on actor id and not the token to avoid storing the config for each token.
 	data, hit := s.cache.Get(token)
 	if !hit {
-		return s.fetchAndCache(ctx, token)
+		return s.fetchAndCache(ctx, token, nil)
 	}
 
 	var act *actor.Actor
@@ -149,11 +180,12 @@ func (s *Source) get(ctx context.Context, token string, bypassCache bool) (*acto
 		// Delete the corrupted record.
 		s.cache.Delete(token)
 
-		return s.fetchAndCache(ctx, token)
+		return s.fetchAndCache(ctx, token, nil)
 	}
 
-	if act.LastUpdated == nil || time.Since(*act.LastUpdated) > defaultUpdateInterval {
-		return s.fetchAndCache(ctx, token)
+	// if force fetch of rate-limit data from upstream or act cache older than defaultUpdateInterval (15 minutes)
+	if bypassCache || act.LastUpdated == nil || time.Since(*act.LastUpdated) > defaultUpdateInterval {
+		return s.fetchAndCache(ctx, token, act)
 	}
 
 	act.Source = s
--- a/cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
+++ b/cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
@@ -82,41 +82,28 @@ func (s *Source) Update(ctx context.Context, act *actor.Actor) error {
 	return err
 }
 
-// fetchAndCache fetches the dotcom user data for the given user token and caches it
+// fetchAndCache fetches the dotcom user data for the given user token and caches it.
+// We compare against the previous state of the actor (if available) to trigger logic based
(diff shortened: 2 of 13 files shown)


cmd/cody-gateway/internal/actor/actor.go
```diff
@@ -189,19 +195,47 @@ func (a *Actor) Limiter(
 		},
 	}
 
+	return baseLimiterConfig{limit: limit, featurePrefix: featurePrefix, baseLimiter: baseLimiter}, true
+}
+
+func (a *Actor) BaseLimiter(
```
chrsmith, line 201: (Sorry if you were planning to address this later, as the PR is still in draft.)

Would you mind adding a doc comment here on `BaseLimiter` and below on `Limiter`? It is unclear to me what the difference is, or when I should call one method over the other.

I suspect the `StaticLimiter` is used in some situations, but the other `Limiter` should be used more generally? But how/why? (The concern here is just making sure that callers know how to use these methods correctly and don't inadvertently call the wrong one.)

cmd/cody-gateway/internal/actor/actor.go
```diff
@@ -148,24 +148,30 @@ func WithActor(ctx context.Context, a *Actor) context.Context {
 	return context.WithValue(ctx, actorKey, a)
 }
 
-func (a *Actor) Limiter(
+type baseLimiterConfig struct {
+	limit         RateLimit
+	featurePrefix string
```
chrsmith, line 153: What is this used for? Without knowing more about the surrounding code, it seems odd that there is a `feature codygateway.Feature` parameter passed to `getBaseLimiterConfig(...)` and then the returned object includes that as `"%s:"` for the `featurePrefix`.

Would it be net cleaner to just omit the `featurePrefix` field from this struct, and require the caller to generate `featurePrefix` as-needed? It seems like it's redundant, since the caller essentially already knows what the feature prefix is. (The `feature` it passed, with a colon after it.)

I don't feel strongly about it, so feel free to keep as-is and/or explain why this is a bad suggestion :)

cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
```diff
@@ -91,6 +98,20 @@ func (s *Source) fetchAndCache(ctx context.Context, token string) (*actor.Actor,
...
 	} else {
+		// TODO(naman) check if usage cache TTL > now + interval, if yes, reset usage cache.
+
+		for _, feature := range codygateway.AllFeatures {
+			if rl, ok := act.RateLimits[feature]; ok {
+				if l, ok := act.BaseLimiter(s.log, s.rs, feature, s.rateLimitNotifier); ok {
+
+					if usageTTL, err := l.Redis.TTL(l.Identifier); err != nil && usageTTL > int(rl.Interval.Seconds()) {
+						// reset usage cache
+						err := l.Redis.Del(l.Identifier)
```
chrsmith, line 109: Perhaps this just needs to be called out in the comment above for clarity, but just so I understand:

Here we are purging the Redis _usage_ cache. (i.e. the number of LLM requests or completions a user has made.)

The reason we need to add this code here is that currently, we rely on the Redis TTL to _automatically_ delete the usage data. However, in the case where the user's interval has changed (_AFTER_ the Redis TTL has been set), we are in for a bad time.

So we proactively delete the usage data here, so we can start fresh.

Right?

But if that is the case, does it make sense to not use the Redis TTL at all? And instead opportunistically purge the Redis Usage cache whenever we notice that the usage data is too stale? (Or is it we don't persist that anywhere, and the source of truth for when we need to reset a given (user x feature) usage data is the Redis TTL value for that key?

chrsmith (commented):
Mostly questions to better understand things.

thenamankumar:
@chrsmith Thanks for your comments. It was still WIP when you reviewed it earlier. I have addressed all your concerns. After getting this draft validated, I'll progress with adding tests.

cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
```diff
@@ -91,6 +98,26 @@ func (s *Source) fetchAndCache(ctx context.Context, token string) (*actor.Actor,
 		s.log.Error("failed to marshal actor",
 			log.Error(err))
 	} else {
+		for _, feature := range codygateway.AllFeatures {
+			if rl, ok := act.RateLimits[feature]; ok {
+				if l, ok := act.BaseLimiter(s.rs, feature, s.rateLimitNotifier); ok {
+					usageTTL, err := l.Redis.TTL(l.Identifier)
+
+					if err != nil {
```
rafax, line 106: Would it be easier / cleaner to push the TTL updating logic to the Limiter class (that manages Identifier)?

cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
```diff
@@ -91,6 +98,26 @@ func (s *Source) fetchAndCache(ctx context.Context, token string) (*actor.Actor,
...
+						return nil, errors.Wrap(err, fmt.Sprintf("failed to get usage cache TTL for feature: %s and id: %s", feature, l.Identifier))
+					}
+
+					// If the usage cache TTL is greater than the rate limit interval, then we need to reset the usage cache.
+					// It is used to handle the case when free users with upto 30 days usage counter TTL upgrade to a paid plan with 1 day rate limit interval.
+					if usageTTL > int(rl.Interval.Seconds()) {
```
rafax, line 112: Note: we already have some [logic](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/cmd/cody-gateway/internal/limiter/limiter.go?L166) that handles this case differently (if usage TTL is longer than interval, it updates usage TTL to the interval instead of deleting usage altogether) - I think we should make sure that we either have one implementation of this, or if we have two, that they are in sync.

thenamankumar:
NOTES: meeting with @rafax 

- TLL should only change when the limit changes. 
- The logic should belong to the source and not inside the limiter. 
- We need to remove the logic from the limiter and the last 5 mins logic. 
- Should we update TTL or reset the counter? 
  - Delete rather than update TTL
-  What about downgrade?
  - They might be blocked for rest of the day if they have already consumed the free limits that day before downgrade. 

-------

NEW IDEA
- store the actor in the cache for 30 days
- on fetchAndCache every 15 mins, check if the new intervalSeconds is different from the old one is cache
- if yes, then delete the usage.

Additional improvements
- do not store config for each token separately
- may be have a token <> user_id mapping
- may be SAMS token will be also the embed the user_id in it.

cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
thenamankumar: I have added a `StaticLimiter.ResetUsage` func so that we don't need to work with the internals here.

cmd/cody-gateway/internal/actor/actor.go
```diff
@@ -189,18 +189,46 @@ func (a *Actor) Limiter(
 		},
 	}
 
+	return &baseLimiter, limit, true
+}
+
+// BaseLimiter is the core Limiter that naively applies the specified
+// rate limits. This will get wrapped in various other layers of limiter
+// behaviour. Do not use this directly to apply rate limits, use Limiter instead.
+func (a *Actor) BaseLimiter(
```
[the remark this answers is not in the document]
thenamankumar: We can not, we need to reset the usage during the sync and not on every request. So the logic must be present in side the source update code.

cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
thenamankumar: After another look, I found that we still need that logic in place because that is how the limiter works. With the updated code in the PR, we now only reset the usage if the interval changes.

cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
```diff
@@ -149,11 +180,12 @@ func (s *Source) get(ctx context.Context, token string, bypassCache bool) (*acto
...
 		s.cache.Delete(token)
 
-		return s.fetchAndCache(ctx, token)
+		return s.fetchAndCache(ctx, token, nil)
 	}
 
-	if act.LastUpdated == nil || time.Since(*act.LastUpdated) > defaultUpdateInterval {
-		return s.fetchAndCache(ctx, token)
+	// if force fetch of rate-limit data from upstream or act cache older than defaultUpdateInterval (15 minutes)
+	if bypassCache || act.LastUpdated == nil || time.Since(*act.LastUpdated) > defaultUpdateInterval {
```
rafax, line 187: Moving `bypassCache` here will mean that we will fail to refresh the cache if the current record can't be parsed (defeating the purpose of bypassCache)

cmd/cody-gateway/internal/httpapi/featurelimiter/featurelimiter.go
```diff
@@ -223,7 +223,9 @@ func RefreshLimitsHandler(baseLogger log.Logger) http.Handler {
 		if err := act.Update(r.Context()); err != nil {
 			logger := act.Logger(trace.Logger(r.Context(), baseLogger))
 			if actor.IsErrActorRecentlyUpdated(err) {
-				response.JSONError(logger, w, http.StatusTooManyRequests, err)
+				// If the actor was recently updated, we can just return a 200.
+				// Client will keep on retrying otherwise.
+				w.WriteHeader(http.StatusOK)
```
rafax, line 228: I disagree with this change - returning 429 is a clear signal that:

- the action requested by the client has not been performed
- client should not retry this error

Returning 200 instead makes it look like the refresh succeeded (where we didn't really refresh).

cmd/cody-gateway/internal/limiter/limiter.go
```diff
@@ -86,6 +86,18 @@ func (l StaticLimiter) TryAcquire(ctx context.Context) (_ func(context.Context,
 		return nil, NoAccessError{}
 	}
 
+	// If the expiry on the key is greater than the intervalSeconds, update the TTL.
+	// This is here as a safeguard for the case where the TTL is wrongly set or the
+	// usage cache is not reset when the intervalSeconds config change. This only covers
+	// the case where the new intervalSeconds in shorter than the present TTL, i.e Free -> Pro.
```
rafax, line 92:
```suggestion
	// the case where the new intervalSeconds in shorter than the present TTL
```
Free / Pro are not items that this code deals with

cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
```diff
@@ -91,6 +103,27 @@ func (s *Source) fetchAndCache(ctx context.Context, token string) (*actor.Actor,
...
+		for _, feature := range codygateway.AllFeatures {
+			rl, ok := act.RateLimits[feature]
+			if !ok {
+				continue
+			}
+
+			// reset the usage cache if:
+			// 1. old actor is not present in the cache
+			// 2. new rl interval is different from old rl interval
+			if oldAct == nil || oldAct.ID == "" || rl.Interval != oldAct.RateLimits[feature].Interval {
```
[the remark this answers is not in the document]
rafax: Ex. see https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@dc0d555bf15b8957f4af3f35058e63e6255698b3/-/blob/cmd/cody-gateway/internal/actor/actor.go?L65

cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
```diff
@@ -37,24 +40,28 @@ var (
...
-	defaultRefreshInterval = 5 * time.Minute
+	defaultRefreshInterval = 0 * time.Minute
 )
 
 type Source struct {
 	log               log.Logger
 	cache             httpcache.Cache // cache is expected to be something with automatic TTL
 	dotcom            graphql.Client
 	concurrencyConfig codygateway.ActorConcurrencyLimitConfig
+	rs                limiter.RedisStore
```
rafax, line 51:
```suggestion
	redisStore                limiter.RedisStore
```

cmd/cody-gateway/internal/limiter/limiter.go
```diff
@@ -86,6 +86,18 @@ func (l StaticLimiter) TryAcquire(ctx context.Context) (_ func(context.Context,
...
+	if ttl, err := l.Redis.TTL(l.Identifier); err == nil {
```
rafax, line 93: This is adding 1-2 Redis round-trips to every rate-limit check, so (currently) 2-4 Redis round-trips (X0ms) to every Cody request - is this extra check really necessary?

cmd/cody-gateway/internal/actor/actor.go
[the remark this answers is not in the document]
rafax: Understood, but we could make the syncing code do:
```
act.Limiter(...).ResetUsage(feature)
``` 
and we'd avoid exposing two very similar methods (BaseLimiter and Limiter) and maybe could hide some of the parameters we pass around (I initially had the same question as @chrsmith above).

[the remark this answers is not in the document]
rafax: I'll try to propose a more specific fix tomorrow.

cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
```diff
@@ -91,6 +98,27 @@ func (s *Source) fetchAndCache(ctx context.Context, token string) (*actor.Actor,
...
+		for _, feature := range codygateway.AllFeatures {
+			rl, ok := act.RateLimits[feature]
+			if !ok {
+				continue
+			}
+
+			// reset the usage cache if:
+			// 1. old actor is not present in the cache
+			// 2. new rl interval is different from old rl interval
+			if oldAct.IsEmpty() || rl.Interval != oldAct.RateLimits[feature].Interval {
```
chrsmith, line 110: Would it ever be the case that the old actor is missing some new `feature` the the new actor has? e.g. suppose we introduce a new Cody feature. Will `oldAct` be loaded from a permanent store (and likely be serialized missing the new feature), or will it _always_ come from memory? And so we can assume it was initialized with the same set of features?

cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
```diff
@@ -91,6 +98,27 @@ func (s *Source) fetchAndCache(ctx context.Context, token string) (*actor.Actor,
...
+				// get the base limiter for the feature which implements the core rate limiting based on the config
```
chrsmith, line 111: Can we be more specific than "core rate limiting based on the config"? I mean, it's clear what that is referring too.. I just cannot say off the top of my head what it _means_. e.g. what types of things are we limiting based on the config, and what we would call these "other" types of rate limiters (dependent on user subscriptions, etc.)

Would it be helpful to try and introduce new terms here? Like "anti-abuse limiting" or "subscription-usage limiting" or something?

```suggestion
				// Get the base limiter for the feature which implements the core rate limiting based on the config.
```

cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
```diff
@@ -91,6 +98,27 @@ func (s *Source) fetchAndCache(ctx context.Context, token string) (*actor.Actor,
...
+				l, ok := act.BaseLimiter(s.redisStore, feature, s.rateLimitNotifier)
+				if !ok {
+					return nil, errors.Wrap(err, fmt.Sprintf("failed to create base limiter for feature: %s and actor id: %s", feature, l.Identifier))
+				}
+				if err := l.ResetUsage(); err != nil {
```
chrsmith, line 116: This is probably fine... but something doesn't sit right with me that a type named `*Static*Limiter` has a function called `ResetUsage`. (Something that is static would be presumably immutable, right?)

Would a better way for me to think about it be that the "static" nature of the `StaticLimiter` is that the _limit_ is the thing that is static. (But not the _usage_ data, allowing you to go up to that limit.) Is that what is going on here?

cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
```diff
@@ -91,6 +98,27 @@ func (s *Source) fetchAndCache(ctx context.Context, token string) (*actor.Actor,
...
+					return nil, errors.Wrap(err, fmt.Sprintf("failed to reset usage cache for feature: %s and actor id: %s", feature, l.Identifier))
```
chrsmith, line 117: Is `l.Identifier` the right thing here? (Same as on line 114.)

`l.Identifier` is defined as: "Identifier is the key used to identify the rate limit counter.".

But wouldn't you be able to infer the key based on the actor ID and feature? Like the actual error text, do we want to have this be `act.GetID()` or something?

cmd/cody-gateway/internal/limiter/limiter.go
chrsmith: Thinking aloud, if this is just a "safeguard" type of check to ensure things are correct/consistent, could we move this to only happen in the async background job that refreshes limits every 15m? (And not for interactive requests initiated by a user?)

cmd/cody-gateway/internal/limiter/limiter.go
```diff
@@ -155,15 +167,15 @@ func (l StaticLimiter) TryAcquire(ctx context.Context) (_ func(context.Context,
 		}
 
 		// Set expiry on the key. If the key didn't exist prior to the previous INCR,
-		// it will set the expiry of the key to one day.
+		// it will set the expiry of the key to interval seconds.
```
chrsmith, line 170: Super minor nit: since we are referring to an actual variable name, we should refer to `intervalSeconds`. (And not "interval seconds" which could be interpreted as just English text.)

```suggestion
		// it will set the expiry of the key to `intervalSeconds`.
```

cmd/cody-gateway/internal/limiter/limiter.go
```diff
@@ -182,6 +194,10 @@ func (l StaticLimiter) TryAcquire(ctx context.Context) (_ func(context.Context,
 	}, nil
 }
 
+func (l StaticLimiter) ResetUsage() error {
```
chrsmith, line 197: Could you please ad a doc comment here, explaining what this method does and when it should be used?

e.g. when does it makes sense to call `ResetUsage()`? Are there any downsides? Will calling this in the wrong place lead to panics or crashes?

cmd/cody-gateway/shared/main.go
```diff
@@ -94,6 +94,29 @@ func Main(ctx context.Context, obctx *observation.Context, ready service.ReadyFu
 		return errors.Wrap(err, "failed to initialize external http client")
 	}
 
+	// Add a prefix to the store for globally unique keys and simpler pruning.
+	rs := limiter.NewPrefixRedisStore("rate_limit:", newRedisStore(redispool.Cache))
+
+	// Ignore the error because it's already validated in the config.
+	dotcomURL, _ := url.Parse(config.Dotcom.URL)
+	dotcomURL.Path = ""
```
chrsmith, line 102: What is the significance of this? Does the `SlackRateLimitNotifier` only care about the domain or something? The parameter on `NewSlackRateLimitNotifier` is `dotcomURL`, so perhaps by removing the path we are introducing a potential for ambiguity?

Since this is just existing code moved around, "🤷 no idea dawg" is a perfectly acceptable answer :)

chrsmith (approved):
Nothing seems obviously wrong with the code, and most of my feedback is just trying to better understand things. (It may be fair to say that this part of the system is more complex than it ought to be, or at least we'd like it to be.)

Anyways, I'll chat with @bobheadxi and see how much risk there would be in trying to address PR feedback and getting this merged/deployed today. Since the customer experience because of this is pretty bad. And if the overhead of the additional Redis calls doesn't increase latency too much, we can improve things in a follow-up PR.

cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
vdavid: @chrsmith, to answer the question, yes, we currently don't store the period start date anywhere in Gateway's cache, so we can't notice the period's end. We _could_ of course do that and not rely on Redis's TTL. Not sure what drove the design decision to rely on the TTL. I guess the requirements were a lot simpler and it was a good enough solution at the time. But I think with this PR's change, we can still roll with it.

cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
```diff
@@ -91,6 +98,27 @@ func (s *Source) fetchAndCache(ctx context.Context, token string) (*actor.Actor,
...
+			}
+
+			// reset the usage cache if:
+			// 1. old actor is not present in the cache
+			// 2. new rl interval is different from old rl interval
+			if oldAct.IsEmpty() || rl.Interval != oldAct.RateLimits[feature].Interval {
+				// get the base limiter for the feature which implements the core rate limiting based on the config
+				l, ok := act.BaseLimiter(s.redisStore, feature, s.rateLimitNotifier)
+				if !ok {
+					return nil, errors.Wrap(err, fmt.Sprintf("failed to create base limiter for feature: %s and actor id: %s", feature, l.Identifier))
```
chrsmith, line 114: ⚠️ This is actually a bug.

Here we are wrapping `err` which is defined earlier in the call to `json.Marshal(act)`. We can assume it is `nil`, so `errors.Wrap(nil, ...)` IIRC will always return `nil`. So this code would silently swallow the problem getting the base limiter.

The right fix might just be to replace `errors.Wrap` with `errors.Errorf("failed to create base limiter..."`

cmd/cody-gateway/internal/actor/actor.go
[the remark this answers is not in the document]
bobheadxi: Or, just give `limiter.RedisStore` to the actor source? The limiter identifier is provided by the actor source, so exposing Limiter back to Source makes the dependency cyclic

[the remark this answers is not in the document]
bobheadxi: The desired reset usage func is just: `return l.Redis.Del(l.Identifier)`s

cmd/cody-gateway/internal/limiter/limiter.go
bobheadxi: If you want to safeguard, you should do it in `updateOnErrorLimiter` - it currently is used to trigger behaviour if a user hits a rate limit, out of the hotpath of requests

bobheadxi: i.e. if a user hit a rate limit, double-check that TTL etc is right

cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
```diff
@@ -76,11 +83,11 @@ func (s *Source) Update(ctx context.Context, act *actor.Actor) error {
...
 
 // fetchAndCache fetches the dotcom user data for the given user token and caches it
-func (s *Source) fetchAndCache(ctx context.Context, token string) (*actor.Actor, error) {
+func (s *Source) fetchAndCache(ctx context.Context, token string, oldAct *actor.Actor) (*actor.Actor, error) {
 	var act *actor.Actor
 	resp, checkErr := s.checkAccessToken(ctx, token)
 	if checkErr != nil {
-		// Generate a stateless actor so that we aren't constantly hitting the dotcom API
+		// Generate a stateless actor and save in the cache so that we aren't constantly hitting the dotcom API
 		act = newActor(s, token, dotcom.DotcomUserState{}, s.concurrencyConfig)
```
bobheadxi, line 91: I made this comment in some other places but this zero-state code path should trigger if and only if we get `actor.ErrAccessTokenDenied` (see source of `checkAccessToken`). We don't want to set a zero state if dotcom is down.

bobheadxi: I think you can also make it so that any `gqlerror.List` will trigger this code path, since that should be a "real" application error and not just some downtime

cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
bobheadxi: The identifier is always the actor ID (I think we should make this clearer)

cmd/cody-gateway/internal/limiter/limiter.go
vdavid: @chrsmith's idea sounds good to me

bobheadxi: Or, the probably better approach, we make the TTL-setting code currently in `defer` always run, so that we don't have 2x the TTL checks

Reading this code again, I'm also unsure whether we should keep `UpdateRateLimitTTL`, and instead we should encode the behaviour in this PR (making resetting usage a first-class citizen)

rafax: Yeah, what @chrsmith said is what I meant but didn't fully express - this code feels like an "extra" check not the main solution, so it'd be great to avoid adding it to every single request.

cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
vdavid: Nice catch!

cmd/cody-gateway/internal/actor/actor.go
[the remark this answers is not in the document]
bobheadxi: Making the limiter's store available to actor sources would also be in-line with making usage resetting a first-class citizen: https://github.com/sourcegraph/sourcegraph/pull/59330/files#r1446585468

chrsmith:
@thenamankumar , @rafax I made some changes to this PR and added a test, pushing it to a new feature branch.
https://github.com/sourcegraph/sourcegraph/pull/59453

Feel free to merge that PR into this feature branch (`naman/reset-cody-usage-cache`), changes that are useful, or ignore it entirely. But hopefully it will be somewhat helpful for understanding/validating the expected behavior here.

Again, AFAICT, everything seems reasonable and correct. Though @bobheadxi and I didn't feel comfortable enough merging this as-is and rolling it out today. And instead would want to make sure Rafal and Erik got a chance to sign off.

cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
bobheadxi: Also, are we certain `s.redisStore` is the right one to use? At the root, the limiter's RedisStore is a prefixed one: https://sourcegraph.com/github.com/sourcegraph/sourcegraph@09b46b44a2c3d36ba3921b248c0d48715d98fb33/-/blob/cmd/cody-gateway/internal/httpapi/handler.go?L74-75

cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
```diff
@@ -45,16 +48,20 @@ type Source struct {
 	cache             httpcache.Cache // cache is expected to be something with automatic TTL
 	dotcom            graphql.Client
 	concurrencyConfig codygateway.ActorConcurrencyLimitConfig
+	redisStore        limiter.RedisStore
```
bobheadxi, line 51:
```suggestion
	usageStore        limiter.RedisStore
```

rafax (commented):
2 areas that I think we should address:

- https://github.com/sourcegraph/sourcegraph/pull/59330/files#r1442799857 and 
@bobheadxi  specific suggestion https://github.com/sourcegraph/sourcegraph/pull/59453/files#r1446764596
    - this is “just” code quality, so it can happen as a quick follow-up PR if that’s what you prefer
- https://github.com/sourcegraph/sourcegraph/pull/59330/files#r1446204614 is unresolved - I think we should apply one of the suggestions from that comment thread (remove extra checks / move to defer/move to background)
    - this adds X0ms of latency to every Cody request and can ~double the traffic to Redis, which feels like a high price to pay for something that is not really needed to fix the issue

thenamankumar:
I have addressed the feedback shared here https://github.com/sourcegraph/sourcegraph/pull/59330#discussion_r1446204614

And merged the PR as other are non-critical feedbacks around refactoring or grammar. Will be addressing all of the open comments in a separate PR.

cc: @vdavid @rafax @bobheadxi @chrsmith

cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
rafax: @thenamankumar this still needs to be addressed in follow-up PRs

thenamankumar (commented):
I forgot to post these comments. Github UX sucks. Thanks @bobheadxi for teaching me this. TIL.

PR #59331, opened 2024-01-04
Searcher: simplify the 'matcher' interface

jtibshirani:
This simplifies the 'matcher' interface in preparation for adding two new implementations, `orMatcher` and `andMatcher`. Changes:
* Remove `MatchesAllContent`. Now the matcher is just nil if it matches everything
* Remove `AddAttributes` in favor of a `String` method
* Remove `IgnoreCase` method, since this is not really a property of a specific matcher/ clause

## Test plan

Refactor is covered by existing unit tests

### DIFF
--- a/cmd/searcher/internal/search/matcher.go
+++ b/cmd/searcher/internal/search/matcher.go
@@ -15,9 +15,6 @@ import (
 )
 
 type matcher interface {
-	// IgnoreCase returns whether matches will ignore case
-	IgnoreCase() bool
-
 	// MatchesString returns whether the string matches
 	MatchesString(s string) bool
 
@@ -148,10 +145,6 @@ func longestLiteral(re *syntax.Regexp) string {
 
 func (rm *regexMatcher) String() string {
 	return fmt.Sprintf("re: %q", rm.re)
-}
-
-func (rm *regexMatcher) IgnoreCase() bool {
-	return rm.ignoreCase
 }
 
 func (rm *regexMatcher) MatchesString(s string) bool {
--- a/cmd/searcher/internal/search/matcher.go
+++ b/cmd/searcher/internal/search/matcher.go
@@ -2,27 +2,19 @@ package search
 
 import (
 	"bytes"
+	"fmt"
 	"regexp/syntax"
 	"strings"
 
 	"github.com/grafana/regexp"
 	"github.com/sourcegraph/zoekt/query"
-	"go.opentelemetry.io/otel/attribute"
+	zoektquery "github.com/sourcegraph/zoekt/query"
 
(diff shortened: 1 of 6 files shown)


cmd/searcher/internal/search/matcher.go
```diff
@@ -166,9 +153,6 @@ func (rm *regexMatcher) MatchesString(s string) bool {
 }
 
 func (rm *regexMatcher) matchesString(s string) bool {
-	if rm.re == nil {
```
stefanhengl, line 169: Reading this, I assume the invariant is "regexMatcher.re is never nil". I know you use `rm = nil` instead of `re.re = nil` as signal for "matches all" so that makes sense. How do we enforce the invariant?

I think the docstring for rm.re needs an update, doesn't it?

stefanhengl (approved):
LGTM!

cmd/searcher/internal/search/search_regex.go
```diff
@@ -52,14 +53,17 @@ func regexSearch(
...
 	contextLines int32,
 ) (err error) {
 	tr, ctx := trace.New(ctx, "regexSearch")
 	defer tr.EndWithErr(&err)
 
-	m.AddAttributes(tr)
 	tr.SetAttributes(attribute.Stringer("path", pm))
+	if m != nil {
+		tr.SetAttributes(attribute.String("matcher", m.String()))
+	}
```
camdencheek, line 66: I generally find it a pretty dangerous pattern to allow interface values to be `nil` because it forces us to do `nil` checks (like this) everywhere we want to use them.

Instead of using `nil` as a sentinel value to mean "matches everything", would it be possible to make a "matches everything" implementation of `matcher`? We do this in both Zoekt and gitserver commit search with "const" nodes in the match tree.

cmd/searcher/internal/search/matcher.go
jtibshirani: Indeed that is the invariant. It's enforced in `compilePattern`, which is the only method through which `regexMatcher` should be created. Did you have another idea for enforcing it in mind?

Good catch with the docstring -- I updated it.

cmd/searcher/internal/search/search_regex.go
jtibshirani: Very good point. It's not a huge deal for private struct fields to be `nil`, but when interfaces can have `nil` values it creates a large "blast radius" for nil deferences.

I introduced an `allMatcher` instead and it turned out well... just requires a type check in one place. (Note to self: type checks in Go are clean compared to many other languages and don't seem as discouraged.)

camdencheek: That looks much nicer, thanks!

camdencheek (approved):

PR #59334, opened 2024-01-04
search: json encode repo metadata for csv export

stefanhengl:
We received the feedback that it is not always obvious that metadata is line-separated in the CSV export of search results. Depending on the tool used to display the CSV, only the first metadata value might be visible. Additionally the current format can be hard to parse. 

Here we add a new column "Repository metadata JSON" to the CSV export of search results, which encodes metadata into a single line JSON. We keep "Repository metadata" for backward compatibility. 

image

[example.csv](https://github.com/sourcegraph/sourcegraph/files/13834555/example.csv)

## Test plan
- updated unit test
- manual testing: 
  - validated the JSON encoding with a python script that reads the CSV and parses the JSON field.


jtibshirani (approved):
Do you think we could just replace "Repository metadata" to keep the output clean? It seems like a low probability someone would have a script relying on the contents of this column?

jtibshirani:
> However, I know of at least 2 customers that rely on this field

I see, that's what I was wondering about. Then your approach makes sense to me to just deprecate it for now.

jtibshirani (approved):

PR #59338, opened 2024-01-04
monitoring: remove very long description

bobheadxi:
Spotted this alert the other day, we have dedicated fields for extended descriptions, and some parts (like "5s") is redundant because the alerting rules already provide that in the template message

image

## Test plan

n/a

## Preview 🤩
[Preview Link](https://docs.sourcegraph.com/@remove-paragraph-description)

### DIFF
--- a/doc/admin/observability/alerts.md
+++ b/doc/admin/observability/alerts.md
@@ -231,18 +231,19 @@ Generated query for warning alert: `max((histogram_quantile(0.9, sum by (le) (ra
 
 ## frontend: blob_load_latency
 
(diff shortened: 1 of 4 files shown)


pjlast (approved):

pjlast:
As a side note: I question the usefulness of this alert. I've been digging into what's triggering it over the last few days, and the moment we have ~4 slow requests in a row, the alert is triggered. Sometimes these requests are unusual, like 5 requests being made to the same `/-/blob/...` route within milliseconds of each other, and it immediately resolves itself after that. But then someone has already received a very loud OpsGenie alert in the middle of the night

bobheadxi:
> the moment we have ~4 slow requests in a row, the alert is triggered. Sometimes these requests are unusual, like 5 requests being made to the same /-/blob/... route within milliseconds of each other, and it immediately resolves itself after that

Maybe you can play around with adding a `.For(10 * time.Minute)` or similar on the threshold 🤔 so that this alert fires only on extended periods

PR #59340, opened 2024-01-04
Logging previous values during setting changes

mohammadualam:
To be able to do deeper investigation, we want to log previous settings in cases of changes. this PR will add arguments to current logging that will capture those previous settings before new settings are applied. 

## Test plan

locally tested


cmd/frontend/graphqlbackend/external_services.go
```diff
@@ -172,33 +179,40 @@ func (r *schemaResolver) UpdateExternalService(ctx context.Context, args *update
...
 			DisplayName *string
 			UpdaterID   *int32
+			PrevConfig  string
+			NewConfig   *string
 		}{
 			ID:          args.Input.ID,
 			DisplayName: args.Input.DisplayName,
 			UpdaterID:   &userID,
+			PrevConfig:  prevConfig,
+			NewConfig:   &latestConfig,
```
evict, line 208: nit: Is `NewConfig` accurate if we call the arg `latestConfig`?

internal/auth/userpasswd/reset_password.go
```diff
@@ -151,7 +151,7 @@ func HandleResetPasswordCode(logger log.Logger, db database.DB) http.HandlerFunc
 			} else {
 
 				anonymousID, _ := cookie.AnonymousUID(r)
-				if errsec := db.SecurityEventLogs().LogSecurityEvent(r.Context(), database.SecurityEventNameEmailVerified, r.URL.Path, uint32(params.UserID), anonymousID, "BACKEND", nil); errsec != nil {
+				if errsec := db.SecurityEventLogs().LogSecurityEvent(r.Context(), database.SecurityEventNameEmailVerified, r.URL.Path, uint32(params.UserID), anonymousID, "BACKEND", params.Email); errsec != nil {
```
evict, line 154:
```suggestion
				if err := db.SecurityEventLogs().LogSecurityEvent(r.Context(), database.SecurityEventNameEmailVerified, r.URL.Path, uint32(params.UserID), anonymousID, "BACKEND", params.Email); err != nil {
```

cmd/frontend/graphqlbackend/user.go
```diff
@@ -957,7 +969,7 @@ func (r *schemaResolver) SetUserCompletionsQuota(ctx context.Context, args SetUs
 	if featureflag.FromContext(ctx).GetBoolOr("auditlog-expansion", false) {
```
evict, line 970:
```suggestion
```

cmd/frontend/graphqlbackend/user.go
```diff
@@ -946,6 +946,18 @@ func (r *schemaResolver) SetUserCompletionsQuota(ctx context.Context, args SetUs
 		return nil, err
 	}
 
+	// Capture values for logging prior to change
+	prevQuota, _ := r.db.Users().GetChatCompletionsQuota(ctx, user.ID)
```
evict, line 950: Do we really need to catch the previous quota? Is capturing what changed and who did it enough?

evict (approved):
Nice work! I do think it might be unnecessary to catch the previous config/value/state. It adds an extra database call, and besides helping us revert, I don't see any added benefit of logging the previous state.

cmd/frontend/graphqlbackend/external_services.go
mohammadualam: fixed

cmd/frontend/graphqlbackend/external_services.go
```diff
@@ -172,33 +179,40 @@ func (r *schemaResolver) UpdateExternalService(ctx context.Context, args *update
 		return nil, err
 	}
 
-	if featureflag.FromContext(ctx).GetBoolOr("auditlog-expansion", false) {
+	// Fetch from database again to get all fields with updated values.
+	es, err = r.db.ExternalServices().GetByID(ctx, id)
```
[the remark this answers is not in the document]
mohammadualam: This was actually original code, i just moved my logging code below it to be able to utilize the new pulled values so its mistakenly showing it as new code.

mohammadualam:
> Nice work! I do think it might be unnecessary to catch the previous config/value/state. It adds an extra database call, and besides helping us revert, I don't see any added benefit of logging the previous state.

yeah there is an added call to get the previous quota. Ill get more feedback if this is something we want. if not then ill take it out.

cmd/frontend/graphqlbackend/user.go
mohammadualam: removed as discussed

PR #59344, opened 2024-01-04
Filters panel: refine type icons

camdencheek:
- Changed the "Code" icon to curly braces so we didn't have so many vaguely-rectangular icons
- Changed the "Repositories" icon to a fork icon because the normal "Source Repository" icon stands out much more than the other icons and it looked out of place. 
- Changed the "Symbols" icon to the function icon, which is widely used to represent generic "symbols" elsewhere.

Before:
CleanShot 2024-01-04 at 15 58 56@2x

After:
CleanShot 2024-01-04 at 15 58 45@2x

# Test Plan

See screenshots. We'll let design have a pass at this too once Taiyab is back.

### DIFF
--- a/client/branded/src/search-ui/results/filters/components/filter-type-list/FilterTypeList.tsx
+++ b/client/branded/src/search-ui/results/filters/components/filter-type-list/FilterTypeList.tsx
@@ -1,13 +1,6 @@
 import { FC } from 'react'
 
-import {
-    mdiBook,
-    mdiCodeBrackets,
-    mdiFileOutline,
-    mdiPlusMinus,
-    mdiShapeSquareRoundedPlus,
-    mdiSourceCommit,
-} from '@mdi/js'
+import { mdiSourceFork, mdiCodeBraces, mdiFileOutline, mdiPlusMinus, mdiFunction, mdiSourceCommit } from '@mdi/js'
 import classNames from 'classnames'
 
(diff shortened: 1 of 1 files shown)


jasonhawkharris (approved):
Very nice. LGTM!

client/branded/src/search-ui/results/filters/components/filter-type-list/FilterTypeList.tsx
```diff
@@ -1,13 +1,6 @@
...
 
-import {
-    mdiBook,
-    mdiCodeBrackets,
-    mdiFileOutline,
-    mdiPlusMinus,
-    mdiShapeSquareRoundedPlus,
-    mdiSourceCommit,
-} from '@mdi/js'
+import { mdiSourceFork, mdiCodeBraces, mdiFileOutline, mdiPlusMinus, mdiFunction, mdiSourceCommit } from '@mdi/js'
```
jasonhawkharris, line 3: I bet the linter get's mad at you for not having a new line for each of these.

camdencheek: this is weird. I ran `pnpm format:changed` but it didn't do anything

camdencheek: oh, turns out it was just a transitive failure

PR #59349, opened 2024-01-05
sg: add 'fix all' choice in setup

burmudar:
Add the choice to `fix all`

"Fix everything" choice
Screenshot 2024-01-05 at 12 12 27

Other choices still work 😄 
Screenshot 2024-01-05 at 12 12 41

Closes https://github.com/sourcegraph/devx-support/issues/512
## Test plan
Tested locally


jamesmcnamara (approved):
Couldn't have said it better myself.

jamesmcnamara (changes requested):
One discrepancy here is that the numbering is off. In the existing impl, the failure numbers are zero indexed, but the numbers displayed to the user are 1-indexed. So currently it looks like this:
Screenshot 2024-01-05 at 9 53 35 AM

So not only are the numbers off but I don't have the option to fix the first problem singularly.

burmudar:
> One discrepancy here is that the numbering is off. In the existing impl, the failure numbers are zero indexed, but the numbers displayed to the user are 1-indexed. So currently it looks like this: Screenshot 2024-01-05 at 9 53 35 AM
> 
> So not only are the numbers off but I don't have the option to fix the first problem singularly.

Ah yeah good catch. Missed that. I'll fix that

burmudar:
And fixed 
Screenshot 2024-01-08 at 11 48 57

burmudar:
> @burmudar can you add a line in https://github.com/sourcegraph/team-infra-newsletter/blob/main/newsletters/2024-01-05.md as well? I'll post it this Friday :)

Added 🚀

PR #59219, opened 2023-12-22
use vite for web builds (#58228)

jamesmcnamara:
* remove little-used `web-standalone-http-prod`

This let you run a local web app built in production mode against a remote Sourcegraph endpoint. You can still run a local web app built in *dev* mode.

* add `sg test bazel-backend-integration`

* fix DeveloperDialog positioning

It was taking up 100% width and was translated -50% so the left half of it was off-viewport.

* use vite for web builds

[Vite](https://vitejs.dev/) is a fast web bundler. With Vite, a local dev server is available as fast as with esbuild, and incremental builds are (1) hot (no page reload needed) and (2) much faster (<500ms).

* fix "manifestPlugin.d.ts" was not created

* sg lint

* remove little-used `web-standalone-http-prod`

This let you run a local web app built in production mode against a remote Sourcegraph endpoint. You can still run a local web app built in *dev* mode.

* add `sg test bazel-backend-integration`

* fix DeveloperDialog positioning

It was taking up 100% width and was translated -50% so the left half of it was off-viewport.

* use vite for web builds

[Vite](https://vitejs.dev/) is a fast web bundler. With Vite, a local dev server is available as fast as with esbuild, and incremental builds are (1) hot (no page reload needed) and (2) much faster (<500ms).

* fix "manifestPlugin.d.ts" was not created

* sg lint

* small lint fix

* added events shim to client/web/BUILD.bazel

* updated via bazel configure

* added in side-effectful import for EventEmitter

* added in side-effectful import for EventEmitter

* ran bazel configure

* re-run bazel configure

* pnpm dedupe

---------

PR #59351, opened 2024-01-05
Fix broken monaco code completion

pjlast:
Code completion in the site config on dotcom has been broken for a while now. The monaco editor depends on some worker scripts being available at `/.assets/scripts` for code completion suggestions to work in the online editor.

This is not an issue locally because we run an esbuild dev server that proxies requests to `/.assets`

This PR attempts to fix it by adding the `scripts` folder to the bazel build target for `//client/web/dist`

## Test plan

Tested locally by inspecting the output of `bazel-bin/client/web/dist` after running `bazel build //client/web/dist`, but I'm not sure how thorough this test is.

### DIFF
--- a/client/web/dist/scripts/BUILD.bazel
+++ b/client/web/dist/scripts/BUILD.bazel
@@ -11,6 +11,6 @@ filegroup(
 
 copy_to_bin(
     name = "copy",
-    srcs = ["//client/web/dist/scripts"],
+    srcs = [":scripts"],
     visibility = ["//client/web-sveltekit:__pkg__"],
 )
--- a/client/web/dist/scripts/BUILD.bazel
+++ b/client/web/dist/scripts/BUILD.bazel
@@ -2,7 +2,10 @@ load("@aspect_bazel_lib//lib:copy_to_bin.bzl", "copy_to_bin")
 
 filegroup(
     name = "scripts",
-    srcs = glob(["**/*.*"]),
+    srcs = glob(
+        ["**/*.*"],
+        ["BUILD.bazel"],
+    ),
     visibility = ["//visibility:public"],
 )
 
--- a/client/web/dist/scripts/BUILD.bazel
+++ b/client/web/dist/scripts/BUILD.bazel
@@ -1,0 +1,13 @@
+load("@aspect_bazel_lib//lib:copy_to_bin.bzl", "copy_to_bin")
+
+filegroup(
+    name = "scripts",
+    srcs = glob(["**/*.*"]),
+    visibility = ["//visibility:public"],
+)
+
+copy_to_bin(
+    name = "copy",
+    srcs = ["//client/web/dist/scripts"],
+    visibility = ["//client/web-sveltekit:__pkg__"],
+)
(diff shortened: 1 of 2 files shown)


client/web/dist/scripts/BUILD.bazel
```diff
@@ -0,0 +1,16 @@
+load("@aspect_bazel_lib//lib:copy_to_bin.bzl", "copy_to_bin")
+
+filegroup(
+    name = "scripts",
+    srcs = glob(
+        ["**/*.*"],
+        ["BUILD.bazel"],
```
Strum355, line 7: Does this need to be included? Id be surprised if it does but what do I know :clueless:

client/web/dist/scripts/BUILD.bazel
```diff
@@ -0,0 +1,16 @@
...
+    ),
+    visibility = ["//visibility:public"],
+)
+
+copy_to_bin(
+    name = "copy",
+    srcs = ["//client/web/dist/scripts"],
```
Strum355, line 14: Can be simplified to `":scripts"`

client/web/dist/scripts/BUILD.bazel
pjlast: idk man I've tried various builds to get an iteration that might work. I just tried to copy what the `/img` directory is doing

valerybugakov (commented):
Hey @Strum355, based on [the diff history](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/sourcegraph%24+type:diff+buildMonaco%28&patternType=standard&sm=1&groupBy=author), the `buildMonaco` function was [introduced](https://github.com/sourcegraph/sourcegraph/commit/bf6fadfed9ddb66dac4bfd598a28b6bf522d5f70#diff-9fb5a352314757868e817dfdb8a23d523bb8331c7adeb8b385980af7f67a9591R67) into the web package quite some time ago, accompanied by a `TODO` note suggesting a switch to a more suitable alternative for production builds. Subsequently, during our transition from `webpack` to `esbuild` for production builds, this function [was not utilized](https://github.com/sourcegraph/sourcegraph/commit/e1f9a6f27f75588768f929afb05376ca6dc43859#diff-f7c1ab1acb5e753bccdd4b092b0098fa58855899e8764b68829ada2ba9cea760), leading to an unresolved bug since that change, as far as I understand. It appears that `buildMonaco` is now only triggered when the `esbuild/build.ts` script is executed — specifically for development builds outside of Bazel.

The `vite` [migration PR](https://github.com/sourcegraph/sourcegraph/pull/59219) addresses this issue by [resolving](https://github.com/sourcegraph/sourcegraph/blob/86180de04a599ce847540badeb0e5dacf116c83b/client/web/src/settings/MonacoSettingsEditor.tsx#L370) the initial `TODO` mentioned above the `buildMonaco` function. Therefore, we can potentially forego a specific fix for esbuild if the migration PR can be merged soon.

pjlast:
Closing this in favour of #59219

PR #59357, opened 2024-01-05
Return an empty embeddings job list when embeddings are disabled

chwarwick:
When embeddings are disabled and admins view the web client no repositories were listed because the embeddings jobs query would return an error indicating embeddings are disabled and the first nullable resolver is the root resolver.

This updates the embeddings jobs resolver to return an empty list when embeddings are disabled.  Ideally we would allow and handle the error but it would be a major refactor since currently nothing is nullable.

Additionally this moves the security checks prior to return any errors regarding enable/disable.

## Test plan
Added unit test
disabled embeddings and logged in as admin verified that repo list available in web chat.


slimsag (approved):

chwarwick:
replaced by https://github.com/sourcegraph/sourcegraph/pull/59392

PR #59358, opened 2024-01-05
Code Navigation: Fix isLimitHit on individual filters

jasonhawkharris:
Before, filters were updated individually using the filters.Add() function. This caused an issue where not all filters would show that searches were incomplete, only some of them. 
 
Now, filters are set universally based on the value of SearchFilters.isLimitHit.

## Test plan
Shipped with passing unit tests
CI should take care of the rest.


camdencheek (approved):
LGTM! Looks like CI is complaining about formatting though

jasonhawkharris:
Closing this because we've replaced it with [this](https://github.com/sourcegraph/sourcegraph/pull/59365).

PR #59360, opened 2024-01-05
Update unified cody context to only use search when embeddings are disabled

chwarwick:
Currently when embeddings are disabled the unified cody context will continue to try to use embeddings because they were previously generated.  This results in errors for each repo, stating that embeddings are disabled.

This updates the repo partitioning to only use search if embeddings are disabled. 
## Test plan
Disabled embeddings
Used cody web client - verified additional context added to the request

### DIFF
--- a/cmd/frontend/internal/context/resolvers/context_test.go
+++ b/cmd/frontend/internal/context/resolvers/context_test.go
@@ -33,10 +33,20 @@ import (
 func TestContextResolver(t *testing.T) {
 	logger := logtest.Scoped(t)
 	ctx := context.Background()
-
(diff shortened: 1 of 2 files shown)


camdencheek (approved):

arafatkatze (approved):

arafatkatze:
I don't really explore this PRs code in a proper testing frame but it seems simple enough so I trust to approve it. I am aware of the embeddings not present issues which can be annoying so this is better. 


The steps look reasonable
```

Disabled embeddings
Used cody web client - verified additional context added to the request
```

PR #59362, opened 2024-01-05
Fix storybook type errors and add CI test

jamesmcnamara:
EDIT: @valerybugakov found an example of where [custom gazelle groups](https://github.com/aspect-build/aspect-cli/blob/5.5.2/gazelle/js/tests/groups_deps/BUILD.in) in javascript (unfortunately it doesn't appear to be documented).
So I've updated the PR to use that.

ORIGINAL:
I think the expectation for this ticket was that we could use [custom gazelle groups](https://github.com/sourcegraph/sourcegraph/blob/main/client/web/BUILD.bazel#L45) but from what I can see in the [documentation](https://docs.aspect.build/cli/help/directives/), I don't think that has been added.

I tried a lot of different approaches here none of which worked, before trying the dumb thing of just creating a ts_project which I didn't realize created a typing test target automatically. This makes me think there was probably a reason we didn't do this in the first place? 

## Test plan
Added test target run in CI


client/web/BUILD.bazel
```diff
@@ -2157,3 +2141,281 @@ bundlesize_bin.bundlesize_test(
         "WEB_BUNDLE_PATH": "$(rootpath //client/web:bundle)",
     },
 )
+
+ts_project(
+    name = "stories",
```
valerybugakov, line 2146: Beautiful

valerybugakov (approved):
Should we do the same for other client packages with Storybook stories?

jamesmcnamara:
@valerybugakov I think that'd be really helpful for preventing these type errors to creep from creeping into all our storybook setups. If there's no objections I can grep around for storybook files and add this.

jhchabran (approved):

PR #59363, opened 2024-01-05
Searcher: Implement 'and' and 'or' match trees

jtibshirani:
This change implements the searcher logic for 'and' and 'or' queries:
* Renames `matcher` to `matchTree` to be more consistent with Zoekt and
gitserver
* Adds `andMatchTree` and `orMatchTree`, plus a bunch of tests

This code is not yet hooked in end-to-end. In a final follow-up PR, we'll
update the jobs code to pass entire basic queries to searcher, and translate
them to these new match trees.

Relates to #59038

## Test plan

Added a bunch of new unit tests to cover and/or logic, path matching,
overlapping ranges, and match limits.


cmd/searcher/internal/search/search_regex.go
```diff
@@ -155,7 +155,8 @@ func regexSearch(
 					casetransform.BytesToLowerASCII(fileMatchBuf, fileBuf)
 				}
 
-				match, locs := m.MatchesFile(fileMatchBuf, sender.Remaining())
+				// find limit+1 matches so we know whether we hit the limit
```
jtibshirani, line 158: I pulled this up out of `regexMatchTree` to make it more obvious what was happening.

cmd/searcher/internal/search/matchtree.go
```diff
@@ -231,9 +233,143 @@ func (rm *regexMatcher) ToZoektQuery(matchContent bool, matchPath bool) (zoektqu
...
+	return &zoektquery.Or{Children: children}, nil
+}
+
+func (om *orMatchTree) String() string {
+	return fmt.Sprintf("OR (%d children)", len(om.children))
+}
+
+// mergeMatches sorts the matched ranges, merges overlapping ranges,
+// and finally truncates the list to obey 'limit'
+func mergeMatches(matches [][]int, limit int) [][]int {
```
jtibshirani, line 332: I wasn't sure if I should be merging overlapping ranges here. It felt like the most correct behavior, and there's [downstream logic](https://github.com/sourcegraph/sourcegraph/blob/8801a54ae5e2f492ba218e8977c878b7d1ae7944/cmd/searcher/internal/search/search_regex.go#L238) that assumes the ranges are non-overlapping. However, I don't think it's consistent with what we do elsewhere:
* In Zoekt, we merge matches if we're using "file matches". But if we use "chunk matches" then we [skip overlapping matches](https://github.com/sourcegraph/zoekt/blob/e92f6c764b567d9b359633fa4045cfba9d37eb3c/eval.go#L604-L605).
* For diff/ commit search, we [avoid merging overlapping matches altogether](https://github.com/sourcegraph/sourcegraph/blob/main/internal/search/result/range.go#L136)

I'm not sure of the motivation behind these different approaches. Do you have thoughts on what to do here?

camdencheek: I personally think either skipping overlapping matches or retaining all matches is the most correct behavior.
- Skipping overlapping matches more closely follows what the regex package will return (e.g. `(abc|cde)` will only match once on the string `abcde`). 
- Retaining all matches ensures that we have any piece of the file that matches one of our input queries is  highlighted in the UI (as does merging matches)
- But merging matches means that our matched ranges aren't guaranteed to match any single atom of our query, which is a little surprising. 

FWIW, that comment on the downstream logic is only because it made it easier to write an efficient algorithm for counting newlines and that invariant is maintained naturally if we're getting our ranges directly from the `regexp` package. It would be possible to change the behavior of that function to remove that requirement.

cmd/searcher/internal/search/matchtree.go
```diff
@@ -231,9 +233,143 @@ func (rm *regexMatcher) ToZoektQuery(matchContent bool, matchPath bool) (zoektqu
...
+	if len(matches) == 0 {
+		return matches
+	}
+
+	sort.Sort(matchSlice(matches))
+	newMatches := make([][]int, 0, len(matches))
```
camdencheek, line 338: very minor: we can avoid the allocation here and do the merging in-place
```suggestion
	newMatches := matches[:0]
```

camdencheek (approved):
This looks great!

cmd/searcher/internal/search/matchtree.go
jtibshirani: Oops, I meant to do this but forgot! I was so excited to use this new Go trick I learned.

cmd/searcher/internal/search/matchtree.go
camdencheek: Sounds good to me 👍 We should do at least some cursory manual end-to-end testing of this as well since it's certainly possible that otehr parts of the stack have baked-in assumptions here. There are just none that I know of.

PR #59365, opened 2024-01-05
New Filters Panel: Fix limitHit value on filters

jasonhawkharris:
Before, filters were updated individually using the filters.Add() function. This caused an issue where not all filters would show that searches were incomplete, only some of them.

Now, filters are set universally based on the value of SearchFilters.isLimitHit.

## Test plan
- passing unit tests
- ci should take care of the rest.


camdencheek (approved):
Pushed a couple more commits to simplify some things and fix the build, but LGTM!

PR #59366, opened 2024-01-06
Search: integrate new backend filters

camdencheek:
The first version of the search filters panel did client-side aggregation for commit author, commit date, and symbol type. Backend support for these filters was added in https://github.com/sourcegraph/sourcegraph/pull/59283. This PR updates the client to take advantage of the new server-side-aggregated filters. 

Detailed comments inline, but in summary:
- No more static filters. All filters come from the backend.
- Separate out `Filter['kind']` because it's not guaranteed to correspond to a search query filter type
- Filters are no longer marshaled to a query string for the URL bar, allowing us to retain more information and making 
merging filters _much_ easier.

## Test plan

Lots of manual testing. Found a few bugs with the current implementation during testing that I fixed along the way.


client/branded/src/search-ui/results/filters/NewSearchFilters.tsx
```diff
@@ -23,28 +23,24 @@ import {
...
-    const [filterQuery, setFilterQuery] = useFilterQuery()
+export const NewSearchFilters: FC<NewSearchFiltersProps> = ({ query, filters, onQueryChange }) => {
+    const [selectedFilters, setSelectedFilters] = useFilterQuery()
 
     const type = useMemo(() => {
         const tokens = scanSearchQuery(query)
 
         if (tokens.type === 'success') {
-            const filters = tokens.term.filter(token => token.type === 'filter') as Filter[]
+            const filters = tokens.term.filter((token): token is QueryFilter => token.type === 'filter')
```
camdencheek, line 43: unrelated change, but it makes this filter operation type-safe

client/branded/src/search-ui/results/filters/components/dynamic-filter/SearchDynamicFilter.tsx
```diff
@@ -28,149 +23,83 @@ interface SearchDynamicFilterProps {
      * Specifies which type filter we want to render in this particular
      * filter section, it could be lang filter, repo filter, or file filters.
      */
-    filterType: string | string[]
+    filterKind: Filter['kind']
```
camdencheek, line 26: This must be one of the kinds returned by the backend. Note that this is _not_ necessarily a valid search filter. For example, the kind might be `symbol type` or `commit date` or `utility`. The `Filter['kind']` is just a grouping for the filters.

client/branded/src/search-ui/results/filters/NewSearchFilters.tsx
```diff
@@ -57,8 +53,6 @@ export const NewSearchFilters: FC<NewSearchFiltersProps> = props => {
         return SearchFilterType.Code
     }, [query])
 
-    const authorFilters = useMemo(() => generateAuthorFilters(results ?? []), [results])
```
camdencheek, line 60: This is no longer needed since we do this on the backend.

client/branded/src/search-ui/results/filters/hooks.ts
```diff
@@ -1,10 +1,28 @@
...
+    return JSON.stringify(filters.map(f => [f.kind, f.label, f.value]))
+}
+
+export function deserializeURLQueryFilters(serialized: string | null): URLQueryFilter[] {
+    if (!serialized) {
+        return []
+    }
+    const parsed = JSON.parse(serialized) as [Filter['kind'], string, string][]
+    return parsed.map(([kind, label, value]) => ({ kind, label, value }))
+}
```
camdencheek, line 19: Significant change: do not store filters in the URL as a query string. The information in the filters is important because it's possible to have filters from multiple groups that contain the same query fragment. We cannot safely map from a query fragment back to a filter. So instead, we encode the kind (group), the label, and the value (query fragment) in the URL so we can easily pull that information back off. 

This makes the URL a little more verbose, but I think it's worth it for the sake of robust roundtrips.

client/web/src/search/results/SearchResultsCacheProvider.tsx
```diff
@@ -89,17 +97,13 @@ export function useCachedSearchResults(props: CachedSearchResultsInput): Aggrega
             // and is discussed extensively in github issues. Instead, we just manually
             // merge throttleTime with only leading values and the final value.
             // See: https://github.com/ReactiveX/rxjs/issues/5732
-            return merge(stream.pipe(throttleTime(500)), stream.pipe(last())).pipe(
-                tap(results => {
-                    const previousCache = cachedResults.current?.cache ?? {}
-                    cachedResults.current = { query, options, cache: { ...previousCache, [filterQuery]: results } }
-                })
-            )
+            return merge(stream.pipe(throttleTime(500)), stream.pipe(last(), tap(saveToCache)))
```
camdencheek, line 100: There was a bug here where we would copy the cached results even when the query had changed, which meant it was possible to do things like show old cached repo results for a path query. I fixed this (in `saveToCache`). Additionally, I moved the tap to the `last()` event so that we're only updating the cache on a complete result set.

internal/search/streaming/search_filters.go
```diff
@@ -164,7 +164,7 @@ func (s *SearchFilters) Update(event SearchEvent) {
 	}
 
 	addCommitAuthorFilter := func(commit gitdomain.Commit) {
-		filter := fmt.Sprintf(`author:%s`, commit.Author.Email)
+		filter := fmt.Sprintf(`author:%s`, regexp.QuoteMeta(commit.Author.Email))
```
camdencheek, line 167: Unrelated bug that I came across: emails can contain regex special characters like `+` which we should treat literally.

BolajiOlajide (approved):

client/branded/src/search-ui/results/filters/hooks.ts
vovakulikov: This is was my main concern that URL with filter query will be very very long, maybe it's okay since I don't think that filter set can grow that big (I mean how many filters usually will be set, like 3-4 maybe). Thanks for the change!

PR #59373, opened 2024-01-08
svelte: Fix issues reported by svelte-check

fkling:
Fixes TS/format issues introduced by recent commits.

## Test plan

`pnpm check` doesn't report any issues with files inside `client/web-sveltekit`.


client/web-sveltekit/src/lib/fileIcons.ts
fkling, line 1: This file wasn't used anymore.

PR #59376, opened 2024-01-08
Improves locals queries for Java

kritzcreek:
Fixes the Java specific part of https://github.com/sourcegraph/sourcegraph/issues/57791

No longer reports a number of non-local entities as locals. Also stops recording field access as local. I tried splitting the changes into individual commits, to make the changes to the snapshot easy to review.

Added you as a reviewer as well @keynmol as you originally wrote these queries and might know a few more places I've missed, or should improve?

~There's one TODO here, @varungandhi-src WDYT about adding `@reference.skip` to the DSL? `@definition.skip` works just fine here, but it's a bit unintuitive.~ Fixed

## Test plan

Look at the updated snapshots


docker-images/syntax-highlighter/crates/scip-syntax/queries/java/scip-locals.scm
```diff
@@ -37,10 +37,9 @@
 ;     name: (identifier) @definition.function
 ; )
 
-
-(enum_constant
-  name: (identifier) @definition.term
-)
+; (enum_constant
+;   name: (identifier) @definition.term
+; )
```
varungandhi-src, line 42:
```suggestion
```

docker-images/syntax-highlighter/crates/scip-syntax/queries/java/scip-locals.scm
```diff
@@ -74,5 +82,7 @@
     (type_identifier) @definition.type
 )
 
+;; TODO(Christoph): Do we need @reference.skip?
+(field_access field: (identifier) @definition.skip)
```
varungandhi-src, line 86: Alternately, maybe we should rename `@definition.skip` to `@occurrence.skip` (and update the corresponding Rust code so that the names aren't out-of-sync)? Or are there places where a `@definition.skip` entry would be correctly marked as a reference?

varungandhi-src (approved):

docker-images/syntax-highlighter/crates/scip-syntax/queries/java/scip-locals.scm
kritzcreek: Oh yeah that's much better. I'll go with that

docker-images/syntax-highlighter/crates/scip-syntax/queries/java/scip-locals.scm
kritzcreek: I kept it in, because it ties in with all the other non-local definitions commented out above, which are prefixed with:
```
; NOTE: The definitions below are commented out
; as they overlap with global symbol indexing
; marking type declarations as locals causes
; various confusions, for example around constructors
;
; They are kept here for reference and to avoid re-introducing them
```

PR #59381, opened 2024-01-08
search: revert glob filter values from new keyword pattern type

keegancsmith:
This reverts the changes made to support glob filters in the new keyword search patternType. We have had mixed feedback on the change, so would rather focus on the core improvement which is interpretting via AND.

This PR was generated by looking at the commits by @stefanhengl and reverting the ones that looked related. Please double check this Stefan! :) Below is the list of commits to main in the last month, and I have marked those I picked:

```
  79486a0599d * search: don't highlight operators inside "..." (#59328)
  74c51a73a91 * changelog: display GitHub/GitLab topics (#59211)
  bdc45fa09f2 * search jobs: handle unknown users (#59201)
x 14e382e5686 * search: respect patternType when navigating to search results (#59164)
x 4f6cdb658c1 * search: don't convert rev to regex for kw search (#59165)
x c6f0aa4b5d4 * search: mention glob syntax in search mode picker (#59168)
  2f6286d5f69 * dev: revert to old search mode for local dev environment (#59129)
x a0e660e1a8b * search: use literal suggestions for keyword search (#59120)
x 92cfe03c101 * search: support glob syntax for repo and file filters (#59080)
  f883d66bc7c * search: support "exact match" pattern using "..." (#59057)
```

Test Plan: CI and spun up in local dev. Browsed around with new keyword search mode turned on checking we still used regex. Then did a few searches using regex syntax and confirmed it worked.

Part of https://github.com/sourcegraph/sourcegraph/issues/58815


stefanhengl (approved):
Looks good to me!

PR #59382, opened 2024-01-08
Adjust blobstore latency alert

pjlast:
The >5s blobstore latency alert triggers _very_ eagerly, and it's also self-resolving. I don't think the current threshold works very well as an alert system, as it turns into a boy-who-cried-wolf scenario.

This PR requires that the blobstore latency be >5s for 10 minutes before an alert is triggered.

## Test plan

Alert update


eseliger (approved):

eseliger (approved):

bobheadxi (approved):
thanks!

PR #59383, opened 2024-01-08
svelte: Towards a better data fetching and GraphQL authoring experience

fkling:
This PR refactors almost all of the prototypes GraphQL queries to take advantage of GraphQLs compossibility via fragments. The goal is to provide a more structured approach to authoring and executing GraphQL queries, with the following advantages:
- Data dependency co-location makes it easier to maintain/extend individual components.
- Data fetching happens in specific, predictable places (layout and page loaders).

On a high level it works like this:

- Components declare their data dependencies in `<Component>.gql` files next to them. Thanks to GraphQL code generation they can import the corresponding TypeScript types via `import type { SomeFragment} from './<Component>.gql'`.
- Higher level components compose the fragments of their children.
- At the page/layout level `page.gql`/`layout.gql` files define the queries, composed from the data dependencies of the page/layout.
- The page/layout data loaders can import queries directly from the corresponding `.gql` file.

Authoring the `.gql` files should be relatively easily if the graphql language server is setup. The changes in the `.graphlrc` file make all fragments globally available which means that every fragment needs to be unique and we don't need to use unofficial `#import` directives inside `.gql` files.

There are a couple of things to consider though:

- Caching: If different pages/layouts fetch the same data with different queries, we won't leverage caching without additional setup. That's something I still need to look into. That also means that sometimes we might want to use a shared queries instead of composition/co-location, if caching is more important.
- Shared layout data: Some data fetched in layouts is accessed by sub-layouts/sub-pages, but the layout doesn't know which sub-layout/sub-page is loaded, making query composition more difficult. So it far it seems that data shared this way is rather limited/constrained. So my current approach is to have components/pages define fragments following a specific naming convention, and have the loaders that provide this data compose them. Examples for this is `SearchInput_AuthorizedUser` and `RepoPage_ResolvedRevision`.
  This is not ideal because it has to be remembered to embed this fragment in the right place, but it's not worse than the current situation (where we often don't know where the query providing some data is defined).
- On demand data fetching: Not all data is necessary for rendering a page, some data is only fetched when in response to some user interaction. The layout/page loader should still be the place that executes the query, but instead of doing it on page load it passes a function to the page to fetch the data on demand. This way we can maintain data co-location and fetching in loaders. For an example see the `fetchCommitHistory` function.

NOTE: I expect there to be changes to this approach as we uncover more data loading requirements.

## Test plan

Browsed pages in dev mode and checked output of `pnpm check`.


client/web-sveltekit/.graphqlrc
```diff
@@ -0,0 +1,4 @@
+{
+  "schema": "../../cmd/frontend/graphqlbackend/*.graphql",
+  "documents": "src/**/*.gql"
```
fkling, line 3: This allows the GraphQL language server to know about all defined fragments, making it easy to reference fragments from any `.gql` file and to get validation in the editor.

client/web-sveltekit/src/lib/repo/FileDiff.gql
```diff
@@ -1,5 +1,3 @@
-#import './FileDiffHunks.gql'
```
fkling, line 1: No need for this anymore, it didn't work with the GraphQL language server anyway (the LSP would complain about 'unknown fragments').

client/web-sveltekit/src/lib/repo/blob.ts
```diff
@@ -16,7 +16,7 @@ import {
     type DocumentInfo,
 } from '$lib/web'
 
-import type { BlobFileFields } from './api/blob'
+import type { BlobPage_Blob } from '../../routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/page.gql'
```
fkling, line 19: This is not ideal. I'll have to iterate on that.

client/web-sveltekit/src/testing/testdata.ts
fkling, line 1: I did the minimum to keep this file working. I want to change how testdata is generated in the context of adding integration tests.

fkling:
@camdencheek 

> Do I understand this correctly that if a fragment is required for multiple components, it will show up in the query multiple times?

I think so. I tried to play around with different code generation options to merge/inline fragments, but some threw errors.

> Right now, our backend doesn't do deduplication/caching within a query. Does anything like that happen in the client (e.g. in apollo?)

I don't think so. But this might not be a problem yet because so far most component's data needs don't overlap. I'd be curious to learn more about how the graphql queries are processed on the backend.

camdencheek (approved):
This is very cool!

PR #59397, opened 2024-01-08
Bump go-mockgen to include lenient type checking

eseliger:
Camden added https://github.com/derision-test/go-mockgen/commit/4fea34868d8b36024f9a62eca1387596efe110dc to the tool so that we don't have to have a functioning build when mocks need to be regenerated, which can sometimes cause a cyclic issue.

### Test plan

Ran generate on broken code and it worked, while it failed on main.


camdencheek (approved):
woot woot!

PR #59401, opened 2024-01-08
sg/msp: add 'sg msp tfc graph' for resources

bobheadxi:
We likely need architecture diagrams eventually, especially for SOC2, and I thought it might be good to explore what we can generate, since MSP infra architecture is a bit conditional on service specification, documenting by hand can prove rather difficult, outside of `sg msp ops`.

I tried walking the `cdktf.TerraformStack` graph, but couldn't figure out how to get dependencies correctly. In the end @michaellzc pointed me to `terraform graph`, which uses TF plans to prepare a graph. It includes stuff that doesn't feel very important, so I added a bunch of crude filtering to make the graph a bit more usable.

The layout is not _great_ - I tried the various [dot layout engines](https://graphviz.org/docs/layouts/) but none of them worked very well for our use case - but the information is actually kind of useful, and does illustrate a realistic graph of the various pieces involved.

A default rendering of the graph is available with `sg msp tfc graph`, and you can get the dot-format configuration with `sg msp tfc graph -dot`.

Part of https://github.com/sourcegraph/managed-services/issues/361 and https://github.com/sourcegraph/managed-services/issues/328

## Test plan

image

### DIFF
--- a/dev/sg/msp/BUILD.bazel
+++ b/dev/sg/msp/BUILD.bazel
@@ -16,6 +16,9 @@ go_library(
         "//dev/managedservicesplatform/stacks",
         "//dev/managedservicesplatform/stacks/cloudrun",
         "//dev/managedservicesplatform/stacks/iam",
+        "//dev/managedservicesplatform/stacks/monitoring",
+        "//dev/managedservicesplatform/stacks/project",
+        "//dev/managedservicesplatform/stacks/tfcworkspaces",
         "//dev/managedservicesplatform/terraformcloud",
         "//dev/sg/cloudsqlproxy",
         "//dev/sg/internal/category",
@@ -29,6 +32,7 @@ go_library(
         "//lib/errors",
         "//lib/output",
         "//lib/pointers",
+        "@com_github_sourcegraph_run//:run",
         "@com_github_urfave_cli_v2//:cli",
     ],
 )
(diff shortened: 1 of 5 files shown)


unknwon:
The diagram in https://docs.sourcegraph.com/dev/background-information/architecture#diagram looks like is also using dot (https://github.com/sourcegraph/sourcegraph/blob/main/doc/dev/background-information/architecture/architecture.dot)? That one looks substantially more readable 🤔

bobheadxi:
> The diagram in https://docs.sourcegraph.com/dev/background-information/architecture#diagram looks like is also using dot (https://github.com/sourcegraph/sourcegraph/blob/main/doc/dev/background-information/architecture/architecture.dot)? That one looks substantially more readable 🤔

@unknwon  yeah, those are hand-crafted and this is generated, so that looks a lot better. Overall I think this is a useful in-depth reference and probably requires some guidance to use well - if it doesn't work for security purposes then I'll try and hand-draw a few (maybe in `sg msp operations`), but even this could be useful for analyzing generated CDKTF

bobheadxi:
As per PR description I spent probably too much time with the layouts and preprocessing of the generated graph 😆 including a tool called `unflatten` to try and make the graph less spread out, but in the end they all looked kind of bad, so maybe not worth investing too much in this direction

unknwon (approved):
Approve as it's working per se, maybe a good hackathon project to make the rendering more modern/beautiful/readable (i.e. power point quality)

PR #21310, opened 2021-05-24
Remove deprecated icon field

camdencheek:
The icon field was deprecated two releases ago, and should be safe to
remove for the next release.

Closes #19892

PR #59402, opened 2024-01-08
Chore: clean up `RepositoryResolver`

camdencheek:
For reasons I don't fully understand even though it's probably my doing, `RepositoryResolver` depends on the search result type `RepoMatch`. There was some really convoluted logic about how only the `name` and `id` fields of `RepoMatch` were set, which is particularly confusing because a `RepoMatch` isn't even required to construct the `RepositoryResolver`. The only information being provided by `RepoMatch` was `rev`, which it really shouldn't be since a repository domain object should not even have a rev (that should be relegated to the `GitCommitResolver` or `GitRefResolver`, which contain a `RepositoryResolver`). 

Anyways, this removes the dependency on the search result type.

## Test plan

CI.

### DIFF
--- a/cmd/frontend/graphqlbackend/repository.go
+++ b/cmd/frontend/graphqlbackend/repository.go
@@ -28,7 +28,6 @@ import (
 	"github.com/sourcegraph/sourcegraph/internal/gitserver/gitdomain"
 	"github.com/sourcegraph/sourcegraph/internal/gitserver/protocol"
 	"github.com/sourcegraph/sourcegraph/internal/gqlutil"
-	"github.com/sourcegraph/sourcegraph/internal/search/result"
 	"github.com/sourcegraph/sourcegraph/internal/trace"
 	"github.com/sourcegraph/sourcegraph/internal/types"
 	"github.com/sourcegraph/sourcegraph/lib/errors"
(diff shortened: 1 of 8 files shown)


internal/cmd/search-blitz/search.graphql
```diff
@@ -73,9 +73,6 @@ fragment RepositoryFields on Repository {
         serviceType
         url
     }
-    label {
```
eseliger, line 76: Was this unused?

cmd/frontend/graphqlbackend/repository.go
```diff
@@ -424,18 +416,8 @@ func (r *RepositoryResolver) ExternalURLs(ctx context.Context) ([]*externallink.
 	return externallink.Repository(ctx, r.db, repo)
 }
 
-func (r *RepositoryResolver) Rev() string {
-	return r.RepoMatch.Rev
-}
-
 func (r *RepositoryResolver) Label() (Markdown, error) {
```
eseliger, line 419: What is the purpose of this resolver now? 😅 is it used anywhere? Feels weird to have a method to get a HTML anchor to the repo

eseliger (approved):
Nice! That was definitely a strange pattern

internal/cmd/search-blitz/search.graphql
camdencheek: Yeah, actually most of the fields in this request are unused. Basically [just the match count](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@fd32368e195b30baf14210307641f9ae680ddf12/-/blob/internal/cmd/search-blitz/client.go?L52-52) is extracted.

cmd/frontend/graphqlbackend/repository.go
camdencheek: AFAIK, it's not used anywhere. But it is a public part of our API, so we'd need to deprecate it, especially since I think some of our example queries used it. 

>  Feels weird to have a method to get a HTML anchor to the repo

Yes, definitely. This used to be a much more common pattern. We used to send HTML snippets of lots of things. Check out the [base-64-encoded svg](https://github.com/sourcegraph/sourcegraph/pull/21310/files) we used to send on every commit search result 😬

camdencheek: Will deprecate in a followup PR

PR #59405, opened 2024-01-08
Chore: clean up `RepositoryResolver`, part II

camdencheek:
Stacked on https://github.com/sourcegraph/sourcegraph/pull/59402

This PR does a few things to further clean up `RepositoryResolver`: 
1) It splits `NewMinimalRepositoryResolver` out of `NewRepositoryResolver` to make it more obvious when we're creating a fully-resolved `RepositoryResolver` and a lazy `RepositoryResolver`. It's pretty error prone to pass in a `*types.Repo` with only the `Name` and `ID` fields set.
2) It moves the always-available fields `ID` and `Name` directly onto the struct to make it more clear that those are safe to use directly
3) It refactors the lazy hydration to the more standard `sync.Once` pattern
4) It changes a bunch of call sites to use `api.RepoName` instead of `*types.Repo` when a fully-resolved repo is not actually needed (removing a few DB calls in the process)

## Test plan

CI. I'll also be keeping a close eye out for panic logs in prod afterwards, because touching this stuff is always bit risky.

### DIFF
--- a/cmd/frontend/backend/repos.go
+++ b/cmd/frontend/backend/repos.go
@@ -33,7 +33,7 @@ type ReposService interface {
 	GetByName(ctx context.Context, name api.RepoName) (*types.Repo, error)
 	List(ctx context.Context, opt database.ReposListOptions) ([]*types.Repo, error)
 	ListIndexable(ctx context.Context) ([]types.MinimalRepo, error)
-	GetInventory(ctx context.Context, repo api.RepoName, commitID api.CommitID, forceEnhancedLanguageDetection bool) (*inventory.Inventory, error)
+	GetInventory(ctx context.Context, repoName api.RepoName, commitID api.CommitID, forceEnhancedLanguageDetection bool) (*inventory.Inventory, error)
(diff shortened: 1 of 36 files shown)


cmd/frontend/backend/repos.go
```diff
@@ -33,10 +33,10 @@ type ReposService interface {
 	GetByName(ctx context.Context, name api.RepoName) (*types.Repo, error)
 	List(ctx context.Context, opt database.ReposListOptions) ([]*types.Repo, error)
 	ListIndexable(ctx context.Context) ([]types.MinimalRepo, error)
-	GetInventory(ctx context.Context, repo *types.Repo, commitID api.CommitID, forceEnhancedLanguageDetection bool) (*inventory.Inventory, error)
+	GetInventory(ctx context.Context, repo api.RepoName, commitID api.CommitID, forceEnhancedLanguageDetection bool) (*inventory.Inventory, error)
```
vdavid, line 36: Nit: I think using the argument name `repoName` would increase readability after this change.

camdencheek: good call, thanks!

PR #59408, opened 2024-01-09
Adding Hubspot Load Script Resiliency

arafatkatze:
Solves https://github.com/sourcegraph/cody/issues/2557 

image
image

## Test plan
Trying in SG setup


thenamankumar (changes requested):
What are we going to show in the UI if the Hubspot form fails to open? Will users move to the next step automatically? I don't see any such changes being made in the PR.

It will help the reviewers understand the changes in the flow a bit more if you can attach a Loom video. 

It will be also helpful if you can write a descriptive test plan.

arafatkatze:
@thenamankumar makes sense will open a new PR

PR #59411, opened 2024-01-09
Refactor calcUnrestricted function

### LINKED ISSUE #59412: Refactor calcUnrestricted a bit
https://github.com/sourcegraph/sourcegraph/pull/58795#discussion_r1417030865

pjlast:
Closes #59412 

Refactors the calcUnrestricted function so that the logic is a bit more clear.

## Test plan

Unit tests added to ensure refactor is safe

### DIFF
--- a/internal/database/external_services_test.go
+++ b/internal/database/external_services_test.go
@@ -2789,3 +2789,48 @@ func TestExternalServices_CleanupSyncJobs(t *testing.T) {
 		},
 	}, syncJobs)
 }
+
+func TestCalcUnrestricted(t *testing.T) {
+	// Separate test for dotcom mode to test a mix of cases
+	t.Run("dotcom mode always returns false", func(t *testing.T) {
+		dotComMode := envvar.SourcegraphDotComMode()
+		envvar.MockSourcegraphDotComMode(true)
+		defer envvar.MockSourcegraphDotComMode(dotComMode)
+
+		require.False(t, calcUnrestricted(""))
+		require.False(t, calcUnrestricted(`{"authorization": {}}`))
+		require.False(t, calcUnrestricted(`{"authorization": {}, "enforcePermissions": false}`))
+	})
+
+	otherTests := map[string]struct {
+		authorization          bool
+		enforcePermissions     bool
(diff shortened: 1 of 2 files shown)


internal/database/external_services_test.go
```diff
@@ -2789,3 +2789,48 @@ func TestExternalServices_CleanupSyncJobs(t *testing.T) {
...
+			defer envvar.MockSourcegraphDotComMode(dotComMode)
+
+			var vals []string
+			if test.authorization {
+				vals = append(vals, `"authorization": {}`)
+			}
+			if test.enforcePermissions {
+				vals = append(vals, `"enforcePermissions": true`)
+			}
+			globals.SetPermissionsUserMapping(&schema.PermissionsUserMapping{Enabled: test.permissionsUserMapping})
```
eseliger, line 2830: should this be unset after the test?

internal/database/external_services_test.go
```diff
@@ -2789,3 +2789,48 @@ func TestExternalServices_CleanupSyncJobs(t *testing.T) {
...
+	// Separate test for dotcom mode to test a mix of cases
+	t.Run("dotcom mode always returns false", func(t *testing.T) {
+		dotComMode := envvar.SourcegraphDotComMode()
+		envvar.MockSourcegraphDotComMode(true)
+		defer envvar.MockSourcegraphDotComMode(dotComMode)
+
+		require.False(t, calcUnrestricted(""))
+		require.False(t, calcUnrestricted(`{"authorization": {}}`))
+		require.False(t, calcUnrestricted(`{"authorization": {}, "enforcePermissions": false}`))
+	})
```
eseliger, line 2803: I see this is how it was before, but would we ever want to have non-unrestricted repos on dotcom now that we don't have private code there anymore? 🤔

eseliger (approved):

internal/database/external_services_test.go
pjlast: Keeping `calcUnrestricted` like this ensures that no-one accidentally adds private code. i.e. the repos need to be marked as public to be visible

Issue #57399, opened 2023-10-05
Implement clear for MATLAB local code navigation


SuperAuguste:
See https://www.mathworks.com/help/matlab/ref/clear.html

PR #59415, opened 2024-01-09
Improves locals queries for MATLAB

kritzcreek:
Fixes the MATLAB specific part of https://github.com/sourcegraph/sourcegraph/issues/57791

Removes a couple non-local occurrences from being captured. Just excluding the first function as a non-local definition is a bit cute, but it matches how MATLAB's "module system" works. WDYT @varungandhi-src? For users that are just working on individual scripts it might be a bit surprising that Sourcegraph works differently on the first function in their script.

## Test plan

Compare the snapshot test output


docker-images/syntax-highlighter/crates/scip-syntax/queries/matlab/scip-locals.scm
```diff
@@ -14,6 +11,18 @@
  (identifier) @definition.var
  (#set! "def_ref"))
 
+;; Matlab exports the _first_ function from a module as a non-local,
+;; which this query matches using tree-sitter's anchor syntax (the
+;; dot).
+(source_file . (function_definition
+ name: (identifier) @occurrence.skip))
```
varungandhi-src, line 18: It would be helpful to have (a) a citation to some official source describing the semantics OR (b) a link to a code example where this is being used OR (c) a small code example that someone can paste elsewhere to test this out.

varungandhi-src:
> Removes a couple non-local occurrences from being captured. Just excluding the first function as a non-local definition is a bit cute, but it matches how Matlab's "module system" works. WDYT @varungandhi-src? For users that are just working on individual scripts it might be a bit surprising that Sourcegraph works differently on the first function in their script.

My rule of thumb is that we should match the language semantics as closely as possible, in the absence of explicit direction from users that they actually want behavior which doesn't quite match the language semantics.

Since we don't have the latter here, let's err on the side of being faithful to the semantics.

docker-images/syntax-highlighter/crates/scip-syntax/queries/matlab/scip-locals.scm
```diff
@@ -14,6 +11,18 @@
...
+
+;; Properties and methods on classes are non-local occurrences
```
varungandhi-src, line 20:
```suggestion
```

Comment seems to be repeating the content of the query.

docker-images/syntax-highlighter/crates/scip-syntax/queries/matlab/scip-locals.scm
```diff
@@ -32,4 +41,9 @@
 
 (lambda (arguments (identifier) @definition.term))
 
+;; Neither class names nor field accesses are local references
```
varungandhi-src, line 44:
```suggestion
```

Comment seems to be repeating the content of the query.

docker-images/syntax-highlighter/crates/scip-syntax/queries/matlab/scip-locals.scm
```diff
@@ -14,6 +11,18 @@
  (identifier) @definition.var
  (#set! "def_ref"))
 
+;; Matlab exports the _first_ function from a module as a non-local,
```
varungandhi-src, line 14:
```suggestion
;; MATLAB exports the _first_ function from a module as a non-local,
```

docker-images/syntax-highlighter/crates/scip-syntax/src/snapshots/scip_syntax__locals__test__matlab_occurrences.snap
```diff
@@ -2,141 +2,138 @@
...
-//    ^ definition local 3
+//    ^ definition local 1
       f = 6;
-//    ^ definition local 4
+//    ^ definition local 2
   end
   
+  a = 1;
+//^ definition local 3
+
```
varungandhi-src, line 14: Is marking this as a local correct or an unfixed bug? Are top-level variables not accessible to other files like Python?

docker-images/syntax-highlighter/crates/scip-syntax/src/snapshots/scip_syntax__locals__test__matlab_occurrences.snap
```diff
@@ -2,141 +2,138 @@
...
-//        ^ definition local 8
+//        ^ definition local 7
       end
   end
   
   global i j
-//       ^ definition local 9
-//         ^ definition local 10
+//       ^ definition local 8
+//         ^ definition local 9
```
varungandhi-src, line 29: We should also skip these for locals.

varungandhi-src (commented):
I'm a little confused by the PR description:

> Fixes the MATLAB specific part of ...
>
> Removes a couple non-local occurrences from being captured.

Is this PR meant to fix _all_ known issues with MATLAB locals? Right now, it looks like some cases still need handling. If those are harder to do for some reason, and you want to split that into a separate PR, that's fine too (but please update the PR description to reflect that).

docker-images/syntax-highlighter/crates/scip-syntax/src/snapshots/scip_syntax__locals__test__matlab_occurrences.snap
kritzcreek: I think making a variable accessible from anywhere outside this file, means it needs to be declared `global` which is not a feature we can cover with our current mechanisms.

From what I can tell by looking at a couple repositories on Github anything that is supposed to be imported from somewhere else declares exactly one function per file. Classes are rarely used, but when they are, every file declares exactly one class and nothing else.

- https://github.com/PRML/PRMLT
- https://github.com/aludnam/MATLAB/
- https://github.com/ZJULearning/MatlabFunc

The only place where top-level variables are used, is in script files that aren't meant to be used anywhere else, e.g.: https://github.com/PRML/PRMLT/tree/master/demo

docker-images/syntax-highlighter/crates/scip-syntax/src/snapshots/scip_syntax__locals__test__matlab_occurrences.snap
kritzcreek: While we can skip these, that just means the references to them further down become definitions, which is a downgrade IMO.

kritzcreek (commented):
> Is this PR meant to fix all known issues with MATLAB locals? Right now, it looks like some cases still need handling. If those are harder to do for some reason, and you want to split that into a separate PR, that's fine too (but please update the PR description to reflect that).

It does fix all issues known _to me_. It sounds like you know of more issues I'm not covering in here. Could you write the ones down you immediately know of?

docker-images/syntax-highlighter/crates/scip-syntax/queries/matlab/scip-locals.scm
kritzcreek: It's 'defined' in the Anchors section in here: https://tree-sitter.github.io/tree-sitter/using-parsers

I wouldn't want to directly link it from here, because I don't trust the tree-sitter documentation to not move around/restructure. The most likely way to find the relevant documentation is to Google for 'tree sitter anchor' (It's how I find it, whenever I try to remember how it works)

varungandhi-src: Sorry, I should've been clearer. I was referring to the unusual MATLAB semantics, not the tree-sitter grammar syntax.

docker-images/syntax-highlighter/crates/scip-syntax/src/snapshots/scip_syntax__locals__test__matlab_occurrences.snap
varungandhi-src: > While we can skip these, that just means the references to them further down become definitions, which is a downgrade IMO.

Can we fix both of these then, by making some modifications to the Rust code/adding a new kind of annotation? Python also has a `global` annotation.

varungandhi-src:
> It does fix all issues known to me. It sounds like you know of more issues I'm not covering in here. Could you write the ones down you immediately know of?

Only this one for now: https://github.com/sourcegraph/sourcegraph/pull/59415/files/4090b981162ecc679ba0c95c590a8657600f3b4c#r1445974965

There is also https://github.com/sourcegraph/sourcegraph/issues/57399, but let's not handle that right now, it seems much more complex to support and it's unclear if it is worth supporting without user feedback.

docker-images/syntax-highlighter/crates/scip-syntax/src/snapshots/scip_syntax__locals__test__matlab_occurrences.snap
varungandhi-src: It seems to me that the MATLAB code example is similar to the following Python code in terms of scoping (ignoring the special behavior of the first function in MATLAB)

```
def __implicit_top_level__():
  a = 10 # put inside a function because it's a proper local in MATLAB
  
  def fun():
    nonlocal a # so that the first '=' is interpreted as a ref instead of a def to mimic MATLAB
    a = 20
    def nested():
       global a
       a = 30
```

So if we were able to support both `nonlocal` and `global` in Python (it would be nice to be able to do that, unless there is a compelling reason otherwise), then it seems like we should also be able to support the MATLAB `global` keyword via "desugaring" (not literally implemented like that, but based on the same core mechanisms/annotations).

Does that make sense?

kritzcreek: I have some thoughts on how to implement nonlocal/global for Python. Maybe something similar to`hoist` but call it `link` instead and have it create a definition that acts as a link/proxy to some other definition.

But right now I think that's out of scope for this PR. I'd say we design that feature for Python and keep Matlab in the back of our mind while doing so.

varungandhi-src (approved):

PR #59420, opened 2024-01-09
Check for nodes length before access

### LINKED ISSUE #59421: Error while reading 'commit' on Perforce repo
https://sourcegraph.sourcegraph.com/perforce-sgdev-org/go

### LINKED ISSUE #59422: Investigate missing history nodes on perforce depots
Follow-up to https://github.com/sourcegraph/sourcegraph/pull/59420

Figure out if we should be concerned about this

pjlast:
Closes #59421 
Closes #59422 

At the moment, access to Perforce depots can be broken, as seen on S2: https://sourcegraph.sourcegraph.com/perforce-sgdev-org/go

image

This PR resolves the issue by checking the array length of the history nodes before indexing it. However, there might be a different underlying issue here, since I'm not sure if these nodes should ever be empty.

## Test plan

Tested on my local instance


eseliger (approved):

eseliger:
> However, there might be a different underlying issue here, since I'm not sure if these nodes should ever be empty.

Can you take a look at that, or file a ticket for us?

pjlast:
@eseliger yep will do!

ggilmore (approved):

PR #59437, opened 2024-01-09
Filters panel: fix label for commit dates

camdencheek:
This just fixes the string label for commit date filters. It didn't match what we expected in the client, so that panel wasn't showing.

## Test plan

Quick manual test


jasonhawkharris (approved):
Nice!

PR #59438, opened 2024-01-09
chore: remove unused bundle diff reporter

Strum355:
Referenced an npm package thats no longer in the package.json closure, so this isnt runnable in its current state. 

From Valery: 
> This dependency [was used](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@aa6f683b6f86577a8ddf5287372fa26c22dfa81e/-/blob/client/web/scripts/report-bundle-diff.ts?L24) to generate detailed bundlesize diff reports for pull requests. It [does not](https://github.com/statoscope/statoscope/issues/148) support esbuild or vite yet, so we cannot use it with the new build infra. I'm not aware of any migration plans.

## Test plan

Checked for any references (none)


valerybugakov (approved):
It would great to have some kind of a replacement for it in the future 👍

PR #59445, opened 2024-01-09
Filters: update repo rendering

camdencheek:
This updates the filters panel to show only the `org/name` from the repo, removing the code host from the text by default and replacing it with a representative icon if it is a known codehost. Additionally, it adds a hover tooltip that shows the full repo name (including code host) so that a user can still see what repo it is even if it's a long repo name. 

In an ideal world, we would fetch the code host kind from the backend and use that to render the icon. However, it's non-trivial to add followup requests and caching here, so I'm just copying what we do elsewhere in the UI. So this [only renders special icons](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@6b1979f7b6267b67d72a744ca8048bdc06e1c3e3/-/blob/client/branded/src/search-ui/components/CodeHostIcon.tsx?L8-10) for `github.com`, `bitbucket.org` and `gitlab.com`, falling back to the generic icon.

Fixes https://github.com/sourcegraph/sourcegraph/issues/59444
Fixes https://github.com/sourcegraph/sourcegraph/issues/59446

Before:

CleanShot 2024-01-09 at 11 02 57@2x

After:

CleanShot 2024-01-09 at 11 23 27@2x

## Test plan

Manual testing.

### DIFF
--- a/client/branded/src/search-ui/results/filters/components/dynamic-filter/SearchDynamicFilter.tsx
+++ b/client/branded/src/search-ui/results/filters/components/dynamic-filter/SearchDynamicFilter.tsx
@@ -8,7 +8,7 @@ import { UserAvatar } from '@sourcegraph/shared/src/components/UserAvatar'
 import type { Filter } from '@sourcegraph/shared/src/search/stream'
 import { useExperimentalFeatures } from '@sourcegraph/shared/src/settings/settings'
 import { SymbolKind } from '@sourcegraph/shared/src/symbols/SymbolKind'
-import { Badge, Button, Icon, H4, Input, LanguageIcon, Code } from '@sourcegraph/wildcard'
+import { Badge, Button, Icon, H4, Input, LanguageIcon, Code, Tooltip } from '@sourcegraph/wildcard'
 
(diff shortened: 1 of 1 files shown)


camdencheek:
One thing I couldn't figure out is that our tooltips seem to have a fixed width that's pretty small, so long repo names tend to wrap in the hover text. That's a pretty low-priority issue, and I figure we can fix that in a followup.

BolajiOlajide (approved):

vovakulikov:
> One thing I couldn't figure out is that our tooltips seem to have a fixed width that's pretty small
@camdencheek we don't have fixed width in tooltips I think the culprit here is scrollable sidebar block where filters are rendered, tooltips are rendered outside of main body DOM tree but by default they try to fit in the closest scroll container of the element to which you assign tooltip. And this is what happening here I think, we have a prop in Popover API (which tooltip is based) which can change this behaviour and tooltip won't respect filters panel scroll block in size calculations. I can take a look on this

camdencheek:
Aha! That makes sense. Thanks @vovakulikov!

PR #59448, opened 2024-01-09
Perf: cache linker on `RepositoryResolver`

camdencheek:
Currently, repository links are calculated on the fly. The problem with this is it requires fetching the default branch and whether the repo is a perforce repo. We currently do this once per commit when we are listing commits. 

Instead, this moves the state into a new `RepositoryLinker` struct that can be cached on the `RepositoryResolver` and reused rather than being recalculated repeatedly. 

For fetching the commit history of one file, this reduced the number of spans from 628 to 264. These are all tiny spans, so it doesn't massively improve the latency, but it should help reduce load on the DB. 

## Test plan

CI and some manual testing. Existing unit tests have good coverage for this section.


unknwon (approved):
Not a very careful review.

PR #59453, opened 2024-01-09
Comments, Cleanups, Tests for `naman/reset-cody-usage-cache`

chrsmith:
This PR adds some comments, minor cleanups, and adds tests to https://github.com/sourcegraph/sourcegraph/pull/59330.

But in general, this PR:

- Adds some more comments to clarify the interactions of various limiters. e.g. making `(a *Actor) Limiter(...)` easier to follow by moving the 3x chained limiters into local variables.
- Move the logic for resetting a user's usage cache into a dedicated `(s *Source) maybeResetUsageData(...)`
- Add a unit test verifying the side effects of calling `dotcomuser.Store::get(...)`.
- Add a new mock/fake for `limiter.RedisStore` that records the history of operations. (So we can confirm the specific calls to `Del(key)` as needed.)

@thenamankumar , feel free to accept/reject any part of this PR as you think makes sense. Hopefully you think it is a strict improvement. But some changes (e.g. `s/i.e./e.g./` are definitely unimportant things that I care more about than I ought. Similarly, feel free to push back if you think the extra comments are unnecessary or excessive!

## Test plan

Added new tests


cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
```diff
@@ -130,6 +117,43 @@ func (s *Source) fetchAndCache(ctx context.Context, token string, oldAct *actor.
...
+	for _, feature := range codygateway.AllFeatures {
+		rl, ok := current.RateLimits[feature]
+		if !ok {
+			continue
+		}
+
+		// The actual RateLimit will be a zero value in the event of an error reading the
+		// Actor from dotcom. In these cases, rl.IsValid() is false and so we just ignore
+		// resetting the usage data to avoid cascading failure.
+		if !rl.IsValid() {
```
chrsmith, line 133: This `if`-statement is the only new-new thing I added on top of @thenamankumar 's PR.

I encountered this when running the unit tests:
- If we refetch the user from dotcom and get an error, we initialize a stubbed out actor via `newActor(s, token, dotcom.DotcomUserState{}, s.concurrencyConfig)`.
- However, the `RateLimits` associated with that actor are all invalid. (Since we default to a limit of 0, no allowable LLMs, etc.)
- So in this case, the call to `l, ok := current.BaseLimiter(...)` will always fail, because we refuse to create a limiter for an invalid rate limit.

I don't know what the right thing to do here is.

We could change the definition of a "valid" rate limit to include a limit of 0? Or, like I have here, if the rate limit is invalid, we don't bother with resetting the usage limit. (Since something is clearly wrong from the get-go.)

cmd/cody-gateway/internal/actor/actor.go
```diff
@@ -222,24 +222,35 @@ func (a *Actor) Limiter(
 	feature codygateway.Feature,
 	rateLimitNotifier notify.RateLimitNotifier,
 ) (limiter.Limiter, bool) {
+	// Start with the base limiter. For most users, this is governed by the app's configuration.
+	// But for users on the Cody Pro plan, their limits may be different.
```
bobheadxi, line 226: Does this comment apply at this level? We should avoid anything Cody-Pro specific as this is the shared Actor abstraction

cmd/cody-gateway/internal/actor/actor.go
```diff
@@ -222,24 +222,35 @@ func (a *Actor) Limiter(
...
+		logger:      logger.Scoped("updateOnError"),
+		actor:       a,
+		nextLimiter: baseLimiter,
+	}
+
+	// Finally return a concurrency limiter, to ensure that a user cannot have too many
+	// requests in-flight at a time. Note that we are setting the user's limits based on
+	// what we got from baseLimiterAndLimit, as to honor a Cody Pro subscription as
+	// applicable.
+	concurrentStorePrefix := fmt.Sprintf("concurrent:%s", featurePrefix(feature))
```
bobheadxi, line 244: Likewise

cmd/cody-gateway/internal/actor/actor.go
chrsmith: It's possible the comment is totally wrong and I don't fully understand the code. But if it is factually correct, I don't think _mentioning_ Cody Pro would be a problem.

I agree that we don't want Cody Gateway to have any awareness of "Cody Pro" or "Cody Free" etc. And that any rate limits are just opaque values that come from the Actor source. (i.e. it is dotcom that knows if someone is on Cody-Pro or not, and then returns different values for Actor.RateLimits, etc.)

But as far as mentioning it in comments like here, IMHO it's helpful to call out why we have the base limiter and/or why rate limits would need to be refreshed at different intervals for different users.

bobheadxi: Reading this again, I find it pretty confusing IMO - a better one would be:

```suggestion
	// Start with the base limiter. This limiter enforces rate limits provided
	// the actor's Source, and each Source implementation owns ensuring that
	// Cody Gateway has an up-to-date view of the appropriate rate limits for
	// a particular actor based on the application (e.g. users from enterprise product
	// subscriptions, Self-Serve-Cody's tiers, etc.
```

bobheadxi: The comment makes it sound like we have 2 particular type of limiters and rate limits, when in reality, Cody Gateway just wants to enforce whatever rate limits the Source of an Actor says the Actor has. It doesn't care about plans or why limits are different or why they change

cmd/cody-gateway/internal/actor/actor.go
bobheadxi:
```suggestion
	// Finally return a concurrency limiter, to ensure that a user cannot have too many
	// requests in-flight at a time. This is generally a percentage of the rate limit
	// assigned to an Actor by its Source - see RateLimit for more details.
	concurrentStorePrefix := fmt.Sprintf("concurrent:%s", featurePrefix(feature))
```

cmd/cody-gateway/internal/actor/actor.go
chrsmith: I believe we have both participated in [Cunningham's Law](https://en.wikipedia.org/wiki/Ward_Cunningham#%22Cunningham's_Law%22)-driven development. 🙃😭

cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
```diff
@@ -130,6 +117,43 @@ func (s *Source) fetchAndCache(ctx context.Context, token string, oldAct *actor.
...
+		if oldAct.IsEmpty() || rl.Interval != oldAct.RateLimits[feature].Interval {
+			// Get the base limiter for the feature which implements the core rate limiting based on the config.
+			l, ok := current.BaseLimiter(s.redisStore, feature, s.rateLimitNotifier)
+			if !ok {
+				return errors.Errorf("failed to create base limiter for feature: %s and actor id: %s", feature, current.ID)
+			}
+			if err := l.ResetUsage(); err != nil {
+				message := fmt.Sprintf("failed to reset usage cache for feature: %s and actor id: %s", feature, current.ID)
+				return errors.Wrap(err, message)
+			}
```
bobheadxi, line 150: Up to you, but maybe we can apply my suggestion here? https://github.com/sourcegraph/sourcegraph/pull/59330#discussion_r1446551170

1. Encode the standard that the usage store ID === actor ID (already the case practically)
2. Make manipulation of usage a first-class citizen of Source
3. Remove `BaseLimiter` leaky abstraction
4. Remove `rateLimitNotifier` from Source

chrsmith: All of those sound like reasonable things to do. But given that this PR is in service of solving a user-facing bug, I'm inclined to try to keep refactoring to a minimum.

But if those suggestions make sense to Rafal and Naman, I'd be happy to send out a PR later to do that. Or try and apply any other suggestions to streamline this part of the code.

cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
```diff
@@ -82,41 +82,28 @@ func (s *Source) Update(ctx context.Context, act *actor.Actor) error {
...
 	if checkErr != nil {
-		// Generate a stateless actor and save in the cache so that we aren't constantly hitting the dotcom API
+		// Generate a stateless actor and save in the cache so that we aren't constantly hitting the dotcom API.
 		act = newActor(s, token, dotcom.DotcomUserState{}, s.concurrencyConfig)
 	} else {
 		act = newActor(s, token,
 			resp.Dotcom.CodyGatewayDotcomUserByToken.DotcomUserState, s.concurrencyConfig)
 	}
 
+	// Marshall the actor into JSON so we can persist it.
```
rafax, line 99:
```suggestion
	// Marshal the actor into JSON so we can persist it.
```

rafax (approved):
I still think we need to address [this](https://github.com/sourcegraph/sourcegraph/pull/59330#discussion_r1442799857) and [this](https://github.com/sourcegraph/sourcegraph/pull/59453/files#r1446764596) - if you prefer to do it as a fast-follow PR, that's fine with me.

chrsmith:
@rafax, @thenamankumar , I'll take care of addressing those suggested refactoring in a follow-up PR this week. Thank you both for pushing through and getting the underlying bug fixed!

PR #59454, opened 2024-01-09
New Filters Panel: replace 'show all' logic with 'show more/less' logic

jasonhawkharris:
Before, filters with many results would show either all, or no filters. Now, the `show more [filter]s` button adds 5 at a time until `allFiltersAreDisplayed`, at which point the button will change to a `Show less [filter]s`, collapsing back to 5 when clicked. 

https://github.com/sourcegraph/sourcegraph/assets/62355966/5a2cce74-1707-470d-b171-006cddaee081

## Test plan
CI/CD should be enough.

### DIFF
--- a/client/branded/src/search-ui/results/filters/components/dynamic-filter/SearchDynamicFilter.tsx
+++ b/client/branded/src/search-ui/results/filters/components/dynamic-filter/SearchDynamicFilter.tsx
@@ -95,7 +95,7 @@ export const SearchDynamicFilter: FC<SearchDynamicFilterProps> = ({
     const filtersToShow =
         mergedFilters.length <= visibleFilters ? filteredFilters : filteredFilters.slice(0, visibleFilters)
 
-    const moreOrLessFilters = () => {
+    const moreOrLessFilters = (): void => {
         const filtersDisplayed = visibleFilters + MAX_FILTERS_NUMBER
 
         if (allFiltersDisplayed) {
--- a/client/branded/src/search-ui/results/filters/components/dynamic-filter/SearchDynamicFilter.tsx
+++ b/client/branded/src/search-ui/results/filters/components/dynamic-filter/SearchDynamicFilter.tsx
@@ -15,7 +15,7 @@ import { URLQueryFilter } from '../../hooks'
 
 import styles from './SearchDynamicFilter.module.scss'
 
-const MAX_FILTERS_NUMBER = 7
+const MAX_FILTERS_NUMBER = 5
 
 interface SearchDynamicFilterProps {
     /** Name title of the filter section */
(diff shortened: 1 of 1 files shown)


client/branded/src/search-ui/results/filters/components/dynamic-filter/SearchDynamicFilter.tsx
```diff
@@ -91,7 +92,26 @@ export const SearchDynamicFilter: FC<SearchDynamicFilterProps> = ({
 
     const lowerSearchTerm = searchTerm.toLowerCase()
     const filteredFilters = mergedFilters.filter(filter => filter.label.toLowerCase().includes(lowerSearchTerm))
-    const filtersToShow = showAllFilters ? filteredFilters : filteredFilters.slice(0, MAX_FILTERS_NUMBER)
+    const filtersToShow =
+        mergedFilters.length <= visibleFilters ? filteredFilters : filteredFilters.slice(0, visibleFilters)
```
fkling, line 96: Should this be `filteredFilters.length` instead of `mergedFilters.length`?

fkling (commented):
Three cases that come to mind that should be tested against:

- The number of available filters is equal to or less than five. I expect the button not to be shown then.
- How does this behavior interact with filtered filters?
  - If I type into the input and less than five filters match, what happens?
  - What if more than five filters match? Do the buttons properly operate on the filtered filters only?

client/branded/src/search-ui/results/filters/components/dynamic-filter/SearchDynamicFilter.tsx
jasonhawkharris: Yup! Changing it.

client/branded/src/search-ui/results/filters/components/dynamic-filter/SearchDynamicFilter.tsx
```diff
@@ -85,13 +86,34 @@ export const SearchDynamicFilter: FC<SearchDynamicFilterProps> = ({
         }
     }
 
+    useEffect(() => {}, [searchTerm])
+
```
fkling, line 90: Remove
```suggestion
```

client/branded/src/search-ui/results/filters/components/dynamic-filter/SearchDynamicFilter.tsx
```diff
@@ -118,8 +140,8 @@ export const SearchDynamicFilter: FC<SearchDynamicFilterProps> = ({
                 ))}
             </ul>
             {filteredFilters.length > MAX_FILTERS_NUMBER && (
```
fkling, line 142: Ah, I missed that this condition existed when I posted my first comment. What threw me off was that `allFiltersDisplayed` is initialized to `false`. I wondered how that would work if there are less than five filters. Now it makes sense.

fkling (approved):

vovakulikov (commented):
I'm gonna test this manually, left a few comments about state and logic

client/branded/src/search-ui/results/filters/components/dynamic-filter/SearchDynamicFilter.tsx
```diff
@@ -85,13 +86,34 @@ export const SearchDynamicFilter: FC<SearchDynamicFilterProps> = ({
         }
     }
 
+    useEffect(() => {}, [searchTerm])
```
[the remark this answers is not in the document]
jasonhawkharris: yup! my bad.

PR #59464, opened 2024-01-10
sg: update-ops helm images continue loop when skipping image

burmudar:
When the UpdateOperation tells us to skip we return `NoUpdateNeeded`, and instead of continueing the loop we return making it that if we intend skip something we would skip it ... but also the rest of the images
## Test plan
Tested locally


jhchabran (approved):
Ah damn, good catch 🙏

Issue #59463, opened 2024-01-10
Update our tree-sitter-perl fork


kritzcreek:
Our fork at https://github.com/sourcegraph/tree-sitter-perl is trailing upstream by >60 commits. We should update it.

PR #59465, opened 2024-01-10
Improves locals queries for Perl

kritzcreek:
Fixes the Perl specific part of https://github.com/sourcegraph/sourcegraph/issues/57791

The only thing I can do at the moment is to exclude `our` bindings from locals as these are exported definitions. I've also removed tracking assignment of globals as these are, well, globals.

Unfortunately the Perl tree-sitter grammar is pretty useless for our purposes as it skips function signatures with a C lexer so we don't get a tree for the bit that contains most local definitions.

## Test plan

Snapshot tests


varungandhi-src:
> Unfortunately the Perl tree-sitter grammar is pretty useless for our purposes as it skips function signatures with a C lexer so we don't get a tree for the bit that contains most local definitions.

Could you elaborate on this with one or two code examples? I see there are `sub` declarations and locals in those are being tracked, and I don't know any Perl to be able to tell what's wrong.

Aside from that, we're using a fork of grammar, and that is already ~60 commits behind upstream. It would be valuable to have the following: (by digging yourself and/or discussing with TJ)
- A summary of what the main changes from upstream are (and can we upstream them)
- Can we incorporate the recent changes from upstream
- Take a brief look as to what it'd take to fix the shortcoming you're running into with the grammar

kritzcreek:
> Could you elaborate on this with one or two code examples? I see there are sub declarations and locals in those are being tracked, and I don't know any Perl to be able to tell what's wrong.

image

The `prototype_or_signature` node here spans the function signature and it's an external token defined in the C lexer that just skips matching parens:

https://github.com/tree-sitter-perl/tree-sitter-perl/blob/f2effa6c4a5f88e600d316f08a02666a396bcc3a/src/scanner.c#L762-L794

which means there's no node in the tree spanning  the definition of`$phrase`.

> Aside from that, we're using a fork of grammar, and that is already ~60 commits behind upstream. It would be valuable to have the following: (by digging yourself and/or discussing with TJ)

Way ahead of you https://github.com/sourcegraph/sourcegraph/issues/59463

I'll ask TJ and get some more details.

varungandhi-src:
> The prototype_or_signature node here spans the function signature and it's an external token defined in the C lexer that just skips matching parens:

So we're missing parameters, but not other function-local variables, is that right?

kritzcreek:
Yes, that's right

kritzcreek:
It also means we get things like:

```perl
my $local = 1

sub func($local) {
  say $local
}
```
wrong, because we don't notice that the local variable gets shadowed in the subroutine.

kritzcreek:
@varungandhi-src Could you stamp this one? I know it doesn't fix all known issues with the Perl locals, but it does make it so we don't report non-locals as locals, which was the main goal of this round of query PRs.

varungandhi-src (approved):

varungandhi-src:
> I know it doesn't fix all known issues with the Perl locals, but it does make it so we don't report non-locals as locals, which was the main goal of this round of query PRs

Makes sense, I kinda' got a bit distracted with the other bits, so I lost track of this.

Issue #59329, opened 2024-01-04
Search jobs: Switch from CSV to JSON


jtibshirani:
### Requirements:
  - [x] Add individual matches
  - [ ] Add checksum to header
  - [ ] Add tags (metadata, topics) as new column

### Design
- Switch from CSV to JSON
- Serve JSON in the format of the Steam API IE ChunkMatches with byte offsets

```[tasklist]
### Tasks
- [x] Update SearchQuery interface to take a (more generic) Writer interface
- [x] Implement a JSON Writer
- [x] Switch storage layer and download format to JSON
- [x] Serve JSON
```

Context:
https://sourcegraph.slack.com/archives/C05KARB1F33/p1702502762016359

PR #59468, opened 2024-01-10
search jobs: refactor SearchQuery interface

stefanhengl:
Relates to #59329 

This is mostly a refactor to prepare for a JSON writer that replaces our CSV writer.

The high-level plan is to switch the internal (object storage) AND the external (download) format from CSV to JSON.

Test plan:
- CI
- Note: I had to update `SearcherFake`, hence some of the test data changed. It only affects a few tests and the character of the tests remains the same.

### DIFF
--- a/cmd/worker/internal/search/exhaustive_search_test.go
+++ b/cmd/worker/internal/search/exhaustive_search_test.go
@@ -118,9 +118,9 @@ func TestExhaustiveSearch(t *testing.T) {
 		}
 		sort.Strings(vals)
 		require.Equal([]string{
-			"repo,revspec,revision\n1,spec,rev1\n",
-			"repo,revspec,revision\n1,spec,rev2\n",
-			"repo,revspec,revision\n2,spec,rev3\n",
+			"repository,revision,file_path,match_count,first_match_url\n1,rev1,path/to/file.go,0,/1@rev1/-/blob/path/to/file.go\n",
+			"repository,revision,file_path,match_count,first_match_url\n1,rev2,path/to/file.go,0,/1@rev2/-/blob/path/to/file.go\n",
+			"repository,revision,file_path,match_count,first_match_url\n2,rev3,path/to/file.go,0,/2@rev3/-/blob/path/to/file.go\n",
 		}, vals)
 	}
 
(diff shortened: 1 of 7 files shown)


cmd/worker/internal/search/exhaustive_search_test.go
```diff
@@ -118,9 +118,9 @@ func TestExhaustiveSearch(t *testing.T) {
 		}
 		sort.Strings(vals)
 		require.Equal([]string{
-			"repo,revspec,revision\n1,spec,rev1\n",
-			"repo,revspec,revision\n1,spec,rev2\n",
-			"repo,revspec,revision\n2,spec,rev3\n",
+			"repository,revision,file_path,match_count,first_match_url\n1,rev1,path/to/file.go,0,/1@rev1/-/blob/path/to/file.go\n",
+			"repository,revision,file_path,match_count,first_match_url\n1,rev2,path/to/file.go,0,/1@rev2/-/blob/path/to/file.go\n",
+			"repository,revision,file_path,match_count,first_match_url\n2,rev3,path/to/file.go,0,/2@rev3/-/blob/path/to/file.go\n",
```
stefanhengl, line 123: We don't have "spec" anymore, but that was anyway hardcoded and didn't add much value to the test.

internal/search/exhaustive/service/search.go
```diff
@@ -90,7 +98,12 @@ type CSVWriter interface {
 	WriteRow(...string) error
 }
 
-// NewBlobstoreCSVWriter creates a new BlobstoreCSVWriter which writes a CSV to
+type CSVWriterCloser interface {
```
jtibshirani, line 101: Do we need this new interface? Or could we just extend the `CSVWriter` interface to have a close method?

jtibshirani (approved):

keegancsmith (approved):

internal/search/exhaustive/service/search.go
stefanhengl: We can remove it and extend `CSVWriter`. [Inspired by the std lib](https://sourcegraph.sourcegraph.com/github.com/golang/go@e9b3ff15f40d6b258217b3467c662f816b078477/-/blob/src/io/io.go?L137) I try to keep the interfaces as small as possible and compose them when needed. However, in this case, it is not really necessary, because we don't expect the `CSVWriter` interface to be used anywhere else.

PR #59471, opened 2024-01-10
handle deleted users when listing team members

BolajiOlajide:
While fetching members of a team, we currently don't check if the user is deleted which leads to an error on S2 when fetching the member's information.

CleanShot 2024-01-10 at 14 42 03@2x

```graphql
query ListTeamMembers($first: Int, $after: String, $search: String, $teamName: String!) {
    team(name: $teamName) {
        id
        members(first: $first, after: $after, search: $search) {
            totalCount
            pageInfo {
                endCursor
                hasNextPage
            }
            nodes {
                ...ListTeamMemberFields
            }
        }
    }
}

fragment ListTeamMemberFields on TeamMember {
    __typename
    ... on User {
        __typename
        id
        url
        username
        displayName
        avatarURL
    }
}
```

## Test plan

* Navigate to S2
* Access the [`everyone` team on S2](https://sourcegraph.sourcegraph.com/teams/everyone/members)
* You should see the list of team members instead of the current error page on S2.

### DIFF
--- a/internal/database/teams.go
+++ b/internal/database/teams.go
@@ -320,6 +320,9 @@ func (s *teamStore) ListTeamMembers(ctx context.Context, opts ListTeamMembersOpt
 		opts.Limit++
 	}
 
+	joins = append(joins, sqlf.Sprintf("LEFT JOIN users ON team_members.user_id = users.id"))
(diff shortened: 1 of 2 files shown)


eseliger (approved):
good catch - could we get a test for that?

PR #59476, opened 2024-01-10
chore: update universal-ctags GHA workflow

Strum355:
Bringing in-line with versions from https://github.com/sourcegraph/p4-fusion/blob/master/.github/workflows/nix-build-and-upload.yaml

## Test plan

With `act` locally

### DIFF
--- a/.github/workflows/universal-ctags.yml
+++ b/.github/workflows/universal-ctags.yml
@@ -20,30 +20,28 @@ jobs:
     runs-on: macos-latest
     steps:
       - uses: actions/checkout@v3
-      - name: Install Nix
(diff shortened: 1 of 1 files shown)


.github/workflows/universal-ctags.yml
```diff
@@ -53,33 +51,31 @@ jobs:
           glob: 'universal-ctags-*'
   aarch64-darwin:
     name: Build ctags aarch64-darwin
-    runs-on: macos-latest
+    runs-on: macos-latest-xlarge
```
Strum355, line 54: Moving from cross-compiling to native building

jhchabran (approved):

burmudar (approved):

PR #59483, opened 2024-01-10
sg msp: use static list of stack names, remove --tfc=false

bobheadxi:
This change adds a static list of all workspaces we have. This is unlikely to change much more in the future. This static list can be used for:

1. Syncing Terraform Cloud workspaces (we no longer need to render the stack to do so)
2. CLI completions where appropriate

To make sure the static list holds, I've also removed the option to not use TFC as the backend.

## Test plan

`sg msp gen msp-testbed test` has no diff


unknwon (approved):
>I've also removed the option to not use TFC as the backend.

Somewhat uncomfortable but it might also simplify many things when we can assume TFC is always in place. 🤷‍♂️ 😂

jac (approved):

bobheadxi:
> Somewhat uncomfortable but it might also simplify many things when we can assume TFC is always in place. 🤷‍♂️ 😂

It makes it simpler, and also, the local apply is unlikely to work easily due to the degree of permissions you need

Issue #59308, opened 2024-01-03
Disable embeddings for all instances except dotcom


chwarwick:
Move enterprise embeddings behind a experimental feature so they can be disabled for enterprise but enabled for sourcegraph.com

PR #59493, opened 2024-01-10
Disables embeddings for all instances that are not dotcom

chwarwick:
Overrides the embeddings config to default embeddings to disabled unless the site is running in dotcom mode.  This is a first minimal step to removing embeddings, dotcom is still dependent on the embeddings config to supply PLG rate limits for local generated embeddings and the current extension versions still use remote embeddings from dotcom as an additional context source.  

When embeddings are no longer need for dotcom the code related to embeddings can be removed. 

closes https://github.com/sourcegraph/sourcegraph/issues/59308

## Test plan
unit tests updated
manual test using `sg start` - extension didn't attempt to search embeddings because they are disabled.
manual test using `sg start dotcom` - extension saw embeddings enabled for repo and searched successfully


eseliger (approved):
nice! it's gonna be great as well to be able to remove embeddings rate limiting from cody gateway, plus the management features in licensing.

PR #59498, opened 2024-01-10
fix code actions color mismatch

BolajiOlajide:
[Thread](https://sourcegraph.slack.com/archives/C05EA9KQUTA/p1704903083081249)

This fixes the color mismatch in the repo actions navbar.

CleanShot 2024-01-10 at 22 46 58@2x

## Test plan

Navigate to a file view `/github.com/sourcegraph/code-intel-extensions@525da1bc7a1dd5c068091f881c38b6b8210d3351/-/blob/schema/schema.graphql`


mrnugget:
Appreciate it!

PR #59500, opened 2024-01-10
Symbol search: clean up job logic

jtibshirani:
The search planning code uses the same `TextPatternInfo` struct for unindexed
search, structural search, and symbol search. However, most of the parameters
don't make sense for symbol search. This PR pulls out a dedicated
`SymbolSearchRequest` struct for symbol search to use instead.

This uncovered an error with symbol search where negation was allowed, but gave
the wrong results. With this change, we just return an error.

Relates to https://github.com/sourcegraph/sourcegraph/issues/59038

## Test plan

Existing tests for symbol search, as well as manual testing

### DIFF
--- a/cmd/frontend/internal/search/search.go
+++ b/cmd/frontend/internal/search/search.go
@@ -29,6 +29,7 @@ import (
 	"github.com/sourcegraph/sourcegraph/internal/lazyregexp"
 	"github.com/sourcegraph/sourcegraph/internal/search"
 	"github.com/sourcegraph/sourcegraph/internal/search/client"
+	"github.com/sourcegraph/sourcegraph/internal/search/query"
 	"github.com/sourcegraph/sourcegraph/internal/search/result"
 	"github.com/sourcegraph/sourcegraph/internal/search/streaming"
 	streamclient "github.com/sourcegraph/sourcegraph/internal/search/streaming/client"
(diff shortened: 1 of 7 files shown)


jtibshirani:
Here's an example of a negated symbol search.

**Before**
Screenshot 2024-01-10 at 3 31 17 PM

**After**
Screenshot 2024-01-10 at 3 09 29 PM

cmd/frontend/internal/search/search.go
```diff
@@ -178,6 +179,11 @@ func (h *streamHandler) serveHTTP(r *http.Request, tr trace.Trace, eventWriter *
 
 		return h.searchClient.Execute(ctx, batchedStream, inputs)
 	}()
+
+	if err != nil && errors.HasType(err, &query.UnsupportedError{}) {
```
jtibshirani, line 183: I made some changes here so that the error shows up nicely in the search UX.

As an alternative, we could try to check this earlier during query validation. I preferred returning an error during job creation, as it's the most lightweight and natural place to check this.

internal/search/job/jobutil/job.go
```diff
@@ -624,6 +629,31 @@ func mapSlice(values []string, f func(string) string) []string {
...
+func toSymbolSearchRequest(f query.Flat) (*searcher.SymbolSearchRequest, error) {
+	if f.Pattern.Negated {
+		return nil, &query.UnsupportedError{
+			Msg: "symbol search does not support negation.",
+		}
+	}
+
+	// We convert literal searches to regexes, since the symbol search logic
+	// assumes that a literal pattern is an escaped regular expression.
+	regexpPattern := f.ToBasic().PatternString()
```
jtibshirani, line 641: I simply preserved this logic from `toTextPatternInfo`. It was surprising to me that we always convert the query to a regex pattern, when we could be accepting literals directly. I filed https://github.com/sourcegraph/sourcegraph/issues/59503 suggesting to clean this up.

keegancsmith (approved):
nice!

PR #59502, opened 2024-01-10
db: Cast some slices of api.RepoID and api.RepoName to primitive slices

eseliger:
When using primitive slices, the pq implementation is more optimized, and we also get better handling in tracing for the maximum length.

As mentioned on Slack here: https://sourcegraph.slack.com/archives/C02UC4WUX1Q/p1704928801529719?thread_ts=1704913928.968849&cid=C02UC4WUX1Q, this is not an ultimate solution. It fixes it for some of our queries, but won't be a sustainable long-term solution.

Alternative ideas:
- Remove these two type aliases
- Fork the pq.Array implementation and add native support for those two

## Test plan

CI should find issues with the DB serialization.

### DIFF
--- a/internal/database/codeowners.go
+++ b/internal/database/codeowners.go
@@ -3,6 +3,7 @@ package database
 import (
 	"context"
 	"fmt"
+	"unsafe"
 
 	"github.com/jackc/pgconn"
 	"github.com/keegancsmith/sqlf"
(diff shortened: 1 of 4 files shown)


internal/database/codeowners.go
```diff
@@ -180,7 +181,15 @@ LIMIT 1
...
 		conds := []*sqlf.Query{
-			sqlf.Sprintf("repo_id = ANY (%s)", pq.Array(ids)),
+			sqlf.Sprintf(
+				"repo_id = ANY (%s)",
+				// We use this trick to convert the []api.RepoID to []int32 (the underlying
+				// type of api.RepoID). This is safe because the underlying type of api.RepoID
+				// is always int32.
+				// When passing an []api.RepoID to pq.Array directly, a more costly
+				// conversion when Value is called is required.
+				pq.Array(*(*[]int32)(unsafe.Pointer(&ids))),
```
camdencheek, line 191: I wonder if we should just create a transmute package with a safe API. It's fairly common across the codebase to have to make a new allocation for a slice even though they have the same underlying type. It'd be pretty straightforward to expose transmutes for all the builtin types.


```go
package main

import (
	"fmt"
	"unsafe"
)

func TransmuteInt32s[T, R ~int32](input []T) []R {
	return *(*[]R)(unsafe.Pointer(&input))
}

func TransmuteToInt32s[T ~int32](input []T) []int32 {
	return TransmuteInt32s[T, int32](input)
}

type RepoID int32

func main() {
	a := []RepoID{1, 2, 3, 4}
	fmt.Println(TransmuteToInt32s(a))
	/// prints [1 2 3 4]
}
```

camdencheek: tbh, seems like a pretty simple package we could make OSS and get some publicity for too 🤷

eseliger: that's interesting, and thanks for finding this draft PR 😆 

Yeah I'm all-in for making this a proper helper vs inline keeping to copy this pattern and comments around it being safe.

PR #54079, opened 2023-06-23
this one is for cesar

tjdevries:
## Test plan

- [ ] @Numbers88s gives me as many different go symbols as he wants me to look at and i make sure they are all covered

PR #59515, opened 2024-01-11
Delete unused symbols extraction

kritzcreek:
Removes code that was added in https://github.com/sourcegraph/sourcegraph/pull/54079 but never ended up used.

[Relevant Slack discussion](https://sourcegraph.slack.com/archives/C05GBLHM4P8/p1704816045788289)

## Test plan

CI should cover this


kritzcreek:
@keynmol @tjdevries could someone please take a look? Thanks

PR #59519, opened 2024-01-11
search jobs: implement JSON writer

stefanhengl:
WIP

This adds a JSON writer, which, just like the already existing CSV writer, writes chunks of 100MB files to the object storage. 
The new writer is not hooked up yet, but I added a test that covers its behavior.

## Test plan
new unit test


keegancsmith (approved):
LGTM.

One high level question? What is a better format for consumers? A line of json per match, or a single array of the matches.

My prefence is the latter since most json tools support that format (eg jq, go's json library), and it is usually easier to stream those formats. Additionally it is more greppable. You can also more easily combine the multiple shards by just catting them together.

However, I wonder what others think about the format?

jtibshirani:
I have slight preference for emitting well-formed JSON instead of JSON lines format (which is one JSON line per match, separated by a newline). JSON lines often requires special handling, you can't just call a single method in a standard library. So it can be a bit more work when scripting, etc.

On the other hand, the output is visually clean and has some benefits. But I'm guessing this format will mostly be used to help filter/ process results, so optimizing for ease of scripting makes sense?

stefanhengl:
I am leaning towards a properly formatted JSON, too. We don't know how the users will consume the API so it seems better to default to something any JSON library can handle. Personally I like the line-based format because it is a bit easier to handle on the command line.

PR #59520, opened 2024-01-11
Remove no longer used cody-autocomplete-default-starcoder-hybrid-sourcegraph feature flag

philipp-spiess:
This feature flag was only used while we were evaluating the single tenant instance. 

Going forward, we will add support for Cody Gateway to resolve the virtual model strings, but for now we translate to our current MT defaults.

## Test plan

- Trivial change, CI will test


fkling:
That must have been the longest feature flag name 😅

rafax (approved):

PR #59522, opened 2024-01-11
Add StarCoder enterprise virtual model string

philipp-spiess:
This PR adds a new `fireworks/starcoder` virtual model string for use with the licensing data (so that adding one concise `fireworks/starcoder` string allows us to give access to all needed fireworks model) and the downstream clients (so that we can later propagate this as the default code completion model from the SG instance to the Cody clients.

Detailed list of changes after this PR:

- Creation of a new `fireworks/starcoder` virtual model string
  - This is the only starcoder related model string that is now rendered as valid in the UI (existing configurations with the specific strings will still work but will be rendered with a red background. This is to ensure GTM will be able to only set the correct values)
  - Cody Gateway can translate the fireworks/starcoder model string to a specific model
    - For now this is 100% mapping to accounts/fireworks/models/starcoder-16b-w8a16
  - When validating the model for a request, we now expand virtual model strings so that `fireworks/starcoder` in the list of allowed models is enough to allow all Fireworks related models
  - PLG users now report `fireworks/starcoder` as a valid coder model in their license data
    - This will allow us to remove the specific models listed in this payload in the future (since Cody Gateway knows that this is a virtual model tag and how to expand it). However, the existing list of specific models is not changed (small asterisks here that I’ll discuss later)
   - Clients can now use `fireworks/starcoder` as a custom model. The SG instance will simply forward this to Cody Gateway
- Removal of more places that mentioned the old ST model name.
  - Here, it's important to know that the validation always happens with the _resolved_ model string. So after adding the `disableSingleTenant` flag, the resolved model will no longer be the old single-tenant model names. Thus we can remove support for it (as long as we keep the rewrite)

Here's a list of what changed from an actor perspective:

- PLG user
  - PLG traffic to dotcom will still send the same virtual model strings that we have added before (`fireworks/starcoder-16b`, `fireworks/starcoder-7b`). This is still going to be translated on the SG instance. However, that translation was also duplicated into Cody Gateway so that, after a Cody Gateway rollout, the SG instance can just pass-through this model string. 
  - We do not want to change the model strings PLG uses for now as we still need to be able to separate between the faster 7b and the more accurate 16b model in multi-tenant environments. 
  - The PLG user license ping will no longer report the old outdated single-tenant model as allowed (since Cody Gateway rewrites those anyways and checks against the resolved model string) and now includes the new `fireworks/starcoder` virtual mode. After a Cody Gateway rollout, it is safe to remove the old specific model strings as Cody Gateway can expand the new virtual one.
  - In the Cody extensions, nothing changes for PLG users. They will still opt-into Fireworks using the feature flag on the dotcom instance. When this FF is true, they will use the `fireworks/starcoder-16b` and `fireworks/starcoder-7b` virtual model names just as they are now.
- Enterprise user
  - It is now possible to add `fireworks/starcoder` to the allowed model list in the license data. When this is enabled, Cody Gateway starts to accept all specific starcoder model strings and the virtual one (`fireworks/starcoder`) from the model. When these are used, Cody Gateway translates the request to a specific model.
  - ⚠️ It is not possible yet for clients to pick up when the default model in the site config changes. This is going to be a separate PR in the Cody repo. Thus, Enterprise users would not be able to use Fireworks yet (unless they also set the feature flag). This is done here: https://github.com/sourcegraph/cody/pull/2714
  - ⚠️ This PR does not change the default models for licenses with no overwrite yet. I will do this in a follow-up PR after 2023-01-14 to be 100% safe. This is done here: https://github.com/sourcegraph/sourcegraph/pull/59586
     
### Enterprise instance with `fireworks/starcoder` and nothing else in the license data:

```
curl 'https://sourcegraph.test:3443/.api/completions/code' \
-X POST \
-H 'authorization: token TOKENTOKEN' \
--data-raw '{"messages":[{"speaker":"human","text":"function bubbleSort(){"}],"maxTokensToSample":30,"temperature":0.2,"stopSequences":[],"timeoutMs":5000,"stream":false,"model":"fireworks/starcoder"}'
{"completion":"\\n\\tvar len = arr.length;\\n\\tvar temp;\\n\\tfor(var i = 0; i \u003c","stopReason":"length"}%      
```

### Enterprise instance with no rate limit overwrites and no `fireworks/starcoder` in the license data

```
curl 'https://sourcegraph.test:3443/.api/completions/code' \
-X POST \
-H 'authorization: token TOKENTOKEN' \
--data-raw '{"messages":[{"speaker":"human","text":"function bubbleSort(){"}],"maxTokensToSample":30,"temperature":0.2,"stopSequences":[],"timeoutMs":5000,"stream":false,"model":"fireworks/starcoder"}'
Sourcegraph Cody Gateway: unexpected status code 400: {"error":"model \"fireworks/accounts/fireworks/models/starcoder-16b-w8a16\" is not allowed, allowed: []"}
```

### Dotcom users with the same license data as our production dotcom license key

Screenshot 2024-01-12 at 15 28 13

```
curl 'https://sourcegraph.test:3443/.api/completions/code' \
-X POST \
-H 'authorization: token TOKENTOKEN' \
--data-raw '{"messages":[{"speaker":"human","text":"function bubbleSort(){"}],"maxTokensToSample":30,"temperature":0.2,"stopSequences":[],"timeoutMs":5000,"stream":false,"model":"fireworks/starcoder-16b"}'
```


rafax:
> starcoder/fireworks

Note: PR description uses `starcoder/fireworks` (model first), while code (and all other models) use the "provider-first" convention - I think it makes sense to update the description to avoid confusion.

cmd/cody-gateway/internal/httpapi/completions/fireworks.go
```diff
@@ -117,17 +117,35 @@ func (f *FireworksHandlerMethods) transformBody(body *fireworksRequest, _ string
...
+			body.Model = "starcoder-7b"
 		}
 		if oldModel != body.Model {
 			f.baseLogger.Debug("rewriting model", log.String("old-model", oldModel), log.String("new-model", body.Model))
 		}
 	}
+
+	// Enterprise virtual model string
+	if body.Model == "starcoder" {
+		body.Model = "accounts/fireworks/models/starcoder-16b-w8a16"
```
rafax, line 139: WDYT about putting the model strings in a const (all, or at least the long ones)? One day, we'll make a typo in one of those, and it's going to be pretty hard to spot.

rafax:
> Dotcom users with the same license data as our production dotcom license key

Note - we don't use Dotcom licence data for anything nowadays, only the PLG model lists.

cmd/cody-gateway/internal/httpapi/completions/fireworks.go
```diff
@@ -117,17 +117,35 @@ func (f *FireworksHandlerMethods) transformBody(body *fireworksRequest, _ string
 	if body.N > 1 {
 		body.N = 1
 	}
+
+	// This code rewrites an older model identifier that we've used when testing a single-tenant
+	// deployment and that is no longer live. Since the clients used to hard code these, some
+	// outdated clients might still be sending these requests and we want to make sure they are
+	// compatible with the new virtual model strings.
 	if f.disableSingleTenant {
```
rafax, line 125: I think this has gotten rare enought that we can ignore it [altogether](https://console.cloud.google.com/monitoring/metrics-explorer;duration=P14D?pageState=%7B%22domainObjectDeprecationId%22:%22A0DA0488-2553-48A7-9317-E8F18BA31C70%22,%22xyChart%22:%7B%22constantLines%22:%5B%5D,%22dataSets%22:%5B%7B%22plotType%22:%22LINE%22,%22targetAxis%22:%22Y1%22,%22timeSeriesFilter%22:%7B%22aggregations%22:%5B%7B%22crossSeriesReducer%22:%22REDUCE_SUM%22,%22groupByFields%22:%5B%5D,%22perSeriesAligner%22:%22ALIGN_RATE%22%7D%5D,%22apiSource%22:%22DEFAULT_CLOUD%22,%22crossSeriesReducer%22:%22REDUCE_SUM%22,%22filter%22:%22metric.type%3D%5C%22logging.googleapis.com%2Fuser%2Frewritten_st_requests%5C%22%20resource.type%3D%5C%22cloud_run_revision%5C%22%22,%22groupByFields%22:%5B%5D,%22minAlignmentPeriod%22:%2260s%22,%22perSeriesAligner%22:%22ALIGN_RATE%22%7D%7D%5D,%22options%22:%7B%22mode%22:%22COLOR%22%7D,%22y1Axis%22:%7B%22label%22:%22%22,%22scale%22:%22LINEAR%22%7D%7D%7D&project=cody-gateway-prod)

rafax: I can remove it in a follow-up PR

internal/completions/client/fireworks/fireworks.go
```diff
@@ -12,6 +12,19 @@ import (
...
+
+// Multi-tenant model identifiers
+const Starcoder16bMultiTenant = "accounts/fireworks/models/starcoder-16b-w8a16"
+const Starcoder7bMultiTenant = "accounts/fireworks/models/starcoder-7b-w8a16"
+const Llama27bCodeMultiTenant = "accounts/fireworks/models/llama-v2-7b-code"
+const Llama213bCodeMultiTenant = "accounts/fireworks/models/llama-v2-13b-code"
+const Llama213bCodeInstructMultiTenant = "accounts/fireworks/models/llama-v2-13b-code-instruct"
+const Llama234bCodeInstructMultiTenant = "accounts/fireworks/models/llama-v2-34b-code-instruct"
+const Mistral7bInstructMultiTenant = "accounts/fireworks/models/mistral-7b-instruct-4k"
+const Mixtral8x7bInstructMultiTenant = "accounts/fireworks/models/mixtral-8x7b-instruct"
```
philipp-spiess, line 26: @rafax What do you think about this? 🙂

rafax: const LGTM, but I'd drop the MultiTenant suffix (most models have it as it is the default, so adding it to each name doesn't add much from my POV)

philipp-spiess: Sounds good 👍  Should we keep it for the singleTenant models? (I can remove the one for now as it's not used yet, will add a dice roll and % rollout feature flag in a follow-up PR)

client/web/src/enterprise/site-admin/dotcom/productSubscriptions/ModelBadges.tsx
```diff
@@ -47,14 +47,11 @@ function modelBadgeVariant(model: string, mode: 'completions' | 'embeddings'): '
...
             case 'openai/gpt-4-1106-preview':
-            // Bespoken models hosted for us by Fireworks. These are also allowed on the Cody
-            // Gateway side
-            case 'fireworks/accounts/fireworks/models/starcoder-16b-w8a16':
-            case 'fireworks/accounts/fireworks/models/starcoder-7b-w8a16':
-            case 'fireworks/accounts/fireworks/models/starcoder-3b-w8a16':
-            case 'fireworks/accounts/fireworks/models/starcoder-1b-w8a16':
-            case 'fireworks/accounts/sourcegraph/models/starcoder-7b':
-            case 'fireworks/accounts/sourcegraph/models/starcoder-16b':
+            // Virtual models that are translated by Cody Gateway and allows access to all StarCoder
```
rafax, line 50:
```suggestion
            // Virtual models that are translated by Cody Gateway and allow access to all StarCoder
```

rafax (commented):
LGTM overall, I will test it locally first thing tomorrow morning (with fresh eyes)

rafax (approved):

PR #59527, opened 2024-01-11
Searcher: improve how matches are built

jtibshirani:
This PR improves how searcher creates matches, making it more consistent with
how it's done in Zoekt.

Changes:
* Pull chunking logic out of structural search code and into its own file
`chunk.go`
* Remove overlapping ranges (this is what Zoekt does when chunk matches are
enabled)
* Optimize the column calculation using the same strategy from Zoekt ([zoekt#711](https://github.com/sourcegraph/zoekt/pull/711))

## Test plan

Adapted existing unit tests and added some new cases


cmd/searcher/internal/search/search_regex.go
```diff
@@ -242,6 +241,10 @@ func locsToRanges(buf []byte, locs [][]int) []protocol.Range {
 	prevStart := 0
 	prevStartLine := 0
 
+	c := columnHelper{
```
jtibshirani, line 244: I noticed some other opportunities for optimizations around newline handling. I've run out of steam for searcher improvements though, so just stopped here :)

cmd/searcher/internal/search/matchtree.go
```diff
@@ -327,15 +327,29 @@ func (om *orMatchTree) String() string {
...
 }
 
-// mergeMatches sorts the matched ranges and truncates the list
-// to obey limit. Consistent with behavior for diff/ commit search,
-// it does not merge or remove overlapping ranges.
+// mergeMatches sorts the matched ranges, removes overlaps, and truncates the list
+// to obey limit. This is consistent with the chunk match behavior in Zoekt, which
+// also removes overlapping ranges.
 func mergeMatches(matches [][]int, limit int) [][]int {
+	if len(matches) == 0 {
```
jtibshirani, line 334: This revisits the [decision I made in a previous PR](https://github.com/sourcegraph/sourcegraph/pull/59363/commits/9dd00203f61b045c36762048452252b0f447f656#r1445310225) but is still in line with our discussion.

cmd/searcher/internal/search/chunk.go
```diff
@@ -0,0 +1,195 @@
...
+	} else {
+		// Need to count from the beginning of line
+		runeCount = utf8.RuneCount(c.data[lineOffset:offset])
+	}
+
+	c.lastLineOffset = lineOffset
+	c.lastOffset = offset
+	c.lastRuneCount = runeCount
+
+	return runeCount
```
jtibshirani, line 194: In Zoekt, this line reads `return runeCount + 1`. However, that caused searcher tests to fail! It seems that searcher considers the column value start at 0, whereas Zoekt starts it at 1.

I plan to push a commit to update this to be consistent with Zoekt, and adapt all the tests. Since this discrepancy has been around for a while, I'm guessing we don't have frontend logic relying on the column value. Question for reviewer: could you double-check me on this?

cmd/searcher/internal/search/chunk.go
```diff
@@ -0,0 +1,195 @@
...
+				cover:  rr,
+				ranges: ranges[:1],
+			})
+			continue
+		}
+
+		lastChunk := &chunks[len(chunks)-1] // pointer for mutability
+		if lastChunk.cover.End.Line+interChunkLines >= rr.Start.Line {
+			// The current range overlaps with the current chunk, so merge them
+			lastChunk.ranges = ranges[i-len(lastChunk.ranges) : i+1]
```
keegancsmith, line 46: this logic is kinda tricky to get right. Why not just use append? If I am not mistaken append won't allocate, it will reuse the same underlying array as ranges. I guess the tricky thing there is on another level in that you worry about reusing the slice.

But either way I like that you re-use the underlying array. We don't do this in zoekt but should to avoid all the extra allocations. Did the benchmarks in searcher make you go down this path?

cmd/searcher/internal/search/chunk.go
```diff
@@ -0,0 +1,195 @@
...
+	return protocol.Range{
+		Start: protocol.Location{
+			Offset: firstLineStart,
+			Line:   inputRange.Start.Line,
+			Column: 0,
+		},
+		End: protocol.Location{
+			Offset: lastLineEnd,
+			Line:   inputRange.End.Line,
+			Column: int32(utf8.RuneCount(buf[lastLineStart:lastLineEnd])),
```
keegancsmith, line 95: Just to confirm, this doesn't have the hidden n^2 because you only calculate this column once per line?

cmd/searcher/internal/search/chunk.go
keegancsmith: cc @camdencheek who would know if column is used since I believe he included it in the chunkmatch API. If we don't actually use this value, can't we just remove it?

I double checked, looks good.

keegancsmith (approved):
LGTM!

cmd/searcher/internal/search/chunk.go
jtibshirani: This is a straight refactor, I didn't write any new logic here. I just moved these methods from `search_structural` to `chunk` since they were shared between core searcher logic and structural search. Here's the commit: https://github.com/sourcegraph/sourcegraph/pull/59527/commits/d4f7087913c9413d61187a0a3d20b34858d8a03f.

I think there's a lot of room for improvement here, but I'd like to scope this PR and not touch this logic.

cmd/searcher/internal/search/chunk.go
jtibshirani: That's my understanding as well. Also to clarify, this is pre-existing logic that I didn't touch, it's just moved: https://github.com/sourcegraph/sourcegraph/commit/d4f7087913c9413d61187a0a3d20b34858d8a03f.

cmd/searcher/internal/search/chunk.go
camdencheek: Making it consistent with Zoekt sounds good to me. IIRC, this was all just trying to preserve historical behavior.

We do use column in the web client. However, we could probably get rid of it in the API because both column and line can be calculated on the fly if we we have the chunk content (and the line number it starts on). The downside to dropping it from the API is that Zoekt can calculate lines more efficiently because it stores the offsets of newlines.

jtibshirani: Thanks for the context. I looked into this further, and the logic here is actually correct:
* The GraphQL API converts the column to `character`, [which is 0-indexed](https://github.com/sourcegraph/sourcegraph/blob/6dd737f15656c492e2a8e43aad7101f289c67686/cmd/frontend/graphqlbackend/schema.graphql#L10560-L10572)
* When frontend processes Zoekt matches, it [subtracts 1 from the column](https://github.com/sourcegraph/sourcegraph/blob/6dd737f15656c492e2a8e43aad7101f289c67686/internal/search/zoekt/indexed_search.go#L501)

Definitely surprising, but no bug to fix at the moment.

cmd/searcher/internal/search/chunk.go
jtibshirani: Note to self: in the future, I'll pull these moves out into their own PR since it makes the commit history clearer

PR #59529, opened 2024-01-11
Rename 'keyword' patterntype to 'codyContext'

jtibshirani:
This renames the experimental `keyword` pattern type to `codyContext` to
clarify its use case, and to distinguish it from the  "keyword search" project.
It also updates some internal naming (packages, methods, etc.)

## Test plan

Covered by existing tests


chwarwick (approved):
Thank you

keegancsmith (approved):

keegancsmith:
Is cody client dependent on the keyword patternType yet? When it does use it, are we sure it will always be speaking to a sourcegraph instance which contains this rename? If not, we may need to either support an alias to the keyword type on the backend or on the client detect which version of sourcegraph the endpoint is.

chwarwick:
> Is cody client dependent on the keyword patternType yet? When it does use it, are we sure it will always be speaking to a sourcegraph instance which contains this rename? If not, we may need to either support an alias to the keyword type on the backend or on the client detect which version of sourcegraph the endpoint is.

@keegancsmith  The cody clients do not directly use the search endpoint or have awareness of the patterntype they access this through the `getCodyContext`  resolver.  Currently Cody web is the only consumer.

jtibshirani:
Echoing what @chwarwick said, there are no external consumers of this pattern type. Clients use the GraphQL endpoint `getCodyContext`, which abstracts away the details of what search is used. (Thanks @camdencheek for the foresight here :))

PR #59533, opened 2024-01-11
otel: Only log type for unknown arguments in sql

eseliger:
This reverts commit 24fc9f5923328233c3257e72b10f74fb7682d658.

It doesn't have the intended effect on []int, and instead prints even longer values. 

## Test plan

CI.

### DIFF
--- a/internal/database/dbconn/open.go
+++ b/internal/database/dbconn/open.go
@@ -376,7 +376,7 @@ func argsAsAttributes(ctx context.Context, _ otelsql.Method, _ string, args []dr
 			attrs[i] = attribute.StringSlice(key, strings)
 
 		default: // in case we miss anything
-			attrs[i] = attribute.String(key, fmt.Sprintf("%v", v))
+			attrs[i] = attribute.String(key, fmt.Sprintf("%T", v))
 		}
 	}
 	return attrs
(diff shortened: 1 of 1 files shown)


bobheadxi (approved):

internal/database/dbconn/open.go
```diff
@@ -376,8 +376,7 @@ func argsAsAttributes(ctx context.Context, _ otelsql.Method, _ string, args []dr
 			attrs[i] = attribute.StringSlice(key, strings)
 
 		default: // in case we miss anything
-			// To prevent very large attributes, render at most 128 characters.
-			attrs[i] = attribute.String(key, fmt.Sprintf("%.128v", v))
+			attrs[i] = attribute.String(key, fmt.Sprintf("%v", v))
```
bobheadxi, line 379: Maybe we should just log the type here?

```suggestion
			attrs[i] = attribute.String(key, fmt.Sprintf("%T", v))
```

bobheadxi: If we encounter something we want to see we should write a special handler for it

eseliger: makes sense to me, let's do that 👍

PR #59534, opened 2024-01-12
Blob view: scroll search results to center of view

### LINKED ISSUE #57897: File/blob view: scroll viewport when finding in file
# Feature request description

When using Cmd+F to search inside a file, we scroll the viewport only far enough to make the highlighted match appear at the bottom of the blob view. It would be nice if the highlighted match appeared instead the middle of the blob view similar to how VS Code does it

# Additional context

## VSCode behavior
VSCode behavior

## Code Search behavior
Code Search behavior

camdencheek:
Fixes #57897 

This adds an effect on `findNext` that scrolls the newly-selected item into the middle third of the view rather than just scrolling the minimum amount for it to show up in the view. If it is already visible in the view, no scrolling happens.

https://github.com/sourcegraph/sourcegraph/assets/12631702/121ec818-4eb4-4dd9-9351-c36c375d253b

## Test plan

Manually tested that the scrolling feels right.

### DIFF
--- a/client/web/src/repo/blob/codemirror/search.tsx
+++ b/client/web/src/repo/blob/codemirror/search.tsx
@@ -322,10 +322,24 @@ class SearchPanel implements Panel {
 
     private findNext = (): void => {
         findNext(this.view)
+        // Scroll the selection into the middle third of the view
+        this.view.dispatch({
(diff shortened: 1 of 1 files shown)


peterguy (approved):
love it!

BolajiOlajide (approved):

PR #59535, opened 2024-01-12
msp/spec: validate CPU/memory configurations

bobheadxi:
As with many GCP resources, this fails only at TF apply-time - when planning, no error is raised. This probably changes often enough to warrant some up-front checking to avoid some frustration (we already ran into this once).

Cody was able to write a good chunk of this nicely 😁 

## Test plan

Unit test


dev/managedservicesplatform/spec/environment.go
```diff
@@ -283,18 +285,79 @@ func (e *EnvironmentDomainCloudflareSpec) ShouldProxy() bool {
...
+	} else if gib > 4 && s.CPU < 2 {
+		errs = append(errs, errors.New("resources.memory > 4GiB requires resources.cpu >= 2"))
+	}
+
+	// Enforce min memory: https://cloud.google.com/run/docs/configuring/services/cpu#cpu-memory
+	if s.CPU > 6 && gib < 4 {
+		errs = append(errs, errors.New("resources.cpu > 6 requires resources.memory >= 4GiB"))
+	} else if s.CPU > 4 && gib < 2 {
+		errs = append(errs, errors.New("resources.cpu > 4 requires resources.memory >= 2GiB"))
+	}
```
unknwon, line 356: nit: each can be a loop of map? e.g. pseudo code

```
minCPU := map[gib]cpu{
 24: 8,
 16: 6,
 ...
}
```

unknwon (approved):

dev/managedservicesplatform/spec/environment.go
```diff
@@ -283,18 +285,79 @@ func (e *EnvironmentDomainCloudflareSpec) ShouldProxy() bool {
...
+
+	var errs []error
+
+	// https://cloud.google.com/run/docs/configuring/services/cpu
+	if s.CPU < 1 {
+		errs = append(errs, errors.New("resources.cpu must be >= 1"))
+	} else if s.CPU > 8 {
+		errs = append(errs,
+			errors.New("resources.cpu > 8 not supported - considering decreasing scaling.maxRequestConcurrency and increasing scaling.maxCount instead"))
```
jac, line 323:
```suggestion
			errors.New("resources.cpu > 8 not supported - consider decreasing scaling.maxRequestConcurrency and increasing scaling.maxCount instead"))
```

dev/managedservicesplatform/spec/environment.go
```diff
@@ -283,18 +285,79 @@ func (e *EnvironmentDomainCloudflareSpec) ShouldProxy() bool {
...
+	if err != nil {
+		errs = append(errs, errors.Wrap(err, "resources.memory is invalid"))
+	}
+	if units.Base2Bytes(bytes)/units.MiB < 512 {
+		errs = append(errs, errors.New("resources.memory must be >= 512MiB"))
+	}
+	gib := units.Base2Bytes(bytes) / units.GiB
+	if gib > 32 {
+		errs = append(errs,
+			errors.New("resources.memory > 32GiB not supported - considering decreasing scaling.maxRequestConcurrency and increasing scaling.maxCount instead"))
```
jac, line 337:
```suggestion
			errors.New("resources.memory > 32GiB not supported - consider decreasing scaling.maxRequestConcurrency and increasing scaling.maxCount instead"))
```

jac (commented):
feels a bit strange to have all the error messages contain `MiB` and `GiB` if the unit should actually be `Mi` or `Gi`

bobheadxi:
> feels a bit strange to have all the error messages contain MiB and GiB if the unit should actually be Mi or Gi

Ah, good catch!

dev/managedservicesplatform/spec/environment.go
bobheadxi: is this easier to understand? 😆 or maybe if you have more extended pseudocode I might see it better?

one thing I think we might run into with map is, we want to make sure the error message shows the highest "category" you fit into, e.g. show the error message for `24: 8` even if you match `16: 6`. We would have to iterate the map keys, find the highest we match, and then get the corresponding threshold, which feels... a bit wordy

bobheadxi:
Fixed the error messages, merging as-is - happy to follow up with any other suggestions

dev/managedservicesplatform/spec/environment.go
unknwon: Ah yeah, order is a problem. should use a list, a nit anymore, not seeing a problem with current number of `if`s 😁

PR #59548, opened 2024-01-12
licensing: surface cody feature status

BolajiOlajide:
Similar to #59528, this surfaces the cody feature status.

## Test plan

`features.cody` should be available in the JSContext.

### DIFF
--- a/cmd/frontend/hooks/hooks.go
+++ b/cmd/frontend/hooks/hooks.go
@@ -23,6 +23,7 @@ type FeatureBatchChanges struct {
 // enabled/disabled on the current license.
 type LicenseFeatures struct {
 	CodeSearch bool `json:"codeSearch"`
+	Cody       bool `json:"cody"`
 }
 
 // LicenseInfo contains non-sensitive information about the legitimate usage of the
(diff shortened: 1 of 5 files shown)


cmd/frontend/hooks/hooks.go
```diff
@@ -23,6 +23,7 @@ type FeatureBatchChanges struct {
 // enabled/disabled on the current license.
 type LicenseFeatures struct {
 	CodeSearch bool `json:"codeSearch"`
+	Cody       bool `json:"cody"`
```
eseliger, line 26: There are already these three fields: 

https://sourcegraph.com/-/editor?remote_url=github.com%2Fsourcegraph%2Fsourcegraph&branch=HEAD&file=cmd%2Ffrontend%2Finternal%2Fapp%2Fjscontext%2Fjscontext.go&editor=VSCode&version=2.2.16&start_row=181&start_col=0&end_row=189&end_col=0

Should we somehow consolidate those?

BolajiOlajide: I believe those fields are unrelated to licensing and only have to do with site configuration and access control.

I would want to leave those separated since they're unrelated to the license and AFAICT; if the CodyFeature isn't available on a license, then those fields are more or less redundant.

fkling (approved):

PR #59551, opened 2024-01-12
DB: do not pre-allocate massive amounts of storage in the inserter

camdencheek:
`NewInserterWithReturn` currently accounts for 26% of the allocated bytes in `frontend` on sourcegraph.com. I'm like 90% sure it's because every time we call `NewInserterWithReturn`, we allocate two slices with capacity 65535 (the max number of parameters for Postgres). 

In the _vast_ majority of cases, these slices only need to be a few elements long (most uses are coming from event logs). In the case where we need more capacity, `append` will take care of it at the cost of a few extra allocations.

CleanShot 2024-01-12 at 11 00 29@2x

## Test plan

CI.


bobheadxi (approved):

bobheadxi:
(I think one of the heavy callers is the new telemetry, which does use this in a `conc.WaitGroup`)

camdencheek:
Yeah, I noticed that. Allocating a megabyte per log is probably not ideal 😄

camdencheek:
A profile from after deploying:

CleanShot 2024-01-12 at 15 29 19@2x

eseliger:
great stuff! thanks :)

PR #59553, opened 2024-01-12
[Backport 5.2] accesstokens: allow last_used_at to be semi-stale (#59284)

bobheadxi:
This backports #59284, which changes the `last_used_at` column of access tokens to only be updated after a period of time so that we avoid significant database contention in periods of high token-auth'd API traffic that we've been observing in Sourcegraph.com due to the Cody PLG launch.

I'm backporting this because it turns out we are seeing similar problems placing significant database load on Cloud instances that have adopted Cody as they start increasing their API access traffic in a way that probably never happened before as most Code Search usage happens via UI - I've seen this alert fire many times on multiple Cloud instances. On a spot-check, at least several of the spikes in Cloud `pgsql` database (equivalent to the dotcom `sg` database) are entirely attributable to access token lookups - see [Slack message](https://sourcegraph.slack.com/archives/C05SZB829D0/p1705084254251499?thread_ts=1704221033.599109&cid=C05SZB829D0)

This was a tricky backport, so would appreciate a careful review 🙏 

(cherry picked from commit 3d9f69df4fd43c87b2b3216db81ae0e94a4eadb7)

Test plan: automated tests

### DIFF
--- a/cmd/frontend/internal/dotcom/productsubscription/tokens_db.go
+++ b/cmd/frontend/internal/dotcom/productsubscription/tokens_db.go
@@ -2,10 +2,15 @@ package productsubscription
 
 import (
 	"context"
+	"database/sql"
 	"encoding/hex"
 	"strings"
+	"time"
 
 	"github.com/keegancsmith/sqlf"
+
+	"github.com/sourcegraph/sourcegraph/internal/accesstoken"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
 
 	"github.com/sourcegraph/sourcegraph/internal/database"
 	"github.com/sourcegraph/sourcegraph/internal/database/basestore"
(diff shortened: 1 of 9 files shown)


cmd/frontend/internal/dotcom/productsubscription/tokens_db.go
```diff
@@ -82,37 +87,56 @@ func (e dotcomUserNotFoundError) NotFound() bool {
...
-// dotcomUserGatewayAccessTokenPrefix is the prefix used for identifying tokens
-// generated for dotcom users to access the cody-gateway.
-const dotcomUserGatewayAccessTokenPrefix = "sgd_"
-
 // LookupDotcomUserIDByAccessToken returns the userID
 // corresponding to a token, trimming token prefixes if there are any.
 func (t dbTokens) LookupDotcomUserIDByAccessToken(ctx context.Context, token string) (int, error) {
-	if !strings.HasPrefix(token, dotcomUserGatewayAccessTokenPrefix) {
+	if !strings.HasPrefix(token, accesstoken.DotcomUserGatewayAccessTokenPrefix) {
 		return 0, dotcomUserNotFoundError{reason: "invalid token with unknown prefix"}
```
bobheadxi, line 94: It looks like this shared const is available in 5.2 but was only adopted at this callsite here later. The definition is:

```go
// DotcomUserGatewayAccessTokenPrefix is the prefix used for identifying tokens
// generated for a dotcom api token .
const DotcomUserGatewayAccessTokenPrefix = "sgd_" // "(S)ource(g)raph (d)otcom user key"
```

which matches the existing const that I've removed in this backport

bobheadxi: Either way, this is a dotcom-specific features, so the changes in this entire function (`LookupDotcomUserIDByAccessToken`) is not used in any enterprise/Cloud/on-prem use cases

chrsmith (approved):
This looks right to me. There are a few changes that seem a little dubious, but I'm guessing that they are changes that my PR was built on top of. (And needed to be included for this back port, such as ` getInternalToken()`.)

Functionality wise, the new behavior is pretty well covered by the unit tests. So if those are passing I'd feel confident merging the changes.

PR #59554, opened 2024-01-12
ci: only run prettier on files changed between <commit> and <PR base branch>

Strum355:
`sg lint`'s `format` target runs `./dev/ci/pnpm-run.sh format:check`, which runs prettier against _every_ file. We can do better than that in CI by only running it against files that have been modified between the current branch and the base branch of the PR.

~**TODO:** this probably breaks when run on `main`, so we need to fix that first.~ Format step isnt run on main, so I _think_ this is safe

Before (left) After (right)
image

## Test plan

Tested on a branch based off https://github.com/sourcegraph/sourcegraph/tree/vk/add-mobile-layout-for-new-filters (https://github.com/sourcegraph/sourcegraph/tree/nsc-vk/lint-test_add-mobile-layout-for-new-filters

https://buildkite.com/sourcegraph/sourcegraph/builds/257779#_

### DIFF
--- a/dev/sg/linters/prettier.go
+++ b/dev/sg/linters/prettier.go
@@ -2,6 +2,7 @@ package linters
 
 import (
 	"context"
+	"os"
 
 	"github.com/sourcegraph/run"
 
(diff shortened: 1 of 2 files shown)


jamesmcnamara (approved):
LGTM

PR #59558, opened 2024-01-12
Search: bump default streaming limit to 1000

camdencheek:
In order to show users exhaustive search counts more frequently, I'm experimenting with bumping up the default search limit. Note, this does not change the display limit (the number of full results sent to the client). I hope to increase this more significantly, but I'm starting small to limit risk. 

[Related slack thread](https://sourcegraph.slack.com/archives/C05EA9KQUTA/p1704725763094299)

## Test plan

Will keep an eye on perf indicators on s2 and dotcom


jasonhawkharris (approved):
LGTM!

PR #59560, opened 2024-01-12
gitlab: fallback to http when custom git clone protocol is used

michaellzc:

### DIFF
--- a/internal/repos/gitlab.go
+++ b/internal/repos/gitlab.go
@@ -254,7 +254,28 @@ func (s *GitLabSource) remoteURL(proj *gitlab.Project) string {
 	if s.config.GitURLType == "ssh" {
 		return proj.SSHURLToRepo // SSH authentication must be provided out-of-band
 	}
+	// gitlab supports special protocol to support custom git remote plugin, e.g., iap://
+	// Sourcegraph does not support custom git remote plugin nor handle custom authentication protocol in addition to code host authentication,
+	// so we need to ensure we fallback to http
+	if (s.config.GitURLType == "" || s.config.GitURLType == "http") && !strings.HasPrefix(proj.HTTPURLToRepo, "http") {
+		uri, err := url.Parse(proj.HTTPURLToRepo)
+		if err != nil {
+			s.logger.Warn("parse GitLab project URL", log.String("url", proj.HTTPURLToRepo), log.Error(err))
+			return proj.HTTPURLToRepo
+		}
+		if s.isHTTPS() {
+			uri.Scheme = "https"
+		} else {
+			uri.Scheme = "http"
+		}
+		return uri.String()
+	}
 	return proj.HTTPURLToRepo
+}
+
+// isHTTPS returns true if the configured GitLab connection uses HTTPS.
+func (s *GitLabSource) isHTTPS() bool {
+	return strings.HasPrefix(s.config.Url, "https")
 }
 
 func (s *GitLabSource) excludes(p *gitlab.Project) bool {
--- a/internal/repos/testdata/gitlab-repos.json
+++ b/internal/repos/testdata/gitlab-repos.json
@@ -34,5 +34,17 @@
     "archived": false,
     "default_branch": "main",
     "topics": ["topic1", "topic2"]
+  },
+  {
+    "id": 4,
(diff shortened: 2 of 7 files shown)


michaellzc:
* `main`
  * **PR #59560** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/59560?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/59560?utm_source=stack-comment).

internal/repos/gitlab.go
```diff
@@ -254,9 +254,30 @@ func (s *GitLabSource) remoteURL(proj *gitlab.Project) string {
 	if s.config.GitURLType == "ssh" {
 		return proj.SSHURLToRepo // SSH authentication must be provided out-of-band
 	}
+	// gitlab supports special protocol to support custom git remote plugin, e.g., iap://
+	// Sourcegraph does not support custom git remote plugin nor handle custom authentication protocol in addition to code host authentication,
+	// so we need to ensure we fallback to http
+	if (s.config.GitURLType == "" || s.config.GitURLType == "http") && !strings.HasPrefix(proj.HTTPURLToRepo, "http") {
```
michaellzc, line 260: this implies `http` is the default, is that okay?

eseliger:
> not sure how to test against a real GitLab instance, I can't even find the place to configure custom git clone protocl

I guess they're using https://docs.gitlab.com/ee/administration/settings/visibility_and_access_controls.html#customize-git-clone-url-for-https, TIL that this exists. Might be that you can overwrite the protocol there? I'm surprised that works. 

See https://git-scm.com/docs/gitremote-helpers, git calls a helper binary called git-remote-<protocolname> to do the fetch. Default providers for git-remote-http, git-remote-https, git-remote-ftp and git-remote-ftps are provided out of the box. 
I wonder if they use a custom binary locally that makes it work for their envs to pull from these URLs? 

Update: I confirmed this setting actually does that:
Screenshot 2024-01-14 at 21 29 27@2x

Given that, I wonder if that isn't misconfiguration of the GitLab instance. Reading the above linked page from gitlab, it clearly sounds like it is meant to modify the HTTP hostname or path prefix, but *not* the protocol, and a paragraph above they mention that gitlab explicitly supports HTTP(S) and SSH, full stop. Nothing else.

The fact that the field "http_url_to_repo" has HTTP in the name but isn't an HTTP URL kind of backs me up here, IMO.

For example, I wouldn't be sure that just replacing the protocol will make things work. What if the hostname is also changed, and that actually only speaks a custom protocol, and not HTTP. 

So while this might fix it for this very specific setup where this error was first observed, it feels a bit like working around a configuration issue and not an actually wrong setting, and this fallback might cause other hidden issues, for example if the host that handles the IAP doesn't work for plain HTTP clones.

--- 

That said, we need a solution for this problem. I propose that this is NOT a product bug, or feature gap, unless someone can prove otherwise that GitLab anticipated someone to use a non-https protocol here.

I think the right things to do:

- Reach out to GitLab and ask about that, so we know what to expect 
- Clearly document that even though this setting exists, sourcegraph ONLY supports HTTP and SSH clones, while SSH has to be enabled manually
- We provide instructions for a workaround. I propose that if they actually want to use IAP, that they should add a git-remote-iap binary to their gitserver image to $PATH. If that binary existed and was configured correctly, I believe their setup should "just work" - although disclaimer, this is not official. 

How do you feel about that?

Also cc @sourcegraph/source for visibility.

michaellzc:
@eseliger adding more context:

The customer is on Cloud and they are hosting their own GitLab behind IAP. They have a git remote plugin to handle the IAP protocol to authenticate with the IAP tunnel.

When Cloud instance is connecting to their GitLab, it talks to a different entrypoint that doesn’t enforce IAP, hence we need to use HTTP.

Deferring next step to the Source team.

eseliger (commented):
Is this okay to close?

michaellzc:
> Is this okay to close?

yes. sorry for leaving this dangling!

Issue #59545, opened 2024-01-12
P0: Expiry for a token


chwarwick:
Add expiration for api tokens

PR #59565, opened 2024-01-13
Implement expiry for access tokens

chwarwick:
Add expiry to access tokens. Users can now select a maximum timespan for which a token is valid. Tokens will automatically lose access after this period.

By default all new tokens will require expiration; however there is a site configuration setting to allow users to create tokens without expiration if necessary.  The preconfigured options when creating a token are 7, 14, 30, 60, or 90 days with the selection defaulting to 90 days.

Tokens that are generated in the background for example from the Cody VS Code extension will use the default expiration period for the instance.

Defaults
Screenshot 2024-01-15 at 2 55 18 PM

Options when allowing no expiration
Screenshot 2024-01-15 at 2 54 53 PM

Allow no expiration but show warning
Screenshot 2024-01-15 at 2 54 38 PM

Created from VS Code shows using the default 
Screenshot 2024-01-15 at 2 56 17 PM

closes https://github.com/sourcegraph/sourcegraph/issues/59545
## Test plan
  - Updated existing and added new tests for token lookup and creation
  - Update integration tests
  - Manual testing of configuration options and creating from clients.


CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ All notable changes to Sourcegraph are documented in this file.
 - Supports custom ChatCompletion models in Cody clients for dotcom users. [#58158](https://github.com/sourcegraph/sourcegraph/pull/58158)
 - Topics synced from GitHub and GitLab are now displayed for repository matches in the search results and on the repository tree page. [#58927](https://github.com/sourcegraph/sourcegraph/pull/58927)
 - Added a new column "Repository metadata JSON" to the CSV export of repository search results, which includes the JSON encoded object of metadata key-value pairs. [#59334](https://github.com/sourcegraph/sourcegraph/pull/59334)
+- Expiry to access tokens. Users can now select a maximum timespan for which a token is valid. Tokens will automatically lose access after this period. Defaults can be configured in the `auth.accessTokens` section of the site configuration. [#59565](https://github.com/sourcegraph/sourcegraph/pull/59565)
```
slimsag, line 25: Maybe make note here of how non-expiring access tokens can still be created with a certain site config?

slimsag:
Have you checked what happens if a token is expired in the VS Code client? e.g. do you get an obscure error or is it clear about what you need to do to get the Cody client working again?

slimsag (approved):
Looks correct & well-tested to me, great work

chwarwick:
> Have you checked what happens if a token is expired in the VS Code client? e.g. do you get an obscure error or is it clear about what you need to do to get the Cody client working again?

VS Code isn't too bad, JetBrains you need to look a little deeper, I opened issues for it:
https://github.com/sourcegraph/cody/issues/2788
https://github.com/sourcegraph/jetbrains/issues/311

chwarwick:
@unknwon @vdavid tagging you here because I don't want to introduce any new complications in the SAMS/SSC PLG workflows.

client/web/dev/utils/create-js-context.ts
```diff
@@ -26,6 +26,9 @@ export const createJsContext = ({ sourcegraphBaseUrl }: { sourcegraphBaseUrl: st
         temporarySettings: null,
         externalURL: sourcegraphBaseUrl,
         accessTokensAllow: 'all-users-create',
+        accessTokensAllowNoExpiration: false,
+        accessTokensExpirationDaysDefault: 90,
+        accessTokensExpirationDaysOptions: [90, 7, 14, 60],
```
unknwon, line 31: nit unless `90` as the first element has a special meaning... which should then accompanied with a comment.
```suggestion
        accessTokensExpirationDaysOptions: [7, 14, 60, 90],
```

client/web/src/user/settings/accessTokens/UserSettingsCreateAccessTokenCallbackPage.tsx
```diff
@@ -286,8 +304,10 @@ export const UserSettingsCreateAccessTokenCallbackPage: React.FC<Props> = ({
                                 <Text>{requester.name} access token successfully generated.</Text>
                                 <CopyableText className="test-access-token" text={newToken} />
                                 <Text className="form-help text-muted" size="small">
-                                    This is a one-time access token to connect your account to {requester.name}. You
-                                    will not be able to see this token again once the window is closed.
+                                    This is a one-time access token to connect your account to {requester.name}. This
+                                    token will expire in {defaultAccessTokenExpiryDays}{' '}
+                                    {pluralize('day', defaultAccessTokenExpiryDays)}. You will not be able to see this
+                                    token again once the window is closed.
```
unknwon, line 310: What will this message say, when a user created a token without expiry?

cmd/frontend/graphqlbackend/access_tokens.go
```diff
@@ -59,6 +63,24 @@ func (r *schemaResolver) CreateAccessToken(ctx context.Context, args *createAcce
...
+	var expiresAt time.Time
+	if args.ExpiresAt != nil {
+		expiresAt, err = time.Parse(time.RFC3339, *args.ExpiresAt)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to parse expires at as time")
+		}
+		switch duration := time.Until(expiresAt); {
+		case duration < 0:
+			return nil, errors.New("expiry must be in the future")
+		case duration > maxTokenDuration:
```
unknwon, line 75: 🤔 Shouldn't we compute the the "max" based on the list of allowed duration? It seems potentially confusing to customer admins that they thought they've only allowed `15, 30, 90` days but someone knows GQL can still create a year-long expiry.

internal/database/schema.md
```diff
@@ -34,6 +34,7 @@ Foreign-key constraints:
  creator_user_id | integer                  |           | not null | 
  scopes          | text[]                   |           | not null | 
  internal        | boolean                  |           |          | false
+ expires_at      | timestamp with time zone |           |          |
```
unknwon, line 37: Maybe worth checking whether we need an update to the `access_tokens_lookup` index 
 (line 41)?

client/web/dev/utils/create-js-context.ts
chwarwick: I did mention it in the site config docs.  This is admittedly kind of awkward but I wanted a way to ensure the default was actually a value that was in the list of options.  I guess I could add the default option into the list if it didn't already exist. Open to suggestions on something more obvious to users.

client/web/src/user/settings/accessTokens/UserSettingsCreateAccessTokenCallbackPage.tsx
chwarwick: @unknwon This is the callback token generation page that is used by the extensions, the user doesn't have the ability to pick an expiration here. The token in generated without their input, it will use whatever the default is and currently I didn't provide a way to default to no expiration, since enforcing an expiration for this pathway is the primary purpose of this pr.

unknwon: Ahh, I see, NVM then!

cmd/frontend/graphqlbackend/access_tokens.go
eseliger: Yes that makes sense to me

client/web/src/user/settings/accessTokens/UserSettingsCreateAccessTokenCallbackPage.tsx
eseliger: What does one-time access token mean in this context? I'd expect it to work for only one request

eseliger (approved):

client/web/src/user/settings/accessTokens/UserSettingsCreateAccessTokenCallbackPage.tsx
chwarwick: It's the 2nd half of this page, when you expand to see the token.  That was the original copy but that isn't accurate  so I'll remove the `one-time`.
image

vdavid:
Thanks for tagging me, @chwarwick!
Cody PLG and Cody Web use internal access tokens to talk to Cody Gateway.
If I understand [this](https://github.com/sourcegraph/sourcegraph/pull/59565/files#diff-01e47a9f18b465559ebf3de20d7ae7190ee348325fe4b7ac9cfd30ff714b8cadR175) and [this](https://github.com/sourcegraph/sourcegraph/pull/59565/files#diff-01e47a9f18b465559ebf3de20d7ae7190ee348325fe4b7ac9cfd30ff714b8cadR216) correctly, internal access tokens are created with NULL expiration, which means they never expire, which makes our case easy. @chwarwick, please confirm that my understanding is correct.
I can't think of other ways this affects PLG specifically.

@sourcegraph/cody-plg, do you think introducing token expirations affects us in any other way?

chwarwick:
@unknwon  @eseliger  @slimsag  I updated based on the feedback, changes were a bit more extensive since I changed to mutation to accept a duration instead of an expiration time.  I plan to merge later today if you want a 2nd look.

cmd/frontend/graphqlbackend/access_tokens.go
```diff
@@ -273,3 +297,12 @@ func (r *accessTokenConnectionResolver) PageInfo(ctx context.Context) (*graphqlu
 	}
 	return graphqlutil.HasNextPage(r.opt.LimitOffset != nil && len(accessTokens) > r.opt.Limit), nil
 }
+
+func getMaxExpiryDuration(allowedOptionsInDays []int) (int32, error) {
+	if len(allowedOptionsInDays) == 0 {
+		return 0, errors.New("no expiry options available")
+	}
+
+	maxDaysAllowed := allowedOptionsInDays[len(allowedOptionsInDays)-1]
```
unknwon, line 306: 🤔 This seems too much of a shortcut... this is only technically `getLastExpiryDuration`. Not sure it's worth saving a quick for loop here just to do it right.

Or if we want to rely on some implicit guarantees from call site, we need a comment saying that.

alexAtSourcegraph:
@chwarwick What would happen to existing access tokens configured prior to this change once this site configuration is set?

chwarwick:
> @chwarwick What would happen to existing access tokens configured prior to this change once this site configuration is set?

@alexAtSourcegraph  They do not expire, this only effects **new** tokens created.

PR #59566, opened 2024-01-13
msp: validate that domains are unique across environments

bobheadxi:
Made the mistake of forgetting to change the domain when setting up a new test environment in `msp-testbed`

## Test plan

n/a

### DIFF
--- a/dev/managedservicesplatform/spec/spec.go
+++ b/dev/managedservicesplatform/spec/spec.go
@@ -10,6 +10,7 @@ import (
 	"gopkg.in/yaml.v3"
 
 	"github.com/sourcegraph/sourcegraph/lib/errors"
+	"github.com/sourcegraph/sourcegraph/lib/pointers"
 )
 
 // Spec is a Managed Services Platform (MSP) service.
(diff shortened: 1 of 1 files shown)


dev/managedservicesplatform/spec/spec.go
```diff
@@ -157,6 +159,16 @@ func (s Spec) Validate() []error {
...
 				env.ID, env.ProjectID))
 		}
+
+		if domain := pointers.DerefZero(env.EnvironmentServiceSpec).Domain.GetDNSName(); domain != "" {
+			_, exists := configuredDomains[domain]
+			if exists {
+				errs = append(errs, errors.Newf("domain %q is configured more than once across environments",
+					domain))
+			} else {
+				configuredDomains[domain] = struct{}{}
```
chrishalbert, line 169: Is using an empty struct convention? Just curious..

chrishalbert (approved):

dev/managedservicesplatform/spec/spec.go
jac: @chrishalbert  Yes, as an empty struct has a size of 0 bytes it allows one to efficiently create a HashSet
https://www.sohamkamani.com/golang/sets/

afaik this is a common pattern across languages e.g. Zig where a Set is a HashMap with `void` values
https://devlog.hexops.com/2022/zig-hashmaps-explained/#hashmaps-are-also-sets

jac (approved):

PR #59567, opened 2024-01-13
Do not use locally built `sg` on aspect workflows

jamesmcnamara:
Our aspect workflows skip running bazel prechecks, which means the dependency that sg-lint has is unsatisfied, causing all sg lint jobs to fail.

Personally I think we need a more robust solution for determining dependency satisfaction. This isn't currently possible because steps are completely programmatic and so a missing dependency cannot be synthesized in all cases. 

I think we should try to work towards doing that, but this hot-fix removes the dependency in aspect manually.

## Test plan
Runs in CI

### DIFF
--- a/dev/ci/internal/ci/misc_operations.go
+++ b/dev/ci/internal/ci/misc_operations.go
@@ -43,7 +43,7 @@ func bazelGoModTidy() func(*bk.Pipeline) {
 func addSgLints(targets []string, opts CoreTestOperationsOptions) func(*bk.Pipeline) {
 	cmd := "./sg"
 	if opts.AspectWorkflows {
-		cmd = "go run ./dev/sg"
+		cmd = "go run ./dev/sg "
 	}
 
 	if retryCount := os.Getenv("BUILDKITE_RETRY_COUNT"); retryCount != "" && retryCount != "0" {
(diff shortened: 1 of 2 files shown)


dev/ci/internal/ci/misc_operations.go
```diff
@@ -40,8 +40,11 @@ func bazelGoModTidy() func(*bk.Pipeline) {
 }
 
 // addSgLints runs linters for the given targets.
-func addSgLints(targets []string) func(*bk.Pipeline) {
-	cmd := "./sg "
+func addSgLints(targets []string, opts CoreTestOperationsOptions) func(*bk.Pipeline) {
+	cmd := "./sg"
```
Strum355, line 44: Quick catch, these strings need a space after them for the flags to follow! 🙂

dev/ci/internal/ci/misc_operations.go
```diff
@@ -67,16 +70,22 @@ func addSgLints(targets []string) func(*bk.Pipeline) {
...
 					IncludeNames: true,
 					Type:         bk.AnnotationTypeAuto,
 				},
-			}))
+			}),
+		}
+
+		if !opts.AspectWorkflows {
+			steps = append(steps, bk.DependsOn("bazel-prechecks"),
+				bk.Cmd("buildkite-agent artifact download sg . --step bazel-prechecks"),
```
Strum355, line 85: This Cmd would have to come before the one on L75 (the AnnotatedCmd)

PR #59571, opened 2024-01-15
Searcher: natively support AND/ OR patterns

### LINKED ISSUE #59038: Searcher: optimize AND/ OR patterns

jtibshirani:
This is the final PR to handle AND/ OR queries directly in searcher, instead of
running each clause separately and combining the results in frontend. Major
changes:
1. Update `job.go` to pass entire query to searcher, instead of creating flat
jobs
2. Update `TextPatternInfo` and `protocol.PatternInfo` to hold the whole query
instead of the string pattern directly. This was tricky, since these types are
also used for structural search, which does not support AND/ OR directly.
3. Within searcher, convert this query to a `matchTree` (added in [#59363](#59363))

Closes #59038

### Test Plan

Added to existing end-to-end and unit tests. Light manual testing.


jtibshirani:
This change required a lot of small refactors and bug fixes, which were hard to break into individual changes. So apologies for the big review! I'll add a few review comments to help explain what's going on.

internal/search/job/jobutil/exhaustive_job.go
```diff
@@ -60,10 +60,9 @@ func NewExhaustive(inputs *search.Inputs) (Exhaustive, error) {
 		return Exhaustive{}, errors.Errorf("regex search with .* is not supported")
 	}
 
-	planJob, err := NewFlatJob(inputs, query.Flat{Parameters: b.Parameters, Pattern: &term})
-	if err != nil {
-		return Exhaustive{}, err
-	}
+	repoOptions := toRepoOptions(b, inputs.UserSettings)
+	resultTypes := computeResultTypes(b, inputs.PatternType)
+	planJob := NewTextSearchJob(b, inputs, resultTypes, repoOptions)
```
jtibshirani, line 65: I pulled out the shared method `NewTextSearchJob` so we can use it both for unindexed search and here in search jobs (exhaustive search). After this PR, search jobs could handle boolean queries.

internal/search/job/jobutil/job.go
```diff
@@ -497,39 +497,35 @@ func NewFlatJob(searchInputs *search.Inputs, f query.Flat) (job.Job, error) {
...
-			if patternInfo.IsCaseSensitive {
-				pathRegexps = append(pathRegexps, regexp.MustCompile(pattern))
-			} else {
-				pathRegexps = append(pathRegexps, regexp.MustCompile(`(?i)`+pattern))
-			}
+// getPathRegexps parses the search and returns a set of regular expressions that represent
+// how it matches file paths. These regexes are later used to create result matches on the
+// file paths. We pass in the original query.Basic in addition to the TextPatternInfo just
+// for convenience, so we can use methods like query.VisitPattern.
+func getPathRegexps(b query.Basic, p *search.TextPatternInfo) (pathRegexps []*regexp.Regexp) {
```
jtibshirani, line 504: I needed to refactor this so it walks the whole query to find path patterns.

internal/search/types.go
```diff
@@ -297,12 +298,10 @@ type SearcherParameters struct {
 // TextPatternInfo defines the search request for unindexed and structural search
 // (the 'searcher' service). Keep it in sync with pkg/searcher/protocol.PatternInfo.
 type TextPatternInfo struct {
-	// Values dependent on pattern atom.
```
jtibshirani, line 300: In a follow-up, it could make sense to rename `TextPatternInfo` and `protocol.PatternInfo` since they don't hold a single pattern anymore.

cmd/searcher/internal/search/search_structural.go
```diff
@@ -211,34 +210,26 @@ func lookupMatcher(language string) string {
...
-		PatternMatchesPath:           p.PatternMatchesPath,
-		Languages:                    p.Languages,
-	}
-
 	if p.Branch == "" {
 		p.Branch = "HEAD"
 	}
 	branchRepos := []zoektquery.BranchRepos{{Branch: p.Branch, Repos: roaring.BitmapOf(uint32(p.RepoID))}}
-	err = zoektSearch(ctx, logger, indexed, patternInfo, branchRepos, p.NumContextLines, time.Since, p.Repo, sender)
+	err = zoektSearch(ctx, logger, indexed, &p.PatternInfo, branchRepos, p.NumContextLines, time.Since, p.Repo, sender)
```
jtibshirani, line 217: I refactored the `zoektSearch` method to work with `protocol.PatternInfo` directly, since it was cleaner.

cmd/searcher/internal/search/search_structural.go
```diff
@@ -250,17 +241,26 @@ func filteredStructuralSearch(
 	sender matchSender,
 	contextLines int32,
 ) error {
+	a, err := extractQueryAtom(p)
```
jtibshirani, line 244: This is the one hacky part of the PR! Structural search can only handle leaf patterns. However, it uses the same types `TextPatternInfo` and `protocol.PatternInfo` everywhere. Switching it over to use different types would be a big refactor, and I didn't want to invest this much in structural search. Instead, when we create a structural search we make sure to pass a single `PatternNode`, and then parse it out here.

internal/search/job/jobutil/job.go
```diff
@@ -654,43 +650,20 @@ func toSymbolSearchRequest(f query.Flat) (*searcher.SymbolSearchRequest, error)
...
 	filesInclude = append(filesInclude, mapSlice(langInclude, query.LangToFileRegexp)...)
 	filesExclude = append(filesExclude, mapSlice(langExclude, query.LangToFileRegexp)...)
+
 	selector, _ := filter.SelectPathFromString(b.FindValue(query.FieldSelect)) // Invariant: select is validated
 	count := b.MaxResults(defaultLimit)
 
-	// Ugly assumption: for a literal search, the IsRegexp member of
-	// TextPatternInfo must be set true. The logic assumes that a literal
-	// pattern is an escaped regular expression.
-	isRegexp := b.IsLiteral() || b.IsRegexp()
```
jtibshirani, line 674: Here, we converted literal searches to regexes by setting `isRegexp: true`, and passing `Pattern: b.PatternString()`, which quotes meta characters if the query is a literal. I could have tried to preserve this logic by traversing the query tree and doing these conversions. But I believe we can just remove this logic, since the searcher backend handles literals directly: https://github.com/sourcegraph/sourcegraph/blob/c83aca70e927af470ac6fac657291b0b59addebb/cmd/searcher/internal/search/matchtree.go#L46-L48

cmd/searcher/internal/search/search_structural.go
```diff
@@ -211,34 +210,26 @@ func lookupMatcher(language string) string {
...
 	}
 
 	return nil
 }
 
+func extractQueryAtom(p *protocol.PatternInfo) (*protocol.PatternNode, error) {
+	if a, ok := p.Query.(*protocol.PatternNode); ok {
+		return a, nil
+	} else {
+		return nil, errors.New("Searcher only supports leaf regex patterns")
```
camdencheek, line 229: This error seems a little misleading. Should this be "structural search only supports leaf regex patterns"?

cmd/searcher/protocol/query.go
```diff
@@ -0,0 +1,131 @@
...
+	children := make([]*proto.QueryNode, 0, len(an.Children))
+	for _, child := range an.Children {
+		children = append(children, child.ToProto())
+	}
+	return &proto.QueryNode{
+		Value: &proto.QueryNode_And{
+			And: &proto.AndNode{Children: children},
+		},
+	}
+}
```
camdencheek, line 71: Is it valuable to have this indirection layer for the query type? Or would it be better to just have `FromJobNode` return a `*proto.QueryNode` directly? 

For context, we have this pattern all over the codebase right now for two reasons:
1) The generated proto types can be awkward to use, and
2) We wanted to maintain the existing structs to minimize the blast radius of the migration to gRPC

So, if neither of those are true, it might be more ergonomic to just construct the proto message directly. Totally up to you though.

cmd/searcher/protocol/query.go
```diff
@@ -0,0 +1,131 @@
...
+		panic(fmt.Sprintf("unexpected query node type (%T)", n))
+	}
+}
+
+type AndNode struct {
+	Children []QueryNode
+}
+
+func (an *AndNode) String() string {
+	return fmt.Sprintf("AND (%d children)", len(an.Children))
```
camdencheek, line 58: nit: any reason the `String()` implementation does not recurse? for debugging, it's often very convenient to see the full contents, and I don't expect these to be extremely deep trees.

cmd/searcher/protocol/searcher_test.go
```diff
@@ -29,7 +71,57 @@ func TestRequestProtoRoundtrip(t *testing.T) {
 	}
 }
 
-func TestProtoFileMatchProtoRoundTrip(t *testing.T) {
+// We need to implement this explicitly, because quickcheck cannot generate random
+// protocol.QueryNode structs.
+type queryGenerator struct {
+	Query protocol.QueryNode
+}
```
camdencheek, line 78: thanks for implementing a generator! `testing/quick` can be really nice for things like this

internal/search/job/jobutil/job.go
```diff
@@ -119,16 +120,23 @@ func NewBasicJob(inputs *search.Inputs, b query.Basic) (job.Job, error) {
...
+						containsRefGlobs: query.ContainsRefGlobs(b.ToParseTree()),
+					})
 				}
-				addJob(&repoPagerJob{
-					child:            &reposPartialJob{searchJob},
-					repoOpts:         repoOptions,
-					containsRefGlobs: query.ContainsRefGlobs(b.ToParseTree()),
-				})
+
+				// searcher to use full deadline if timeout: set or we are not batch.
```
camdencheek, line 136: I don't understand this comment. Mind elaborating?

internal/search/job/jobutil/job.go
camdencheek: That's a very nice simplification!

internal/search/job/jobutil/job.go
jtibshirani: Oops, this comment got separated from the proper line when I shuffled things around. Will fix. It should look like this:

```go
	// searcher to use full deadline if timeout: set or we are not batch.
	useFullDeadline := f.GetTimeout() != nil || f.Count() != nil || searchInputs.Protocol != search.Batch
```

cmd/searcher/internal/search/search_structural.go
jtibshirani: Indeed, will fix this "typo"

cmd/searcher/protocol/query.go
jtibshirani: Good point, I updated this and added a test.

cmd/searcher/protocol/query.go
jtibshirani: I think (1) is in fact true: the generated types contain two layers of indirection because of how proto handles subtyping. Also, this type is used `TextPatternInfo`, which is fairly high level ... so it's nice so use a higher-level type here where we get nice `String()` output for tests, etc.

keegancsmith (approved):
Nice work!

jtibshirani:
> Are we tracking unindexed search performance of AND and OR? It would be really nice to see the before and after of this change.

This is a good question! I looked at our monitoring, including search blitz, and unfortunately don't see anything specifically for unindexed search. Maybe I could add a search blitz query for unindexed search before merging this.

cmd/searcher/internal/search/search.go
```diff
@@ -209,12 +207,9 @@ func validateParams(p *protocol.Request) error {
 	if len(p.Commit) != 40 {
 		return errors.Errorf("Commit must be resolved (Commit=%q)", p.Commit)
 	}
-	if p.Pattern == "" && p.ExcludePattern == "" && len(p.IncludePatterns) == 0 {
+	if p.Query == nil && p.ExcludePattern == "" && len(p.IncludePatterns) == 0 {
```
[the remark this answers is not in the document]
jtibshirani: Good catch! The invariant I chose was that `Query` is never nil. It's nice when interface values are guaranteed non-nil, so you don't need guards everywhere. The `FromJobNode` method enforces this.

I believe this is the only place we make the wrong assumption -- I'll fix it and add a test.

jtibshirani:
Thanks for the reviews! I will merge this after https://github.com/sourcegraph/sourcegraph/pull/59652 is merged, and we have the chance to collect data on the current unindexed search perf. I don't expect to see a big improvement, since it's likely rare for users to specify AND and OR patterns directly. But it'll be nice to check there's no regression.

PR #59577, opened 2024-01-15
Uses `src code-intel` over the deprecated `src lsif`

kritzcreek:
Just spotted this when reading through the codebase. Given that this function is commented with
```
// upload invokes `src code-intel upload` on the host and returns the graphql identifier of
// the uploaded resource.
```
this seemed like a safe change to make.

## Test plan

I'm not sure how to check this works, maybe you do?


varungandhi-src (approved):
I think `bazel test //dev/...` should run the tests which should invoke this code path?

PR #59579, opened 2024-01-15
licensing: add utility for checking license features

BolajiOlajide:
Utilities to be used to check for license features

## Test plan

Added unit tests


fkling (approved):
Accepting to unblock. I think you need to extend the type of the window context somewhere.

BolajiOlajide:
> Accepting to unblock. I think you need to extend the type of the window context somewhere.

@fkling Yes, this is dependent on #59548 getting merged that's why there are CI failures. Thanks

fkling:
I saw the change later in the other PR.

PR #59583, opened 2024-01-15
Do not forward hop-by-hop headers

rafax:
When proxying upstream requests, we shouldn't be forwarding hop-by-hop headers as defined in [RFC 2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec13.html) and redefined later. 

## Test plan

- tested locally with a fake backend


philipp-spiess (approved):
Nice!!

bobheadxi (approved):

PR #59588, opened 2024-01-15
Add percentage based rollouts for Starcoder single-tenant setups

philipp-spiess:
Based on https://github.com/sourcegraph/sourcegraph/pull/59586

Add percentage based rollout options to shift PLG and Enterprise traffic to the single-target instance

## Test plan

- Add logging for the resolved metric (e.g. `fmt.Println("resolved model", model)`)
- Test in the various combinations:
   - `cody-pro` flag disabled and no rate limit overwrites
     - no gateway feature flag overwrites: **resolved model accounts/fireworks/models/starcoder-16b-w8a16**
     - `CODY_GATEWAY_FIREWORKS_STARCODER_ENTERPRISE_SINGLE_TENANT_PERCENT=100`: **resolved model accounts/sourcegraph/models/starcoder-16b**
     - `CODY_GATEWAY_FIREWORKS_STARCODER_ENTERPRISE_SINGLE_TENANT_PERCENT=50`:
		Screenshot 2024-01-16 at 12 51 18
   - `cody-pro` flag disabled
          - no gateway feature flag overwrites: **resolved model accounts/fireworks/models/starcoder-7b-w8a16**
          - `CODY_GATEWAY_FIREWORKS_STARCODER_COMMUNITY_SINGLE_TENANT_PERCENT=100`: **resolved model accounts/sourcegraph/models/starcoder-16b**
          -  `CODY_GATEWAY_FIREWORKS_STARCODER_COMMUNITY_SINGLE_TENANT_PERCENT=50`:
                Screenshot 2024-01-16 at 12 59 26


cmd/cody-gateway/internal/httpapi/completions/fireworks.go
```diff
@@ -136,16 +148,27 @@ func (f *FireworksHandlerMethods) transformBody(body *fireworksRequest, _ string
 
 	// Enterprise virtual model string
 	if body.Model == "starcoder" {
-		body.Model = fireworks.Starcoder16b
+		d100Roll := rand.Intn(99)
```
rafax, line 151: Two points:

1. [Intn](https://pkg.go.dev/math/rand#Intn) return values from [0,n) (half-open interval), so you really need Intn(100) 
2. (nit) D100 is [1,100], this is [0,99] :)

philipp-spiess: > [Intn](https://pkg.go.dev/math/rand#Intn) return values from [0,n) (half-open interval), so you really need Intn(100)

Ohh gotcha, I only saw that it was inclusive of `0` and thought it is also inclusive of `n` 🤦 

> (nit) D100 is [1,100], this is [0,99] :)

I will rename this then, it's easier to use this range I suppose  since we need it to be:

- always false if percent is `0` 
- always true if percent is 100 

so checking for `<` makes sense here as `0 < 0` is false but `99 < 100` is true (for the end of the ranges)

rafax: Agreed that using `<` and comparing `rand.Intn(100)` to a number from [0-100] range should work.

cmd/cody-gateway/internal/httpapi/completions/fireworks.go
```diff
@@ -136,16 +148,27 @@ func (f *FireworksHandlerMethods) transformBody(body *fireworksRequest, _ string
...
+
 	// PLG virtual model strings
-	if body.Model == "starcoder-16b" {
-		body.Model = fireworks.Starcoder16b
-	}
-	if body.Model == "starcoder-7b" {
-		body.Model = fireworks.Starcoder7b
+	if (body.Model == "starcoder-16b" || body.Model == "starcoder-7b" ) {
+		d100Roll := rand.Intn(99)
+     		if (d100Roll < f.starcoderCommunitySingleTenantPercent) {
```
rafax: One thing we could do here is extract a function like `pickModelBasedOnTrafficSplit` that will reduce the 
(minor) duplication between PLG and Enterprise and can be used to document the logic around ranges / Intn params.

cmd/cody-gateway/internal/httpapi/completions/fireworks.go
```diff
@@ -247,3 +260,22 @@ func (f *FireworksHandlerMethods) parseResponseAndUsage(logger log.Logger, reqBo
 
 	return promptUsage, completionUsage
 }
+
+// Picks a model based on a specific percentage split. If the percent value is 0, the
+// zeroPercentModel is always picked. If the value is 100, the hundredPercentModel is always picked.
+func pickModelBasedOnTrafficSplit(percentage int, hundredPercentModel string, zeroPercentModel string) string {
+	// Create a value inside the range of [0, 100). This means you are doing a 1 out of 100 roll.
```
rafax, line 267:
```suggestion
	// Create a value inside the range of [0, 100).
```

cmd/cody-gateway/internal/httpapi/completions/fireworks.go
```diff
@@ -247,3 +260,22 @@ func (f *FireworksHandlerMethods) parseResponseAndUsage(logger log.Logger, reqBo
...
+	roll := rand.Intn(100)
+
+	// Check if the roll is within the target percentage:
+	//
+	// - If the percentage is `0`, the roll will always be smaller than 0 (since it's at least 0)
```
rafax, line 272: > `roll` will never be smaller than 0 

Did you mean 'roll will never be smaller than percentage' (roll will never be smaller than 0)

cmd/cody-gateway/internal/httpapi/completions/fireworks.go
```diff
@@ -247,3 +260,22 @@ func (f *FireworksHandlerMethods) parseResponseAndUsage(logger log.Logger, reqBo
...
+	// - If the percentage is `100`, the roll will always be greater than 100 (since it's at most 99)
```
rafax, line 273: Similarly, "roll will always be smaller than percentage"

cmd/cody-gateway/internal/httpapi/completions/fireworks.go
philipp-spiess: 😬

cmd/cody-gateway/internal/httpapi/completions/fireworks.go
```diff
@@ -136,16 +149,20 @@ func (f *FireworksHandlerMethods) transformBody(body *fireworksRequest, _ string
...
 	// PLG virtual model strings
-	if body.Model == "starcoder-16b" {
-		body.Model = fireworks.Starcoder16b
-	}
-	if body.Model == "starcoder-7b" {
-		body.Model = fireworks.Starcoder7b
+	//
+	// TODO: Remove the support for the full 7b MT model names here as soon as we can remove the
+	//       virtual model resolution on the SG instance in codecompletion.go
+	if body.Model == "starcoder-16b" || body.Model == "starcoder-7b" || body.Model == fireworks.Starcoder7b || body.Model == fireworks.Starcoder16b {
```
philipp-spiess, line 159: Added support for rewriting the full model strings for the MT models here. This is because of the code here: https://sourcegraph.com/github.com/sourcegraph/sourcegraph@1a3693f18b01aef62fb371a26f09386fa6c66597/-/blob/internal/completions/httpapi/codecompletion.go?L43-49

We can't remove this until we have deployed a new Cody Gateway version (since updating the instance but not Cody Gateway would leave our production traffic unresolved)

cmd/cody-gateway/internal/httpapi/completions/upstream.go
```diff
@@ -438,7 +438,7 @@ func isAllowedModel(allowedModels []string, model string) bool {
 		}
 
 		// Expand virtual model names
-		if m == "fireworks/starcoder" && (model == "fireworks/"+fireworks.Starcoder7b || model == "fireworks/"+fireworks.Starcoder16b) {
+		if m == "fireworks/starcoder" && (model == "fireworks/"+fireworks.Starcoder7b || model == "fireworks/"+fireworks.Starcoder16b || model == "fireworks/"+fireworks.Starcoder16bSingleTenant) {
```
philipp-spiess, line 441: The `fireworks/starcoder` virtual model should give access to the ST instance too

philipp-spiess (commented):
@rafax Tested this locally now and it seems good from my side. Added my tested configurations in the PR description.

philipp-spiess:
(and yup I also tested that the new ST env works as expected 😎  - Time to roll it out and send some PLG traffic there?)

PR #59591, opened 2024-01-15
msp: check matching EnvironmentDeploySpec is provided

jac:
Old code tried to check `s.Subscription.Tag`  even if `s.Subscription` was `nil`. It also did not verify whether the provided `s.Subscription` or `s.Manual` was actually expected based on `s.Type`
## Test plan

- CI
- tested locally


bobheadxi (approved):
Thanks!

PR #59598, opened 2024-01-15
msp-example: close redis client

bobheadxi:
@jac spotted a profile indicating we might have a resource leak on the clients created here. The client implements a closer, so we should use it.

## Test plan

n/a


jac (approved):
I think the a single client which gets passed as state to the diagnostics handler is probably the correct way to do this but for this example opening and closing the client for each request is _good enough_

PR #59604, opened 2024-01-15
msp-example: close BigQuery client

bobheadxi:
Similar to #59598, I think this is another leak in the example service. This is less of an issue in services that create the `Writer` once, but in general things should be shut down cleanly.

## Test plan

n/a


unknwon (approved):

cmd/msp-example/internal/example/bigquery.go
```diff
@@ -16,6 +16,7 @@ func writeBigQueryEvent(ctx context.Context, contract runtime.Contract, eventNam
 	if err != nil {
 		return errors.Wrap(err, "BigQuery.GetTableWriter")
 	}
+	defer bq.Close()
```
unknwon, line 19: nit

```suggestion
	defer func() { _ = bq.Close() }
```

PR #59615, opened 2024-01-16
Search Filters: use codeintel languages to determine language

### LINKED ISSUE #59611: Search filters: incorrect language detection
e.g. `.yaml` files are being detected as "MiniYAML", which is just weird

CleanShot 2024-01-15 at 17 30 41@2x

camdencheek:
Fixes #59611

Previously, we were using `internal/inventory` to determine the language for a file name. This led to weird things like `.yaml` files being detected as `MiniYAML`. Instead, this switches filters to use the more accurate `lib/codeintel/languages` package, which we should be standardizing on as our source of truth for language detection.

Note: the codeintel package requires the file contents for best accuracy, but we do not want to put network requests in this hot path in the search stream, nor do I want to thread a gitserver client through to this caller. Ideally, the language would be detected by our search backends and streamed back on search results instead of detecting it at the API boundary because we already have the file contents readily available. cc @sourcegraph/search-platform in case you want to tackle that at some point. 

also cc @sourcegraph/team-graph  for visibility

## Test plan

Added a test for detecting YAML as a language, also for coverage of the `lang` filter in general.

CleanShot 2024-01-15 at 19 07 11@2x

### DIFF
--- a/internal/search/streaming/search_filters.go
+++ b/internal/search/streaming/search_filters.go
@@ -2,7 +2,6 @@ package streaming
 
 import (
 	"fmt"
-	"path"
 	"strconv"
 	"strings"
 	"time"
@@ -13,9 +12,9 @@ import (
 	"github.com/grafana/regexp"
 	"github.com/sourcegraph/sourcegraph/internal/api"
 	"github.com/sourcegraph/sourcegraph/internal/gitserver/gitdomain"
-	"github.com/sourcegraph/sourcegraph/internal/inventory"
 	"github.com/sourcegraph/sourcegraph/internal/lazyregexp"
 	"github.com/sourcegraph/sourcegraph/internal/search/result"
+	"github.com/sourcegraph/sourcegraph/lib/codeintel/languages"
 )
 
 // SearchFilters computes the filters to show a user based on results.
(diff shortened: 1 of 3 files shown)


internal/search/streaming/search_filters.go
```diff
@@ -148,16 +147,22 @@ func (s *SearchFilters) Update(event SearchEvent) {
...
-				if strings.Contains(language, " ") {
-					language = strconv.Quote(language)
-				}
-				value := fmt.Sprintf(`lang:%s`, language)
-				s.filters.Add(value, rawLanguage, lineMatchCount, "lang")
+		// NOTE: this does not provide file contents to determine the language,
+		// so accuracy may be limited. Ideally, we would stream the language
+		// on results from our search backends, where the file contents are
+		// readily available.
+		languages, _ := languages.GetLanguages(fileMatchPath, nil)
```
varungandhi-src, line 154: I see that the `inventory.GetLanguageByFilename` API is also being used by `inventory.getLang` which is used in a couple of other places. Thoughts on pros and cons of replacing the internal implementation `GetLanguageByFilename` by `languages.GetLanguages`, versus getting rid of that API altogether?

camdencheek: The `inventory` code is old, buggy, and barely used anymore. I'd like to eventually get rid of that package entirely, but I'd be happy to just swap out remaining uses of `GetLanguageByFilename` and minimize its public API surface. I'll make a followup PR that does that

internal/search/streaming/search_filters.go
```diff
@@ -148,16 +147,22 @@ func (s *SearchFilters) Update(event SearchEvent) {
...
+		if len(languages) == 0 {
+			return
+		}
+		rawLanguage := languages[0]
```
varungandhi-src, line 158: Could you describe where exactly/how this is used? I remember there is one code path where we convert `lang:` filters into searches for file extensions for the the known extensions for that language, so it seems a bit odd that we have the opposite too.

I can't quite tell whether it is justified to drop the other entries in the `languages` slice if the length is greater than 1 (should there be an `OR` filter instead?)

camdencheek: Yeah, that's exactly right: this is the opposite. This is what populates the "languages" section of the filters panel. So it's taking a file and converting it into a `lang:` filter that would select that file.

In the case the slice has length greater than 1, the file name cannot uniquely identify a language, so we want to count it as the most likely language for the filters panel. Incrementing each possible language would lead to a mismatch between the counts in the filters panel and the total result count, which would reduce trust in the product. 

When the filter is clicked, it will add a `lang:` filter for that language, which will get converted to the file path regexes like you mentioned, which _should_ include a pattern that matches the original path.

Again, we should ideally _not_ be doing this translate-language-to-filename-and-back-again dance, but we're not quite there yet. I might be able to make some quick progress there since my head's already in this space though.

CleanShot 2024-01-15 at 20 43 55@2x

camdencheek:
Looking into this more closely, I think I can do better. As it turns out, Zoekt already returns a language on its search results, so we can use that when it's available. I've got a branch locally that uses that when search results come back from Zoekt and uses path heuristics when that's not available. I'll push a different PR tomorrow.

PR #59616, opened 2024-01-16
Search: add language to the `FileMatch` type

### LINKED ISSUE #59611: Search filters: incorrect language detection

camdencheek:
The language of a file cannot reliably be determined with just the file name. We have numerous examples of this failing and giving surprising results. 

We can provide much more reliable language detection if we have the file contents available to provide more information during language detection. However, fetching the contents of files to detect the language can be expensive, especially for operations like search where there may be _many_ results we would need to fetch contents for. 

Instead, this adds the language as a field on the `FileMatch` result. We already get this information from Zoekt and symbols, and it will be straightforward to add language detection in searcher, where the file contents are already available when constructing matches.

This PR replaces https://github.com/sourcegraph/sourcegraph/pull/59615
Fixes #59611 

## Test plan

Tested that this fixes incorrect languages in the filters panel. Added a couple of tests. Updated existing tests.


internal/search/result/file.go
```diff
@@ -20,6 +21,32 @@ type File struct {
...
+	PreciseLanguage string
+}
+
+// Language returns the most likely language for the file.
+// In the case that the file has PreciseLanguage set, it uses that.
+// Otherwise, it falls back to language matching on the path.
+// It will return an empty string in the case language matching fails.
+//
+// TODO(camdencheek): this lazy detection can be removed once
+// we add language detection in searcher.
```
camdencheek, line 38: If we add language detection to searcher, drop this whole method since `PreciseLanguage` should always be populated.

cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -6207,6 +6207,10 @@ type FileMatch {
     Whether or not the limit was hit.
     """
     limitHit: Boolean!
+    """
+    The most likely language for the file, or unset if the language cannot be determined.
+    """
+    language: String
```
varungandhi-src, line 6213: From an API standpoint, I think we should avoid limiting the result to 'most likely', in case the client wants to apply any additional heuristics to the results. We can let the client determine whether it wants to subset out the results or do something else.

AIUI, currently Zoekt only stores a single language for a file, but I think it should store a list for correctness.

camdencheek: > in case the client wants to apply any additional heuristics to the results

In practice, what heuristics could the client provide that we wouldn't be able to answer more accurately with the extra information we have on the backend? If we wanted the ability to add extra heuristics (e.g. "in this codebase, all .h files are actually Objective-C"), it seems like we'd be better served by allowing users to configure overrides that inform the detection of the "most likely" language. 

Lightly pushing back because I don't ever expect a Sourcegraph client to care about anything other than the first language, and I think it makes it more confusing if we only have a single string from most of our sources and always narrow down to a single string in the end.

internal/search/streaming/http/events.go
```diff
@@ -30,6 +30,7 @@ type EventContentMatch struct {
 	Hunks           []DecoratedHunk  `json:"hunks"`
 	LineMatches     []EventLineMatch `json:"lineMatches,omitempty"`
 	ChunkMatches    []ChunkMatch     `json:"chunkMatches,omitempty"`
+	Language        string           `json:"language,omitempty"`
```
varungandhi-src, line 33: Background question, not exactly related to the contents of this PR.

I thought all of our APIs were exposed as GraphQL, but it seems like based on this, we also have a (public? internal-use only?) REST API also (?), and we're maintaining the types between the two by hand instead of using code generation, is that right?

camdencheek: Yes, that's correct. The streaming search endpoint uses server-sent events, and the types are, in fact, maintained by hand. We allow customers to use it, but with a "this might change" warning label. I have it on my personal backlog to experiment with defining these types in protobuf so we can have a better-defined API here.

cmd/frontend/graphqlbackend/schema.graphql
varungandhi-src: > In practice, what heuristics could the client provide that we wouldn't be able to answer more accurately with the extra information we have on the backend?

The backend code is working with atomic units (single files), whereas client-side code has aggregate information available (folder/repo). E.g. if there are `.cpp` files and `.h` files, then it's likely that `.h` files are also C++. But sometimes C++ code is written in a C style, in which case a text classifier won't be able to distinguish between the two.

The client needs to anyways handle both the 0 and 1 cases. Handling the more than 1 case is at max an extra 2-3 lines of code.

I'll give a slightly different example. The prospect that we've been talking to which is interested in writing a CLI tool to do find references, and someone writing similar tooling might be interested in language information.

camdencheek: Got it, thanks for the example! I'll update the API

varungandhi-src: Another slightly different example. If someone reports a bug, then having a bug report with:

1. The languages list has the desired language, but it's not the first one, so it is getting dropped by the client. (Or conversely, there is a single element list, and it is missing the desired language altogether.)
2. The language returned is wrong.

Out of these, 1. is a more useful informative bug report.

varungandhi-src (approved):
> Got it, thanks for the example! I'll update the API

No other blocking concerns, so approving since you said you'll make this change. Tests look OK. 👍🏽

keegancsmith (approved):
LGTM

internal/search/result/file.go
```diff
@@ -20,6 +21,44 @@ type File struct {
...
 	Path     string
+
+	// A language, as determined by lib/codeintel/languages.
+	// May be empty, indicating a precise language is not known
+	// for this file, in which case we should fall back to
+	// language detection on the file name.
+	PreciseLanguage string
+}
+
+// Language returns the most likely language for the file.
```
BolajiOlajide, line 32:
```suggestion
// Languages returns the most likely language for the file.
```

BolajiOlajide (approved):

Issue #59352, opened 2024-01-05
☂️  Search Jobs


stefanhengl:
P0: for 5.3 release
- [x] #59329  
- [x] Make search jobs available in UI by default

P1: future release
- [ ] support type:repo
- [ ] support type:diff, type:commit ([context](https://sourcegraph.slack.com/archives/C05KARB1F33/p1703094344907599))
- [ ] #59354

jtibshirani:
5.3 release status: ✅ All set, not aware of any remaining work.

PR #59619, opened 2024-01-16
search jobs: switch from CSV to JSON

stefanhengl:
Relates to #59352 

We switch the format of the results download from CSV to line-separated JSON. 
Each line corresponds to a JSON object containing chunk matches. 
The JSON object has the same format as the matches served by the [Stream API](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@1a3693f18b01aef62fb371a26f09386fa6c66597/-/blob/internal/search/streaming/http/events.go?L12).

[example.json](https://github.com/sourcegraph/sourcegraph/files/13948979/example.json)

This is a **breaking change** motivated by customer feedback. 
Search Jobs is still released as EAP so a breaking change is acceptable. 

Pros:
- richer information (matches, content, positions)
- supports all result types
- same format as Stream API

Cons:
- Requires more storage
- Not as easy to parse by a human as a CSV

The commits can be reviewed separatedly. 
The first 2 commits contain the core of this change. 
In the third commit I delete code and update tests

Next: update documentation

## Test plan
- updated and new units tests
- manual testing
  - I used this Python script to get the result counts from the downloaded json and compared them to the counts in the web app for a couple of examples.
```python
"""
Usage:
    python main.py <path/to/json>
"""
import json
import argparse


def count_results(filename):
    return sum(len(chunks['ranges']) for line in open(filename) for chunks in json.loads(line)['chunkMatches'])


if __name__ == '__main__':
    parser = argparse.ArgumentParser()
    parser.add_argument('filename', type=str, help='Search Jobs result json')
    args = parser.parse_args()

    print(f'{count_results(args.filename)} results')
```


cmd/worker/internal/search/exhaustive_search_repo_revision.go
```diff
@@ -72,13 +72,13 @@ func (h *exhaustiveSearchRepoRevHandler) Handle(ctx context.Context, logger log.
 		return err
 	}
```
stefanhengl, line 74: This is "write" part. We replace the CSV writer with a JSON writer.

internal/search/exhaustive/service/service.go
```diff
@@ -401,6 +398,35 @@ func discardUntil(br *bufio.Reader, delim byte) error {
 	}
 }
 
+func writeSearchJobJSON(ctx context.Context, iter *iterator.Iterator[string], uploadStore uploadstore.Store, w io.Writer) (int64, error) {
```
stefanhengl, line 386: This is the "read" part. We simply concatenate the result blobs from the various repo-revisions we searched.

internal/search/exhaustive/service/matchjson.go
```diff
@@ -25,11 +25,11 @@ func NewJSONWriter(ctx context.Context, store uploadstore.Store, prefix string)
 	}
 
 	return &MatchJSONWriter{
-		w: http.NewJSONArrayBuf(1024*1024*100, blobUploader.write)}, nil
+		w: newBufferedWriter(1024*1024*100, blobUploader.write)}, nil
 }
 
 type MatchJSONWriter struct {
-	w *http.JSONArrayBuf
+	w *bufferedWriter
```
stefanhengl, line 32: switching to line-separted JSON for the internal format. See [discussion](https://github.com/sourcegraph/sourcegraph/pull/59519#issuecomment-1891563287)

cmd/frontend/internal/httpapi/httpapi.go
```diff
@@ -161,7 +161,7 @@ func NewHandler(
 	m.Path("/src-cli/{rest:.*}").Methods("GET").Handler(trace.Route(newSrcCliVersionHandler(logger)))
 	m.Path("/insights/export/{id}").Methods("GET").Handler(trace.Route(handlers.CodeInsightsDataExportHandler))
 	m.Path("/search/stream").Methods("GET").Handler(trace.Route(frontendsearch.StreamHandler(db)))
-	m.Path("/search/export/{id}.csv").Methods("GET").Handler(trace.Route(handlers.SearchJobsDataExportHandler))
+	m.Path("/search/export/{id}.json").Methods("GET").Handler(trace.Route(handlers.SearchJobsDataExportHandler))
```
keegancsmith, line 164: I guess technically we could support something which converts the json into csv in the future? I quite liked the csv support, but we should actually get feedback it is useful.

internal/search/exhaustive/service/matchjson.go
```diff
@@ -66,3 +66,56 @@ func (b *blobUploader) write(p []byte) error {
...
+	b.buf.Grow(min(flushSize, 1024*1024)) // 1 MiB
+	return b
+}
+
+// Append marshals v and adds it to the buffer. If the size of the buffer
+// exceeds flushSize the buffer is written out.
+func (j *bufferedWriter) Append(v any) error {
+	oldLen := j.buf.Len()
+
+	enc := json.NewEncoder(&j.buf)
```
keegancsmith, line 93: was wondering if it was worth saving the returned encoder for use between calls. But after reading the implementation, it seems all the real state that can be inferred between calls is stored in a sync.pool. So the only downside of this is a tiny allocation which won't outlive this function call. LGTM.

keegancsmith (approved):
LGTM

cmd/frontend/internal/httpapi/httpapi.go
stefanhengl: Agreed. I also like the CSV. However, I think it might get a unwieldy quickly once we want to support multiple result types or return individual result chunks with their position.

PR #59624, opened 2024-01-16
Enable attribution through site admin.

cbart:
Add query:

```graphql
site {
  codyConfigFeatures {
    attribution
  }
}
```
to explicitly enable attribution for users.

Pending: Corresponding client-side change.

## Test plan

GraphQL-level testing with mocked site config.


chwarwick (commented):
@slimsag I thought the config based features were going to be removed in favor of RBAC is that still the plan?

arafatkatze:
@chwarwick To my understanding that is still the plan the only thing is that we wanna have easy placeholder gQL to be able to test things on all ends really fast. Once we have that the RBAC shift is super easy(this is also why the current configfeatures exists) its only for testing.

arafatkatze (approved):
image
Tests pass fine and the GQL querying works perfectly too.

cbart:
> @slimsag I thought the config based features were going to be removed in favor of RBAC is that still the plan?

I was following @slimsag 's advice here: https://sourcegraph.slack.com/archives/C05JDP433DL/p1704931118495999?thread_ts=1704408478.471069&cid=C05JDP433DL

cbart:
But yes! @chwarwick is right that if we want to pick an API we should do it now :)

cbart:
Yep! Sorry I meant to link it in the PR description, but then forgot.

schema/site.schema.json
```diff
@@ -2909,6 +2908,10 @@
         "commands": {
           "description": "Enable/Disable special commands for the clients",
           "type": "boolean"
+        },
+        "attribution": {
+          "description": "Enable/Disable attribution search for Cody-generated snippets",
+          "type": "boolean"
```
slimsag, line 2914: Specifically in the site config schema, `configFeatures` will be removed. The GraphQL resolver and everything else will remain, but the site config schema will not.

You should put this as your own top-level config key. i.e. instead of site admins writing:

```
"configFeatures": {
    "attribution": true,
},
```

You should have them write something that is consistent with how other site config 'enabled' options look like, you can pick from one of these two options (I have no preference):

```
// Option 1
"batchChanges.enabled": true,
"search.index.symbols.enabled": true,
"codeIntelAutoIndexing.enabled": true,

// Option 2
"auth.passwordPolicy": {
    "enabled": true,
},
"auth.allowedIpAddress": {
    "enabled": true,
}
```

slimsag (approved):
All this code looks good to me, with the exception of just the site.schema.json changes I mentioned inline.

cbart:
Thanks! I'll proceed in amending the site config and then land this. ❤️

schema/site.schema.json
cbart: Done now!

arafatkatze: Makes sense @cbart thank you

Issue #59451, opened 2024-01-09
Search Results: Prepare feature flag set to the Feb release


vovakulikov:
At the moment, search result page has a lot of dev-only feature flags that we have to remove or enable by default before the Feb release. 

Short list of feature flags to be

**removed:**
- `newSearchResultsUI` - the new search result UI update is good enough to roll out it for all users without further disabled feature flag

**enable by default**
- `newSearchResultFiltersPanel` - we want to roll out new filters for all users automatically but still want to keep feature flag in case if there would be some regression (will be removed in the next release if everything is alright with most of our customers)

**refine, make it with visible toggle in the user menu**
- `newSearchNavigationUI` - the new navigation helps with vertical space and reduces amount of navigation in general but there is a counter point that we don't want to disrupt natural feature enablement (with old navigation main features like insights, monitoring and batch changes are visible in all screens in the app), so we want to make the new navigation is optional for those who want to have a simple UI, we put visual toggle UI in the user menu to simplify enablement of this new navigation (similar to what we did with new search input UI)

vovakulikov:
We should add feature flag (in order to support URL overrides) for the new features such as 
- File preview 
- Filters panel 


Ideally we should support URL overrides on experimental feature flags level as well

Issue #59386, opened 2024-01-08
Search: Enable new search result looks by default 


vovakulikov:
At the moment, newly added search result page layout is under experimental feature flag (turned off by default), since we're okay with general layout ideas and with file preview feature that we introduced there, we should enable this layout by default on this page. 

In the next release cycle after February major release, we will get rid of old layout completely if we don't encounter any big problems with new one.

Issue #59385, opened 2024-01-08
Search: Polish the new search result filter panel 


vovakulikov:
In this PR https://github.com/sourcegraph/sourcegraph/pull/59178 we added the new filter panel UI but there were some leftovers, in this issue we should fix most of them, here is a short list of important things

- Filtering the language list doesn't seem to work (it ignores case register)  
- Replace "Show all" with "Show more [X]" in the truncated filter list UI
- Use only repository name in the repo filter panel (ignore host part of URL)
- Add `count:all` filter if results got their limit 
- Make search filter input with standard size (not small)
- Repository icons should be outlined 
- Repo names should have a hover with the full name
- The numbers should respect the limitHit value returned and show a + next to the number if it’s not a complete count. I think we might also want to provide fuzzier numbers for an imcomplete count, like 1+, 5+, 10+, 50+, 100+, 500+. I think that communicates the “this number might not be complete” better than 237+.
- Fix icons in the footer link panel (@taiyab can you provide a names of icons that you used in this UI, it looks like they are not standard to mdi icons that we use? not sure)
- Mobile/tablet layout version for filter UI

camdencheek:
> Clicking a repository currently fails when the new search mode is enabled because repo: no longer takes a regex in that mode

The revert for this is [here](https://github.com/sourcegraph/sourcegraph/pull/59381)

taiyab:
> * Fix icons in the footer link panel (@taiyab can you provide a names of icons that you used in this UI, it looks like they are not standard to mdi icons that we use? not sure)

All the icons are non-standard. I want to start introducing a new icon set to the design system, specifically [this one](https://www.streamlinehq.com/icons/core-line).

How difficult/easy would this be?

vovakulikov:
@taiyab It's not difficult at all from a tech point of view, if we have SVG sources of these icons we're good. I can add some of them that we need here in filters panel UI. The problem is about consistency with other icon set that we currently use. But I think this part of the problem should be solved within design department

camdencheek:
> Add count:all filter if results got their limit

I don't think this is as obvious as it sounds. The rest of our filters are subtractive: they can only return a strict subset of the results previously returned. `count` is additive. It changes the mental model of the filters panel from reducing the set of results returned by the user-entered query.

taiyab:
@camdencheek Implementing this thing should help in alleviating some of the pain of discovering `count:all` 
[...]

PR #59641, opened 2024-01-16
Search results: Adjust search navigation flags

vovakulikov:
Closes https://github.com/sourcegraph/sourcegraph/issues/59451
Сloses https://github.com/sourcegraph/sourcegraph/issues/59386
Part of https://github.com/sourcegraph/sourcegraph/issues/59385

## Test plan
- Check that Search Filters panel is accessible to you by default (but you can turn it off by disabling `newSearchFiltersPanel: false` in the experimental settings 
- Check that the new Search Result layout is accessible by default (the feature flag was removed completely because we don't want to maintain multiple version of layouts on the page anymore)
- Check that the new search navigation panel is disabled by default but you can enable it through user menu toggle UI (the new search box UI has been removed/replaced by new search navigation toggle)


dev/linters/logging/BUILD.bazel
```diff
@@ -7,7 +7,6 @@ go_library(
     visibility = ["//visibility:public"],
     deps = [
         "//dev/linters/nolint",
-        "@com_github_openpeedeep_depguard_v2//:go_default_library",
         "@org_golang_x_tools//go/analysis",
```
vovakulikov, line 10: Changes in `/dev` directory are interesting because they were produced by `bazel configure` on this PR branch, they don't have anything related to this PR changes, looks like main got some outdated changes about it.

camdencheek: cc @sourcegraph/team-dev-infra

camdencheek (approved):

dev/linters/logging/BUILD.bazel
burmudar: @vovakulikov if I check out this branch and run `bazel configure` I don't get these changes.

what is the output of `git clean -nfx dev/linters` ?

vovakulikov: Hmm, this is weird, @burmudar I don't have any which isn't in working tree in the dev directory, git clean produces nothing, I reverted changes in the dev folder and run bazel configure again, it again produced these updates to bazel build files. 

I think I can ignore them and just don't commit, if it doesn't happen on your machine locally maybe something weird things happens on my machine only, I will try to reset the sourcegraph repo by cloning the most recent version once again.

burmudar: Ah damn. Yeah we've seen this happen with some peole and still trying to figure out why it happens 😞

PR #59643, opened 2024-01-16
Code Navigation: cache history panel data 

jasonhawkharris:
Before, we refetched data every time we visited the history panel, even if it the data hadn't changed from the last time we visited. This resulted in a sluggish experience when switching between tabs in the reference panel. 

Now, we cache `commit-history` after the first visit, and use that as the data source, resulting in a much snappier experience. 

## Test plan
CI/CD passing


jasonhawkharris:
I worked with what I knew here. Completely open to this not being the correct solution!

client/web/src/repo/RepoRevisionSidebarCommits.tsx
```diff
@@ -99,10 +101,41 @@ export const RepoRevisionSidebarCommits: FC<Props> = props => {
...
+                ls.setItem('commit-history', deserializedNodes)
+            }
+        }
+
+        const cachedNodes = ls.getItem('commit-history')
+        if (cachedNodes !== null) {
+            setCommitHistory(JSON.parse(cachedNodes))
+        }
+    }, [connection, props.filePath])
+
```
BolajiOlajide, line 134: Is there a reason we aren't leveraging Apollo's caching mechanism instead of storing this in localstorage?

camdencheek: Another path we could take is to use a react hook (`useMemo`?) to only re-fetch the history if `props.filePath` changes. In general, using local storage is a bit of a code smell since it's global and untyped.

jasonhawkharris: @BolajiOlajide I would absolutely love to do that, and would welcome any contribution to that end, whether it be advice or actual code.

camdencheek: Just to capture some progress @jasonhawkharris and I made: 
- The component is being re-rendered on ever load because it is being instantiated inside a `useMemo` with a bunch of unnecessary dependencies. Moving out into its own, independent `useMemo` helped with this
- For some reason, the query is still being re-run. React is reacting to changes in the props inside the component, but outside the component, the props are unchanged. We have no idea why this is happening, but it's enough to explain the requests being re-run.

PR #59461, opened 2024-01-10
msp/monitoring: add external uptime check and alert, rework health probes configuration

bobheadxi:
The new configuration is mostly based on Cody Gateway - if a service has an external domain, we create an uptime check and alert on failures.

The uptime check uses MSP standards, which depends on whether or not service health probes are configured. Since we use this in several places now, I've also reworked the health probes configuration to make it easier to reason with:

1. `healthProbes` now configures all healthchecks. `startupProbe` and `livenessProbe` has been removed
2. `disabled` is now `healthzProbes` - this configures if MSP healthchecks should be used, instead of default `/` ones.
3. By default, if no config is provided, MSP healthchecks are not used
4. If config is provided, MSP healthchecks must be explicitly disabled

Closes https://github.com/sourcegraph/managed-services/issues/350

This is required for our upcoming vendor evaluations as well.

This PR also includes a variety of internal improvements to alert policies.

## Test plan

https://github.com/sourcegraph/managed-services/pull/383:

image
image

https://github.com/sourcegraph/managed-services/actions/runs/7509458664

PR #59644, opened 2024-01-16
msp/spec: document requirements for healthzProbes: false

bobheadxi:
If `/-/healthz` is not implemented, services _must_ respond on `/` for external healthchecks (the Cloud Run default only does a TCP connect, which is why it generally works OOTB while the recently launched external healthchecks #59461 do not).

## Test plan

n/a

### DIFF
--- a/dev/managedservicesplatform/spec/environment.go
+++ b/dev/managedservicesplatform/spec/environment.go
@@ -407,19 +407,21 @@ type EnvironmentServiceAuthenticationSpec struct {
 
 type EnvironmentServiceHealthProbesSpec struct {
 	// HealthzProbes configures whether the MSP-standard '/-/healthz' service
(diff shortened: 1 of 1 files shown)


unknwon (approved):
Will there be a sister change in https://github.com/sourcegraph/managed-services/blob/main/schema/service.schema.json? Thought comments here are not visible to users that are writing `service.yaml`?

bobheadxi:
> Will there be a sister change in https://github.com/sourcegraph/managed-services/blob/main/schema/service.schema.json? Thought comments here are not visible to users that are writing service.yaml?

Yeah , steps here: https://github.com/sourcegraph/managed-services?tab=readme-ov-file#update-service-spec-schema - I will follow up post merge

PR #59646, opened 2024-01-16
Code Navigation: fix broken `copy` link

jasonhawkharris:
Fixes the copy link that was copying the url without a base url. 

## Test plan
CI/CD

### DIFF
--- a/client/web/src/repo/actions/CopyPermalinkAction.tsx
+++ b/client/web/src/repo/actions/CopyPermalinkAction.tsx
@@ -49,6 +49,7 @@ interface CopyPermalinkActionProps extends RepoHeaderContext, TelemetryProps {
 export const CopyPermalinkAction: React.FunctionComponent<CopyPermalinkActionProps> = props => {
     const { revision, commitID, actionType, repoName, telemetryService } = props
 
+    const rootUrl = window.location.host
     const navigate = useNavigate()
     const location = useLocation()
     const fullURL = location.pathname + location.search + location.hash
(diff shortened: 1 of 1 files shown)


client/web/src/repo/actions/CopyPermalinkAction.tsx
```diff
@@ -49,6 +49,7 @@ interface CopyPermalinkActionProps extends RepoHeaderContext, TelemetryProps {
 export const CopyPermalinkAction: React.FunctionComponent<CopyPermalinkActionProps> = props => {
     const { revision, commitID, actionType, repoName, telemetryService } = props
 
+    const rootUrl = window.location.host
```
BolajiOlajide, line 52: I think we usually make use of `window.context.externalURL` in these kind of cases.

BolajiOlajide (approved):
Asides my comment, this looks good to me.

client/web/src/repo/actions/CopyPermalinkAction.tsx
jasonhawkharris: Out of curiosity do you know why we default to that?

jasonhawkharris: ahhh... I think I see why. window.context.externalURL includes the transfer protocol.

BolajiOlajide: Unfortunately, I can't remember.

BolajiOlajide: Oh, that too.

vovakulikov: I think this might be because we used to running Sourcegraph logic in different environments where we couldn't use location because it was different from where we host sourcegraph instance.

vovakulikov (approved):

cbart:
👌

cbart (approved):

PR #59647, opened 2024-01-16
Revert "otel-collector: upgrade to 0.92.0 (#59613)"

ggilmore:
This is causing panics to occur when tracing is enabled. 

```
[       frontend] INFO frontend.overrideGlobalSettings cli/config.go:202 Global settings is already up to date, skipping writing a new entry
[       frontend] 2024/01/16 19:03:52 tmpfriend: Removing /var/folders/5g/kp3xgqds5wn7mwtcqf1vpn280000gn/T/tmpfriend-66510-2924445689
[       frontend] INFO frontend.permission_update bg/update_permissions.go:37 RBAC Permissions update {"added": 0, "deleted": 0}
[       frontend] panic: runtime error: invalid memory address or nil pointer dereference
[       frontend] [signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x10723cd98]
[       frontend] goroutine 178 [running]:
[       frontend] go.opentelemetry.io/collector/internal/sharedcomponent.(*Component[...]).Start.func1()
[       frontend] 	/Users/ggilmore/.asdf/installs/golang/1.21.4/packages/pkg/mod/go.opentelemetry.io/collector@v0.92.0/internal/sharedcomponent/sharedcomponent.go:94 +0xb8
[       frontend] sync.(*Once).doSlow(0x124fa0a68?, 0x12814a980?)
[       frontend] 	/Users/ggilmore/.asdf/installs/golang/1.21.4/go/src/sync/once.go:74 +0x100
[       frontend] sync.(*Once).Do(...)
[       frontend] 	/Users/ggilmore/.asdf/installs/golang/1.21.4/go/src/sync/once.go:65
[       frontend] go.opentelemetry.io/collector/internal/sharedcomponent.(*Component[...]).Start(0x124fa0a68?, {0x124fa0a68?, 0x12814a980?}, {0x124fa2b28?, 0x14002fd4f00?})
[       frontend] 	/Users/ggilmore/.asdf/installs/golang/1.21.4/packages/pkg/mod/go.opentelemetry.io/collector@v0.92.0/internal/sharedcomponent/sharedcomponent.go:89 +0x84
[       frontend] github.com/sourcegraph/sourcegraph/cmd/frontend/internal/app/otlpadapter.(*signalAdapter).Start(0x140034c8a80, {0x124fa0a68, 0x12814a980}, {0x124fa2b28, 0x14002fd4f00})
[...]

PR #59613, opened 2024-01-16
otel-collector: upgrade to 0.92.0

bobheadxi:
This change upgrades the Open Telemetry Collector to 0.92.0 as a follow-up to #59564 and #59561

While doing this, I found 2 copies of the `builer.template.yaml` file. I'm not sure when the Wolfi one is used and when the other is, but for now I updated both to ensure they are in sync at least.

## Test plan

CI

PR #59649, opened 2024-01-16
tracing: fix regressions from OpenTelemetry upgrades

bobheadxi:
This change reverts https://github.com/sourcegraph/sourcegraph/pull/59647 and fixes regressions from recent OpenTelemetry upgrades:

1. #59564 introduces a regression on trace propagation in gRPC
2. #59613 introduces a panic on the OTLP adapter for clients

## Test plan

```
sg start
```

image


internal/grpc/defaults/defaults.go
```diff
@@ -180,6 +181,7 @@ func buildServerOptions(logger log.Logger, opts serverOptions) []grpc.ServerOpti
 			propagator.StreamServerPropagator(requestinteraction.Propagator{}),
```
eseliger, line 181: don't you need to drop line 174 as well?

eseliger (approved):

internal/grpc/defaults/defaults.go
bobheadxi: Thanks, dropped!

eseliger (approved):

PR #59651, opened 2024-01-16
Fix format:ci arg ordering

jamesmcnamara:
The order for git-diff is from-commit to-commit, so we just had them flipped. This isn't a problem for 95% of cases, but it breaks when a file is deleted in your branch because the flipped filter will think of it as added.

## Test plan
Tested locally on branch with deleted files


Strum355 (approved):
love that this worked for every other case, besides this one case. Youd think itd fail more frequently lol

PR #59660, opened 2024-01-17
[Cody] remove cody-pro feature flag dependency

thenamankumar:
Remove dependency on `cody-pro` feature flag which was added during Dec GA release. Things are now stable and it is a good time to cleanup. 

I won't be cleaning the feature flag settings on dotcom even after this PR is merged and deployed as the clients still use that feature flag. 

Once this PR is merged, it will be good to remove the dependency from the clients as well. cc: @kalanchan 

**Update:** I created a similar PR for clients here https://github.com/sourcegraph/sourcegraph/pull/59660

## Test plan

1. sg start dotcom
2. Go to `/site-admin/feature-flags` and remove the `cody-pro` feature flag.
3. Visit `/cody/manage` and make sure it loads correctly.
4. Try upgrade and downgrade from subscriptions page and verify that limits on dashboard page are reflected correctly.
5. Visit `/cody/chat` and verify that the counters on `/cody/manage` increments.


vdavid (approved):
Thanks for the cleanup! #delete-week!

PR #59663, opened 2024-01-17
licensing: hide code-search related links when on Global Navbar Cody-only license

### LINKED ISSUE #59662: remove search + search-related links from navbar

BolajiOlajide:
Closes #59662

| Cody Only  | 
|---|
| CleanShot 2024-01-19 at 12 15 26@2x  |  

| Code Search Only  | 
|---|
| CleanShot 2024-01-17 at 11 34 04@2x |  

| Code Search + Cody  | 
|---|
| CleanShot 2024-01-17 at 11 34 04@2x |  

## Test plan

* CI checks.
* Manual testing

### DIFF
--- a/client/web/src/nav/GlobalNavbar.test.tsx
+++ b/client/web/src/nav/GlobalNavbar.test.tsx
@@ -46,10 +46,6 @@ describe('GlobalNavbar', () => {
     })
 
     test('default', () => {
-        // vi.spyOn(liceseUtils, 'isCodeSearchOnlyLicense').mockReturnValue(false)
-        // vi.spyOn(liceseUtils, 'isCodeSearchPlusCodyLicense').mockReturnValue(true)
-        // vi.spyOn(liceseUtils, 'isCodyOnlyLicense').mockReturnValue(false)
-
         vi.mock('../util/license', () => ({
             isCodeSearchOnlyLicense: () => false,
             isCodeSearchPlusCodyLicense: () => true,
--- a/client/web/src/nav/GlobalNavbar.test.tsx
+++ b/client/web/src/nav/GlobalNavbar.test.tsx
@@ -1,6 +1,6 @@
 import React from 'react'
 
-import { describe, expect, test, vi } from 'vitest'
+import { describe, expect, test, vi, afterAll, beforeAll } from 'vitest'
 
 import { MockedTestProvider } from '@sourcegraph/shared/src/testing/apollo'
 import {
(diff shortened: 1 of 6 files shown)


camdencheek (approved):

client/web/src/nav/GlobalNavbar.test.tsx
```diff
@@ -41,7 +41,47 @@ const PROPS: React.ComponentProps<typeof GlobalNavbar> = {
 }
 
 describe('GlobalNavbar', () => {
+    afterAll(() => {
+        vi.restoreAllMocks()
+    })
+
     test('default', () => {
+        vi.mock('../util/license', () => ({
```
vovakulikov, line 49: @BolajiOlajide so mocking routes didn't work out here? Mocking license might be bad because in this case we don't test the full flow and all logic about gating and licensing (I mean license utils functions)

client/web/src/nav/GlobalNavbar.tsx
```diff
@@ -140,14 +141,19 @@ export const GlobalNavbar: React.FunctionComponent<React.PropsWithChildren<Globa
...
+    const disableCodeSearchFeatures = isCodyOnlyLicense()
     // Search context management is still enabled on .com
     // but should not show in the navbar. Users can still
     // access this feature via the context dropdown.
-    const showSearchContext = searchContextsEnabled && !isSourcegraphDotCom
-    const showCodeMonitoring = codeMonitoringEnabled && !isSourcegraphDotCom
-    const showSearchNotebook = notebooksEnabled && !isSourcegraphDotCom
-    const isLicensed = !!window.context?.licenseInfo
-    const showBatchChanges = props.batchChangesEnabled && isLicensed && !isSourcegraphDotCom
+    const showSearchContext = searchContextsEnabled && !isSourcegraphDotCom && !disableCodeSearchFeatures
```
vovakulikov, line 149: This isn't blocker but licensing is another feature that we don't do well in my opinion. We have to add checks on the low level to hide/show some UI which is okay but these checks should be less complex I think, we probably need to have a centralised place where we gather some core-based checks about search/cody/specific features. 

Food for though for svelte project, how can we do it different in the fresh start. cc @fkling

vovakulikov (approved):
Left a few comments, nothing big, thanks

client/web/src/nav/GlobalNavbar.test.tsx
BolajiOlajide: Mocking the route works but won't suffice for my test case, the routes are computed based on the license features, so mocking the routes explicitly doesn't make the test case valid.

I want to test what routes are displayed in the navbar based on the license features.

PR #59668, opened 2024-01-17
Add an option for disabling HTTP/2 when communicating with Cody Gateway

rafax:
Add a way to disable HTTP2 for sourcegraph.com -> Cody Gateway communication. This should unblock experimentation with features that require low time-to-first-event for streaming requests (currently affected by GOAWAY errors, see https://github.com/sourcegraph/cody/issues/2180 and [Slack](https://sourcegraph.slack.com/archives/C05497E9MDW/p1704734964553649) for context).

This is unlikely to be the real/permanent fix, but should unblock us for now (until we find a way to reproduce and fix the issue). 
## Test plan

- tested locally 
- off-by-default

### DIFF
--- a/dev/Caddyfile
+++ b/dev/Caddyfile
@@ -18,3 +18,9 @@
 :3081 {
 	reverse_proxy 127.0.0.1:3082
 }
+
+# Caddy (:3083) -> Cody Gateway (:9992)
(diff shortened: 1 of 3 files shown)


unknwon (approved):
Assuming it works and nothing stands out to me.

PR #59671, opened 2024-01-17
bazel: remove explicit running of backend_integration_test

burmudar:
they are executed automatically by bazel as part of //...

## Test plan
CI

### DIFF
--- a/.aspect/workflows/config.yaml
+++ b/.aspect/workflows/config.yaml
@@ -31,7 +31,6 @@ tasks:
       include_eternal_tests: true
       targets:
           - //...
-          - //testing:backend_integration_test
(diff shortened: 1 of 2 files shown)


Strum355:
Are you sure? They have the `manual` tag so they shouldnt be run with `bazel test //...`

burmudar:
> Are you sure? They have the `manual` tag so they shouldnt be run with `bazel test //...`

AFAIK only the following targets are tagged as manual in //testing:
- `codeintel_integration_test_local`
- `codeintel_integration_test`

Strum355:
turns out I was thinking of something else :clueless:

Strum355 (approved):

PR #59673, opened 2024-01-17
Blob UI: Use memoize rendering behavior for the tabbed panels

vovakulikov:
https://github.com/sourcegraph/sourcegraph/assets/18492575/a98c6394-98eb-4097-9cc1-0623d94514e3

## Test plan
- Check that you load only data that you need for open tab on the initial rendering
- Check that as you switch to the new tab (if it's first render of the panel you should see loading) if it's second render you should see tab content UI immediately with no loaders

@jasonhawkharris I didn't link any issue here but if we have some can you add it to this PR description

### DIFF
--- a/client/branded/src/components/panel/TabbedPanelContent.tsx
+++ b/client/branded/src/components/panel/TabbedPanelContent.tsx
@@ -136,7 +136,7 @@ export const TabbedPanelContent = React.memo<TabbedPanelContentProps>(props => {
     const activeTab: Panel | undefined = panels[tabIndex]
 
     return (
-        <Tabs className={styles.panel} index={tabIndex} onChange={handleActiveTab}>
+        <Tabs lazy={true} behavior="memoize" className={styles.panel} index={tabIndex} onChange={handleActiveTab}>
             <TabList
                 wrapperClassName={classNames(styles.panelHeader, 'sticky-top')}
                 actions={
@@ -185,7 +185,7 @@ export const TabbedPanelContent = React.memo<TabbedPanelContentProps>(props => {
                             className={styles.tabsContent}
                             data-testid="panel-tabs-content"
                         >
-                            {id === activeTab.id ? element : null}
+                            {({ shouldRender }) => shouldRender && element}
                         </TabPanel>
                     ))
                 ) : (
--- a/client/branded/src/components/panel/TabbedPanelContent.tsx
+++ b/client/branded/src/components/panel/TabbedPanelContent.tsx
@@ -118,13 +118,15 @@ export const TabbedPanelContent = React.memo<TabbedPanelContentProps>(props => {
         [currentTabLabel, navigate, panels, pathname, search]
     )
 
-    const tabIndex = useMemo<number>(() => {
-        return panels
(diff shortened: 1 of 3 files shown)


client/branded/src/components/panel/TabbedPanelContent.tsx
```diff
@@ -119,15 +118,13 @@ export const TabbedPanelContent = React.memo<TabbedPanelContentProps>(props => {
...
-                : 0
-        )
-    }, [panels, hash, currentTabID])
+    const tabIndex = useMemo<number>(() => {
+        return panels
+            ? panels.findIndex(({ id, matchesTabID }) =>
+                  matchesTabID ? matchesTabID(currentTabID) : id === currentTabID
+              )
+            : 0
+    }, [panels, currentTabID])
```
camdencheek, line 127: This is just a simplification to remove an unnecessary `useEffect`, right? Or was this part of the cause of the problem?

client/branded/src/components/panel/TabbedPanelContent.tsx
```diff
@@ -136,7 +133,7 @@ export const TabbedPanelContent = React.memo<TabbedPanelContentProps>(props => {
     const activeTab: Panel | undefined = panels[tabIndex]
 
     return (
-        <Tabs className={styles.panel} index={tabIndex} onChange={handleActiveTab}>
+        <Tabs lazy={true} behavior="memoize" className={styles.panel} index={tabIndex} onChange={handleActiveTab}>
```
camdencheek, line 136: aha, so it's the combination of `lazy={true}` _and_ `behavior="memoize"`. Neat!

camdencheek (approved):

client/branded/src/components/panel/TabbedPanelContent.tsx
BolajiOlajide: I think it's part of the cause. The `useEffect` fires when a prop changes but the `setTabIndex` also causes a re-render , and `panels` being an array isn't exactly comparable.

My gut feeling about this lol

camdencheek: Wouldn't a change in `panels` or `currentTabID` also cause the `useMemo` to fire, causing a re-render of all dependencies of `tabIndex` anyways?

camdencheek: I guess it no longer depends on `hash`, which could make a difference 🤔

vovakulikov: @camdencheek you're right, but I think it's a mistake we had. I don't see how hash change wouldn't have an affect on panels if we take `currentTabID ` from the `hash` prop. In other words I don't believe that this solves anything here, if there would be some problem will be happy to revert this. 

>Or was this part of the cause of the problem?

Yeah, just a simplification, I don't like any reactive change with state on the consumer level of `panels` stream in this case. It should be more safe to calculate it on demand. This is something that we should always do in my opinion (if you can't not having state - then don't have it)

camdencheek: Cool, always happy to get cleanups along the way!

PR #59675, opened 2024-01-17
bazel: patches to allow use_default_shell_env for e2e testing

Strum355:
Another step towards https://github.com/sourcegraph/sourcegraph/issues/59155, previously `bazel test //...` would error at analysis time on `//client/web/src/end-to-end:e2e` due to it attempting to perform variable substitution for env vars e.g. `"HEADLESS": "$(E2E_HEADLESS)"`, for values not defined via `--define` (we only set these explicitly in .aspect/bazelrc/ci.sourcegraph.bazelrc and some `sg` targets).

By leveraging https://bazel.build/rules/lib/builtins/actions#run.use_default_shell_env, we can allow the test to read values from `--action_env` while _also_ having explicit values set via `env` macro parameter. Previously, setting `env` macro parameter would completely shadow any `--action_env` values. 
Unfortunately, we cant use `--test_env` for this, as `js_run_binary` is an action not a test (or something like that?).

We also cant do env renaming anymore, meaning we have to drop the `E2E_` prefix for some env vars. At least one script needed some reworking to accommodate that `e2e_test.sh` 

image

## Test plan

:eye: CI once again :eye:

### DIFF
--- a/dev/mocha.bzl
+++ b/dev/mocha.bzl
@@ -65,7 +65,7 @@ def mocha_test(name, tests, deps = [], args = [], data = [], env = {}, is_percy_
     ]
 
     env = dict(env, **{
-        "HEADLESS": "$$E2E_HEADLESS",
+        "HEADLESS": "$E2E_HEADLESS",
         # Add environment variable so that mocha writes its test xml
         # to the location Bazel expects.
         "MOCHA_FILE": "$$XML_OUTPUT_FILE",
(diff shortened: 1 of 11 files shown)


dev/mocha.bzl
```diff
@@ -90,7 +88,7 @@ def mocha_test(name, tests, deps = [], args = [], data = [], env = {}, is_percy_
         "INTEGRATION_TESTS": "true",
 
         # Puppeteer config
-        "DISPLAY": "$(DISPLAY)",
+        # "DISPLAY": "$(DISPLAY)",
```
burmudar, line 91: We can remove these now?

Strum355: We might want to change some of them to e.g. `"SOURCEGRAPH_BASE_URL": "$$E2E_SOURCEGRAPH_BASE_URL"` if that works, but yup!

burmudar: Ah right 👍🏼

burmudar (approved):

Strum355:
Main concern is that e2e tests dont complain about missing env vars, instead resulting in the tests failing 

https://sourcegraph.slack.com/archives/C05FDP58LRF/p1705583636603129?thread_ts=1705581394.840339&cid=C05FDP58LRF

valerybugakov (commented):
> js_run_binary is an action not a test (or something like that?)

I feel you 😀

valerybugakov (approved):
LGTM

PR #59678, opened 2024-01-17
Chore: bump default search limit to 2000

camdencheek:
5 days ago, I [bumped the default search limit from 500 to 1000](https://github.com/sourcegraph/sourcegraph/pull/59558) and there have been no apparent issues. I've looked at Zoekt memory usage, CPU usage, stream latencies, and frontend provisioning metrics and have not detected any meaningful change after bumping the limit. This is not a surprise, but provides some nice validation. 

This PR doubles the limit again to 2000. Again, I do not expect there will be significant consequences, but I will keep an eye on the same perf indicators.

## Test plan

CI. Will continue watching perf numbers.


keegancsmith (approved):
the only place I expect implications is slower searches like diff/commit. Otherwise zoekt does a _lot_ more work than this limit anyways, so I wouldn't expect much impact even if you 10x it.

keegancsmith:
> > the only place I expect implications is slower searches like diff/commit.
> 
> With the increase in limit, I wonder if we should decrease the default timeout. 20 seconds is quite a bit of work. Most any text searches should finish before then, and the time taken is a better gauge of "work" than the number of results. @keegancsmith WDYT?

Agreed. I think 20s came from when zoekt didn't exist and everything was searcher :) Having a higher limit for diff/commit sounds good to me. I don't see much benefit of lowering for text search given streaming + if searcher is used its useful to give it a bit of time.

I wouldn't go higher than say 55s as a default, since lots of installations have load balancers which kills connections at 1m unless configured otherwise.

PR #59679, opened 2024-01-17
Search: add 'AND' queries to search blitz

jtibshirani:
This adds and 'AND' query to search blitz. The performance of 'AND' queries
will be important going forward, since we're changing the default search
strategy to interpret spaces as AND instead of performing an exact match.

It adds both an indexed search (hitting Zoekt) and unindexed search (hitting
searcher).

## Test plan

Tested these queries on dot com.


eseliger (approved):

keegancsmith (commented):
lgtm. worth tracking boolean on a smaller repo like sg as well?

jtibshirani:
@keegancsmith that makes sense, I'll follow up. I think we should take an overall look at search blitz to make sure the queries reflect common usage patterns, given we're switching to the "keyword" patterntype by default. I added an item here: https://github.com/sourcegraph/sourcegraph/issues/58815.

PR #59684, opened 2024-01-17
Remove struct duplication

rafax:
Stop duplicating and packing/unpacking structs (use the structs from `config`).  [Context](https://sourcegraph.slack.com/archives/C05497E9MDW/p1701948428450979)

## Test plan

- e2e tests ;) 
- compiles correctly
- tested locally


cmd/cody-gateway/shared/config/config.go
```diff
@@ -7,13 +7,12 @@ import (
...
 
 	"github.com/sourcegraph/sourcegraph/internal/codygateway"
-	"github.com/sourcegraph/sourcegraph/internal/completions/client/fireworks"
 	"github.com/sourcegraph/sourcegraph/internal/env"
 	"github.com/sourcegraph/sourcegraph/internal/trace/policy"
 	"github.com/sourcegraph/sourcegraph/lib/errors"
 )
 
-type Config struct {
+type GatewayConfig struct {
```
unknwon, line 15: Technically this rename is not needed, since this `config` package is under `cody-gateway`, but doesn't hurt.

unknwon (approved):

cmd/cody-gateway/shared/config/config.go
rafax: Fair point, reverted

bobheadxi (approved):
Woo, thanks for the cleanup!!

Issue #57797, opened 2023-10-23
Render Jupyter Notebooks


peterguy:
# Feature request description

GitHub renders Juypter Notebooks, but Sourcegraph Code Search shows them as raw JSON.

Comparison screenshots, and to see current rendering, [search for ipynp files](https://sourcegraph.com/search?q=context%3Aglobal+file%3A%2F.*%5C.ipynb%24&patternType=standard&sm=1&groupBy=repo)

**GitHub:**
![GitHub render](https://github.com/sourcegraph/sourcegraph/assets/129280/40c09170-9ba5-4e33-bd98-ba30cacc5a3c)

**Sourcegraph Code Search:**
![Code Search render](https://github.com/sourcegraph/sourcegraph/assets/129280/ab1fee1c-bdd3-4570-bfd6-63f8bbf861a3)

# Additional context

GitHub renders them, [but not great](https://twitter.com/HamelHusain/status/1716266805097070712). Sourcegraph could take inspiration from [nbsanity](https://github.com/hamelsmu/nbsanity) to do a better job.

peterguy:
See also #10203

bevzzz:
@peterguy I've published the first version of [**`nb`**](https://github.com/bevzzz/nb/tree/main) and tried using it in the Sourcegraph's backend. Here's a short demo from my local deployment:

<details><summary>Click to expand</summary>

![recording](https://github.com/sourcegraph/sourcegraph/assets/53943884/7ce63167-4497-45ef-bf8c-eac309a95ef1)

</details>

Of course, the results need polishing (there's no additional CSS) and some features are missing at the moment.   
I am [planning to add them](https://github.com/bevzzz/nb/tree/main#roadmap-) in the coming days, but thought I'd get some feedback already.

camdencheek:
Hey @bevzzz! This is impressive work -- thank you for the contribution! I don't have time to take a closer look at this immediately, but I'll try to find some time in the next few days to dig in.

bevzzz:
That's great, @camdencheek, you'll be right in time for `v0.2.0`! 
I'm planning to release it this weekend alongside two other smaller extension packages and I'd love to hear your thoughts on it.   

Updates to the PR will follow ⏭️

bevzzz:
Hi everyone!  
Happy to announce that `nb` [v0.2.0](https://github.com/bevzzz/nb/releases/tag/v0.2.0) is now out, which means I was able to extend the renderer with several other packages.  
[...]

Issue #10203, opened 2020-04-27
Render Jupyter Notebook files (.ipynb) like markdown


dadlerj:
Requested by https://app.hubspot.com/contacts/2762526/company/554275594

>ipynb files are structured as JSON, but are not particularly useful to read in that format.

>Internally it was suggested that in rendered form, the python snippets should be appropriately highlighted as python, with code intel applied

unknwon:
https://github.com/jsvine/notebookjs and https://github.com/markedjs/marked are what I used to render Jupyter Notebook, however be aware that

>Marked does not sanitize the output HTML. Please use a sanitize library, like DOMPurify (recommended), sanitize-html or insane on the output HTML! 🚨

_https://github.com/markedjs/marked#warning--marked-does-not-sanitize-the-output-html-please-use-a-sanitize-library-like-dompurify-recommended-sanitize-html-or-insane-on-the-output-html-_

casassg:
Would be nice to have this in sourcegraph!

pdubroy:
Here's an [example](https://github.com/barbagroup/AeroPython/blob/master/lessons/01_Lesson01_sourceSink.ipynb) on GitHub.

To me it looks like a significant chunk of work would be required to deliver a high-quality solution here. Given what @sourcegraph/frontend-platform has on the roadmap for Q3, I don't think we'd be able to work on this until Q4 at the earliest.

sourcegraph-bot-2:
Heads up @taylorsperry @jasongornall - the "team/frontend-platform" label was applied to this issue.

umpox:
Re-added our team label, I think this would still be our area of ownership as we [own](https://handbook.sourcegraph.com/departments/engineering/dev/process/engineering_ownership/#frontend-platform) Non-code file view.

Some additional information:

Here's an example file: https://sourcegraph.com/github.com/donnemartin/system-design-primer/-/blob/solutions/object_oriented_design/hash_table/hash_map.ipynb

Here's that same example on GitHub: https://github.com/donnemartin/system-design-primer/blob/master/solutions/object_oriented_design/hash_table/hash_map.ipynb

We would likely want to do this on the **backend**, not client side if possible. We handle this for markdown through the `richHTML` query returned from GQL.

taylorsperry:
[Related request](https://sourcegraph.slack.com/archives/C0W2E592M/p1660318083027239)

jdorfman:
[...]

PR #59685, opened 2024-01-17
Render Jupyter Notebooks

bevzzz:
Hi everyone!  

This PR uses [`nb`](https://github.com/bevzzz/nb), a package for rendering Jupyter notebooks, to address #57797.  
I wrote `nb` because I could not find an existing implementation in Go and it also seemed like a fun project to do.   
It's still lacking some important features, but I wanted to get your feedback early on.  

Regarding styling and CSS: it is currently mirroring the [standard Jupyter theme](https://unpkg.com/browse/@jupyterlab/nbconvert-css@4.0.2/style/), and is more of a POC.
It needs some polish and should definitely be reconciled with the existing branding. I could use some help / advice here :)

[This discussion](https://github.com/sourcegraph/sourcegraph/issues/10203) has some reference links (I'll be sure to add these examples to the demo when the PR is ready).   
@umpox @pdubroy (since you posted the above examples) would you give me some pointers?

It goes without saying that I would really appreciate any feedback regarding the package itself 🙌  

## Test plan

- package `nb` has good [unit test coverage](https://codecov.io/gh/bevzzz/nb/tree/main)
- `internal/htmlutils`: unit tests in `policy_test.go`
- unit test for loading and removing MathJax script in `Markdown.tsx`
- other changes are tested manually, since there's very little logic there

## TODO
- [x] upgrade `nb` to `v0.2.0` (once it's released) as it [will add some rich formatting capabilities](https://github.com/bevzzz/nb/tree/main#roadmap-)
- [x] use `goldmark.Markdown` to render "markdown" cells in the notebooks
- [x] convert ANSI colors to HTML in "stream" output cells
- [x] add CSS styling
- [x] render mathematical expressions


bevzzz:
I think this PR is now ready to be reviewed.
In summary:
- Markdown is rendered correctly
- Code, JSON/XML, and console output (stream cells) are highlighted
- Mathematical expressions are rendered correctly

Some changes that I still have in the pipeline for `nb`:
- Support older notebook versions (`v1`, `v2`, `v3`, other minor `v.4.*`)
- Customizable CSS (class prefix / classes)

but that should not affect this PR much when the changes are released.

camdencheek (commented):
Great work! Left a few comments on the backend side of things. @vovakulikov or @fkling, could you take a look at the client changes?

client/wildcard/src/components/Markdown/Markdown.tsx
```diff
@@ -5,17 +5,29 @@ import { useNavigate } from 'react-router-dom'
...
+     *
+     * @default false
+     */
+    enableMathJax?: boolean
 }
 
 export const Markdown = React.forwardRef<HTMLElement, MarkdownProps>(
-    ({ wrapper: RootComponent = 'div', className, dangerousInnerHTML, testId }, reference) => {
+    ({ wrapper: RootComponent = 'div', className, dangerousInnerHTML, testId, enableMathJax = false }, reference) => {
+        if (enableMathJax) {
```
vovakulikov, line 28: We can't have any conditions over hooks in react, if you need to run some hook-based logic by some condition my advice would be move condition within hook implementation where conditions are allowed. 

So in this case it would be
```
//...
useMathJax({ enabled: mathJaxEnabled })
//...
```

client/wildcard/src/components/Markdown/mathjax.ts
```diff
@@ -0,0 +1,52 @@
+import { useEffect } from 'react'
+
+const mathjaxURL = 'https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js'
```
vovakulikov, line 3: Can we use a published npm package here?
So we could freeze the version of mathjax renderer and also make sure that we don't have 
any security issues. What I mean here that having URL to directly load something from CDN link
isn't secure enough if we don't have control over this remote asset. 

I think MathJax has npm publish package here https://github.com/mathjax/MathJax
NPM infamous that they also very often become target of switching original assets with 
corrupted but at least they have some level of security checks.

client/wildcard/src/components/Markdown/mathjax.ts
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * On component unmount, removes the script via useEffect destructor.
+ */
+export const useMathJax = () => {
+    useEffect(() => {
+        const mj = document.createElement('script')
+
+        mj.setAttribute('type', 'text/javascript')
+        // mj.setAttribute('src', mathjaxURL)
+        mj.setAttribute('async', '')
```
vovakulikov, line 28: If it was npm package we could use lazy-loading `import()` method instead of manually creation script tag.

client/wildcard/src/components/Markdown/mathjax.ts
vovakulikov: Even if we put aside security issues, Sourcegraph can't load anything but Sourcegraph-hosted assets since Sourcegraph could be run in on premises mode where we don't necessarily have a external DNS connections.
So in this case loading from CDN will fail

vovakulikov (dismissed):
Thanks for contribution, I left a few suggestions about MathJax loading, as soon as you can switch to the npm package instead of CDN you have greenlight from my side, thanks again.

vovakulikov (commented):
Sorry misclicked Approve and comment radio

bevzzz:
Thanks for your review, I'm going to implement these suggestions in the coming days 🙌

internal/htmlutil/syntax.go
```diff
@@ -0,0 +1,29 @@
...
+	for k, v := range origTypes {
+		if k == chroma.PreWrapper {
+			sourcegraphTypes[k] = v
+		} else {
+			sourcegraphTypes[k] = fmt.Sprintf("chroma-%s", v)
+		}
+	}
+	chroma.StandardTypes = sourcegraphTypes
+}
+
```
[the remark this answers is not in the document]
bevzzz: Yes, I've noticed those options too but misunderstood how they were applied and thought they weren't a better solution for the problem. But you're totally right, they actually simplify the config a lot.

These `ClassPrefix` and `WithCustomCSS` are also something I'm [planning to add](https://github.com/bevzzz/nb?tab=readme-ov-file#roadmap-) to `nb` too, so I'll be happy to take a closer look.

internal/htmlutil/policy.go
```diff
@@ -0,0 +1,66 @@
...
+var (
+	once sync.Once
+	p    *bluemonday.Policy
+
+	chromaOnly   = regexp.MustCompile(`^(?:chroma-[a-zA-Z0-9\-]+)|chroma$`)
+	chromaOrAnsi = regexp.MustCompile(`^(?:(chroma|ansi)-[a-zA-Z0-9\-]+)|chroma$`)
+)
+
+// policy configures a standard HTML sanitization policy.
+func policy() *bluemonday.Policy {
```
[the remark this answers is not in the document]
bevzzz: They are indeed. 
And I like how with `sync.OnceValue` you can have multiple objects like that (configured once) without concocting names like `onceThis` and `onceThat`

internal/htmlutil/syntax.go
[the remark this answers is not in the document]
bevzzz: UPD: after taking a closer look I found that:

- `WithCustomCSS` does not affect the class names, only the `style` attribute (if either inline styles or `CSSWriter` are used, which is not our case)
- `ClassPrefix` is applied to _all_ token types, including `chroma.PreWrapper`, for which we really want to use the unmodified `"chroma"` class -- basically what the code above handles

Seeing how these 2 options do not offer enough flexibility, I think we're better off sticking with the current approach.

> I'd prefer this stays with the code that actually uses it.

With this PR there will now be 2 places that rely on this logic, as both Markdown and Jupyter renderers will use `chroma` for syntax highlighting internally.  So, while I share your reluctance to modify global state in an external package,  I don't know of a better way to do this for all of its clients.  

Perhaps we could restructure the code so that these modification are applied exactly before `chroma` is used:

```go
var patchChromaClasses = sync.OnceFunc(func() {
  // Add `chroma-` prefix here
})

func SyntaxHighlightingOptions() []html.Option {
  patchChromaClasses()
  return []html.Option{...}
}
```

That should make it easier to `keep track of why it exists in the first place`.

WDYT?

internal/htmlutil/policy.go
```diff
@@ -0,0 +1,66 @@
...
+		p.AddTargetBlankToFullyQualifiedLinks(true)
+	})
+	return p
+}
+
+// Policy returns the standard sanitization policty that can be further extended.
+func Policy() *bluemonday.Policy {
+	cp := *policy() // copy to avoid modifying the original policy.
+	return &cp
+}
```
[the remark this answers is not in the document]
bevzzz: picked 👌

client/wildcard/src/components/Markdown/mathjax.ts
bevzzz: I see. Loading from CDN was my first option cuz it allowed me to showcase this feature quickly, but I expected it to have a number of drawbacks.

Looking into this

bevzzz: I've updated this to not rely on a CDN to get the scripts. Instead, we now [host our own copy](https://docs.mathjax.org/en/latest/web/hosting.html#hosting-your-own-copy-of-mathjax) of the assets.

It is still  loaded via a `script` tag because I found this to be a bit more straight forward and in-line with the documentation. There are [examples](https://github.com/mathjax/MathJax-demos-node) for using `mathjax` as a library in a Node application; those would require us to configure some MathJax's internals and it just felt like something we should offload to a pre-built component if we can.

Wrt lazy-loading: as far as I understand, the MathJax assets are also loaded "lazily", since the tag is only added to DOM when the file is opened (please correct me if I'm wrong)

client/wildcard/src/components/Markdown/mathjax.ts
bevzzz: Addressed this in the comment above ⬆️

bevzzz:
After adding `mathjax-full` to the project I've ran into an issue with committing the related changes to `pnpm-lock.yaml`, because one of the added entries uses syntax that `pre-commit-hooks` [falsely marks as invalid](https://github.com/pre-commit/pre-commit-hooks/issues/984). 

I didn't want to skip it with `--no-verify` (the next person changing this file would've run into the same issue)  so I added `pnpm-lock.yaml` to the list of files excluded from the `check-yaml` hook.  
Because a) the pre-commit's maintainers were very clear about not doing anything about this (see the ticket) and b) `pnpm-lock` is a generated file, so the chances of it being incorrectly formatted are rather low.

client/wildcard/src/components/Markdown/mathjax.ts
bevzzz: What do you think about this solution?

internal/htmlutil/syntax.go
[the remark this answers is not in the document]
bevzzz: I implemented the change in [this commit](https://github.com/sourcegraph/sourcegraph/pull/59685/commits/c62f192af8a6810528b4ac1ea48d718e06ba430a) to illustrate. Let me know if you want to keep it or switch to a different one

bevzzz:
@camdencheek @vovakulikov thanks again for your reviews 🙌 
I've implemented your suggestions and also updated `nb` to `v0.3.0`, which supports earlier Jupyter Notebook versions (`v4.0`, `v3.0`, `v2.0`, and `v1.0`).

Let me know if you think the PR needs anything else. Cheers!

camdencheek (approved):

camdencheek:
Hey @bevzzz! Really sorry about taking so long to get back to ya. I went ahead and just pushed a commit with some changes so we can get this merged.

In short:
- I removed the MathJax dependency for now. It was causing trouble with our Svelte rewrite and added quite a bit of complexity to builds and distribution. We will look at adding it back later, but for now just getting a rendered version of the code in notebooks is a big win.
- I moved the styles to global so they can be applied in the Svelte webapp in addition to the React webapp
- I cleaned up a handful of lint errors and bazel build errors

Will merge as soon as I can appease CI!

bevzzz:
Hey @camdencheek! Nvmnd, I am glad you had the time to look at the PR again.  
Regarding the Mathjax, it's a bit hairy, yes. Should you need help with adding it back to the project, let me know, I'd be happy to contribute.  

Thanks for handling the CI issues too -- some errors there seem to be slightly "above my paygrade" (`unauthorized upload` in the `scip-` jobs and such)

camdencheek:
@bevzzz It's merged and [live on sourcegraph.com](https://sourcegraph.com/github.com/langchain-ai/langchain@56c6b5868b54754509934d18d9a43f510cb63121/-/blob/cookbook/smart_llm.ipynb)! Sorry, we seem to be having issues merging PRs from forks, so I had to recreate this #62583.

Thanks again for all your work on this -- I know at least a few people who will be very excited about the ability to view jupyter notebooks on Sourcegraph 🙂

bevzzz:
That's so cool! I also know a couple of people who're gonna love it :) 
Thanks, @camdencheek

jdorfman:
Thank you @bevzzz for your contribution! We gave you a shout-out: done! https://sourcegraph.com/blog/code-search-now-supports-jupyter-notebook-rendering

Would love to send you some swag too: community@sourcegraph.com

bevzzz:
@jdorfman that's very kind of you! I've sent an email to community@sourcegraph.com.

Glad I could contribute :)

PR #59686, opened 2024-01-17
Search: rename newStandardRC1 -> keyword

jtibshirani:
This renames the `newStandardRC1` patterntype to `keyword` before the release,
to better reflect the meaning of the patterntype.

It also renames the feature flag `search-new-keyword` to `search-keyword`.

Finally as a follow up to #59529 it also renames `codyContext` -> `codycontext` to
make it clear capitalization doesn't matter, and makes sure to update the GraphQL
API.

## Test plan

Covered by existing tests.


jtibshirani:
Let me know if you have an alternate proposal for the name :)

keegancsmith (approved):
LGTM!

PR #59690, opened 2024-01-18
msp/loadbalancer: add http-to-https proxy

bobheadxi:


bobheadxi:
Looks like it's better to do this in Cloudflare directly (we found the configuration for it) - doing it here would cost us 0.025 * 24 * 30 = ~$20/MSP env/month

PR #59691, opened 2024-01-18
grpc: fix retry callback logic

ggilmore:
The `retry` package's WithOnRetryCallback functionality should only run a user defined callback if the library is about to retry a request. Howevever, there was a a pre-existing bug in the library that ran the callback whenever any sort of error occurred - regardless of whether the error was actually a retriable one. (Example: the callback would fire even if the error was a context cancellation error that can never be retried). 

This had some unintended side-effects, including inaccurate Prometheus / tracing metrics (which relied on this functioanity). 

This PR fixes the issue - now the callback only fires when there is a retriable error. 

## Test plan

CI (new tests)

### DIFF
--- a/internal/grpc/retry/retry.go
+++ b/internal/grpc/retry/retry.go
@@ -261,7 +261,6 @@ func (s *serverStreamingRetryingStream) SendMsg(m any) error {
 }
 
 func (s *serverStreamingRetryingStream) CloseSend() error {
-
 	s.mu.Lock()
 	s.wasClosedSend = true
 	s.mu.Unlock()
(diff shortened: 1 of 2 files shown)


ggilmore:
* `main`
  * **PR #59691** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/59691?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/59691?utm_source=stack-comment).

eseliger (approved):

PR #59699, opened 2024-01-18
release: 5.3.0 config

keegancsmith:
These are the dates based on what Kalan posted in announce-engineering. I wrote them based on my own timezone.

Test Plan: n/a


dev/release/release-config.jsonc
```diff
@@ -22,6 +22,21 @@
       "current": "5.2.0",
       "captainGitHubUsername": "unknwon",
       "captainSlackUsername": "joe"
+    },
+    "5.3.0": {
+      "codeFreezeDate": "2024-02-09T10:00:00.000+02:00",
+      "securityApprovalDate": "2024-02-02T10:00:00.000+02:00",
```
keegancsmith, line 28: I just made this the day before releaseDate. I am not sure what is a good day for this? I kind of do it as I am prepping the final release, but I suppose it is good practice to get them to atleast vet an RC by this date?

BolajiOlajide: Security must approve the final RC before we can finalize the release. Going with the same approach we took the last time, we''ll be continually deploying the RC to a test instance for QA so it'd be redundant to be asking for approval every day.

I'd change this to be the same date as the release. date (13th February, 2024)

BolajiOlajide (approved):

dev/release/release-config.jsonc
keegancsmith:
```suggestion
      "securityApprovalDate": "2024-02-13T10:00:00.000+02:00",
```

PR #59702, opened 2024-01-18
search: hide regex toggle for keyword search

stefanhengl:
The regex toggle introduces an insonsitency in combination with keyword search, because spaces are interpreted differenly in both modes. This hides the toggle and encourages users to use our inline regex `/.../` instead.

Test plan:
manual testing

Kapture 2024-01-18 at 12 21 13


keegancsmith (approved):

jtibshirani (approved):
Looks much cleaner! We can see if we get any feedback on this from S2.

stefanhengl:
Discussed with @jtibshirani in a a sync meeting. This will be postponed to a later release.

PR #59703, opened 2024-01-18
sg/msp add senty to ops docs

jac:
Add link to service Sentry project in handbook ops page
## Test plan

CI

### DIFF
--- a/dev/managedservicesplatform/operationdocs/testdata/TestRender/basic.golden
+++ b/dev/managedservicesplatform/operationdocs/testdata/TestRender/basic.golden
@@ -30,6 +30,7 @@ If you need assistance with MSP infrastructure, reach out to the [Core Services]
 | Category   | **test**                                                                                            |
 | Resources  |                                                                                                     |
 | Alerts     | [GCP monitoring](https://console.cloud.google.com/monitoring/alerting?project=msp-testbed-dev-xxxx) |
+| Sentry     | [`msp-testbed-dev`](https://sourcegraph.sentry.io/projects/msp-testbed-dev/)                        |
 
 MSP infrastructure access needs to be requested using Entitle for time-bound privileges. Test environments may have less stringent requirements.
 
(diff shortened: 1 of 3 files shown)


dev/managedservicesplatform/operationdocs/testdata/TestRender/basic.golden
```diff
@@ -30,6 +30,7 @@ If you need assistance with MSP infrastructure, reach out to the [Core Services]
 | Category   | **test**                                                                                            |
 | Resources  |                                                                                                     |
 | Alerts     | [GCP monitoring](https://console.cloud.google.com/monitoring/alerting?project=msp-testbed-dev-xxxx) |
+| Sentry     | [`msp-testbed-dev`](https://sourcegraph.sentry.io/projects/msp-testbed-dev/)                        |
```
unknwon, line 33: CleanShot 2024-01-18 at 11 34 54@2x

Is it because I don't have access? 🤔

unknwon (commented):
Assuming it works

unknwon (approved):

dev/managedservicesplatform/operationdocs/testdata/TestRender/basic.golden
jac: The `msp-testbed` environments are `test` and `robert`. There isn't a `dev` environment even though that is what the test is set up to use

unknwon: 🤔 Not sure I follow. Would it be confusing to render a link that we know it's gonna broken, or is `msp-testbed` just a special case?

jac: The Go test uses an environment that doesn't actually exist so any link is going to be broken.

We could switch to using `msp-testbed-test` instead of `msp-testbed-dev` and then stuff would work

unknwon: Ahhh, I see. Gotcha!

bobheadxi: Ah, yeah the test isn't "real" MSP services - we might want to make the test case use a realistic setup

PR #59707, opened 2024-01-18
search: update query examples on homepage

stefanhengl:
This updates the examples on the homepage depending on whether keyword search is activated or not.

Test plan:
manual testing

image

### DIFF
--- a/client/branded/src/search-ui/components/QueryExamples.tsx
+++ b/client/branded/src/search-ui/components/QueryExamples.tsx
@@ -7,41 +7,41 @@ import { SearchPatternType } from '@sourcegraph/shared/src/graphql-operations'
 import type { TelemetryProps } from '@sourcegraph/shared/src/telemetry/telemetryService'
 import { buildSearchURLQuery } from '@sourcegraph/shared/src/util/url'
 import {
-    H2,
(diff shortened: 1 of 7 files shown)


client/branded/src/search-ui/components/QueryExamples.tsx
```diff
@@ -17,14 +29,20 @@ import styles from './QueryExamples.module.scss'
...
     selectedSearchContextSpec?: string
     isSourcegraphDotCom?: boolean
+    keywordSearch?: boolean
 }
 
 export const QueryExamples: React.FunctionComponent<QueryExamplesProps> = ({
     selectedSearchContextSpec,
     telemetryService,
     isSourcegraphDotCom = false,
+    keywordSearch = false,
```
jtibshirani, line 39: Small comment: the more common naming pattern seems to be `keywordSearchEnabled`

jtibshirani (approved):
The logic looks good to me.

Documenting my understanding of the behavior: the home page examples depend on the state of the keyword search toggle. If it's disabled, then we show the old examples.

rrhyne (approved):
Looks great to me.

PR #59708, opened 2024-01-18
bazel: fetch code-intel-extensions bundle as bazel repository

Strum355:
Reimplement fetching & building code-intel-extension bundle as a bazel `http_archive` repository & refactored `js_run_binary`. This should allow the extensions bundle to be cached in the repository cache, instead of being affected by changing action environments

## Test plan

`bazel build //client/browser:code-intel-extensions` and `cd client/browser && pnpm run create-source-zip` run without error

### DIFF
--- a/client/shared/dev/buildCodeIntelExtensions.js
+++ b/client/shared/dev/buildCodeIntelExtensions.js
@@ -12,7 +12,7 @@ const bundlesRepoName = 'sourcegraph-extensions-bundles'
  * extensions bundles to the specified path. These bundles may be included into IDE/browser extensions bundles
  * in order not to require access to the extensions registry for the code intel features to work.
  */
-function buildCodeIntelExtensions({ pathToExtensionBundles, revision }) {
+function fetchAndBuildCodeIntelExtensions({ pathToExtensionBundles, revision }) {
   const pathToDistributionRevisionFile = path.join(pathToExtensionBundles, 'revision.txt')
 
   const currentRevision =
@@ -32,6 +32,7 @@ function buildCodeIntelExtensions({ pathToExtensionBundles, revision }) {
   if (result.code !== 0) {
     console.error('Curl command failed with exit code:', result.code)
     console.error('Error message:', result.stderr)
+    return
   }
 
   // when repo archive is unpacked the leading 'v' from tag is trimmed: v1.0.0.zip => sourcegraph-extensions-bundles-1.0.0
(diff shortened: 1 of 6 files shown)


shell.nix
```diff
@@ -71,6 +71,7 @@ mkShell {
   # The packages in the `buildInputs` list will be added to the PATH in our shell
   nativeBuildInputs = with pkgs; [
     bashInteractive
+    zip
```
jhchabran, line 74: Is that accidental?

jhchabran (approved):
Looks good to me, I'll let @valerybugakov take a look at the client code, though nothing particular jumped at me when I read it.

burmudar:
big brain energy

client/shared/dev/build_code_intel_extensions.bzl
```diff
@@ -17,12 +16,10 @@ def build_code_intel_extensions(name, out, revision):
         log_level = "info",
         silent_on_success = False,
         args = [
-            revision,
+            "$(execpath @sourcegraph_extensions_bundle//:bundle)",
```
valerybugakov, line 19: Noice

client/shared/dev/buildCodeIntelExtensions.js
```diff
@@ -92,30 +107,25 @@ function buildCodeIntelExtensions({ pathToExtensionBundles, revision }) {
...
 function main(args) {
   if (args.length !== 2) {
-    throw new Error('Usage: <revision> <outputPath>')
+    throw new Error('Usage: <inputPath> <outputPath>')
   }
 
-  const [revision, outputPath] = args
+  const [inputPath, outputPath] = args
   const output = path.join(process.cwd(), outputPath)
+  const input = path.join(process.env['JS_BINARY__EXECROOT'], inputPath)
```
valerybugakov, line 126: Would this throw outside of Bazel when `process.env['JS_BINARY__EXECROOT'] === undefined`?

valerybugakov (approved):

client/shared/dev/buildCodeIntelExtensions.js
valerybugakov: It would be good to verify that the whole script works outside of Bazel locally since we do not remove these pnpm commands.

valerybugakov: You're right. I assumed the whole file could be executed outside of Bazel.

PR #58675, opened 2023-11-30
OpenCodeGraph (WIP)

sqs:
See https://docs.google.com/document/d/1AmSH0fiA2RdJpKdLfA5UJQbtuOeSsyTEoCRht5lW53w/edit.

## Test plan

TODO

Issue #58007, opened 2023-10-31
Double line selection when line wrapping is enabled


fkling:
#### Steps to reproduce:

1. Enable line wrapping in the blob view
2. Selected a line

#### Expected behavior:

Only the line is selected.

#### Actual behavior:

It looks like two lines are selected:

![2023-10-31_09-51](https://github.com/sourcegraph/sourcegraph/assets/179026/79d8f627-e7bd-4712-baaa-f587b0e84601)

This seems to be an issue with how draw line selections via a separate layer. Unclear whether it's a bug in our code or in CodeMirror. It's worth upgrading CodeMirror to the latest version to see if the bug still exists.

jasonhawkharris:
@fkling I have not been able to replicate this behavior yet. Could this be browser-specific? What browser were you using when testing? I've tried in Firefox Safari and Chrome and haven't been able to repro.

fkling:
Mmh... it only seems to happen for some lines. E.g. the one in the screenshot: https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/cmd/frontend/internal/own/resolvers/resolvers.go?L145

jasonhawkharris:
@fkling https://www.loom.com/share/cc16795e997b4eb497800c486717c7e2

jasonhawkharris:
@fkling can you try replicating this once more? I haven't been able to do so.

fkling:
Sorry, I didn't see the updates. I can replicate it in Firefox on that specific file, but not in Chrome. If it's a super isolated issue then it might not be worth spending time on it now.

jasonhawkharris:
I'll look into it. I think it's definitely within the scope of the work we're doing.

jasonhawkharris:
I found some really weird behavior when comparing the DOM in firefox vs safari/chrome/arc. Firefox seems to add a third div to the `selected-lines-layer` element when selecting lines where the other browsers only add 2. Video: 

https://www.loom.com/share/558259054f2844cf99149203e540e0e4

jasonhawkharris:
Here's where the selected-lines-layer is generated: https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/client/web/src/repo/blob/codemirror/linenumbers.ts?L141-172

PR #47378, opened 2023-02-03
blob: fix selected lines and text styles conflict

taras-yemets:
Closes https://github.com/sourcegraph/sourcegraph/issues/47364

1. Refactors selected lines highlighting to use CodeMirror `layer` extension instead of the line decorations.
With this approach, both selected lines and editor selection layers may be visible (with the latter taking precedence).
It makes selected text highlighted even if it is on a selected line. This approach is similar to how JetBrains IDE handles line breakpoint and text selection styles conflict (video attached to [this comment](https://github.com/sourcegraph/sourcegraph/pull/47378#issuecomment-1416079025)). 
2. Introduces another CSS variable: `--code-selection-bg-2`. It's used for user-selected text in the code editor. Its value is taken from my current VSCode theme as I couldn't find the suitable color in [`colors.scss`](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/client/wildcard/src/global-styles/colors.scss?subtree=true).

Feedback on design (colors) and implementation is highly appreciated!

https://user-images.githubusercontent.com/25318659/216649664-494d17d9-f930-4588-a337-2b307c58894f.mov

## Test plan
Tested manually (video attached).

## App preview:

- [Web](https://sg-web-taras-yemets-cm-blob-fix-selected.onrender.com/search)

Check out the [client app preview documentation](https://docs.sourcegraph.com/dev/how-to/client_pr_previews) to learn more.

PR #46191, opened 2023-01-06
Make line-number layout always align to top

vovakulikov:
With center aligned numbers in multiline mode it's hard to read the numbers from top to bottom and map it to actual line of code on the right. So make it always top aligned, this should help with visual mapping.

| Before | After |
| -------- | ------- |
| Screenshot 2023-01-05 at 23 36 55 | Screenshot 2023-01-05 at 23 36 23 |

## Test plan
- Check that line numbers layout looks correct in the blob view with line wrapping enabled/disabled

## App preview:

- [Web](https://sg-web-vk-fix-multiline-linenumber.onrender.com/search)

Check out the [client app preview documentation](https://docs.sourcegraph.com/dev/how-to/client_pr_previews) to learn more.

PR #59712, opened 2024-01-18
blob: Fix text selection visibility over selected lines and line number alignment

### LINKED ISSUE #59096: Line highlight is the same as user selection highlight
**Sourcegraph version: 255040_2023-12-17_5.2-222da82631d3 (dotcom)**

When selecting text in a file that is already highlighted by the line selection, the selected line highlight is the same as the mouse/keyboard selection highlight, so the user cannot tell what the user is selecting using the mouse or keyboard.

Contrast this behavior with Google Code Search, where the line selection highlight is a different shade than the mouse/keyboard selection highlight, so the user can clearly see what the user is selecting.

https://github.com/sourcegraph/sourcegraph/assets/129280/402efd93-3564-43b1-a46e-c64045520b82

fkling:
Closes #59096

It's been reported multiple times that text selection conflicts with line selection, i.e. the selected line background color would be rendered on top of text selection background color and thus it seemed that no text was selected.

## A little history

I remembered that this was fixed at some point so I was curious why it didn't work anymore.

The original fix was made in #47378. The `selected-line` CSS class was removed from the selected line decoration (effectively removing the background color) and instead the selected line(s) background was drawn with [CodeMirror's layer feature](https://codemirror.net/docs/ref/#view.layer). Layers are drawn *behind* the text and therefore not conflicting with the text selection color.

#47378 added *back* the `selected-line` CSS class to the selected line decoration (I'm not quite sure why) but also set the `.cm-line.selected-line`'s background color to `transparent` to keep the text selection visible. I.e. it was still the layer that would draw the selected line's background color.

#58675 removed `background: transparent` because due to how the layer's rectangle marker was set up, it would only cover the actual text of the line, not the whole line. This causes issues with block widgets that appeared in the line above the text. They would not appear to be part of the selected line:

By accident this seemingly innocent change brought back the text selection issue. 

## Rectangle marker computation problems

Reading the [documentation for `RectangleMarker.forRange`](https://codemirror.net/docs/ref/#view.RectangleMarker^forRange) it seems reasonable to assume that the computed rectangle would cover the area of the whole line. However, it seems that it really just covers the area of the text itself (the documentation does mention "content"). This was tried to counteract by adding margins and paddings but there where always situations where this didn't look right, especially with the blame view open, which changes the height of each line:

## Solution

As can been seen in this PR it's possible to derive the real boundaries of a line, so that the rectangle marker can be drawn correctly. Removing the `selected-line` class from the selected line decoration doesn't appear to result in any visual differences, which, once again, fixes the text selection problem.
Because the rectangle marker is now drawn on line boundaries, block widgets are "covered" by it as well and there are no small shifts anymore.

https://github.com/sourcegraph/sourcegraph/assets/179026/fbe5d877-fa8d-4eab-bf48-35e2953f8aed

## Bonus: Line number and folding marker alignment

One thing I've noticed (and had noticed before) is that folding markers are not aligned well when the blame view is open. And I was surprised to see that line numbers seem wrongly aligned when line wrapping was turned on.

There is also a little bit of history here:

- #44287 always centered line numbers so that it looks properly aligned with the blame view enabled (blame view increases line height). But that didn't work well with line wrapping.
- #46191 somewhat reverted that change, making line numbers top aligned again. That fixed line wrapping but "broke" the blame view again.
- #58675 on the other hand caused line numbers to be always bottom aligned. I assume that was done to make the line number align with the text and not with the block widget that appears before it. However this caused multiple problems:
  - With line wrapping enabled it appears like there is an unnumbered line.
    
  - The folding indicator is potentially completely "detached" from its line number if line wrapping is turned on or a block widget is shown.
    
  - Misalignment in the blame view
    

### Solution

The simplest solution is to not mess with the line number alignment at all. That means in the opencodegraph use case the line number will be shown next to the line number, which I think is OK (there might also be a way to render the opencodegraph widget between two lines if that's more desirable).
For the blame view it appears to suffice to set the line number and folding gutter elements to the same line height as the code line. 

## Test plan

Manual testing, see video.

Opencodegraph block widget: 

I also verified that https://github.com/sourcegraph/sourcegraph/issues/58007 isn't reintroduced.

### DIFF
--- a/client/web/src/repo/blob/codemirror/blame-decorations.tsx
+++ b/client/web/src/repo/blob/codemirror/blame-decorations.tsx
@@ -139,7 +139,6 @@ const blameDecorationTheme = EditorView.theme({
         left: '0',
         height: '100%',
         display: 'inline-block',
-        background: 'var(--body-bg)',
         verticalAlign: 'bottom',
         width: 'var(--blame-decoration-width)',
 
@@ -156,6 +155,15 @@ const blameDecorationTheme = EditorView.theme({
     },
     '.no-line-break-msg': {
         marginLeft: 'var(--blame-decoration-width)',
+    },
+
+    // This ensures that line numbers and folding controls are reasonably aligned
+    // with the top of the line, even if line wrapping is enabled.
+    '.cm-lineNumbers .cm-gutterElement': {
+        lineHeight: '1.5rem',
+    },
+    '.cm-foldGutter .cm-gutterElement': {
+        lineHeight: '1.5rem',
     },
 })
 
--- a/client/web/src/end-to-end/code-intel/repository-component.test.ts
+++ b/client/web/src/end-to-end/code-intel/repository-component.test.ts
@@ -499,7 +499,7 @@ describe('Repository component', () => {
                 await link.click()
 
                 const selectedLine = await driver.page.waitForSelector(
-                    `[data-testid="repo-blob"] .cm-line:nth-child(${line}).selected-line`
+                    `[data-testid="repo-blob"] .cm-line:nth-child(${line})[data-testid="selected-line"]`
                 )
                 expect(selectedLine).not.toBeNull()
             })
--- a/client/web/src/repo/blob/codemirror/linenumbers.ts
+++ b/client/web/src/repo/blob/codemirror/linenumbers.ts
@@ -1,6 +1,5 @@
 import {
     Annotation,
-    EditorSelection,
     type Extension,
     type Range,
     RangeSet,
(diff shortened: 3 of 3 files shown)


client/web/src/repo/blob/codemirror/blame-decorations.tsx
```diff
@@ -139,7 +139,6 @@ const blameDecorationTheme = EditorView.theme({
         left: '0',
         height: '100%',
         display: 'inline-block',
-        background: 'var(--body-bg)',
```
fkling, line 142: This didn't seem to be necessary and removing it actually lets the selected line color "come through".

peterguy (approved):
I would approve this PR based on the comprehensive description alone, but the code changes look good also!

PR #59713, opened 2024-01-18
Perf: cache whether subrepo perms are enabled for a repo

camdencheek:
Currently, in the `SubRepoPermsJob`, we filter each search event in the stream serially. This is problematic because these checks require a database call, which means we're making a serial database call for each unique repo in our search results. We pay this cost whether or not the repo is a perforce repo because we're making the DB call specifically to check whether the repo is a perforce repo. This is currently adding ~2ms per unique repo in a set of search results when subrepo permissions are enabled (such as on S2).

Example trace: 

image

We had various layers of caching of this information that I consolidated into a cache on the checker itself so the caller doesn't need to worry about cache logic and so the cache can be shared between dependents of a client. Whether a repo is a perforce repo is likely to be _quite_ stable, so I set a default cache TTL of 1 hour. This can realistically probably be higher, but I'd rather err on the side of caution since this is a security feature. The cache itself is very dense since it uses roaring bitmaps. I am not concerned about memory usage given that we already have a much larger cache for the actual rules. 

Since the cache lives on the client, it persists for the lifetime of the client. In most cases, this is the lifetime of the process since we use the global `DefaultSubRepoPermsClient` in most cases.

## Test plan

Added unit tests for the cache itself. We've got a handful of backend integration tests that cover this pretty well. The change itself is small. I'll do some followup checks of S2 traces to double check that they do, in fact, look better than the screenshot above.

### DIFF
--- a/internal/authz/subrepoperms/sub_repo_perms.go
+++ b/internal/authz/subrepoperms/sub_repo_perms.go
@@ -48,8 +48,14 @@ type SubRepoPermsClient struct {
 }
 
 const (
-	defaultCacheSize           = 1000
-	defaultCacheTTL            = 10 * time.Second
+	defaultCacheSize = 1000
+	defaultCacheTTL  = 10 * time.Second
+
+	// We maintain a relatively high TTL for the repo enabled cache
+	// because it only stores whether the repo is a perforce repo and
+	// whether the repo is private. These properties should not change
+	// frequently, and are notably not updated when repo permissions
+	// are enabled/disabled.
 	defaultRepoEnabledCacheTTL = time.Hour
 )
 
@@ -354,6 +360,7 @@ func (s *SubRepoPermsClient) EnabledForRepoID(ctx context.Context, id api.RepoID
 		return enabled, nil
 	}
 
+	// TODO(camdencheek): Ideally, refreshing the cache could be handled in a background goroutine so it never blocks a caller
 	enabled, err := s.permissionsGetter.RepoIDSupported(ctx, id)
 	if err != nil {
 		return false, err
--- a/internal/authz/subrepoperms/sub_repo_perms.go
+++ b/internal/authz/subrepoperms/sub_repo_perms.go
@@ -421,7 +421,7 @@ func NewSimpleChecker(repo api.RepoName, paths []string) authz.SubRepoPermission
 type repoEnabledCache struct {
 	mu        sync.Mutex
 	ttl       time.Duration
-	createdAt time.Time
+	lastReset time.Time
 	enabled   *roaring.Bitmap
 	valid     *roaring.Bitmap
 }
@@ -429,7 +429,7 @@ type repoEnabledCache struct {
 func newRepoEnabledCache(ttl time.Duration) repoEnabledCache {
 	return repoEnabledCache{
 		ttl:       ttl,
-		createdAt: time.Now(),
+		lastReset: time.Now(),
 		enabled:   roaring.NewBitmap(),
 		valid:     roaring.NewBitmap(),
 	}
@@ -458,8 +458,9 @@ func (r *repoEnabledCache) SetRepoIsEnabled(repoID api.RepoID, enabled bool) {
 
 // r.mu must be held to safely call this method
 func (r *repoEnabledCache) resetIfExpired() {
-	if time.Since(r.createdAt) >= r.ttl {
+	if time.Since(r.lastReset) >= r.ttl {
 		r.valid.Clear()
 		r.enabled.Clear()
 	}
+	r.lastReset = time.Now()
 }
--- a/internal/authz/subrepoperms/sub_repo_perms.go
+++ b/internal/authz/subrepoperms/sub_repo_perms.go
@@ -5,8 +5,10 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
+	"github.com/RoaringBitmap/roaring"
 	"github.com/gobwas/glob"
 	lru "github.com/hashicorp/golang-lru/v2"
 	"github.com/prometheus/client_golang/prometheus"
@@ -39,14 +41,16 @@ type SubRepoPermsClient struct {
 	clock             func() time.Time
 	since             func(time.Time) time.Duration
 
-	group   *singleflight.Group
-	cache   *lru.Cache[int32, cachedRules]
-	enabled *atomic.Bool
+	group            *singleflight.Group
+	cache            *lru.Cache[int32, cachedRules]
+	enabled          *atomic.Bool
+	repoEnabledCache repoEnabledCache
 }
 
 const (
-	defaultCacheSize = 1000
-	defaultCacheTTL  = 10 * time.Second
+	defaultCacheSize           = 1000
+	defaultCacheTTL            = 10 * time.Second
+	defaultRepoEnabledCacheTTL = time.Hour
 )
 
 // cachedRules caches the perms rules known for a particular user by repo.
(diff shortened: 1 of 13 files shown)


internal/search/job/jobutil/sub_repo_perms_job.go
```diff
@@ -85,33 +85,23 @@ func applySubRepoFiltering(ctx context.Context, checker authz.SubRepoPermissionC
 	// Filter matches in place
 	filtered := matches[:0]
 
-	subRepoPermsCache := map[api.RepoName]bool{}
 	errCache := map[api.RepoName]struct{}{} // cache repos that errored
```
camdencheek, line 88: This cache is no longer necessary. I keep the error cache because in the new implementation, we just do not save the result if the DB call failed. This should be pretty rare, so I'm a little surprised that we need an `errCache`.

internal/own/background/recent_contributors.go
```diff
@@ -90,26 +85,3 @@ func (r *recentContributorsIndexer) indexRepo(ctx context.Context, repoId api.Re
 	commitCounter.Add(float64(len(commitLog)))
 	return nil
 }
-
-func isSubRepoPermsRepo(ctx context.Context, repoID api.RepoID, cache rcache.Cache, checker authz.SubRepoPermissionChecker) (bool, error) {
```
camdencheek, line 94: This function is purely a caching wrapper around `authz.SubRepoEnabledForRepoID`, which now handles its own caching.

internal/authz/subrepoperms/sub_repo_perms.go
```diff
@@ -401,3 +417,50 @@ func NewSimpleChecker(repo api.RepoName, paths []string) authz.SubRepoPermission
 	getter.RepoIDSupportedFunc.SetDefaultReturn(true, nil)
 	return NewSubRepoPermsClient(getter)
 }
+
+type repoEnabledCache struct {
+	mu        sync.Mutex
```
eseliger, line 422: not sure what's the value in practice, but since this might be shared across multiple goroutines, there might be a very slight perf benefit of using an `RWMutex` here, but no blocker.

internal/authz/subrepoperms/sub_repo_perms.go
```diff
@@ -401,3 +417,50 @@ func NewSimpleChecker(repo api.RepoName, paths []string) authz.SubRepoPermission
...
+	}
+}
+
+// r.mu must be held to safely call this method
+func (r *repoEnabledCache) resetIfExpired() {
+	if time.Since(r.lastReset) >= r.ttl {
+		r.valid.Clear()
+		r.enabled.Clear()
+	}
+	r.lastReset = time.Now()
```
eseliger, line 465: for later, but wondering if we couldn't even dump ALL repos in here on reset or in a background worker. even for 4m repos, the amount of memory required is fairly manageable, maybe 10MB.

internal/authz/subrepoperms/sub_repo_perms.go
```diff
@@ -39,14 +41,16 @@ type SubRepoPermsClient struct {
...
+	enabled          *atomic.Bool
+	repoEnabledCache repoEnabledCache
 }
 
 const (
-	defaultCacheSize = 1000
-	defaultCacheTTL  = 10 * time.Second
+	defaultCacheSize           = 1000
+	defaultCacheTTL            = 10 * time.Second
+	defaultRepoEnabledCacheTTL = time.Hour
```
eseliger, line 53: I think when switching on subrepo perms on the code host, this could mean that it takes 1h to reflect. Not necessarily a blocker and also not 100% sure here, but wondering if this could cause some hidden problems. The gains are significant though so I'm thinking this should be fine. 
Might be cool to document this though?

Another side note, the method that returns whether it's enabled for a repo currently returns `false, nil` for repos that aren't found in the DB, so this could cause some "invalid" state in the cache, although realistically unlikely. Not sure we need to do something about it, just figured I'd call it out.

eseliger (approved):
Thanks a lot, that's a great improvement and also makes the code a bunch simpler 👍

internal/authz/subrepoperms/sub_repo_perms.go
keegancsmith: I reckon we can be much more conservative here and cache for 30s? An extra DB call once every 30s seems fine? Maybe even add some jitter if we are worried about every repo getting invalidated at the same time.

internal/authz/subrepoperms/sub_repo_perms.go
```diff
@@ -128,6 +132,7 @@ func NewSubRepoPermsClient(permissionsGetter SubRepoPermissionsGetter) *SubRepoP
 		group:             &singleflight.Group{},
 		cache:             cache,
 		enabled:           enabled,
+		repoEnabledCache:  newRepoEnabledCache(defaultRepoEnabledCacheTTL),
```
keegancsmith, line 135: do we create a new subrepopermsclient per request? If so, might it be better to use a global singleton for the cache? Or is the cost low enough that we can pay it once per query?

internal/authz/subrepoperms/sub_repo_perms.go
keegancsmith: @eseliger RWMutex is tricky due to cache invalidation upgrading a read to a write.

keegancsmith (approved):
nice

internal/authz/subrepoperms/sub_repo_perms_test.go
```diff
@@ -417,3 +418,25 @@ func TestSubRepoPermsPermissionsCache(t *testing.T) {
 		t.Fatal("Should have been called twice")
 	}
 }
+
+func Test_repoEnabledCache(t *testing.T) {
```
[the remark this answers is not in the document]
camdencheek: oh, yup. Old habits die hard

internal/authz/subrepoperms/sub_repo_perms.go
camdencheek: Yeah, the upgrade path is why I decided not to implement it this way in the first place. And I don't expect there to be much contention here since mthe operations are very fast and mostly sequential from a single client anyways.

internal/authz/subrepoperms/sub_repo_perms.go
camdencheek: Yeah, that definitely seems ideal. I didn't feel like messing with goroutine lifecycle management for something as  low-hanging as this, which is why I didn't go that direction, but feel free to take this further if you'd like 🙂

internal/authz/subrepoperms/sub_repo_perms.go
camdencheek: >  switching on subrepo perms on the code host

I don't _think_ that's the case since if you look at the [SQL that is run](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@e445eed8ce888e63923cd289728f3691f305d15b/-/blob/internal/database/sub_repo_perms_store.go?L229-234), it's just checking 1) that the repo is private, and 2) that the repo is a perforce repo. I don't expect those to change very often. 

IIUC, enabling subrepo permissions should only affect calculating the rules, which touches a _different_ cache. This one just handles "are subrepo permissions even possibly enabled for this repo?" Maybe we should clarify the wording a bit 🙂

camdencheek: > An extra DB call once every 30s seems fine?

It would be, but this is a potential N extra DB calls every 30s, where N is the number of repos on the instance

PR #59715, opened 2024-01-19
Search preview panel: use static highlight ranges from search results

camdencheek:
Not all searches can be represented as a single regex pattern, so we can't easily translate a search query into a code mirror search that shows the same results. However, we already have the set of matched ranges from our search results, so we can use that to provide a static set of highlights in the preview panel.

This PR does exactly that. Note that it implements the change within the CodeMirror extensions framework, so it won't require much change to get it working in svelte as well.

[Demo loom](https://www.loom.com/share/f53b9dab21404d9a8d06e173addc1f7e)

~Remaining TODOs:~
1) ~We should scroll the preview panel to the first match. I couldn't quite figure out how to apply an effect on extension application though, so I could use some help there.~
2) ~A nice-to-have would be to retain the search panel's previous/next functionality even if the ranges are static~

## Test plan

Manual testing. What would be the appropriate types of tests to add for a PR like this?


client/web/src/repo/blob/codemirror/static-highlights.tsx
```diff
@@ -0,0 +1,251 @@
...
+}
+
+function findSelectedIdx(ranges: HighlightedRange[]): number | undefined {
+    const selectedIdx = ranges.findIndex(range => range.selected)
+    return selectedIdx === -1 ? undefined : selectedIdx
+}
+
+export const STATIC_HIGHLIGHTS_CONTAINER_ID = 'static-highlights-navigation-container'
+
+class StaticHighlightsPanel implements Panel {
```
camdencheek, line 139: The panel structure is largely copy/pasted from the search panel. I just heavily pruned it to only the relevant pieces.

client/web/src/repo/blob/codemirror/static-highlights.tsx
```diff
@@ -0,0 +1,251 @@
...
+                    from: toCodeMirrorLocation(state, range.start),
+                    to: toCodeMirrorLocation(state, range.end),
+                }))
+            ),
+            showPanel.of(view => new StaticHighlightsPanel(view, navigate)),
+            scrollToFirstRange(),
+        ],
+    })
+    return facet.of(ranges)
+}
```
camdencheek, line 54: This is the primary entrypoint: a facet that provides four extensions: a theme, state, a panel, and a helper that scrolls to the first match on open.

client/web/src/repo/blob/codemirror/static-highlights.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+/**
+ * scrollToFirstRange is a small utility extension that just scrolls to
+ * the selected range on startup.
+ */
+function scrollToFirstRange(): Extension {
```
camdencheek, line 60: This felt a little hacky, but it was the best way I could figure out how to do it. If anyone knows of a better way, let me know.

client/web/src/repo/blob/codemirror/static-highlights.tsx
```diff
@@ -0,0 +1,251 @@
...
+            Decoration.set(
+                ranges.map(({ selected, from, to }) =>
+                    selected
+                        ? staticHighlightSelectedDecoration.range(from, to)
+                        : staticHighlightDecoration.range(from, to)
+                ),
+                true
+            )
+        ),
+})
```
camdencheek, line 116: The state calculates the current selected range from the main selection, then uses that information to provide a set of decorations.

client/web/src/repo/blob/codemirror/static-highlights.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+const staticHighlightTheme = EditorView.theme({
+    '.cm-sg-static-highlight': {
+        backgroundColor: 'var(--mark-bg)',
+    },
+    '.cm-sg-static-highlight-selected': {
+        backgroundColor: 'var(--oc-orange-3)',
+    },
```
camdencheek, line 124: Copied from the search implementation.

client/web/src/repo/blob/codemirror/static-highlights.tsx
vovakulikov: If I understood this correctly the problem comes from the fact that we don't have a good hook/place to run/dispatch scrollToAction for the panel. This is something that I also struggled with when I was trying 
to do a first implementation of the panel. If you run this too soon in constructor it won't work obviously but if it's too late you will see jump state from line 0 to line what you need to scroll 

cc @fkling maybe you know the idiomatic code mirror way to scroll to the line logic?

client/web/src/repo/blob/codemirror/static-highlights.tsx
```diff
@@ -0,0 +1,251 @@
...
+ * staticHighlights is an extension that highlights a static set of ranges
+ * and opens a panel that allows navigation between these highlights.
+ */
+export function staticHighlights(navigate: NavigateFunction, ranges: Range[]): Extension {
+    const facet = Facet.define<Range[], Range[]>({
+        combine: ranges => ranges.flat(),
+        enables: () => [
+            staticHighlightTheme,
+            staticHighlightState.init(state =>
+                sortBy(ranges, ['from']).map((range, i) => ({
```
vovakulikov, line 43: Not sure this is a problem here or something else, but sorting probably doesn't work as expected in some cases, for example here the 

https://github.com/sourcegraph/sourcegraph/assets/18492575/1c95e8a9-f49b-4cdc-886f-38d6dc71183a

vovakulikov (commented):
Found one problem with matches ordering

client/web/src/repo/blob/codemirror/static-highlights.tsx
camdencheek: Yep, that's exactly what it is. I'll just sort after the map.

camdencheek: Fixed in the latest commit 👍

vovakulikov (approved):

PR #59719, opened 2024-01-19
Add creator and updater info to external service byline

gabtorre:
- Added creator and updater field to external service byline if available
- Updated the byline to handle external services differently:
  - Display creator if available
    - Last Updater if available
  - Show "Last synced" instead of "Updated." This differentiates when external services are synced automatically versus edited directly by a user.

## Test plan
Update existing tests.


client/web/src/components/externalServices/ExternalServicePage.tsx
```diff
@@ -165,6 +165,7 @@ export const ExternalServicePage: FC<Props> = props => {
             )
         }
         const externalServiceCategory = resolveExternalServiceCategory(externalService)
+        const noAuthor: boolean | undefined = externalService.creator === null ? true : undefined
```
pjlast, line 168: Why `undefined` instead of just `false`?

pjlast: Also, since we're adding a `createdBy` field, that tells us if there's an author or not. So we can get rid of the `noAuthor` field entirely

pjlast (approved):
Looks good! Just asking about the `noAuthor` field there because it seems like that information is now encoded in the `createdBy` field, so if we can get rid of a redundant field that would be great.

As a side note: screenshots are always nice with changes like this where there's a visual aspect to the change.

client/web/src/components/externalServices/ExternalServicePage.tsx
gabtorre: Thanks for the feedback! noAuthor used nullish coalescing, so it was set up to handle null or undefined. I've switched it to a ternary for better clarity. 
Also, I removed the extra noAuthor and rendered the author field differently for external services, so it won't show as 'a deleted user' when the author is null for existing connections.

Screenshot 2024-01-23 at 2 42 40 PM
Screenshot 2024-01-23 at 2 39 47 PM

pjlast (approved):
Awesome!

PR #59723, opened 2024-01-19
Expose 'sourcegraph-accounts-dev' as external sign-in option

chrsmith:
I would like to add a new, hidden option login option for Sourcegraph.com. We will soon add a `Sourcegraph-accounts` identity provider, but to support testing being able to use the -dev instance from Sourcegraph.com is required. (See https://github.com/sourcegraph/accounts.sourcegraph.com/issues/298 for more details.)

I just searched for `sourcegraph-operator` and made similar changes for a new `sourcegraph-accounts-dev`, but I wasn't able to verify the new option showed up when running locally. (So either I missed something or wasn't able to get the local registration configured.)

@unknwon , @bobheadxi is this something you folks can help me out with? 😬 👉 👈 ?

## Test plan

I tried to test the changes out locally, but couldn't confirm it worked as expected.

I think this might be by design, since somehow the list of available identity providers comes from the backend. But even after adding the following snippet to `jscontext.go`, I still didn't see the new option.

```go
	// Super awesome hack to confirm that sourcegraph-accounts-dev is wired up.
	authProviders = append(authProviders, authProviderInfo{
		IsBuiltin:         true,
		DisplayName:       "Sourcegraph Accounts (dev) #yolo-edition",
		DisplayPrefix:     nil,
		ServiceType:       "mock",
		AuthenticationURL: "https://accounts.sgdev.org/auth",
		ServiceID:         "service-id",
		ClientID:          "client-id",
	})
```

### DIFF
--- a/client/web/src/auth/SignInPage.tsx
+++ b/client/web/src/auth/SignInPage.tsx
@@ -60,7 +60,7 @@ export const SignInPage: React.FunctionComponent<React.PropsWithChildren<SignInP
             if (provider.serviceType !== 'openidconnect') {
                 return false
             }
-            if (provider.displayName.indexOf("Sourcegraph Accounts (dev)") == -1) {
+            if (!provider.displayName.includes('Sourcegraph Accounts (dev)')) {
                 return false
             }
             return true
--- a/client/web/src/auth/SignInPage.tsx
+++ b/client/web/src/auth/SignInPage.tsx
@@ -56,8 +56,21 @@ export const SignInPage: React.FunctionComponent<React.PropsWithChildren<SignInP
     )
 
     const shouldShowProvider = function (provider: AuthProvider): boolean {
+        const isSourcegraphAccountsDev = (provider: AuthProvider): boolean => {
+            if (provider.serviceType !== 'openidconnect') {
+                return false
+            }
+            if (provider.displayName.indexOf("Sourcegraph Accounts (dev)") == -1) {
+                return false
+            }
+            return true
+        }
+
         // Hide the Sourcegraph Operator authentication provider by default because it is
         // not useful to customer users and may even cause confusion.
+        if (isSourcegraphAccountsDev(provider)) {
+            return searchParams.has('sourcegraph-accounts-dev')
+        }
         if (provider.serviceType === 'sourcegraph-operator') {
             return searchParams.has('sourcegraph-operator')
         }
--- a/client/web/src/jscontext.ts
+++ b/client/web/src/jscontext.ts
@@ -19,7 +19,6 @@ export interface AuthProvider {
         | 'http-header'
         | 'openidconnect'
         | 'sourcegraph-operator'
-        | 'sourcegraph-accounts-dev'
         | 'saml'
         | 'builtin'
         | 'gerrit'
--- a/client/web/src/jscontext.ts
+++ b/client/web/src/jscontext.ts
@@ -19,6 +19,7 @@ export interface AuthProvider {
         | 'http-header'
         | 'openidconnect'
         | 'sourcegraph-operator'
+        | 'sourcegraph-accounts-dev'
         | 'saml'
         | 'builtin'
         | 'gerrit'
--- a/client/web/src/user/settings/auth/ExternalAccountsSignIn.tsx
+++ b/client/web/src/user/settings/auth/ExternalAccountsSignIn.tsx
@@ -37,7 +37,6 @@ const getNormalizedAccounts = (
     if (
         authProvider.serviceType === 'builtin' ||
         authProvider.serviceType === 'http-header' ||
-        authProvider.serviceType === 'sourcegraph-accounts-dev' ||
         authProvider.serviceType === 'sourcegraph-operator'
     ) {
         return []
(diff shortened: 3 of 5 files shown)


client/web/src/user/settings/auth/ExternalAccountsSignIn.tsx
```diff
@@ -37,6 +37,7 @@ const getNormalizedAccounts = (
     if (
         authProvider.serviceType === 'builtin' ||
         authProvider.serviceType === 'http-header' ||
+        authProvider.serviceType === 'sourcegraph-accounts-dev' ||
         authProvider.serviceType === 'sourcegraph-operator'
```
bobheadxi, line 41: @unknwon I think we don't need to add `sourcegraph-accounts-dev` here right? Only a user who knew about how to sign in would have this external account

I'm also surprised `sourcegraph-operator` is here so I might be off the mark - but it seems useful to be able to see

client/web/src/auth/SignInPage.tsx
```diff
@@ -58,6 +58,9 @@ export const SignInPage: React.FunctionComponent<React.PropsWithChildren<SignInP
     const shouldShowProvider = function (provider: AuthProvider): boolean {
         // Hide the Sourcegraph Operator authentication provider by default because it is
         // not useful to customer users and may even cause confusion.
+        if (provider.serviceType === 'sourcegraph-accounts-dev') {
+            return searchParams.has('sourcegraph-accounts-dev')
+        }
```
unknwon, line 63: I think we can't use `provider.serviceType` in its current form because the type of SAMS is actually `openidconnect`. We need to do something like 

```ts
if (provider.serviceId.hasPrefix('https://accounts.sgdev.org')) {
	return searchParams.has('sourcegraph-accounts-dev')
}
```

client/web/src/components/externalAccounts/externalAccounts.ts
```diff
@@ -11,7 +11,8 @@ import { GerritIcon } from '../externalServices/GerritIcon'
 
 export type ExternalAccountKind = Exclude<
     AuthProvider['serviceType'],
-    'http-header' | 'builtin' | 'sourcegraph-operator'
+    'http-header' | 'builtin' |
+    'sourcegraph-accounts-dev' | 'sourcegraph-operator'
```
unknwon, line 15: No need to update here and other places

client/web/src/user/settings/auth/ExternalAccountsSignIn.tsx
unknwon: See my comment: https://github.com/sourcegraph/sourcegraph/pull/59723#discussion_r1459716726

client/web/src/auth/SignInPage.tsx
chrsmith: OK, I'll make the change to have the SAMS-dev cases be part of where `serviceType === 'openidconnect'`, but it does seem a little odd because the definition for `serviceType` is:

```ts
        | 'github'
        | 'gitlab'
        | 'bitbucketCloud'
        | 'http-header'
        | 'openidconnect'
```

So presumably {`github`, `gitlab`, `bitbucketCloud`} should all be just `oauth`? In other words, it looks like `AuthProvider.serviceType` is already a list of _specific integrations_ and not the actual type of service being used.

I presume this was just an oversight from a long time ago, and not something to be worried about now. (And moving forward, having the `serviceType` be `openidconnect` would be the correct thing to use.)

chrsmith:
@unknwon , @bobheadxi PTAL.

I reworked the code so that we will show the sign-in option IFF the service type is `'openidconnect'` and the display name contains `'Sourcegraph Accounts (dev)'`. (Which I constraints how we need to register the SAMS identity provider in the Sourcegraph dotcom instance.)

Does this look right now? I didn't try running this locally, since I can only assume there are other registration steps that are required in order for this new query param to be honored. (e.g. the SAMS-dev identity provider is already registered, and passed to the `SignInPage` via some sort of frontend runtime config.)

client/web/src/jscontext.ts
```diff
@@ -19,6 +19,7 @@ export interface AuthProvider {
         | 'http-header'
         | 'openidconnect'
         | 'sourcegraph-operator'
+        | 'sourcegraph-accounts-dev'
```
unknwon, line 22: Place 1

client/web/src/user/settings/auth/ExternalAccountsSignIn.tsx
```diff
@@ -37,6 +37,7 @@ const getNormalizedAccounts = (
     if (
         authProvider.serviceType === 'builtin' ||
         authProvider.serviceType === 'http-header' ||
+        authProvider.serviceType === 'sourcegraph-accounts-dev' ||
```
unknwon, line 40: Place 2

unknwon (approved):
Thanks for the followup! I think all we need now is to revert the two "Places" and we are good to go.

The `serviceType` is `openidconnect`, there is no `sourcegraph-accounts-dev`.

unknwon:
No worries at all 😁 let's go!

bobheadxi (approved):

chrsmith:
Ugh, so CI is failing because I added new tests but didn't regenerate `/client/web/src/auth/__snapshots__/SignInPage.test.tsx.snap`. I'll do that now.

chrsmith:
For posterity, to update the page's snapshot file I ran:

```zsh
npx vitest --dir ./client/web/src/auth 
```

And then pressed `u` when prompted and `vitest` updated the snapshot file.

PR #59724, opened 2024-01-19
msp/runtime: use scoped loggers for healthz, expose EnvironmentID in contract

bobheadxi:
As titled. Also downgrades healthz failures to `warn` - we have additional alerting for if this fails persistently at an infrastructure level.

## Test plan

n/a

### DIFF
--- a/lib/managedservicesplatform/runtime/contract.go
+++ b/lib/managedservicesplatform/runtime/contract.go
@@ -22,7 +22,8 @@ type Contract struct {
 	// environment. In local development, this should generally be false.
 	MSP bool
 	// EnvironmentID is the ID of the MSP environment this service is deployed
-	// in.
+	// in. In local development, this should be 'unknown' if ENVIRONMENT_ID is
+	// not set.
 	EnvironmentID string
 	// Port is the port the service must listen on.
 	Port int
--- a/lib/managedservicesplatform/runtime/contract.go
+++ b/lib/managedservicesplatform/runtime/contract.go
@@ -138,7 +138,9 @@ func (c Contract) RegisterDiagnosticsHandlers(r HandlerRegisterer, state Service
 				diagnosticsLogger.Scoped("healthz"))
 
 			if err := state.Healthy(r.Context(), r.URL.Query()); err != nil {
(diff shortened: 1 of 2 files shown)


lib/managedservicesplatform/runtime/contract.go
```diff
@@ -21,6 +21,9 @@ type Contract struct {
 	// MSP indicates if we are running in a live Managed Services Platform
 	// environment. In local development, this should generally be false.
 	MSP bool
+	// EnvironmentID is the ID of the MSP environment this service is deployed
+	// in.
+	EnvironmentID string
```
bobheadxi, line 26: @chrsmith since you asked for this - @jac added this as an internal var in a previous PR, this just makes that same var public for service implementors

unknwon (approved):

lib/managedservicesplatform/runtime/contract.go
chrsmith: Thanks for exposing this!

So just to confirm, we would expect this to be `dev` or `prod` for SAMS and it is defined here:

https://github.com/sourcegraph/managed-services/blob/main/services/sams/service.yaml#L16
https://github.com/sourcegraph/managed-services/blob/main/services/sams/service.yaml#L85

And when running locally, I presume it would be `""`? Would it make sense to call this out in the comment? (e.g. `Will be "" if MSP is false` or something?)

chrsmith (approved):

lib/managedservicesplatform/runtime/contract.go
bobheadxi: > So just to confirm, we would expect this to be dev or prod for SAMS and it is defined here:

Yep!

> And when running locally, I presume it would be ""? Would it make sense to call this out in the comment? (e.g. Will be "" if MSP is false or something?)

I added a docstring - it will be "unknown" if ENVIRONMENT_ID is not set. In MSP, we just magically populate the env var for you to provide this value (most of the MSP contract is like this)

chrsmith: You should have the default value be "unknwon" and just pretend like it is a typo.

PR #59728, opened 2024-01-21
licensing: hide code-search features on user and org settings page when on Cody-only license

### LINKED ISSUE #59626: Remove search + search-related links from settings page

BolajiOlajide:
Closes #59626

When an instance runs a Cody-only license, we want to hide code-search-related features to avoid confusion. The links in the Settings page (both User and Org) shouldn't contain code-search-related settings like:

* Batch Changes
* Executor Secrets
* Saved Searches
e.t.c

[Walkthrough / Demo](https://www.loom.com/share/c7832f37f6014ce5883ad3adbc857a09?sid=cb3d94df-9fa7-4a5a-a957-8ebfd7debb30)

## Test plan

* Configure your local dev instance to run a Cody-only license.
* The settings page shouldn't contain any links to code-search-related features

### DIFF
--- a/client/web/src/LegacyLayout.tsx
+++ b/client/web/src/LegacyLayout.tsx
@@ -96,7 +96,7 @@ export const LegacyLayout: FC<LegacyLayoutProps> = props => {
     // so that Layout can always render the navbar.
     const needsSiteInit = window.context?.needsSiteInit
     const disableFeedbackSurvey = window.context?.disableFeedbackSurvey
-    const isSiteInit = location.pathname === PageRoutes.SiteAdminInit
+    const isSiteInit = location.pathname === PageRoutes.SiteAdminInit.toString()
     const isSignInOrUp =
         routeMatch &&
         [
@@ -106,8 +106,8 @@ export const LegacyLayout: FC<LegacyLayoutProps> = props => {
             PageRoutes.Welcome,
             PageRoutes.RequestAccess,
         ].includes(routeMatch as PageRoutes)
-    const isGetCodyPage = location.pathname === PageRoutes.GetCody
-    const isPostSignUpPage = location.pathname === PageRoutes.PostSignUp
+    const isGetCodyPage = location.pathname === PageRoutes.GetCody.toString()
+    const isPostSignUpPage = location.pathname === PageRoutes.PostSignUp.toString()
 
     const [newSearchNavigation] = useNewSearchNavigation()
     const [enableContrastCompliantSyntaxHighlighting] = useFeatureFlag('contrast-compliant-syntax-highlighting')
--- a/client/web/src/LegacyLayout.tsx
+++ b/client/web/src/LegacyLayout.tsx
@@ -188,6 +188,7 @@ export const LegacyLayout: FC<LegacyLayoutProps> = props => {
         !isSearchNotebooksPage &&
         !isCodySearchPage &&
         !isSearchJobsPage
+    console.log({ props })
 
     return (
         <div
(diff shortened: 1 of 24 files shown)


client/web/src/LegacyLayout.tsx
```diff
@@ -96,7 +96,7 @@ export const LegacyLayout: FC<LegacyLayoutProps> = props => {
     // so that Layout can always render the navbar.
     const needsSiteInit = window.context?.needsSiteInit
     const disableFeedbackSurvey = window.context?.disableFeedbackSurvey
-    const isSiteInit = location.pathname === PageRoutes.SiteAdminInit
+    const isSiteInit = location.pathname === PageRoutes.SiteAdminInit.toString()
```
BolajiOlajide, line 99: Minor cleanup to ensure we are correctly comparing two variables of the same type.

client/web/src/user/settings/sidebaritems.ts
```diff
@@ -23,13 +26,19 @@ export const userSettingsSideBarItems: UserSettingsSidebarItems = [
         to: '/batch-changes',
         label: 'Batch Changes',
         condition: ({ batchChangesEnabled, user: { viewerCanAdminister }, authenticatedUser }) =>
-            batchChangesEnabled && viewerCanAdminister && canWriteBatchChanges(authenticatedUser),
+            !disableCodeSearchFeatures &&
+            batchChangesEnabled &&
```
eseliger, line 30: I wonder if we couldn't make `batchChangesEnabled` to be always false when it's a cody only license, and solely rely on that vs two booleans that we have to check here 🤔

client/web/src/user/area/navitems.ts
```diff
@@ -22,7 +25,7 @@ export const userAreaHeaderNavItems: readonly UserAreaHeaderNavItem[] = [
         to: '/searches',
         label: 'Saved searches',
         icon: FeatureSearchOutlineIcon,
-        condition: ({ user: { viewerCanAdminister } }) => viewerCanAdminister,
+        condition: ({ user: { viewerCanAdminister } }) => viewerCanAdminister && !disableCodeSearchFeatures,
```
eseliger, line 28: nit: would be cool if disableCodeSearchFeatures would also be part of the context vs some imported global.

BolajiOlajide: I initially added it as context, but then I realized I had to prop drill via several components to get it here. 
Considering how large context currently is, I hesitated against putting it there.

client/web/src/user/settings/sidebaritems.ts
BolajiOlajide: Oh! I see. I think this is related to the CodyEnabled comment you left on one of my PRs some weeks ago.

I do see the reasoning behind doing that, but I wonder if this won't be confusing for folks in the future.

eseliger: Hmm I think "checking if cody only" should only be a temporary thing.
Instead, we should be checking for available features only.

Is batch changes available? Cool, show it in the nav.
Is Cody also enabled? Show that as well! 

That way we will not have to change a lot of business logic when the enterprise plans change and their contents differ.
Also, this will make it easier to build stage 2 of this - different feature sets per user.

Does that make sense? 

I don't intend on blocking here since we're on a tight deadline, just wanted to raise that. LMK if you disagree or need to move faster and I'll stamp it :)

camdencheek (approved):

PR #59731, opened 2024-01-21
New setting to specify the maximum number of active access tokens a user is allowed

chwarwick:
Add a setting to limit the number of active access tokens a user can have at a one time, the default value is 25 tokens.  Internal access tokens do not count towards the active limit.  Override of this default is available using the `maxTokensPerUser` setting in the `auth.AccessTokens` section of the site config.

This change does not effect existing tokens, if a user is over the limit no tokens are invalidated however they will not be able to create new tokens without first removing enough to get below the limit. 

closes https://github.com/sourcegraph/sourcegraph/issues/59544
## Test plan
tests add
manual test of access token page and call back page
Screenshot 2024-01-21 at 3 04 46 PM
Screenshot 2024-01-21 at 3 05 09 PM

### DIFF
--- a/internal/conf/computed.go
+++ b/internal/conf/computed.go
@@ -71,6 +71,15 @@ func AccessTokensAllow() AccessTokenAllow {
 	}
 }
 
+func AccessTokensMaxPerUser() int {
+	defaultValue := 25
+	cfg := Get().AuthAccessTokens
+	if cfg == nil || cfg.MaxTokensPerUser == nil {
+		return defaultValue
+	}
+	return *cfg.MaxTokensPerUser
+}
+
 // AccessTokensAllowNoExpiration returns whether access tokens can be created without expiration.
 func AccessTokensAllowNoExpiration() bool {
 	cfg := Get().AuthAccessTokens
--- a/internal/database/access_tokens.go
+++ b/internal/database/access_tokens.go
@@ -214,7 +214,7 @@ WITH subject_user AS (
 		  subject_user_id = $1
 		  AND deleted_at IS NULL
 		  AND (expires_at IS NULL OR expires_at > NOW())
-		  AND internal IS NOT True
+		  AND internal IS NOT TRUE
 	) AS active_tokens
    FROM users WHERE id=$1 AND deleted_at IS NULL FOR UPDATE
 ),
--- a/internal/database/access_tokens.go
+++ b/internal/database/access_tokens.go
@@ -44,6 +44,9 @@ type AccessToken struct {
 // but it does not exist.
 var ErrAccessTokenNotFound = errors.New("access token not found")
 
+// ErrTooManyAccessTokens is returned when creating an access token would exceed the configured maximum number of active access tokens.
+var ErrTooManyAccessTokens = errors.New("number of active access tokens exceeds limit")
+
 // MaxAccessTokenLastUsedAtAge is the maximum amount of time we will wait before updating an access token's
 // last_used_at column. We are OK letting the value get a little stale in order to cut down on database writes.
 const MaxAccessTokenLastUsedAtAge = 5 * time.Minute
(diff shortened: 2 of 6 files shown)


internal/conf/computed.go
```diff
@@ -71,6 +71,15 @@ func AccessTokensAllow() AccessTokenAllow {
 	}
 }
 
+func AccessTokensMaxPerUser() int {
+	defaultValue := 25
+	cfg := Get().AuthAccessTokens
+	if cfg == nil || cfg.MaxTokensPerUser == nil {
```
eseliger, line 77: should we say that <= 0 is also a failure case and should instead use `defaultValue`? There's a different toggle that prevents access token creation altogether already, and setting it to zero means effectively the same, but with less clear error messages.

internal/database/access_tokens.go
```diff
@@ -202,19 +205,40 @@ func (s *accessTokenStore) createToken(ctx context.Context, subjectUserID int32,
...
+  SELECT
+    id,
+	(
+		SELECT COUNT(*)
+		FROM access_tokens
+		WHERE
+		  subject_user_id = $1
+		  AND deleted_at IS NULL
+		  AND (expires_at IS NULL OR expires_at > NOW())
+		  AND internal IS NOT True
```
eseliger, line 217: pythonic nit:

```suggestion
		  AND internal IS NOT true
```

or

```suggestion
		  AND internal IS NOT TRUE
```

internal/database/access_tokens.go
```diff
@@ -202,19 +205,40 @@ func (s *accessTokenStore) createToken(ctx context.Context, subjectUserID int32,
...
   FROM subject_user, creator_user
+  WHERE subject_user.active_tokens < $8::int OR $7::boolean
 )
 INSERT INTO access_tokens(subject_user_id, scopes, value_sha256, note, creator_user_id, expires_at, internal) SELECT * FROM insert_values RETURNING id
 `,
-		subjectUserID, pq.Array(scopes), hashutil.ToSHA256Bytes(b[:]), note, creatorUserID, dbutil.NullTimeColumn(expiresAt), internal,
+		subjectUserID, pq.Array(scopes), hashutil.ToSHA256Bytes(b[:]), note, creatorUserID, dbutil.NullTimeColumn(expiresAt), internal, conf.AccessTokensMaxPerUser(),
 	).Scan(&id); err != nil {
+		// if creation failed check to see if it was because too many tokens already
+		count, countErr := s.Count(ctx, AccessTokensListOptions{SubjectUserID: subjectUserID})
```
eseliger, line 234: I guess technically we should be doing those two DB calls in a transaction for maximum correctness, but not sure it's worth the extra work.

schema/site.schema.json
```diff
@@ -1168,6 +1168,16 @@
...
+        },
+        "maxTokensPerUser": {
+          "description": "The maximum number of active access tokens a user may have.",
+          "type": "integer",
+          "!go": {
+            "pointer": true
+          },
+          "default": 25,
+          "maximum": 100,
+          "minimum": 1
```
eseliger, line 1180: These two aren't actually enforced - should we add that?

eseliger (approved):
thanks for tests as well!

internal/database/access_tokens.go
chwarwick: Yeah I debated that but figured since it's already failed and the likely reason is that it's too many tokens it's just a matter of this supplying a nicer error.

schema/site.schema.json
chwarwick: When I run locally with `sg start` if I'm outside of the bounds frontend doesn't start.  Is that a dev only behavior?

```
 FATAL sourcegraph svcmain/svcmain.go:167 failed to start service {"service": "frontend", "error": "failed to apply site config overrides: writing site config overrides to database: ConfStore.SiteCreateIfUpToDate: site configuration is invalid: auth.accessTokens.maxTokensPerUser: Must be less than or equal to 100"}
```

eseliger: hmmmm interesting, I'm not sure 🤔 I know many of the properties that JSON schemas offer are not actually honored if we don't add custom validation logic for them - but maybe min/max actually are? Might be good to check if from the UI it also doesn't let you submit the invalid config

chwarwick: image
Looks like the UI checks this .

eseliger: great - thanks for checking!

PR #59732, opened 2024-01-21
Don't send users access token created email when creation fails

chwarwick:
The error from creation of access tokens was unchecked with caused emails to be sent when creation of the access token failed.  This becomes a more common situation after https://github.com/sourcegraph/sourcegraph/pull/59731 which adds a limit to the number of access tokens a user may have.
## Test plan
manually tested


arafatkatze (approved):

unknwon (approved):
😂

PR #59735, opened 2024-01-22
gomod: update zoekt

keegancsmith:
https://github.com/sourcegraph/zoekt/compare/7487a0d53131...b82a981f8240

There are quite a few changes, but most are internal e2e test changes. Here are the commits which change behaviour. Only the scoring change will have a minor effect on users in rare cases:

- https://github.com/sourcegraph/zoekt/commit/7e44ea78f6 all: adjust field order for match structs
- https://github.com/sourcegraph/zoekt/commit/7dcc7979f6 api: implement succinct output for SearchOptions.String
- https://github.com/sourcegraph/zoekt/commit/1adab43b4c score: remove ZOEKT_NOVELTY_DISABLE
- https://github.com/sourcegraph/zoekt/commit/18b6999367 api: JSON omitempty on optional fields in FileMatch
- https://github.com/sourcegraph/zoekt/commit/b82a981f82 score: do scoring on candidateMatches

And below are the e2e test changes:

- https://github.com/sourcegraph/zoekt/commit/155050eb98 zoekt-archive-index: split out ranking tests and archive indexing
- https://github.com/sourcegraph/zoekt/commit/344e33ae3c e2e: introduce optional shard cache
- https://github.com/sourcegraph/zoekt/commit/6463096a03 e2e: recall and visual indication of ranks
- https://github.com/sourcegraph/zoekt/commit/b35d8a202d e2e: add ranking tests for atom boosting
- https://github.com/sourcegraph/zoekt/commit/4be2929fd0 e2e: add a ranking test case for "zoekt searcher"

Test Plan: tested in zoekt repo and this CI

### DIFF
--- a/deps.bzl
+++ b/deps.bzl
@@ -5741,8 +5741,8 @@ def go_dependencies():
         name = "com_github_sourcegraph_zoekt",
         build_file_proto_mode = "disable_global",
         importpath = "github.com/sourcegraph/zoekt",
-        sum = "h1:TB4KxGtcJm98TfyDHhSkbffngM+EVlTu0epG3SxUU0U=",
-        version = "v0.0.0-20240110094557-7487a0d53131",
+        sum = "h1:CumaleyciZzwsNOf1ewSC2hgKdFaBuO0yzTeIUcp64Y=",
+        version = "v0.0.0-20240122093209-b82a981f8240",
     )
     go_repository(
         name = "com_github_spaolacci_murmur3",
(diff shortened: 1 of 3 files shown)


jtibshirani (approved):
Looks good to me. Could you clarify which scoring change will have a minor impact on users? Is it "score: remove ZOEKT_NOVELTY_DISABLE", since this variable defaulted to false?

keegancsmith:
> Could you clarify which scoring change will have a minor impact on users? 

It was to do with how we score. The minor difference is detailed here https://github.com/sourcegraph/zoekt/pull/723#discussion_r1461570249

> Is it "score: remove ZOEKT_NOVELTY_DISABLE", since this variable defaulted to false?

That should be no change since this did default to false, but it is a "DISABLE", so in fact we have been doing novelty boosting for a while.

Issue #59733, opened 2024-01-22
Hovers not getting triggered for #include statements


varungandhi-src:
https://sourcegraph.com/github.com/sourcegraph/scip-clang@1995a07d7a14baa6a2f78541b6d37a12d83d9eca/-/blob/indexer/Indexer.cc?L32-42

![](https://github.com/sourcegraph/sourcegraph/assets/93103176/61d0365f-0f10-4a76-ba20-1b7e0135b79b)

Reproduces in Firefox and Chrome

varungandhi-src:
Previously, the way file navigation worked in scip-clang is that:
1. It would emit a 'definition' occurrence for the file at the first token. Typically, this would be a keyword or a comment.
2. It would emit a 'reference' occurrence for the range for the include (the string `"a.h"` in `#include "a.h"`).

As such, this was written to take advantage of the old implementation of code navigation without making client-side changes.

The reason hovers stopped working is because we use syntax highlighting data as evidence for click-ability. We only check for hover information for certain kinds of syntactic elements such as identifiers, not for keywords, comments or strings. Consequently:
- When you hover over the first token in a file, you no longer get a 'You are at the definition' for the file itself.
- When you hover on an include path, that's marked as a string in the grammar, and hence is considered ineligible for code nav.

There are a few different fixes possible here:
1. Provide a different blessed way for supporting code navigation between files -- update client-side code and indexers based on that.
2. Make the precise SCIP data available client-side, and prefer that for determining hover-ability instead of syntax highlighting data.
3. Add some other workaround specific to C/C++ to match the current behavior of scip-clang. (E.g. I don't exactly know the old behavior, but maybe we can copy that for C/C++.)

The attempted fix in https://github.com/sourcegraph/sourcegraph/pull/59736 enabled hovers for all strings and keywords, but since it wasn't limited on a per-language basis, it would lead to hovers being triggered for actual strings/keywords too.

varungandhi-src:
We can potentially have a partial fix for this by perhaps modifying the highlighting query to add something special for imports, so that you can do jump to definition.

PR #59736, opened 2024-01-22
client: Re-enable file nav related hovers for C and C++

varungandhi-src:
Fixes https://github.com/sourcegraph/sourcegraph/issues/59733

## Test plan

Manually checked the hovers start showing up at the start of the file,
and when hovering over the header path in the `#include`.

### DIFF
--- a/client/shared/src/codeintel/scip.ts
+++ b/client/shared/src/codeintel/scip.ts
@@ -244,7 +244,7 @@ export enum SyntaxKind {
     PunctuationBracket = 3,
 
     // `if`, `else`, `return`, `class`, etc.
-    IdentifierKeyword = 4,
+    Keyword = 4,
 
     // `+`, `*`, etc.
     IdentifierOperator = 5,
--- a/client/web/src/repo/blob/codemirror/occurrence-utils.ts
+++ b/client/web/src/repo/blob/codemirror/occurrence-utils.ts
@@ -32,7 +32,6 @@ const INTERACTIVE_OCCURRENCE_KINDS = new Set([
     // a comment or a keyword.
     SyntaxKind.Keyword,
     SyntaxKind.Comment,
-
 ])
 
 export const isInteractiveOccurrence = (occurrence: Occurrence): boolean => {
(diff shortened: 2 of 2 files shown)


client/shared/src/codeintel/scip.ts
```diff
@@ -244,7 +244,7 @@ export enum SyntaxKind {
     PunctuationBracket = 3,
 
     // `if`, `else`, `return`, `class`, etc.
-    IdentifierKeyword = 4,
+    Keyword = 4,
```
varungandhi-src, line 247: I renamed this because `IdentifierKeyword` was deprecated in SCIP in favor of just `Keyword`. https://github.com/sourcegraph/scip/blob/main/scip.proto#L537-L539

client/web/src/repo/blob/codemirror/occurrence-utils.ts
```diff
@@ -23,6 +24,14 @@ const INTERACTIVE_OCCURRENCE_KINDS = new Set([
...
     SyntaxKind.IdentifierType,
     SyntaxKind.IdentifierAttribute,
+    // In #include "foo.h", "foo.h" gets marked as a string literal
+    SyntaxKind.StringLiteral,
+    // For C and C++, scip-clang emits an occurrence for the very
+    // first token in the source file to support 'Go to def' and
+    // 'Find refs' for header files. This first token is usually
+    // a comment or a keyword.
+    SyntaxKind.Keyword,
+    SyntaxKind.Comment,
```
fkling, line 34: Unfortunately it doesn't seem to be that simple because that will cause hovercards to trigger in other places, for examples:

varungandhi-src: Before we added this fast path, we didn't trigger the hover for keywords and strings, but we did support code navigation for includes. So how did we handle it then?

Ideally, we would have access to the precise SCIP document for the file, and use that to check if a range has a precise occurrence.

varungandhi-src:
I'm closing this PR for now since this isn't the right fix.

PR #59739, opened 2024-01-22
Add Exclude field to gerrit code host connections

pjlast:
Adds an `exclude` field to Gerrit code host connections that acts as the inverse of the `projects` field. Any repos in this list will never ever be added by this code host connection.

Docs update: https://github.com/sourcegraph/docs/pull/53

## Test plan

Added unit tests


schema/gerrit.schema.json
```diff
@@ -38,6 +38,15 @@
...
       ]
     },
+    "exclude": {
+      "description": "An array of project strings specifying which Gerrit projects to exclude from mirroring on Sourcegraph. If empty, no projects will be excluded. Useful when using an empty 'projects' field. Exclusion takes precedence over inclusion.",
+      "type": "array",
+      "items": { "type": "string" },
+      "examples": [
+        ["name", "owner/name"],
+        ["docs", "kubernetes/kubernetes"]
+      ]
```
eseliger, line 48: Both GitHub and GitLab configs use objects here, to also allow excluding repos by other properties, like size, pattern, empty repos, external id, archived status etc. 
We won't need that now, but will we ever want that, we might regret that we didn't make these objects now, do you think that'd be a huge hassle to add?

So proposed schema would instead be

```
"exclude": [{"name": "src-cli"}]
```

pjlast: Yeah I think this is a good idea even if only for consistency

eseliger (approved):

PR #59742, opened 2024-01-22
WIP: Consume Cody Subscription Info from SSC in dotcom

thenamankumar:
In-Progress

Make a request to SSC from dotcom with the SAMS account_id to fetch user's subscription info. Combine the SSC subscription details with the Cody Pro trial opted info from dotcom based on `users.cody_pro_enabled_at` to expose a new `user.codySubscription` graphql API which can be consumed by Cody Clients and Cody Management dashboard. 

This PR also makes dotcom's rate limit resolver consumed by the SSC API to fetch user's subscription.

## Test plan
TODO

### DIFF
--- a/client/web/src/cody/subscription/CodySubscriptionPage.tsx
+++ b/client/web/src/cody/subscription/CodySubscriptionPage.tsx
@@ -22,6 +22,7 @@ import {
 import type { AuthenticatedUser } from '../../auth'
 import { Page } from '../../components/Page'
 import { PageTitle } from '../../components/PageTitle'
+import { useFeatureFlag } from '../../featureFlags/useFeatureFlag'
 import type { UserCodyPlanResult, UserCodyPlanVariables } from '../../graphql-operations'
 import { eventLogger } from '../../tracking/eventLogger'
 import { EventName } from '../../util/constants'
@@ -47,6 +48,8 @@ export const CodySubscriptionPage: React.FunctionComponent<CodySubscriptionPageP
 
     const utm_source = parameters.get('utm_source')
 
+    const [sscEnabled] = useFeatureFlag('ssc-enabled', false)
+
     useEffect(() => {
         eventLogger.log(EventName.CODY_SUBSCRIPTION_PAGE_VIEWED, { utm_source }, { utm_source })
     }, [utm_source])
@@ -68,7 +71,7 @@ export const CodySubscriptionPage: React.FunctionComponent<CodySubscriptionPageP
         return null
     }
 
-    const { codyProEnabled } = data.currentUser
+    const isPro = data.currentUser?.codySubscription?.isPro || false
 
     return (
         <>
@@ -191,7 +194,7 @@ export const CodySubscriptionPage: React.FunctionComponent<CodySubscriptionPageP
                                 <Text className="mb-3 text-muted" size="small">
                                     Free until Feb 2024, <strong>no credit card needed</strong>
                                 </Text>
-                                {codyProEnabled ? (
+                                {isPro ? (
                                     <div>
                                         <Text
                                             className="mb-0 text-muted d-inline cursor-pointer"
@@ -206,10 +209,17 @@ export const CodySubscriptionPage: React.FunctionComponent<CodySubscriptionPageP
                                                         tier: 'free',
                                                     }
                                                 )
+                                                if (sscEnabled) {
+                                                    window.open(
+                                                        'https://accounts.sourcegraph.com/cody/subscription?pro=true',
+                                                        '_blank'
+                                                    )
+                                                    return
+                                                }
                                                 setShowCancelPro(true)
                                             }}
                                         >
-                                            Cancel
+                                            {sscEnabled ? 'Manage' : 'Cancel'} subscription
                                         </Text>
                                     </div>
                                 ) : (
@@ -222,11 +232,19 @@ export const CodySubscriptionPage: React.FunctionComponent<CodySubscriptionPageP
                                                 { tier: 'pro' },
                                                 { tier: 'pro' }
                                             )
+                                            if (sscEnabled) {
+                                                window.open(
+                                                    'https://accounts.sourcegraph.com/cody/subscription',
+                                                    '_blank'
+                                                )
+                                                return
+                                            }
+
                                             setShowUpgradeToPro(true)
                                         }}
                                     >
                                         <Icon svgPath={mdiTrendingUp} className="mr-1" aria-hidden={true} />
-                                        Get Pro trial
+                                        {sscEnabled ? 'Get Pro' : 'Get Pro trial'}
                                     </Button>
                                 )}
                             </div>
--- a/cmd/frontend/graphqlbackend/schema.graphql
+++ b/cmd/frontend/graphqlbackend/schema.graphql
@@ -6361,6 +6361,24 @@ type NewUsersConnection implements Connection {
     pageInfo: ConnectionPageInfo!
 }
 
+enum CodySubscriptionStatus {
+    active
+    past_due
+    unpaid
+    canceled
+    trialing
+    pending
+    other
+}
+
+type CodySubscription {
+    status: CodySubscriptionStatus!
+    isPro: Boolean!
+    applyProRateLimits: Boolean!
+    currentPeriodStartAt: DateTime!
+    currentPeriodEndAt: DateTime!
+}
+
 """
 A user.
 """
@@ -6573,18 +6591,26 @@ type User implements Node & SettingsSubject & Namespace {
     """
     codeCompletionsQuotaOverride: Int
     """
+    The Cody subscription details for the user.
+    """
+    codySubscription: CodySubscription
+    """
+    DEPRICATED
     Whether the user has enrolled for Cody Pro.
     """
     codyProEnabled: Boolean!
     """
+    DEPRICATED
     The date when the user enrolled for Cody Pro.
     """
     codyProEnabledAt: DateTime
     """
+    DEPRICATED
     The start date of current billing period of Cody.
     """
     codyCurrentPeriodStartDate: DateTime
     """
+    DEPRICATED
     The end date of current billing period of Cody.
     """
     codyCurrentPeriodEndDate: DateTime
--- a/cmd/frontend/graphqlbackend/user.go
+++ b/cmd/frontend/graphqlbackend/user.go
@@ -12,6 +12,7 @@ import (
 	"github.com/graph-gophers/graphql-go/relay"
 
 	"github.com/sourcegraph/sourcegraph/internal/accesstoken"
+	"github.com/sourcegraph/sourcegraph/internal/ssc"
 
 	"github.com/keegancsmith/sqlf"
 	"github.com/sourcegraph/log"
@@ -79,7 +80,8 @@ type UserResolver struct {
 	// actor containing current user (derived from request context), which lets us
 	// skip fetching actor from context in every resolver function and save DB calls,
 	// because user is fetched in actor only once.
-	actor *actor.Actor
+	actor                 *actor.Actor
+	codySubscriptionCache *ssc.UserCodySubscription
 }
 
 // NewUserResolver returns a new UserResolver with given user object.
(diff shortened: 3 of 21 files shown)


client/web/src/cody/subscription/CodySubscriptionPage.tsx
```diff
@@ -47,6 +48,8 @@ export const CodySubscriptionPage: React.FunctionComponent<CodySubscriptionPageP
 
     const utm_source = parameters.get('utm_source')
 
+    const [sscEnabled] = useFeatureFlag('ssc-enabled', false)
```
chrsmith, line 51: `sec-enabled` isn't especially clear. Can we change this to be something more descriptive about what it actually enables? Perhaps:

`overlay-subscription-data-from-scc` or `use-scc-for-subscription-data`?

client/web/src/cody/subscription/CodySubscriptionPage.tsx
```diff
@@ -206,10 +209,17 @@ export const CodySubscriptionPage: React.FunctionComponent<CodySubscriptionPageP
                                                         tier: 'free',
                                                     }
                                                 )
+                                                if (sscEnabled) {
+                                                    window.open(
+                                                        'https://accounts.sourcegraph.com/cody/subscription?pro=true',
```
chrsmith, line 214: Nit: Could we move all of the URLs that are fixed to anything on accounts.sourcegraph.com to another file and exposed as a constant?

e.g. this would be `SelfServiceCodyManageSubscriptionAlreadySignedUpOnDotcom`. (Yes, it's annoying long.) That way when we later migrate SSC to `cody.sourcegraph.com` we only need to make the changes in one-place. And also the places where we do direct the user to new locations are clearer about what the purpose of that page is.

client/web/src/cody/subscription/CodySubscriptionPage.tsx
```diff
@@ -206,10 +209,17 @@ export const CodySubscriptionPage: React.FunctionComponent<CodySubscriptionPageP
                                                         tier: 'free',
                                                     }
                                                 )
+                                                if (sscEnabled) {
+                                                    window.open(
```
chrsmith, line 213: Does `window.open` navigate the user? Or does it open a new tab? In order to have a more seamless experience, we should just navigate the user's browser rather than opening a new tab/window. Right?

cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -6361,6 +6361,24 @@ type NewUsersConnection implements Connection {
     pageInfo: ConnectionPageInfo!
 }
 
+enum CodySubscriptionStatus {
+    active
+    past_due
+    unpaid
+    canceled
+    trialing
+    pending
```
chrsmith, line 6370: Let's remove `pending` from these states. We have no plans for ever supporting this, and would just hopefully just pass this over as `other`. By having `pending` as a valid enum state, linters may require us to handle it in `switch` statements and it's just going to be confusing down the road.

cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -6361,6 +6361,24 @@ type NewUsersConnection implements Connection {
...
+    other
+}
+
+type CodySubscription {
```
chrsmith, line 6374: Could you add some comments here to describe what these various fields mean and how to interpret them? Otherwise it will be harder to use this type correctly.

For example:

- `currentPeriodEndAt` may be _in the past_ if the subscription is in the `canceled` state.
- What is the purpose of `isPro`? Will there ever be a situation where there is a subscription but `isPro` is false? (When `status === 'canceled'`.) Are there any others?
- When would `applyProRateLimits` be true? Is it always going to match `isPro`?

cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -6573,18 +6591,26 @@ type User implements Node & SettingsSubject & Namespace {
     """
     codeCompletionsQuotaOverride: Int
     """
+    The Cody subscription details for the user.
```
chrsmith, line 6594: Could you expand on this doc comment and clarify when it will be `null/undefined`? And if so, what the behavior should be?

AFAIK it should be something like:

```suggestion
    The Cody subscription details for the user. If null, the user has not signed up
    for Cody Pro, and will only receive "free" tier limits. However, it's possible
    there is a subscription but the user should still only receive free tier limits. 
    In the case that `subscription.isPro` is false or `subscription.status` is 
    canceled'.
```

cmd/frontend/graphqlbackend/user.go
```diff
@@ -176,22 +178,70 @@ func (r *UserResolver) URL() string {
...
+	return gqlutil.DateTime{Time: r.subscription.CurrentPeriodStartAt}
+}
+
+func (r *CodySubscriptionResolver) CurrentPeriodEndAt() gqlutil.DateTime {
+	return gqlutil.DateTime{Time: r.subscription.CurrentPeriodEndAt}
+}
+
+func (r *UserResolver) getCodySubscription(ctx context.Context) (*ssc.UserCodySubscription, error) {
+	if r.codySubscriptionCache == nil {
+		subscription, err := ssc.CodySubscriptionForUser(ctx, *r.user)
```
chrsmith, line 207: Could this result in us making multiple calls to `ssc.CodySubscriptionForUser`? Or is the GraphQL scheme/implementation written in such a way that this can only ever be called once?

For reference, on the SSC side we use `sync.Once` to lazily load and memoize an API call. (Which TBH is kinda tricky to get right. But probably preferable to making extra API calls if they aren't required.) See:
https://github.com/sourcegraph/accounts.sourcegraph.com/blob/main/cody/backend/graph/model/user.go#L28

internal/ssc/ssc.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/sourcegraph/sourcegraph/internal/httpcli"
+	"github.com/sourcegraph/sourcegraph/internal/types"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+)
+
+const SSC_API_BASE_URL = "http://127.0.0.1:9982/cody"
+const SSC_ADMINISTRATIVE_SECRET_TOKEN = "naman"
+
+// FetchSubscriptionBySAMSAccountID returns the user's Cody subscription for the sams_account_id
+func FetchSubscriptionBySAMSAccountID(samsAccountID string) (*SSCSubscription, error) {
```
chrsmith, line 21: I'd suggest defining a proper "client interface" for SSC. This would make it amenable to mocking and easier unit testing. (As well as putting all of the key functions in one place to easily read and update.)

```go
type Client interface {
   FetchSubscriptionByAccountID(accountID) (*Subscription, error)
...
```

internal/ssc/ssc.go
```diff
@@ -0,0 +1,134 @@
...
+	"net/http"
+	"time"
+
+	"github.com/sourcegraph/sourcegraph/internal/featureflag"
+	"github.com/sourcegraph/sourcegraph/internal/httpcli"
+	"github.com/sourcegraph/sourcegraph/internal/types"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+)
+
+const SSC_API_BASE_URL = "http://127.0.0.1:9982/cody"
```
chrsmith, line 17: We should avoid using `SSC_*`, `SSC*` since the package name is `ssc` it "stutters". (Probably is a lint warning already.)

internal/ssc/types.go
```diff
@@ -0,0 +1,49 @@
...
+	SubscriptionStatusPastDue  SubscriptionStatus = "past_due"
+	SubscriptionStatusUnpaid   SubscriptionStatus = "unpaid"
+	SubscriptionStatusCanceled SubscriptionStatus = "canceled"
+	SubscriptionStatusTrialing SubscriptionStatus = "trialing"
+	SubscriptionStatusPending  SubscriptionStatus = "pending"
+	SubscriptionStatusOther    SubscriptionStatus = "other"
+)
+
+type SSCSubscription struct {
+	SAMSAccountID string
```
chrsmith, line 28: Having a field here that isn't part of the REST API seems kinda dangerous.

Would you like us to include this in the SSC API? (It doesn't quite make sense, since a subscription is actually tied to a _team_, which will have 1-or-more users.)

If you need to bundle the two, I would suggest embedding the struct to make it clear which one is part of the REST API and which one is just for simplifying our implementation. e.g.

```go
// SubscriptionResponse is the Subscription object returned by the SSC backend.
type SubscriptionResponse struct {
    Status SubscriptionStatus `json:"status"`
    ...
}

// userSubscription bundles a subscription with a user ID for convenience.
type userSubscription struct {
    Subscription 
    SAMSAccountID string     
}
```

chrsmith (commented):
Left some notes and nitpicks.

Overall this seems reasonable, though I think we can clean up the `sec` package a bit. (Which I assume was the plan since this PR is only a draft after all.)

My main feedback is just making sure we add an obnoxious amount of comments to clarify what's going on. e.g. which parts of a user's "cody pro" info come from the dotcom backend, and which parts come from "the SSC backend". As without that, anybody new to the code will have trouble making sense of it. (e.g. how we overlap the two sets of data somewhere in the `graphql` API.)

Also, there is some redundancy with fields like `isProUser` and `applyProRateLimits`. And it isn't clear to me when/why they would differ. I'm definitely a fan of having more fields to avoid clients needing to reimplement the same logic and checks. But we need to do a really good job of documenting these things so that they can be used correctly.

cmd/frontend/graphqlbackend/schema.graphql
kalanchan: @chrsmith I think Naman added `pending` to help with the transition period from free -> pro. 

IF I UNDERSTAND THE [THREAD CORRECTLY](https://sourcegraph.slack.com/archives/C059N5FRYG3/p1705936522245649?thread_ts=1705432802.097489&cid=C059N5FRYG3), he's suggesting that clients can use `status === pending` as the state where a user has added a credit card, and a payment is pending -> which doesn't require a notification that their pro trial is ending

but I think the above can also be done if we just check `status === trialing` and we can check the `currentPeriodEndAt` for the end date

chrsmith: > he's suggesting that clients can use status === pending as the state

TL;DR: We _definitely_ should remove `pending`, and just use `status == trialing` to check if a user has onboarded into Cody Pro before 2/15.

🛑 I must have missed that but, DO NOT DO THIS. We need the subscription status value to have extremely precise meaning, and unless we make it crystal clear that this is a "Dotcom's interpretation" of the subscription status, we don't want to conflate this status with what is being returned by SSC or Stripe.

If the user has entered their credit card on the SSC site, we will create a Stripe subscription and it will be in the `trialing` state until 2/15. (After which time it will be `active` if they convert, `canceled` if they end up not wanting to renew, or `past_due`/`unpaid` if we are trying to collect payment... and after X days it will go to either `active` or `canceled`.)

Anyways, we will never expose a state where "payment is pending". (Since that makes it sound like we have scheduled an invoice, but it hasn't been paid yet, and/or waiting for a check to clear... and that's definitely not the level of granularity about the subscription we want to be exposing from the dotcom GraphQL API.)

client/web/src/cody/subscription/CodySubscriptionPage.tsx
kalanchan: @chrsmith  `ssc-enabled`**** 😛

vdavid: Once the feature flag name is more final, please also update it in the [launch plan](https://docs.google.com/document/d/1bQkgy4g6wZyZOCgh5lGyjigiwIHNZyNH-Tc2vhXt-pU/edit).

cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -6573,18 +6591,26 @@ type User implements Node & SettingsSubject & Namespace {
...
+    """
+    codySubscription: CodySubscription
+    """
+    DEPRICATED
```
vdavid, line 6598: Typo:
```suggestion
    DEPRECATED
```

Same below 3x

vdavid (commented):
@thenamankumar, I read through the code and Chris's comments, but it got pretty late and I don't have time left for taking a deeper look. Will get back to it in my morning. I'll ping you before it to make sure I'm not reviewing a stale version.

cmd/frontend/graphqlbackend/schema.graphql
thenamankumar: @chrsmith `pending` is not for payment pending at SSC. It is only for dotcom to know that there is no subscription on SSC for the user or they have not even connected to SAMS yet.

We need this state in dotcom for the UI & clients.

thenamankumar:
Closed in favour of creating 3 separate smaller PRs and addressing the review there.

mkondratek:
@thenamankumar can you please link the new PRs here when you have them? 🙏

thenamankumar:
@mkondratek The first one is here: https://sourcegraph.slack.com/archives/C05SZB829D0/p1706030325615199

PR #59744, opened 2024-01-22
search jobs: graduate from EAP to beta

stefanhengl:
Relates to #59352 

This updates the badges from "experimental" to "beta" and enables search jobs by default.

## Test plan
Just a change of defaults. I checked the badges in the Navbar and on the Search Jobs page manually

image

image


jtibshirani (approved):
Looks good to me!

PR #59747, opened 2024-01-22
WIP - worker bootstrap and launch configuration

keynmol:

### DIFF
--- a/dev/scip-treesitter-dev
+++ b/dev/scip-treesitter-dev
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+
+# This script is a wrapper around `scip-ctags`.
+#
+# It checks if scip-ctags has been installed through ./dev/scip-ctags-install.sh or prints an error and explain
+# how to install it.
+#
+# To use your own `scip-ctags` binary instead of this wrapper in your local dev server, use
+# `SCIP_CTAGS_COMMAND=path/to/ctags sg start`.
+
+root="$(dirname "${BASH_SOURCE[0]}")/.." >/dev/null
+TARGET=$("$root/dev/scip-treesitter-install.sh" which)
+
+if [ ! -f "${TARGET}" ]; then
+    echo "scip-treesitter is not installed, please run ./dev/scip-treesitter-install.sh"
+    echo "Alternatively you can use SCIP_TREESITTER_COMMAND=path/to/scip-treesitter to use your own binary."
+    exit 1
+else
+  ${TARGET} "$@"
+fi
--- a/dev/scip-treesitter-install.sh
+++ b/dev/scip-treesitter-install.sh
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+
+set -euf -o pipefail
+pushd "$(dirname "${BASH_SOURCE[0]}")/.." >/dev/null
+mkdir -p .bin
+
+# TODO: add similar task to zoekt alpine
+
+NAME="scip-treesitter"
+TARGET="$PWD/.bin/${NAME}"
+
+if [ $# -ne 0 ]; then
+  if [ "$1" == "which" ]; then
+    echo "$TARGET"
+    exit 0
+  fi
+fi
+
+function ctrl_c() {
+  printf "[-] Installation cancelled.\n"
+  exit 1
+}
+
+trap ctrl_c INT
+
+function build_scip_treesitter {
+  cd docker-images/syntax-highlighter/crates/scip-treesitter-cli
+  cargo build --bin scip-treesitter-cli --target-dir target
+  cp ./target/release/scip-treesitter-cli "$TARGET"
+}
+
+build_scip_treesitter
+
+popd >/dev/null
--- a/sg.config.yaml
+++ b/sg.config.yaml
@@ -59,6 +59,7 @@ env:
       { "Name": "embeddings", "Host": "127.0.0.1:6099" },
       { "Name": "zoekt-index-0", "Host": "127.0.0.1:6072" },
       { "Name": "zoekt-index-1", "Host": "127.0.0.1:6073" },
+      { "Name": "outline-scip-codeintel-worker", "Host": "127.0.0.1:6075" },
       { "Name": "zoekt-web-0", "Host": "127.0.0.1:3070", "DefaultPath": "/debug/requests/" },
       { "Name": "zoekt-web-1", "Host": "127.0.0.1:3071", "DefaultPath": "/debug/requests/" }
     ]
@@ -182,7 +183,7 @@ commands:
       fi
       go build -gcflags="$GCFLAGS" -o .bin/gitserver github.com/sourcegraph/sourcegraph/cmd/gitserver
     checkBinary: .bin/gitserver
-    env: &gitserverenv
+    env:
       HOSTNAME: 127.0.0.1:3178
     watch:
       - lib
@@ -197,7 +198,6 @@ commands:
   gitserver-0:
     <<: *gitserver_template
     env:
-      <<: *gitserverenv
       GITSERVER_EXTERNAL_ADDR: 127.0.0.1:3501
       GITSERVER_ADDR: 127.0.0.1:3501
       SRC_REPOS_DIR: $HOME/.sourcegraph/repos_1
@@ -206,7 +206,6 @@ commands:
   gitserver-1:
     <<: *gitserver_template
     env:
-      <<: *gitserverenv
       GITSERVER_EXTERNAL_ADDR: 127.0.0.1:3502
       GITSERVER_ADDR: 127.0.0.1:3502
       SRC_REPOS_DIR: $HOME/.sourcegraph/repos_2
@@ -280,6 +279,7 @@ commands:
         -e QDRANT_INIT_FILE_PATH=/data/.qdrant-initialized \
         --entrypoint /usr/local/bin/qdrant \
         sourcegraph/qdrant:insiders
+
   worker:
     cmd: |
       export SOURCEGRAPH_LICENSE_GENERATION_KEY=$(cat ../dev-private/enterprise/dev/test-license-generation-key.pem)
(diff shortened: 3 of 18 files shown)


dev/scip-treesitter-install.sh
```diff
@@ -0,0 +1,34 @@
...
+  printf "[-] Installation cancelled.\n"
+  exit 1
+}
+
+trap ctrl_c INT
+
+function build_scip_treesitter {
+  cd docker-images/syntax-highlighter/crates/scip-treesitter-cli
+  cargo build --bin scip-treesitter-cli --target-dir target
+  cp ./target/release/scip-treesitter-cli "$TARGET"
```
varungandhi-src, line 29: These need updating to remove the `-cli` suffix, right? It would be useful to have the Bazel names and Cargo names be the same.

sg.config.yaml
```diff
@@ -182,7 +183,7 @@ commands:
       fi
       go build -gcflags="$GCFLAGS" -o .bin/gitserver github.com/sourcegraph/sourcegraph/cmd/gitserver
     checkBinary: .bin/gitserver
-    env: &gitserverenv
+    env:
```
varungandhi-src, line 186: Random other change that snuck in...? If it's not necessary, could you remove these?

sg.config.yaml
```diff
@@ -1195,6 +1221,20 @@ commandsets:
       - grafana
       - prometheus
 
+  codeintel-syntactic:
```
varungandhi-src, line 1224: Thoughts on adding `codeintel-syntactic-worker` to other existing command sets like `enterprise` and `enterprise-codeintel` instead of adding a new command set, so that we're working with a close-to-prod system to a greater extent?

Also, maybe we should follow the gitserver naming convention and use `syntactic-code-intel-worker-0` and `syntactic-code-intel-worker-1` for the commands (instead of the variation `codeintel-syntactic-worker` so that grepping also shows hits in `sg.config.yaml`

sg.config.yaml
```diff
@@ -59,6 +59,7 @@ env:
       { "Name": "embeddings", "Host": "127.0.0.1:6099" },
       { "Name": "zoekt-index-0", "Host": "127.0.0.1:6072" },
       { "Name": "zoekt-index-1", "Host": "127.0.0.1:6073" },
+      { "Name": "syntactic-codeintel-worker", "Host": "127.0.0.1:6075" },
```
varungandhi-src, line 62: We will want this to be horizontally scalable (I think we should have this be part of the Q2 design goals).

So maybe we should have the default configuration with two replicas? That would help catch issues where we accidentally design/implement stuff without taking horizontal scaling into account.

```suggestion
      { "Name": "syntactic-code-intel-worker-0", "Host": "127.0.0.1:6075" },
      { "Name": "syntactic-code-intel-worker-1", "Host": "127.0.0.1:6076" },
```

cmd/syntactic-code-intel-worker/shared/shared.go
```diff
@@ -0,0 +1,45 @@
...
+const addr = ":3188"
+
+func Main(ctx context.Context, observationCtx *observation.Context, ready service.ReadyFunc, config Config) error {
+	logger := observationCtx.Logger
+
+	// Initialize tracing/metrics
+	// observationCtx = observation.NewContext(logger, observation.Honeycomb(&honey.Dataset{
+	// 	Name: "syntactic-code-intel-worker",
+	// }))
+
```
varungandhi-src, line 27:
```suggestion
```

Since it is passed as an argument.

cmd/syntactic-code-intel-worker/README.md
```diff
@@ -0,0 +1,3 @@
+# Precise code intel worker
+
+The precise-code-intel-worker service converts LSIF upload file into Postgres data. This service is horizontally scalable.
```
varungandhi-src, line 3:
```suggestion
# Syntactic code intel worker

🚧 WORK IN PROGRESS 🚧

Stateless service that handles generating SCIP data for codebases
using Tree-sitter for powering syntax-based code navigation.

Repository contents are cached for performance.

Horizontally scalable.

[Design docs](https://docs.google.com/document/d/14MHauv52o4zTFiV6gC6NOJZxcJpglK-ElWa64gqeKDo/edit) (Sourcegraph internal)
```

sg.config.yaml
keynmol: This is unfortunately necessary - the linter for yaml files has a bug with duplicate anchors, and as I have modified the config, I had to work around the bug.

sg.config.yaml
keynmol: I agree, but IMO for MVP we should keep things simple. Especially as this will be part of Q2 - there we can focus on making this scalable, at the moment it'd be more of a distraction IMO, as we need to verify the core parts of this worker first.

docker-images/syntax-highlighter/crates/scip-treesitter-cli/BUILD.bazel
```diff
@@ -2,8 +2,8 @@ load("@crate_index//:defs.bzl", "aliases", "all_crate_deps")
 load("@rules_rust//rust:defs.bzl", "rust_binary", "rust_library", "rust_test")
 
 rust_binary(
-    name = "scip-treesitter-cli",
-    srcs = glob(["src/main.rs"]),
+    name = "scip-treesitter",
+    srcs = ["src/main.rs"],
```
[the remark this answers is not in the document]
kritzcreek: I think some consolidation is in order, but I don't know if the `scip-treesitter-cli` crate is the worst offender.
From what I can tell:

- `scip-macros` contains two random macros
- `scip-treesitter` contains a test utility and a shared type for PackedRanges
- `scip-syntax` contains most of the logic for producing syntax driven scip data
- `scip-treesitter-languages` contains code for loading tree-sitter grammars, but also queries for syntax-highlights
- `sg-syntax` contains more highlighting logic and integrates with syntect
- `scip-treesitter-cli` contains the CLI _and_ the evaluation code for scip files

I don't think it's a problem that the binary has a different name than the crate it's defined in. -cli crates are pretty common. I would like to merge the macro stuff witht the test utilities and maybe move the evaluation code into one of the library crates.
I'm fine with having `scip-treesitter-languages` as the single point for adding grammars, but I think we should move the highlighting specific code somewhere else (maybe sg-syntax?)

dev/scip-treesitter-install.sh
keynmol: Don't think we can do this without consolidation of crates - the TARGET is still named `scip-treesitter` (as it should), but crate and the bin target are named `-cli` to avoid collision with the `scip-treesitter` crate.

It's a bit of a mess and would be good to clean up in a separate PR. The binary installed in the container and visible to the worker is still named `scip-treesitter` so it's fine.

dev/scip-treesitter-dev
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+# This script is a wrapper around `scip-ctags`.
+#
+# It checks if scip-ctags has been installed through ./dev/scip-ctags-install.sh or prints an error and explain
+# how to install it.
+#
+# To use your own `scip-ctags` binary instead of this wrapper in your local dev server, use
+# `SCIP_CTAGS_COMMAND=path/to/ctags sg start`.
```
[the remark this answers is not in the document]
keynmol: 👍 

Additionally in a separate PR I think I will modify the script to hash the binary and compare to some saved value.

This way we can add the crate's `target` folder to the list of watched folders and rebuild binary on-demand, restarting the whole thing automatically.

Should be a better dev ex in case we need to make concurrent modifications to Rust and Go code.

dev/scip-treesitter-install.sh
varungandhi-src: > but crate and the bin target are named -cli to avoid collision with the scip-treesitter crate.

The crate name needs to be different, but IIUC since the targets are in different folders (`scip-treesitter` vs `scip-treesitter-cli`), it's fine if a library target in one folder has the same name as the binary target in another folder.

sg.config.yaml
varungandhi-src: My bad, I misspoke, I was thinking it would be part of the _Q1_ design goals. Not necessarily to arbitrary numbers of workers, but at least a handful, similar to Zoekt/gitserver.

So slightly pushing back.

If it's adding significant complexity to the implementation, we can change this back. But from the point of view of avoiding having to rework things, it'd be valuable to have this catch things like making sure we're propagating the worker identity consistently when work is done (instead of implicitly coding against a simpler configuration with a single worker).

dev/scip-treesitter-install.sh
keynmol: Ah, I see what you mean now. I think this commit addresses it: https://github.com/sourcegraph/sourcegraph/pull/59747/commits/d731bee46697ea24bfb03cebda2dcea39143f16f

sg.config.yaml
keynmol: Of the complexities I can imagine right now:

1. Load balancing API (no idea how it's currently done for services that are triggered from public API)
2. Sharding repo cache - we will probably neet some consistent hashing to make sure workers can scale up and down and not reclone repositories that are handled by other instances

For now, I'll add the two instances to the local configuration and we'll see if it causes any significant complexity.

keynmol: Done in https://github.com/sourcegraph/sourcegraph/pull/59747/commits/29345383fa477488e0a0373b5ce787b1e4771fe9

docker-images/syntax-highlighter/crates/scip-treesitter-cli/BUILD.bazel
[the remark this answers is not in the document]
varungandhi-src: Filed an issue here with one potential way of simplifying this to 2 library and 2 binary crates.

https://github.com/sourcegraph/sourcegraph/issues/59792

varungandhi-src (approved):
No major outstanding concerns apart from existing comments. Let's get this ball rolling.

sg.config.yaml
varungandhi-src: For those aspects, we can have a simpler/no-op implementation for the first pass.
- Hard-code number of workers (no automatic scaling).
- No extra load-balancing/redundancy.
- Decide repo simply based on a simple `hash(repo_uri) % N` instead of consistent hashing.

PR #59752, opened 2024-01-22
Search: reduce display limit to 500

camdencheek:
This reduces the default display limit from 1500 to 500. Note: this is _not_ the same as the default `count:`, so we will still collect and aggregate up to the default `count:` on the backend for the purpose of calculating more exhaustive filters, but we will only send up to 500 results to the client for display. 

Now that our default `count:` is higher and we're streaming larger chunk matches, stream sizes can get unacceptably large. For the purpose of searching code, 500 results should be enough to get an idea of whether your search is refined enough, and our new filters panel will help gently suggest ways to refine it better, so I think reducing the default display limit is an acceptable tradeoff here.

Another option would be to have a separate cutoff where we stream the matched ranges back to the client, but not the actual contents so the client could fetch the contents lazily if someone scrolls down that far. However, that's a significantly larger change, and this seems like a reasonable trade-off for now.

## Test plan

N/A

### DIFF
--- a/client/shared/src/search/stream.ts
+++ b/client/shared/src/search/stream.ts
@@ -504,7 +504,7 @@ function initiateSearchStream(
         zoektSearchOptions,
         featureOverrides,
         searchMode = SearchMode.Precise,
-        displayLimit = 1500,
+        displayLimit = 500,
         sourcegraphURL = '',
         chunkMatches = false,
     }: StreamSearchOptions,


BolajiOlajide (approved):

camdencheek:
@taiyab I noticed while looking at this that we do not have a "display limit hit" message at the bottom of search results, so there is no indication that the results we actually show are different from the counts we display. I _think_ this is already on your radar, but wanted to ping ya about it in case since this change will make it more likely that users see the bottom of the search results page.

camdencheek:
Closing since the search platform team is exploring alternatives, and the PR itself is trivial.

taiyab:
> @taiyab I noticed while looking at this that we do not have a "display limit hit" message at the bottom of search results, so there is no indication that the results we actually show are different from the counts we display. I _think_ this is already on your radar, but wanted to ping ya about it in case since this change will make it more likely that users see the bottom of the search results page.

It actually wasn't on my list so thanks for bringing it up.

keegancsmith:
Re-opening after investigation. A lower display limit will just make this harder to trigger, but if your query hits these 1mb lines even with a 500 display limit life sucks. So our fix here needs to be more content size based. I think for now lets go with reducing the limit since it is harder to hit.

keegancsmith:
An interesting note from my investigation about why lower display limits can be helpful:

> Interestingly the display=1 did a lot more work! This is because there was no back pressure from the client. So unexpectedly sending less to the client results in better filters. I suspect we can fix this by introducing a bigger buffer at the HTTP layer to help reduce the back pressure

PR #59753, opened 2024-01-23
Search: fix source for honeycomb logging

camdencheek:
This updates the search stream to properly calculate the source for honeycomb events. Currently, our sentinel metrics in Honeycomb only represent GraphQL requests because we do not properly set the source string. This means that we're not getting any good metrics on latency (which we don't measure for GraphQL since it doesn't make sense), only duration.

## Test plan

Attached a debugger and saw that the string was correctly set.

### DIFF
--- a/cmd/frontend/internal/search/search.go
+++ b/cmd/frontend/internal/search/search.go
@@ -91,11 +91,13 @@ func (h *streamHandler) serveHTTP(r *http.Request, tr trace.Trace, eventWriter *
 	if err != nil {
 		return err
 	}
+	source := GuessSource(r)
 	tr.SetAttributes(
 		attribute.String("query", args.Query),
 		attribute.String("version", args.Version),
 		attribute.String("pattern_type", args.PatternType),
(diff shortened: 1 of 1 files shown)


cmd/frontend/internal/search/search.go
```diff
@@ -206,7 +218,7 @@ func logSearch(ctx context.Context, logger log.Logger, alert *search.Alert, err
 		_ = searchhoney.SearchEvent(ctx, searchhoney.SearchEventArgs{
 			OriginalQuery: originalQuery,
 			Typ:           "stream",
-			Source:        string(trace.RequestSource(ctx)),
```
camdencheek, line 209: `trace.RequestSource(ctx)` only works if a middleware has already set the request source on the context. However, we don't apply that middleware to stream searches, only GraphQL. Instead, just calculate it directly because this doesn't need to be a middleware thing anyways.

jtibshirani (approved):

PR #59755, opened 2024-01-23
Remove smart search preview on no results page

jtibshirani:
We're planning to ship the new `keyword` patterntype search as default, and
remove the smart search toggle. So we also need to remove the smart search
preview on the "no results" page.

For simplicity, this change removes it completely (not only when the keyword
feature flag is enabled).

Relates to #58815

## Test plan

Manual testing


jtibshirani:
Here's a before-and-after. I think the change stands on its own, even without considering the "keyword search" feature. Since Smart Search is on by default, it seems wasteful and not helpful to run an extra smart search on the "no results" page.

Screenshot 2024-01-22 at 5 05 37 PM
Screenshot 2024-01-22 at 5 04 47 PM

camdencheek (approved):
LGTM!

PR #59759, opened 2024-01-23
svelte: Upgrade to SvelteKit v2

fkling:
This includes upgrading to Vite 5. I followed https://kit.svelte.dev/docs/migrating-to-sveltekit-2 and ran `pnpx svelte-migrate@latest sveltekit-2` which took care of the Svelte changes (of which there were only a few).

Because SK2 doesn't automatically await top-level promises anymore there is no need to have the artificaly `deferred` second level. I removed that from all loaders. I also took the time to clean up the svelte config file a bit.

The switch to Vite 5 is tricker and I had to resort to setting `proxySsrExternalModules: true` to make CommonJS modules imported as ES modules work.
However using `pnpm` to alias `lodash` to `lodash-es` generally seems to be a good improvemnt, even if not strictly necessary for the upgrade (until we remove `proxySsrExternalModules: true`).

NOTE: Best reviewed with whitspace changes ignored.

## Test plan

`pnpm dev` and browsing pages
`pnpm build && pnpm preview` and browsing pages
`bazel build //client/web-sveltekit:web-sveltekit`
`pnpm vitest`
`pnpm test`
`pnpm check`


camdencheek (approved):
Nice!

PR #59760, opened 2024-01-23
Auto-redirect to SAMS

### LINKED ISSUE #59634: Add Dotcom auto-redirect to SAMS (unless operator)
The logic should be: “IF Sourcegraph op is the only signin option besides SAMS THEN auto-redirect EXCEPT if the ?sourcegraph-operator flag is there in the URL”

See [this thread](https://sourcegraph.slack.com/archives/C05SZB829D0/p1705337441038019?thread_ts=1705335253.881089&cid=C05SZB829D0)

vdavid:
- Fixes #59634

Notes:
- I've used an "if dotcom" check to avoid affecting enterprise or cloud instances.
- It redirects to any single non-built-in provider, but practically, this will only be SAMS on dotcom.
- I created a simple custom redirect note to make the UX nice.

## Test plan

See me testing this (except operators) in this [1-min Loom](https://www.loom.com/share/64499f7dd0c848e595a35eadb98f78c2)

Tested it with:
- ✅ several auth methods enabled → the sign-in page shows up
- ✅ a single auth provider enabled → I got auto-redirected
- ✅ a single auth provider plus Sourcegraph Operators enabled → I got auto-redirected
- ✅ a single auth provider plus Sourcegraph Operators enabled with the `sourcegraph-operators` query attribute set → the sign-in page shows up

### DIFF
--- a/client/web/src/auth/SignInPage.tsx
+++ b/client/web/src/auth/SignInPage.tsx
@@ -78,6 +78,28 @@ export const SignInPage: React.FunctionComponent<React.PropsWithChildren<SignInP
     }
 
     const thirdPartyAuthProviders = nonBuiltinAuthProviders.filter(provider => shouldShowProvider(provider))
+    // If there is only one auth provider on dotcom, we want to redirect to it directly.
+    if (
+        context.sourcegraphDotComMode &&
+        thirdPartyAuthProviders.length === 1 &&
+        !thirdPartyAuthProviders[0].isBuiltin
+    ) {
+        window.location.replace(thirdPartyAuthProviders[0].authenticationURL)
+        return (
+            <>
+                <PageTitle title="Signing in..." />
+                <AuthPageWrapper
+                    title="Redirecting to sign in..."
+                    sourcegraphDotComMode={context.sourcegraphDotComMode}
+                    className={styles.wrapper}
+                >
+                    <Alert className="mt-3" variant="info">
+                        You are being redirected to sign in with {thirdPartyAuthProviders[0].displayName}.
+                    </Alert>
+                </AuthPageWrapper>
+            </>
+        )
+    }
     const primaryProviders = thirdPartyAuthProviders.slice(0, context.primaryLoginProvidersCount)
     const moreProviders = thirdPartyAuthProviders.slice(context.primaryLoginProvidersCount)
 


client/web/src/auth/SignInPage.tsx
```diff
@@ -78,6 +78,28 @@ export const SignInPage: React.FunctionComponent<React.PropsWithChildren<SignInP
     }
 
     const thirdPartyAuthProviders = nonBuiltinAuthProviders.filter(provider => shouldShowProvider(provider))
+    // If there is only one auth provider on dotcom, we want to redirect to it directly.
```
unknwon, line 81:
```suggestion
    // If there is only one auth provider that is going to be displayed on dotcom, we want to redirect to it directly.
```

unknwon (approved):

thenamankumar (approved):
LGTM

client/web/src/auth/SignInPage.tsx
```diff
@@ -78,6 +78,27 @@ export const SignInPage: React.FunctionComponent<React.PropsWithChildren<SignInP
     }
 
     const thirdPartyAuthProviders = nonBuiltinAuthProviders.filter(provider => shouldShowProvider(provider))
+    // If there is only one auth provider that is going to be displayed on dotcom, we want to redirect to it directly.
+    if (
+        context.sourcegraphDotComMode &&
+        thirdPartyAuthProviders.length === 1
+    ) {
+        window.location.replace(thirdPartyAuthProviders[0].authenticationURL)
```
evict, line 86: Can we add a check to ensure that this starts with `https`?

evict (commented):
I am a bit concerned about potential for XSS so I'd recommend some verification that it's indeed a real URL. In addition to that, could we use React router for navigation instead of manipulating `window.location` directly?

vdavid:
> In addition to that, could we use React router for navigation instead of manipulating window.location directly?

No because that does just a URL rewrite and the redirection doesn't kick in. If I go to the nav bar and submit the URL again then the redirection happens, so I assume it happens on the server-side. Thus, we need a hard redirect.

client/web/src/auth/SignInPage.tsx
evict: My concern is XSS so I want some sort of assurance that it will always be what we want it to be. So if it's a relative URL can we make it always starts with `/`? Because if `javascript:alert(PAYLOAD)` is provided it will currently be executed.

evict: I just saw your comment here: https://github.com/sourcegraph/sourcegraph/pull/59760#issuecomment-1906285775. Resolved.

evict:
> > I am a bit concerned about potential for XSS so I'd recommend some verification that it's indeed a real URL.
> 
> @evict It comes from the server and it starts with `/.auth` (uses [this constant](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/cmd/frontend/auth/auth.go?L13)), so I believe we can trust it.
> 
> I'm grateful for the thorough review, btw! Please let me know if I answered everything.

Gotcha! Looks good, thank you! 👍

PR #59761, opened 2024-01-23
Guardrails autocomplete filtering.

### LINKED ISSUE #59277: Wire guardrails with autocomplete
- For an autocomplete of over 10 lines
- issue an attribution search request
- if positive (any hits) - hide the autocomplete

API for attribution search is being developed here: https://github.com/sourcegraph/cody/pull/2521/files#diff-457c009f122a22d3183157a2b15c5db7e3c556531922296846568bc38d9e8de1R30

cbart:
Closes #59277

This pull request adds a back-end autocomplete filtering on the code completions streaming handler.

In English:

- Completion events come in with increasingly long prefixes of the completion (think of message content when ChatGPT keeps on typing)
- At first completions are small, and are not considered for attribution (10 lines needed)
- The first 10-line-long completion prefix arrives
  - At this point attribution search fires (this one time) with the 10-lines long prefix
- If another completion event comes
  - Then its value is memoized, but not forwarded to the client
- Once attribution search finishes,
  - If result is permissive (no attribution), then forward the most recently memoized completion event to the client, and keep forwarding the next ones
  - If result is not permissive (attribution found) stop forwarding anything to the client.

The filtering logic can be described using a Petri Net for easier reasoning:
Screenshot 2024-01-24 at 3 06 50 PM

## TODO

- [x] Check how client reacts to the error event.

## Test plan

- Unit test the concurrent stream filtering flow.
- Manual check with VS Code and printfs.
- Drawing a Petri Net with the filtered completion event flow ([RFC](https://docs.google.com/document/d/1zSxFDQPxZcn5b6yKx40etpJayoibVzj_Gnugzln1weI/edit#bookmark=id.u41qhnii8xx2))

## How to read a Petri Net

**tl;dr it's like a board game**

* _places_ are the circles; think of them as places where you can put a token (indicated by a dot) in a board game
* _transitions_ are the squares; transitions can _fire_ and then they take one token from each input place, and put token to each output place (inputs and outputs indicated by arrows, like you would think)
  * Specifically each input place _needs to_ have a token, and each output place _needs to_ be empty.
  * Firing is atomic.
* This is a simple model and allows only one token in each place.


internal/guardrails/attribution_filter_test.go
```diff
@@ -0,0 +1,378 @@
...
+		nextLine("5"),
+		nextLine("6"),
+		nextLine("7"),
+		nextLine("8"),
+		nextLine("9"),
+		nextLine("10"),
+		searchFinishes{search: search, canUseSnippet: false},
+		nextLine("11"),
+		nextLine("12"),
+	}
```
eseliger, line 195: very nice test bed!

eseliger (approved):
Thanks for the diagram and the nice test bed! Stared at this for a bit now, and nothing really stood out as outright wrong - stamping to approve, let me know if you want me to look at this harder (but probably tomorrow)

internal/guardrails/attribution_filter.go
```diff
@@ -0,0 +1,179 @@
...
+		return nil
+	}
+	return a.sink(e)
+}
+
+// smallEnough indicates whether snippet size is small enough not to consider
+// it for attribution search. At this point we run attribution search for
+// snippets 10 lines long or longer.
+func (a *CompletionsFilter) smallEnough(e types.CompletionResponse) bool {
+	return len(strings.Split(e.Completion, "\n")) < 10
```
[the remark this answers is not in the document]
cbart: 👍

cbart (commented):
Thanks!

internal/guardrails/attribution_filter_test.go
```diff
@@ -0,0 +1,431 @@
...
+	require.NoError(t, o.replay(context.Background(), f))
+	got := client.trimmedDiffs()
+	want := []string{
```
chwarwick, line 226: Does this match the intention of what guardrails is supposed to do? This reads to me like we stream over data to the client and once it hits a threshold we trigger the attribution request pausing any additional lines to be sent.  If we can attribute it to something know the client keeps the up to the threshold, and we discard anything additional without notifying that the snippet can be attributed.

I would think you would need to immediately quarantine results, and only release them if the response ended before the 10th line, or the attribution search returned empty.

cmd/frontend/internal/guardrails/init.go
```diff
@@ -50,3 +51,35 @@ func initDotcomAttributionService(observationCtx *observation.Context, db databa
 	searchClient := client.New(observationCtx.Logger, db, gitserver.NewClient("http.guardrails.search"))
 	return attribution.NewLocalSearch(observationCtx, searchClient)
 }
+
+func alwaysTrue(context.Context, string) (bool, error) {
```
chwarwick, line 55: I found `awaysTrue` confusing, thinking it meant that Attribution was true, not that the request was allowed.  `alwaysAllowed` might make that clearer to future readers

internal/guardrails/attribution_filter_test.go
cbart: Nope. This is as designed. The reason for this is that threshold is the attribution limit. According to legal noone is going to sue for less than 10 lines (in practice probably the limit is way more flexible). This is the tradeoff we strike between UX/latency and guarantees.

cmd/frontend/internal/guardrails/init.go
cbart: Good point!

PR #59763, opened 2024-01-23
Add Self Serve Cody API client to dotcom

thenamankumar:
This PR is the first of many to fetch Cody's subscription data from SSC (Self-Serve Cody) into dotcom and integrate at the required places. This PR only takes care of the "fetching" part. 

Right now, we know if the user has opted for a pro trial by looking at the `users.cody_pro_enabled_at` column in the dotcom database. 

From Feb 1st, tentatively, we will be allowing users to put in their credit cards on SSC, which will create a `trialling` subscription for them on SSC. We will need the combined info of dotcom and SSC to support the UI.

After the Feb 15th release, SSC will be the only source of truth for the user's Cody subscription. Part of the code, added as part of this PR, will be removed after the release.

Cody Clients and Cody Management page on dotcom uses dotcom's API to know about the user's Cody subscription info. To support that, we need to fetch subscription info from SSC into dotcom.

**As part of this PR, we are:**
1. Introducing SSC client in `/internal/ssc`, which allows us to fetch the user's Cody Subscription from SSC.
2. Consolidating data from dotcom & SSC to figure out the user's subscription in `internal/cody/subscription`.
3. Make fetching info from SSC controlled by a feature flag: `use-ssc-for-cody-subscription`.
4. Updating `user.codyProEnabled` in graphql API to use `internal/cody/subscription`, which integrates it with subscription data from SSC.
5. Removing `user.codyProEnabledAt` from graphql API. It is a redundant field and not used by any Cody Clients or the web. Confirmed by this [search](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/*+codyProEnabledAt&patternType=standard&sm=1) and double checking locally.
6. Adding unit tests.

**Next steps as part of follow-up PRs:**
1. Expose subscription status info from the graphql API. 
2. Update rate-limit logic at various places to use `interna/cody/subscription` for the user's SSC subscription info rather than dotcom's DB.

The above 2 next steps will make the PLG flow functional from the dotcom side. As a follow-up, we will also work on creating a whole new actor source for Cody Gateway to reach SSC for rate limit info directly.

## Test plan

Unit tests are added as part of this PR.

**Manual Testing Plan**
1. Set up and run SAMS locally.
4. Integrate SAMS as an openidconnect auth provider with dotcom locally. 
5. Create a user using the SAMS auth provider.
6. Enable feature flag `use-ssc-for-cody-subscription` from site admin.
7. Make a graphql API [request](https://sourcegraph.test:3443/api/console#%7B%22query%22%3A%22%7B%5Cn%20%20currentUser%20%7B%5Cn%20%20%20%20codyProEnabled%5Cn%20%20%7D%5Cn%7D%5Cn%22%7D) to confirm `false`. 
8. Visit `localhost:9992/cody` and go through the subscription flow to sign up for Cody Pro.
9. Make a graphql API [request](https://sourcegraph.test:3443/api/console#%7B%22query%22%3A%22%7B%5Cn%20%20currentUser%20%7B%5Cn%20%20%20%20codyProEnabled%5Cn%20%20%7D%5Cn%7D%5Cn%22%7D) again to confirm `true`.

### DIFF
--- a/cmd/frontend/graphqlbackend/schema.graphql
+++ b/cmd/frontend/graphqlbackend/schema.graphql
@@ -6577,10 +6577,6 @@ type User implements Node & SettingsSubject & Namespace {
     """
     codyProEnabled: Boolean!
     """
-    The date when the user enrolled for Cody Pro.
-    """
-    codyProEnabledAt: DateTime
-    """
     The start date of current billing period of Cody.
     """
     codyCurrentPeriodStartDate: DateTime
--- a/cmd/frontend/graphqlbackend/user.go
+++ b/cmd/frontend/graphqlbackend/user.go
@@ -6,12 +6,12 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
-	"time"
 
 	"github.com/graph-gophers/graphql-go"
 	"github.com/graph-gophers/graphql-go/relay"
 
 	"github.com/sourcegraph/sourcegraph/internal/accesstoken"
+	"github.com/sourcegraph/sourcegraph/internal/cody"
 
 	"github.com/keegancsmith/sqlf"
 	"github.com/sourcegraph/log"
@@ -180,20 +180,17 @@ func (r *UserResolver) CreatedAt() gqlutil.DateTime {
 	return gqlutil.DateTime{Time: r.user.CreatedAt}
 }
 
-func (r *UserResolver) CodyProEnabledAt(ctx context.Context) *gqlutil.DateTime {
+func (r *UserResolver) CodyProEnabled(ctx context.Context) (bool, error) {
 	if !envvar.SourcegraphDotComMode() {
-		return nil
+		return false, errors.New("this feature is only available on sourcegraph.com")
 	}
 
-	if r.user.CodyProEnabledAt == nil {
-		return nil
+	subscription, err := cody.SubscriptionForUser(ctx, r.db, *r.user)
+	if err != nil {
+		return false, err
 	}
 
-	return &gqlutil.DateTime{Time: *r.user.CodyProEnabledAt}
-}
-
-func (r *UserResolver) CodyProEnabled(ctx context.Context) bool {
-	return r.CodyProEnabledAt(ctx) != nil
+	return subscription.Plan == cody.UserSubscriptionPlanPro, nil
 }
 
 func (r *UserResolver) CodyCurrentPeriodChatLimit(ctx context.Context) (int32, error) {
@@ -229,11 +226,7 @@ func (r *UserResolver) CodyCurrentPeriodCodeLimit(ctx context.Context) (int32, e
 }
 
 func (r *UserResolver) CodyCurrentPeriodChatUsage(ctx context.Context) (int32, error) {
-	if !envvar.SourcegraphDotComMode() {
-		return 0, errors.New("this feature is only available on sourcegraph.com")
-	}
-
-	currentPeriodStartDate, err := r.CodyCurrentPeriodStartDate(ctx)
+	currentPeriodStartDate, _, err := r.codyCurrentPeriodDateRange(ctx)
 	if err != nil {
 		return 0, err
 	}
@@ -279,11 +272,7 @@ func (r *UserResolver) CodyCurrentPeriodChatUsage(ctx context.Context) (int32, e
 }
 
 func (r *UserResolver) CodyCurrentPeriodCodeUsage(ctx context.Context) (int32, error) {
-	if !envvar.SourcegraphDotComMode() {
-		return 0, errors.New("this feature is only available on sourcegraph.com")
-	}
-
-	currentPeriodStartDate, err := r.CodyCurrentPeriodStartDate(ctx)
+	currentPeriodStartDate, _, err := r.codyCurrentPeriodDateRange(ctx)
 	if err != nil {
 		return 0, err
 	}
@@ -314,22 +303,6 @@ func (r *UserResolver) CodyCurrentPeriodEndDate(ctx context.Context) (*gqlutil.D
 	return endDate, err
 }
 
-type currentTimeCtxKey int
-
-const mockCurrentTimeKey currentTimeCtxKey = iota
-
-func currentTimeFromCtx(ctx context.Context) time.Time {
-	t, ok := ctx.Value(mockCurrentTimeKey).(*time.Time)
-	if !ok || t == nil {
-		return time.Now()
-	}
-	return *t
-}
-
-func withCurrentTimeMock(ctx context.Context, t time.Time) context.Context {
-	return context.WithValue(ctx, mockCurrentTimeKey, &t)
-}
-
 func (r *UserResolver) codyCurrentPeriodDateRange(ctx context.Context) (*gqlutil.DateTime, *gqlutil.DateTime, error) {
 	if !envvar.SourcegraphDotComMode() {
 		return nil, nil, errors.New("this feature is only available on sourcegraph.com")
@@ -341,48 +314,10 @@ func (r *UserResolver) codyCurrentPeriodDateRange(ctx context.Context) (*gqlutil
 		return nil, nil, err
 	}
 
-	// to allow mocking current time during tests
-	currentDate := currentTimeFromCtx(ctx)
+	// TODO(naman): integrate this with cody.SubscriptionForUser
+	startDate, endDate := cody.PreSSCReleaseCurrentPeriodDateRange(ctx, *r.user)
 
-	subscriptionStartDate := r.user.CreatedAt
-	gaReleaseDate := time.Date(2023, 12, 14, 0, 0, 0, 0, subscriptionStartDate.Location())
-
-	if !currentDate.Before(gaReleaseDate) && subscriptionStartDate.Before(gaReleaseDate) {
-		subscriptionStartDate = gaReleaseDate
-	}
-
-	if r.user.CodyProEnabledAt != nil {
-		subscriptionStartDate = *r.user.CodyProEnabledAt
-	}
-
-	targetDay := subscriptionStartDate.Day()
-	startDayOfTheMonth := targetDay
-	endDayOfTheMonth := targetDay - 1
-	startMonth := currentDate
-	endMonth := currentDate
-
-	if currentDate.Day() < targetDay {
-		// Set to target day of the previous month
-		startMonth = currentDate.AddDate(0, -1, 0)
-	} else {
-		// Set to target day of the next month
-		endMonth = currentDate.AddDate(0, 1, 0)
-	}
-
-	daysInStartingMonth := time.Date(startMonth.Year(), startMonth.Month()+1, 0, 0, 0, 0, 0, startMonth.Location()).Day()
-	if startDayOfTheMonth > daysInStartingMonth {
-		startDayOfTheMonth = daysInStartingMonth
-	}
-
-	daysInEndingMonth := time.Date(endMonth.Year(), endMonth.Month()+1, 0, 0, 0, 0, 0, endMonth.Location()).Day()
-	if endDayOfTheMonth > daysInEndingMonth {
-		endDayOfTheMonth = daysInEndingMonth
-	}
-
-	startDate := &gqlutil.DateTime{Time: time.Date(startMonth.Year(), startMonth.Month(), startDayOfTheMonth, 0, 0, 0, 0, startMonth.Location())}
-	endDate := &gqlutil.DateTime{Time: time.Date(endMonth.Year(), endMonth.Month(), endDayOfTheMonth, 23, 59, 59, 59, endMonth.Location())}
-
-	return startDate, endDate, nil
+	return &gqlutil.DateTime{Time: startDate}, &gqlutil.DateTime{Time: endDate}, nil
 }
 
 func (r *UserResolver) UpdatedAt() *gqlutil.DateTime {
--- a/internal/cody/subscription.go
+++ b/internal/cody/subscription.go
@@ -12,6 +12,10 @@ import (
 )
 
 const USE_SSC_FEATURE_FLAG = "use-ssc-for-cody-subscription"
+const CODY_PRO_TRIAL_ENDED = "cody-pro-trial-ended"
+
+const SAMS_SERVICE_ID = "https://accounts.sgdev.org"
+const SAMS_SERVICE_TYPE = "openidconnect"
 
 type UserSubscriptionPlan string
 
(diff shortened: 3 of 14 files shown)


internal/cody/utils.go
```diff
@@ -0,0 +1,69 @@
...
+		return time.Now()
+	}
+	return *t
+}
+
+func withCurrentTimeMock(ctx context.Context, t time.Time) context.Context {
+	return context.WithValue(ctx, mockCurrentTimeKey, &t)
+}
+
+func PreSSCReleaseCurrentPeriodDateRange(ctx context.Context, user types.User) (time.Time, time.Time) {
```
thenamankumar, line 25: These tests are old and simply moved here from the graphql resolver, along with the function.

internal/cody/subscription.go
```diff
@@ -0,0 +1,147 @@
...
+	// ApplyProRateLimits is true if the user has a valid subscription status.
+	ApplyProRateLimits bool
+	// CurrentPeriodStartAt is the start date of the current billing cycle.
+	CurrentPeriodStartAt time.Time
+	// CurrentPeriodEndAt is the end date of the current billing cycle.
+	CurrentPeriodEndAt time.Time
+}
+
+// consolidateSubscriptionDetails consolidates the subscription details from dotcom and SSC.
+// Post Feb 15, 2024, this function will be removed and the logic will be simplified to us the SSC info only.
```
chrsmith, line 39: Super-minor suggestion: Your comment was good enough, but I would (personally, by preference) word it slightly differently.

```suggestion
// consolidateSubscriptionDetails merges the subscription data available on dotcom and SCC.
// This is needed while transitioning to use SCC as the source of truth for all subscription data. (Which should happen ~Q1/2024.)
```

Also, could you please file a ticket to track actually doing this work? That way we can replace the bottom line with the following. (So we can actually track and follow-through with our plan here.)

```go
// TODO[sourcegraph#12345]: Update dotcom to use SSC as the source of truth for all subscription data.
```

internal/cody/subscription.go
```diff
@@ -0,0 +1,147 @@
...
+func consolidateSubscriptionDetails(ctx context.Context, user types.User, subscription *ssc.Subscription) (*UserSubscription, error) {
+	now := currentTimeFromCtx(ctx)
+	febReleaseDate := time.Date(2024, 02, 14, 23, 59, 59, 59, now.Location())
```
chrsmith, line 42: 1. It is best to never, ever, ever deal with locals and timezones since it is an infinite source of bugs. All times are in UTC. Even if we wanted to be more friendly here, the behavior would be inconsistent and lead to bugs anyways.
2. "Midnight on 2/14" is a little too cute, IMHO. Let's just say 2/15.

```suggestion
	febReleaseDate := time.Date(2024, 02, 15, 0, 0, 0, 0, time.UTC)
```

chrsmith: Actually, rather than hard-coding a date at all, why don't you introduce a new boolean feature flag. This way we can determine the behavior ourselves, and not be forced into it at a certain time.

e.g.

```go
offerCodyProFreeTrailFromDotcom := flags.Bool(true)`
```

internal/cody/subscription.go
```diff
@@ -0,0 +1,147 @@
...
+	isAfterFebReleaseDate := now.After(febReleaseDate)
+
+	// The user has put in their cc and signed up for Cody Pro on SSC
```
chrsmith, line 45:
```suggestion
	// If subscription information is available from SSC, we use that.
	// And just ignore what is stored in dotcom. (Since they've already
	// been migrated so to speak.)
```

internal/cody/subscription.go
```diff
@@ -0,0 +1,147 @@
...
+	if subscription != nil {
+		currentPeriodStart, err := time.ParseInLocation(time.RFC3339, subscription.CurrentPeriodStart, now.Location())
```
chrsmith, line 47: The ISO string sent across the wire will include the timezone already. But unless this is something we already do in the dotcom backend, I would strongly urge us to not localize that time to the user's location. And just pass it around as being in UTC. That way, it's clear what the frontend is receiving. (And the _browser_ is responsible for putting that into the user's locale.) ... and I'd be shocked if the dotcom server's clock is not using UTC for its timezone/location either.

```suggestion
		currentPeriodStart, err := time.ParseInLocation(time.RFC3339, subscription.CurrentPeriodStart, time.UTC)
```

internal/cody/subscription.go
```diff
@@ -0,0 +1,147 @@
...
+		return &UserSubscription{
+			Status:               subscription.Status,
+			Plan:                 UserSubscriptionPlanPro,
+			ApplyProRateLimits:   applyProRateLimits,
+			CurrentPeriodStartAt: currentPeriodStart,
+			CurrentPeriodEndAt:   currentPeriodEnd,
+		}, nil
+	}
+
+	currentPeriodStartAt, currentPeriodEndAt := PreSSCReleaseCurrentPeriodDateRange(ctx, user)
```
chrsmith, line 68:
```suggestion
	// If the user doesn't have a subscription in the SSC backend, then we need
	// synthesize one using the data available on dotcom.
	currentPeriodStartAt, currentPeriodEndAt := PreSSCReleaseCurrentPeriodDateRange(ctx, user)
```

internal/cody/subscription.go
```diff
@@ -0,0 +1,147 @@
...
+
+	// after the release, all the users will be on the free plan if they have no subscription on SSC
+	if isAfterFebReleaseDate || user.CodyProEnabledAt == nil {
```
chrsmith, line 71: For clarity I think we should rearrange this block of code so that it is the "last" part of the function.

e.g. structuring the function so it is:

```go
// If subscription information is available from SSC, we use that.
// And just ignore what is stored in dotcom. (Since they've already
// been migrated so to speak.)
if subscription != nil {
    ...
}

// The user hasn't registered with SSC yet, but has signed up for the Cody Pro free trial.
// So we give them access to the higher usage limits. At least until we stop offering
// the free trial offer on .com.
if user.CodyProEnabledAt != nil && offerCodyProFreeTrailFromDotcom {
    ...
}

// The user didn't opt for the Cody free trial, or we no longer offer it.
// So we default to the user being on the free tier.
// (Though as per TODO[sourcegraph#1234], we'll stop hitting this case once
// we have migrated all Sourcegraph users to SAMS/SSC. And will fetch subscription
// information exclusively from that service.)
return &UserSubscription {
   ...
}
```

IMHO that is easier to follow. WDYT?

internal/cody/subscription.go
```diff
@@ -0,0 +1,147 @@
...
+	}, nil
+}
+
+// getSAMSAccountIDForUser returns the user's SAMS account ID from users_external_accounts table.
+func getSAMSAccountIDForUser(ctx context.Context, db database.DB, user types.User) (string, error) {
+	// TODO: make service_id configurable between accounts.sourcegraph.com and accounts.sgdev.org using a feature flag for testing.
+	accounts, err := db.UserExternalAccounts().List(ctx, database.ExternalAccountsListOptions{
+		UserID:      user.ID,
+		ServiceType: "openidconnect",
+		ServiceID:   "https://accounts.sourcegraph.com",
```
chrsmith, line 105: Can you move this to a `const` defined in at the top of this file? And I suspect `cody/subscription` won't be the final location of `SAMSProductionServiceID`...

Also, for now, let's set this to `https://accounts.sgdev.org`, as we haven't set up SSC for production yet. And so there's no way we can even test things out yet.

internal/cody/subscription.go
```diff
@@ -0,0 +1,147 @@
...
+		Plan:                 UserSubscriptionPlanPro,
+		ApplyProRateLimits:   true,
+		CurrentPeriodStartAt: currentPeriodStartAt,
+		CurrentPeriodEndAt:   currentPeriodEndAt,
+	}, nil
+}
+
+// getSAMSAccountIDForUser returns the user's SAMS account ID from users_external_accounts table.
+func getSAMSAccountIDForUser(ctx context.Context, db database.DB, user types.User) (string, error) {
+	// TODO: make service_id configurable between accounts.sourcegraph.com and accounts.sgdev.org using a feature flag for testing.
```
chrsmith, line 101: Could you please file a ticket to track this work? Let's try to be clearer about all of the undone things so that we can have a better understanding of things in-flight and their sequencing.

internal/cody/subscription.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	if len(accounts) > 0 {
+		return accounts[0].AccountID, nil
+	}
+
+	return "", nil
+}
+
+type SSCClient interface {
```
chrsmith, line 121:
```suggestion
// SSCClient is the interface for making requests to the Self-Service Cody backend.
// This uses a REST API exposed from the service, and not the GraphQL API (which is
// instead used for user-facing frontend queries.)
type SSCClient interface {
```

internal/ssc/ssc.go
```diff
@@ -0,0 +1,73 @@
...
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/sourcegraph/sourcegraph/internal/conf"
+	"github.com/sourcegraph/sourcegraph/internal/httpcli"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+)
+
+type Client struct {
```
chrsmith, line 14: I know it is annoying boilerplate, but could you make this an `interface` and then a separate struct with its implementation? That way we can use `mockgen` for tests.

```go
type Client interface {
    FetchSubscriptionBySAMSAccountID(samsAccountID string) (*Subscription, error)
}

type sscClient struct {
	baseUrl     string
	secretToken string
}

// Trigger compiler error if sscClient doesn't implement Client.
var _ Client = (*sscClient)(nil)
```

internal/ssc/ssc.go
```diff
@@ -0,0 +1,73 @@
...
+func (c *Client) FetchSubscriptionBySAMSAccountID(samsAccountID string) (*Subscription, error) {
+	if c.baseUrl == "" {
+		return nil, errors.New("SSC base url is not set")
+	}
+
+	if c.secretToken == "" {
+		return nil, errors.New("SSC secret token is not set")
+	}
+
+	req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("%s/rest/svc/subscription/%s", c.baseUrl, samsAccountID), nil)
```
chrsmith, line 30: There's enough HTTP-specific boilerplate that I'd suggest you tease the first N-lines of this function into a separate method. So the body of `FetchSubscriptionBySAMSAccountID` can be distilled to just the relevant parts:

```go
// FetchSubscriptionBySAMSAccountID returns the user's Cody subscription for the sams_account_id.
// It returns nil, nil if the user does not have a Cody Pro subscription.
func (c *Client) FetchSubscriptionBySAMSAccountID(samsAccountID string) (*Subscription, error) {
    var responseObj Subscription
    url := fmt.Sprintf("%s/rest/svc/subscription/%s", c.baseUrl, samsAccountID)
    code, err := c.sendRequest(http.MethodGet, url, &responseObj)
    switch code {
         ...
    }
}
```

internal/ssc/ssc.go
```diff
@@ -0,0 +1,73 @@
...
+	"io"
+	"net/http"
+
+	"github.com/sourcegraph/sourcegraph/internal/conf"
+	"github.com/sourcegraph/sourcegraph/internal/httpcli"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+)
+
+type Client struct {
+	baseUrl     string
```
chrsmith, line 15: Golang naming conventions would have us use `baseURL`. That is, acronyms are in ALLCAPS and not camelCased.

chrsmith (commented):
Lots and lots of PR feedback, but the majority of it is just tweaking comments to improve clarity. This all looks good to me, but I'd like to take another pass. (e.g. I didn't take a close look at the specific of the unit test in this review.)

- 👍 to moving the "dotcom + SSC subscription" logic into its own `cody/subscription` package. That's a great way to make it easier for us to transition towards using SSC subscription data exclusively.
- 👍 to having all of the SSC logic in a dedicated `ssc` package. As we already know we will need to rework the authentication to use SAMS in the near future.

internal/ssc/ssc.go
thenamankumar: I have created a separate interface but in `/internal/cody/subscription`.

I'll move the interface here and swap the interface <> struct names.

I didn't know about mockgen, will use that in the tests.

cmd/frontend/graphqlbackend/user.go
```diff
@@ -279,11 +272,7 @@ func (r *UserResolver) CodyCurrentPeriodChatUsage(ctx context.Context) (int32, e
 }
 
 func (r *UserResolver) CodyCurrentPeriodCodeUsage(ctx context.Context) (int32, error) {
-	if !envvar.SourcegraphDotComMode() {
```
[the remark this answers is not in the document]
thenamankumar: We aren't, the check was getting applied twice, so I removed it from here. It is also there in the `r.codyCurrentPeriodDateRange` func.

cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -6577,10 +6577,6 @@ type User implements Node & SettingsSubject & Namespace {
     """
     codyProEnabled: Boolean!
     """
-    The date when the user enrolled for Cody Pro.
-    """
-    codyProEnabledAt: DateTime
```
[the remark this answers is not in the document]
thenamankumar: We are not using this in a single place. I have performed this [search](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/*+codyProEnabledAt&patternType=standard&sm=1) and double checking locally to confirm

internal/cody/subscription.go
thenamankumar: I'll update. 

But only that you don't need to submit any PRs to toggle feature flags. You should be "entitled" to do so from site-admin.

cmd/frontend/graphqlbackend/user.go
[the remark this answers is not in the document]
chrsmith: Hrm. In that case, I'd suggest actually keeping this code in here.

We would want these types of checks to happen as soon as possible. (And not rely on a side effect of an unexpected method.)

Having redundant checks like this isn't the end of the world, and IMO it's far more important to make the code clearer and easier to reason about. (e.g. identify that yes, this particular GraphQL resolver only applies to dotcom mode.)

internal/cody/subscription.go
thenamankumar: It mostly handles timezone craziness within a unit test in a CI environment.

internal/ssc/ssc.go
chrsmith: Gotcha. Sorry, I must have missed the interface definition in that other package. (But it seems like the canonical definition of the SSC client service area should be in the `ssc` package anyways.)

Mockgen is kinda neat, but as far as any tests you write now, it would probably be overkill to configure it for this `ssc` package. Since having the test case define an implementation of the `Client` interface with its one-method would probably be a lot simpler. (But as the SSC client grows, that'll change, etc. etc.)

If you want examples though, we have it set up in the `accounts.sourcegraph.com` repo and it's been pretty easy to work with!

chrsmith (approved):
Approving to unblock @thenamankumar and make way for other PRs to wire things together.

Meta points being to be diligent about filing tickets and calling undone things out in `TODO[#xxx]: ...` comments so that we can more quickly resolve them.

internal/cody/subscription.go
vdavid: @thenamankumar If we need to flip flags at certain times, please add the date/time for those actions in the [launch plan](https://docs.google.com/document/d/1bQkgy4g6wZyZOCgh5lGyjigiwIHNZyNH-Tc2vhXt-pU/edit#heading=h.v5j7go2zayy), too, with also marking who does it, or clearly marking that it's unassigned. (Let's assume that this PR will be merged.) Thanks!

cmd/frontend/graphqlbackend/user.go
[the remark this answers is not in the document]
thenamankumar: To keep the surface area of this PR limited. I'll make the suggested improvements in the next PR, which will be particularly on graphql API changes.

cmd/frontend/graphqlbackend/user.go
```diff
@@ -341,48 +314,10 @@ func (r *UserResolver) codyCurrentPeriodDateRange(ctx context.Context) (*gqlutil
...
-
-	daysInEndingMonth := time.Date(endMonth.Year(), endMonth.Month()+1, 0, 0, 0, 0, 0, endMonth.Location()).Day()
-	if endDayOfTheMonth > daysInEndingMonth {
-		endDayOfTheMonth = daysInEndingMonth
-	}
-
-	startDate := &gqlutil.DateTime{Time: time.Date(startMonth.Year(), startMonth.Month(), startDayOfTheMonth, 0, 0, 0, 0, startMonth.Location())}
-	endDate := &gqlutil.DateTime{Time: time.Date(endMonth.Year(), endMonth.Month(), endDayOfTheMonth, 23, 59, 59, 59, endMonth.Location())}
+	// TODO(naman): integrate this with cody.SubscriptionForUser
+	startDate, endDate := cody.PreSSCReleaseCurrentPeriodDateRange(ctx, *r.user)
```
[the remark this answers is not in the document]
thenamankumar: We do not need both, and as mentioned in the comment, I'll update it in the graphql API specific upcoming PR. Wanted to keep the API specific changes minimal in this PR.

internal/cody/subscription.go
thenamankumar: https://github.com/sourcegraph/sourcegraph/issues/59785

internal/cody/subscription.go
thenamankumar: I will be resolving this as part of the next PR.

internal/cody/subscription.go
thenamankumar: Updated.

internal/cody/subscription.go
thenamankumar: I have decided the name of the feature flag to be `cody-pro-trial-ended` and won't be changing this as clients depends on it. 

Updated the doc here: https://docs.google.com/document/d/1p8V_c5wkmNgnCshRdlnjbHqxgkYLVlvpTyDdbHu4VSY/edit#heading=h.4r8kivvuokd0

thenamankumar: And not wasting any time, I am rather including the changes in this PR only.

cmd/frontend/graphqlbackend/user.go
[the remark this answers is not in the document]
vdavid: This seems like a small change to me (revert removing the check and add a comment), but if you create an issue to track it, I'm fine if we merge it like this. Please link the issue here if you go with that to make sure we don't forget. (I know it's a small thing but issues are cheap, too)

internal/cody/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+	"time"
+
+	"github.com/sourcegraph/sourcegraph/internal/database"
+	"github.com/sourcegraph/sourcegraph/internal/featureflag"
+	"github.com/sourcegraph/sourcegraph/internal/ssc"
+	"github.com/sourcegraph/sourcegraph/internal/types"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+)
+
+const USE_SSC_FEATURE_FLAG = "use-ssc-for-cody-subscription"
```
vdavid, line 14: Would be great to have comments on both of these flags.

internal/cody/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+const CODY_PRO_TRIAL_ENDED = "cody-pro-trial-ended"
```
vdavid, line 15: Is there a reason why the first feature flag has "FEATURE_FLAG" in its name and this doesn't? I suggest using the same pattern for both.

vdavid (approved):
Left a few comments. Approved to unblock

PR #59765, opened 2024-01-23
svelte: Improve search stream handling wrt filters

fkling:
Due to data preloading, hovering over sidebar filters will send a new, possibly unnecessary, search query. Additionally we only make use of client side caching when navigating back in the browser history. I.e. clicking on a filter in the sidebar will always trigger a new query.

This PR addresses both of these issues:
- We now keep track of ongoing requests and cancel them if a new one was started and the previous one isn't complete.
- I added a new, opaque, URL parameter to indicate whether or not the client side cache should be used. Now re-selecting/unselecting filters feels instant. This opens the new problem of letting users re-run the current query with applied filters, but I think we can solve that iteratively. Likewise for purging the cache (if deemed necessary).

Note: Unlike the React app, this does not only cache the previous query but any query. Until now the cache was only used when using the browser back button to ensure that the user gets back to the same page. Navigating to search results in any other way (except now also filters) will _not_ use the cache.

## Test plan

Quickly hovered over multiple filters and saw that all but the last request didn't complete (which seems to indicate that they are canceled).

If a search result is in the cache, selecting/deselecting a filter is instant.


client/web-sveltekit/src/lib/search/index.ts
```diff
@@ -0,0 +1,21 @@
...
+    /**
+     * Original query + filters.
+     */
+    filteredQuery: string | undefined
+}
+
+export function parseExtendedSearchURL(search: string): ExtendedParsedSearchURL {
+    const parsedQuery = parseSearchURL(search)
+    const filters = parseSearchFilters(search)
+    const filteredQuery = filters ? parsedQuery.query + ' ' + filters : parsedQuery.query
```
camdencheek, line 14: note: this is another spot where we should paranthesize the query if there are and/or nodes. I don't think we do this for the react app yet either though. I've added an issue to the sprint to follow up on this.

client/web-sveltekit/src/routes/search/SearchResults.svelte
```diff
@@ -123,13 +124,13 @@
...
                         <a
-                            href={getQueryURL({
+                            href="{getQueryURL({
                                 searchMode: $queryState.searchMode,
                                 patternType: $queryState.patternType,
                                 caseSensitive: $queryState.caseSensitive,
                                 searchContext: $queryState.searchContext,
                                 query: filter.getQuery($queryState.query),
-                            })}
+                            })}&{USE_CLIENT_CACHE_QUERY_PARAMETER}"
```
camdencheek, line 133: why not add this as a parameter to `getQueryURL`?

client/web-sveltekit/src/routes/search/+page.ts
```diff
@@ -14,23 +15,95 @@ import {
...
-const cache: Record<string, Observable<AggregateStreamingSearchResults | undefined>> = {}
+interface SearchStreamCacheEntry {
+    searchStream: Observable<AggregateStreamingSearchResults | undefined>
+    complete: boolean
+}
+
+/**
+ * CachingStreamManager helps caching and canceling search streams in the browser.
+ */
+class CachingStreamManager {
```
camdencheek, line 30: Am I reading this right that we never clear the cache? Result sets can get pretty large, so I'd be a little concerned about a search/browsing session collecting all the search results in memory.

camdencheek (approved):

client/web-sveltekit/src/lib/search/index.ts
vovakulikov: @camdencheek I added "smart" merge helper after your feedback. At the moment, react version handlers it properly if original query has "or" or "and" operators

@fkling here is an example, maybe we could reuse mergeFiltersAndQuery utility https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@736af892a39306ed0f23b7605af7a39a30ab62e1/-/blob/client/web/src/search/results/SearchResultsCacheProvider.tsx?L90:44-90:64

client/web-sveltekit/src/routes/search/+page.ts
fkling: Yep. Happy to implement a cache eviction strategy eventually...

PR #59766, opened 2024-01-23
truncate commit message in commit view if necessary

jasonhawkharris:
Before, we showed the full commit message on the commit view. Now, we give the user the option to see more, or less of the commit message.

## Test plan
Previewed locally. Passing CI/CD.

### DIFF
--- a/client/web/src/repo/commits/GitCommitNode.tsx
+++ b/client/web/src/repo/commits/GitCommitNode.tsx
@@ -21,6 +21,7 @@ import { GitCommitNodeByline } from './GitCommitNodeByline'
 
 import styles from './GitCommitNode.module.scss'
 
+const TRUNCATED_COMMIT_MESSAGE_LENGTH = 240
 export interface GitCommitNodeProps {
     node: GitCommitFields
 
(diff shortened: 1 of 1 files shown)


client/web/src/repo/commits/GitCommitNode.tsx
```diff
@@ -160,9 +161,35 @@ export const GitCommitNode: React.FunctionComponent<React.PropsWithChildren<GitC
...
+                        <Button
+                            variant="link"
+                            size="sm"
+                            display="inline"
+                            onClick={() => setTruncatedCommitMessage(true)}
+                        >
+                            see less
+                        </Button>
+                    </>
+                )}
```
fkling, line 192: You could shorten this a lot, no need for two JSX "trees". Store the commit message in a variable (full or sliced) and render that. Conditionally render the text of the button. And the event handler can be `setTruncatedCommitMessage(value => !value)`

jasonhawkharris: Good call. While optimizing I found another bug as well. Thanks for the call out!

client/web/src/repo/commits/GitCommitNode.tsx
```diff
@@ -83,6 +83,7 @@ export const GitCommitNode: React.FunctionComponent<React.PropsWithChildren<GitC
     wrapperElement: WrapperElement = 'div',
 }) => {
     const [showCommitMessageBody, setShowCommitMessageBody] = useState<boolean>(false)
+    const [truncatedCommitMessage, setTruncatedCommitMessage] = useState<boolean>(true)
```
camdencheek, line 86: nit: `truncatedCommitMessage` reads like this is the commit message that has been truncated. I think `truncateCommitMessage` is more clear.

camdencheek (approved):

client/web/src/repo/commits/GitCommitNode.tsx
```diff
@@ -157,11 +158,22 @@ export const GitCommitNode: React.FunctionComponent<React.PropsWithChildren<GitC
         </div>
     )
 
+    const handleTruncate = () => setTruncateCommitMessage(!truncateCommitMessage)
```
fkling, line 161: I'd just inline this function below.

client/web/src/repo/commits/GitCommitNode.tsx
```diff
@@ -157,11 +158,22 @@ export const GitCommitNode: React.FunctionComponent<React.PropsWithChildren<GitC
...
+    const showCommitMessage = expandCommitMessageBody || showCommitMessageBody
+    const commitContent =
+        truncateCommitMessage && node.body && node.body.length > 240 ? `${node.body.slice(0, 240)}...` : node.body
+    const truncationNeeded = commitContent && commitContent.length > 240
```
fkling, line 165: I'd pull out the `240` to be a constant defined at the top of the file (e.g. `TRUNCATED_COMMIT_MESSAGE_LENGTH`). If this number ever changes it only needs to be changed once.

I'd also consider pulling out the commit message into a separate variable sol that you don't have to deal with the "undefinedness" multiple times. Maybe something like this:

```
const commitMessage = node.body ?? ''
const truncationNeeded = commitMessage.length > TRUNCATED_COMMIT_MESSAGE_LENGTH
const truncatedCommitMessage = truncateCommitMessage && truncationNeeded 
  ? `${commitMesage.slice(0, TRUNCATED_COMMIT_MESSAGE_LENGTH)}...` 
  : commitMessage
```

And then use `truncatedCommitMessage` below.

fkling (approved):

PR #59769, opened 2024-01-23
sg/msp: fix monitoring config struct always being zero value

jac:
Adding response code ratio alerts to the service spec was not generating the cdktf. The code was always setting the monitoring spec to empty struct.

The fix is to only set to empty struct if it parsed as nil (was not provided)

Noticed when experimenting for https://github.com/sourcegraph/accounts.sourcegraph.com/issues/291
## Test plan

- CI
- locally-tested


bobheadxi (approved):
thanks 😂

PR #59772, opened 2024-01-23
gitserver: Simplify client

eseliger:
This PR consolidates and simplifies the client code a lot, in anticipation of moving more of the logic to the server-side. This allows us to have a better overview of what is actually used, and keeps the client interface as simple as possible.

Most notable changes:

- `Head` has been removed in favor of `GetDefaultBranch`. Currently, this is 1 gRPC -> 2 gRPC but with specialized endpoints, we should get it down to 1 again. They were otherwise the same, and I even noticed there's a slight bug in the `Head` implementation that didn't really handle the "empty repo" case. 
- `GetCommit` no longer exposes the option to also ensure the revision exists. 99% of call sites disabled this feature and I opted for an explicit call to EnsureRevision instead. This will keep the endpoint performant and not cause latency spikes.
- `ListBranches` no longer takes options. They were unused.
- The `BatchLog` endpoint has been deprecated.
- `CommitsExist` has been removed, it's been the only method that makes use of the batch log method and accounts for probably 1500 of the deleted lines in this PR. We're now making individual requests instead. *We should monitor this, it doesn't seem like we're making crazy request counts on dotcom, our largest instance. And we have much worse offenders, like the indexserver endpoint that runs rev-parse for 3M repos.*. I do believe that with the new API, we're able to do better generic batching implementations though that aren't this specific, that also make use of gRPC streaming instead of creating very large responses and latency. 
- `CommitDate` has been removed in favor of `GetCommit` which returns the date as well, one less endpoint.
- `GetCommits` has been unexported from the client API, it was only used internally.
- `Addrs` has been unexported.
- The unused properties `RefGlob` and `ExcludeRefGlob` have been removed from `RevisionSpecifier`, they were unused.

## Test plan

CI still passes, including integration tests.

## Preview 🤩
[Preview Link](https://docs.sourcegraph.com/@es/gitserver-cleanup-client)


cmd/gitserver/internal/batchlog.go
eseliger, line 1: just moved the existing logic into a single file so it'll be easier to rip out.

cmd/gitserver/internal/server.go
eseliger, line 1: removed contents have been moved to a new file

cmd/searcher/internal/search/store.go
```diff
@@ -431,16 +431,14 @@ func (s *Store) watchAndEvict() {
...
-		limit := 10 * len(s.GitserverClient.Addrs())
+		limit := 10 * len(conf.Get().ServiceConnections().GitServers)
 		if limit == 0 {
 			limit = 15
 		}
 		s.fetchLimiter.SetLimit(limit)
-
-		time.Sleep(10 * time.Second)
-	}
+	})
```
eseliger, line 441: Came here to use conf instead of gitserverClient.Addrs, which I wanted to remove from the API surface - it's not really something someone should assert on, as we might want to replace the logic to schedule repos to different shards later on. 

While I did that, I noticed that we could use conf.Watch here instead for more instantaneous propagation and less polling.

internal/codeintel/codenav/service.go
eseliger, line 1: reported as unused

internal/codeintel/codenav/utils.go
eseliger, line 1: reported unused

internal/gitserver/client.go
```diff
@@ -306,7 +293,7 @@ type Client interface {
 	ListRefs(ctx context.Context, repo api.RepoName) ([]gitdomain.Ref, error)
 
 	// ListBranches returns a list of all branches in the repository.
-	ListBranches(ctx context.Context, repo api.RepoName, opt BranchesOptions) ([]*gitdomain.Branch, error)
+	ListBranches(ctx context.Context, repo api.RepoName) ([]*gitdomain.Branch, error)
```
eseliger, line 296: the options were never used.

internal/codeintel/codenav/commit_cache.go
```diff
@@ -138,24 +139,24 @@ func (c *commitCache) commitsExist(ctx context.Context, commits []RepositoryComm
...
+	for i, rc := range repoCommits {
+		_, err := c.gitserverClient.GetCommit(ctx, rc.repoName, rc.commitID)
+		if err != nil {
+			if errors.HasType(err, &gitdomain.RevisionNotFoundError{}) {
+				exists[i] = false
+				continue
+			}
+			return nil, err
+		}
+		exists[i] = true
```
eseliger, line 159: Here, we're doing more RPC calls now than before.

We do thousands of those for zoekt repo state though, so I don't think this should be too bad.
Dotcom is our largest codenav instance, so should show any performance issues reliably.

pjlast: Do we know what graph/metric to watch for possible regressions here?

eseliger: This is the amount of requests to batchLog we're making today: https://sourcegraph.com/-/debug/grafana/d/gitserver/git-server?viewPanel=100110&orgId=1

And we track all requests including the origin of the request here: https://sourcegraph.com/-/debug/grafana/d/gitserver/git-server?viewPanel=100610&orgId=1 so we should see this thing go to zero for BatchLog, and go up for GetCommit instead.

pjlast (approved):

PR #59773, opened 2024-01-23
licensing: hide cody-related settings on SiteAdmin page when on code search-only license 

### LINKED ISSUE #59727: Remove search + search-related links from site-admin

BolajiOlajide:
Closes #59727

When an instance runs a Code Search-only license, we want to hide Cody-related settings on the Site Admin page to avoid confusion. The links in the Site Admin page shouldn't contain cody-related settings like:

* Embeddings

## Test plan

* Configure your local dev instance to run a Code Search-only license.
* The SiteAdmin page shouldn't contain any links to cody-related features

### DIFF
--- a/client/web/src/site-admin/routes.tsx
+++ b/client/web/src/site-admin/routes.tsx
@@ -10,7 +10,6 @@ import { OwnAnalyticsPage } from '../enterprise/own/admin-ui/OwnAnalyticsPage'
 import type { SiteAdminRolesPageProps } from '../enterprise/rbac/SiteAdminRolesPage'
 import type { RoleAssignmentModalProps } from '../enterprise/site-admin/UserManagement/components/RoleAssignmentModal'
 import { checkRequestAccessAllowed } from '../util/checkRequestAccessAllowed'
-import { isCodyOnlyLicense, isCodeSearchOnlyLicense } from '../util/license'
 
 import { isPackagesEnabled } from './flags'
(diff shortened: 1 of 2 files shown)


client/web/src/site-admin/routes.tsx
```diff
@@ -192,7 +193,10 @@ const CodyConfigurationPage = lazyComponent(
     'CodyConfigurationPage'
 )
 
-const codyIsEnabled = (): boolean => Boolean(window.context?.codyEnabled && window.context?.embeddingsEnabled)
+const disableCodeSearchFeatures = isCodyOnlyLicense()
+const disableCodyFeatures = isCodeSearchOnlyLicense()
+const codyIsEnabled = (): boolean =>
```
eseliger, line 198: why would we return `codyEnabled` for instances where we have a cody only license?

BolajiOlajide: This is mostly needed on a Code Search + Cody license, where Cody can be disabled via the site configuration.

`codyIsEnabled` here has to do with whether the site admin enables it via the Site Configuration while the `isCodeSearchOnlyLicense` and `isCodyOnlyLicense` have to do with whether the license allows for such features.

I added this here because we conditionally render some UI based on whether Cody is enabled in the Site Configuration, and I want the license features to supercede this check.

eseliger: hmm yeah, would a Code Search + Cody license have a different UI here though? Shouldn't it just be gone from site-admin in both cases?

PR #59779, opened 2024-01-23
Search: Add initial keyword search toggle

jtibshirani:
This change adds a simple toggle to enable/ disable keyword search, and removes
the prototype toggle called "Precise (NEW)". It also removes the smart search
toggle, since smart search doesn't pair well with the new keyword search
behavior. When the feature is disabled, we revert to the old UI and defaults
(smart search is still present, and we default to the 'standard' patterntype).
This helps mitigate risk for the initial release, since we can always disable
the change completely.

The toggle behaves like other toggles in the nav bar: the setting sticks unless
the user opens a new tab or refreshes the page. Then it reverts back,
defaulting to "keyword search".

Specific changes:
* Remove the `search-keyword` feature flag in favor of a new site setting
called `experimentalFeatures.keywordSearch`. I believe this is equally
convenient for admins, and it makes it simpler to do things like automatically
change the default search mode.
* When the feature is enabled, hide smart search and introduce a new toggle
called "Search language update" that switches between the 'keyword' and
'standard' pattern types

## Test plan

Adapted unit tests, and tested manually


jtibshirani:
Here's what the toggle looks like now. In a follow-up, we'll style it properly and a docs link.

Screenshot 2024-01-23 at 4 25 40 PM

jtibshirani:
Here's what the toggle looks like now. In a follow-up, we'll style it properly and a docs link.

Screenshot 2024-01-23 at 4 25 40 PM

I tested interactions of various toggles (enable keyword, then disable regex, etc.) and they felt natural to me.

The one weird thing I noticed: let's the keyword search feature is enabled, but you click on search query URL from when smart search was the default. It will contain the query parameter `sm=1`. This causes us to run the search in "smart search mode", which can may return additional results along with a banner saying that smart search was run (see below)

Screenshot 2024-01-23 at 4 33 22 PM
. But we have no smart search toggle anymore, so it's a bit strange.

client/web/src/util/settings.ts
```diff
@@ -50,6 +57,12 @@ export function defaultSearchModeFromSettings(settingsCascade: SettingsCascadeOr
  * configured by the user.
  */
 export function defaultPatternTypeFromSettings(settingsCascade: SettingsCascadeOrError): SearchPatternType | undefined {
+    // When the 'keyword search' language update is enabled, default to the 'keyword' patterntype
+    const features = getFromSettings(settingsCascade, 'experimentalFeatures') as SettingsExperimentalFeatures
```
jtibshirani, line 61: Moving from a feature flag to 'settings' allowed me to adjust the default pattern type and search mode.

client/web/src/search/results/components/search-results-info-bar/SearchResultsInfoBar.tsx
```diff
@@ -227,6 +231,18 @@ export const SearchResultsInfoBar: FC<SearchResultsInfoBarProps> = props => {
 
                 <div className={styles.expander} />
 
+                {props.showKeywordSearchToggle && (
+                    <Label className={styles.toggle}>
+                        Search language update{' '}
```
jtibshirani, line 236: I went with this copy for now, we can always tweak it when we polish the UI and docs.

schema/settings.schema.json
```diff
@@ -222,6 +222,11 @@
           "!go": {
             "pointer": true
           }
+        },
+        "keywordSearch": {
```
jtibshirani, line 226: Before the release, we'll update this to default to "true". The setting exists as an escape hatch, so admins can revert to the old UX if anything goes wrong.

Not sure if it's weird to list this under `experimentalFeatures`... it felt okay as this will go away in a future release.

stefanhengl (approved):
I played with it locally and this looks good to me. Left a minor comment.

rrhyne (approved):
Great start

schema/settings.schema.json
jtibshirani: Indeed, let's make sure to update that too when we default to `true`. (I don't think there's a change needed in this PR though, am I understanding your comment right?)

PR #59782, opened 2024-01-24
gitserver: Unify filereader API

eseliger:
This PR consolidates two methods that were effectively doing the same thing: `ReadFile` and `NewFileReader` - where `ReadFile` used the latter under the hood, and wrapped an `io.ReadAll`.
Given entire files are returned from this function that can be of arbitrary size, exposing a simple high-memory-use footgun made me uneasy here, and having to use io.ReadAll explicitly as the consumer hopefully makes the person using this API feel the same way.
Sometimes, we can currently not avoid it (GQL layer), but where possible, using the reader is very likely the better option. 

Since there would've been a slight change in how errors are reported, I also fixed that. Previously, `NewFileReader` would NOT return `os.ErrNotExist` errors, those would only be discovered when the returned reader is called.
Now, like `ReadFile`, and like `os.Open` we immediately return that error.

This comes at the cost of making a second gRPC request. This is not great, but we will migrate the FileReader API to a proper gRPC hopefully ASAP, which will let us optimize it better on the server-side.
Since Varun also asked for perf improvements here, this is probably one of the first things that would be easy to switch over to go-git, eliminating probably 50%+ of the overhead here.

## Test plan

Adjusted unit and integration tests. Things still work.


internal/gitserver/commands.go
```diff
@@ -1474,17 +1429,6 @@ func (br *blobReader) convertError(err error) error {
 	if strings.Contains(err.Error(), "exists on disk, but not in") || strings.Contains(err.Error(), "does not exist") {
 		return &os.PathError{Op: "open", Path: br.name, Err: os.ErrNotExist}
 	}
-	if strings.Contains(err.Error(), "fatal: bad object ") {
```
eseliger, line 1477: this is now handled by the blobOID call - so no need to do another roundtrip in the blobreader.

internal/gitserver/commands.go
```diff
@@ -1403,54 +1366,46 @@ func (c *clientImplementor) blobOID(ctx context.Context, repo api.RepoName, comm
...
-		//   - error: object $SHA is a tree, not a commit
-		//   - fatal: Invalid symmetric difference expression $SHA:$name
-		//   - outputting a diff instead of the file
-		//
-		// The last point is a security issue for repositories with sub-repo
-		// permissions since the diff will not be filtered.
-		blobOID, err := c.blobOID(ctx, repo, commit, name)
-		if err != nil {
-			return nil, err
+	blobOID, err := c.blobOID(ctx, repo, commit, name)
```
eseliger, line 1380: This is a slight regression in the simple case, we now make 2 gRPC calls here instead of just one. The upside is that it improves the error reporting so that we can expose an API that is similar to the API of `os.Open`.
We can fix that when we switch to the backend for this call. Also, simplicity (yay)!

internal/gitserver/commands_test.go
```diff
@@ -2421,41 +2421,6 @@ func TestRead(t *testing.T) {
 			}
 		}
 
-		t.Run(name+"-ReadFile", func(t *testing.T) {
```
eseliger, line 2424: these tests were basically redundant with the NewFileReader tests just below, no need to adjust them as this method is gone.

PR #59783, opened 2024-01-24
graphqlbackend: send webhook to SSC when upgrade to Cody Pro

unknwon:
## Test plan

Running https://github.com/sourcegraph/accounts.sourcegraph.com/pull/347 alongside locally.

```graphql
mutation ChangeCodyPlan {
  # User:1
  changeCodyPlan(user: "VXNlcjox", pro: true) {
    id
    codyProEnabled
    codyProEnabledAt
    __typename
  }
}
```

CleanShot 2024-01-23 at 17 16 19@2x


cmd/frontend/graphqlbackend/user.go
```diff
@@ -556,14 +558,68 @@ func (r *schemaResolver) ChangeCodyPlan(ctx context.Context, args *changeCodyPla
...
+			return
+		}
+
+		// TODO(ssc): send webhook to the right endpoint
+		req, err := http.NewRequest(http.MethodPost, fmt.Sprintf("%s/cody/api/rest/test-sams-token-auth", samsProvider.Issuer), nil)
+		if err != nil {
+			r.logger.Error("Error creating request", log.Error(err))
+			return
+		}
+		req.Header.Add("Authorization", "Bearer "+token.AccessToken)
```
bobheadxi, line 596: I'm pretty sure you can just get the HTTP client from `clientcredentials.Config` and use that directly, e.g. https://github.com/sourcegraph/accounts.sourcegraph.com/commit/e459febf95b6b20c38d8784db03003cbee023c06#diff-0ca42f72dd3de735c69be864375eee64bf7db07dcffe68d1f8f29308d64984bdR97-R99 , then you don't need to add `Authorization` by hand

This might not work very well with `httpcli.ExternalClient` - another alternative is to have first-class support for this in `httpcli.ExternalClient` with some wrapping thingamabob with `oauth2.Transport`, where you can set `Source` to the `TokenSource` implemented by `clientcredentials.Config`

chrsmith (commented):
Some nit picks and questions, but if this will work as we think it will then awesome!

❓ @unknwon , do you plan on cleaning this up and sending it out for a full preview? Would you like me to do that since? (As-is I'd suggest creating a dedicated `sendSSCWebhook(eventType)` function, etc.)

❓ So what, um, is this code doing exactly?

We are using OIDC libraries to create a new token. And then using that token as in the authorization header to a request to SSC.

So the principal here is the SAMS service AKA OIDC issuer? Right?

cmd/frontend/graphqlbackend/user.go
```diff
@@ -556,14 +558,68 @@ func (r *schemaResolver) ChangeCodyPlan(ctx context.Context, args *changeCodyPla
...
+
+		var samsProvider *schema.OpenIDConnectAuthProvider
+		for _, p := range conf.Get().AuthProviders {
+			if p.Openidconnect != nil && strings.HasPrefix(p.Openidconnect.ClientID, "sams_cid_") {
+				samsProvider = p.Openidconnect
+				break
+			}
+		}
+		if samsProvider == nil {
+			r.logger.Warn("No SAMS client found, skipped sending webhook to SSC")
```
bobheadxi, line 574: Presumably this might happen in dev, should we dump some metadata about the request we would have sent?

bobheadxi: Also let's put `trace.Logger` on the log entries here, so that we can follow a request to a webhook that it should have sent in case something breaks

chrsmith:
PSA: Will pull this PR locally, rebase, and do some cleanup to get it ready to be checked-in.

cmd/frontend/graphqlbackend/user.go
chrsmith: Personally, setting up the Authorization header seems more straightforward than using the `clientcredentials.Client(context.Context)` function. After stepping through the source, I still don't quite understand what it was doing. Similarly if we were to create an `*http.Client` and muck with the transport layers and what not.

I honestly don't know what the advantages to that are over, say, just using the stock `http.DefaultClient`. Is it so that the underlying `http` package machinery can reuse the connection or something?

chrsmith:
@unknwon , @bobheadxi I pushed a change to do a little refactoring. I'll make sure to fix build errors and address some of the feedback not covered already. But take a look. The main thing I'm not sure about is if this will actually work? (e.g. that the SAMS instance(s) are available in the dotcom configuration, and if specifying the "client.dotcom" scope will be supported by SAMS.)

cmd/frontend/graphqlbackend/user.go
chrsmith: > should we dump some metadata about the request we would have sent?

I think this would be a "If you like it then you shoulda put a ring on it" situation. If webhooks aren't being delivered at all, and that isn't fixed immediately, then it's pretty clear that we didn't care about their contents in the first place.

But having the `trace.Logger` makes sense in case we ever need to debug/troubleshoot connection woes or something.

chrsmith:
Meta point, before we spend any more time on this:

I do not believe we actually need this PR. (It's a nice forcing function to require thinking through how we will use SAMS access tokens for authorizing requests between different services.) But as far as the functionality being described, it isn't actually needed. (At least with the code that checked into the repo right now.)

A "dotcom->SSC" webhook would be required if we were to have SSC be the source of truth for all subscription data before 2/15. This is because after we migrate user data from dotcom to SAMS on 1/29, the metadata in SAMS would be out of date for any users who have opted to sign up for the Cody Pro free trial on Sourcegraph.com.

What we've done to address this without using a Webhooks, was to simply merge the "dotcom" and "SSC" information together. (So that if a user has opted to use Cody Pro on dotcom, but hasn't yet converted to paying for Cody Pro on the SSC backend, dotcom's GraphQL API will give the expected results.)

The crux of this is in the behavior of:
https://github.com/sourcegraph/sourcegraph/blob/572065bd1f7398ee28c53e6008bb6c434f0b67e4/internal/cody/subscription.go#L51-L54

I personally would greatly prefer that approach over relying on a dotcom->SSC Webhooks, as adding a Webhooks adds another moving part. And without a guarantee of reliable delivery, it just opens up the possibility of bugs. Also, the webhook would be at best a temporary solution.

Because after 2/15, we will no longer off the Cody Pro free trial from Sourcegraph.com. And can have the SSC backend be the source of truth for all Subscription information. (Without any caveats or exceptions.)

So if you would like to keep going down this road, I can work on adding the necessary API on the SAMS/SSC backend side to _receive_ these webhooks and do the book keeping there. But it's probably fine to just close this PR out.

However, we will need to put the logic for generating SAMS access tokens and sending requests to the SSC backend here:
https://github.com/sourcegraph/sourcegraph/blob/main/internal/ssc/ssc.go#L32

That is where (behind a feature flag) we have a SSC client for retrieving subscription information. It currently is using the "shared secret" auth approach. But when https://github.com/sourcegraph/accounts.sourcegraph.com/pull/347 gets merged, that API will no longer required the shared secret, and instead require a valid SAMS token. (At which time we can remove support for the SSC shared secret auth, and be glad it didn't need to be long enough to get wired into MSP/GSM, etc.)

Does that make sense?

TL;DR: We should probably close this out, but send out a new PR to move the few lines of code for generating a SAMS token into the `internal/ssc` package.

bobheadxi:
It's certainly an awkward situation. Ideally, 2/1 should have launched with SSC as the sole source of rate limits, and had this webhook - but since that's not the case (though IMO what ended up being done is more work than this + SSC as the sole rate limit source, but we're past that point, so that's fine 🙃), let's go as you proposed: keep aiming for the right state ready by 2/15, and launch with a happier architecture when we abandon the dotcom pro state.

I hope this was a productive learning experience regardless, thanks for bearing with us and hacking on this 😄

chrsmith:
After discussing this with folks, we'll not move forward with merging this PR.

However, I'll tease out the part where we update the `internal/ssc.Client` to use a SAMS access token to contact the SSC backend. (The demonstration of which was the genesis of this PR.)

cmd/frontend/graphqlbackend/user.go
bobheadxi: I'm pretty sure `clientcredentials.Client` handles deciding when to create a new token for you, ensuring that you never accidentally send an expired token, and helping you avoid excessive "create a new token" requests

bobheadxi: Never mind, I lied 😆 https://sourcegraph.com/github.com/golang/oauth2/-/blob/transport.go?L30-56 It's actually super simple, doesn't do much - in https://sourcegraph.com/github.com/golang/oauth2/-/blob/clientcredentials/clientcredentials.go?L92-124 it looks like it requests every time anyway

PR #59806, opened 2024-01-24
svelte: Fix search results rendering issues

fkling:
This commit fixes a couple of issues I noticed with search results rendering:

It was possible that we were rendering fewer plaintext lines than highlighted lines, which caused a jump as soon as the highlighted version was available. That's because we might show a truncated version of the chunk, but we'll always fetch the highlighting information for the full chunk. We need to account for that by truncating the highlighted lines too.

There have been several issues when search result components have been reused for different search results. Reuse can happen when navigating between cached results via the browser back/forward buttons. There are multiple ways to fix this. I choose to change the components themselves so that any state would be updated/recomputed when the result changes. In particular `FileContentSearchResult` will now refetch highlighting information when its input changes.
Using `{#await ...}` inside the loop doesn't seem ideal but I didn't notice any problems with it.

There have been some inconsistencies with capturing/restoring the scroll position of the search results list. I switched to using SvelteKit's native snapshot feature instead.

## Test plan

Manual testing


client/web-sveltekit/src/lib/search/CodeExcerpt.svelte
```diff
@@ -1,3 +1,5 @@
+<svelte:options immutable />
```
camdencheek, line 1: Oh, this is interesting. How does svelte normally determine whether an object was mutated? I always assumed that all equality checks were referential.

client/web-sveltekit/src/lib/search/CodeExcerpt.svelte
```diff
@@ -38,31 +40,32 @@
...
-            </tbody>
-        </table>
-    {:else}
-        <table use:highlightMatches={matches}>
-            <tbody>
-                {@html highlightedHTMLRows.join('')}
-            </tbody>
-        </table>
-    {/if}
+    {#key matches}
```
camdencheek, line 43: Why would `matches` ever change such that we need this to be keyed?

client/web-sveltekit/src/routes/search/+page.svelte
```diff
@@ -2,19 +2,34 @@
...
+            return {
+                searchResults: searchResults?.capture(),
+            }
+        },
+        restore(value) {
+            if (value) {
+                searchResults?.restore(value.searchResults)
+            }
+        },
+    }
```
camdencheek, line 22: This is so cool 😄

client/web-sveltekit/src/routes/search/RepoSearchResult.svelte
```diff
@@ -62,6 +63,11 @@
             </span>
         {/if}
     </div>
+    <svelte:fragment slot="info">
+        {#if result.repoStars}
+            <RepoStars repoStars={result.repoStars} />
+        {/if}
+    </svelte:fragment>
```
camdencheek, line 70: Nice little addition 🙂

client/web-sveltekit/src/routes/search/SearchResults.svelte
```diff
@@ -40,6 +40,16 @@
...
 
+    export function capture(): SearchResultsCapture {
+        return resultContainer?.scrollTop ?? 0
+    }
+
+    export function restore(capture?: SearchResultsCapture): void {
+        if (resultContainer) {
+            resultContainer.scrollTop = capture ?? 0
+        }
+    }
```
camdencheek, line 51: What's the difference between this style and exporting `snapshot` like we do in the `+page.svelte`?

client/web-sveltekit/src/routes/search/StreamingProgress.svelte
```diff
@@ -51,8 +51,8 @@
             {progress.matchCount}
             {pluralize('result', progress.matchCount)}
             {#if progress.repositoriesCount !== undefined}
-                from {progress.repositoriesCount} {pluralize('repository', progress.repositoriesCount, 'repositories')}
-            {/if}.
+                from {progress.repositoriesCount} {pluralize('repository', progress.repositoriesCount, 'repositories')}.
+            {/if}
```
camdencheek, line 55: was this period intentionally moved inside the if branch? Seems like it's still a sentence even without `repositoriesCount`

client/web-sveltekit/src/lib/search/CodeExcerpt.svelte
fkling: So whenever something sets the value of a prop that is an object it would assume that it has potentially changed. But we basically don't mutate data and especially when rendering something in a loop it seems to make sense to do as little work as necessary.

client/web-sveltekit/src/routes/search/SearchResults.svelte
fkling: The snapshot export only works for pages/layouts. This is my own convention for "normal" components to provide a unified interface for data that the component needs to preserve when going through the browser history.

client/web-sveltekit/src/routes/search/StreamingProgress.svelte
fkling: Yeah the period looked weird when the condition wasn't met and only `Found X results.` was rendered. But we could go either way.

camdencheek (approved):

client/web-sveltekit/src/lib/search/CodeExcerpt.svelte
camdencheek: Got it, thanks!

client/web-sveltekit/src/routes/search/StreamingProgress.svelte
camdencheek: I don't really care, just making sure it was intentional

client/web-sveltekit/src/routes/search/SymbolSearchResult.svelte
```diff
@@ -39,19 +35,21 @@
         {/if}
     </svelte:fragment>
     <svelte:fragment slot="body">
-        <div use:observeIntersection on:intersecting={onIntersection}>
+        <div use:observeIntersection on:intersecting={event => (visible = event.detail)}>
```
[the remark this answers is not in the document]
fkling: Yep, that's right. However, the result is cached in Apollo, which means Apollo will return the exact same object and due to immutability the CodeExcerpt component won't re-render. But even if it did it would be unnoticable. Another option would to not refetch when the value is already set but to also clear the value when `result` changes. But that's not necessarily obvious either. I'll leave comment about the caching.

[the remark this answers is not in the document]
camdencheek: That's putting a lot of trust in the Apollo client 😅

client/web-sveltekit/src/lib/search/CodeExcerpt.svelte
fkling: When we expand search results we might expand a partially rendered chunk match, which would include more matches. `highlightMatches` isn't written to handle updates and I don't believe it's called again when the inner content of the associated element changes. Additionally, since `highlightMatches` mutates the DOM it seems safer to recreate it when anything changes.

camdencheek: aaah, got it. Thanks!

client/web-sveltekit/src/routes/search/SymbolSearchResult.svelte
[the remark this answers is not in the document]
fkling: We should only use dependencies that we trust 😉

Issue #59623, opened 2024-01-16
On “Upgrade to Pro”, redirect from dotcom to SSC


vdavid:

chrsmith:
What is the URL we are planning on redirecting users to? We may want to have two different URLs depending on the context? (Or just change the UI based on the user account.)

As a proposal, why don't we just redirect users to:
https://accounts.sourcegraph.com/cody/subscription

And then when that page loads, we:

- If the user doesn't have a Cody Pro subscription, show them the "signup flow".
- If the user has an active or canceled Cody Pro subscription, show the "subscription management" page.

thenamankumar:
Makes perfect sense to me @chrsmith

Issue #59620, opened 2024-01-16
Dotcom: Expose Subscription details through GraphQL


vdavid:
Expose these pieces of info:
- “Is user on Cody Pro”
- “is subscription active/past_due”
- “when will it cancel”.

Why?
- For clients and `/cody/manage` to consume.

How?
- Just proxy the requests to SSC. On the SSC side, we already have these exposed—see [this](https://github.com/sourcegraph/accounts.sourcegraph.com/pull/259/files#diff-b01dfe047d5c62ac4cb93d7420beefb09bc65dbbc7ad22ab6f3a4155c4e19938R15).

chrsmith:
Adding a note with a few more requirements, @thenamankumar:

We need to do this in such a way that we don't require all dotcom / Cody Pro users be migrated or existing in the SSC/SAMS database. So the logic is slightly tricky.

If the user has signed up for the cody_pro trial, but dotcom doesn't have any SAMS info for the user, then we do _not_ need to fetch their subscription information from SSC. Instead, we just hard-code them being on a free trial that will end on 2/15. (And after 2/15, we say that their Cody Pro subscription has been canceled.)

For users that are on cody_pro AND have SAMS info, _then_ we should fetch their subscription information from SSC. This will also account for users that have entered their payment information on the SSC website prior to 2/15.

[This Slack Thread](https://sourcegraph.slack.com/archives/C05SZB829D0/p1705100026684459?thread_ts=1705085764.652129&cid=C05SZB829D0) captures the context here.

Issue #58123, opened 2023-11-06
Let Cody Gateway receive the usage limits from the Cody Management site


vdavid:

vdavid:
@thenamankumar and @chrsmith: I wonder how dotcom overrides will apply if we introduce a new actor source for SSC.
My idea is to create the new actor source in a way that it will actually make requests to both dotcom and SSC.

chrsmith:
> I wonder how dotcom overrides will apply if we introduce a new actor source for SSC.

They won't, full-stop.

The SSC backend is the source of truth for all things Cody usage limits. Any dotcom overrides will no longer be applicable.

If this means we need to take on a new requirement on the SSC-side to support setting per-user limit overrides or what not, then we can/should/will do that.

> My idea is to create the new actor source in a way that it will actually make requests to both dotcom and SSC.

We should definitely not do that. It just adds more latency or at the very least complexity. And instead, limit information should always come from the SSC backend.

Issue #59786, opened 2024-01-24
Make SAMS service id configurable for testing in cody/subscription


thenamankumar:
Make SAMS service ID configurable between accounts.sourcegraph.com and accounts.sgdev.org for testing in `getSAMSAccountIDForUser`.

thenamankumar:
We are closing this issue. We will rather add feature flagging support on SSC and let SSC prod switch between stripe prod and stripe test.

PR #59816, opened 2024-01-24
Expose new Cody Subscriptions API & integrate rate limits with SSC subscription data

thenamankumar:
Follow-up PR of https://github.com/sourcegraph/sourcegraph/pull/59763

closes: https://github.com/sourcegraph/sourcegraph/issues/58123
closes: https://github.com/sourcegraph/sourcegraph/issues/59622
closes: https://github.com/sourcegraph/sourcegraph/issues/59620
closes: https://github.com/sourcegraph/sourcegraph/issues/59623

The previous PR:
- added a new SSC API client in dotcom to fetch user's subscription info
- added a func `cody.SubscriptionForUser` to be used to find cody subscription info for the user. 
- added unit tests

This PR:
- updates all the graphql resolvers wherever applicable to use the `cody.SubscriptionForUser` function rather than directly checking dotcom's DB for `users.cody_pro_enabled_at` column.
- updated the rate limits logic in `dotcom_user_resolver` to use cody.SubscriptionForUser` function as well. 
- Added the `currentUser.codySubscription` field to graphql api
```graphql
currentUser { 
  codySubscription { 
    status 
    plan 
    applyProRateLimtis 
    currentPeriodStartAt 
    currentPeriodEndAt 
  }
}
```
- Removed the following fields from the graphql API
```graphql
currentUser { 
  codyCurrentPeriodStartDate
  codyCurrentPeriodEndDate
}
```
Both the removed fields are not used by any of the Cody Clients and their usage is updated to follow the new schema for dotcom web. 

## Test plan
Unit tests present

**Manual testing plan**

1. Set up and run SAMS locally.
2. Integrate SAMS as an openidconnect auth provider with dotcom locally.
3. Create a user using the SAMS auth provider.
4. Enable feature flag use-ssc-for-cody-subscription from site admin.
5. Make a graphql API [request](https://sourcegraph.test:3443/api/console#%7B%22query%22%3A%22%7B%5Cn%20%20currentUser%20%7B%5Cn%20%20%20%20codyProEnabled%5Cn%20%20%7D%5Cn%7D%5Cn%22%7D) to confirm false.
6. Visit localhost:9992/cody and go through the subscription flow to sign up for Cody Pro.
7. Make a graphql API [request](https://sourcegraph.test:3443/api/console#%7B%22query%22%3A%22%7B%5Cn%20%20currentUser%20%7B%5Cn%20%20%20%20codyProEnabled%5Cn%20%20%7D%5Cn%7D%5Cn%22%7D) again to confirm true.

### DIFF
--- a/client/web/src/cody/management/CodyManagementPage.tsx
+++ b/client/web/src/cody/management/CodyManagementPage.tsx
@@ -79,7 +79,7 @@ export const CodyManagementPage: React.FunctionComponent<CodyManagementPageProps
     const subscription = data?.currentUser?.codySubscription
 
     useEffect(() => {
-        if (enrollPro && data?.currentUser && subscription?.plan !== CodySubscriptionPlan.pro) {
+        if (enrollPro && data?.currentUser && subscription?.plan !== CodySubscriptionPlan.PRO) {
             changeCodyPlan({ variables: { pro: true, id: data?.currentUser?.id } })
         }
     }, [data?.currentUser, changeCodyPlan, enrollPro, subscription])
@@ -98,7 +98,7 @@ export const CodyManagementPage: React.FunctionComponent<CodyManagementPageProps
 
     const codeLimitReached = codyCurrentPeriodCodeUsage >= codyCurrentPeriodCodeLimit && codyCurrentPeriodCodeLimit > 0
     const chatLimitReached = codyCurrentPeriodChatUsage >= codyCurrentPeriodChatLimit && codyCurrentPeriodChatLimit > 0
-    const userIsOnProTier = subscription.plan === CodySubscriptionPlan.pro
+    const userIsOnProTier = subscription.plan === CodySubscriptionPlan.PRO
 
     const showUpgradeBanner = !userIsOnProTier && (codeLimitReached || chatLimitReached)
 
@@ -271,12 +271,12 @@ export const CodyManagementPage: React.FunctionComponent<CodyManagementPageProps
                                     <TrialPeriodIcon />
                                     <div className="mb-2 mt-4">
                                         <Text weight="bold" className={classNames('d-inline mb-0', styles.counter)}>
-                                            {subscription.status === CodySubscriptionStatus.pending
+                                            {subscription.status === CodySubscriptionStatus.PENDING
                                                 ? 'Free trial'
-                                                : subscription.status.toUpperCase()}
+                                                : subscription.status}
                                         </Text>
                                     </div>
-                                    {subscription.status === CodySubscriptionStatus.pending ? (
+                                    {subscription.status === CodySubscriptionStatus.PENDING ? (
                                         <Text className="text-muted mb-0" size="small">
                                             Until Feb 14, 2024
                                         </Text>
@@ -290,7 +290,7 @@ export const CodyManagementPage: React.FunctionComponent<CodyManagementPageProps
                                             weight="bold"
                                             className={classNames('d-inline mb-0 text-danger', styles.counter)}
                                         >
-                                            {subscription.status === CodySubscriptionStatus.pending
+                                            {subscription.status === CodySubscriptionStatus.PENDING
                                                 ? 'Free trial ended'
                                                 : subscription.status.toUpperCase()}
                                         </Text>
--- a/client/web/src/cody/management/CodyManagementPage.tsx
+++ b/client/web/src/cody/management/CodyManagementPage.tsx
@@ -24,6 +24,7 @@ import {
 import type { AuthenticatedUser } from '../../auth'
 import { Page } from '../../components/Page'
 import { PageTitle } from '../../components/PageTitle'
+import { CodySubscriptionStatus, CodySubscriptionPlan } from '../../graphql-operations'
 import type {
     ChangeCodyPlanResult,
     ChangeCodyPlanVariables,
@@ -76,7 +77,7 @@ export const CodyManagementPage: React.FunctionComponent<CodyManagementPageProps
     const enrollPro = parameters.get('pro') === 'true'
 
     useEffect(() => {
-        if (enrollPro && data?.currentUser && !data?.currentUser?.codyProEnabled) {
+        if (enrollPro && data?.currentUser && data?.currentUser?.codySubscription?.plan !== CodySubscriptionPlan.pro) {
             changeCodyPlan({ variables: { pro: true, id: data?.currentUser?.id } })
         }
     }, [data?.currentUser, changeCodyPlan, enrollPro])
@@ -93,12 +94,15 @@ export const CodyManagementPage: React.FunctionComponent<CodyManagementPageProps
         return null
     }
 
-    const { codyProEnabled } = data.currentUser
+    const subscription = data.currentUser.codySubscription
+    if (!subscription) {
+        return null
+    }
 
     const codeLimitReached = codyCurrentPeriodCodeUsage >= codyCurrentPeriodCodeLimit && codyCurrentPeriodCodeLimit > 0
     const chatLimitReached = codyCurrentPeriodChatUsage >= codyCurrentPeriodChatLimit && codyCurrentPeriodChatLimit > 0
 
-    const showUpgradeBanner = !codyProEnabled && (codeLimitReached || chatLimitReached)
+    const showUpgradeBanner = subscription.plan !== CodySubscriptionPlan.pro && (codeLimitReached || chatLimitReached)
 
     return (
         <>
@@ -140,10 +144,12 @@ export const CodyManagementPage: React.FunctionComponent<CodyManagementPageProps
                         <div>
                             <H2>My subscription</H2>
                             <Text className="text-muted mb-0">
-                                You are on the {codyProEnabled ? 'Pro' : 'Free'} tier.
+                                You are on the{' '}
+                                {subscription.plan === CodySubscriptionPlan.pro ? CodySubscriptionPlan.pro : 'Free'}{' '}
+                                tier.
                             </Text>
                         </div>
-                        {codyProEnabled ? (
+                        {subscription.plan === CodySubscriptionPlan.pro ? (
                             <div>
                                 <ButtonLink to="/cody/subscription" variant="secondary" outline={true} size="sm">
                                     Manage subscription
@@ -160,7 +166,7 @@ export const CodyManagementPage: React.FunctionComponent<CodyManagementPageProps
                     </div>
                     <div className={classNames('d-flex align-items-center mt-3', styles.responsiveContainer)}>
                         <div className="d-flex flex-column align-items-center flex-grow-1 p-3">
-                            {codyProEnabled ? (
+                            {subscription.plan === CodySubscriptionPlan.pro ? (
                                 <ProTierIcon />
                             ) : (
                                 <Text className={classNames(styles.planName, 'mb-0')}>Free</Text>
@@ -172,7 +178,7 @@ export const CodyManagementPage: React.FunctionComponent<CodyManagementPageProps
                         <div className="d-flex flex-column align-items-center flex-grow-1 p-3 border-left border-right">
                             <AutocompletesIcon />
                             <div className="mb-2 mt-3">
-                                {codyProEnabled ? (
+                                {subscription.applyProRateLimits ? (
                                     <Text weight="bold" className={classNames('d-inline mb-0', styles.counter)}>
                                         Unlimited
                                     </Text>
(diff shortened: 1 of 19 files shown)


client/web/src/cody/management/CodyManagementPage.tsx
```diff
@@ -93,12 +94,15 @@ export const CodyManagementPage: React.FunctionComponent<CodyManagementPageProps
...
+    const subscription = data.currentUser.codySubscription
+    if (!subscription) {
+        return null
+    }
 
     const codeLimitReached = codyCurrentPeriodCodeUsage >= codyCurrentPeriodCodeLimit && codyCurrentPeriodCodeLimit > 0
     const chatLimitReached = codyCurrentPeriodChatUsage >= codyCurrentPeriodChatLimit && codyCurrentPeriodChatLimit > 0
 
-    const showUpgradeBanner = !codyProEnabled && (codeLimitReached || chatLimitReached)
+    const showUpgradeBanner = subscription.plan !== CodySubscriptionPlan.pro && (codeLimitReached || chatLimitReached)
```
vdavid, line 105: We do the `subscription.plan !== CodySubscriptionPlan.pro` check at so many places. How about we introduce a variable `userHasProPlan` for this around line 78 and use that throughout the component?

client/web/src/cody/management/CodyManagementPage.tsx
```diff
@@ -252,30 +258,52 @@ export const CodyManagementPage: React.FunctionComponent<CodyManagementPageProps
...
-                                        Free trial
-                                    </Text>
+                        {subscription.plan === CodySubscriptionPlan.pro &&
+                            (subscription.applyProRateLimits ? (
+                                <div className="d-flex flex-column align-items-center flex-grow-1 p-3 border-left">
+                                    <TrialPeriodIcon />
+                                    <div className="mb-2 mt-4">
+                                        <Text weight="bold" className={classNames('d-inline mb-0', styles.counter)}>
+                                            {subscription.status === CodySubscriptionStatus.pending
+                                                ? 'Free trial'
```
vdavid, line 279: This is for removal after launch, right? If so, maybe mark it with a TODO and a date?

client/web/src/cody/subscription/CodySubscriptionPage.tsx
```diff
@@ -222,11 +230,17 @@ export const CodySubscriptionPage: React.FunctionComponent<CodySubscriptionPageP
                                                 { tier: 'pro' },
                                                 { tier: 'pro' }
                                             )
+                                            if (sscEnabled) {
+                                                window.location.href =
+                                                    'https://accounts.sourcegraph.com/cody/subscription'
```
vdavid, line 235: I suggest extracting this into a constant

vdavid (commented):
The changes make sense to me.

One thing we could discuss sync: What are your thoughts about performance on the completion hotpath? How long is the caching timeout on the dotcom side for SSC subscription data? When we refresh it, is it blocking (causing some completion requests to take longer)?

client/web/src/cody/management/CodyManagementPage.tsx
```diff
@@ -93,12 +94,15 @@ export const CodyManagementPage: React.FunctionComponent<CodyManagementPageProps
         return null
     }
 
-    const { codyProEnabled } = data.currentUser
+    const subscription = data.currentUser.codySubscription
+    if (!subscription) {
+        return null
```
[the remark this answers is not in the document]
thenamankumar: Yes it is a blank page, for now.

client/web/src/cody/management/CodyManagementPage.tsx
```diff
@@ -205,10 +211,10 @@ export const CodyManagementPage: React.FunctionComponent<CodyManagementPageProps
                             <H4 className={classNames('mb-0', codeLimitReached ? 'text-danger' : 'text-muted')}>
                                 Autocomplete suggestions
                             </H4>
-                            {!codyProEnabled &&
-                                (codeLimitReached && usageData?.currentUser?.codyCurrentPeriodEndDate ? (
+                            {!subscription.applyProRateLimits &&
```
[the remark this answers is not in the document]
thenamankumar: We apply pro limits only if the user has a valid subscription status. For cancelled or pending or other, we do not allow pro limits.

client/web/src/cody/management/CodyManagementPage.tsx
thenamankumar: Not sure, when we will do the removal. There is no mention of removing notifications regarding the same from Cody Clients. BUt I have an issue tracking removing of "pending" status and it should be covered under that.

client/web/src/cody/management/CodyManagementPage.tsx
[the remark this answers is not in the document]
chrsmith: Unrelated to this PR, here's how I think we should approach the topic of "null subscriptions".

Within the SSC backend, we should expect a given user does not have a subscription. (That is, they haven't converted and have subscription data in Stripe. And so there is no row in the `subscriptions` table, etc.)

However, outside of the SSC backend (e.g. here in dotcom, or even in the REST/GraphQL APIs of SSC) we should have the invariant that we will _always_ return a subscription for a user. (Even if there isn't any data in Stripe.) That way, we simplify clients so they don't need to know how to treat the "no subscription" case. And instead, just return enough data to clearly state what behavior a user should get with regard to rate limits, etc.

As a desired end-state, does that sound reasonable?

client/web/src/cody/subscription/CodySubscriptionPage.tsx
```diff
@@ -38,6 +40,7 @@ interface CodySubscriptionPageProps {
     isSourcegraphDotCom: boolean
     authenticatedUser?: AuthenticatedUser | null
 }
+const MANAGE_SUBSCRIPTION_REDIRECT_URL = 'https://accounts.sourcegraph.com/cody/subscription'
```
chrsmith, line 43: This is OK for this PR, but we'll need to revisit this in order to test the SSC backend in advance of the February launch. For example, something like:

```ts
// Returns the URL the user should navigate to in order to manage their Cody subscription.
// It is dependent upon which versions of SAMS the user is using for their identity.
function getCodySubscriptionManagementUrl(samsSessionId: string): string {
```

cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -6361,6 +6361,53 @@ type NewUsersConnection implements Connection {
...
+    trialing
+    pending
+    other
+}
+
+"""
+A plan for cody subscription.
+"""
+enum CodySubscriptionPlan {
+    free
```
chrsmith, line 6381: I don't know if this is 100% authoritative, but I've always seem enum values in GraphQL expressed in "SCREAMING_SNAKE_CASE".
https://www.apollographql.com/docs/technotes/TN0002-schema-naming-conventions/

cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -6361,6 +6361,53 @@ type NewUsersConnection implements Connection {
...
+    """
+    Whether pro rate limits are applied for the user.
+    """
+    applyProRateLimits: Boolean!
+    """
+    The current billing period start time.
+    """
+    currentPeriodStartAt: DateTime!
+    """
+    The current billing period end time.
```
chrsmith, line 6406: No seriously, we need to call this out EVERY GOSH DARNED OPPORTUNITY we get, because I 100% guarantee you that if we don't we can and will have bugs related to this.

```suggestion
    The current billing period end time. This may be in the past if the
    subscription is in a canceled state.
```

client/web/src/cody/management/CodyManagementPage.tsx
```diff
@@ -252,30 +254,52 @@ export const CodyManagementPage: React.FunctionComponent<CodyManagementPageProps
                             <H4 className={classNames('mb-0', chatLimitReached ? 'text-danger' : 'text-muted')}>
                                 Chat messages and commands
                             </H4>
-                            {!codyProEnabled &&
-                                (chatLimitReached && usageData?.currentUser?.codyCurrentPeriodEndDate ? (
+                            {!subscription.applyProRateLimits &&
+                                (chatLimitReached && subscription.currentPeriodEndAt ? (
                                     <Text className="text-danger mb-0" size="small">
-                                        Renews <Timestamp date={usageData?.currentUser?.codyCurrentPeriodEndDate} />
+                                        Renews <Timestamp date={subscription.currentPeriodEndAt} />
```
chrsmith, line 260: I don't quite understand this line. What if the subscription is in the canceled state? The subscription won't ever "renew". Or if the user is on the free tier.

Does "renew" in this case refer to when the usage's _usage_ limits will reset?

internal/cody/subscription.go
```diff
@@ -11,8 +11,11 @@ import (
 	"github.com/sourcegraph/sourcegraph/lib/errors"
 )
 
-const USE_SSC_FEATURE_FLAG = "use-ssc-for-cody-subscription"
-const CODY_PRO_TRIAL_ENDED = "cody-pro-trial-ended"
+// use-ssc-for-cody-subscription is a feature flag that enables the use of SSC as the source of truth for Cody subscription data.
+const USE_SSC_FOR_SUBSCRIPTION_FF = "use-ssc-for-cody-subscription"
+
+// cody-pro-trial-ended is a feature flag that indicates if the Cody Pro Trial has ended.
```
chrsmith, line 17: I'd suggest clarifying this a bit, since "Cody Pro Trial" leaves some room for ambiguity.

```suggestion
// cody-pro-trial-ended is a feature flag that indicates if the Cody Pro "Free Trial"  has ended.
// (Enabling users to use Cody Pro for free for 3-months starting in late Q4'2023.)
```

chrsmith (approved):
This seems reasonable, and the GraphQL schema looks right. But I think we need to do some more serious testing around the edge cases. (In particular how the UI will reflect subscription states such as canceled.)

Also, it isn't clear how we intend to test this with Sourcegraph dotcom using sgdev.org. It is critically important we have the ability to test the end-to-end experience on dotcom using subscription data fetched from the same-dev environment. (Since otherwise we are prohibitively constrained by the number of people capable and willing to run both SAMS and dotcom locally, and that is only compounded by the difficulty in setting up more sophisticated edge cases for Stripe-based subscriptions.)

client/web/src/cody/management/CodyManagementPage.tsx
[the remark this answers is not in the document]
vdavid: Sounds good. But for now, I think we should do better than displaying a blank page if the user has no subscription. We also return `null` if we're not on dotcom, but that's acceptable because they won't be links to this page on enterprise instances. But failing with a blank page if no subscription, that sounds like really bad UX.

client/web/src/cody/subscription/CodySubscriptionPage.tsx
vdavid: @thenamankumar, could you add a post-Feb issue to track this on the board?

client/web/src/cody/management/CodyManagementPage.tsx
thenamankumar: All such states will be fixed after proper consideration and design input. Will be collaborating with @danielmarquespt on that: https://sourcegraph.slack.com/archives/C05PC7AKFQV/p1706108698005529

client/web/src/cody/subscription/CodySubscriptionPage.tsx
thenamankumar: As mentioned in the previous PR, this will be handled separately and is already being tracked under issue: https://github.com/sourcegraph/sourcegraph/issues/59786

client/web/src/cody/management/CodyManagementPage.tsx
vdavid: @thenamankumar Please create a pre-Feb P1 issue to track this gap. We really don't want to say "Renews" for a subscription that will cancel at the end of the period. Saying this would confuse the hell out of users.

thenamankumar: Tracking here: https://github.com/sourcegraph/sourcegraph/issues/59823

client/web/src/cody/management/CodyManagementPage.tsx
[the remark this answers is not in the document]
thenamankumar: subscription is actually never null, unless there is a 5XX. We do send a subscription object even for free users.

PR #59834, opened 2024-01-24
Search Results: new styling

camdencheek:
This updates the styling of the search results to match the designs. As it turns out, there wasn't a ton to do other than styling and spacing tweaks that we needed to do anyways.

<table>
<th>
<td>Light</td>
<td>Dark</td>
</th>
<tr>
<td>Content</td>
<td>

CleanShot 2024-01-24 at 16 45 39@2x

</td>
<td>

CleanShot 2024-01-24 at 16 45 48@2x

</td>

</td>
</tr>
<tr>
<td>Repo</td>
<td>

CleanShot 2024-01-24 at 16 46 31@2x

</td>
<td>

CleanShot 2024-01-24 at 16 46 23@2x

</td>
</tr>
<tr>
<td>Path</td>
<td>

CleanShot 2024-01-24 at 16 46 54@2x

</td>
<td>

CleanShot 2024-01-24 at 16 47 00@2x

</td>
</tr>
<tr>
<td>Symbol</td>
<td>

CleanShot 2024-01-24 at 16 47 34@2x

</td>
<td>

CleanShot 2024-01-24 at 16 47 27@2x

</td>

</tr>
<tr>
<tr>
<td>Commit</td>
<td>

CleanShot 2024-01-24 at 16 48 27@2x

</td>
<td>

CleanShot 2024-01-24 at 16 48 33@2x

</td>
</tr>
<tr>
<td>Diff</td>
<td>

CleanShot 2024-01-24 at 16 49 07@2x

</td>
<td>

CleanShot 2024-01-24 at 16 49 02@2x

</td>
</tr>
</table>

A video of scroll behavior. Notice that the first header does not move when scrolled.

https://github.com/sourcegraph/sourcegraph/assets/12631702/feb598cc-6ba9-4a72-9602-bb19a4fadaa2

@taiyab The header color for dark mode looks nice, but does not contrast well with the border color:

CleanShot 2024-01-24 at 16 19 18@2x

## Test plan

Lots and lots of manual visual testing and tweaking. There should be no changes in behavior, only in style. I'll be doing an audit of the other places we display search results as well to make sure nothing surprising broke there before merge, but right now it's dinner time.

### DIFF
--- a/client/branded/src/search-ui/components/CopyPathAction.module.scss
+++ b/client/branded/src/search-ui/components/CopyPathAction.module.scss
@@ -1,5 +1,5 @@
 .button {
-    padding: 0.5rem;
+    padding: 0 0.5rem;
 }
 
 .copy-icon {
--- a/client/branded/src/search-ui/components/FilePathSearchResult.tsx
+++ b/client/branded/src/search-ui/components/FilePathSearchResult.tsx
@@ -65,6 +65,6 @@ export const FilePathSearchResult: FC<FilePathSearchResult & TelemetryProps> = (
             className={classNames(styles.copyButtonContainer, containerClassName)}
             repoLastFetched={result.repoLastFetched}
             actions={<SearchResultPreviewButton result={result} />}
-        ></ResultContainer>
+        />
     )
 }
--- a/client/branded/src/search-ui/components/FilePathSearchResult.tsx
+++ b/client/branded/src/search-ui/components/FilePathSearchResult.tsx
@@ -65,10 +65,6 @@ export const FilePathSearchResult: FC<FilePathSearchResult & TelemetryProps> = (
             className={classNames(styles.copyButtonContainer, containerClassName)}
             repoLastFetched={result.repoLastFetched}
             actions={<SearchResultPreviewButton result={result} />}
-        >
-            <div className={classNames(styles.searchResultMatch, 'p-2')}>
-                <small>{result.pathMatches ? 'Path match' : 'File contains matching content'}</small>
-            </div>
-        </ResultContainer>
+        ></ResultContainer>
     )
 }
--- a/client/branded/src/search-ui/components/ResultContainer.module.scss
+++ b/client/branded/src/search-ui/components/ResultContainer.module.scss
@@ -3,7 +3,10 @@
     display: flex;
     align-items: center;
     flex-wrap: wrap;
+
     position: sticky;
+    top: 0;
+
     z-index: 1; // Show on top of search result contents
 
     border-bottom: 1px solid var(--border-color);
--- a/client/branded/src/search-ui/components/ResultContainer.module.scss
+++ b/client/branded/src/search-ui/components/ResultContainer.module.scss
@@ -1,18 +1,18 @@
-.result-container {
+li.result-container {
     &:last-child {
         border-bottom-width: 1px;
     }
 
-    &:not(:last-of-type) {
-        // Prevents the sticky items below from affecting
-        // anything outside of the result container.
-        isolation: isolate;
-        margin-bottom: 1rem;
-    }
+    // &:not(:last-of-type) {
+    //     // Prevents the sticky items below from affecting
+    //     // anything outside of the result container.
+    //     isolation: isolate;
+    //     margin-bottom: 0.75rem;
+    // }
 }
 
 .header {
-    padding: 0.5rem 0.5rem 0.25rem 0;
+    padding: 0.5rem 0.75rem;
     display: flex;
     align-items: center;
     flex-wrap: wrap;
@@ -22,7 +22,15 @@
     // header and top of the scrolling block
     top: -1px;
     z-index: 1; // Show on top of search result contents
-    background-color: var(--body-bg);
+    border-bottom: 1px solid var(--border-color);
+
+    // TODO(taiyab): this needs to be fit into the design system
+    :global(.theme-light) & {
+        background-color: #f4f7f9;
+    }
+    :global(.theme-dark) & {
+        background-color: #090909;
+    }
 
     &-title {
         flex: 1 1 auto;
@@ -34,9 +42,9 @@
         margin-bottom: 0;
     }
 
-    &:not(:only-of-type) {
-        border-bottom: none;
-    }
+    // &:not(:only-of-type) {
+    //     border-bottom: none;
+    // }
 
     :global(.match-highlight) {
         color: var(--text-muted-highlighted);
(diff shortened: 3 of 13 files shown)


client/branded/src/search-ui/components/CopyPathAction.module.scss
```diff
@@ -1,5 +1,5 @@
 .button {
-    padding: 0.5rem;
+    padding: 0 0.5rem;
```
camdencheek, line 2: Changed the copy path button to only have horizontal padding so it is not forcing the padding for the whole header bar.

client/branded/src/search-ui/components/FilePathSearchResult.tsx
```diff
@@ -65,10 +65,6 @@ export const FilePathSearchResult: FC<FilePathSearchResult & TelemetryProps> = (
             className={classNames(styles.copyButtonContainer, containerClassName)}
             repoLastFetched={result.repoLastFetched}
             actions={<SearchResultPreviewButton result={result} />}
-        >
-            <div className={classNames(styles.searchResultMatch, 'p-2')}>
-                <small>{result.pathMatches ? 'Path match' : 'File contains matching content'}</small>
-            </div>
-        </ResultContainer>
```
camdencheek, line 72: Removed the currently-mostly-useless card for file matches

client/branded/src/search-ui/components/ResultContainer.module.scss
```diff
@@ -1,56 +1,37 @@
...
-        border-bottom-width: 1px;
-    }
-
-    &:not(:last-of-type) {
-        // Prevents the sticky items below from affecting
-        // anything outside of the result container.
-        isolation: isolate;
-        margin-bottom: 1rem;
-    }
-}
```
camdencheek, line 12: Moved the borders off the result container so they still make sense even when we do not have any contents for the results (like for path matches). 

I'm honestly not sure what the `isolate` was for, but AFAICT it isn't doing anything now.

client/branded/src/search-ui/components/ResultContainer.module.scss
```diff
@@ -1,56 +1,37 @@
...
+    padding: 0.5rem 0.75rem;
     display: flex;
     align-items: center;
     flex-wrap: wrap;
+
     position: sticky;
-    // With 0 value there is a rendering bug in Safari where this block
-    // doesn't fit tight enough and hence it's leaving a gap between sticky
-    // header and top of the scrolling block
-    top: -1px;
```
camdencheek, line 23: Tested on Safari and this is no longer needed. And it was causing the header to shift on scroll, which made it feel a tiny bit janky.

client/branded/src/search-ui/components/ResultContainer.module.scss
```diff
@@ -1,56 +1,37 @@
...
+
+    border-bottom: 1px solid var(--border-color);
+
+    // TODO(taiyab): this needs to be fit into the design system
+    :global(.theme-light) & {
+        background-color: #f4f7f9;
+    }
+    :global(.theme-dark) & {
+        background-color: #22283b;
+    }
```
camdencheek, line 20: There weren't good colors in our existing variables, so we hard-coded colors here. @taiyab is to follow up and update our color palette

client/branded/src/search-ui/components/ResultContainer.module.scss
```diff
@@ -1,56 +1,37 @@
...
 }
 
 .result {
-    border-radius: var(--border-radius);
-    border: 1px solid var(--border-color);
-    margin-top: 0.25rem;
-
-    &:hover {
-        border-color: var(--border-color-2);
-    }
```
camdencheek, line 53: @taiyab the border color change on hover no longer made sense to me since our results are much flatter now, so I removed it. Do we want search results to still react on hover somehow?

client/branded/src/search-ui/components/SymbolSearchResult.module.scss
```diff
@@ -7,22 +7,17 @@
     position: relative;
     display: flex;
     flex-direction: column;
-    background-color: var(--body-bg) !important;
```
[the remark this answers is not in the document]
BolajiOlajide: >>>>

client/branded/src/search-ui/results/StreamingSearchResultsList.tsx
```diff
@@ -267,7 +267,7 @@ export const StreamingSearchResultsList: React.FunctionComponent<
             <VirtualList<SearchMatch>
                 as="ol"
                 aria-label="Search results"
-                className={classNames('mt-2 mb-0', styles.list)}
+                className={classNames(styles.list)}
```
BolajiOlajide, line 270:
```suggestion
                className={styles.list}
```

We can safely remove the `classNames` function call since we aren't merging styles anymore.

BolajiOlajide (approved):

client/branded/src/search-ui/results/StreamingSearchResultsList.tsx
camdencheek: Thanks!

client/branded/src/search-ui/results/StreamingSearchResultsList.tsx
```diff
@@ -267,7 +263,7 @@ export const StreamingSearchResultsList: React.FunctionComponent<
             <VirtualList<SearchMatch>
                 as="ol"
                 aria-label="Search results"
-                className={classNames('mt-2 mb-0', styles.list)}
+                className={styles.list}
```
vovakulikov, line 266: FIY this is what in my opinion we should do, ideally we should not have inline classes, even small spacing classes aren't good, thanks!

Why? Because it multiple sources of styles hence 
- It might be hard to debug since there is no good way to override inline classes in the browsers freely 
- Even this set of spacings makes spacing inconsistent enough (some use 0.25rem some use just spaces instead)
 

Ideally we should rely on properties like `gap` in the layout.

vovakulikov (approved):
First of all thanks @camdencheek, it looks like a small changes in code but it feels very good, and I think it matches with other panels better too (like file preview and filters panel) 

Not a blocker but one thing that I've noticed is that we probably want to simplify repository card when we have just private badge without description, so this card won't take two rows cc @taiyab 

Screenshot 2024-01-24 at 21 53 52

client/branded/src/search-ui/components/ResultContainer.module.scss
```diff
@@ -13,7 +13,7 @@
 
     // TODO(taiyab): this needs to be fit into the design system
     :global(.theme-light) & {
-        background-color: #f4f7f9;
+        background-color: #f5f8fa;
```
camdencheek, line 16: FYI @taiyab, CI was complaining because the background color was _just barely_ below the contrast threshold, so I bumped this just a tiny bit lighter.

Issue #59682, opened 2024-01-17
Keyword search respects codyignore


cbart:
Ensure the `getCodyContext` respects cody ignore.

PR #59836, opened 2024-01-24
Add .cody/ignore support to remote context 

chwarwick:
Updates the getCodyContext resolver to filter out context if a `.cody/ignore` file is present and it matches an ignore rule. The behavior is behind an experimental feature in the site config. 

closes https://github.com/sourcegraph/sourcegraph/issues/59682
closes https://github.com/sourcegraph/sourcegraph/issues/59976

## Test plan
updated existing context resolver tests
add new tests for filtering logic

Manual testing 
 - Created a test repo with a main.go file and printed hello world
 - Asked Cody what it printed
Screenshot 2024-01-25 at 10 01 29 AM

- Added a .cody/ignore file with main.go in it
- waiting for repo sync
- Asked Cody what the program prints - it did not read the file anymore.
Screenshot 2024-01-25 at 9 56 29 AM

### DIFF
--- a/internal/codycontext/context.go
+++ b/internal/codycontext/context.go
@@ -20,6 +20,7 @@ import (
 	vdb "github.com/sourcegraph/sourcegraph/internal/embeddings/db"
 	"github.com/sourcegraph/sourcegraph/internal/embeddings/embed"
 	"github.com/sourcegraph/sourcegraph/internal/featureflag"
+	"github.com/sourcegraph/sourcegraph/internal/gitserver"
 	"github.com/sourcegraph/sourcegraph/internal/metrics"
 	"github.com/sourcegraph/sourcegraph/internal/observation"
 	"github.com/sourcegraph/sourcegraph/internal/search"
@@ -41,7 +42,7 @@ type FileChunkContext struct {
 	EndLine   int
 }
 
-func NewCodyContextClient(obsCtx *observation.Context, db database.DB, embeddingsClient embeddings.Client, searchClient client.SearchClient, getQdrantSearcher func() (vdb.VectorSearcher, error)) *CodyContextClient {
+func NewCodyContextClient(obsCtx *observation.Context, db database.DB, embeddingsClient embeddings.Client, searchClient client.SearchClient, getQdrantSearcher func() (vdb.VectorSearcher, error), gitClient gitserver.Client) *CodyContextClient {
 	redMetrics := metrics.NewREDMetrics(
 		obsCtx.Registerer,
 		"codycontext_client",
@@ -64,6 +65,7 @@ func NewCodyContextClient(obsCtx *observation.Context, db database.DB, embedding
 		embeddingsClient:  embeddingsClient,
 		searchClient:      searchClient,
 		getQdrantSearcher: getQdrantSearcher,
+		gitClient:         gitClient,
 
 		obsCtx:                 obsCtx,
 		getCodyContextOp:       op("getCodyContext"),
@@ -76,6 +78,7 @@ type CodyContextClient struct {
 	db                database.DB
 	embeddingsClient  embeddings.Client
 	searchClient      client.SearchClient
+	gitClient         gitserver.Client
 	getQdrantSearcher func() (vdb.VectorSearcher, error)
 
 	obsCtx                 *observation.Context
@@ -125,6 +128,11 @@ func (c *CodyContextClient) GetCodyContext(ctx context.Context, args GetContextA
 		return nil, err
 	}
 
+	contextFilter, err := NewCodyIgnoreFilter(ctx, c.gitClient, append([]types.RepoIDName{}, args.Repos...))
+	if err != nil {
+		return nil, err
+	}
+
 	// NOTE: We use a pretty simple heuristic for combining results from
 	// embeddings and keyword search. We use the ratio of repos with embeddings
 	// to decide how many results out of our limit should be reserved for
@@ -151,11 +159,11 @@ func (c *CodyContextClient) GetCodyContext(ctx context.Context, args GetContextA
 	// Fetch keyword results and embeddings results concurrently
 	p := pool.New().WithErrors()
 	p.Go(func() (err error) {
-		embeddingsResults, err = c.getEmbeddingsContext(ctx, embeddingsArgs)
+		embeddingsResults, err = c.getEmbeddingsContext(ctx, embeddingsArgs, contextFilter)
 		return err
 	})
 	p.Go(func() (err error) {
-		keywordResults, err = c.getKeywordContext(ctx, keywordArgs)
+		keywordResults, err = c.getKeywordContext(ctx, keywordArgs, contextFilter)
 		return err
 	})
 
@@ -188,7 +196,7 @@ func (c *CodyContextClient) partitionRepos(ctx context.Context, input []types.Re
 	return embedded, notEmbedded, nil
 }
 
-func (c *CodyContextClient) getEmbeddingsContext(ctx context.Context, args GetContextArgs) (_ []FileChunkContext, err error) {
+func (c *CodyContextClient) getEmbeddingsContext(ctx context.Context, args GetContextArgs, filter RepoContextFilter) (_ []FileChunkContext, err error) {
 	ctx, _, endObservation := c.getEmbeddingsContextOp.With(ctx, &err, observation.Args{Attrs: args.Attrs()})
 	defer endObservation(1, observation.Args{})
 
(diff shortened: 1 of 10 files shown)


internal/codycontext/filter.go
```diff
@@ -0,0 +1,73 @@
...
+		if !found {
+			continue
+		}
+		ignoreFile, err := client.ReadFile(ctx, repo.Name, api.CommitID(head), codyIgnoreFile)
+		if err != nil {
+			// We do not ignore anything if the ignore file does not exist.
+			if strings.Contains(err.Error(), "file does not exist") {
+				continue
+			}
+			return nil, err
```
chwarwick, line 49: I'm undecided on this case and the error parsing the ignore file if I should allow it to continue but return a function that blocks all context from the repo.

internal/codycontext/filter.go
```diff
@@ -0,0 +1,72 @@
+package context
+
+import (
+	"bytes"
+	"context"
+	"strings"
+
+	"github.com/sourcegraph/zoekt/ignore"
```
chwarwick, line 8: Is it wrong to use this from zoekt?

internal/codycontext/filter.go
```diff
@@ -0,0 +1,88 @@
...
+	"os"
+
+	"github.com/sourcegraph/zoekt/ignore"
+
+	"github.com/sourcegraph/sourcegraph/internal/api"
+	"github.com/sourcegraph/sourcegraph/internal/gitserver"
+	"github.com/sourcegraph/sourcegraph/internal/types"
+)
+
+const codyIgnoreFile = ".cody/ignore"
```
chwarwick, line 16: @abeatrix  this is correct for the ignore file right?

abeatrix: Yea this is the name we use for the .cody/ignore file

abeatrix: @chwarwick would this work with nested directory?

chwarwick: > @chwarwick would this work with nested directory?

I don't think that is an issue since I'm checking directly against the repo from gitserver and I thought it had to be at the root of the repo.  Is there a scenario are you thinking of?  I shouldn't need to handle any of the complexities that the extensions have were multiple repos can be open etc.. 

The api is "get me context from repos [A, B, C]" so it runs a search against those repos and checks for ignore files in each of them to apply to the search results.

internal/codycontext/filter.go
```diff
@@ -0,0 +1,88 @@
...
+		filters: make(map[types.RepoIDName]filterFunc),
+	}
+	for _, repo := range repos {
+		_, commit, err := client.GetDefaultBranch(ctx, repo.Name, true)
+		if err != nil {
+			return nil, err
+		}
+		// this is an empty repo, there won't be anything to filter
+		if commit == "" {
+			continue
```
slimsag, line 41: I'm not 100% convinced this is true, it seems like an edge case that could potentially bite us. Are you completely sure this cannot ever be the case with a Git repo somehow? How about just removing the continue?

internal/codycontext/filter.go
slimsag: If there is an error parsing/reading the ignore file, we should definitely block everything and scream in the server logs whatever the error is (so admins can find it)

If you're worried there could be a bug that blocks context for customers/users that do not care about context blocking at all, then we could put it behind a site config flag - so that e.g. if context blocking is enable then it's very aggressive about blocking context, and otherwise relaxed (but still respects codyignore)?

internal/codycontext/filter.go
```diff
@@ -0,0 +1,88 @@
...
+// content based on the .cody/ignore file at the head of the default branch
+// for the given repositories. If no .cody/ignore file exists, no filtering is done.
+func NewCodyIgnoreFilter(ctx context.Context, client gitserver.Client, repos []types.RepoIDName) (RepoContentFilter, error) {
+	f := &repoFilter{
+		filters: make(map[types.RepoIDName]filterFunc),
+	}
+	for _, repo := range repos {
+		_, commit, err := client.GetDefaultBranch(ctx, repo.Name, true)
+		if err != nil {
+			return nil, err
```
slimsag, line 37: I would treat this case the same way you treat an error from getIgnoreFileBytes, or parsing the ignore file. e.g. if gitserver has an intermittent failure and doesn't respond well to this query, how do we want it to behave.

slimsag (approved):
Generally looks good

internal/codycontext/filter.go
abeatrix: >  thought it had to be at the root of the repo

I was wondering how it'd work with monorepos that contain sub repo but I think we should be good as long as we communicate that with the users?

abeatrix (approved):
Lgtm! Can't wait to test it out in the clients!

arafatkatze (commented):
Love the process and the idea of this PR. I wanna approve after the resolution of 
1. [Caching](https://github.com/sourcegraph/sourcegraph/pull/59836/files#r1467134734)
2. [Edge Case of git](https://github.com/sourcegraph/sourcegraph/pull/59836/files#r1467296447)
3. [Blocking on errors](https://github.com/sourcegraph/sourcegraph/pull/59836/files#r1467295375)

internal/codycontext/filter.go
chwarwick: I am going to move it behind an experimental flag like it is in the extensions.  I do think I need to fail the whole context call, if I just trying to remove all results from 1 repo it could be the only repo and when we have a lot of long running searches that end up providing nothing.

internal/codycontext/context.go
```diff
@@ -125,6 +128,11 @@ func (c *CodyContextClient) GetCodyContext(ctx context.Context, args GetContextA
 		return nil, err
 	}
 
+	contextFilter, err := NewCodyIgnoreFilter(ctx, c.gitClient, append([]types.RepoIDName{}, args.Repos...))
```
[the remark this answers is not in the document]
chwarwick: Thank you! yes I'll do that now.

internal/codycontext/context.go
```diff
@@ -302,7 +314,7 @@ func (c *CodyContextClient) getKeywordContext(ctx context.Context, args GetConte
 
 			for _, res := range e.Results {
 				if fm, ok := res.(*result.FileMatch); ok {
-					collected = append(collected, fileMatchToContextMatches(fm)...)
+					collected = append(collected, filter.Filter(fileMatchToContextMatches(fm))...)
```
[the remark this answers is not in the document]
chwarwick: @jtibshirani I went through a few iterations here and I am not fully happy with any.  At first I let the searches run and then filtered the results at the very end, then realized that it wouldn't provide the # of results requested because the searches would get cut off, so I moved the filtering logic here.  I agree that this could cause the search to run a lot longer and still may never get as any results as requested.

I'm interested in your idea about passing it as part of the request,  there are few potential issues I can think of:

1. The ignore file is globs and the -file:  is regex.  I assume there is a reliable way to convert but I didn't look yet.
2. I think it complicates building the search.  Right now it is one repo: filter with a regex or of all the repos, I think to pass the filters we would need to do something like:
 ```
(repo:^repoA$ AND -file:secret.go -file.... ) OR (repo:^repoB$ AND -file:^private/) ...
```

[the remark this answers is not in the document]
Comment (jtibshirani) [reply internal/codycontext/context.go:317]: We discussed offline, here are our conclusions:
* The current behavior seems fine for a first version. We already have this downside with sub-repo permissions.
* Eventually we should consider pushing the filtering logic into the search backends, as we do for `.sourcegraph-ignore`. We can do this in a follow-up though, nothing in this PR prevents us from doing that.

File: internal/codycontext/filter.go (hunk 4 above)
Comment (chwarwick) [reply internal/codycontext/filter.go:98]: I refactored this logic, if we hit any of these scenarios where it can't determine if there is an ignore file, then we exclude that repo from the request.  The signature for GetDefaultBranch says empty values without an error mean it's either empty or cloning so this route is safer since potential it could be cloning but then finish prior to searching.

File: internal/codycontext/filter.go [hunk 8]
```diff
@@ -0,0 +1,187 @@
...
+func (f *repoFilter) getFilter(repos []types.RepoIDName) ([]types.RepoIDName, FileChunkFilterFunc) {
+	filters := make(map[api.RepoName]filterFunc, len(repos))
+	filterableRepos := make([]types.RepoIDName, 0, len(repos))
+	// use the internal actor to ensure access to repo and ignore files
+	ctx := actor.WithInternalActor(context.Background())
+	for _, repo := range repos {
+
+		_, commit, err := f.client.GetDefaultBranch(ctx, repo.Name, true)
+		if err != nil {
+			continue
```
slimsag, line 93: I suggest still logging all of these errors. If e.g. important context ends up getting filtered out, and we need to debug, how would we identify the error reason it was excluded?

slimsag (approved):

PR #59838, opened 2024-01-24
Control SAMS origin via feature flag for testing

thenamankumar:
closes: https://github.com/sourcegraph/sourcegraph/issues/59786

This PR enables site admins to control if cody subscription flow for a particular user will utilise `accounts.sourcegraph.com` or `accounts.sgdev.org` using a feature flag. 

This will enable us to test the complete SSC flow, even before SAMS is fully deployed and migrated to. 

loom: https://www.loom.com/share/efd9c3dbda1249288627234d7092ba86

## Test plan
Unit testes added
Follow the loom demo for manual testing: https://www.loom.com/share/efd9c3dbda1249288627234d7092ba86

### DIFF
--- a/client/web/src/cody/subscription/CodySubscriptionPage.tsx
+++ b/client/web/src/cody/subscription/CodySubscriptionPage.tsx
@@ -40,7 +40,16 @@ interface CodySubscriptionPageProps {
     isSourcegraphDotCom: boolean
     authenticatedUser?: AuthenticatedUser | null
 }
-const MANAGE_SUBSCRIPTION_REDIRECT_URL = 'https://accounts.sourcegraph.com/cody/subscription'
+
+const useManageSubscriptionRedirectURL = (pro: boolean): string => {
+    const [useSAMSTestInstance] = useFeatureFlag('use-sams-test-instance', false)
+
+    if (useSAMSTestInstance) {
+        return 'https://accounts.sgdev.org/cody/subscription' + (pro ? '?pro=true' : '')
+    }
+
+    return 'https://accounts.sourcegraph.com/cody/subscription' + (pro ? '?pro=true' : '')
+}
 
 export const CodySubscriptionPage: React.FunctionComponent<CodySubscriptionPageProps> = ({
     isSourcegraphDotCom,
@@ -51,7 +60,6 @@ export const CodySubscriptionPage: React.FunctionComponent<CodySubscriptionPageP
     const utm_source = parameters.get('utm_source')
 
     const [sscEnabled] = useFeatureFlag('use-ssc-for-cody-subscription', false)
-
     useEffect(() => {
         eventLogger.log(EventName.CODY_SUBSCRIPTION_PAGE_VIEWED, { utm_source }, { utm_source })
     }, [utm_source])
@@ -60,6 +68,10 @@ export const CodySubscriptionPage: React.FunctionComponent<CodySubscriptionPageP
 
     const [showUpgradeToPro, setShowUpgradeToPro] = useState<boolean>(false)
     const [showCancelPro, setShowCancelPro] = useState<boolean>(false)
+
+    const manageSubscriptionRedirestURL = useManageSubscriptionRedirectURL(
+        data?.currentUser?.codySubscription?.plan === CodySubscriptionPlan.PRO
+    )
 
     const navigate = useNavigate()
 
@@ -210,7 +222,7 @@ export const CodySubscriptionPage: React.FunctionComponent<CodySubscriptionPageP
                                                     }
                                                 )
                                                 if (sscEnabled) {
-                                                    window.location.href = MANAGE_SUBSCRIPTION_REDIRECT_URL
+                                                    window.location.href = manageSubscriptionRedirestURL
                                                     return
                                                 }
 
@@ -231,7 +243,7 @@ export const CodySubscriptionPage: React.FunctionComponent<CodySubscriptionPageP
                                                 { tier: 'pro' }
                                             )
                                             if (sscEnabled) {
-                                                window.location.href = MANAGE_SUBSCRIPTION_REDIRECT_URL
+                                                window.location.href = manageSubscriptionRedirestURL
                                                 return
                                             }
 
--- a/internal/cody/subscription.go
+++ b/internal/cody/subscription.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"time"
 
+	"github.com/sourcegraph/sourcegraph/internal/conf"
 	"github.com/sourcegraph/sourcegraph/internal/database"
 	"github.com/sourcegraph/sourcegraph/internal/featureflag"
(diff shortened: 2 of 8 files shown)


thenamankumar:
Added issue to track the clean up of testing specific logic, once it is no longer required: https://github.com/sourcegraph/sourcegraph/issues/59839

client/web/src/cody/subscription/CodySubscriptionPage.tsx
```diff
@@ -40,7 +40,16 @@ interface CodySubscriptionPageProps {
     isSourcegraphDotCom: boolean
     authenticatedUser?: AuthenticatedUser | null
 }
-const MANAGE_SUBSCRIPTION_REDIRECT_URL = 'https://accounts.sourcegraph.com/cody/subscription'
+
+const useManageSubscriptionRedirectURL = (pro: boolean): string => {
```
chrsmith, line 44: Nit: TypeScript naming conventions is to only capitalize the first letter of an acronym or initialism. (Unlike Golang.) While I personally kinda like this more, it definitely looks out of place and would be inconsistent with the rest of the codebase.

chrsmith: As an unimportant styleistic question, is there a reason why you are creating a "value with a function type" via the lambda (`(pro: boolean): string => {`) as opposed to just creating a function (`function useManageSubscriptionRedirectUrl(pro: string): string {`)?

I don't know if there is any performance or efficiency difference between the two. But personally the `function` approach looks better to me. But I might be in the minority on this. So I'm just wondering if this was intentional on your part, and if this is the "more standard" approach.

chrsmith: ⚠️ Could you please add a doc comment explaining what `pro: boolean` is used for? It isn't immediately obvious, and would probably be confusing to the uninitiated. I assume the following would be correct, right?

```ts
// Returns a URL the user can be directed to in order to manage their Cody Subscription.
// The `pro` parameter should be true if the user has already upgraded to
// use Cody Pro on Sourcegraph.com.
```

client/web/src/cody/subscription/CodySubscriptionPage.tsx
```diff
@@ -61,6 +69,10 @@ export const CodySubscriptionPage: React.FunctionComponent<CodySubscriptionPageP
     const [showUpgradeToPro, setShowUpgradeToPro] = useState<boolean>(false)
     const [showCancelPro, setShowCancelPro] = useState<boolean>(false)
 
+    const manageSubscriptionRedirestURL = useManageSubscriptionRedirectURL(
```
chrsmith, line 72: Nit: I'd suggest we use a different variable name here, as `manageSubscriptionRedirestURL` and `useManageSubscriptionRedirectURL` are awfully similar. (Also I there is a typo, "Retire[s/c]tURL".)

For readability, perhaps:

```ts
const manageSubUrl = useManageSubscriptionRedirectUrl(...)
```

Also, there isn't any redirection going on, right? It's just a regular link? So perhaps we can drop that part from the name?

internal/cody/subscription.go
```diff
@@ -100,13 +100,36 @@ func consolidateSubscriptionDetails(ctx context.Context, user types.User, subscr
 	}, nil
 }
 
+func getSAMSServiceID(ctx context.Context) (string, error) {
+	sgconf := conf.Get().SiteConfig()
+
+	if featureflag.FromContext(ctx).GetBoolOr(ssc.USE_SAMS_TEST_INSTANCE_FF, false) {
```
chrsmith, line 106: ⚠️ Will this actually work? We need the behavior to be the following:

1. Get every SAMS identity associated with the dotcom user account. (The service type is 'openidconnect' and the client ID contains "sams".)
- If the user account has only one SAMS identity associated with it, then use that identity provider's service ID. (For 99% of users, this will be the SAMS production instance.)
- If the user account has 2 SAMS identities associated with it then default to using the SAMS-dev instance.
- We honor a per-user feature flag, allowing someone to override this behavior. (e.g. you are done testing SSC-dev and need to verify something in production.)

I speced out this behavior in a doc I wrote a couple hours ago. (Sorry for not being having done that sooner!)
https://docs.google.com/document/d/1IIdeJqQ_Xwkgw8m0dydCYJTRMByowoymFXzth3fqRoI/edit#bookmark=id.do0bgngt01uz

chrsmith: Having a per-user flag to have them always use the SAMS-dev instance is certainly simpler than what I am proposing. But I think going with the more nuanced approach is required.

Only Sourcegraph.com admins can control setting/modifying feature flags, right? This isn't something that a user can do for themselves? If so, then requiring a Sourcegraph.com admin to "opt you in" to using the SSC-dev environment is a prohibitive barrier for testing our service.

Worth pointing out that **~everyone at the company** will need to do this in order to get access to Cody Pro starting in February. (We could offer Sourcegraph employees a discount code or something, but doing so would skew metrics enough that it isn't worth it.)

So that's one of the reasons why I'm more eager to use this other approach. Also, if you set the "use SAMS/SSC-dev" flag, and have not yet signed in, then the lookup to get the user's SAMS account ID would fail. (With I assume a really poor UX in tracking down the source of the failure and how to fix it.)

thenamankumar: I went through the doc, and I have some security concerns with the idea mentioned there. I have left a comment in the doc: https://docs.google.com/document/d/1IIdeJqQ_Xwkgw8m0dydCYJTRMByowoymFXzth3fqRoI/edit?disco=AAABD9U75ME

Lets discuss it before me move ahead with the implementation. 

Here is the same comment:

```
🚨 How are we controlling the access to who can connect their account to sgdev.org? Having it hidden under a query param for open source software isn't secure enough. 

Given that the mentioned idea is to automatically give precedence to sgdev.org connection by default (🚨), it makes it more serious because that will allow users to use Cody Pro for free going through the Stripe test env flow. 

The logic below lets people test on prod without any control from our end. It is suggested to set a feature flag to get out of the test environment, which IMHO is the opposite of how it should be.

Even if we decide to go ahead with this particular logic mentioned in the doc, when exactly are we planning to remove it?
```

chrsmith: To recap the discussion from the doc:

We'll simplify things to just having this feature flag "opt-in" to using SAMS-dev on a user-by-user basis. (But otherwise always use SAMS-prod.)

This is content on our ability to apply a single "flag override" for all users in an organization (e.g. opt-in the entire sourcegraph team) AND we can provide a user-level flag to _undo_ that (e.g. opt-out someone who needs to use the SAMS-prod instance).

chrsmith (changes requested):
Let's close this out.

I'm working on a PR now to update the `ssc.Client` to send SAMS access tokens to SSC. And I'll try to add this flag so that everything will "just work" on the backend.

PR #59843, opened 2024-01-25
auth: retry user creation once with random suffix when in collision

unknwon:
Make the users' life a little bit better.

## Test plan

CI


sqs (approved):
Idea LGTM and thank you! I am not familiar with this part of the code anymore.

eseliger (approved):

PR #59846, opened 2024-01-25
Move hardcoded config to jscontext

BolajiOlajide:
## Test plan

We override some of the site configuration with the `hardcodedConfig` , this PR moves those set of variables into JSContext so we can determine their values based on license features.


camdencheek (approved):
LGTM. Stamping to unblock, but would be nice to dig up the history of why these were hardcoded in the first place

valerybugakov (approved):
LGMT. I [left](https://sourcegraph.slack.com/archives/C07KZF47K/p1706151177737639?thread_ts=1706147005.647759&cid=C07KZF47K) a comment in the Slack thread about the history of these changes.

PR #59847, opened 2024-01-25
build: Bump go-tree-sitter version and fix Squirrel queries

varungandhi-src:
See individual commit messages for context. Without these, we get
two kinds of test failures:
- Hover docs for Java don't get picked up correctly
- Local variable navigation for Python doesn't work when the local
   variable is bound in an `except` clause

## Test plan

Added new tests + covered by existing tests.


Strum355 (approved):
Many thanks 🙏

kritzcreek (approved):
LGTM, thank you for the well structured commits

Issue #59792, opened 2024-01-24
Consolidate crates under syntax-highlighter


varungandhi-src:
We currently have the following dependency structure (generated via `cargo depgraph`)

![](https://github.com/sourcegraph/sourcegraph/assets/93103176/92af52a2-07ca-4819-8cd7-090a8cf807c0)

<details>
<summary> GraphViz code </summary>

```graphviz
digraph {
    graph [fontname = "Handlee"];
    node [fontname = "Handlee"];
    edge [fontname = "Handlee"];

    0 [ label = "sg-syntax" shape = box]
    1 [ label = "scip-syntax" shape = box]
    2 [ label = "scip-macros" shape = box, color = green3]
    3 [ label = "scip-treesitter" shape = box]
    4 [ label = "scip-treesitter-languages" shape = box]
    5 [ label = "scip-treesitter-cli" shape = box]
    6 [ label = "syntect_server" shape = box]
    0 -> 1 [ ]
    0 -> 3 [ ]
    0 -> 4 [ ]
    1 -> 2 [ color = green3]
    1 -> 3 [ ]
    1 -> 4 [ ]
    4 -> 2 [ color = green3]
    5 -> 1 [ ]
    5 -> 3 [ ]
    5 -> 4 [ ]
    6 -> 1 [ ]
    6 -> 3 [ ]
    6 -> 4 [ ]
    6 -> 0 [ ]
}
```

</details>

- `syntect_server` is the main entrypoint for highlighting:
  - Uses Tree-sitter for highlighting in some cases
  - Is different from `syntax-highlighter` which is the name of the overall service/Docker image (however, there is technically `http-server-stabilizer` sitting in front of it).
  - It also provides locals data
  - Proposal: Rename this to `syntax-server` (similar to `gitserver`).
- `scip-treesitter` is a library and doesn't have code for actually producing a SCIP index using Tree-sitter (unlike other `scip-*` indexers)
  - It has some utilities related to scip, unrelated to Tree-sitter.
      - Proposal: These can be folded back into the official SCIP bindings or into `scip-syntax` (or future `syntax-analysis`).
  -  Some utilities related to Tree-sitter.
      - Proposal: Fold this into `scip-syntax` (or future `syntax-analysis`).
        - [x] https://github.com/sourcegraph/sourcegraph/pull/59862
- `sg-syntax` has a bunch of highlighting related code that bridges between scip-syntax and syntax-highlighter.
[...]

PR #59850, opened 2024-01-25
build: Delete scip-macros crate

varungandhi-src:
Makes partial progress towards https://github.com/sourcegraph/sourcegraph/issues/59792

## Test plan

Builds and tests pass.


docker-images/syntax-highlighter/crates/scip-treesitter-languages/src/highlights.rs
```diff
@@ -143,25 +149,25 @@ lazy_static::lazy_static! {
...
+            (Nickel, "nickel"),
+            (Perl, "perl"),
+            (Pod, "pod"),
+            (Python, "python"),
+            (Ruby, "ruby"),
+            (Rust, "rust"),
+            (Scala, "scala"),
+            (Sql, "sql"),
+            (Xlsg, "xlsg"),
+            (Zig, "zig")
```
varungandhi-src, line 170: This is the only downside of removing the macros. We could use a [separate crate](https://crates.io/crates/casey) for this which ships a proc macro for lowercasing identifiers (and then stringify it), but I thought it's fine to have this list.

kritzcreek: I'm very happy with just having this list. Changes to it will be nice and local. I still feel like we're overusing macros in some places.

kritzcreek (approved):
Thank you!

PR #59851, opened 2024-01-25
gitserver: Migrate Blame to rpc call

eseliger:
This PR migrates the first more sophisticated call to a gRPC equivalent.

I had to generate quite a bunch of mocks and convert some things to interfaces to properly test this, so the diff looks scary but it's mostly generated code. Positive: Found a few bugs in the new gitcli backend by implementing these tests :)

## Test plan

Added a ton of tests, tried blame manually.


cmd/gitserver/internal/batchlog.go
```diff
@@ -16,7 +16,7 @@ import (
 // Note: The BatchLog endpoint has been deprecated. This file shall be removed after
 // the 5.3 release has been cut.
 
-func (s *Server) batchGitLogInstrumentedHandler(ctx context.Context, req *proto.BatchLogRequest) (resp *proto.BatchLogResponse, err error) {
+func (s *Server) BatchGitLogInstrumentedHandler(ctx context.Context, req *proto.BatchLogRequest) (resp *proto.BatchLogResponse, err error) {
```
eseliger, line 19: Temporary: Had to export the things that grpcServer uses, so that I could make it an interface and provide mocks. 
I expect that to be nicer one we're done with a few other things, and the server struct will be removed in favor of just the gRPC struct, plus a proper service layer.

internal/gitserver/v1/gitserver.proto
eseliger, line 1: Please roast my gRPC decisions here!

internal/gitserver/gitdomain/common_test.go
eseliger, line 1: Help, I would love to uncomment this test, but have been struggling to get `quick` to work with a `time.Time` in the struct.

internal/gitserver/commands_test.go
eseliger, line 1: nice: much fewer critical tests in this package, and no real git operations needed at this layer to actually verify that things work.

cmd/gitserver/internal/git/gitcli/clibackend.go
```diff
@@ -84,14 +88,15 @@ func (g *gitCLIBackend) gitCommand(ctx context.Context, args ...string) (wrexec.
 		return nil, cancel, ErrBadGitCommand
 	}
 
-	cmd := exec.Command("git", args...)
+	cmd := exec.CommandContext(ctx, "git", args...)
```
eseliger, line 91: oops, forgot to actually honor the context here.

cmd/gitserver/internal/git/gitcli/clibackend.go
```diff
@@ -148,6 +155,14 @@ func (rc *cmdReader) Read(p []byte) (n int, err error) {
 	return n, err
 }
 
+func (rc *cmdReader) Close() error {
+	if rc.closed {
+		return nil
+	}
+
+	return rc.ReadCloser.Close()
+}
```
eseliger, line 164: fixes error with double close

cmd/gitserver/internal/git/gitcli/blame_test.go
```diff
@@ -0,0 +1,451 @@
...
+func TestGitCLIBackend_Blame(t *testing.T) {
+	rcf := wrexec.NewNoOpRecordingCommandFactory()
+	reposDir := t.TempDir()
+
+	// Make a new bare repo on disk.
+	p := filepath.Join(reposDir, "repo")
+	require.NoError(t, os.MkdirAll(p, os.ModePerm))
+	dir := common.GitDir(filepath.Join(p, ".git"))
+
+	// Prepare repo state:
```
eseliger, line 36: note: actual low-level git tests are only needed at this package level - this feels much easier than the LocalGitCommand mock in the gitserver client, or always having to spin up whole a gitserver to test in the inttests package.

internal/gitserver/commands.go
```diff
@@ -825,9 +799,58 @@ func (c *clientImplementor) StreamBlameFile(ctx context.Context, repo api.RepoNa
...
+		endObservation(1, observation.Args{})
+
+		if s, ok := status.FromError(err); ok {
+			if s.Code() == codes.PermissionDenied {
+				return nil, errUnauthorizedStreamBlame{Repo: repo}
+			}
+		}
+
+		return nil, err
+	}
```
eseliger, line 846: how do we like this pattern to work around the issue that we faced in ArchiveReader before?

cmd/gitserver/internal/server_grpc.go
```diff
@@ -15,25 +15,65 @@ import (
...
 	proto "github.com/sourcegraph/sourcegraph/internal/gitserver/v1"
 	"github.com/sourcegraph/sourcegraph/internal/grpc/streamio"
 	"github.com/sourcegraph/sourcegraph/internal/trace"
 	"github.com/sourcegraph/sourcegraph/lib/errors"
+	"github.com/sourcegraph/sourcegraph/lib/pointers"
 )
 
-type GRPCServer struct {
-	Server *Server
+type service interface {
```
eseliger, line 33: I made this temporary interface to be able to test the grpc server more easily.

cmd/gitserver/internal/server_grpc.go
```diff
@@ -686,3 +724,92 @@ func (gs *GRPCServer) MergeBase(ctx context.Context, req *proto.MergeBaseRequest
...
+			CloneProgress:   notFoundPayload.CloneProgress,
+			Repo:            req.GetRepoName(),
+		})
+		if err != nil {
+			return err
+		}
+		return s.Err()
+	}
+
+	// First, verify that the actor has access to the given path.
```
eseliger, line 763: I decided that I believe that authz is an API concern and not a git backend concern after I initially had this check in the gitcli package, because these kinds of checks will not depend on the backend implementation (ie. a go-git version of this would not need to do a _different_ check to verify perms).

Thoughts?

cmd/gitserver/internal/server_grpc_test.go
eseliger, line 1: first gRPC server test!

internal/gitserver/commands.go
pjlast: So the normal Go way seems to be to only return errors once you start reading, not on creation of the reader. But I'm not particularly married to this idea if it causes confusion. Would be nice to see a "What the code looks like the normal way" vs. a "What the code looks like with this" next to each other.

pjlast (approved):

internal/gitserver/commands.go
eseliger: Oh, is that true? Maybe I've run myself into something here then.. 🤔 
I went by how the fs APIs work, and they return notfound errors from `fs.Open` instead of when reading from the file.

eseliger: I'm going to merge this for now, but please let's keep this conversation going before we migrate too many things :)

pjlast: Ah right, perhaps it's not quite so black and white then. I was thinking of permission errors (like trying to read from a file which you opened with only write permissions). In which case, returning a not found error here makes a lot more sense and you can ignore me :)

eseliger: gotcha, yeah the lines are blurry 🤔 I'm not 1000% convinced if this way or another way is The Right Way ™️, so I'm going to stick to this for now until I hate it 😆

PR #59861, opened 2024-01-25
Fix: ensure event.Argument is json

evict:
When we pass `nil` to `LogSecurityEvent` `json.Marshal`  turns it into `"null"`, that's why before we edit it with `jsonc.Edit` we need to make sure it's a valid JSON object string instead. We added an additional condition that checks if the argument is `"null"` and return `"{}"` instead.

## Test plan
CI tests, local test.

### DIFF
--- a/internal/database/security_event_logs.go
+++ b/internal/database/security_event_logs.go
@@ -112,8 +112,8 @@ type SecurityEvent struct {
 	Timestamp       time.Time
 }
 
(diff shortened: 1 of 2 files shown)


willdollman (approved):
Fix seems good and we tested it during pairing. Suggest we get a second pair of eyes though.

eseliger (approved):
Welcome back Will!

Issue #59593, opened 2024-01-15
Search Filters: do not reset the filters on query change


camdencheek:
_This is a proposal and is not ready for implementation._

When using the filters panel, I found myself regularly going through the following flow:
- Submit a relatively broad search query
- Realize that it returns more results than I intended and use the filters panel to narrow the search
- Either 1) realize my initial query is not quite what I'm looking for, or 2) find what I'm looking for and want to submit a followup search for a term that I discovered during my initial search
- Changing the query resets the filters, which means I lose the scope I was working in

So, what if changing the query did not reset the selected filters?

Pros:
- Relevant scope is likely to be "sticky" within a search session, so it saves the user some clicks
- It's not immediately obvious that changing the search query resets the filters, so it may be less surprising to the user. (I didn't expect to be surprised by this behavior, but I found I was)
- Query adjustments don't require re-selecting filters

Cons:
- Changing the search query without resetting filters would likely cause more zero-result states

camdencheek:
IIRC, we decided to reset the filters on query change because it was the easiest-to-implement option and it didn't have the problem that a filter added to the query could be mutually exclusive with a sidebar filter, leading to zero results.

I do not think it's high priority to consider this (it seems reasonable to ship the current resetting implementation), but I do think it might be a nice-to-have.

camdencheek:
Closing this since we didn't get to discussing it in more detail and I think we'd want more time to see it running before the code freeze if we were to experiment with it. If we want to pursue this, we can tackle it as a followup

vovakulikov:
It makes more relevant the fact that we're about to add `count:all` which changes the query and if we want this filter working properly we have to preserve filters when query is changing. Otherwise it would be. Broad search → narrow down → search limit hit → apply count all filter → filter panel is empty because query has been changed. 

In theory I could make an exception in the logic specially for the `count` filter but it feels like it would introduce more confuse about why my filters has been reseted when I did X

cc @taiyab @camdencheek

vovakulikov:
I'm going to reopen this one since I changes described in this issue need to be done for the `count:all` filter support on the client

vovakulikov:
Completed by https://github.com/sourcegraph/sourcegraph/pull/59868

PR #59868, opened 2024-01-25
Search Filters: Add count:all client filter

vovakulikov:
Support behaviour from https://github.com/sourcegraph/sourcegraph/issues/59593

Quick facts about `count:all` filter
- It isn't supported on the backend, so client should extend backend generated filters with new count kind
- `Count:all` filter should change the original query, we do not store anything in filters query param about count:all filters
- Since `count:all` will change the query we should support filter preservation as we change the query (prior to this PR any changes about the query reset filters completely) 
- Filter preservation tries to be a bit smart and resets filters if they no longer makes sense with new query, for example if you're picked Paths type in the filter panel and then select `Exclude go vendor` filter and then switch to the repository type, this filter will be reseted, because there is no such filter as exclude go vendor with repositories types
- This smart logic won't catch all cases, so sometimes there will be some empty state
- Because of possible empty states we no longer have any logic about show/hide filters in different filter types, if we have selected filter we will show it to you in any filter type. This is our safe way to reset filters if they don't make sense anymore.
- Browser history catches all changes as it was before.
- We show `count:all` filter unconditionally, because conditions for this might be very tricky (don't show if search limit wasn't hit may work but I found it very difficult to figure out from user point of view, we probably should apply this show/hide logic when we have more visible limit hit UI in the progress info header cc @taiyab)

## Test plan
- Check that `count:all` filter works as you expect it.
- Check that smart logic about preservation filters value works as you think it should work (any feedback is welcome)
- Check that browser history catches all filter and query change and there is neither trash or skipped states in browser history

### DIFF
--- a/client/wildcard/src/hooks/useSyncedWithUrlState.ts
+++ b/client/wildcard/src/hooks/useSyncedWithUrlState.ts
@@ -68,7 +68,7 @@ export function useSyncedWithURLState<State, SerializedState>(
 
             return `?${urlSearchParameters.toString()}`
         },
-        [navigate, serializer, urlKey, urlSearchParameters, replace]
+        [serializer, urlKey, urlSearchParameters]
     )
 
     return [queryParameter, setNextState, serializeURL]
--- a/client/wildcard/src/hooks/useSyncedWithUrlState.ts
+++ b/client/wildcard/src/hooks/useSyncedWithUrlState.ts
@@ -18,7 +18,7 @@ export interface UpdateSearchQueryOptions {
 export type SetStateResult<State> = [
     state: State,
     dispatch: (state: State, options?: UpdateSearchQueryOptions) => UpdatedSearchQuery,
-    serilize: (state: State) => UpdatedSearchQuery
+    serialize: (state: State) => UpdatedSearchQuery
 ]
 
 /**
--- a/client/wildcard/src/hooks/useSyncedWithUrlState.ts
+++ b/client/wildcard/src/hooks/useSyncedWithUrlState.ts
@@ -10,7 +10,16 @@ export interface URLStateOptions<State, SerializedState> {
 }
 
 export type UpdatedSearchQuery = string
-export type SetStateResult<State> = [state: State, dispatch: (state: State) => UpdatedSearchQuery]
+
+export interface UpdateSearchQueryOptions {
+    replace: boolean
+}
+
+export type SetStateResult<State> = [
+    state: State,
+    dispatch: (state: State, options?: UpdateSearchQueryOptions) => UpdatedSearchQuery,
+    serilize: (state: State) => UpdatedSearchQuery
+]
 
 /**
  * React hook analog standard react useState hook but with synced value with URL
@@ -30,7 +39,7 @@ export function useSyncedWithURLState<State, SerializedState>(
     )
 
     const setNextState = useCallback(
-        (nextState: State) => {
+        (nextState: State, options?: UpdateSearchQueryOptions) => {
             const serializedValue = serializer(nextState)
 
             if (serializedValue === null) {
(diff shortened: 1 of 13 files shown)


camdencheek (commented):
Is it possible to only show the filter if we've hit a limit? `count:all` is not needed unless a limit or timeout has been hit.

vovakulikov:
@camdencheek it's possible, I didn't do it because I was thinking that it should be unconditional filter since you can include count:all in your query any time. If you have concerns that this will increase number of searches with count all and hence increase load on the backend we can show it conditionally as limit is hit

camdencheek:
> you can include count:all in your query any time

That's true, but it doesn't always make _sense_ to include it. If you're already getting exhaustive results, there's no need to add `count:all`, so it's just extra noise in the UI. And I definitely do have some reservations about making it _that_ much more obvious of a feature 🙂

vovakulikov:
@taiyab @camdencheek there was a suggestion from @peterguy about preserve logic over filters. So imagine the following case 
- You're on the code filter type and you select `lang:` filter in the filters UI
- Then you switch to the repository filter type and we remove lang filters since it's no longer valid anymore (there is no lang filter when you have `type:repo`)
- After this you switch to the path and there lang: filter makes sense again but we don't have it because we removed it on the previous step when we were switching from Code to Repo filters types

Question: would you expect to have lang filters after this switching? 

My biggest concern here that we never remove filters on query change in this case (only if use unselect filter manually) this means that we preserve them always in the query even if them don't make sense and we don't show them. This would lead to the situation when you clicked some filters then change the query and get some other set of filters, then when you share link with someone and this someone also switching types in their UI they will see your old filters that you had when you were switching between filters types.

vovakulikov:
@camdencheek I added filter show/hide logic by condition of the progress and state status

client/wildcard/src/hooks/useSyncedWithUrlState.ts
```diff
@@ -10,7 +10,16 @@ export interface URLStateOptions<State, SerializedState> {
...
-export type SetStateResult<State> = [state: State, dispatch: (state: State) => UpdatedSearchQuery]
+
+export interface UpdateSearchQueryOptions {
+    replace: boolean
+}
+
+export type SetStateResult<State> = [
+    state: State,
+    dispatch: (state: State, options?: UpdateSearchQueryOptions) => UpdatedSearchQuery,
+    serilize: (state: State) => UpdatedSearchQuery
```
camdencheek, line 21:
```suggestion
    serialize: (state: State) => UpdatedSearchQuery
```

camdencheek (approved):
LGTM. Just to check: @taiyab, the plan here is to still move the `count:` functionality into a redesigned result counter at some point, right?

camdencheek:
> Question: would you expect to have lang filters after this switching?

I don't feel strongly about this, but I would not expect to have lang filters after switching, and I think I'd actually find it surprising if we did. It's a piece of state that is not visible to the user when switched to a different result type. Honestly, I think even having repo stick around when changing result types is surprising given the layout of the page, but I personally think the convenience outweights the minor confusion. (Ideally, I'd still like to see us tackle the "better context" problem and move repo filters into its own, special UI, but that's a conversation for another time). 

Anyways, I think this feels nice. Thanks for the updates!

taiyab:
> LGTM. Just to check: @taiyab, the plan here is to still move the `count:` functionality into a redesigned result counter at some point, right?

Yes. Probably not 5.3.

>Question: would you expect to have lang filters after this switching?

My biggest concern here that we never remove filters on query change in this case (only if use unselect filter manually) this means that we preserve them always in the query even if them don't make sense and we don't show them. This would lead to the situation when you clicked some filters then change the query and get some other set of filters, then when you share link with someone and this someone also switching types in their UI they will see your old filters that you had when you were switching between filters types.

I don't think we should preserve filter states in the filter panel at all when switching types. I wouldn't expect it to, and we should treat the filters below a particular type option to uniquely "belong" to that type.

PR #59885, opened 2024-01-26
Update cody/ssc packages to use SAMS tokens; support multiple SAMS environments

chrsmith:
We are trying to introduce the ability for dotcom to pull Cody subscription information from the SSC service, as part of eventually having it be the source of truth for this information.

https://github.com/sourcegraph/sourcegraph/pull/59763 was the first step towards this. It introduced a `cody` package to be the sole location where we fetch Cody subscription information. And then as an implementation detail, that package referred to an `ssc` package with an API client for contacting the SSC backend.

The next step is to update the SSC client to use SAMS access tokens for authorization. As it is configured today, it is using a "shared secret" which is no longer supported on the SSC backend. Nor has it ever been configured for the live dotcom environment. (https://github.com/sourcegraph/sourcegraph/pull/59783/files was a prototype showing how to do that.)

This PR updates the SSC client to use SAMS tokens. But also does it in such a way that dotcom can support TWO versions of SAMS (`-dev` and `-prod`) concurrently. In order to facilitate end-to-end testing and dogfooding of Cody, and without a Sourcegraph.com test environment, we need to expose a per-user feature flag to enable a user having their subscription information pulled from a lower SSC environment.

... and to authenticate requests to that SSC-dev environment, we need to use a SAMS token for the SAMS-dev environment. (Yes that is confusing. But fortunately the two services share a hostname. So that's nice.)

> **Why is it important to support BOTH -dev and -prod SAMS instance concurrently?**
> Otherwise, we aren't able to test Cody :'( See https://docs.google.com/document/d/1IIdeJqQ_Xwkgw8m0dydCYJTRMByowoymFXzth3fqRoI/edit.

I also did a bunch of cleanup and refactoring to just get my bearings for how the code was supposed to work. But can tease that out if it seems too risky or would make this more difficult to review.

## Questions for Reviewers

I'm assuming that lookups to get the status of a feature flag are essentially free, since the flag set is part of the `context.Context`. Is that correct?

Needing every SSC request to first hit the database to look up the user's attached identities, doesn't seem great. But this is only done as part of GraphQL queries that fetch the user's subscription information. (So presumably not exactly on the critical path.) But what options are available for making this faster? Initializing the SAMS configuration lazily, but keeping them in a package-level variable?

I am renaming some feature flags to (arguably) make them easier to understand. But how bad of an idea is this? e.g. does having orphaned feature flags cause major problems?

## Test plan

Not yet. I want to confirm that he approach is sound before updating the tests.

### DIFF
--- a/internal/cody/subscription.go
+++ b/internal/cody/subscription.go
@@ -11,15 +11,16 @@ import (
 	"github.com/sourcegraph/sourcegraph/lib/errors"
 )
 
-// use-ssc-for-cody-subscription is a feature flag that enables the use of SSC as the source of truth for Cody subscription data.
-const USE_SSC_FOR_SUBSCRIPTION_FF = "use-ssc-for-cody-subscription"
+const (
+	// FeatureFlagCodyProTrialStillAvailable controls whether or not we grant users who have signed up for the "Cody Pro"
+	// free trial access to Pro features. The default for this is true, but we flip it to false on February 15, 2024.
+	// (And in a subsequent release, remove it and all references.)
+	FeatureFlagCodyProTrialStillAvailable = "ssc.cody-pro-trial-still-available"
 
-// cody-pro-trial-ended is a feature flag that indicates if the Cody Pro "Free Trial"  has ended.
-// (Enabling users to use Cody Pro for free for 3-months starting in late Q4'2023.)
-const CODY_PRO_TRIAL_ENDED_FF = "cody-pro-trial-ended"
-
-const SAMS_SERVICE_ID = "https://accounts.sgdev.org"
-const SAMS_SERVICE_TYPE = "openidconnect"
+	// FeatureFlagReadCodySubscriptionFromSSC controls whether or not we should attempt to read a user's Cody Pro
+	// subscription information from the SSC backend. Defaults to false while the SSC integration is in development.
+	FeatureFlagReadCodySubscriptionFromSSC = "ssc.read-cody-subscription-from-ssc"
+)
 
 type UserSubscriptionPlan string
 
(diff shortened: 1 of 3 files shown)


internal/cody/subscription.go
```diff
@@ -48,100 +49,107 @@ type UserSubscription struct {
...
+func SubscriptionForUser(ctx context.Context, db database.DB, user types.User) (*UserSubscription, error) {
 	var subscription *ssc.Subscription
 
-	// Fetch subscription from SSC only if the feature flag is enabled and the user has a SAMS account ID
-	if samsAccountID != "" && featureflag.FromContext(ctx).GetBoolOr(USE_SSC_FOR_SUBSCRIPTION_FF, false) {
-		subscription, err = sscClient.FetchSubscriptionBySAMSAccountID(samsAccountID)
+	tryReadSSCSubscription := featureflag.FromContext(ctx).GetBoolOr(FeatureFlagReadCodySubscriptionFromSSC, false)
+	if tryReadSSCSubscription {
+		// The SSC client shouldn't be created on every call.
+		sscClient := ssc.NewClient()
```
bobheadxi, line 133: Maybe the `ssc` package's `NewClient` constructor can use a `sync.OnceValue`? https://pkg.go.dev/sync#OnceValue

Then, every call to `ssc.NewClient` will actually return the same client

bobheadxi (commented):
> I'm assuming that lookups to get the status of a feature flag are essentially free, since the flag set is part of the context.Context. Is that correct?

You should be fine here - these flags are widely used, and if they present a problem, let's optimize when we have evidence via metrics/traces that his is a problem

> Needing every SSC request to first hit the database to look up the user's attached identities, doesn't seem great. But this is only done as part of GraphQL queries that fetch the user's subscription information. (So presumably not exactly on the critical path.) But what options are available for making this faster? Initializing the SAMS configuration lazily, but keeping them in a package-level variable?

IIRC we do external account lookups in a variety of other points as well, so the status quo is "it's okay" - as above, let's optimize when we have an identified problem 😁 @unknwon might have other ideas on this front though

> I am renaming some feature flags to (arguably) make them easier to understand. But how bad of an idea is this? e.g. does having orphaned feature flags cause major problems?

It depends if these flags are configured anywhere - nothing will migrate flags for you automatically, so you'll want to check if these are configured on dotcom and make sure the new FFs are configured as well before rolling this out

Other than the above, I also left a comment about how to share the SSC client - does that some some of the todos/bug comments in the diff?

internal/ssc/ssc.go
```diff
@@ -1,35 +1,125 @@
...
+		return nil, errors.Wrap(err, "creating SAMS token")
+	}
+
+	// Issue the request.
 	req, err := http.NewRequest(method, url, nil)
 	if err != nil {
 		return nil, err
 	}
-	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", c.secretToken))
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", samsToken))
```
bobheadxi, line 122: I realized I didn't respond to your question about this, but I followed up again in https://github.com/sourcegraph/sourcegraph/pull/59783#discussion_r1464372564 (we can also continue discussion here)

bobheadxi: Ah, nevermind - see https://github.com/sourcegraph/sourcegraph/pull/59783#discussion_r1467116309, it looks like the oauth2 client doesn't do that much for you, it requests a new token every time, so it's not materially different from what you do here

internal/ssc/ssc.go
```diff
@@ -1,35 +1,125 @@
...
+	}
+	if sams == nil {
+		return "", errors.New("no appropriate SAMS auth provider found")
+	}
+
+	// BUG: We shouldn't create a new token for every request, and instead use the proper
+	// libraries to cache and refresh automatically. (But that would require that the
+	// SSC client is long-lived, meaning we need to fix that in the cody package first.)
+	clientCreds := &clientcredentials.Config{
+		ClientID:     sams.ClientID,
```
bobheadxi, line 96: Looks like the fix is to use https://sourcegraph.com/github.com/golang/oauth2@adffd94437cee5e033b67282039fc9c21708ffc7/-/blob/oauth2.go?L367-379

internal/cody/subscription.go
chrsmith: That's a good suggestion, but looking closer at the implementation of `NewClient`, I'll just remove the comment. (Calling new client repeatedly doesn't look like it would be a problem.)

I'm assuming that fetching `SiteConfig()` is just loading something from memory, and so this is just copying some data and doing an allocation. (So unless we are striving for optimal performance, `sync.OnceValue` might just be overkill.) 

```go
	sgconf := conf.Get().SiteConfig()
	return &client{
		baseURL: sgconf.SscApiBaseUrl, // [sic] generated code
	}
```

bobheadxi: Just for memory's sake it's not worth it, but maybe you want it to implement https://github.com/sourcegraph/sourcegraph/pull/59885#discussion_r1467117288 ?

internal/ssc/ssc.go
chrsmith: Cool. I'll look into it. As long as the HTTP client and whatever caching/transport layer are persisted in memory (rather than needing to be recreated every time we call `NewClient`) then that would probably be a big improvement. (Which should be the case? Since we get the client ID from the site config, so I assume it isn't reinitialized or anything.)

internal/ssc/ssc.go
bobheadxi: Relevant issue: https://github.com/golang/oauth2/issues/84#issuecomment-1080690104

It might be less straight-forward than it seems, given the length of this thread 😬

bobheadxi (approved):
No blocking comments, just stuff about holding the OAuth2 client, but I don't see anything that might cause this to _not_ work - we can optimize later. 

The rate limit evaluation is a bit complex, I trust that tests/your understanding of the desired behaviour here ensures this is what you want 😆

internal/cody/subscription.go
chrsmith: OK, never mind, yeah we'll want to initialize this once so that we can have the different HTTP clients for the SAMS instances live longer.

I'll beat this PR with `sync.Once` until this is fixed.

internal/cody/subscription.go
```diff
@@ -11,15 +11,16 @@ import (
...
-const USE_SSC_FOR_SUBSCRIPTION_FF = "use-ssc-for-cody-subscription"
-
-// cody-pro-trial-ended is a feature flag that indicates if the Cody Pro "Free Trial"  has ended.
-// (Enabling users to use Cody Pro for free for 3-months starting in late Q4'2023.)
-const CODY_PRO_TRIAL_ENDED_FF = "cody-pro-trial-ended"
-
-const SAMS_SERVICE_ID = "https://accounts.sgdev.org"
-const SAMS_SERVICE_TYPE = "openidconnect"
+const (
+	// FeatureFlagCodyProTrialStillAvailable controls whether or not we grant users who have signed up for the "Cody Pro"
```
thenamankumar, line 15: 🚨 We can not change feature flag names all the time. Cody's Clients are already using them. I suggest not changing them and introducing additional work for the Clients. 

Moreover, it is always better to have a feature flag to reflect the change we tend to make from the current state. 

`ssc. Cody-pro-trial-still-available` will have to be set to `true` by default, which is not preferred.

We should only care about this feature flag on 15th Feb when we do the release, and the action at that time will be to end the free trial, so the feature flag should be set to end it and be named accordingly.

internal/cody/subscription.go
```diff
@@ -48,100 +49,107 @@ type UserSubscription struct {
...
+		CurrentPeriodEndAt:   periodEnd,
 	}
 
-	if len(accounts) > 0 {
-		return accounts[0].AccountID, nil
+	// Check if the user has signed up for the Cody Pro free trial.
+	// We'll give them the benefits of a Cody Pro subscription until we end the trial offer.
+	//
+	// At that time, we will be pulling all subscription information from the SSC backend.
+	coryProTrialStillAvailable := !featureflag.FromContext(ctx).GetBoolOr(FeatureFlagCodyProTrialStillAvailable, true)
```
thenamankumar, line 116: 🚨 even if the pro trial has ended, we need to return the plan as 'pro' because clients need to show notifications to the users. Here we are only returning the plan as pro is the trial is active. This will break all the things we are planning. 

IMHO, the readability is degraded by extracting the free trial signup & active logic here and updating the particular fields. It makes it difficult to visualise exactly what is being returned.

It will become more complex when introducing a pending logic of return "free" subscription state if the subscription from SSC has "cancelled" status. In that case, we need to start the free tier period from the end date of the cancelled subscription period. That is being worked on in ||.

thenamankumar (changes requested):
🚨 I have not reviewed the SAMS token changes in depth yet, but there are major bugs in the changes that are made to refactor the present code. 

I highly recommend on splitting the refactoring and SAMS token changes into 2 separate PRs.

internal/cody/utils.go
```diff
@@ -22,48 +21,25 @@ func withCurrentTimeMock(ctx context.Context, t time.Time) context.Context {
...
-		startDayOfTheMonth = daysInStartingMonth
+// fakeSubscriptionDateRange returns a fake date range for a Cody subscription.
+// Since this isn't backed by Stripe or anything, the exact dates don't matter.
+//
+// If set, we peg the end date to February 15. (To align with the expected Cody
+// Pro release, when we expect existing free trials to end.) Otherwise the end
+// of the month is chosen.
+func fakeSubscriptionBillingCycle(endOnFeb15 bool) (time.Time, time.Time) {
+	now := time.Now()
+	startOfTheMonth := time.Date(now.Year(), now.Month(), 1, 0, 0, 0, 0, time.UTC)
```
thenamankumar, line 32: 🚨 We can not just set the startDate as the start of the month because it won't reflect when the usage stats will be reset for them.

Here in dotcom, this data is strictly used to tell free users when their usage will reset. The gateway cache is set to 30 days TTL when the user is created. Ideally, the billing period should start when the user gets created and should be tied to the user creation date. 

Otherwise, people will hit rate limits, and we will be showing them the wrong current period start day. 

To remove this whole logic from dotcom after the release, which we all want, we need SSC to take care of the free tier subscription; right now, it just returns null, which should not be the case. SSC should take care of the billing cycle of free users as well as resetting their rate limits. 

It will be good, if you can create a ticket to track this.

internal/cody/subscription.go
chrsmith: Gotcha. I was unaware that Cody Clients were already depending on this. And the explanation makes sense. Thanks.

internal/cody/subscription.go
chrsmith: I don't quite follow. Could you clarify what you mean?

>  even if the pro trial has ended, we need to return the plan as 'pro' because clients need to show notifications to the users. Here we are only returning the plan as pro is the trial is active. This will break all the things we are planning.

Are you saying that _after_ a trial has ended, when the user should no longer still be receiving the Cody Pro functionality (`ApplyProRateLimits == false`) that you _still_ want the subscription to have the the plan be `UserSubscriptionPlanPro`?

I don't know "all the things we are planning". But is what you are expecting is that after the trial ends the fields will change like this:

```
Plan              : Pro        -> Pro
ApplyProRateLimits: true       -> false
Status            : "trialing" -> "canceled"
```

If so, then, yes, that seems totally reasonable and perhaps less confusing users.

> It will become more complex when introducing a pending logic of return "free" subscription state if the subscription from SSC has "cancelled" status. In that case, we need to start the free tier period from the end date of the cancelled subscription period.

This doesn't make sense to me. (I was also unaware that it was being worked on in parallel. Are you working on that?)

What is this "pending logic of return 'free' subscription state"? Just like the note above, when are you expecting the "free" status to be returned? It seems like we can either:

1. Have it change from "Pro" to "Free" as soon as the trial ends, and the subscription status be "active". (i.e. you are on an active "free" plan, and can upgrade to "pro" if you wish.)
2. Have it remain on "Pro" indefinitely (but with the status of "canceled"). (i.e. you are on a "pro" plan, but it has been canceled, so you no longer get the benefits of the subscription.)

I think we can convince ourselves that either approach would work. But it sounds like you were assuming we would go with one of those? (Or is there a third option I hadn't considered?)

internal/cody/utils.go
chrsmith: OK, yes, I was unaware that the subscription billing cycle anchors were being used to for how usage stats are managed... but if that is the case, that sounds like we shouldn't be doing that.

> Here in dotcom, this data is strictly used to tell free users when their usage will reset. The gateway cache is set to 30 days TTL when the user is created.

And then what? After it gets reset we have their limit tracked for another 30 days from that point on? That doesn't seem especially robust, as it will inevitably drift over time. (As opposed to the more complicated notion of "on the Nth of the month", like how Stripe handles subscriptions.)

> Ideally, the billing period should start when the user gets created and should be tied to the user creation date.

Sure, but I this will not be the case. Not ever. It isn't feasible, and even if it were, it would very much be a bad idea.

> people will hit rate limits, and we will be showing them the wrong current period start day.

... so this is the root problem:

Cody Gateway has a time frame for which usage limits get reset. And that is in no-way related to the data that is coming from SSC. Right?

So we can choose to try to have these "synthesized/fake" subscriptions match whatever logic Cody Gateway has. Or, we need to have the Gateway cache match these dates exactly. (And make sure that they are stable.)

Do I have that right? Would having the Cody Gateway TTL be based on the `currentUser.codySubscription.endDate` be the right fix for this?

> We need SSC to take care of the free tier subscription; right now, it just returns null, which should not be the case. SSC should take care of the billing cycle of free users as well as resetting their rate limits.

Yes. We can add that logic any time we want. (Have SSC _ALWAYS_ return a subscription for a user, even if the user account is unknown or something.)

But unless I am mistaken, that still wouldn't be enough, because the gateway cache's TTL needs to align with the billing cycle. And so we'd need to update that piece too? (So in other words, getting Gateway to respect the subscription's billing cycle anchor is the problem to solve. Having the SSC backend return "free" tier subscriptions would just make that easier to implement.)

chrsmith:
> I have not reviewed the SAMS token changes in depth yet, but there are major bugs in the changes that are made to refactor the present code.

Yes, thanks @thenamankumar. Clearly you and I aren't on the same page about how you are expecting this all to fit together. (And even from the PR description it was clear that it was trying to do too much.)

I'll close this out and tease out the SAMS token changes into a separate PR. (Perhaps without the added complexity of the SAMS-dev override.)

But as far as the subscription-related behaviors, I still don't quite follow what you are thinking there. I will file a ticket and try to capture the relevant questions. 🙏

PR #59890, opened 2024-01-26
build: Merge sg-syntax into scip-syntax

varungandhi-src:
The bulk of the logic in the sg-syntax crate is related to highlighting.
Since that is generic information related to highlighting, move
it to scip-syntax which will be renamed to syntax-analysis in the future.

Moving it inside the syntax-highlighter itself was another option,
but I didn't go for that, so that we can separately benchmark the
syntax-analysis crate, instead of introducing a new library target
in syntax-highlighter and then benchmarking that.

But I'm open to moving this logic into syntax-highlighter too,
it's not a big deal IMO.

## Test plan

Covered by existing build & tests.


kritzcreek:
I think it would make sense to group the highlighting logic into syntax-highlighter, but my main objection is that I'd rather not depend on a webserver in the syntax-analysis crate. Maybe it even makes sense to keep the webserver portion as its own crate?

varungandhi-src:
> but my main objection is that I'd rather not depend on a webserver in the syntax-analysis crate

I can move this out in a follow-up PR. Sound reasonable? This PR is primarily for removing one crate, so most of the changes are renaming, not splitting the code into more appropriate locations. Similarly, I haven't moved some highlighting related code from tree-sitter-all-languages to syntax-analysis yet.

kritzcreek (approved):

PR #59900, opened 2024-01-26
search jobs: change extension to .jsonl

stefanhengl:
Search Jobs returns results as line separated JSON, but with a .json extension. This can cause confusion because users might exepect a well formated json file.

This changes the extension to [.jsonl](https://jsonlines.org/) and updates content-type to `application/jsonlines`.

Test plan:
manual testing

### DIFF
--- a/client/branded/src/search-ui/results/progress/StreamingProgressSkippedPopover.tsx
+++ b/client/branded/src/search-ui/results/progress/StreamingProgressSkippedPopover.tsx
@@ -374,7 +374,8 @@ export const ExhaustiveSearchMessage: FC<ExhaustiveSearchMessageProps> = props =
             )}
 
             <Text className={classNames(validationError && 'text-muted', styles.exhaustiveSearchText)}>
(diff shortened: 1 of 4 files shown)


keegancsmith (approved):
Maybe update mime type as well? Not standardized, but saw a suggestion of `application/jsonl+json` since then chrome treats the file as plain text.

keegancsmith:
> I couldn't confirm that Shopify uses `plain/jsonl`. Did you mean [`text/jsonl`](https://shopify.dev/docs/api/usage/bulk-operations/imports#generate-the-uploaded-url-and-parameters)?

Yes sorry `text/jsonl`.

> AWS seems to use [`application/jsonlines`](https://docs.aws.amazon.com/sagemaker/latest/dg/cdf-inference.html) which they are also discussing in the [issue](https://github.com/wardi/jsonlines/issues/19) linked on the JSON lines page.

I think pick whatever you like the most. Looking at that issue lots of examples of people picking different things. AWS's choice seems like a good bet since they likely have some of the most consumed APIs in the world :)

PR #59902, opened 2024-01-26
search: factor out display limit logic into displayFilter struct

keegancsmith:
We are going to make this logic more complicated. This is a first step to better document and isolate the logic from the core http streaming logic.

The intention is to add limiter (or possibly mutators) based on ChunkMatch.Content sizes.

Test Plan: go test

### DIFF
--- a/cmd/frontend/internal/search/display.go
+++ b/cmd/frontend/internal/search/display.go
@@ -1,0 +1,46 @@
+package search
+
+import "github.com/sourcegraph/sourcegraph/internal/search/result"
+
+// displayFilter mutates event matches before sending them to the end user
+// (after all statistics/etc have been calculated).
+//
+// In particular displayFilter is responsible for ensuring we only send
+// display limit results (?display URL parameter).
+//
+// Note: this is different to the search limit (MaxResults). Search limit
+// decides when to stop searching, this decides when to stop sending the
+// actual matches to the user.
+type displayFilter struct {
+	// MatchLimit is the computed display limit (?display URL parameter).
+	// After MatchLimit results have been sent to the user we stop streaming
+	// more matches.
+	MatchLimit int
+
+	matchesRemaining int
+}
+
+// newDisplayFilter returns the displayFilter for args. maxResults is the
+// maximum results we are allowed to search. This is used to bound the display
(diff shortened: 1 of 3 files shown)


cmd/frontend/internal/search/display.go
```diff
@@ -0,0 +1,46 @@
...
+		matchesRemaining: displayLimit,
+	}
+}
+
+// Limit will limit m to enforce the display limits.
+//
+// Note: this should only be called after aggregating statistics.
+func (d *displayFilter) Limit(m *result.Matches) {
+	d.matchesRemaining = m.Limit(d.matchesRemaining)
+}
```
camdencheek, line 46: Should we just make this a general-purpose "LimitedStream" or something that can be located with and compose with our other [stream infrastructure](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@3bc52d0a2451988f09e956341670489842c3ef77/-/blob/internal/search/streaming/stream.go)?

camdencheek (approved):

cmd/frontend/internal/search/display.go
keegancsmith: Can do that, but is there an example of where else we would use it? If not, I'd prefer to keep it non-exported living next to were it does its job. I'll go ahead and land this, but if you have an idea will open up a PR for it.

camdencheek: Well, we already have a `limitStream` used for the `LimitJob`. We could ideally reuse that so that the display limit and the result limit are just to different layers of the same limiter.

camdencheek: (I don't feel strongly about this though. Happy to be told "nah, this is a case reuse doesn't make sense")

keegancsmith: Yeah I'm not sure this makes sense, since the display limit lives right at the http layer and needs to run after the http layer has computed its statistics and filters. So kinda different to the stream. However, once I have finished the work in this area I will relook at your suggestion. We could compose this stuff in a more data flow way, but I kinda like it being a bit more procedural in this area and a bit less RxJS like

PR #59904, opened 2024-01-26
build: Simplify tree-sitter-all-languages & collect queries in syntax-analysis

varungandhi-src:
Makes partial progress towards https://github.com/sourcegraph/sourcegraph/issues/59792

At this point, `tree-sitter-all-languages` basically only has some ad-hoc language
detection stuff (hopefully, we can get rid of it at some point) and is mainly just a magnet
to hold all the grammars together.

## Test plan

Build & existing tests


kritzcreek (approved):
LGTM

PR #59911, opened 2024-01-26
Update SSC client to use SAMS access tokens

chrsmith:
Update the `ssc.Client` code to use SAMS access tokens for authentication rather than a shared secret. (Having SSC use the shared secret for auth was only going to be a temporary solution, and SAMS apparently supports OAuth clients and issuing tokens well enough to use this now(*).)

(*) Using the SSC client is behind a feature flag while the system is under development, but also because I do not believe the dotcom instance actually has the SAMS-prod identity provider wired up. (So if this were enabled, I don't think it would actually work on Sourcegraph.com as-is.)

This PR also performs some minor stylistic cleanups, adds some comments, and fixes a bug in the unit tests that should have been caught by the linter in CI (❗❓) .

> Re: bugs, the issue was that since `mockSSCClient.FetchSubscriptionBySAMSAccountID` was on a non-pointer receiver, assignment to `m.Called` would have been visible. My IDE identified this, and is easy enough to fix.
>
> I'm surprised that it got through CI. Do we need to opt-in to Golang linting or something? If I run `golangci-lint run` locally it... has some problems with the repo. Is this expected for the `sourcegraph/sourcegraph` repo?

## Test plan

None (⚠️) I haven't ran this locally. I will try to get my local setup running and configured to work with SAMS running locally. (I'll update the PR description when I do.)

I would love to have unit tests for the SSC client, but I don't know if it makes sense to mock out all the layers of the HTTP client, its transport, just to confirm the `clientcredentials.Config` type works as expected. (Since that's all outside of our control.)

### DIFF
--- a/internal/cody/subscription.go
+++ b/internal/cody/subscription.go
@@ -2,6 +2,7 @@ package cody
 
 import (
 	"context"
+	"strings"
 	"sync"
 	"time"
 
@@ -109,20 +110,23 @@ func consolidateSubscriptionDetails(ctx context.Context, user types.User, subscr
 func getSAMSAccountIDForUser(ctx context.Context, db database.DB, dotcomUserID int32) (string, error) {
 	// TODO(sourcegraph#59786): Support pulling a user's SAMS-dev identity so we can
 	// fetch subscription information from SSC-dev, for dogfooding.
-	accounts, err := db.UserExternalAccounts().List(ctx, database.ExternalAccountsListOptions{
+	oidcAccounts, err := db.UserExternalAccounts().List(ctx, database.ExternalAccountsListOptions{
 		UserID:      dotcomUserID,
-		ServiceID:   "https://accounts.sgdev.org", // SAMS-prod
 		ServiceType: "openidconnect",
 		LimitOffset: &database.LimitOffset{
-			Limit: 1,
+			Limit: 3,
 		},
 	})
 	if err != nil {
 		return "", errors.Wrap(err, "listing external accounts")
 	}
 
-	if len(accounts) > 0 {
-		return accounts[0].AccountID, nil
+	for _, account := range oidcAccounts {
+		// The ServiceID will be be something like "https://accounts.sourcegraph.com",
+		// reusing the constant from the ssc package to avoid duplication.
+		if strings.Contains(account.ServiceID, ssc.SAMSProdHostname) {
+			return account.AccountID, nil
+		}
 	}
 	return "", nil
 }
@@ -157,6 +161,10 @@ func SubscriptionForUser(ctx context.Context, db database.DB, user types.User) (
 // getSSCClient returns a self-service Cody API client. We only do this once so that the stateless client
 // can persist in memory longer, so we can benefit from the underlying HTTP client only needing to reissue
 // SAMS access tokens when needed, rather than minting a new token for every request.
+//
+// BUG: If the SAMS configuration is added or changed during the lifetime of the this process, the returned
+// client will be invalid. (As it would be using the original SAMS client configuration data.) The process
+// will need to be restarted to correct this situation.
 var getSSCClient = sync.OnceValue[ssc.Client](func() ssc.Client {
 	return ssc.NewClient()
 })
--- a/internal/cody/subscription.go
+++ b/internal/cody/subscription.go
@@ -2,7 +2,7 @@ package cody
 
 import (
 	"context"
-	"strings"
+	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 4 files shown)


internal/cody/subscription.go
```diff
@@ -11,15 +13,12 @@ import (
 	"github.com/sourcegraph/sourcegraph/lib/errors"
 )
 
-// use-ssc-for-cody-subscription is a feature flag that enables the use of SSC as the source of truth for Cody subscription data.
-const USE_SSC_FOR_SUBSCRIPTION_FF = "use-ssc-for-cody-subscription"
+// featureFlagUseSCCForSubscription determines if we should attempt to lookup subscription data from SSC.
+const featureFlagUseSCCForSubscription = "use-ssc-for-cody-subscription"
```
chrsmith, line 17: @thenamankumar , Golang uses a camel/Pascal cae naming convention. But is there a particular reason why you went for the ALL_CAPS? e.g. is there some codeine or regex-based tool working behind the scenes?

Similarly, I assume it is preferable to not export these flags to other packages. Is that a good idea? Or are feature flags typically exported by convention?

internal/cody/subscription.go
```diff
@@ -100,41 +102,50 @@ func consolidateSubscriptionDetails(ctx context.Context, user types.User, subscr
...
+//
+// If the user has not associated a SAMS identity with their dotcom user account,
+// will return ("", nil). After we migrate all dotcom user accounts to SAMS, that
+// should no longer be possible.
+func getSAMSAccountIDForUser(ctx context.Context, db database.DB, dotcomUserID int32) (string, error) {
+	// TODO(sourcegraph#59786): Support pulling a user's SAMS-dev identity so we can
+	// fetch subscription information from SSC-dev, for dogfooding.
+	oidcAccounts, err := db.UserExternalAccounts().List(ctx, database.ExternalAccountsListOptions{
+		UserID:      dotcomUserID,
+		ServiceType: "openidconnect",
```
chrsmith, line 115: Looking at the `List(...)` functionality, it appears that `self` will only add the `AND` clauses if the option is provided. So rather than matching the full service ID ("https://accounts.sgdev.org") we loop through all OIDC identities and identify which one is SAMS-prod.

Happy to change it back to the way it was, but I was thinking this will maybe make it easier to do the update to somehow support SAMS-dev and SAMS-prod concurrently. Also, since we do something similar -- but not on the "external identity" but looking for the "SAMS auth provider" -- in the `ssc` package, I figured reusing the hostname would be slightly better. (But it's arguable.)

internal/ssc/ssc.go
```diff
@@ -1,96 +1,151 @@
...
+// sendRequest issues an HTTP request to SSC. If supplied, the response will be unmarshalled into outBody as JSON.
+func (c *client) sendRequest(ctx context.Context, method string, url string, outBody *Subscription) (*int, error) {
+	// Build and send the request.
+	req, err := http.NewRequest(method, url, nil /* body */)
 	if err != nil {
 		return nil, err
 	}
 
+	tr, _ := trace.New(ctx, "sendSSCRequest", attribute.String("url", url))
+	resp, err := c.httpClient.Do(req)
```
bobheadxi, line 65: You usually want to keep the returned `ctx` from `trace.New` and pass it to `req` (`req.WithContext` IIRC) so that you get propagation to other layers. This requires a bit of careful variable assignment usually (you can check other callsites of `trace.New` for examples)

internal/ssc/ssc.go
```diff
@@ -1,96 +1,151 @@
...
+	tr.SetError(err)
+	tr.End()
+	if err != nil {
```
bobheadxi, line 68: If this is intended to only wrap the `httpClient`, you would usually use instrumentation _in_ the thing you call, e.g. https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp or instrument in a helper

bobheadxi: I think a better trace here might be to encapsulate _all_ of `sendRequest`, and capture the error/close the span in a `defer`

internal/ssc/ssc.go
chrsmith: Good catch! I didn't see `http.Client.DoWithContext()` but didn't look hard enough to see `NewRequestWithContext`. I'll double check other calls to `trace.New` to confirm that's the typical idiom though.

internal/ssc/ssc.go
chrsmith: 🤦 yeah, that sounds like a much better solution. Since then we'd only need to put the tracing around a single function call, rather than potentially needing to add `tr.End()` to multiple exit points for the function.

Given that we are already using a customized `http.Client` from `clientcredentials.Config.Client()` I would prefer not to try and use the `otelhttp` client though, since it seems like an easy thing to wire up incorrectly?

chrsmith:
Addressed feedback from Robert, rebased on top of `main`, and then force-pushed.

internal/cody/subscription.go
thenamankumar: meh, I do not have a good reason, switching too much between ts and go doesn't help. 

Thanks for the renames.

internal/cody/subscription.go
```diff
@@ -100,41 +102,50 @@ func consolidateSubscriptionDetails(ctx context.Context, user types.User, subscr
...
 		},
 	})
 	if err != nil {
-		return "", errors.Wrap(err, "error while fetching user's external account")
+		return "", errors.Wrap(err, "listing external accounts")
 	}
 
-	if len(accounts) > 0 {
-		return accounts[0].AccountID, nil
+	for _, account := range oidcAccounts {
```
thenamankumar, line 124: We can restore this change based on the comment [above](https://github.com/sourcegraph/sourcegraph/pull/59911/files#r1470013716).

thenamankumar (approved):
Overall LGTM, dropped 2 comments. Approving to unblock.

PR #59913, opened 2024-01-26
truncate commit message in commit view if necessary

jasonhawkharris:
Before, we showed the full commit message on the commit view. Now, we give the user the option to see more, or less of the commit message.

## Test plan
Manual testing/passing CI/CD

### DIFF
--- a/client/web/src/repo/commits/GitCommitNode.tsx
+++ b/client/web/src/repo/commits/GitCommitNode.tsx
@@ -21,6 +21,7 @@ import { GitCommitNodeByline } from './GitCommitNodeByline'
 
 import styles from './GitCommitNode.module.scss'
 
+const TRUNCATED_COMMIT_MESSAGE_LENGTH = 240
 export interface GitCommitNodeProps {
     node: GitCommitFields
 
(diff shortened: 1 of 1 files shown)


client/web/src/repo/commits/GitCommitNode.tsx
```diff
@@ -353,7 +371,6 @@ export const GitCommitNode: React.FunctionComponent<React.PropsWithChildren<GitC
                             {!extraCompact && <Link to={canonicalURL}>{oidElement}</Link>}
                             {afterElement}
                         </div>
-                        {commitMessageBody}
```
camdencheek, line 356: I'm surprised to see this line deleted. Doesn't this just make it impossible to actually show the commit message body?

client/web/src/repo/commits/GitCommitNode.tsx
```diff
@@ -157,14 +159,30 @@ export const GitCommitNode: React.FunctionComponent<React.PropsWithChildren<GitC
...
+                <Linkified input={truncatedCommitMessage} externalURLs={node.externalURLs} />
+                {truncationNeeded && (
+                    <Button
+                        variant="link"
+                        size="sm"
+                        display="inline"
+                        onClick={() => setTruncateCommitMessage(!truncateCommitMessage)}
+                    >
+                        {truncateCommitMessage ? 'see more' : 'see less'}
+                    </Button>
```
camdencheek, line 181: We also use the `commitMessageBody` to render the byline below. I do not think we ever want an expand/collapse button in the byline

Issue #59898, opened 2024-01-26
Search Results: Search result snippet hover


taiyab:
I want to be able to hover an individual match and get a hover effect.

https://www.loom.com/share/01da2f468c2a4436912981c4a1340c70?sid=352345ff-965e-4c8f-b326-72579ceda14b

PR #59914, opened 2024-01-26
Search results: improve interactivity

camdencheek:
This adds hover reactivity and cleans up focus state in the search results page.

It also includes a bit of a CSS cleanup rampage. We had a ton of different overlapping, duplicate, or unused CSS classes that were interacting unpredictably, so I consolidated. Now, most shared classes live in `ResultContainer.module.scss` and any result-type-specific classes live in the corresponding result type modules.

[Loom video walkthrough](https://www.loom.com/share/8e6a0232f6b14993afc47e6973fb8e85)

Fixes https://github.com/sourcegraph/sourcegraph/issues/59898
Stacked on https://github.com/sourcegraph/sourcegraph/pull/59908

## Test plan

Manual testing. Relying on screenshot testing to uncover unexpected changes.


client/branded/src/search-ui/components/ResultContainer.module.scss
```diff
@@ -31,10 +31,94 @@
...
+    }
+}
+
+.clickable {
+    cursor: pointer;
+    &:hover {
+        text-decoration: none;
+        background-color: var(--subtle-bg-2);
+    }
+}
```
camdencheek, line 65: These are the primary new classes that are shared among the result types. 

`.horizontal-divider-between` is for symbol results and file content results, where we have multiple sub-sections that need to have a separator. Previously, we duplicated this styling for each result type.

`.focusable-block` is a block that can be focused. Importantly, this allows us to _massively_ clean up our padding and margins because previously, we were specifically leaving space outside the components for the focus ring to display. Instead, this implements the focus ring as an _inset_ box shadow so that 1) the component does not get resized at all on focus, and 2) we don't need to predict the size of the focus ring and leave space for it. This also makes it render much more predictably since before, shenanigans with z-index meant that the hover ring was often cut off. 

`.clickable` is the shared styling for any clickable block. `--subtle-bg-2` was the best color I could find that looked decent in both light and dark mode, but I'm not at all tied to the color.

client/branded/src/search-ui/components/SymbolSearchResult.module.scss
```diff
@@ -1,8 +1,3 @@
-.symbols-override {
-    border: none !important;
-    padding: 0 !important;
-}
```
camdencheek, line 4: 👋 `!important`

client/branded/src/search-ui/components/ResultContainer.module.scss
```diff
@@ -31,10 +31,94 @@
...
```
camdencheek, line 65: These are the primary new classes that are shared among the result types. 

`.horizontal-divider-between` is for symbol results and file content results, where we have multiple sub-sections that need to have a separator. Previously, we duplicated this styling for each result type.

`.focusable-block` is a block that can be focused. Importantly, this allows us to _massively_ clean up our padding and margins because previously, we were specifically leaving space outside the components for the focus ring to display. Instead, this implements the focus ring as an _inset_ box shadow so that 1) the component does not get resized at all on focus, and 2) we don't need to predict the size of the focus ring and leave space for it. This also makes it render much more predictably since before, shenanigans with z-index meant that the hover ring was often cut off. 

`.clickable` is the shared styling for any clickable block. `--subtle-bg-2` was the best color I could find that looked decent in both light and dark mode, but I'm not at all tied to the color.

taiyab:
Great work @camdencheek! 

I'm tempted to make the focus state for an individual result even more prominent.

image

Something like this: https://www.figma.com/file/EPgl0qVbqGSumvf52brhda/Code-Search?type=design&node-id=1381%3A23333&mode=design&t=6E5IayNkoI2L75cG-1

Light and dark mode examples provided.

taiyab (approved):
LGTM. Approved! 👍

PR #59925, opened 2024-01-29
search: remove V4

stefanhengl:
We don't need V4 for the next release. We always send the pattern type if keyword search is active. 

Test plan:
CI

### DIFF
--- a/internal/search/client/client.go
+++ b/internal/search/client/client.go
@@ -262,10 +262,8 @@ func detectSearchType(version string, patternType *string) (query.SearchType, er
 			searchType = query.SearchTypeLiteral
 		case "V3":
 			searchType = query.SearchTypeStandard
-		case "V4":
-			searchType = query.SearchTypeKeyword
 		default:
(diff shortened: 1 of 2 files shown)


keegancsmith (approved):
I would keep rc1 since we may wanna make more changes soon to the query language. Once we promote to v4 I think that means we support it for a long time?

Lets do a quick chat about it at our sync today

internal/search/client/client.go
```diff
@@ -262,10 +262,10 @@ func detectSearchType(version string, patternType *string) (query.SearchType, er
 			searchType = query.SearchTypeLiteral
 		case "V3":
 			searchType = query.SearchTypeStandard
-		case "V4-rc1":
```
jtibshirani, line 265: Maybe we can entirely back out the version bump for this change, since we decided not to make any API changes for this release. I'm not seeing a benefit to having a new version right now (no matter what it's called, "V4-rc1" or "V4") ... what do you think?

keegancsmith: aah I like this suggestion. Gets my vote.

PR #59928, opened 2024-01-29
client: Update lsif-{clang,dart,php}->scip-{clang,dart,php}

varungandhi-src:
We should not be recommending the old LSIF indexers

## Test plan

n/a

### DIFF
--- a/internal/codeintel/autoindexing/internal/inference/libs/update-shas.sh
+++ b/internal/codeintel/autoindexing/internal/inference/libs/update-shas.sh
@@ -15,6 +15,7 @@ fi
 
 SCRIPT_DIR="$(dirname "${BASH_SOURCE[0]}")"
 
+# No scip-clang as that doesn't have a Docker image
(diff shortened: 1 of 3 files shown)


internal/codeintel/autoindexing/internal/inference/libs/update-shas.sh
```diff
@@ -15,7 +15,7 @@ fi
 
 SCRIPT_DIR="$(dirname "${BASH_SOURCE[0]}")"
 
-for indexer in lsif-clang scip-go lsif-rust scip-rust scip-java scip-python scip-typescript scip-ruby; do
```
kritzcreek, line 18: Should we be adding `scip-clang` to this list?

varungandhi-src: No, because there is no Docker image for scip-clang.

kritzcreek (approved):

PR #59929, opened 2024-01-29
Stops updating lsif-rust docker image

kritzcreek:
We're not actually using this anywhere, as the default autoindexing config for Rust uses `scip-rust` instead.

## Test plan

n/a

### DIFF
--- a/internal/codeintel/autoindexing/internal/store/enqueuer_test.go
+++ b/internal/codeintel/autoindexing/internal/store/enqueuer_test.go
@@ -143,9 +143,9 @@ func TestInsertIndexes(t *testing.T) {
 			},
 			LocalSteps:  nil,
 			Root:        "/baz",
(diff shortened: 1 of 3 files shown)


internal/codeintel/autoindexing/internal/store/enqueuer_test.go
```diff
@@ -143,9 +143,9 @@ func TestInsertIndexes(t *testing.T) {
 			},
 			LocalSteps:  nil,
 			Root:        "/baz",
-			Indexer:     "sourcegraph/lsif-rust:15",
```
kritzcreek, line 146: This change isn't strictly necessary, but I figured I'd just get rid of `lsif-rust` references. (The last remaining one is in the `doc/` directory, which I'm not allowed to change)

varungandhi-src (approved):

PR #59930, opened 2024-01-29
removes dead code in codeintel upload

kritzcreek:
This code was only "used" when testing itself. (At least as far as I can tell, my Go navigation skills are still nascent)

## Test plan

CI passing


varungandhi-src (approved):
If there are no tests for this, it should be fine to delete it.

Issue #59738, opened 2024-01-22
Strict Gitserver API surface


eseliger:
To improve debuggability and implement meaningful performance improvements, we're going to provide a strictly-typed git interface from gitserver that doesn't split the responsibilities of dealing with git output and data between the server and client packages.
The outcome will be that we can switch simple commands over to methods like libgit2 or go-git as a follow-up because we don't rely on the log output format anymore on the client side, without introducing further breaking changes to the API.
This will also help increase the security of gitserver, as we will be able to move subrepo perm checks into the server, so we never even return data that shouldn't be read by the given actor.

List of gitserver.Client methods that we likely want to migrate:

- [x] `ArchiveReader(ctx context.Context, repo api.RepoName, options ArchiveOptions) (io.ReadCloser, error)` @pjlast 
  - Consolidate error handling on server-side so that notfound errors are returned before any messages are streamed
- [x] `BatchLog(ctx context.Context, opts BatchLogOptions, callback BatchLogCallback) error` @eseliger 
  - Already a separate endpoint, although the data is very low-level, might consider rolling this into other requests.
- [x] `StreamBlameFile(ctx context.Context, repo api.RepoName, path string, opt *BlameOptions) (HunkReader, error)` @eseliger 
- [x] `GetDefaultBranch(ctx context.Context, repo api.RepoName, short bool) (refName string, commit api.CommitID, err error)` @eseliger 
- [x] `GetObject(ctx context.Context, repo api.RepoName, objectName string) (*gitdomain.GitObject, error)` @ggilmore 
- [x] `HasCommitAfter(ctx context.Context, repo api.RepoName, date string, revspec string) (bool, error)` @eseliger 
  - Could this be implemented by getting the latest commit on branch X, and compare the date then? If so, let's not create a separate endpoint here. There's only a single user of this method, that could encapsulate this logic.
- [x] `ListRefs(ctx context.Context, repo api.RepoName) ([]gitdomain.Ref, error)` @eseliger 
- [x] `ListBranches(ctx context.Context, repo api.RepoName, opt BranchesOptions) ([]*gitdomain.Branch, error)` @eseliger 
- [x] `MergeBase(ctx context.Context, repo api.RepoName, base, head string) (api.CommitID, error)` @eseliger 
- [x] `ResolveRevision(ctx context.Context, repo api.RepoName, spec string, opt ResolveRevisionOptions) (api.CommitID, error)` @eseliger 
[...]

PR #59933, opened 2024-01-29
Move ArchiveReader's git exec command to gitcli

pjlast:
This PR migrates the ArchiveReader function's use of the exec endpoint to be in line with the rest of #59738

Additionally, sub-repo permission checks are now done on the server instead of in the client code.

## Test plan

New tests added and moved as appropriate


cmd/gitserver/internal/git/gitcli/archivereader.go
```diff
@@ -0,0 +1,49 @@
...
+func (g *gitCLIBackend) ArchiveReader(ctx context.Context, format, treeish string, pathspecs []string) (io.ReadCloser, error) {
+	// This is a long time, but this never blocks a user request for this
+	// long. Even repos that are not that large can take a long time, for
+	// example a search over all repos in an organization may have several
+	// large repos. All of those repos will be competing for IO => we need
+	// a larger timeout.
+	ctx, cancel := context.WithTimeout(ctx, conf.GitLongCommandTimeout())
+
+	archiveArgs := buildArchiveArgs(format, treeish, pathspecs)
+	cmd, _, err := g.gitCommand(ctx, archiveArgs...)
```
eseliger, line 20: the cancel func must always be called - I know very specifically here this is not true because there's a context passed right above but it does scare me a bit that we could forget this if we ever changed something about the timeout 😬

cmd/gitserver/internal/server_grpc.go
```diff
@@ -191,37 +191,70 @@ func (gs *grpcServer) Archive(req *proto.ArchiveRequest, ss proto.GitserverServi
...
 
-	// This is a long time, but this never blocks a user request for this
-	// long. Even repos that are not that large can take a long time, for
-	// example a search over all repos in an organization may have several
-	// large repos. All of those repos will be competing for IO => we need
-	// a larger timeout.
-	ctx, cancel := context.WithTimeout(ss.Context(), conf.GitLongCommandTimeout())
-	defer cancel()
-
-	return gs.doExec(ctx, execReq, w)
```
eseliger, line 224: love that this is gone :wooo:

cmd/gitserver/internal/server_grpc.go
```diff
@@ -191,37 +191,70 @@ func (gs *grpcServer) Archive(req *proto.ArchiveRequest, ss proto.GitserverServi
...
+		return s.Err()
 	}
 
-	if req.GetFormat() == string(gitserver.ArchiveFormatZip) {
-		execReq.Args = append(execReq.Args, "-0")
+	if !actor.FromContext(ctx).IsInternal() {
+		if enabled, err := gs.subRepoChecker.EnabledForRepo(ctx, repoName); err != nil {
+			return errors.Wrap(err, "sub-repo permissions check")
+		} else if enabled {
+			return errors.New("archiveReader invoked for a repo with sub-repo permissions")
```
eseliger, line 215: might be nice to augment this with a status -- perhaps codes.Unimplemented?

cmd/gitserver/internal/server_grpc.go
```diff
@@ -191,37 +191,70 @@ func (gs *grpcServer) Archive(req *proto.ArchiveRequest, ss proto.GitserverServi
...
+		}
 	}
 
-	execReq.Args = append(execReq.Args, req.GetTreeish(), "--")
-	execReq.Args = append(execReq.Args, req.GetPathspecs()...)
+	backend := gs.getBackendFunc(repoDir, repoName)
+
+	r, err := backend.ArchiveReader(ss.Context(), req.GetFormat(), req.GetTreeish(), req.GetPathspecs())
```
eseliger, line 221: can use `ctx` here

cmd/gitserver/internal/server_grpc.go
```diff
@@ -191,37 +191,70 @@ func (gs *grpcServer) Archive(req *proto.ArchiveRequest, ss proto.GitserverServi
 		return status.Error(codes.InvalidArgument, "empty repo or format")
 	}
```
eseliger, line 192: could we split this check in two so we get accurate error messages? 😬

cmd/gitserver/internal/server_grpc.go
```diff
@@ -176,6 +174,8 @@ func (gs *grpcServer) Exec(req *proto.ExecRequest, ss proto.GitserverService_Exe
 }
 
 func (gs *grpcServer) Archive(req *proto.ArchiveRequest, ss proto.GitserverService_ArchiveServer) error {
+	ctx := ss.Context()
+
 	// Log which which actor is accessing the repo.
 	accesslog.Record(ss.Context(), req.GetRepo(),
```
eseliger, line 180: `ctx`
```suggestion
	accesslog.Record(ctx, req.GetRepo(),
```

cmd/gitserver/internal/server_grpc.go
```diff
@@ -191,37 +191,70 @@ func (gs *grpcServer) Archive(req *proto.ArchiveRequest, ss proto.GitserverServi
...
+		}
 	}
 
-	execReq.Args = append(execReq.Args, req.GetTreeish(), "--")
-	execReq.Args = append(execReq.Args, req.GetPathspecs()...)
+	backend := gs.getBackendFunc(repoDir, repoName)
+
+	r, err := backend.ArchiveReader(ss.Context(), req.GetFormat(), req.GetTreeish(), req.GetPathspecs())
+	if err != nil {
+		if os.IsNotExist(err) {
```
eseliger, line 223: do we actually return this error type from the gitcli?

cmd/gitserver/internal/server_grpc.go
```diff
@@ -191,37 +191,70 @@ func (gs *grpcServer) Archive(req *proto.ArchiveRequest, ss proto.GitserverServi
...
+				Repo: string(repoName),
+			})
+			if err != nil {
+				gs.logger.Error("failed to marshal status", log.Error(err))
+				return err
+			}
+			return s.Err()
+		}
+		var e *gitdomain.RevisionNotFoundError
+		if errors.As(err, &e) {
```
eseliger, line 234: do we return revisionnotfounderrors properly?

cmd/gitserver/internal/server_grpc_test.go
```diff
@@ -396,6 +396,123 @@ func TestGRPCServer_ReadFile(t *testing.T) {
...
+			assert.NoError(t, err)
+			mockassert.Called(t, srp.EnabledForRepoFunc)
+		})
+
+		t.Run("subrepo perms are enabled, returns error", func(t *testing.T) {
+			srp.EnabledForRepoFunc.SetDefaultReturn(true, nil)
+			err := gs.Archive(&v1.ArchiveRequest{Repo: "therepo", Format: "zip"}, mockSS)
+			assert.Error(t, err)
+			require.Contains(t, err.Error(), "archiveReader invoked for a repo with sub-repo permissions")
+			mockassert.Called(t, srp.EnabledForRepoFunc)
```
eseliger, line 451: would be good to assert on the code as well

internal/gitserver/commands.go
```diff
@@ -2247,14 +2247,6 @@ func (c *clientImplementor) ArchiveReader(
 	})
 	defer endObservation(1, observation.Args{})
```
eseliger, line 2248: Would be nice to fix this observability thing, the NewFileReader did that

internal/gitserver/commands.go
```diff
@@ -2247,14 +2247,6 @@ func (c *clientImplementor) ArchiveReader(
...
 
-	if authz.SubRepoEnabled(c.subRepoPermsChecker) {
-		if enabled, err := authz.SubRepoEnabledForRepo(ctx, c.subRepoPermsChecker, repo); err != nil {
-			return nil, errors.Wrap(err, "sub-repo permissions check:")
-		} else if enabled {
-			return nil, errors.New("archiveReader invoked for a repo with sub-repo permissions")
-		}
-	}
-
 	if ClientMocks.Archive != nil {
```
eseliger, line 2250: let's get rid of mocks in this layer, the idea is that people should mock the Client itself, not individual methods of it. 
The NewFileReader PR https://github.com/sourcegraph/sourcegraph/pull/59919/files#diff-8391844b0f80cedc2822763af030c766f674322a7458266479d0e7b94bed0ffcR2424-R2505 did this by mocking the grpc client instead. Would also be nice to get rid of the hack: comment further down about firstChunk buffering, because that's gonna be standard practice now.

eseliger (commented):
Would be great to get some additional doc string added to the proto file, that mentions the clone enqueueing and what errors are returned, like here: https://github.com/sourcegraph/sourcegraph/pull/59919/files#diff-651d337c467c327550ce3c4b4ba643acaf9a424bdb02372c328d842d08bbb1a1R100-R108

------

Sorry for the plethora of comments, happy to also just hack on it together :D 
Just trying to use this opportunity to resolve a lot of testing and general tech debt as we touch things anyways :)

eseliger (approved):
Reviewed over zoom.

PR #59934, opened 2024-01-29
search: Don't trigger git fetch in symbol search job

eseliger:
This is in line with other jobs, and makes more sense IMO wrt. the comment right above.

Let me know if this assessment is wrong.

## Test plan

CI.


keegancsmith (approved):
LGTM

PR #59936, opened 2024-01-29
gitserver: Fixup and improve error handling

eseliger:
- Use new standard mechanism for starting clones on the backend for search, so no repoCloned check is needed in the search handler.
- Make sure the same error conversion logic that was applied with exec calls is applied to pure gRPC call handlers as well.

We could also try to wrap the automaticRetryClient in another layer that auto-transforms errors from gRPC, and from Recv functions.

**Update**: Converted to a wrapper for the client, so we never forget any calls.

## Test plan

Existing test suites, ran stack locally and clicked through.


ggilmore (approved):
Looks good!

I do wonder if we can use generics to implement all the stream wrappers that do the error handling? It could save a bit of typing.

PR #59937, opened 2024-01-29
gitserver: Fixup error conversion to RepoNotExistError in search

eseliger:
Seems to be a flipped condition, fixing here.

## Test plan

Test suite and code review.


camdencheek (approved):
thank you!

PR #59940, opened 2024-01-29
search: boost phrase queries

stefanhengl:
This uses the new `query.Boost` operator exposed by Zoekt to boost phrase queries. The feature is behind the feature-flag "search-boost-phrase".

The heuristic to detect phrases is very simple: We detect phrases if the query is a top-level AND query without nesting or negated terms.

Test plan:
- New unit test
- Manual testing: I checked locally with the feature flag turned off and on and compared the results.


internal/search/query/transformer.go
```diff
@@ -617,3 +617,50 @@ func ToBasicQuery(nodes []Node) (Basic, error) {
 	}
 	return Basic{Parameters: parameters, Pattern: pattern}, nil
 }
+
+// ExperimentalPhraseBoost appends a phrase query to the original query but only
+// if the original query consists of a single top-level AND expression. The
+// purpose is to improve ranking of exact matches by adding a phrase
+// query for the entire query string.
```
keegancsmith, line 624: you additionally need to only do this if there are 3 or more terms.

internal/search/query/transformer.go
```diff
@@ -617,3 +617,50 @@ func ToBasicQuery(nodes []Node) (Basic, error) {
...
+//
+// Example:
+//
+//	foo bar -> (or (and foo bar) ("foo bar"))
+func ExperimentalPhraseBoost(node Node) Node {
```
keegancsmith, line 629: should we be using some sort of Map thing so we apply this to all possible And nodes?

keegancsmith (approved):
LGTM

internal/search/query/transformer.go
stefanhengl: Good idea. We might have to do this manually though. Looking at `Visitor`, I suspect it won't like that we modify the tree as we go. I will do this in a follow-up PR.

PR #59944, opened 2024-01-29
Enterprise attribution client refreshes on site config changes

### LINKED ISSUE #59576: Allow configuring gateway when customer uses BYOK
At this point gateway works in the enterprise instance only when embeddings or completions use gateway. https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Asourcegraph%2Fsourcegraph.git&branch=cb%2F59262%2Fenterprise-instance-uses-gateway-for-attribution&file=internal%2Fcodygateway%2Fclient.go&editor=VSCode&version=2.2.16&start_row=46&start_col=19&end_row=46&end_col=19

For guardrails, we need to allow for separate config in case customer is BYOK

### LINKED ISSUE #59701: Properly handle errors in gateway client setup on attribution endpoint in enterprise instance
Currently if gateway is not configured, guardrails will panic.

Also consider re-initializing gateway client every time site config changes.
https://github.com/sourcegraph/sourcegraph/pull/59761/files/8871608e07ba2ed5d055478e9ca061995b21f8ea#diff-0380878f2392758d1876992ca20c56203feaa5e923068ca9855db918dddfbe3fR59

cbart:
Closes #59576
Closes #59701

- Introduce `attribution.gateway { endpoint, token }` site config.
- For attribution this endpoint can be used for BYOK customers which do not configure gateway.
- Initialization of attribution service changes to `func () attribution.Service`. When these factory-functions are called:
  - For dotcom, nothing changes: We initialize client at startup and keep returning it.
  - For enterprise, we re-initialize client on demand the moment site-config changes.

TODO

- [ ] Redact secret.
- [ ] (optional) Move updates to attribution service not to duplicate between attribution test and graphQL implementation.

## Test plan

- [x] Manually check via graphQL after site config changes
- [x] Automated test based on graphQL queries.
- [x] Test against gateway with no Cody subscription


cmd/frontend/internal/guardrails/init.go
```diff
@@ -28,53 +29,73 @@ func Init(
...
+type enterpriseInitialization struct {
+	observationCtx *observation.Context
+	mu             sync.Mutex
+	client         codygateway.Client
+	endpoint       string
+	token          string
+}
+
+func (e *enterpriseInitialization) Service() attribution.Service {
+	e.mu.Lock()
```
chwarwick, line 55: This seems fine to me, although I'm more used to seeing us do this with a conf.watch.

schema/site.schema.json
```diff
@@ -2987,6 +2987,23 @@
       "!go": {
         "pointer": true
       }
+    },
+    "attribution.gateway": {
+      "description": "Use this gateway parameters for customers that bring their own key. Otherwise gateway endpoint is used.",
```
chwarwick, line 2992: This is always going to be the gateway endpoint right? It's just a matter of if we want to use the dev endpoint internally for local development?

internal/codygateway/client.go
```diff
@@ -48,7 +48,6 @@ func NewClientFromSiteConfig(cli httpcli.Doer) (_ Client, ok bool) {
 	config := conf.Get().SiteConfig()
 	cc := conf.GetCompletionsConfig(config)
 
-	// TODO: What if customer is BYOK? How do we talk to gateway then?
```
chwarwick, line 51: Have you tried this against the Gateway with a subscription that does not have Cody enabled which would be similar to a BYOK customer.   I'm not sure the request will be immediately rejected.

chwarwick (commented):
I think you should remove at least the access token from the config since it's generated push and the default values into computed.go.

internal/codygateway/client.go
cbart: Filed a bug: https://github.com/sourcegraph/sourcegraph/issues/59996

cbart (commented):
Thanks Chris! Please take another look!

cbart:
Still need to redact secret, but hopefully I'll make this last change and merge it in the morrow.

@chwarwick if you don't see anything blocking, please leave a ✅ and I'll amend your remaining comments and land.

Sorry for the size of the PR - it's the test 😓

chwarwick (approved):
Looks good nice job.

PR #59950, opened 2024-01-29
dev: add search-boost-phrase to list of flags

stefanhengl:
This adds "search-boost-phrase" to the list of feature flags and also sorts the list alphabetically (sorry for the diff :-().
This way we can use `feat=search-boost-phrase` for quick testing.

Test plan:
- tested locally
- only affects dev settings


jtibshirani (approved):
Looks good. Hope we can promote this soon to be the default behavior!

PR #59954, opened 2024-01-29
gitserver: Simplify accesslogging

eseliger:
This PR reduces the number of interceptors that need to be invoked to check if we should be logging. We only log if accesslog has been called in the handler. This also makes it so that we cannot forget to add the method here so that access logging works. That was not the case for the perforce handlers before.

## Test plan

Ran manually and checked that access logs are logged correctly.


ggilmore:
So, my understanding is that we explicitly don't want to have access logs for _every_ gitserver method since that will generate a huge log volume. 

I might be missing something, but does this PR maintain this behavior? It looks like it will log for every method if access logs are enabled.

Another way of solving this problem is to change the method specific interceptor to instead take a set (`map[string]struct{}`) of methods that should trigger the access log. The interceptor will check to see if the interceptor is a member of that set. This will avoid the large chain of interceptors problem (now, we just have one)

eseliger:
This should be maintained, because the accesslog.Log method has to be explicitly called in the method handler, otherwise the middleware bails out because no context is found.
At least that's how I understand the code 🤔

pjlast (approved):

PR #59956, opened 2024-01-29
msp: deployment rollout strategies

### LINKED ISSUE #186: SIGTERM does not shut down server
`webpack-serve` registers a SIGTERM/SIGINT handler that doesn't function properly so it hangs forever and you can't shut down the process.

That package is now deprecated so let's switch to something in use instead.

bobheadxi:
Allows services to define a `rollout` spec that ensures new image releases go through a specified sequence and flow. We do this using Cloud Deploy and custom targets that update the Cloud Run service image and configuring Terraform to ignore image changes.

> [!NOTE]
> We use a custom target (as opposed to using the native Cloud Deploy + Cloud Run integration, which wants the entire spec in YAML for releases - see https://github.com/sourcegraph/managed-services/issues/186#issuecomment-1915196511) because everything else we have is generated in Terraform, and the core Cloud Run configuration extensively references Terraform values. It would be an extensive undertaking to change how this works. For the most part, this is to deploy a new version of the service code, and it can be beneficial to tie that to the service repository's CI to make it clear where a piece of code goes - building the custom target to _only_ roll out images allows us to do that.

Custom targets are not yet supported by the GCP Terraform provider, which is unfortunate - instead we have to render some YAML that can be applied with a `gcloud` command. For the most part, this should be a one-time operation. There is generated guidance on what to do with the generated output, and also how to create releases.

Closes https://github.com/sourcegraph/managed-services/issues/186

Kinda rambly, high-level Loom overview: https://www.loom.com/share/55bfa34d173c40a9b78708de2029f34f?sid=6f1b062d-ba02-4bb9-8abe-c9f8f8f9a8fe

### Configuring rollouts

In the top-level service spec:

```yaml
rollout:
  stages:
    - environment: test
    - environment: robert
```

And in each relevant environment:

```yaml
- id: robert
  projectID: msp-testbed-robert-7be9
  category: test
  deploy:
    type: rollout
```

`sg msp generate` will render resources for the "last" stage to house Cloud Deploy infrastructure.

### Creating releases

Creating a release triggers a rollout, which progresses through the specified stages, like so:

image

Creating releases is intended to be run using `gcloud` commands for now - we could introduce a `sg msp` command for this later. The command creates a release targeting the Cloud Deploy pipeline that exists in the final-stage project. Example command (one is also generated in the pipeline YAML file docstrings):

```sh
gcloud deploy releases create manual-test-04-2024-01-31 \
    --project=msp-testbed-robert-7be9 \
    --region=us-central1 \
    --delivery-pipeline=msp-testbed-us-central1-rollout \
    --source='gs://msp-testbed-robert-7be9-cloudrun-skaffold/source.tar.gz' \
    --labels="commit=abc123,author=foo" \
    --deploy-parameters="customTarget/tag=dd34d1be076e_2024-01-31"
```

Promotions can happen at any time - not every release needs to be promoted to the subsequent stage - and currently must happen manually for each stage except the first.

A secret/output is provisioned with a "release creator" SA that can be used to create a [workload identity pool](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/infrastructure/-/blob/managed-services/continuous-deployment-pipeline/main.tf?L5-20) that can be used to run the `gcloud deploy releases create` command in CI.

After the first apply, which now assumes an `insiders` tag, Terraform no longer touches the image via a [lifecycle ignore](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle)

### Rollout execution

Rollouts happen via a new `clouddeploy-executor` SA in the last stage, which is granted sufficient IAM roles to deploy Cloud Run revisions.

The "render" step in Skaffold prepares a release - in our case, generating a `deploy.sh` with the prerequisite arguments. A record is available in the the relevant "rollout" page:

image

image

The "deploy" step just downloads the artifact and executes it.

### Tracing a release

You can include arbitrary labels on releases - this shows up in the release entity in GCP console, but we don't yet propagate anything very well down to the Cloud Run revision. In particular, it seems like we don't get the tag information in the revision UI, but if you click "edit" you see the correct tag populated:

image  image

See `skaffold.yaml` - we're mostly just executing commands with `gcloud`, and reporting expected outputs. We can extend this with more detailed outputs and additional tagging or scripting if we want - examples I've seen often build a custom binary/image to execute more advanced use cases. Also see https://cloud.google.com/deploy/docs/custom-targets

### Rollbacks

[Cloud Deploy has a concept of rollbacks](https://cloud.google.com/deploy/docs/roll-back), which you can apply via UI - it seems this just runs the previous configuration:

image
image

## Test plan

See https://github.com/sourcegraph/managed-services/pull/454 and https://console.cloud.google.com/deploy/delivery-pipelines/us-central1/msp-testbed-us-central1-rollout?project=msp-testbed-robert-7be9 . I also specifically tested that deploying a particular image, and then deploying a change in Terraform, does not overwrite the image, and we do not have infinite drift on the Terraform when releases deploy images.

Also https://github.com/sourcegraph/managed-services/actions/runs/7744296405


michaellzc:
took a quick look at code, it looks good to me.

will try it out tomorrow if I have time.

unknwon:
Nice!

Few high-level questions:
1. For rollbacks, do it create conflicts if I roll back on Cloud Run but not on Cloud Deploy (i.e. should roll back on Cloud Run be prohibited to not confuse Cloud Deploy)?
1. Is rolling back on Cloud Deploy would effectively "pause" subsequent deployments until some sort of "manual-resume"?

jac:
🚀 
Had a read through and it all looks good and seems straightforward enough

I do wonder though if it would make sense to extract the delivery pipeline into a separate project (either a single project for all MSP services or a project per service)

That way it would be easier to alter the stages without having to completely recreate the pipeline if a new stage gets appended.

bobheadxi:
@unknwon:

> For rollbacks, do it create conflicts if I roll back on Cloud Run but not on Cloud Deploy (i.e. should roll back on Cloud Run be prohibited to not confuse Cloud Deploy)?

I don't think so - under the hood, the Cloud Deploy "rollback" rolls back to the previous state, which is recorded in the deploy artifacts, currently a `gcloud` command in a `deploy.sh` file like in the "Rollout execution" screenshot in the PR description. If you roll back in Cloud Run, then do the same roll back in Cloud Deploy, the Cloud Deploy one will just roll to the version indicated in the previous `deploy.sh` artifact which creates an identical revision.

> Is rolling back on Cloud Deploy would effectively "pause" subsequent deployments until some sort of "manual-resume"?

Unfortunately, it requires a spec change to suspend a pipeline - see https://cloud.google.com/deploy/docs/suspend-pipeline . I can add this toggle to this PR (the pipeline is configured in CDKTF). Alternatively the deploying team would implement the pause on the "push" side

> If I click another "edit and deploy" on Cloud Run, would Cloud Deploy auto-correct the deployed version, or would it get confused?

It will auto-correct the version to the one indicated in the `gcloud deploy releases create` publisher, since we just say "deploy this tag" irrespective of the image that is live

bobheadxi:
> I do wonder though if it would make sense to extract the delivery pipeline into a separate project (either a single project for all MSP services or a project per service)
>
> That way it would be easier to alter the stages without having to completely recreate the pipeline if a new stage gets appended.

Maybe, but it seems unlikely you will change the final stage after a service goes live, and this keeps the promotion process in the production service where if an issue happens or a promotion is needed, you're likely in that final stage project already anyways. It hopefully also reduces confusion about what project to look at - the goal of deployments is to eventually end up in the final stage, so you effectively "deploy to the last stage" when targeting the `gcloud deploy releases create` command. WDYT?

unknwon:
> > Is rolling back on Cloud Deploy would effectively "pause" subsequent deployments until some sort of "manual-resume"?
> 
> Unfortunately, it requires a spec change to suspend a pipeline - see https://cloud.google.com/deploy/docs/suspend-pipeline . I can add this toggle to this PR (the pipeline is configured in CDKTF). Alternatively the deploying team would implement the pause on the "push" side

Let's add this toggle and ship this PR, then iterate from here!

unknwon (approved):

PR #59959, opened 2024-01-30
Keyword search: temporarily add feature flag

jtibshirani:
In #59779 we cleaned up the config for keyword search, removing the feature
flag in favor of a user setting. This PR temporarily re-introduces the feature
flag to make it easier to run users tests.

Now, the feature can be enabled simply by adding the URL parameter
`&feat=search-keyword`.

## Test plan

Manual testing

### DIFF
--- a/client/web/src/communitySearchContexts/CommunitySearchContextPage.tsx
+++ b/client/web/src/communitySearchContexts/CommunitySearchContextPage.tsx
@@ -120,7 +120,6 @@ export const CommunitySearchContextPage: React.FunctionComponent<
                     setQueryState={setQueryState}
                     hardCodedSearchContextSpec={props.communitySearchContextMetadata.spec}
                     simpleSearch={false}
-                    showKeywordSearchToggle={false}
                 />
             </div>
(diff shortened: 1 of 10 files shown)


client/web/src/communitySearchContexts/CommunitySearchContextPage.tsx
```diff
@@ -120,6 +120,7 @@ export const CommunitySearchContextPage: React.FunctionComponent<
                     setQueryState={setQueryState}
                     hardCodedSearchContextSpec={props.communitySearchContextMetadata.spec}
                     simpleSearch={false}
+                    showKeywordSearchToggle={false}
```
keegancsmith, line 123: was this a bug before?

keegancsmith (approved):
LGTM

client/web/src/communitySearchContexts/CommunitySearchContextPage.tsx
jtibshirani: No, I just needed to drill the flag through props in some places. I just switched this to an optional boolean to avoid the need to modify this (and some other places).

PR #59970, opened 2024-01-30
trace: Memoize template for trace URL

eseliger:
This showed up in some profiles as being an area of ~high contention, 99% of which is coming from parsing the template string every time. This PR adds some caching to it.

## Test plan

n/a - code review


bobheadxi (approved):
ah nice, thank you!!

PR #59973, opened 2024-01-30
search: Render patterns and filters as "chips" when keyword search is enabled

### LINKED ISSUE #59963: Search input: timeboxed experiment with token highlighting
[Loom](https://www.loom.com/share/3926c3f2b51346b49943b0caad3ed3c9)
[Slack thread](https://sourcegraph.slack.com/archives/C05MHAP318B/p1706228408551399)

fkling:
Closes #59963

This commit makes the following changes:

- When the search pattern type is set to `keyword`, filters and patterns are decorated with a background to more easily identify keywords and sequences of keywords.
- Escaped quotation marks in quoted patterns are decorated differently (same as regex escape sequences). This takes into account the delimiter with which the pattern was quoted (i.e. `"\""` searches for `"` but `'\"'` searches for `\"` and thus `\"` is not an escape sequence).
- Hover tooltips are tweaked to show the real value that a pattern is searching for when it contains escaped delimiters.
- Updates @codemirror/view and @codemirror/state to bring in any potential decoration fixes (might not be necessary).

To make this work correctly I had to extend the `Literal` and `Pattern` token types to contain optional `quotes` and `delimiter` fields, otherwise we wouldn't have access to the delimiter used when decorating tokens.

Note that when decorating tokens we still emit the full pattern token. This was the easiest way to ensure that the hover information would include the full pattern text and not be split up by escape sequences.

https://github.com/sourcegraph/sourcegraph/assets/179026/35d95591-88a3-4936-bcac-fd8555a698b8

## CodeMirror cursor issue

I've intentionally set a very small amount of padding for the chips, because the padding seems to interfere with the cursor position. The cursor is positioned at the end of the chip, i.e. after the padding, not after the last character in the chip. This only happens when there is more content after the chip. The same might apply to the start of the chip but I haven't tested that.
To me this seems like a bug in CodeMirror which I'm happy to file an issue for. In the meantime, having essentially no horizontal padding seems to be an OK workaround.

## TODO

- [ ] Filter chip background color for dark mode (cc @rrhyne, @taiyab)
- [ ] Different chip color for regex patterns?
- [ ] Should we use a different background color for the `context:` filter now that it doesn't stand out anymore?

## Test plan

Manual testing


client/shared/src/search/query/decoratedToken.ts
```diff
@@ -202,6 +203,18 @@ export interface MetaPredicate {
     value: Predicate
 }
 
+enum MetaKeywordKind {
+    EscapedCharacter = 'EscapedCharacter',
+}
```
fkling, line 208: I wanted to keep it simple for now. If we want to decorate more characters (e.g. the delimiters itself like for regex patterns) we can do that later on.

camdencheek:
This looks great to me! During testing, I _thought_ I found an unhandled edge case, but I actually found a misunderstanding of how our query language works. So I guess the highlights are doing their job!

rrhyne:
@fkling I pushed a few changes, feel free to revert. I was finding the chips hard to recognize so I tooled with the colors a bit. I read your note on padding, but I didn't see the issue when I increased it to help with visibility. Not sure why not?

client/branded/src/search-ui/input/experimental/codemirror/syntax-highlighting.ts
```diff
@@ -25,22 +25,26 @@ function getDecorationForToken(token: Token): Decoration | null {
...
     EditorView.baseTheme({
         '.sg-query-token': {
-            borderRadius: '3px',
+            borderRadius: '4px',
             // We only apply little horizontal padding because it appears that
             // the padding interferes with the cursor position (CodeMirror will
             // place the cursor after the padding, not after the last character,
             // which is surprising to the user).
-            padding: '1px',
+            padding: '6px',
```
camdencheek, line 33: Adding this much padding is causing the search input to scroll. I think I see what Felix means about the padding, but it honestly doesn't bother me. It's just a little weird to see things move around when I close quotes or add a space, but I think I'd get used to that.

https://github.com/sourcegraph/sourcegraph/assets/12631702/f7784399-1756-4137-ba6a-40929d6af2b9

camdencheek: I took the liberty of pushing up another commit that adjusts the padding so it doesn't overflow the viewport and so the rounded corners are actually visible (rather than being cut off by the top/bottom of the viewport)

CleanShot 2024-01-30 at 16 04 16@2x

camdencheek (approved):

PR #59974, opened 2024-01-30
search: respect whitespace for phrase boost

stefanhengl:
Terms in phrases might be separated by more than one whitespace. With this change we respect the original query when we construct the phrase query. Before, we always joined terms with " ".  

Additionally we couple this feature to keyword search instead of having it as a feature flag.

Screenshot 2024-01-30 at 14 23 43

Note: The transformation now happens during the planning stage instead of the executing stage. I think this is a more obvious spot as it happens next to other transformations on the query plan.

Test plan:
updated unit test

### DIFF
--- a/internal/search/job/jobutil/job.go
+++ b/internal/search/job/jobutil/job.go
@@ -72,7 +72,11 @@ func NewPlanJob(inputs *search.Inputs, plan query.Plan) (job.Job, error) {
 func NewBasicJob(inputs *search.Inputs, b query.Basic) (job.Job, error) {
 
 	if inputs.Features != nil && inputs.Features.PhraseBoost {
-		b.Pattern = query.ExperimentalPhraseBoost(b.Pattern)
+		newPattern, err := query.ExperimentalPhraseBoost(b.Pattern, inputs.OriginalQuery)
+		if err != nil {
+			return nil, err
+		}
+		b.Pattern = newPattern
 	}
 
 	var children []job.Job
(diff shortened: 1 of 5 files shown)


internal/search/job/jobutil/job.go
```diff
@@ -71,10 +71,6 @@ func NewPlanJob(inputs *search.Inputs, plan query.Plan) (job.Job, error) {
 // NewBasicJob converts a query.Basic into its job tree representation.
 func NewBasicJob(inputs *search.Inputs, b query.Basic) (job.Job, error) {
 
-	if inputs.Features != nil && inputs.Features.PhraseBoost {
```
keegancsmith, line 74: removing feature flag sounds good, I think we are pretty confident this is good.

internal/search/query/transformer_test.go
```diff
@@ -6,30 +6,36 @@ import (
...
 
-		json, _ := ToJSON(query)
+		json, _ := ToJSON(plan.ToQ())
 		return json
 	}
 
 	// expect phrase query
 	autogold.Expect(`[{"or":[{"value":"foo bar bas","negated":false,"labels":null,"range":{"start":{"line":0,"column":0},"end":{"line":0,"column":0}}},{"and":[{"value":"foo","negated":false,"labels":["Literal"],"range":{"start":{"line":0,"column":0},"end":{"line":0,"column":3}}},{"value":"bar","negated":false,"labels":["Literal"],"range":{"start":{"line":0,"column":4},"end":{"line":0,"column":7}}},{"value":"bas","negated":false,"labels":["Literal"],"range":{"start":{"line":0,"column":8},"end":{"line":0,"column":11}}}]}]}]`).Equal(t, test("foo bar bas", SearchTypeKeyword))
 
+	// respect whitespace
```
keegancsmith, line 28: these test cases are pretty hard to parse and understand if they are asserting something correct. Is there maybe a way to restructure the test so as input we have the query and for output we have the extracted phrases?

keegancsmith (approved):
LGTM very minor and ignorable inline comments.

PR #59982, opened 2024-01-30
Fix chicken-and-egg problem in scip-index.sh

keynmol:
1. Fix an issue where `index.scip` symlink was creating issues in... the process that is supposed to produce the destination file for the symlink
2. Remove the container after running, to avoid having lots of lingering containers

Extensive details are added in the script

## Test plan

existing tests cover this


kritzcreek (approved):
LHTM (Looks horrible to me)

The joys of scripting Mac and Linux compatible... Thanks for the explanation in the script

Issue #59698, opened 2024-01-18
SSC: Cody Management Page Fixes & Updates


vdavid:
Designs: https://www.figma.com/file/FMSdn1oKccJRHQPgf7053o/Cody-GA?type=design&node-id=3496-3651&mode=design&t=JcXR4U3zxvfALZxQ-0

Fix: Display subscription renewal / end date on the /cody/manage page

![Image](https://github.com/sourcegraph/sourcegraph/assets/2552265/271f5b8b-018f-456a-8372-14109270d7ea)

Add: don't loose Cody Pro header

![CleanShot 2024-01-29 at 14 18 10@2x](https://github.com/sourcegraph/sourcegraph/assets/22571395/2f63afef-0203-44e4-ace6-acbfe50499bb)

Add: you will be charged header  

![CleanShot 2024-01-29 at 14 16 50@2x](https://github.com/sourcegraph/sourcegraph/assets/22571395/0d58a2fa-5d29-4d1a-8208-efde82241923)

Add the following banner

![CleanShot 2024-01-29 at 14 36 23@2x](https://github.com/sourcegraph/sourcegraph/assets/22571395/6edea0ce-522a-428c-8ef4-c600da474b24)

vdavid:
@thenamankumar, I think your draft PR may tackle this?

PR #59989, opened 2024-01-30
Add Cody Upgrade Banners to Cody Management Page

thenamankumar:
closes: https://github.com/sourcegraph/sourcegraph/issues/59698

I have added a bunch of banners and changes covering all the different user states a user can be:
- Right now, before asking users to put in their cc
- After asking users to put in their cc, but free trial continued. (tentatively after 1st Feb)
- After the trial ends. (after 15th Feb)

Check designs for details: https://www.figma.com/file/FMSdn1oKccJRHQPgf7053o/Cody-GA?type=design&node-id=3496-5821&mode=design&t=rqGpsXz0PDffDRpk-0

Screenshots:

CleanShot 2024-01-30 at 23 12 44@2x
CleanShot 2024-01-30 at 22 48 51@2x
CleanShot 2024-01-30 at 22 47 56@2x
CleanShot 2024-01-30 at 22 44 54@2x

## Test plan

- connect dotcom with SSC locally
- toggle 2 feature flags: `use-ssc-for-cody-subscription` and `cody-pro-trial-ended` to move through various stages.
- Use Cody Management on dotcom and SSC Subscription pages to signup for pro and cancel subscriptions during different stages
- validate that the each state matches the designs on Figma.


client/web/src/cody/subscription/CodySubscriptionPage.tsx
```diff
@@ -50,7 +50,8 @@ export const CodySubscriptionPage: React.FunctionComponent<CodySubscriptionPageP
 
     const utm_source = parameters.get('utm_source')
 
-    const [sscEnabled] = useFeatureFlag('use-ssc-for-cody-subscription', false)
+    const [isPaymentsEnabled] = useFeatureFlag('use-ssc-for-cody-subscription', false)
```
vdavid, line 53: **are**PaymentsEnabled maybe? This looks weird.

client/web/src/cody/subscription/CodySubscriptionPage.tsx
```diff
@@ -191,9 +192,12 @@ export const CodySubscriptionPage: React.FunctionComponent<CodySubscriptionPageP
                                     <H2 className={classNames('text-muted d-inline mb-0', styles.proPricing)}>$9</H2>
                                     <Text className="mb-0 text-muted d-inline">/month</Text>
                                 </div>
-                                <Text className="mb-3 text-muted" size="small">
-                                    Free until Feb 2024, <strong>no credit card needed</strong>
-                                </Text>
+                                {!isTrialEnded && (
+                                    <Text className="mb-3 text-muted" size="small">
+                                        Free until Feb 2024,{' '}
```
vdavid, line 197: Feb 15?

client/web/src/cody/management/CodyManagementPage.tsx
```diff
@@ -55,6 +58,9 @@ export const CodyManagementPage: React.FunctionComponent<CodyManagementPageProps
 
     const utm_source = parameters.get('utm_source')
 
+    const [isPaymentsEnabled] = useFeatureFlag('use-ssc-for-cody-subscription', false)
```
chrsmith, line 61: Do you think it would be a net improvement to put the string literal here into a package level constant, with a doc comment explaining what it controls? This seems a little too magical IMHO. But if you disagree feel free to keep as-is.

client/web/src/cody/management/CodyManagementPage.tsx
```diff
@@ -96,11 +102,13 @@ export const CodyManagementPage: React.FunctionComponent<CodyManagementPageProps
...
     }
 
+    subscription = { ...subscription, cancelAtPeriodEnd: true }
+
     const codeLimitReached = codyCurrentPeriodCodeUsage >= codyCurrentPeriodCodeLimit && codyCurrentPeriodCodeLimit > 0
     const chatLimitReached = codyCurrentPeriodChatUsage >= codyCurrentPeriodChatLimit && codyCurrentPeriodChatLimit > 0
-    const userIsOnProTier = subscription.plan === CodySubscriptionPlan.PRO
-
-    const showUpgradeBanner = !userIsOnProTier && (codeLimitReached || chatLimitReached)
+    const userIsOnProTier =
```
chrsmith, line 109: Is this right? It seems like we shouldn't have `cancelAtPeriodEnd` be used as part of our logic for determining if `userIsOnProTier`.

After 2/15, if I sign up on SCC and create an active Cody Pro subscription. But then hit the cancel button so my active subscription will be canceled at the end of the billing cycle... and at that time, we would start showing this `!userIsOnProTer` "upgrade" banner. Even though the user currently has an active Cody Pro subscription.

Right? Or am I just confused here.

Maybe it would be helpful if we kept adding new boolean values? e.g. `const showUpgradeToProBanner = ...`? WDYT?

client/web/src/cody/management/CodyManagementPage.tsx
```diff
@@ -114,50 +122,134 @@ export const CodyManagementPage: React.FunctionComponent<CodyManagementPageProps
...
+                                }
+                                variant="primary"
+                                size="sm"
+                            >
+                                Upgrade
+                            </ButtonLink>
+                        </div>
+                    </div>
+                )}
+                {isPaymentsEnabled && userIsOnProTier && subscription.status === CodySubscriptionStatus.PENDING && (
```
chrsmith, line 157: These banners are great, but IMO this code is really hard to read. Since we are mixing non-trivial logic with the markup and then with the specifics of the type of banner message / call-to-action.

Can we clean this up by introducing a separate function or component? e.g.

```ts
type CodyProBanner = 'upgrade-to-pro' | 'trial-ending-add-cc' | undefined

function getBannerToDisplay(subscription): CodyProBanner { ... }
function renderBanner(banner, subscription, flags) { ... }
```

Yes there is more "code", but IMHO it'll make it much everything much easier to read.

client/web/src/cody/subscription/CodySubscriptionPage.tsx
chrsmith: +1 to the naming being kinda awkward. Perhaps `paymentOnSccEnabled`?

client/web/src/cody/subscription/CodySubscriptionPage.tsx
chrsmith: The wording we have on the "upgrade to pro" dialog is vague about this, literally saying "Free until February". I'd be inclined to keep this as-is and just say "until Feb" even though that isn't precise and/or doesn't even make sense. "until [sometime during the month of] February" doesn't sound right either. 🙃

cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -6417,6 +6417,10 @@ type CodySubscription {
     subscription is in a canceled state.
     """
     currentPeriodEndAt: DateTime!
+    """
+    Whether the subscrition ends on curren billing period end.
```
chrsmith, line 6421: Typeos, clarifications:

```suggestion
    Whether the subscription will be canceled at the end of the current billing period. (Or automatically renew, if true.)
```

chrsmith (commented):
First off, those banners look beautiful. 👏👏👏

Some suggestions to improve readability and things, but I'll defer to David about React idioms and things.

I'm a little confused by the code, but perhaps you are just making an assumption specific to how the existing free trials are handled. (Where SSC does NOT have a subscription for the user, but the dotcom backend DOES have `codyEnabledAt !== undefined`.)

FWIW, this is how the SSC backend works today. (But if this behavior is going to cause problems, we can change it.)

- If you go to the SSC frontend today, we will create a new subscription AND have it be on a free trial until 2/15.
- Once you have a subscription in the SSC frontend, you are free to toggle the `cancelAtPeriodEnd` setting.

So today, you can have a subscription in SSC that is both `trialing` and has `cancelAtPeriodEnd=true`. (Meaning, that you will not convert at the end of your trial.)

Or, after your trial ends, your subscription may be `active` but you can then set `cancelAtPeriodEnd=true`. (Meaning that at the end of the current billing period, your subscription will not renew.)

Does that match your mental model for how/when `cancelAtPeriodEnd` would be set? Should we just _never_ offer the ability to receive a free trial if you show up to the SSC frontend? (Since it would only apply to users who go to SSC in the next X days...)

client/web/src/cody/subscription/CodySubscriptionPage.tsx
thenamankumar: Keep the old stuff as it is.

client/web/src/cody/management/CodyManagementPage.tsx
thenamankumar: The logic is correct. We discussed this particular logic 2 days back as well. We are checking the cancelAtPeriodEnd only for the Trialing state. Suppose the user is on a Trialing subscription and cancels the subscription. On SSC, it doesn't get immediately cancelled. But we need it to be cancelled immediately. So this logic covers that. 

If later on SSC, we update the behaviour to match the requirements, this logic will still not hurt, but we can remove it then.

client/web/src/cody/subscription/CodySubscriptionPage.tsx
thenamankumar: Clients don't care or know about SSC. `arePaymentsEnabled` is in my opinion a better name, and I am updating it.

client/web/src/cody/management/CodyManagementPage.tsx
thenamankumar: The feature flag names are type-safe and strict. So you can not actually put a wrong one without getting the linter to go crazy. But I have extracted them into react hooks into a separate file.

client/web/src/cody/management/CodyManagementPage.tsx
```diff
@@ -548,3 +623,62 @@ const DashboardIcon = ({ className }: { className?: string }): ReactElement => (
         </defs>
     </svg>
 )
+
+const CodyProIcon = ({ className }: { className: string }): ReactElement => (
```
vdavid, line 627: I'd extract the icon to a separate file.

vdavid (approved):
Looks good to me, left a few comments

client/web/src/cody/management/CodyManagementPage.tsx
thenamankumar: `renderBanner` or anything `renderX` is an empty pattern. There used to be an era of class based react components < V15 when this pattern was used. But now is the time of simple functional components. 

I have extracted all the banners into separate components.

vdavid: That's great, however, these banners are temporary, right? So not sure how much effort it justifies.

thenamankumar:
Addressed all the feedback, enabling auto merge.

PR #59993, opened 2024-01-30
Search: Add telemetry for keyword and regexp toggles

jtibshirani:
This adds a telemetry event to track if users hit the keyword search and regexp
toggles. This will help us understand
1. If users are often frustrated/ surprised with the new search syntax
2. Whether we could potentially remove the regexp toggle in a future release

## Test plan

Manual testing with the `eventLogDebug` setting


jtibshirani:
Ideally, I would have used our new (non-deprecated) `TelemetryRecorder` from our v2 telemetry package. I tried this, but it resulted in a huge amount of prop drilling all over the search code. I think it'd be better to open a dedicated PR to switch to the new telemetry style for the search frontend.

keegancsmith (approved):

Issue #59964, opened 2024-01-30
Lack of container padding for aggregation results


taiyab:
<img width="832" alt="image" src="https://github.com/sourcegraph/sourcegraph/assets/5337876/83c304eb-cd12-49cc-a3e5-ee7265b0d31d">

We should probably add some padding back in for the aggregation stuff.

PR #60003, opened 2024-01-30
Search results: design tweaks

camdencheek:
This implements a few design tweaks from pairing with @taiyab this morning.

Numbered in the screenshot below:
1) Remove the background color from the search results header so hovering over a button actually changes the color. Previously, the hover color was the same as the background color, so hovering did nothing. (@taiyab, I chose not to implement the changes with removing borders and changing background colors since this also affects the other buttons and it had unexpected downstream effects). 
2) Use the `--marketing-gradient-25` for the keyword search popover in dark mode so it doesn't look broken in dark mode (cc @rrhyne)
3) De-emphasize the count badges
4) Fix some spacing between the type filters and the subfilters. Also, move the repositories selector to the top of the filters panel since it's likely to be the most used, most useful, and hierarchically the top level filter type.
5) Fix the spacing of the "Group results by" message

Fixes https://github.com/sourcegraph/sourcegraph/issues/59964

Dark mode:

CleanShot 2024-01-30 at 17 57 39@2x

Light mode:

CleanShot 2024-01-30 at 17 58 50@2x

Group by spacing:

CleanShot 2024-01-30 at 18 20 52@2x

## Test plan

Visual testing.

### DIFF
--- a/client/branded/src/search-ui/components/ResultContainer.module.scss
+++ b/client/branded/src/search-ui/components/ResultContainer.module.scss
@@ -17,14 +17,7 @@
     z-index: 1; // Show on top of search result contents
 
     border-bottom: 1px solid var(--border-color-2);
-
-    // TODO(taiyab): this needs to be fit into the design system
-    :global(.theme-light) & {
-        background-color: #f5f8fa;
-    }
-    :global(.theme-dark) & {
-        background-color: #22283b;
-    }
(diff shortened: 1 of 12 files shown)


BolajiOlajide (approved):

taiyab (commented):
Looks good and happy not removing the borders and stuff as we're getting a little too hacky probably too.

PR #54633, opened 2023-07-05
codeintel: Add symbols method to syntect client

efritz:
Pulled from #51690. This adds a new client method we'll use to provide precise Cody context.

## Test plan

Tested locally.

PR #60008, opened 2024-01-31
Revert "codeintel: Add symbols method to syntect client (#54633)"

varungandhi-src:
This reverts commit a2e6ff458722558a7c2d71d631009c6e8f448a1b.

This is not being used by Cody, so deleting it.

It also doesn't make much sense from a layering perspective anyways;
precise context information doesn't have anything to do with
syntax highlighting.

## Test plan

It's unused and doesn't have any tests


kritzcreek (commented):
This looks good, but do we know if this is the only place the "/symbols" endpoint is called on our syntect_server? (If so we could delete some more code 🤩)

kritzcreek (approved):

varungandhi-src:
> but do we know if this is the only place the "/symbols" endpoint is called on our syntect_server? (If so we could delete some more code 🤩)

All code calling to syntect_server should go through gosyntect, so yes. I've deleted the endpoint code in Rust too.

Issue #60012, opened 2024-01-31
Hook ssc.SAMSProdHostname  with site config to enable testing


thenamankumar:
ask from @chrsmith : https://sourcegraph.slack.com/archives/C05PC7AKFQV/p1706670137824999

PR #60013, opened 2024-01-31
Make SSC SAMS host name configurable

thenamankumar:
closes: https://github.com/sourcegraph/sourcegraph/issues/60012

To enable testing, we need to make the SAMS hostname configuration via site config, so that we can connect dotcom staging to sams dev. 

## Test plan
unit tests cover the changes.

### DIFF
--- a/internal/ssc/ssc.go
+++ b/internal/ssc/ssc.go
@@ -117,6 +117,16 @@ func (c *client) FetchSubscriptionBySAMSAccountID(ctx context.Context, samsAccou
 	}
 }
 
+func GetSAMSHostName() string {
+	sgconf := conf.Get().SiteConfig()
+
+	if sgconf.SscSamsHostName == "" {
+		return SAMSProdHostname
+	}
+
+	return sgconf.SscSamsHostName // [sic] generated code
+}
+
 // NewClient returns a new SSC API client. It is important to avoid creating new
 // API clients if possible, so that it can reuse SAMS access tokens when making
 // requests to SSC. (Otherwise every request would need to create a new token,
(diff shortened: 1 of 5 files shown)


internal/ssc/ssc.go
```diff
@@ -117,6 +117,16 @@ func (c *client) FetchSubscriptionBySAMSAccountID(ctx context.Context, samsAccou
...
 }
 
+func GetSAMSHostName() string {
+	sgconf := conf.Get().SiteConfig()
+
+	if sgconf.SscSamsHostName == "" {
+		return SAMSProdHostname
+	}
+
+	return sgconf.SscSamsHostName // [sic] generated code
```
vdavid, line 127: What's "[sic] generated code"?

schema/site.schema.json
```diff
@@ -12,10 +12,10 @@
       "default": "https://accounts.sourcegraph.com/cody/api",
       "description": "The base URL of the Self-Serve Cody API."
     },
-    "ssc.apiSecret": {
+    "ssc.samsHostName": {
```
vdavid, line 15: Just to confirm, `ssc.apiSecret` was unused, right? (I didn't find any uses.) Might be worth mentioning in the PR description that you've removed it, just for clarity.

vdavid (approved):
Nice! Left a nit

schema/site.schema.json
thenamankumar: yeah it is unused given we migrated to service to serve SAMS token based auth.

internal/ssc/ssc.go
chrsmith: The name `SscSamsHostName` looks WAAAAY out of place because of how standardized Golang naming conventions are. So I believe I added a "[sic] generated code" comment to call out that this isn't something we can control, and it's not a mistake so to speak.

chrsmith:
Post-hoc LGTM! :shipit: 🚀

internal/ssc/ssc.go
vdavid: Ah, gotcha, makes sense.

PR #60014, opened 2024-01-31
Implement license manager RBAC role

eseliger:


cmd/frontend/graphqlbackend/users.go
```diff
@@ -24,9 +25,15 @@ type usersArgs struct {
 }
 
 func (r *schemaResolver) Users(ctx context.Context, args *usersArgs) (*userConnectionResolver, error) {
-	// 🚨 SECURITY: Verify listing users is allowed.
-	if err := checkMembersAccess(ctx, r.db); err != nil {
-		return nil, err
+	// 🚨 SECURITY: Only license manager and site admins can list users on sourcegraph.com.
+	if envvar.SourcegraphDotComMode() {
+		// First, check if they have the rbac permission for license manager.
+		if err := rbac.CheckCurrentUserHasPermission(ctx, r.db, rbac.LicenseManagerReadPermission); err != nil {
```
BolajiOlajide, line 31: Site Admin roles usually have all permissions assigned to them, so if the user is a site admin then they'll already have the permission required.

PR #60017, opened 2024-01-31
honeycomb: remove database-batch from emitting to honeycomb

Strum355:
Doesn't appear to be used by anyone anymore, so we can save some quota

## Test plan

N/A


eseliger (approved):
Okeh

PR #60021, opened 2024-01-31
search: revert to single space phrase boosting

stefanhengl:
Relates to #59974

We revert the logic to a much simpler join by " ".

During manual testing we noticed that queries that include reserved keywords, such as `(foo and bar) and bas` led to undesired phrase queries. For the case of `(foo and bar) and bas` we would search the phrase `foo and bar) and bas`. In hindsight this is an obvious flaw of using the character range of the terms from the original query. 

I also fix another bug which was caused by missing parser flags. Essentially, the missing flags led to parser errors if specials characters were used in the phrase.

Test plan:
- updated unit test
- I tested manually queries such "* func = Function" or "(foo and bar) and bas" and confirmed the results are as expected.

### DIFF
--- a/internal/search/query/transformer.go
+++ b/internal/search/query/transformer.go
@@ -644,7 +644,7 @@ func ExperimentalPhraseBoost(logger sglog.Logger, originalQuery string) BasicPas
 			phrase := ""
 			for _, child := range n.Operands {
 				c, isPattern := child.(Pattern)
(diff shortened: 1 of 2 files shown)


jtibshirani (approved):

jtibshirani:
@stefanhengl requested I merge this before feature freeze ... doing that now!

PR #60022, opened 2024-01-31
Keyword search: add feedback modal

jtibshirani:
This PR adds a "Send feedback" button to the keyword search popover that pulls
up our feedback modal. It prefixes the feedback with "[Source: keyword search]"
so we know it's about keyword search.

## Test plan

Manual testing

### DIFF
--- a/client/web/src/search/results/components/search-results-info-bar/SearchResultsInfoBar.tsx
+++ b/client/web/src/search/results/components/search-results-info-bar/SearchResultsInfoBar.tsx
@@ -2,7 +2,7 @@ import { FC, useCallback, useMemo, useRef, useState } from 'react'
 
 import { mdiChevronDoubleDown, mdiChevronDoubleUp, mdiOpenInNew, mdiThumbDown, mdiThumbUp } from '@mdi/js'
 import classNames from 'classnames'
-import { useLocation, useNavigate } from 'react-router-dom'
+import { useLocation, useMatches, useNavigate } from 'react-router-dom'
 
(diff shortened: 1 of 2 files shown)


jtibshirani:
Here's how it looks. I would love help with the design here... maybe it should be a link instead of a button?
Screenshot 2024-01-31 at 12 28 41 PM
Screenshot 2024-01-31 at 12 28 51 PM

client/web/src/search/results/components/search-results-info-bar/SearchResultsInfoBar.tsx
```diff
@@ -210,13 +212,37 @@ export const SearchResultsInfoBar: FC<SearchResultsInfoBarProps> = props => {
         onTogglePatternType(patternType)
     }, [onTogglePatternType, patternType, telemetryService])
 
+    const [feedbackModalOpen, setFeedbackModalOpen] = useState(false)
+
+    const routeMatches = useMatches()
+    const { handleSubmitFeedback } = useHandleSubmitFeedback({
```
jtibshirani, line 218: I based the route matching logic on this: https://github.com/sourcegraph/sourcegraph/blob/51960eee012c0b5593bab8b123ebe7452fcc4488/client/web/src/storm/pages/LayoutPage/LayoutPage.tsx#L112-L114

jtibshirani:
Here's how it looks ... very open to design suggestions :)
Screenshot 2024-01-31 at 3 17 20 PM

stefanhengl (approved):
I tested it locally and the feedback is received on our end. LGTM!

PR #60024, opened 2024-01-31
Filters Panel: Add reset all filters button

jasonhawkharris:
This adds a button that resets all chosen filters currently selected in the filters panel. It also adds a keybind (option/alt + backspace) to clear all filters. 

https://www.loom.com/share/cf5539324bdb436c9f97c8768e08fb7c?sid=91f06d7c-22f4-4bfd-b535-449b964c69fe

## Test plan
Test manually. CI/CD passing


client/branded/src/search-ui/results/filters/NewSearchFilters.tsx
```diff
@@ -170,15 +171,27 @@ export const NewSearchFilters: FC<NewSearchFiltersProps> = ({
...
+                                <Icon as={ArrowBendIcon} aria-hidden={true} className={styles.moveIcon} />
+                            </Button>
+                        </Tooltip>
+                        <Button variant="link" onClick={() => setSelectedFilters([])}>
+                            <Icon
+                                size="md"
+                                className="pr-1 pt-1"
+                                svgPath={mdiRefresh}
+                                inline={true}
+                                aria-label="Reset all filters"
```
camdencheek, line 190: In this case, I think we should use `aria-hidden={true}` since the purpose of the button exists in the label itself, so the icon is just visual sparkle.

vovakulikov:
@taiyab I really don't know about link-like button. I'm biased here because I always don't like this variation, it looks like a link but it doesn't do anything with URL and it confuses me because of this. Visually it looks a bit off as well since button above has visible border area and this button doesn't. Maybe we could use here outline secondary again. (I think button link was a mistake in our wildcard :) )

Screenshot 2024-01-31 at 18 12 49

camdencheek (commented):
The code itself looks good, but let's take a pass at the design. Seconding Vova's comments about the link-like button, and also it takes up quite a bit of valuable vertical space for what I expect is a pretty rarely-used action.

CleanShot 2024-01-31 at 14 15 51@2x

jasonhawkharris:
@taiyab @camdencheek mentioned the possibility of maybe having the buttons inline and with much less information. Maybe even just the icons. Thoughts?

taiyab:
>@taiyab I really don't know about link-like button. I'm biased here because I always don't like this variation, it looks like a link but it doesn't do anything with URL and it confuses me because of this. Visually it looks a bit off as well since button above has visible border area and this button doesn't. Maybe we could use here outline secondary again. (I think button link was a mistake in our wildcard :) )

I don't think the link button is a mistake at all, the problem here is that we're really missing another variant I'd like which is a "ghost" button.

A ghost button essentially visually fills the button space on hover with a (in this case) light gray but with no border.

Before hover:
image

During hover:
image

If we can implement something like this, it would work IMO.

> Seconding Vova's comments about the link-like button, and also it takes up quite a bit of valuable vertical space for what I expect is a pretty rarely-used action.

I don't necessarily mind the vertical space taken here (with a bit of tweaking). We could also only have it appear once more than one filter is applied.

For reference, my original treatment had this as a small option inline with a "Filter results" title, however, we probably (rightfully) stripped that out:
image

Alternatively, I'm happy to just delete it. It's not providing *that much* value in a world where you can only apply like 3-4 filters via the panel anyway.

jasonhawkharris:
Second pass over the design, and added shortcut for clearing filters.

Uploading Screen Recording 2024-02-02 at 1.29.52 PM.mov…

client/wildcard/src/hooks/useOperatingSystem.ts
```diff
@@ -0,0 +1,21 @@
...
+            if (navigator.userAgent.indexOf('Win') !== -1) {
+                setOs('Windows');
+            } else if (navigator.userAgent.indexOf('Mac') !== -1) {
+                setOs('MacOS');
+            } else if (navigator.userAgent.indexOf('Linux') !== -1) {
+                setOs('Linux');
+            }
+        };
+
+        detectOS();
```
vovakulikov, line 17: why do you need to have a function here? Can you just run this function body in the useEffect directly?

vovakulikov (commented):
@jasonhawkharris a few things that I noticed 
- Since @camdencheek PR about making type filter real filters has been merged we should move filter results title and reset button on the top of the filters panel 

Screenshot 2024-02-02 at 18 09 35

- I think we should also add keybinding since we have `kbd` icon with keyboard combination to be able to reset filters from keyboard
- About button, again feel free to ignore here but I think that we should intro this ghost variation button with hover state and also `kbd` icon should be within this button so if I click it filter will be reset (now you have to aim to the reset filter text, which is link)

client/wildcard/src/hooks/useOperatingSystem.ts
jasonhawkharris: I like reading this line because it's clear what it does:

`const os = useOperatingSystem`

I feel like it's easier to immediately understand than the alternative. I've also found myself needing this info before, so I thought it'd be good to make it a reusable hook.

camdencheek: Why do we need this to be a hook or stateful at all? Couldn't this just be a function that returns the operating system?

client/wildcard/src/components/KbdBadge/KbdBadge.tsx
```diff
@@ -0,0 +1,64 @@
...
+    shortCut: Keybind
+    shadow?: boolean
+    className?: string
+    partnerKeyIsIcon: boolean
+}
+
+/**
+ * KbdBadge Element
+ */
+export const KbdBadge = React.forwardRef(function KbdBadge({
```
camdencheek, line 25: Rather than creating a custom component, we should probably use the [`<kbd>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd) element. That will be more accessible, have better default styling, and will likely require a lot less code

client/wildcard/src/hooks/useOperatingSystem.ts
jasonhawkharris: ahhh... sorry I misunderstood. I see what you're saying.

client/wildcard/src/components/KbdBadge/KbdBadge.tsx
jasonhawkharris: So, @vovakulikov and I actually chatted about this. Personally, I think the <kbd> element doesn't look great. At first, I wanted to introduce a new badge variant (kbd), but that was handling too much logic, so we came to the conclusion that a new wildcard component would be good for this. I've also found myself needing a better kbd element to introduce keymaps and this solves that problem.

taiyab:
@vovakulikov Thanks for clarifying on the other related changes.

I think once those are in place we should be good to go.

We should make sure to test the keyboard shortcut button (and the relevant icons) on Windows/Mac, etc.

taiyab (commented):
@vovakulikov Thanks for clarifying on the other related changes.

I think once those are in place we should be good to go.

We should make sure to test the keyboard shortcut button (and the relevant icons) on Windows/Mac, etc.

client/wildcard/src/components/KbdBadge/KbdBadge.tsx
vovakulikov: I actually missed the fact that this component doesn't use the `kbd` tag.

 I mean I agree that a separate component is okay here because we have to deal with a global `.kbd` CSS class. And we don't have time to migrate all other kbd UI in the Sourcegraph (probably), So we could do a few things here
 - Have a wildcard component for kbd and still use `kbd` element in this component implementation (as @camdencheek said this should help use with a11y) and then fix styling in this component styles. 
 - If for some reason we can't use `kbd` element in this component we need to make sure that this doesn't different in any way from real kbd (maybe we have to sue aria labels and types there but not sure, never tried to listen kbd elements in VoiceOver really)

client/wildcard/src/components/KbdBadge/KbdBadge.tsx
```diff
@@ -0,0 +1,51 @@
...
+                {shortCutKeys[0]} {shortCutKeys[1]}
+            </div>
+        </Component>
+    )
+}) as ForwardReferenceComponent<'span', KbdBadgeProps>
+
+const createKbd = (kb: Keybind): string[] => {
+    let shortcut = [kb.modifier]
+    let keys = ''
+    if (isArray(kb.selector)) {
```
camdencheek, line 41: nit: instead of using lodash, we can just use builtins

```suggestion
    if (kb.selector instanceof Array) {
```

client/wildcard/src/components/KbdBadge/KbdBadge.tsx
jasonhawkharris: @vovakulikov @camdencheek After working with it over the past hour, I think we should irradicate the <kbd> element from the code base. It's difficult to style, the default padding/margins are way off. In general, I think it's just a frustrating element to work on and I think this wildcard component can make all our jobs a lot easier in the future when trying to style these buttons. That's my two cents. I'm happy to do what you all want to do, but just wanted to give my input.

client/wildcard/src/components/KbdBadge/KbdBadge.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { isArray } from 'lodash'
+
+import { type ForwardReferenceComponent } from '../..'
+
+import styles from './KbdBadge.module.scss'
+
+export interface Keybind {
+    modifier: string
+    selector: string[] | string
+}
```
camdencheek, line 13: Why separate modifier and selector? Sometimes we can have multiple modifiers (cmd+ctrl is fairly common). I also think we can simplify the formatting significantly if this type just becomes `[]string`.

client/wildcard/src/components/KbdBadge/KbdBadge.tsx
```diff
@@ -0,0 +1,52 @@
...
+ */
+export const KbdBadge = React.forwardRef(function KbdBadge(
+    { shortCut, className, onClick, as: Component = 'span', ...attributes },
+    reference
+) {
+    const shortCutKeys = createKbd(shortCut)
+
+    return (
+        <Component className={classNames(className)} ref={reference} {...attributes}>
+            <div className={classNames(styles.keybind, onClick && styles.withCursor)} onClick={onClick}>
```
camdencheek, line 32: What happens if `onClick` is undefined? Wouldn't that evaluate to a class name `false`?

client/wildcard/src/components/KbdBadge/KbdBadge.tsx
```diff
@@ -0,0 +1,52 @@
...
+                {shortCutKeys[0]} {shortCutKeys[1]}
+            </div>
+        </Component >
+    )
+}) as ForwardReferenceComponent<'span', KbdBadgeProps>
```
camdencheek, line 37: I don't think we should be casting here. Instead, we should just `export const KbdBadge: ForwardReferenceComponent<'span', KbdBadgeProps>` above so the type system catches any issues.

camdencheek (changes requested):

client/wildcard/src/hooks/useOperatingSystem.ts
```diff
@@ -0,0 +1,17 @@
...
+            setOs('Windows')
+        } else if (navigator.userAgent.indexOf('Mac') !== -1) {
+            setOs('MacOS')
+        } else if (navigator.userAgent.indexOf('Linux') !== -1) {
+            setOs('Linux')
+        }
+    }, [])
+
+    return os
+}
```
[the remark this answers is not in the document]
jasonhawkharris: @camdencheek If this were the case we'd have to account for the value being undefined on mount... isn't it better to return a hook so that the component updates when it determines the OS?

client/wildcard/src/components/KbdBadge/KbdBadge.tsx
jasonhawkharris: Wouldn't it return nothing at all? Since onClick would evaluate to undefined?

client/wildcard/src/components/KbdBadge/KbdBadge.tsx
camdencheek: There's also `satisfies` which could be a good option here as well

client/wildcard/src/components/KbdBadge/KbdBadge.tsx
camdencheek: Maybe? But then the question is just how does `classNames` handle `undefined`? Have you tested to see if this works with no click handler?

camdencheek (approved):
Looks great! Thanks for iterating with me

client/branded/src/search-ui/results/filters/NewSearchFilters.module.scss
```diff
@@ -69,3 +69,29 @@
...
+    display: flex;
+    flex-flow: row nowrap;
+    justify-content: space-between;
+    align-content: center;
+    margin-left: 0.25rem;
+    margin-top: 0.75rem;
+}
+
+.reset-button {
+    color: var(--oc-gray-6);
```
taiyab, line 83: We can't use static colors here. We can try `--color-bg-1` here instead.

client/branded/src/search-ui/results/filters/NewSearchFilters.module.scss
```diff
@@ -69,3 +69,29 @@
...
+    margin-right: 0.5rem;
+}
+
+// @TODO: Jason Harris: <kbd> element is unuseable unless the following CSS selectors are present.
+// We should use these as a base for adjusting the default styling of <kbd> elements.
+.keybind {
+    font-family: var(--font-family-base);
+    margin-left: 0.25rem;
+    line-height: 1rem;
+    color: var(--gray-06);
```
taiyab, line 96: Similarly here, we should try `--text-muted`.

taiyab (commented):
LGTM! Nice work @jasonhawkharris.

client/branded/src/search-ui/results/filters/NewSearchFilters.tsx
```diff
@@ -97,15 +118,45 @@ export const NewSearchFilters: FC<NewSearchFiltersProps> = ({
         telemetryService.log('SearchFiltersApplyFiltersClick')
     }
 
+    const handleKeyDown = useCallback(
```
vovakulikov, line 121: @jasonhawkharris Usually all functions that are used only within `useEffect` stay inside useEffect. 
Something like this in this case 

```ts
 useEffect(() => {
        function handleKeyDown(e: KeyboardEvent): void {
            if (e.altKey && e.key === 'Backspace') {
                setSelectedFilters([])
            }
        } 
        
        document.addEventListener('keydown', handleKeyDown)
        
        return () => {
            document.removeEventListener('keydown', handleKeyDown)
        }
    }, [setSelectedFilters])
```

client/branded/src/search-ui/results/filters/NewSearchFilters.tsx
```diff
@@ -36,6 +39,22 @@ interface NewSearchFiltersProps extends TelemetryProps {
...
+
+    if (userAgent.includes('Mac')) {
+        return 'MacOS'
+    }
+
+    if (userAgent.includes('Linux')) {
+        return 'Linux'
+    }
+
+    return undefined
```
vovakulikov, line 55: There is no need for return undefined, just `return` is enough since `return` without value after returns `undefined` already

PR #60028, opened 2024-01-31
Search filters: do not select a type by default

camdencheek:
This modifies the types section of the dynamic filters sidebar to not have any type selected by default. This is to better represent the default search behavior where we return code, repositories, and paths in the result set.

[Loom walkthrough](https://www.loom.com/share/947ef0784d0245398839e326e30adc35)

## Test plan

Manual testing that the behavior feels correct. Updated backend tests with new `type:` dynamic filters

### DIFF
--- a/client/branded/src/search-ui/results/filters/NewSearchFilters.tsx
+++ b/client/branded/src/search-ui/results/filters/NewSearchFilters.tsx
@@ -4,7 +4,7 @@ import { FilterType, resolveFilter } from '@sourcegraph/shared/src/search/query/
 import { findFilters } from '@sourcegraph/shared/src/search/query/query'
 import { scanSearchQuery, succeedScan } from '@sourcegraph/shared/src/search/query/scanner'
 import type { Filter as QueryFilter } from '@sourcegraph/shared/src/search/query/token'
-import { omitFilter, updateFilter } from '@sourcegraph/shared/src/search/query/transformer'
+import { omitFilter } from '@sourcegraph/shared/src/search/query/transformer'
 import type { Filter } from '@sourcegraph/shared/src/search/stream'
 import { TelemetryProps } from '@sourcegraph/shared/src/telemetry/telemetryService'
 import { Button, Icon, Tooltip } from '@sourcegraph/wildcard'
@@ -18,15 +18,11 @@ import {
     symbolFilter,
     utilityFilter,
 } from './components/dynamic-filter/SearchDynamicFilter'
-import {
-    FilterTypeList,
-    resolveFilterTypeValue,
-    toSearchSyntaxTypeFilter,
-} from './components/filter-type-list/FilterTypeList'
+import { FilterTypeList } from './components/filter-type-list/FilterTypeList'
 import { FiltersDocFooter } from './components/filters-doc-footer/FiltersDocFooter'
 import { ArrowBendIcon } from './components/Icons'
 import { mergeQueryAndFilters, URLQueryFilter, useUrlFilters } from './hooks'
-import { FiltersType, SEARCH_TYPES_TO_FILTER_TYPES, SearchFilterType } from './types'
+import { FilterKind } from './types'
 
 import styles from './NewSearchFilters.module.scss'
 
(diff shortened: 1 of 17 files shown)


client/branded/src/search-ui/results/filters/NewSearchFilters.tsx
```diff
@@ -48,36 +44,29 @@ export const NewSearchFilters: FC<NewSearchFiltersProps> = ({
...
-        if (tokens.type === 'success') {
-            const filters = tokens.term.filter((token): token is QueryFilter => token.type === 'filter')
-            const typeFilters = filters.filter(filter => resolveFilter(filter.field.value)?.type === 'type')
-
-            if (typeFilters.length === 0 || typeFilters.length > 1) {
-                return SearchFilterType.Code
-            }
-
-            return resolveFilterTypeValue(typeFilters[0].value?.value)
+    useEffect(() => {
```
vovakulikov, line 47: Can it be also a calculation-like (with useMemo) logic or we need to react on their changes as we do it here in useEffect. I always prefer useMemo and useEffect only when it's absolutely needed (to avoid possible effect races)

client/branded/src/search-ui/results/filters/NewSearchFilters.tsx
```diff
@@ -48,36 +44,29 @@ export const NewSearchFilters: FC<NewSearchFiltersProps> = ({
...
-
-            // Replace: true is needed here to avoid populating history with
-            // extra entries with completely internal locations update,
-            // Setting filters shouldn't be in the history since onQueryChange
-            // changes URL itself.
-            onQueryChange(newQuery, serializeFiltersURL(newSelectedFilters))
-            telemetryService.log('SearchFiltersTypeClick', { filterType }, { filterType })
+    }, [selectedFilters, query, setSelectedFilters])
+
+    const onClickFilterType = useCallback(
```
vovakulikov, line 53: Nit: This is my personal policy so feel free to ignore this. I try to follow the next pattern when I name props and their handlers 
- `on[handlerName]` is how we name prop callback in some component (like this ` onClick` in the `FilterTypeList`)
- `handle[EventName]` is how we name callback in consumer for this on-Prop, so here it would be `handleFilterTypeClick`

client/branded/src/search-ui/results/filters/NewSearchFilters.tsx
```diff
@@ -257,33 +259,11 @@ const SyntheticCountFilter: FC<SyntheticCountFilterProps> = props => {
...
-        default: {
-            const filters = findFilters(succeedScan(query), FilterType.type)
-            const newQuery = filters.reduce((query, filter) => omitFilter(query, filter), query)
-
-            return updateFilter(newQuery, FilterType.type, toSearchSyntaxTypeFilter(searchType))
-        }
-    }
-}
-
-function omitImpossibleFilters(filters: URLQueryFilter[], searchType: SearchFilterType): URLQueryFilter[] {
```
vovakulikov, line 285: @camdencheek I think removing this one will cause a regression about preserving filters as you go through different filter types. Try to go to File type, set some language filter and then go to the Repositories filter type. 
Prior to this lang filter will be reset in order to avoid empty state (since repo filter type can't have lang filters)

I checked it manually in this PR and I think this is the case here

vovakulikov (changes requested):
Left a few comments and one question about omitting filters

client/branded/src/search-ui/results/filters/NewSearchFilters.tsx
camdencheek: Aaah shoot. Good catch. Fixed in the latest commit -- thank you!

taiyab:
Thanks for these changes @camdencheek, great stuff! :)

A few important things from my perspective:

- **I'm very much happy with treating the type list differently.** When originally designing this, I almost saw those at the top-level "categories" that you could use to switch between types, whilst everything below was an actual filter in the traditional sense (hence even adding the "Filter by" title was below the types in design, along with the reset button). Even the type list was treated visually slightly different in that everything could be slightly bigger, to indicate they're the parent categories that the filters below apply to. It almost meant when you hit reset all, it didn't reset the type, but did reset everything else below.

- Overall, I would want this to default to `Code` (which I know that's not how it works today so it's fine it's like this for now). We should continue to work with Search Suite to get alignment on this and push forward changes in the future.

So yea, I'm happy with these being like this right now, but, I would want us to explore and work with Search Suite to see if we can move things in a simpler direction.

taiyab:
@camdencheek Doing this also means we need to add in a `x` remove filter button on the RHS on the type items in the filter pane.

client/branded/src/search-ui/results/filters/NewSearchFilters.tsx
```diff
@@ -48,36 +44,29 @@ export const NewSearchFilters: FC<NewSearchFiltersProps> = ({
...
-
-            if (typeFilters.length === 0 || typeFilters.length > 1) {
-                return SearchFilterType.Code
-            }
-
-            return resolveFilterTypeValue(typeFilters[0].value?.value)
+    useEffect(() => {
+        if (queryHasTypeFilter(query) && selectedFilters.some(filter => filter.kind === 'type')) {
+            setSelectedFilters(selectedFilters.filter(filter => filter.kind !== 'type'))
         }
```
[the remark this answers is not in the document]
vovakulikov: Maybe we could use `useLayoutEffect` instead since `useEffect` runs after rendering when layout effect is running before jsx render. This should help with render optimisation when user select/deselect types in filter bar

vovakulikov (approved):
Code looks good. I don't have strong preferences about how exactly it should work, I leave it to @taiyab

client/branded/src/search-ui/results/filters/NewSearchFilters.tsx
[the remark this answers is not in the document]
camdencheek: Happy to return to this, but based on what I'm reading [here](https://kentcdodds.com/blog/useeffect-vs-uselayouteffect), it doesn't sound like `useLayoutEffect` is intended for this?

camdencheek:
@taiyab thanks for the feedback!

> I'm very much happy with treating the type list differently

Yes! I am also happy to treat the type list differently where it makes sense.

Where I think it makes sense right now:
- Show all types at all times to improve discoverability
- Give them different visual treatment to clarify their use as top-level categories

Where I do not think it makes sense right now:
- Selecting a type by default. This is a significant behavior change, and it would make invalid some searches that were previously accepted.
- Mutating the query text. This also breaks some previously valid queries. Additionally, adding the `type:` filter to the query bar duplicates visual information, increasing noise and making it unclear where the source of truth is. This is a behavior difference between the type filters and the subfilters that I don't think is adding value, so in that case, I'd prefer for the behavior to be consistent.

Where I think this makes sense in the future:
- Searching only code by default. Agreed that we should work with the platform team after the release. Definitely still some risk here since we'd no longer be surfacing things like repos in broad, exploratory searches like `sourcegraph`, but seems like something we can mitigate with good design.

> Doing this also means we need to add in a x remove filter button on the RHS on the type items in the filter pane.

Good catch -- added in the latest commit 👍 


CleanShot 2024-02-02 at 09 37 02@2x

client/branded/src/search-ui/results/filters/NewSearchFilters.tsx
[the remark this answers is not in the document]
vovakulikov: The difference here that useEffect doesn't block rendering and runs after you already can see something on the screen, so there should be possible to see some flashes if query has gotten some type filters and we should reset some URL filters (search filter panel) with useLayoutEffect since react performs this hook after jsx creation but before actual rendering synchronously this should prevent all possible flashes with inconsistent search filters URL state

PR #60037, opened 2024-02-01
highlighter: Simplify OffsetManager::new a bit

varungandhi-src:
No functional change

## Test plan

Covered by existing tests

PR #60038, opened 2024-02-01
highlighter: Rename PackedRange -> Range

varungandhi-src:
`Packed` doesn't make sense as a prefix as the value is actually unpacked
(we'll duplicate the line number in the common case)

## Test plan

n/a

### DIFF
--- a/docker-images/syntax-highlighter/crates/scip-treesitter-cli/src/evaluate.rs
+++ b/docker-images/syntax-highlighter/crates/scip-treesitter-cli/src/evaluate.rs
@@ -12,7 +12,7 @@ use colored::{ColoredString, Colorize};
 use scip::types::Index;
 use serde::Serializer;
 use string_interner::{symbol::SymbolU32, StringInterner, Symbol};
-use syntax_analysis::range::PackedRange;
+use syntax_analysis::range::Range;
 
 use crate::{io::read_index_from_file, progress::*};
 
@@ -86,7 +86,7 @@ struct SymbolPair {
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
 struct LocationInFile {
-    rng: PackedRange,
+    rng: Range,
     path_id: PathId,
 }
 
@@ -97,7 +97,7 @@ pub struct SymbolOccurrence {
 }
 
 impl SymbolOccurrence {
-    fn range(&self) -> &PackedRange {
+    fn range(&self) -> &Range {
         &self.location.rng
     }
 }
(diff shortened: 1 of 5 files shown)


varungandhi-src:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/60038?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `main`
  * **PR #60037** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/60037?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #60038** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/60038?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/60038?utm_source=stack-comment).

kritzcreek (approved):
I think ideally we'd try to get rid of this type completely...

For now correcting the name is an improvement though

varungandhi-src:
> I think ideally we'd try to get rid of this type completely...

Could you elaborate on why?

kritzcreek:
> Could you elaborate on why?

The encoding is very inefficient, which matters for Ranges because we're allocating so many of them. Also anywhere we deal with tree-sitter we've already got both byte and line/col based offsets/ranges in the Nodes. Finally the only real operations we perform on these are before/after/contains, which is just as easy (and more efficient) with byte ranges.

varungandhi-src:
Looking at the code where the Range values are being used, AFAICT they're mostly only created transiently on the stack for comparisons/lookups, so we shouldn't worry about those cases. The exception to this we create a hashmap where we store them for lookups in `LocationInFile`. However, for that use case, dealing with line/col for debugging is useful compared to having byte offsets, and we're not really operating at the level of Tree-sitter there.

Maybe your overall point is that we should not even materialize the Occurrence values (which do contain a `Vec` each?)? For that, I agree, that seems a bit odd to do early in the pipeline, instead I think we should defer that to the very end during serialization.

kritzcreek:
> AFAICT they're mostly only created transiently on the stack for comparisons/lookups, so we shouldn't worry about those cases.

`globals.rs` uses them to track ranges of idenfiers and scopes (and those directly correspond to tree-sitter nodes).

> However, for that use case, dealing with line/col for debugging

Agreed, line/col is mostly easier for debugging, but at least in the tree-sitter case nodes carry both a byte offset and a line/col range.

PR #60043, opened 2024-02-01
Fix Cody subscription ends at display logic in UI

thenamankumar:
closes: https://sourcegraph.slack.com/archives/C05PC7AKFQV/p1706755128948399

Silly mistake from the previous PR. 

I override the `cancelAtPeriodEnd` property on the frontend to test that state and take a screenshot. I forgot to remove the code. 

CleanShot 2024-02-01 at 13 31 27@2x

## Test plan
- Set `use-ssc-for-cody-subscription` and `cody-pro-trial-ended` feature flags to true
- Upgrade to pro
- No subscription end date should be visible on /cody/manage
- Cancel the subscription
- Subscription end date should be visible now


vdavid (approved):
👍

PR #60047, opened 2024-02-01
search: default keyword search to on

stefanhengl:
This switches the default from off to on. To disable keyword search completely, customers have to set
`experimentalFeatures.keywordSearch: false` in site settings.

## Test plan
Tested manually

### DIFF
--- a/client/web/src/featureFlags/useFeatureFlag.ts
+++ b/client/web/src/featureFlags/useFeatureFlag.ts
@@ -118,7 +118,5 @@ export function useFeatureFlag(
 }
 
 export function useKeywordSearch(): boolean {
-    const [flagEnabled] = useFeatureFlag('search-keyword')
-    const settingEnabled = !!useExperimentalFeatures(features => features.keywordSearch)
-    return flagEnabled || settingEnabled
+    return useExperimentalFeatures<boolean | undefined>(features => features.keywordSearch) !== false
 }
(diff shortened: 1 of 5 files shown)


jtibshirani:
@stefanhengl I reviewed the Percy snapshots and noticed that the Smart Search icon keeps showing up, whereas it should be gone now? Here's an example:

Screenshot 2024-02-01 at 10 04 43 AM

I'm not able to repro locally, I wonder if we're using a different component in tests or something ...

jtibshirani:
> Seems like we should be moving this out of experimental features?

I probably made a bad decision in putting it there! It'd indeed be good to move this to a normal setting (this should be pretty easy, as everything is encapsulated in one function now, `useKeywordSearch()`). Also we should update the docs on how admins can completely disable this.

keegancsmith:
It may be worth keeping it in experimentalSettings since it is easier to remove stuff from there? Additionally given we default to on, maybe the experimentalSetting should be something like "disableKeywordSearch"?

stefanhengl:
> It may be worth keeping it in experimentalSettings since it is easier to remove stuff from there? Additionally given we default to on, maybe the experimentalSetting should be something like "disableKeywordSearch"?

If we rename, we have to keep the deprecated "keywordSearch" around until we are sure we removed it everywhere, no? I guess it's just s2 and .com?

stefanhengl:
> It may be worth keeping it in experimentalSettings since it is easier to remove stuff from there? Additionally given we default to on, maybe the experimentalSetting should be something like "disableKeywordSearch"?

I am for keeping it in experimental settings until we have updated the API. After that I don't think we should keep the setting around at all.

stefanhengl:
> I'm not able to repro locally, I wonder if we're using a different component in tests or something ...

I cannot repro either

PR #60051, opened 2024-02-01
Implement new UI for generating 5.3 licenses

eseliger:
Screenshot 2024-02-05 at 13 05 50@2x
Screenshot 2024-02-05 at 13 06 26@2x
Screenshot 2024-02-05 at 13 06 41@2x

## Test plan

Manual testing.


pjlast:
Any scenario we can think of where we might need some way to add a tag to a license that's not in the list of pre-approved tags? Like the way we added a `cody` tag to the dotcom license?

eseliger:
Ideally, if we really need it, we should add it to the form. That said, there's a collapsible at the end of the form that discourages you from using tags manually, but still allows to do so :D Does that help?

pjlast (approved):

PR #60054, opened 2024-02-01
svelte: Make compatible with keyword search; remove structural and smart search buttons

fkling:
This commit aligns the prototype more closely with the production app. It removes the structural and smart search buttons defaults to keyword search (which can still be overwritten by server settings).

## Test plan

Visual testing


stefanhengl:
How will the prototype behave if users set `experimentalFeatures.keywordSearch: false` in global settings? In this case the legacy client would show the smart search toggle.

fkling:
> How will the prototype behave if users set experimentalFeatures.keywordSearch: false in global settings? In this case the legacy client would show the smart search toggle.

It will not show the smart search toggle anymore. I'd like to have as few experiments in the prototype as possible (it's an experiment itself 😆 ) and keep it in the "ideal" state that we eventually want to reach (of course we might revert something or have to make things backwards compatible (or maybe not?), in which case we can address that when needed.

PR #60055, opened 2024-02-01
Syntactic indexing: add database schema and dbworker

### LINKED ISSUE #59794: Syntactic indexing: add DB migration for tables and views required by dbworker
Depends on #59747 

We should model it closely after the lsif_indexes table, and follow the structure required for dbworker interface: https://docs.sourcegraph.com/dev/background-information/workers#database-backed-stores

keynmol:
Closes #59794 

This PR:

- Adds a basic schema for syntactic code intel indexes (up and down migration)
- Sets up everything required to use `dbworker` interface
- Adds a stub indexing worker implementation
- Adds tests to ensure conformance with dbworker interface (run against real postgres)

## Test plan

New tests to ensure schema conformance.

For manual testing:
1. Start the configuration: `sg start codeintel-syntactic`
2. Execute this SQL query: `insert into syntactic_scip_indexes(id, commit, repository_id, outfile) values (nextval('syntactic_scip_index_id_seq'), '8c0cc755be302107b675824b0fb1b985992092b8', 1, 'index.scip');`
3. Observe the worker pick up the job:

```
[syntactic-c...1] INFO syntactic-code-intel-worker workerutil/worker.go:329 Dequeued record for processing {"TraceId": "a0b9c4a988d9e46ac42b8c4deb63d670", "SpanId": "6483b7e47fa617f2", "name": "syntactic_code_intel_indexer", "id": "1"}
[syntactic-c...1] INFO syntactic-code-intel-worker.Handle shared/indexing_worker.go:34 Stub indexing worker handling record {"TraceId": "a0b9c4a988d9e46ac42b8c4deb63d670", "SpanId": "470708143297b7ba", "handle": {"id": 1, "repository name": "github.com/sourcegraph-testing/zap", "commit": "8c0cc755be302107b675824b0fb1b985992092b8"}}
```


internal/workerutil/dbworker/store/store.go
```diff
@@ -492,7 +492,7 @@ func (s *store[T]) Dequeue(ctx context.Context, workerHostname string, condition
 		s.columnReplacer.Replace("{worker_hostname}"):   workerHostnameExpr,
 	}
 
-	records, err := s.options.Scan(s.Query(ctx, s.formatQuery(
+	query := s.formatQuery(
```
keynmol, line 495: TODO - remove
I was getting nil pointers on the line below because several of the `s.options` fields were nil and apparently were required, so I separated the nested expressions.

sg.config.yaml
```diff
@@ -1256,8 +1256,12 @@ commandsets:
       - git
     commands:
       - frontend
+      - web
       - worker
       - blobstore
+      - repo-updater
+      - gitserver-0
+      - gitserver-1
```
keynmol, line 1264: As we now require the repo table to be populated for the `syntactic_scip_indexes_with_repository_name` view to work in joins, it's useful to have repo updater running locally.

dev/scip-syntax-install.sh
```diff
@@ -26,7 +26,7 @@ trap ctrl_c INT
 function build_scip_syntax {
   cd docker-images/syntax-highlighter/crates/scip-syntax
   cargo build --bin scip-syntax --target-dir target
-  cp ./target/release/scip-syntax "$TARGET"
+  cp ./target/debug/scip-syntax "$TARGET"
```
keynmol, line 29: The `cargo` command above builds the CLI in debug mode.
For local testing I think it's better to keep this in debug mode (this script is not used for building the service images) to iterate quicker.

cmd/syntactic-code-intel-worker/shared/shared.go
```diff
@@ -23,9 +23,13 @@ func Main(ctx context.Context, observationCtx *observation.Context, ready servic
 	}
 
 	logger.Info("Syntactic code intel worker running",
-		log.String("path to scip-syntax CLI", config.CliPath),
+		log.String("path to scip-syntax CLI", config.IndexingWorker.CliPath),
 		log.String("API address", config.ListenAddress))
 
+	db := mustInitializeDB(observationCtx, "syntactic-code-intel-indexer")
+	workerStore, _ := NewStore(observationCtx, db)
```
keynmol, line 30: TODO: surface error

cmd/syntactic-code-intel-worker/shared/config.go
```diff
@@ -5,36 +5,43 @@ import (
...
 }
 
 const DefaultPort = 3188
 
-func (c *Config) Load() {
-	c.SCIPUploadStoreConfig = &lsifuploadstore.Config{}
-	c.SCIPUploadStoreConfig.Load()
+func scopedEnv(name string) string {
+	return "SYNTACTIC_CODE_INTEL_" + name
+}
```
varungandhi-src, line 32: It's not a big deal, but for env vars, it's generally helpful to copy the prefixes in-place because it aids with textual search.

cmd/syntactic-code-intel-worker/shared/config.go
```diff
@@ -5,36 +5,43 @@ import (
...
 	env.BaseConfig
 
-	WorkerPollInterval    time.Duration
-	WorkerConcurrency     int
-	WorkerBudget          int64
-	MaximumRuntimePerJob  time.Duration
-	SCIPUploadStoreConfig *lsifuploadstore.Config
-	CliPath               string
-	ListenAddress         string
+	IndexingWorker *IndexingWorkerConfig
```
varungandhi-src, line 23:
```suggestion
	*IndexingWorkerConfig
```

The usage site reads a bit weird with just `IndexingWorker.Validate()`. It's also a bit confusing because the worker itself is conceptually a different entity from its configuration.

cmd/syntactic-code-intel-worker/shared/indexing_worker.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/sourcegraph/sourcegraph/internal/observation"
+	"github.com/sourcegraph/sourcegraph/internal/workerutil"
+	"github.com/sourcegraph/sourcegraph/internal/workerutil/dbworker"
+	dbworkerstore "github.com/sourcegraph/sourcegraph/internal/workerutil/dbworker/store"
+)
+
+func NewIndexingWorker(ctx context.Context, observationCtx *observation.Context, workerStore dbworkerstore.Store[*SyntacticIndexRecord], config IndexingWorkerConfig) *workerutil.Worker[*SyntacticIndexRecord] {
+
+	name := "syntactic_code_intel_indexer"
```
varungandhi-src, line 16:
```suggestion
	name := "syntactic_code_intel_indexing_worker"
```

To match the existing naming conventions for other workers.

cmd/syntactic-code-intel-worker/shared/store.go
```diff
@@ -0,0 +1,145 @@
...
+	StartedAt      *time.Time `json:"startedAt"`
+	FinishedAt     *time.Time `json:"finishedAt"`
+	ProcessAfter   *time.Time `json:"processAfter"`
+	NumResets      int        `json:"numResets"`
+	NumFailures    int        `json:"numFailures"`
+	RepositoryID   int        `json:"repositoryId"`
+	RepositoryName string     `json:"repositoryName"`
+	Outfile        string     `json:"outfile"`
+	ShouldReindex  bool       `json:"shouldReindex"`
+	EnqueuerUserID int32      `json:"enqueuerUserID"`
```
varungandhi-src, line 38: I know this doesn't match the style of the code elsewhere which is heavily undocumented, but it would be helpful to add doc comments to at least some of the fields describing:
- What the possible states for `State` are (can we use a Go "string-based enum" instead?)
- What does "reset" mean/when does it happen?
- Is `FailureMessage` only meant to store the most recent failure message? Thoughts on making it be a slice of strings instead? That way we'd be able to get rid of `NumFailures` as that could be inferred from the length of `FailureMessage`.

cmd/syntactic-code-intel-worker/shared/store.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func (i SyntacticIndexRecord) RecordID() int {
+	return i.ID
+}
```
varungandhi-src, line 43: Is this method for satisfying an interface? If so, it would be helpful to add a type assertion for that interface for clarity. `var _ InterfaceName = SyntacticIndexRecord{}`.

cmd/syntactic-code-intel-worker/shared/store.go
```diff
@@ -0,0 +1,145 @@
...
+		sqlf.Sprintf("u.repository_name"),
+		sqlf.Sprintf("u.outfile"),
+		sqlf.Sprintf("u.should_reindex"),
+		sqlf.Sprintf("u.enqueuer_user_id"),
+	}
+
+	storeOptions := dbworkerstore.Options[*SyntacticIndexRecord]{
+		Name:      "syntactic_scip_index_store",
+		TableName: "syntactic_scip_indexes",
+		ViewName:  "syntactic_scip_indexes_with_repository_name u",
```
varungandhi-src, line 108: Thoughts on having "job", "jobs" or "queue" somewhere in the name? Right now, this reads as-if this table itself stores the indexes themselves, but IIUC, it is storing information about the jobs that need to be run, with modification to the job status after they are run.

cmd/syntactic-code-intel-worker/shared/store.go
```diff
@@ -0,0 +1,145 @@
...
+	State          string     `json:"state"`
+	FailureMessage *string    `json:"failureMessage"`
+	StartedAt      *time.Time `json:"startedAt"`
+	FinishedAt     *time.Time `json:"finishedAt"`
+	ProcessAfter   *time.Time `json:"processAfter"`
+	NumResets      int        `json:"numResets"`
+	NumFailures    int        `json:"numFailures"`
+	RepositoryID   int        `json:"repositoryId"`
+	RepositoryName string     `json:"repositoryName"`
+	Outfile        string     `json:"outfile"`
```
varungandhi-src, line 36:
```suggestion
```

I think we can remove this field? The indexer is fixed, so the index for the whole repository is guaranteed to be produced at `index.scip`.

cmd/syntactic-code-intel-worker/shared/store.go
```diff
@@ -0,0 +1,145 @@
...
+	// Relevant issue: https://github.com/sourcegraph/sourcegraph/issues/15962
+
+	ctx := context.Background()
+	db := database.NewDB(observationCtx.Logger, sqlDB)
+	go func() {
+		for range time.NewTicker(providers.RefreshInterval()).C {
+			allowAccessByDefault, authzProviders, _, _, _ := providers.ProvidersFromConfig(ctx, conf.Get(), db)
+			authz.SetProviders(allowAccessByDefault, authzProviders)
+		}
+	}()
```
varungandhi-src, line 143: There is something off here. The comment seems to be based on other calls to `authz.SetProviders(true, ...)`. However, that's not what we're doing here.

cmd/syntactic-code-intel-worker/shared/store_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+	assert.Equal(t, 1, rec1.ID)
+	assert.Equal(t, "tangy/tacos", rec1.RepositoryName)
+	assert.Equal(t, "deadbeefdeadbeefdeadbeefdeadbeefdead1111", rec1.Commit)
+
+	rec2, _, _ := store.Dequeue(ctx, "worker2", nil)
+
+	assert.Equal(t, 2, rec2.ID)
+	assert.Equal(t, "salty/empanadas", rec2.RepositoryName)
+	assert.Equal(t, "deadbeefdeadbeefdeadbeefdeadbeefdead2222", rec2.Commit)
```
varungandhi-src, line 83: `assert.Equal` will actually keep going even if the assertion fails (so you'll get N assertion failures if something goes wrong), which is generally not helpful in practice. I recommend `require.Equal` for fast failure.

Also, can we assert something for the other return values? E.g. looks like the last return value is an `error`, so you could do `require.NoError(t, err)`.

keynmol: Hah, didn't know about it
Will address

cmd/syntactic-code-intel-worker/shared/store.go
keynmol: Not sure I understand, this bit is copied from the precise code intel worker, and it seems to exist in this same form across the codebase: https://github.com/search?q=repo%3Asourcegraph%2Fsourcegraph%20SetProviders&type=code

cmd/syntactic-code-intel-worker/shared/store.go
keynmol: > What the possible states for State are (can we use a Go "string-based enum" instead?)

We can make this a bit better in this particular part of the codebase, but IMO that won't make it safer - those state constants are hardcoded in query strings and various constants inside dbworker code. 

So can't _really_ enact meaningful change without refactoring there - and in the syntactic worker it's not likely that we will be accessing those fields directly, as the state of them is managed through the dbworker.Store interface, to ensure all the operation is are consisent across usages..

cmd/syntactic-code-intel-worker/shared/store.go
keynmol: Agree, I copied after `lsif_indexes` but it didn't feel right.

WDYT `syntactic_scip_indexing_jobs` and `syntactic_scip_indexing_jobs_with_repository_name` for table and view respectively?

A bit long but at least it's clear

cmd/syntactic-code-intel-worker/shared/store.go
keynmol: I will add a link to extensive docs about the fields: https://sourcegraph.com/docs/dev/background-information/workers#database-backed-stores

And will add a couple of inline comments for confusing fields.

varungandhi-src: 👍🏽 I didn't realize that the State was shared in other places too, that makes sense.

cmd/syntactic-code-intel-worker/shared/store.go
varungandhi-src: The comment in this file is currently saying:

```
	// This is an internal service, so we rely on the
	// frontend to do authz checks for user requests.
```

So this comment makes it sound like we're not doing any authz check here. If you look in gitserver as one example, there is a similar comment.

https://sourcegraph.com/github.com/sourcegraph/sourcegraph@3c06bc0806981d8de0492c5ce8a74d48c3ca3720/-/blob/cmd/gitserver/shared/shared.go?L265-269

```
	// Gitserver is an internal actor. We rely on the frontend to do authz checks for
	// user requests.
	//
	// This call to SetProviders is here so that calls to GetProviders don't block.
	authz.SetProviders(true, []authz.Provider{})
```

However, in gitserver, the call is `authz.SetProviders(true, []authz.Provider{})`, so the comment makes sense -- the provider in gitserver allows access by default because a different subsystem is responsible for the authz checks.

In this code, the `SetProviders` call looks different.

```
			allowAccessByDefault, authzProviders, _, _, _ := providers.ProvidersFromConfig(ctx, conf.Get(), db)
			authz.SetProviders(allowAccessByDefault, authzProviders)
```

Presumably, it is possible that `allowAccessByDefault` is false, and `authzProviders` is non-empty, right?

So the comment isn't quite consistent with the code -- is the frontend always going to do authz checks? If so, then we don't we just have:

```
	authz.SetProviders(true, []authz.Provider{})
```

here too similar to gitserver?

cmd/syntactic-code-intel-worker/shared/store.go
varungandhi-src: `syntactic_scip_indexing_jobs` seems good. 👍🏽

cmd/syntactic-code-intel-worker/shared/config.go
```diff
@@ -5,36 +5,43 @@ import (
...
 }
 
 const DefaultPort = 3188
 
-func (c *Config) Load() {
-	c.SCIPUploadStoreConfig = &lsifuploadstore.Config{}
-	c.SCIPUploadStoreConfig.Load()
+func scopedEnv(name string) string {
+	return "SYNTACTIC_CODE_INTEL_" + name
+}
```
varungandhi-src, line 32:
```suggestion
```

Unused code since the prefix was inlined

cmd/syntactic-code-intel-worker/shared/store.go
```diff
@@ -0,0 +1,162 @@
...
+	ProcessAfter   *time.Time  `json:"processAfter"`
+	NumResets      int         `json:"numResets"`
+	NumFailures    int         `json:"numFailures"`
+	// Which repository id to index (not part of dbworker fields)
+	RepositoryID int `json:"repositoryId"`
+	// Name of repository being indexed (not part of dbworker fields)
+	RepositoryName string `json:"repositoryName"`
+	ShouldReindex  bool   `json:"shouldReindex"`
+	// Which user scheduled this job (not part of dbworker fields)
+	EnqueuerUserID int32 `json:"enqueuerUserID"`
```
varungandhi-src, line 56:
```suggestion
	State          recordState `json:"state"`
	QueuedAt       time.Time   `json:"queuedAt"`
	StartedAt      *time.Time  `json:"startedAt"`
	FinishedAt     *time.Time  `json:"finishedAt"`
	ProcessAfter   *time.Time  `json:"processAfter"`
	NumResets      int         `json:"numResets"`
	NumFailures    int         `json:"numFailures"`
	FailureMessage *string     `json:"failureMessage"`
	ShouldReindex  bool   `json:"shouldReindex"`
    // The fields below are not part of the standard dbworker fields
 
    RepositoryID int `json:"repositoryId"`
    RepositoryName string `json:"repositoryName"`
	Commit         string      `json:"commit"`
	// Which user scheduled this job
	EnqueuerUserID int32 `json:"enqueuerUserID"`
```

Attempted to rearrange this so that related fields are together and all the non-standard fields are grouped together.

Is the repository name preserved here for aiding with debugging to avoid an extra join?

cmd/syntactic-code-intel-worker/shared/store.go
varungandhi-src: I think we could perhaps shorten it a bit to `syntactic_indexing_jobs` and remove the `scip_` bit, but your call.

migrations/frontend/squashed.sql
```diff
@@ -4646,6 +4646,65 @@ CREATE SEQUENCE survey_responses_id_seq
 
 ALTER SEQUENCE survey_responses_id_seq OWNED BY survey_responses.id;
 
+CREATE TABLE syntactic_scip_indexing_jobs (
+    id bigint NOT NULL,
```
varungandhi-src, line 4650: Should this be marked `PRIMARY KEY` or is that not necessary?

migrations/frontend/squashed.sql
```diff
@@ -4646,6 +4646,65 @@ CREATE SEQUENCE survey_responses_id_seq
...
+    num_resets integer DEFAULT 0 NOT NULL,
+    num_failures integer DEFAULT 0 NOT NULL,
+    execution_logs json[],
+    commit_last_checked_at timestamp with time zone,
+    worker_hostname text DEFAULT ''::text NOT NULL,
+    last_heartbeat_at timestamp with time zone,
+    cancel boolean DEFAULT false NOT NULL,
+    should_reindex boolean DEFAULT false NOT NULL,
+    enqueuer_user_id integer DEFAULT 0 NOT NULL,
+    CONSTRAINT syntactic_scip_indexing_jobs_commit_valid_chars CHECK ((commit ~ '^[a-z0-9]{40}$'::text))
```
varungandhi-src, line 4668:
```suggestion
    CONSTRAINT syntactic_scip_indexing_jobs_commit_valid_chars CHECK ((commit ~ '^[a-f0-9]{40}$'::text))
```

cmd/syntactic-code-intel-worker/shared/store_helpers_test.go
```diff
@@ -0,0 +1,102 @@
...
+			index.NumResets,
+			index.NumFailures,
+			index.RepositoryID,
+			index.ShouldReindex,
+			index.EnqueuerUserID,
+		)
+
+		if _, err := db.ExecContext(context.Background(), query.Query(sqlf.PostgresBindVar), query.Args()...); err != nil {
+			t.Fatalf("unexpected error while inserting index: %s", err)
+		}
```
varungandhi-src, line 62:
```suggestion
		_, err := db.ExecContext(context.Background(), query.Query(sqlf.PostgresBindVar), query.Args()...)
		require.NoError(t, err, "unexpected error while inserting index")
```

Bit more compact. Same could be applied in other places.

cmd/syntactic-code-intel-worker/shared/store.go
keynmol: Right, I see what you are saying.

For this PR I will go with the more permissive version (`SetProviders(true, ...)`) and will create an issue to follow up in precise code intel worker to see if we can simplify there as well.

varungandhi-src (commented):
Added some more comments; will do a third pass shortly over some of the SQL.

cmd/syntactic-code-intel-worker/shared/store_helpers_test.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/sourcegraph/sourcegraph/internal/database"
+)
+
+func insertIndexRecords(t testing.TB, db database.DB, records ...SyntacticIndexingJob) {
+	for _, index := range records {
+		if index.Commit == "" {
+			index.Commit = makeCommit(index.ID)
+		}
+		if index.State == "" {
+			index.State = "completed"
```
varungandhi-src, line 20:
```suggestion
			index.State = Completed
```

migrations/frontend/squashed.sql
keynmol: You're right, it seems lsif_indexes table received those keys over time:

```
Indexes:
    "lsif_indexes_pkey" PRIMARY KEY, btree (id)
    "lsif_indexes_commit_last_checked_at" btree (commit_last_checked_at) WHERE state <> 'deleted'::text
    "lsif_indexes_dequeue_order_idx" btree ((enqueuer_user_id > 0) DESC, queued_at DESC, id) WHERE state = 'queued'::text OR state = 'errored'::text
    "lsif_indexes_queued_at_id" btree (queued_at DESC, id)
    "lsif_indexes_repository_id_commit" btree (repository_id, commit)
    "lsif_indexes_repository_id_indexer" btree (repository_id, indexer)
    "lsif_indexes_state" btree (state)
```

Most of them are relevant to us, so I will add them as part of initial table creation

migrations/frontend/1706710430_add_syntactic_scip_indexes_table/up.sql
```diff
@@ -0,0 +1,59 @@
+
+CREATE TABLE IF NOT EXISTS syntactic_scip_indexing_jobs (
```
varungandhi-src, line 2: Q: Is there a single source of truth in terms of the SQL files and the other SQL files are generated from that, or are they all hand-written?

Also, I noticed this in our docs: https://github.com/sourcegraph/sourcegraph/blob/main/migrations/README.md

```
To pass CI, you'll additionally need to:

    Ensure that your new migrations run against the current Go unit tests
    Ensure that your new migrations can be run up, then down, then up again (idempotency test)
    Ensure that your new migrations do not break the Go unit tests published with the previous release (backwards-compatibility test)
```

Do we need to add separate testing for this by registering this migration somewhere, or is this automatically done by virtue of these files being present in the right places?

varungandhi-src (approved):
No blocking comments, but some questions and suggestions for improving readability.

migrations/frontend/1706710430_add_syntactic_scip_indexes_table/up.sql
keynmol: The idmpotency test is run by `bazel test //internal/database/connections/live:live_test` (I have a few commits here with fixes because of this particular test)

The migrations running against current unit tests seems to be done automatically, we exercise the table in the current sole test.

The breaking previous releases thing I'm not sure - they seem to imply that tests from this repo can be run against a different version of the database. I'll look around the codebase to see if there are some guardrails in place for that.

keynmol: > Is there a single source of truth in terms of the SQL files and the other SQL files are generated from that, or are they all hand-written?

The squashed SQL files (and the json and markdown schema files) are generated by `bazel run @//dev:write_all_generated` but it needs to be run manually

cmd/syntactic-code-intel-worker/shared/store_helpers_test.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+func insertRepo(t testing.TB, db database.DB, id int, name string) {
+	if name == "" {
+		name = fmt.Sprintf("n-%d", id)
+	}
+
+	deletedAt := sqlf.Sprintf("NULL")
+	if strings.HasPrefix(name, "DELETED-") {
+		deletedAt = sqlf.Sprintf("%s", time.Unix(1587396557, 0).UTC())
```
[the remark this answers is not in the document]
keynmol: The DELETED- is the side effect of copypasting it from autoindexing module - I didn't want to to introduce dependency for such small amount of helper methods

Not that I see this being unused, I can add a test that verifies that if the repository doesn't exist anymore, the job is not dequeued by the worker - thus verifying that the SQL view we're using joins with repositories and won't return records for deleted repositories. WDYT?

cmd/syntactic-code-intel-worker/shared/store_test.go
```diff
@@ -0,0 +1,89 @@
...
+	require.Equal(t, "deadbeefdeadbeefdeadbeefdeadbeefdead1111", record1.Commit)
+
+	record2, hasRecord, err := store.Dequeue(ctx, "worker2", nil)
+
+	require.NoError(t, err)
+	require.True(t, hasRecord)
+	require.Equal(t, 2, record2.ID)
+	require.Equal(t, "salty/empanadas", record2.RepositoryName)
+	require.Equal(t, "deadbeefdeadbeefdeadbeefdeadbeefdead2222", record2.Commit)
+
```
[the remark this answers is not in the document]
keynmol: Added a test for this exact case in https://github.com/sourcegraph/sourcegraph/pull/60055/commits/512b133c6cf5d8ab60ee7c0bc2b3154b210741b7

keynmol:
@varungandhi-src once the build is green I will merge this - if you have any more comments, please add them here or raise a separate issue.

PR #59986, opened 2024-01-30
auth: redirect to "/sign-in" when auth provider not found

unknwon:
When there is no matching auth provider, there is almost always something more fucky going on. Although this PR is made specifically for VSCode Cody client, i.e. when users click on the "Sign in with X" button that links to a non-existent auth provider, it is generally more helpful to the end user to get back to the `/sign-in` page then a raw screen with a opaque error message.

## Test plan

```zsh
→ curl https://sourcegraph.test:3443/.auth/openidconnect/login?pc=404&returnTo=%2Fsearch
<a href="/sign-in?returnTo=/search">Found</a>.

→ curl https://sourcegraph.test:3443/.auth/github/login?pc=404&returnTo=%2Fsearch
<a href="/sign-in?returnTo=/search">Found</a>.
```

PR #60065, opened 2024-02-01
auth: be fully compatible with our creative redirect params

unknwon:
https://github.com/sourcegraph/sourcegraph/pull/59986 does not _fully_ work, so here is the follow-up.

## Test plan

```zsh
→ curl https://sourcegraph.test:3443/.auth/openidconnect/login?pc=404&returnTo=%2Fsearch
<a href="/sign-in?returnTo=/search">Found</a>.

→ curl https://sourcegraph.test:3443/.auth/github/login?pc=404&returnTo=%2Fsearch
<a href="/sign-in?returnTo=/search">Found</a>.

→ curl https://sourcegraph.test:3443/.auth/openidconnect/login?pc=404&redirect=%2Fsearch
<a href="/sign-in?returnTo=/search">Found</a>.

→ curl https://sourcegraph.test:3443/.auth/github/login?pc=404&redirect=%2Fsearch
<a href="/sign-in?returnTo=/search">Found</a>.
```

### DIFF
--- a/cmd/frontend/internal/auth/oauth/middleware.go
+++ b/cmd/frontend/internal/auth/oauth/middleware.go
@@ -83,8 +83,13 @@ func newOAuthFlowHandler(serviceType string) http.Handler {
 		id := r.URL.Query().Get("pc")
 		p := GetProvider(serviceType, id)
 		if p == nil {
+			redirect := r.URL.Query().Get("redirect")
+			if redirect == "" {
+				redirect = r.URL.Query().Get("returnTo")
+			}
+
 			log15.Warn("no OAuth provider found with ID and service type", "id", id, "serviceType", serviceType)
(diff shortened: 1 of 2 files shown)


kalanchan (approved):
"creative" haha

this looks good, we'll test it e2e once we land https://github.com/sourcegraph/jetbrains/pull/464 and https://github.com/sourcegraph/cody/pull/2957

eseliger:
What’s the reason that we cannot fix the param on the other end of the wire?

unknwon:
> What’s the reason that we cannot fix the param on the other end of the wire?

There is no reason why we cannot, but the problem is that the other end of the wire is Cody clients, which we have no control on the version rollout, thus we need to keep backwards compatible even if we managed to fix all server-side use cases.

eseliger:
could you leave a super small comment in there that this is the sole reason we do this, and that we can remove it in the future? :)

unknwon:

PR #60066, opened 2024-02-01
Keyword search: support patterntype in other features

jtibshirani:
This PR fixes several issues when you perform a keyword search, then select Actions -> ...
* Code monitoring now allows `patterntype:keyword`
* Saved searches now accept `keyword`
* Insights already allowed `keyword` for "Track changes", now it correctly rejects it for "Detect and track patterns"
* The keyword patterntype no longer has these red squiggly lines in various search inputs:

Screenshot 2024-02-01 at 12 04 49 PM

## Test plan

Manual testing. Also tested searches within Notebooks and Batch changes, and `keyword` is already allowed.

### DIFF
--- a/client/shared/src/search/query/filters.ts
+++ b/client/shared/src/search/query/filters.ts
@@ -271,8 +271,9 @@ export const FILTERS: Record<NegatableFilter, NegatableFilterDefinition> &
         placeholder: '"content"',
     },
     [FilterType.patterntype]: {
-        discreteValues: () => ['regexp', 'structural', 'literal', 'standard'].map(value => ({ label: value })),
-        description: 'The pattern type (standard, regexp, literal, structural) in use',
+        discreteValues: () =>
+            ['keyword', 'literal', 'regexp', 'standard', 'structural'].map(value => ({ label: value })),
+        description: 'The pattern type (standard, keyword, regexp, literal) in use',
         singular: true,
     },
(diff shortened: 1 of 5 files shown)


jtibshirani:
Question for reviewers: previously, Code monitoring only accepted `literal` and `regexp` searches. Notably, it doesn't accept the search bar default of `standard` -- is there a reason for this? Am I missing something where we shouldn't allow `keyword`?

client/shared/src/search/query/filters.ts
```diff
@@ -271,8 +271,8 @@ export const FILTERS: Record<NegatableFilter, NegatableFilterDefinition> &
         placeholder: '"content"',
     },
     [FilterType.patterntype]: {
-        discreteValues: () => ['regexp', 'structural', 'literal', 'standard'].map(value => ({ label: value })),
-        description: 'The pattern type (standard, regexp, literal, structural) in use',
+        discreteValues: () => ['keyword', 'literal', 'regexp', 'standard', 'structural'].map(value => ({ label: value })),
+        description: 'The pattern type (standard, keyword, regexp, literal) in use',
```
jtibshirani, line 275: I removed 'structural' in a few descriptions, since we are trying to advertise it less, and these lists were getting super long.

stefanhengl (approved):

PR #60067, opened 2024-02-01
Search jobs: support boolean patterns

jtibshirani:
This PR adds support for boolean patterns to search jobs. It should be okay in
terms of performance, since we recently updated searcher to handle AND/ OR
patterns natively (https://github.com/sourcegraph/sourcegraph/issues/59038).

This also makes the integration with keyword search a lot nicer. Now when you
have a keyword search with multiple terms and you go to create a search job, it
succeeds. Previously it showed this confusing error:

Screenshot 2024-02-01 at 11 55 55
AM

## Test plan

Updated unit tests, light manual testing


keegancsmith (approved):
LGTM. Worth cherry-picking into the release branch

PR #60075, opened 2024-02-01
Correct discrepancy between "subscription end" and "trial end" on Cody Management Page

### LINKED ISSUE #60071: cody/manage page shows inconsistent data regarding Cody Pro free tiral
Reported [on Slack](https://sourcegraph.slack.com/archives/C05SZB829D0/p1706823068694079)

image

chrsmith:
You can be in a situation where the `/cody/management` page is showing you inconsistent/confusing data. Unfortunately there are several reasons for this, and some we won't be able to fix for another few weeks.

image

This PR introduces some new local variables to tease apart different uses of the `cancelAtPeriodEnd` field. And adds a boat load of comments to clarify what things need to change and how, as we iterate on this page and the larger SSC integration.

> @thenamankumar , as always I reserve the right to be totally wrong. Feel free to make any changes to the comments if they are factually incorrect. Otherwise, if things look good please go ahead and merge it so we can get this deployed ASAP. 🙏 

Fixes #60071 

## Test plan

I did no testing for this. We definitely should have unit tests for the `CodyManagementPage.tsx` (since this is only the tip of the iceberg in terms of annoying complexity and nuance to subscription statuses and their interpretation).

But I'm going to feign ignorance for the time being in not knowing how to create said unit tests for a React component in a reasonable amount of time.

### DIFF
--- a/client/web/src/cody/management/CodyManagementPage.tsx
+++ b/client/web/src/cody/management/CodyManagementPage.tsx
@@ -142,7 +142,7 @@ export const CodyManagementPage: React.FunctionComponent<CodyManagementPageProps
     // change its value accordingly.
     const freeTrialEndString = 'Until Feb 14, 2024'
     if (!hasTrialEnded && userIsOnProTier) {
-        codyProSubscriptionEndTime = new Date(2024, 2, 14, 12, 0, 0)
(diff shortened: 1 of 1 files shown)


vdavid (approved):
This is an improvement, so approving to get it merged. We can follow up on the marked bugs and do more tweaks based on QA.

PR #60076, opened 2024-02-02
Updated Postgres install step for Postgres 15

jamesmcnamara:
As was reported by @arafatkatze, `sg setup` was broken for postgres 15. There were two main reasons for this:
1. brew considers postgres 15 a keg only formula and so doesn't add the utilities to the path (causing issues usually when another install of postgres is visible)
2. brew for Postgres 15 changed how they call initdb, now using the login shell to be the user for the postgres instead of `postgres` or whatever is the value for `PGUSER`

## Test plan
I tested this by removing postgres, and re-running `sg setup --fix`, as well as installing a fresh postgres through brew and running it. Then I installed postgres 14 to make sure it would complain that it was the wrong version.


arafatkatze (approved):
I was facing the migration issues with a fresh install on another new computer

See this

image


So then I followed the steps 
```

brew services stop postgresql@15
brew uninstall postgresql@15
rm -rf /usr/local/var/postgres
rm /usr/local/var/log/postgres.log
rm -f ~/.psqlrc ~/.psql_history


```

And then it almost worked but I had to do 
`
brew services start postgresql@15 `


And then this worked perfectly.

See my entire logs https://gist.github.com/arafatkatze/47df8fe47665e6d9e1fc18ddc7bab233

Still approving the PR because its super awesome and useful and will save people a lot of time. If theres any chance the service start command thing can be added that would be awesome too.


Thanks a lot James 🙇

arafatkatze:
@jamesmcnamara Thanks a lot for handling this with so much diligence.

jamesmcnamara:
No worries @arafatkatze . It won't help you now, but these changes should also handle those two steps you had to do manually, so HOPEFULLY `sg setup --fix` should be all someone needs.

PR #60082, opened 2024-02-02
Test for jansson shared library rather than .la file

willdollman:
Our image tests looked for the presence of a libarchive file. This isn't present in newer versions of the package - presumably as it isn't required at runtime so has been stripped. Replace the test with the actual shared library file.

## Test plan

- CI build

### DIFF
--- a/cmd/symbols/image_test.yaml
+++ b/cmd/symbols/image_test.yaml
@@ -32,5 +32,5 @@ fileExistenceTests:
   gid: 101
   permissions: 'drwxr-xr-x'
 - name: 'jansson package'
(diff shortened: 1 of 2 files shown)


burmudar (approved):

keegancsmith (approved):
we used to try statically link the binary for universal-ctags. But in this new world I don't think that is needed, so LGTM.

For your image tests can you run `universal-ctags --list-features` and check that `interactive` and `json` are in the output? Then I don't think you really need to test for libjansson, and rather just assert on the final binary that it has the features we need?

PR #60084, opened 2024-02-02
ctags: ensure that binary has correct featureset

willdollman:
Confimed that this test validates that ctags was both built with jansson and that the runtime environment has the correct libs:

```
$ universal-ctags --list-features
#NAME             DESCRIPTION
iconv             can convert input/output encodings
interactive       accepts source code from stdin
json              supports json format output
option-directory  TO BE WRITTEN
optscript         can use the interpreter
packcc            has peg based parser(s)
regex             can use regular expression based pattern matching
wildcards         can use glob matching

# Remove jansson shared libs
rm /usr/lib/libjansson.*

$ universal-ctags --list-features
universal-ctags: error while loading shared libraries: libjansson.so.4: cannot open shared object file: No such file or directory

```

## Test plan

- Local testing
- CI

### DIFF
--- a/cmd/symbols/image_test.yaml
+++ b/cmd/symbols/image_test.yaml
@@ -15,8 +15,10 @@ commandTests:
     envVars:
       - key: "SANITY_CHECK"
         value: "true"
-  - name: "universal-ctags has correct features"
-    command: "/usr/bin/universal-ctags --list-features"
+  - name: "ctags has correct features"
+    command: "universal-ctags"
+    args:
+      - --list-features
     expectedOutput: ["interactive", "json"]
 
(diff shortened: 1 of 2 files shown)


cmd/symbols/image_test.yaml
```diff
@@ -15,6 +15,11 @@ commandTests:
     envVars:
       - key: "SANITY_CHECK"
         value: "true"
+  - name: "ctags has correct features"
+    command: "universal-ctags"
+    args:
+      - --list-features
+    expectedOutput: ["interactive", "json"]
```
keegancsmith, line 22: nice

keegancsmith (approved):

PR #60086, opened 2024-02-02
all: use stdlib slices package

keegancsmith:
This is a bigger change than I want, but the version of exp/slices we used had a change from taking a Less func for SortFunc to a Compare. To update a vulnerability we also bring in a new version of exp/slices, so I seperated out this change.

All our uses of exp/slices are now covered by the stdlib, so I migrate all our uses to it. In particular this mostly meant updating custom Less functions to using the stdlib cmp.Compare.

The only functional change was a potential bug I found in filter_file_contains.go. The comment for the sort says sort by Start offset, but the code was incorrectly comparing the end offset for the 2nd argument. This makes sort's Less function break the expected invariants, so the code was quite wrong. I expect in practice we didn't notice it since most ranges did not overlap. This fix makes the compare function respect the expected invariants and just seems correct.

Test Plan: CI and grep returns nothing for "golang.org/x/exp/slices" in go files.

### DIFF
--- a/internal/authz/providers/github/github_test.go
+++ b/internal/authz/providers/github/github_test.go
@@ -5,13 +5,13 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/url"
+	"slices"
 	"strings"
 	"testing"
 	"time"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/gregjones/httpcache"
-	"golang.org/x/exp/slices"
 
 	"github.com/sourcegraph/sourcegraph/internal/api"
 	"github.com/sourcegraph/sourcegraph/internal/authz"
@@ -41,7 +41,7 @@ func mockClientFunc(mockClient client) func() (client, error) {
 }
 
 func stableSortRepoID(v []extsvc.RepoID) {
-	slices.SortStableFunc(v, func(a, b extsvc.RepoID) bool { return strings.Compare(string(a), string(b)) < 1 })
+	slices.Sort(v)
 }
 
 // newMockClientWithTokenMock is used to keep the behaviour of WithToken function mocking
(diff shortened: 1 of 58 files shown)


internal/search/job/jobutil/filter_file_contains.go
```diff
@@ -221,8 +221,10 @@ func (j *fileContainsFilterJob) filterCommitMatch(ctx context.Context, searcherU
 
 func (j *fileContainsFilterJob) removeUnmatchedFileDiffs(cm *result.CommitMatch, matchedFileCounts map[string]int) result.Match {
 	// Ensure the matched ranges are sorted by start offset
-	slices.SortFunc(cm.DiffPreview.MatchedRanges, func(a, b result.Range) bool {
-		return a.Start.Offset < b.End.Offset
+	slices.SortFunc(cm.DiffPreview.MatchedRanges, func(a, b result.Range) int {
+		// TODO(keegancsmith) I changed this from b.End to b.Start since that
+		// matches the comment above.
+		return a.Start.Compare(b.Start)
```
keegancsmith, line 227: @camdencheek can I remove this TODO and assume this change is correct?
```suggestion
		return a.Start.Compare(b.Start)
```

internal/authz/providers/github/github_test.go
```diff
@@ -41,7 +41,7 @@ func mockClientFunc(mockClient client) func() (client, error) {
 }
 
 func stableSortRepoID(v []extsvc.RepoID) {
-	slices.SortStableFunc(v, func(a, b extsvc.RepoID) bool { return strings.Compare(string(a), string(b)) < 1 })
+	slices.Sort(v)
```
eseliger, line 44: I _assume_ SortStable was used here for a reason - is the new Sort stable by default?

eseliger (approved):

internal/authz/providers/github/github_test.go
keegancsmith: I didn't dive into why StableSort was used, I assume at some point the type here wasn't a primitive type. But if you have a primitive type (which this is since they are just strings), then if cmp.Compare(a,b) == 0 then they are equal.

PR #60088, opened 2024-02-02
Uses an actual Path in FileInfo and avoids an unnecessary clone

kritzcreek:
Sorry, I couldn't let it rest so I had to try to make the change. I still think this makes more sense as the only non-path usage is to check whether the string is empty, which is just as easy with `self.path.to_string_lossy().is_empty()`.

Feel free to close this PR and go with the String one if you disagree

## Test plan

CI

### DIFF
--- a/docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting.rs
+++ b/docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting.rs
@@ -4,9 +4,8 @@ pub mod syntect_scip;
 use anyhow::anyhow;
 use protobuf::Message;
 use std::fmt::{Debug, Display, Formatter};
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use syntect::html::ClassStyle;
-
 use syntect::parsing::{SyntaxReference, SyntaxSet};
 
 pub mod tree_sitter;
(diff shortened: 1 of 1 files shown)


docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting.rs
```diff
@@ -152,20 +136,20 @@ impl<'a> FileInfo<'a> {
...
-struct SublimeLanguageName {
-    raw: String,
+struct SublimeLanguageName<'a> {
+    raw: &'a str,
 }
 
-impl SublimeLanguageName {
-    fn into_tree_sitter_name(self, file_info: &FileInfo<'_>) -> TreeSitterLanguageName {
+impl<'a> SublimeLanguageName<'a> {
+    fn to_tree_sitter_name(&self, file_info: &FileInfo<'_>) -> TreeSitterLanguageName {
```
varungandhi-src, line 152: I don't think saving a few characters in some method calls is worth introducing an extra lifetime here; this also means that the two `Name` types now have different field types.

The `into_tree_sitter_name` function was deliberately consuming `self`, because there isn't a good reason to keep the Sublime name around.

kritzcreek: I'll back this change out

varungandhi-src:
> I still think this makes more sense as the only non-path usage is to check whether the string is empty

That's because you've confined the conversions to the `new` and the methods. 😄 

I guess I don't understand what the benefit is for storing the `PathBuf`; generally I think it's better to use the same type for similar things, so if callers are almost always passing in `&str`, then storing `String` makes it clear when you look at the `struct` definition that the bytes are UTF-8 only. Relatedly, when there is a transient call via `Path` (i.e. `&str -> Path -> &str`), by virtue of it being local, one can easily tell "hey, there isn't going to be invalid character replacement going on here".

varungandhi-src (approved):
Look, I don't care about the PathBuf/String thing enough to litigate more details here. If you're not convinced by my above comment, and you want to merge this, go for it, I don't think it's a big deal either way. I went with String because I thought it was simpler.

PR #60104, opened 2024-02-02
licensing: cleanup logic for conditonally mounting route

### LINKED ISSUE #60105: Refactor web routes definition

BolajiOlajide:
Closes #60105

The previous implementaiton had lots of ternaries - I discovered recently that the `<LegacyRoute>` component had a `condition` prop that uses a boolean logic to mount a route or display a 404, so I'm moving the logic to use that instead as it's cleaner and easier to understand.

## Test plan

On a full license, all routes will work as is.

On a Code Search only license, only code-search features will show up.
On a Cody only license, only cody related features will show up.

### DIFF
--- a/client/web/src/routes.tsx
+++ b/client/web/src/routes.tsx
@@ -9,7 +9,6 @@ import { communitySearchContextsRoutes } from './communitySearchContexts/routes'
 import { type LegacyLayoutRouteContext, LegacyRoute } from './LegacyRouteContext'
 import { PageRoutes } from './routes.constants'
 import { isSearchJobsEnabled } from './search-jobs/utility'
-import { isCodyOnlyLicense, isCodeSearchOnlyLicense } from './util/license'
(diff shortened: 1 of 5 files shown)


client/web/src/routes.tsx
```diff
@@ -307,40 +299,93 @@ export const routes: RouteObject[] = [
...
+                    }, [navigate, chatID])
 
+                    return <div />
+                }}
+                condition={({ licenseFeatures }) =>
+                    !window.location.pathname.startsWith('/cody/chat') && licenseFeatures.isCodyEnabled
+                }
+            />
+        ),
+    },
```
BolajiOlajide, line 331: We should be good to remove this. Waiting for a confirmation from Quinn incase there's something I'm missing.

camdencheek (approved):
Much nicer, thanks!

PR #60108, opened 2024-02-02
Fixup and improve Blame error handling

eseliger:
While investigating why our blob page would not render a nicer error message, I realized that we threw an error where we shouldn't. Then, a full can of worms opened up. So I went out and tried to fix it.

What's fixed here:

- Error reporting for blob not found is now correctly showing the right component again

**Before**

Screenshot 2024-02-02 at 19 47 24@2x

**After**

Screenshot 2024-02-02 at 19 45 15@2x

- When the file does not exist, the Blame endpoint will be called exactly once, and receives a 404 error. 

Screenshot 2024-02-02 at 19 47 51@2x

Previously, it would return a 500 error because we had no handling in place that correctly maps to `os.ErrNotExist`. Also, the error handling in the frontend was not working correctly. We now correctly report the error, treat it as non-retriable, and also render the error. 

Before: 
Screenshot 2024-02-02 at 19 49 55@2x

## Test plan

Added test suites, and manually verified what is shown above.


cmd/gitserver/internal/git/gitcli/blame.go
```diff
@@ -16,12 +16,18 @@ import (
 	"github.com/sourcegraph/sourcegraph/lib/errors"
 )
 
-func (g *gitCLIBackend) Blame(ctx context.Context, path string, opt git.BlameOptions) (git.BlameHunkReader, error) {
-	if err := checkSpecArgSafety(string(opt.NewestCommit)); err != nil {
+func (g *gitCLIBackend) Blame(ctx context.Context, startCommit api.CommitID, path string, opt git.BlameOptions) (git.BlameHunkReader, error) {
```
ggilmore, line 19: I think I am missing something, but why is the addition of the `startCommit` parameter here necessary?

ggilmore (approved):

cmd/gitserver/internal/git/gitcli/blame.go
```diff
@@ -16,12 +16,18 @@ import (
...
+	if err := checkSpecArgSafety(string(startCommit)); err != nil {
 		return nil, err
 	}
 
-	cmd, cancel, err := g.gitCommand(ctx, buildBlameArgs(path, opt)...)
+	// Verify that the blob exists.
+	_, err := g.getBlobOID(ctx, startCommit, path)
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
eseliger: The method already returns nicely wrapped errors that map to specific types (gitdomain.RevisionNotFound), so wrapping here IMO just adds noise and the slight chance that some err type check would not account for unwrapping further down the wire.
LMK if you disagree!

cmd/gitserver/internal/git/gitcli/blame.go
eseliger: Since we verify the commits existence now before running the blame to handle the not found error case, we need to be sure it's always set. In practice, all APIs calling this method set it, but to make that real clear ™️, I went with making it a top-level param vs opt (which are all optional).

eseliger:
Since the remaining comments are about style, I'm going to merge this so we can get the backport started, and I'm happy to follow up to comments in a separate PR that we won't need to backport :)

PR #60111, opened 2024-02-02
search input: Add background color to operators/keywords

fkling:
This PR adds a violet background color to keywords. I chose violet-01 for light mode and violet-05 for dark mode.

I also cleaned this up a bit to avoid repeating color values and moved the colors into CSS variables and applied transparency via SCSS. Maybe the alpha value needs to be tweaked.

Light mode: 

Dark mode:

## Test plan

Visual testing

### DIFF
--- a/client/wildcard/src/global-styles/colors.scss
+++ b/client/wildcard/src/global-styles/colors.scss
@@ -15,6 +15,7 @@ $cyan: #72dbe8;
 
 // Exception that's used in $theme-colors.
 $gray-02: #eff2f5;
+$gray-03: #e6ebf2;
 // Exception that's used in [data-reach-dialog-overlay] background.
 $gray-04: #dbe2f0;
 // Exception that's used in $custom-select-indicator as SVG icon fill color.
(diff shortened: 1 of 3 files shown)


rrhyne (approved):
you're the bomb.

client/wildcard/src/global-styles/colors.scss
```diff
@@ -487,6 +505,12 @@ $theme-colors: (
 
     // Header colors
     --result-header-bg: #22283b;
+
+    // Search input "chip" colors
+    // Transparency needs to be added to make the text selection marker visible
+    --search-input-token-pattern: #{rgba($gray-08, 0.8)};
+    --search-input-token-filter: #{rgba($oc-blue-9, 0.4)};
+    --search-input-token-keyword: #{rgba($violet-05, 0.6)};
```
camdencheek, line 513: Oh, that's an interesting pattern to apply transparency to colors. General question: when should we be using CSS variables vs SCSS variables?

camdencheek (approved):

client/wildcard/src/global-styles/colors.scss
fkling: Depends on when you need to use which values. SCSS variables only exist at build time, CSS variables are shipped to the browser. In this case there isn't any easy way to apply opacity to a hex color value in CSS (there are some ways but might not work in all browsers). `rgba` here is a SCSS function, so the values passed to it have to be available at build time (-> literal or SCSS variable; I don't assume that SCSS can resolve CSS variables at build-time but honestly I haven't tried).

Having said that, I feel like there is a lot going on in our global stylesheets, especially around colors, and it would be great to clean that up some time.

PR #60114, opened 2024-02-02
Search: increase default search limit

camdencheek:
This bumps the default search limit once more from 2000 to 10000. This is the last planned increase. We have already saturated the display limit, so we should not see any significant increase in the payload sizes, which has been the source of any perf issues so far.

## Test plan

Will watch search blitz as this rolls out. As with the previous increases, I do not expect a significant effect on performance.


keegancsmith:
@camdencheek just in case you missed it, the backport failed.

camdencheek:
Ooh, thanks for the ping! I did, in fact, miss this

PR #60116, opened 2024-02-03
Search filters: add approximate count tooltip

camdencheek:
We'd talked about this at some point, but I don't remember where we landed on it. It was trivial to implement so I figured we'd just discuss it on a PR to save some time. I am also simultaneously [increasing the default limit](https://github.com/sourcegraph/sourcegraph/pull/60114) so we should hopefully see this even less. 

It might be a little verbose, would appreciate feedback on the copy. 

CleanShot 2024-02-02 at 17 06 36@2x

## Test plan

Hovered over an approximate count and viewed the tooltip.

### DIFF
--- a/client/branded/src/search-ui/results/filters/components/DynamicFilterBadge.tsx
+++ b/client/branded/src/search-ui/results/filters/components/DynamicFilterBadge.tsx
@@ -2,16 +2,25 @@ import { FC } from 'react'
 
 import classNames from 'classnames'
 
-import { Badge } from '@sourcegraph/wildcard'
+import { Badge, Tooltip } from '@sourcegraph/wildcard'
 
 import styles from './DynamicFilterBadge.module.scss'
 
-export const DynamicFilterBadge: FC<{ exhaustive: boolean; count: number }> = ({ exhaustive, count }) =>
-    count !== 0 ? (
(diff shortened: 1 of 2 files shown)


taiyab:
"This is an approx. count of the results returned because you hit a limit. Try increasing the limit using the `count:` filter in the search query, or select `count:all` from the filter list."

taiyab:
We also need to do a better job of code highlighting in the tooltip.

taiyab:
> > We also need to do a better job of code highlighting in the tooltip.
> > just 1 minute of context about tooltip and text
> 
> It was a big fight between FE people and the Design department about should tooltip render anything but just common text or if it should take some generic UI elements, like formatted text or gif and images preview. Last time design department won and we ended up with just common text)
> 
> cc @taiyab

Sounds like an amazing debate :D

It's probably better we're highly restrictive about what can go into a tooltip, one reason being because people will start using them in-lieu of popovers. I definitely don't wanna see images/gifs in them.

However, it still might be useful (like in our case here) to have things like slightly more obvious code highlighting.

taiyab (commented):
@camdencheek My suggestion is definitely more verbose (and could be simplified down a lot more) but I'm good with that for now.

camdencheek:
@taiyab if you're happy with the copy, would appreciate a stamp 🙂

taiyab (approved):

PR #60120, opened 2024-02-03
licensing: ignore redirect check on dotcom

BolajiOlajide:
## Test plan

Ran `sg start dotcom` and confirmed after login, a user is redirected to the `/search` route.


peterguy (approved):
LGTM

PR #60123, opened 2024-02-03
licensing-ui: cody Icon should not be colored until active

### LINKED ISSUE #60081: Cody Icon should not be colored until active
The Cody Icon in the header should only be "colored" when it's the active tab in that navigation. Otherwise, it's the same monocolor as the other icons.

image

BolajiOlajide:
Closes #60081 

## Test plan

Manual testing.
Added a video for that below.

https://github.com/sourcegraph/sourcegraph/assets/25608335/1bcd19f5-e71d-4630-bc3f-6ff866dbb41b


peterguy (approved):
LGTM

PR #60125, opened 2024-02-03
batches: allow refreshing commit signing apps installation

BolajiOlajide:
This gives site admins the ability to refresh a commit signing app, instead of waiting for the periodic worker (which runs every 24 hours) to kick of the installation sync.

CleanShot 2024-02-14 at 17 51 06@2x

## Test plan

Add a GitHub app on Sourcegraph for Commit Signing.
There should be a `Refresh` button to refresh installations on the app.

If the app is installed on any new org or account, the installations will be added to the database after the `Refresh` operation is done.

### DIFF
--- a/client/web/src/enterprise/batches/settings/CommitSigningIntegrationNode.tsx
+++ b/client/web/src/enterprise/batches/settings/CommitSigningIntegrationNode.tsx
@@ -3,12 +3,14 @@ import React, { useState } from 'react'
 import { mdiCheckCircleOutline, mdiCheckboxBlankCircleOutline, mdiCogOutline, mdiDelete, mdiOpenInNew } from '@mdi/js'
 import classNames from 'classnames'
 
-import { AnchorLink, Button, ButtonLink, H3, Icon, Link, Text } from '@sourcegraph/wildcard'
+import { AnchorLink, Button, ButtonLink, H3, Icon, Link, Text, LoadingSpinner, ErrorAlert } from '@sourcegraph/wildcard'
 
 import { defaultExternalServices } from '../../../components/externalServices/externalServices'
 import { AppLogo } from '../../../components/gitHubApps/AppLogo'
 import { RemoveGitHubAppModal } from '../../../components/gitHubApps/RemoveGitHubAppModal'
 import type { BatchChangesCodeHostFields } from '../../../graphql-operations'
+
+import { useRefreshGitHubApp } from './backend'
 
 import styles from './CommitSigningIntegrationNode.module.scss'
 
@@ -70,7 +72,7 @@ interface AppDetailsControlsProps {
 
 const AppDetailsControls: React.FunctionComponent<AppDetailsControlsProps> = ({ baseURL, config, refetch }) => {
     const [removeModalOpen, setRemoveModalOpen] = useState<boolean>(false)
-
+    const [refreshGitHubApp, { loading, error }] = useRefreshGitHubApp()
     const createURL = `/site-admin/batch-changes/github-apps/new?baseURL=${encodeURIComponent(baseURL)}`
     return config ? (
         <>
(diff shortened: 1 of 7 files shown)


cmd/frontend/internal/githubapp/resolver.go
```diff
@@ -350,5 +359,16 @@ func (r *gitHubAppResolver) syncInstallations() {
...
 		r.logger.Warn("Error syncing GitHub App Installations", log.Error(errs))
+		return errs
+	}
+	return nil
+}
+
+// This method only logs errors rather than assigning them to
+// the resolver because they should not block the request from completing.
+func (r *gitHubAppResolver) syncInstallations() {
+	ctx := context.Background()
```
camdencheek, line 370: We should use `context.WithoutCancel()` here so that we retain all the other information in the context (like tracing information, auth information, etc.)

client/web/src/enterprise/batches/settings/CommitSigningIntegrationNode.tsx
```diff
@@ -140,3 +162,33 @@ const ReadOnlyAppDetails: React.FunctionComponent<ReadOnlyAppDetailsProps> = ({
...
+const NodeAlert: React.FunctionComponent<React.PropsWithChildren<NodeAlertProps>> = ({
+    visible,
+    children,
+    variant,
+}) => {
+    const ref = useRef<HTMLDivElement>(null)
+    const style = useSpring({
+        height: visible ? `${(ref.current?.offsetHeight || APPROX_BANNER_HEIGHT_PX) + ONE_REM_IN_PX}px` : '0px',
+        opacity: visible ? 1 : 0,
+    })
```
camdencheek, line 184: This is...hacky looking 😄 What's going on here?

cmd/frontend/internal/githubapp/resolver.go
BolajiOlajide: TIL

client/web/src/enterprise/batches/settings/CommitSigningIntegrationNode.tsx
BolajiOlajide: Lmaooo. It's a copy-pasta solution to having a drawer effect for the alert. This tells `react-spring` how to display the `animated.div` which gives this a nice-looking animation.

cmd/frontend/internal/githubapp/resolver.go
BolajiOlajide: Does `context.WithCancel` retain these information also? I think I want to use that instead, since this is run in a goroutine, we don't want this to exceed more than 1 minute.

camdencheek: Yeah, the problem with `context.WithCancel` is that if the parent context is canceled (e.g., the user's request times out or is canceled), the child context will also get canceled. Based on your code comment and the fact that `syncInstallations` does not have a context argument, I assumed that you wanted this to run in the background even if the request returns and the request context is canceled

BolajiOlajide: Yes, that's correct but [a timeout of 1 minute was added originally](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/cmd/frontend/internal/githubapp/resolver.go?L330). It's an opportunisitic run, we don't need this to take a long time.

client/web/src/enterprise/batches/settings/CommitSigningIntegrationNode.tsx
BolajiOlajide: See that nice looking animation lol.

CleanShot 2024-02-14 at 18 15 56

I've made it more verbose so folks can understand what's going on a bit.

cmd/frontend/internal/githubapp/resolver.go
camdencheek: > And since the context is newly created in the syncInstallations method, it wouldn't affect the goroutine if the request gets canceled

Right, but since the context is newly-created, it also doesn't include any of the other information carried in a context like auth information or tracing information. New contexts generally should not be created in response to a user-initiated request.

BolajiOlajide: CleanShot 2024-02-14 at 18 20 52@2x

Chatted about this with @camdencheek in the DM. We ended up with this approach, added comments for whoever is here next.

PR #60127, opened 2024-02-03
v2 telemetry: add v2 telemetry to src/api

dadlerj:
## Test plan

1. `sg start`
2. visit page
3. check if events appear in event_logs table locally


dadlerj:
The first of many to come. I'll be working through one folder under client/web/src at a time. Please let me know if you see any issues with this model @bobheadxi, otherwise I'll power through these.

I used a lot of resources to get familiar with this structure and build this, but https://github.com/sourcegraph/sourcegraph/pull/57939 most of all.

akalia25 (approved):

PR #60128, opened 2024-02-04
Fixing Hubspot Form Proxy issue

arafatkatze:
Solves https://github.com/sourcegraph/cody/issues/2557 

## Test plan

Use this branch   `ara/test-hubspotfix` and see this loom video to see how I replicated the failure mode of hubspot and how to solve it.

https://www.loom.com/share/ecc0ea965bba4b33863233d4292e7526?sid=311c71f3-ac84-4387-be3f-b0bf1f16cdd5 

I basically blocked the website on my own computer and then saw the behavior in the cases with or without the fix.  

I am still trying to learn frontend development so feel free to make more suggestions on how to improve this. If you believe there is better methods to test this then I am all for it. 

Thank you!!!


arafatkatze:
image
@thenamankumar Percy seems to have an issue with the Cody icon which seems bizarre and unrelated.

client/web/src/marketing/components/HubSpotForm.tsx
```diff
@@ -244,33 +244,62 @@ export const HubSpotForm: FunctionComponent<HubSpotFormProps> = ({
...
-                    onFormReady?.(form)
-                    onHubsportFormReady?.(form, userId, userEmail)
-                },
-                onFormSubmitted,
-                onFormSubmit,
-                inlineMessage,
-            })
+            // Load all scripts
+            // eslint-disable-next-line @typescript-eslint/no-floating-promises
+            loadAllScripts()
```
thenamankumar, line 284: I think we should still call `loadAllScripts()` outside the `if(scriptsLoaded) {...}` because in the case, that nothing is blocked, actually loading scripts won't be blocked by first checking if the scripts are load-able.

client/web/src/marketing/components/HubSpotForm.tsx
```diff
@@ -244,33 +244,62 @@ export const HubSpotForm: FunctionComponent<HubSpotFormProps> = ({
...
+                    })
+            )
+        )
+            .then(() => setScriptsLoaded(true))
+            .catch(() => {
+                // Mark a loading error that returns a minimal react component as a failure mode
+                setLoadError(true)
+                const emptyElement = document.createElement('div')
+                // Do this callback method to increase the value of step
+                onFormSubmitted?.(emptyElement)
```
thenamankumar, line 270: Calling `onFormSubmitted` on failure seems incorrect. Should we not accept additional prop `onFormLoadError` and use it for this purpose? We don't even need to create the empty div in that case.

client/web/src/marketing/components/HubSpotForm.tsx
```diff
@@ -244,33 +244,62 @@ export const HubSpotForm: FunctionComponent<HubSpotFormProps> = ({
...
+            })
+    }, [onFormSubmitted])
```
thenamankumar, line 272: Rather than making additional HTTP requests to verify if the script exists. Can we try with `<script onreadystatechange={cb} />` or  `<script onload={cb} />`callbacks? 

Reference: https://stackoverflow.com/questions/37762047/javascript-onload-and-script-callback-functions-which-takes-the-precedence

I am unsure if that will work, as I have not used it earlier myself, but it seems like a better approach than making additional HHTP requests. 

But let's timebox that, if it will take longer to fix it with the above-mentioned approach then we can merge this PR as a temp hacky fix, to fix UX right now.

thenamankumar (changes requested):
I have left some comments regarding:
1. making additional HTTP requests to verify loading of scripts vs using something like `<script onreadystatechange={cb} />`.
2. calling `onFormSubmitted` on form load error. There should be a separate cb `onFormLoadError` for this case. 


While 2. is a requested change. If it takes longer to fix the 1., then we can merge this solution as a temporary fix to unblock user onboarding and then continue with improving it if possible.

chenkc805 (approved):

chenkc805:
Thanks Ara, just to make sure I understand the end behavior of this PR:
- If a user has an ad blocker that's blocking HubSpot, we'll just skip over the Hubspot form entirely
- If a user doesn't have an ad blocker that's blocking HubSpot, we'll show them the form as normal

Did I get that right?

Also, is there a reason why the HubSpot form loads so slowly? There's like a 1 second delay on clicking "Hi" and then getting the HubSpot form. Can we make that load faster, like load it async in the background when the user first loads the "Hi" modal?

chenkc805 (commented):
Left comments

client/web/src/marketing/components/HubSpotForm.tsx
arafatkatze: The purpose of `onFormSubmitted` is skipping the counter so that we can move ahead to the editor step [here](https://sourcegraph.com/-/editor?remote_url=https%3A%2F%2Fgithub.com%2Fsourcegraph%2Fsourcegraph.git&branch=ara%2Fhubspot-fix&file=client%2Fweb%2Fsrc%2Fcody%2Fonboarding%2FCodyOnboarding.tsx&editor=VSCode&version=2.2.16&start_row=169&start_col=9&end_row=189&end_col=14).

See the `onNext` link

client/web/src/marketing/components/HubSpotForm.tsx
arafatkatze: Makes sense, I changed that.

client/web/src/marketing/components/HubSpotForm.tsx
arafatkatze: I am not opposed to the idea but testing it on different browsers etc might take a while and I am committed to a few tasks at the moment. That being said I am totally okay if that were to be a subsequent addon(assuming it works fine for our case. )

arafatkatze:
@chenkc805 
> If a user has an ad blocker that's blocking HubSpot, we'll just skip over the Hubspot form entirely
Yes

> If a user doesn't have an ad blocker that's blocking HubSpot, we'll show them the form as normal
Yes exactly, we only skip when a Hubspot URL can't be reached from the users browser(that can happen for any reason whatsoever)

>  is there a reason why the HubSpot form loads so slowly? There's like a 1 second delay on clicking "Hi" and then getting the HubSpot form. Can we make that load faster, like load it async in the background when the user first loads the "Hi" modal?

I agree with you and you are right, the loading happens AFTER the user goes through the "hi" modal although it can theoretically happen async. I wish I could take that on but its a time consuming task for now and I have some commitments so I spun up an issue for it.

chenkc805:
Thanks. Yeah that's fine to address the slow form load in a separate issue.

chenkc805 (approved):

arafatkatze:
@thenamankumar I solved the suggested comments

thenamankumar:
@arafatkatze you should be able to approve on Percy yourself. Sign in with Okta.

arafatkatze:
Naman PErcy isn't in Okta
image

But for some reason it got auto approved so we are good.

arafatkatze:
@thenamankumar If there are no blockers on this can you approve so this automerges?

thenamankumar (approved):

PR #60130, opened 2024-02-04
gomod: update or vendor buildkit, docker, hostmatcher and saml to resolve CVEs

keegancsmith:
We had multiple CVEs reported for these dependencies. I don't think this affected us in practice, but this is a step towards a clean scan from trivy. I updated to the minimum version which supports the fix if that worked otherwise latest.

``` shell
go get github.com/moby/buildkit@v0.12.5
go get github.com/docker/docker@v24.0.7
go get -u github.com/crewjam/saml
```

In the case of code.gitea.io/gitea@v1.18.0/modules/hostmatcher we couldn't update it due to lots of issues popping up in random transitive dependencies. However, we don't depend on the whole gitea project, rather just a tiny self contained package in it. So we vendor it in.

Test Plan: CI and "trivy fs go.mod" reporting no issues.


cmd/blobstore/internal/blobstore/blobstore_test.go
```diff
@@ -297,9 +297,11 @@ func initTestStore(ctx context.Context, t *testing.T, dataDir string) (uploadsto
...
 		TTL:          168 * time.Hour,
 		S3: uploadstore.S3Config{
-			Region:       "us-east-1",
-			Endpoint:     ts.URL,
-			UsePathStyle: false,
+			Region:          "us-east-1",
+			Endpoint:        ts.URL,
+			UsePathStyle:    false,
+			AccessKeyID:     "TESTACCESSKEY",
+			SecretAccessKey: "TEST/SECRET/ACCESSKEY",
```
keegancsmith, line 304: @slimsag @eseliger I am asking you as potential blobstore experts. The new version of the s3 SDK requires that the credentials are not empty. This is since the credentials are used to sign the request. Are there cases where we do not configure credentials for blobstore? From what I can tell we always configure this, but I want to confirm.

Example failure from our unit tests:

```
blobstore_test.go:310: Init failed to create bucket: operation error S3: CreateBucket, failed to sign request: failed to retrieve credentials: static credentials are empty
```

keegancsmith: Alright nevermind this will break stuff in practice. Searching for `UPLOAD_BACKEND=blobstore` straight away I see examples of us not setting creds: https://sourcegraph.sourcegraph.com/github.com/sourcegraph/deploy-sourcegraph-docker@ab267a5dfe402af2f450e6250b774a3a967eff58/-/blob/docker-compose/docker-compose.yaml?L145-146

PR #60145, opened 2024-02-05
[dotcom] add refresh token endpoint with sams authenticator

thenamankumar:
This is a follow-up PR of the closed PR: https://github.com/sourcegraph/sourcegraph/pull/60061

It exposes a temporary endpoint on dotcom to refresh users's cody rate limits. It uses SAMS service-to-service tokens for auth. 

This new endpoint will be used by SSC to trigger rate limit refresh whenever there is a subscription change. 

Related SSC PR: https://github.com/sourcegraph/accounts.sourcegraph.com/pull/418

The approach has been recommended and approved under the thread: https://sourcegraph.slack.com/archives/C05SZB829D0/p1706806647849469

## Test plan

Tested locally with the test plan mentioned here: https://github.com/sourcegraph/sourcegraph/pull/60061


vdavid (approved):
All parts look fine to me.

cmd/cody-gateway/internal/auth/sams.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+)
+
+// SAMSAuthenticator provides an auth middleware that uses SAMS service-to-service tokens to authenticate the requests.
+type SAMSAuthenticator struct {
+	Logger     log.Logger
+	SAMSClient sams.Client
+}
+
+func (a *SAMSAuthenticator) Middleware(requiredScopes []sams.Scope, next http.Handler) http.Handler {
```
chrsmith, line 23: Is this required for this PR? My understanding is that we were only going to use SAMS auth for the `ssc->dotcom` portion of this. But for the `dotcom->Gateway` leg, we were going to just keep that working the same as it does today. (Using whatever existing method dotcom pushes subscription update events to Cody Gateway.)

Right? If this is required to add to the Cody Gateway service, could you elaborate on that in the PR description to explain why we are adding this new functionality? If it is optional, or to "inch towards making Cody Gateway SAMS aware", then I would suggest teasing that out into a separate PR.

cmd/frontend/internal/httpapi/httpapi.go
```diff
@@ -179,6 +183,32 @@ func NewHandler(
...
+
+		dotcomConf := conf.Get().Dotcom
+
+		samsClient := sams.NewClient(
+			dotcomConf.SamsServer,
+			clientcredentials.Config{
+				ClientID:     dotcomConf.SamsClientID,
+				ClientSecret: dotcomConf.SamsClientSecret,
+				TokenURL:     fmt.Sprintf("%s/oauth/token", dotcomConf.SamsServer),
+				Scopes:       []string{"openid", "profile", "email"},
```
chrsmith, line 195: This would be correct IFF the SAMS client were just verifying the tokens it receives from SSC. But if the expectation is that it will send requests to Cody Gateway (which I don't think it should, but some of the required code is in this PR), then it would require the `client.cody-gateway` scope.

cmd/frontend/internal/httpapi/ssc.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/sourcegraph/sourcegraph/internal/ssc"
+)
+
+// newSSCRefreshCodyRateLimitHandler returns an http.Handler to trigger cody's rate limit refresh for a user
+// TODO(sourcegraph#59625) remove as part of adding SAMSActor source
+func newSSCRefreshCodyRateLimitHandler(logger log.Logger, db database.DB) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+
+		samsUserID := mux.Vars(r)["samsUserID"]
```
chrsmith, line 20: Do not use "SAMS user ID" as that is not correct. We should only be sending the "SAMS External Account ID", the publicly facing UUID to identify a SAMS account. Which is separate from the internal, and private "SAMS user ID".

Being precise in the terms, while annoying, will cut down on mistakes down the road.

cmd/frontend/internal/httpapi/ssc.go
```diff
@@ -0,0 +1,57 @@
...
+		if samsUserID == "" {
+			w.WriteHeader(http.StatusBadRequest)
+			w.Write([]byte("missing uuid"))
```
chrsmith, line 23: Any reason why we shouldn't use `http.Error`? It's built specifically for this sort of thing.
https://pkg.go.dev/net/http#Error

```go
http.Error(w, "missing UUID", http.StatusBadRequest)
```

cmd/frontend/internal/httpapi/ssc.go
```diff
@@ -0,0 +1,57 @@
...
+			w.Write([]byte(err.Error()))
+			return
+		}
+
+		if len(oidcAccounts) == 0 {
+			w.WriteHeader(http.StatusNotFound)
+			return
+		}
+
+		userID := oidcAccounts[0].UserID
```
chrsmith, line 47: Just for clarity, would it be helpful to rename this to `dotcomUserID`?

internal/sams/client.go
```diff
@@ -0,0 +1,102 @@
...
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/clientcredentials"
+)
+
+type Scope = string
```
chrsmith, line 16: This should be written as `type Scope string`. Using the `Scope = string` syntax, Golang will not complain about certain types of conversions. But we want to force those to be errors when expecting writing code that deals with the `sams.Scope` type.

internal/sams/client.go
```diff
@@ -0,0 +1,102 @@
...
+	)
+	if err != nil {
+		return nil, errors.Wrap(err, "create introspection request")
+	}
+
+	request.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+
+	// Create the OAuth2 HTTP client. This will handle setting up the HTTP headers based on the token source.
+	// (And potentially issue a new token as needed.)
+	tokenSource := c.clientCredentialsConfig.TokenSource(ctx)
```
chrsmith, line 71: This is incorrect. It will not "potentially issue a new token as needed". Because we are creating a _new_ `oauth2.TokenSource` on every call to `IntrospectToken`, we will reissue a new SAMS token on every call.

We should instead, not have the `samsClient` have a `clientCredentialsConfig` field, but instead an `oath.TokenSource` field. So that the `TokenSource` is long-lived, allowing us to benefit from reusing access tokens.

schema/site.schema.json
```diff
@@ -2125,6 +2125,18 @@
...
+          "type": "string",
+          "description": "The server URL for SAMS instance."
+        },
+        "sams.clientID": {
+          "type": "string",
+          "description": "The clientID for SAMS instance."
+        },
+        "sams.clientSecret": {
+          "type": "string",
+          "description": "The clientSecret for SAMS instance."
```
chrsmith, line 2138: Hey @unknwon , does this look right to you?

Essentially we need to add some configuration data so that in the portions of dotcom that interact directly with SAMS know how to contact it. (Sounds reasonable.) But this `server`, `clientID`, and `clientSecret` data is _already_ available via the Authentication Provider configuration for the Sourcegraph instance... so this seems kind of redundant.

Would there ever be a situation where we are not using SAMS for authenticating users, but _would_ use SAMS for sending requests to Cody Gateway? Maybe this would be the case for Sourcegraph Cloud / Managed Instances?

chrsmith (dismissed):
This seems like it is on the right track, but two things seem problematic:

- Are the changes to `/cmd/cody-gateway` required for this PR? Or just left over from an earlier change set?

- Here we are introducing a new `internal/sams` package, which we plan on using for dotcom to wire up authorization handlers. Great. But within the `internal/ssc` we also have a SAMS client... or some thin approximation of one.

It seems like we should work to unify that so there is a single `sams.Client` that can be passed around and used as needed. (Since both are requiring the same client ID and secret, both need to use the same hostname, etc. But both are using two entirely different methods for obtaining that information.)

I would also nudge us towards having a more proper `sams.Client` interface since it's something we would reuse again when it comes to sending SAMS-based requests to Cody Gateway from the dotcom backend. In other words, if we don't do this right, we could end up with 3x versions of the same SAMS boilerplate in the codebase!

cmd/cody-gateway/internal/auth/sams.go
thenamankumar: This file should have been deleted. removed now. I moved the client & middleware to `internal/sams/auth.go`

cmd/frontend/internal/httpapi/httpapi.go
thenamankumar: Only to verify tokens

cmd/frontend/internal/httpapi/ssc.go
thenamankumar: I'll rename it to `samsAccountID`, dotcom knows only about the account id in the externalAccounts table also named as `account_id` so that should work without any confusion.

cmd/frontend/internal/httpapi/ssc.go
thenamankumar: in dotcom's codebase context calling a normal userID "dotcomUserID" will be like me talking about myself in 3rd person.

schema/site.schema.json
thenamankumar: 2 services can talk to each other without having an openidconnect with SAMS. like in future if we need to integrate gateway with ssc, gateway do not implement any opened connect and won't have auth provider config. In my opinion, it makes sense to have separate config params and not couple it with auth provider config.

service-to-service auth has nothing to do with openidconnect auth provider config.

internal/sams/client.go
thenamankumar: done

internal/sams/client.go
thenamankumar: updated client to store both `clientCredentialsConfig` and `tokenSource` separately, providing flexibility for future usecase. `func NewClient` still only accepts `clientCredentialsConfig` to keep the usage simple.

cmd/frontend/internal/httpapi/ssc.go
chrsmith: Yes Naman, you are absolutely correct. But I would argue that in my experience, that isn't enough to stop very smart programmers from getting it wrong and confusing something like this. So even if it looks weird (e.g. "talking about myself in the 3rd person"), I would suggest that in cases like this we might want to err on the side look clarity.

It's OK if you disagree, there isn't a clearly right or wrong answer. But the thing I'm looking for is just imagining someone not super familiar with the codebase were trying to make an edit here. How likely would they be able to do it correctly. That's all. 👍

chrsmith:
> I have addressed all the feedback except having separate config args, which if required can be change separately. Dismissing review to unblock. Will address any additional feedback separately.

Thanks @thenamankumar for pushing forward with this. I agree the config issue isn't worth blocking on. (I'll leave a comment in that particular thread, but it seems like we should just keep that as-is anyways.)

schema/site.schema.json
chrsmith: So to paraphrase, we are saying:

- Yeah, we are specifying the same client credentials in two places.
- No, we would not consider creating "two OAuth apps" for dotcom, we'd just use one.  (For both machine-to-machine, and user-facing logins.)

If that's the case, then it probably isn't worth it to rework the configuration here. (It's just that we are wiring the credentials in two different places, to be used in two different contexts.) 

I'll stop bothering you all then :D

unknwon: >If that's the case, then it probably isn't worth it to rework the configuration here. (It's just that we are wiring the credentials in two different places, to be used in two different contexts.)

To be fair, creating two sources for the identical credentials could be confusing, but maybe it's fine since we will move away from dotcom and then can delete this `dotcom.xxx` thing altogether when that happens.

daxmc99:
@thenamankumar @chrsmith 
This PR is likely causing panics on our cloud instances because 
https://github.com/sourcegraph/sourcegraph/blob/main/cmd/frontend/internal/httpapi/httpapi.go#L194 can be `nil` on non-dotcom instances (I assume)

chrsmith:
Thanks @daxmc99 , will send PR ASAP.

PR #60155, opened 2024-02-05
Implement go-git backend

eseliger:
## Preview 🤩
[Preview Link](https://docs.sourcegraph.com/@es/gogit-backend)


keegancsmith:
can you give some background motivation for this? FYI in my testing go-git isn't that performant and quite heavy on the GC.

eseliger:
Background: Just an experiment, nothing more 😬 
It shows that our new backend interface can be quite pluggable and that it'll allow us to find more optimized versions of what we're currently doing, where helpful. 
This is not intended to be merged, and is also quite incomplete. I just found it a fun exercise to play with this for not even 2h and not break my sg instance :) 

If we ever would go with go-git, we'd do real performance comparisons. For very fast things, it helps get around the majority of the overhead: process spawning. But often, git is more optimized.

PR #60161, opened 2024-02-05
Uses Result over panics in scip evaluation

kritzcreek:
`anyhow` allows us to use Result instead of panics in most places, and lets us attach more context to errors. This way we can eg. print file paths we failed to read, rather than just "Failed to read"

## Test plan

Code continues to compile


keynmol (approved):
LMBTM (looks much better to me), thanks!

PR #60165, opened 2024-02-05
codeintel: Downgrade queue size critical alert to warning

varungandhi-src:
We've been running into spurious alerts on-and-off.

We should add observability here to better narrow down why
we're getting backlogs that are getting cleared later.

In the meantime, it doesn't make sense for this to
be a critical alert.

## Test plan

n/a

### DIFF
--- a/doc/admin/observability/alerts.md
+++ b/doc/admin/observability/alerts.md
@@ -2221,7 +2221,7 @@ Generated query for critical alert: `min((sum by (app) (up{app=~".*(pgsql|codein
 
 **Descriptions**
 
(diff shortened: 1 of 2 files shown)


keynmol (approved):

keynmol:
Seems like you need to run a bazel command to regeneraet the markdown docs: https://buildkite.com/sourcegraph/sourcegraph/builds/260667#018d79a8-10fb-47ba-bda8-b46e51acf1fe/42-265

varungandhi-src:
I was afraid that running that would cause problems with the check that you can't modify files under the docs directories. But I've run it now, let's see

PR #60169, opened 2024-02-05
search: Don't show drop shadow for when autofocused

fkling:
@rrhyne mentioned to remove the dropshadow when the input is autofocused. Personally it feels weird to me that the shadow suddenly appears after user interaction.

https://github.com/sourcegraph/sourcegraph/assets/179026/b06a1026-8c45-46be-95b6-8808b754cd1c

## Update: Without background color change

https://github.com/sourcegraph/sourcegraph/assets/179026/cf1aa98d-5829-4d05-89b6-ec05c1e065a8

## Test plan

Manual testing

### DIFF
--- a/client/branded/src/search-ui/input/experimental/CodeMirrorQueryInputWrapper.module.scss
+++ b/client/branded/src/search-ui/input/experimental/CodeMirrorQueryInputWrapper.module.scss
@@ -98,13 +98,16 @@
     left: 0;
     right: 0;
     border-radius: 8px;
-    background-color: var(--color-bg-1);
 
     // Set a default paddings to the suggestion panel (see Suggestions.module.scss)
     --suggestions-padding: 0.75rem;
 
+    // Only apply the "container" style if the suggestions are open, as indicated
+    // by the "open" class. When autofocus is enabled we do not want the suggestions
+    // container to be visible.
     &.open {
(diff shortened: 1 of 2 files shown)


vovakulikov (commented):
If we go this way maybe we should autofocus but don't render any from suggestions UI instead just dropping shadow? It makes sense to me because this is how usually some pickers/suggestions UI work, you don't see any suggestions until you type something in the search field. This empty suggestion panel looks like a very big focus ring in my opinion.

fkling:
Yeah, I can update the PR to include that, but I think others should actually try to see how it feels.

camdencheek:
+1 to "feels weird that the shadow only shows up after interacting." I personally like the really emphasized drop shadow on autofocus. It reminds me that the bar is now autofocused so I don't reflexively reach for my mouse

vovakulikov:
>  I personally like the really emphasized drop shadow on autofocus

Just to be clear I think autofocus should render a "blue" focus ring regardless that we are in the initial state or not. If focus ring isn't highlighted enough this might a good reason to revisit here, drop black shadow shouldn't be a factor here (again my humble opinion)

rrhyne:
After trying it out, I'd say this is working really, really well now. Meets my expectations.

rrhyne (approved):
After trying it out, I'd say this is working really, really well now!

PR #60186, opened 2024-02-05
Gateway /v1/attribution allowed for BYOK customers

### LINKED ISSUE #59996: Gateway attribution needs to work with product subscriptions with BYOK

cbart:
Closes #59996

Adding extra logic to gateway to allow product subscription auth without Cody enabled (BYOK customers) to access attribution endpoint.

## Test plan

Auth gateway unit test added.


arafatkatze (approved):
PR works and enables auth even in the edge case where  Cody gateway is not enabled.

cmd/cody-gateway/internal/actor/productsubscription/productsubscription.go
```diff
@@ -254,6 +254,9 @@ func newActor(source *Source, token string, s dotcom.ProductSubscriptionState, i
 		ID:            s.Uuid,
 		Name:          name,
 		AccessEnabled: !disallowedLicense && !s.IsArchived && s.CodyGatewayAccess.Enabled,
+		EndpointAccess: map[string]bool{
+			"/v1/attribution": !disallowedLicense && !s.IsArchived,
+		},
```
rafax, line 259: I understand that this is a "P0 for Feb 15th" item, so we don't have other options, but I really think this should live on the ProductSubscription side of the world, with CodyGatewayAccess enabled set to true, and some "feature" marker indicating that Attribution is on (hacky way is setting rate limits to 0).

rafax (approved):

cmd/cody-gateway/internal/actor/productsubscription/productsubscription.go
cbart: Yep. I'll file an issue to track transitioning it.

PR #60189, opened 2024-02-05
Keyword search: add 'beta' labels and fix URL parsing

jtibshirani:
This PR adds 'Beta' labels to the keyword search banner and popover, as we've
decided to ship the feature as 'beta'.

It also sneaks in a bug fix: before, if the URL did not contain a `patterntype`
parameter, we always defaulted to `standard`. Now, we use the correct default
based on the settings. This is important because users often configure a custom
search engine to search Sourcegraph (so for example they can type
`sg "some query"` in the chrome search bar). Now, these searches use `keyword`
search when the feature is enabled, whereas before they always used `standard`.

## Test plan

Added new unit test for bug fix

### DIFF
--- a/client/web/src/storm/pages/SearchPage/KeywordSearchCtaSection.tsx
+++ b/client/web/src/storm/pages/SearchPage/KeywordSearchCtaSection.tsx
@@ -5,7 +5,7 @@ import classNames from 'classnames'
 
 import { useExperimentalFeatures } from '@sourcegraph/shared/src/settings/settings'
 import { useTemporarySetting } from '@sourcegraph/shared/src/settings/temporary'
-import { Code, H2, Icon, Link, Text } from '@sourcegraph/wildcard'
+import { Code, H2, Icon, Link, ProductStatusBadge, Text } from '@sourcegraph/wildcard'
 
 import { MarketingBlock } from '../../../components/MarketingBlock'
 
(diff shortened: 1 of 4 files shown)


jtibshirani:
Here's what the banner and popover look like now 
Screenshot 2024-02-05 at 1 00 26 PM
Screenshot 2024-02-05 at 1 00 18 PM

client/web/src/storm/pages/SearchPage/KeywordSearchCtaSection.tsx
```diff
@@ -19,7 +19,7 @@ interface KeywordSearchCtaSection {
 
 export const KeywordSearchCtaSection: React.FC<KeywordSearchCtaSection> = ({ className }) => {
     const keywordSearchEnabled = useExperimentalFeatures(features => features.keywordSearch)
-    const [isDismissed = true, setIsDismissed] = useTemporarySetting('search.homepage.keywordCta.dismissed', false)
+    const [isDismissed = true, setIsDismissed] = useTemporarySetting('search.homepage.keywordCta.dismissed', true)
```
keegancsmith, line 22: i don't feel qualified enough to know if this was an intentional change or something you did for testing and forgot to undo.

keegancsmith (approved):

client/web/src/storm/pages/SearchPage/KeywordSearchCtaSection.tsx
jtibshirani: Thank you for catching this! This was unintentional, just had a sleepy day yesterday.

PR #60190, opened 2024-02-05
Update Access token defaults

chwarwick:
Update default access token setting for allowing tokens with no expiration to true.  The no expiration option will never be the default selected option in the UI and a warning is displayed to discourage selecting no expiration.  The site config option remains to disable creation of tokens with no expiration. 
## Test plan
Updated unit tests
Manual test of token creation and defaults.


daxmc99 (approved):

daxmc99:
Related to https://github.com/sourcegraph/sourcegraph/pull/59565

PR #60192, opened 2024-02-06
v2 telemetry: add v2 telemetry to src/auth

dadlerj:
Context:

## Test plan

`sg start`
visit page

(This is actually harder with this commit given some of these pages are hard to reach, so more testing would be helpful)

### DIFF
--- a/client/web/src/auth/SignInPage.tsx
+++ b/client/web/src/auth/SignInPage.tsx
@@ -5,6 +5,7 @@ import classNames from 'classnames'
 import { partition } from 'lodash'
 import { Navigate, useLocation, useSearchParams } from 'react-router-dom'
 
+import { TelemetryV2Props } from '@sourcegraph/shared/src/telemetry'
 import { Alert, Icon, Text, Link, Button, ErrorAlert, AnchorLink, Container } from '@sourcegraph/wildcard'
 
 import type { AuthenticatedUser } from '../auth'
@@ -20,7 +21,7 @@ import { UsernamePasswordSignInForm } from './UsernamePasswordSignInForm'
 
 import styles from './SignInPage.module.scss'
 
-export interface SignInPageProps {
+export interface SignInPageProps extends TelemetryV2Props {
     authenticatedUser: AuthenticatedUser | null
     context: Pick<
         SourcegraphContext,
@@ -38,7 +39,10 @@ export interface SignInPageProps {
 
 export const SignInPage: React.FunctionComponent<React.PropsWithChildren<SignInPageProps>> = props => {
     const { context, authenticatedUser } = props
-    useEffect(() => eventLogger.logViewEvent('SignIn', null, false))
+    useEffect(() => {
+        eventLogger.logViewEvent('SignIn', null, false)
+        props.telemetryRecorder.recordEvent('auth.sign-in', 'view')
+    }, [props.telemetryRecorder])
 
     const location = useLocation()
     const [error, setError] = useState<Error | null>(null)
@@ -189,6 +193,7 @@ export const SignInPage: React.FunctionComponent<React.PropsWithChildren<SignInP
                 allowSignup={context.allowSignup}
                 isRequestAccessAllowed={isRequestAccessAllowed}
                 sourcegraphDotComMode={context.sourcegraphDotComMode}
+                telemetryRecorder={props.telemetryRecorder}
             />
         </>
     )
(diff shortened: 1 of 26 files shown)


dadlerj:
This change is ready for review. 

It unfortunately ended up being more complex than expected—I ended up refactoring logging in our `ExternalsAuth` and `SignUpForm` components to avoid double logging events.

One question for @bobheadxi : I see [in the docs that we recommend camel case for event features and actions](https://docs.sourcegraph.com/dev/background-information/telemetry/protocol#event), but looking at our [Cody code](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/cody%24%40a1afc07+%27cody.&patternType=keyword&sm=0), I exclusively see "dash case" features. Should we standardize on the latter now given that?

dadlerj:
Also, I'd love to get to a point where you don't have to review every PR... for now, I'd just love your eyes on this since it's the first sizable webapp frontend one I'm making. Let me know if you want me to start looping in any other teams to get up to speed on this transition so they can review my changes from here on.

bobheadxi:
> Let me know if you want me to start looping in any other teams to get up to speed on this transition so they can review my changes from here on.

@dadlerj I think this might be worth looping in Erika on - it would be valuable to have engineers with experience in the web app be able to provide their input. Sadly we don't have an obvious owner for this like "frontend platform", the way we have the Dev Infra team for backend things.

bobheadxi:
> One question for @bobheadxi : I see [in the docs that we recommend camel case for event features and actions](https://docs.sourcegraph.com/dev/background-information/telemetry/protocol#event), but looking at our [Cody code](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/cody%24%40a1afc07+%27cody.&patternType=keyword&sm=0), I exclusively see "dash case" features. Should we standardize on the latter now given that?

@dadlerj I think you want to look specifically at `recordEvent` calls, because a lot of your search results are not telemetry and are other parts of the code. For example: https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/cody$@a1afc07+recordEvent('cody.&patternType=keyword&sm=0 - this search result indicates most V2 events _are_ using the camelcase convention.

I would recommend sticking to the camelcase convention, but ultimately it doesn't _really_ matter, as long as we use the structures well (prefixing with `cody.`, and using good values for `feature` and `action`, such that we can create useful groupings/indexing on them)

client/web/src/auth/SignInPage.tsx
```diff
@@ -227,16 +232,26 @@ const ProviderIcon: React.FunctionComponent<{ serviceType: AuthProvider['service
...
+    telemetryRecorder,
+}) => {
     const dotcomCTAs = (
         <>
             <Link
                 to="https://sourcegraph.com/get-started?t=enterprise"
-                onClick={() => eventLogger.log('ClickedOnEnterpriseCTA', { location: 'SignInPage' })}
+                onClick={() => {
+                    eventLogger.log('ClickedOnEnterpriseCTA', { location: 'SignInPage' })
+                    telemetryRecorder.recordEvent('auth.enterprise-cta', 'click')
```
bobheadxi, line 253: I assume this retains the v1 logger becasue we use the v1 event in dotcom, is that right? If so we might want to add a TODO here with a comment about this, and indicate it will be removed later

bobheadxi (approved):
Everything LGTM, and thanks for going the extra mile with the safe-enum-values of a few attributes! Also good to see that some events were just removed it seems :)

Approve to unblock. I would personally prefer to see the camel case convention adopted here, since https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/cody$@a1afc07+recordEvent(%27cody.&patternType=keyword&sm=0 indicates _most_ Cody events adhere to camel case already, but strictly speaking this is not blocking - as long as we're using `feature` and `action` values in a useful manner (which for the most part seems to be the case here).

Thanks again for picking this up!

dadlerj:
>@dadlerj I think you want to look specifically at recordEvent calls, because a lot of your search results are not telemetry and are other parts of the code. For example: [https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/cody$@a1afc07+recordEvent('cody.&patternType=keyword&sm=0](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/cody$@a1afc07+recordEvent(%27cody.&patternType=keyword&sm=0) - this search result indicates most V2 events are using the camelcase convention.

🤦  I totally missed this. Yep, this is an easy fix.

> Also good to see that some events were just removed it seems :)

In response to this and your question above, I actually _didn't_ remove any v1 events... My thinking is that until I do all of the transition/v1 removal stuff from the project tracker (e.g. update in-product analytics, update the queries that generate our ping data to use v2 event names, etc), I can't safely remove v1 events, even from self-hosted instances. 

Also, there's something nice about having them both running at the same time for comparison.

My thinking was that the v2 addition and v1 removal would be two separate exercises, and that v1 would only be removed once v2 was fully operational.

What's your take on that approach? Apologies if I missed a discussion about this from earlier in the project.

client/web/src/auth/SignInPage.tsx
dadlerj: Hmm interesting question—see my question below for follow up on this:https://github.com/sourcegraph/sourcegraph/pull/60192#issuecomment-1939735504

bobheadxi:
> In response to this and your question above, I actually _didn't_ remove any v1 events (it was more of a refactor)... My thinking is that until I do all of the transition/v1 removal stuff from the project tracker (e.g. update in-product analytics, update the queries that generate our ping data to use v2 event names, etc), I can't safely remove v1 events, even from self-hosted instances.
> 
> Also, there's something nice about having them both running at the same time for comparison.
> 
> My thinking was that the v2 addition and v1 removal would be two separate exercises, and that v1 would only be removed once v2 was fully operational.
> 
> What's your take on that approach? Apologies if I missed a discussion about this from earlier in the project.

@dadlerj my original concern was that unlike Cody clients, where the customer only receives the V2 events while dotcom receives the V1 events for us to use, the in-instance web app telemetry really does send both events to the backend. This means that as long as we have a v1+v2 setup in the web app, the Sourcegraph instance will actually have double the events. In earlier discussions my recommendation was that each PR to change events does so "end to end": i.e. updates all dependencies in-application, such as in-product analytics

Ultimately though it might not be that problematic, especially if we eventually get to a v2-only state before the next `main` release, but I would be wary about landing a v1+v2 state in-instance for, say, Sourcegraph 5.3.0 or 5.4.0

PR #60204, opened 2024-02-06
code nav: Respect history.preferAbsoluteTimestamps user setting

### LINKED ISSUE #51209: RFE: Enable absolute timestamps in commit history
Requested by https://github.com/sourcegraph/accounts/issues/8205

On the main repo page, we currently have relative timestamps for commit history.
example

Same scenario when looking at a commit.
example 2

With the `history.preferAbsoluteTimestamps` setting, expand the functionality to commit history as well.

Similar to what we've achieved here: https://github.com/sourcegraph/sourcegraph/issues/31712

fkling:
Closes #51209

This commit changes the logic in the file list panel to respect the `history.preferAbsoluteTimestamps` setting. The issue originally referenced the commits panel but that was removed in favor of the combined file/commit panel.

I also noted that `GitCommitNode` is used in most places but the callsites do not provide the value from the setting. Rather than make every callsite read the user settings the component itself uses settings as a fallback when the prop is not explicitly set.

## Test plan

Manual testing.

### DIFF
--- a/client/web/src/repo/tree/TreePagePanels.module.scss
+++ b/client/web/src/repo/tree/TreePagePanels.module.scss
@@ -73,14 +73,14 @@ table.files {
     }
 
     .commit-date-column {
-        /* stylelint-disable-next-line declaration-property-unit-allowed-list */
-        width: 120px;
+        // Width of the column that fits the text "Last commit date" without overflow
+        width: 7rem;
 
         // When showing absolute dates by default we have to give the column
-        // more space to show the full date + time.
+        // more space to show the full date + time. This width fits absolute
+        // timestamps without overflow.
         &.absolute {
-            /* stylelint-disable-next-line declaration-property-unit-allowed-list */
-            width: 160px;
+            width: 10rem;
         }
     }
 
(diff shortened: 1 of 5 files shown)


client/web/src/repo/tree/TreePagePanels.module.scss
```diff
@@ -75,6 +75,13 @@ table.files {
     .commit-date-column {
         /* stylelint-disable-next-line declaration-property-unit-allowed-list */
         width: 120px;
+
+        // When showing absolute dates by default we have to give the column
+        // more space to show the full date + time.
+        &.absolute {
+            /* stylelint-disable-next-line declaration-property-unit-allowed-list */
+            width: 160px;
```
camdencheek, line 83: why do we disallow `px`? It seems to be a pretty noisy lint given that we use it quite a bit around the codebase 🤔

camdencheek (approved):

vovakulikov (approved):

client/web/src/repo/tree/TreePagePanels.module.scss
fkling: Using pixels shouldn't be a problem with browser zoom (see screenshot) but I can change this `rem` too.

PR #60215, opened 2024-02-06
search: update client parser to match backend changes

stefanhengl:
Relates to #60157

This updates the client parser to match the recent changes in the backend. Most importantly, we treat '(...)' as group by default with the exception of `()` which we treat as literal.

Example:

Screenshot 2024-02-06 at 17 32 34

Screenshot 2024-02-06 at 17 31 41

Test plan:
Manual testing


camdencheek (approved):
I honestly don't know this code too well, but this looks fine to me and testing it locally works as described for all my test cases.

cc @fkling since I think you've worked with this code before

stefanhengl:
Talked with @fkling over Zoom. This is ready to merge. While testing, we found more, but unrelated, bugs in our parser which I will fix in a follow-up PR.

PR #60216, opened 2024-02-06
Search filters: visual updates

taiyab:
A bunch of visual updates and tweaks to the filter panel.

## Before
image

## After
image

image

## Test plan
- Manual testing


camdencheek (approved):
LGTM! The more compact spacing looks nice

vovakulikov:
I know that this is probably too late for this kind of comment but it just strikes me, that we decided to have this reset button next to the title because we don't want to take additional vertical space with a sticky reset filter button below the filters. But it seems that this "Filter results" title takes even more vertical space and we render it unconditionally (which means we always take this vertical space from actual valuable filter content)

Also personal feedback, this title seems a bit redundant here (it also forces as to have title over filter types to avoid confusion, which also takes some vertical space)

So my question - do we really need this title and reset button on the top, can we revert to the idea of having this in the footer? cc @taiyab

taiyab:
> I know that this is probably too late for this kind of comment but it just strikes me, that we decided to have this reset button next to the title because we don't want to take additional vertical space with a sticky reset filter button below the filters. But it seems that this "Filter results" title takes even more vertical space and we render it unconditionally (which means we always take this vertical space from actual valuable filter content)
> 
> Also personal feedback, this title seems a bit redundant here (it also forces us to have title over filter types to avoid confusion, which also takes some vertical space)
> 
> So my question - do we really need this title and reset button on the top, can we revert to the idea of having this in the footer? cc @taiyab

The title is also providing more clarity to what you're actually looking at in the LHS. User testing is showing that people are potentially confusing the filters with search on first glance.

Let's try it for a day or so, and we can revert it if it's something we're not liking.

vovakulikov:
I think the reset button also doesn't match with baseline of the filter type title 
I can see what I can do here, but again I think it would be simpler to have it full-size body at the footer 

Screenshot 2024-02-06 at 14 26 19

taiyab:
> Does it bother us that this difference in panels and line height visually doesn't match up Screenshot 2024-02-06 at 14 19 41

Not currently, no.

Let's review today and see if we like it.

PR #60217, opened 2024-02-06
[Cody Manage] make ssc url configurable for testing

thenamankumar:
closes: https://github.com/sourcegraph/sourcegraph/issues/60212

We are blocked from testing complete SSC integration on https://cody-services-dev.sgdev.dev/cody/manage due to hard-coded sams URL in the code. 

For a quick fix, this PR makes the URL origin toggle between `accounts.sourcegraph.com` and `accounts.sgdev.org` using feature flag `cody-payments-testing-mode`

Ideally yes, it should be configurable through site config. But that will take a bit longer to implement and unblock testing. 

## Test plan
tested locally.

### DIFF
--- a/client/web/BUILD.bazel
+++ b/client/web/BUILD.bazel
@@ -220,7 +220,7 @@ ts_project(
         "src/cody/components/ScopeSelector/backend.ts",
         "src/cody/components/ScopeSelector/index.tsx",
         "src/cody/components/ScopeSelector/useRepoSuggestions.ts",
-        "src/cody/featurFlags.ts",
+        "src/cody/featureFlags.ts",
         "src/cody/isCodyEnabled.tsx",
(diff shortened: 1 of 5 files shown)


client/web/src/cody/featurFlags.ts
```diff
@@ -11,3 +11,10 @@ export const useHasTrialEnded = (): boolean => {
 
     return ended
 }
+
```
vdavid, line 14: Not sure where to flag it, but there is a typo in the name of this file

thenamankumar: fixing.

vdavid (approved):
Nice. Feels so good to have this dynamic.

client/web/src/cody/featurFlags.ts
thenamankumar: done

PR #60219, opened 2024-02-06
v2 telemetry:  add v2 telemetry to src/codeintel

dadlerj:
Context:

## Test plan

`sg start`
visit page


dadlerj:
Need to figure out how to deal with this same issue on this PR: https://github.com/sourcegraph/sourcegraph/pull/60192#issuecomment-1928583035

dadlerj:
Waiting on https://github.com/sourcegraph/sourcegraph/pull/60192 to be merged to get the changes to TryCodyWidget.tsx in before this is ready to review

client/web/src/repo/components/TryCodyWidget/TryCodyWidget.tsx
```diff
@@ -58,7 +58,25 @@ function useTryCodyWidget(telemetryService: TelemetryProps['telemetryService']):
...
+        const eventType = provider === 'builtin' ? 'form' : provider
+        const eventPage = type === 'blob' ? 'Blobview' : 'RepositoryPage'
+        const eventArguments = {
+            type: eventType,
+            page: eventPage,
+            description: '',
+        }
+        telemetryService.log(EventName.AUTH_INITIATED, eventArguments, eventArguments)
+        const v2EventPage = type === 'blob' ? 0 : 1
+        telemetryRecorder.recordEvent('auth', 'initiate', { type: V2AuthProviderTypes[provider], page: v2EventPage })
```
vovakulikov, line 77: @dadlerj so if I tracked this right this is where this problem with types comes from. 

I'm not familiar with our new telemetry recorder but it seems like that third parameters with arbitrary 
arguments should be send as metadata property. So It would look like this

```
telemetryRecorder.recordEvent('auth', 'initiate', {
    metadata: { type: V2AuthProviderTypes[provider], page: v2EventPage },
})
```

client/branded/src/search-ui/components/CopyPathAction.tsx
```diff
@@ -16,13 +17,15 @@ export const CopyPathAction: React.FunctionComponent<
...
-> = ({ filePath, className, telemetryService }) => {
+    } & TelemetryProps &
+        TelemetryV2Props
+> = ({ filePath, className, telemetryService, telemetryRecorder }) => {
     const [copied, setCopied] = useState(false)
 
     const onClick = (event: React.MouseEvent<HTMLButtonElement>): void => {
         event.preventDefault()
         telemetryService.log('CopyFilePath')
+        telemetryRecorder.recordEvent('search.header.file-path', 'copy')
```
vovakulikov, line 28: I see that this PR is still in draft mode but just to make sure we're on the same page 
if we change the event's names on the client we also should take care about new names
on the backend as well (in sql event aggregations, migrations for big query for amplitude) 
otherwise it will break existing looker and amplitude dashboards?

client/web/src/repo/components/TryCodyWidget/TryCodyWidget.tsx
dadlerj: Thanks! This was the error I mentioned up above in https://github.com/sourcegraph/sourcegraph/pull/60219#issuecomment-1930938074—I was able to track this down in the other PR but hadn't updated here yet.

client/branded/src/search-ui/components/CopyPathAction.tsx
dadlerj: Yes, this is an important project—cleaning up the old v1 events. It'd be nice to do them as I go, but it's a big project. I'm going to do it selectively, focusing on removing our highest volume events, and push to get v2 set up sooner without fully removing v1. You can see the [workplan](https://docs.google.com/document/d/17REfT9UCu3nANKEIRBUw3HiNpiTYhg5gikMclqMuWaU/edit) I have shows these as two separate steps, but they will probably get merged/happen fluidly given the risk of overloading local databases.

vovakulikov (approved):
I see, CI is finally green, looks like mergin main helped, which makes sense since they passed locally on my machine. Thanks for your patient @dadlerj

Issue #59946, opened 2024-01-29
Search Results: Scroll position jumps if you scroll slow enough 


vovakulikov:
[Original slack thread](https://sourcegraph.slack.com/archives/C05EA9KQUTA/p1706228797872909)

If you scroll slow enough to the top there is a point when the whole scroll container jumps to the top, it doesn't happen thought if you scroll fast, there is some problem with out virtual list and react visible sensor.

Turning off partial visibility in visibility sensor fixes this problem for me [https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@668f19c11e7a00490c3[…]ient/shared/src/components/VirtualList.tsx?L66:25-66:43](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@668f19c11e7a00490c348e90fe4c050e5c0c1267/-/blob/client/shared/src/components/VirtualList.tsx?L66:25-66:43)

But we should research why exactly this happens in this UI

https://github.com/sourcegraph/sourcegraph/assets/18492575/00fc12f8-4f6d-4032-bc4c-8e424cfd0ed5

PR #60222, opened 2024-02-06
Search Result: Fix search result scroll jumps

vovakulikov:
Fixes https://github.com/sourcegraph/sourcegraph/issues/59946

So the problem was that it was our intention logic about preserving focus as we render/hide elements in the virtual list. The problem is that as you focus on element browsers by default try to scroll to the focusable element on the page, this overlaps with our user scroll and as a result, we had jumps. 

This PR still preserves logic we had about saving focus on the first element but doesn't scroll to it when it shows up. 

## Test plan
- Open the search result page with enough results to have a scrollbar on the page
- Scroll to the bottom 
- Scroll to the top slowly 
- Check that there are no scroll jumps when your first search result element shows up in the visible scrollable area


camdencheek (approved):
Good find! Thanks for tracking this down

PR #60226, opened 2024-02-06
Fix panic on startup

chrsmith:
A recent PR introduced a new API specific to the Cody Pro codepath (https://github.com/sourcegraph/sourcegraph/pull/60145). However, the way it pulls the "dotcom configuration" will be `nil` if not set. (Which is the case since this PR just landed.)

Unfortunately this leads to a `nil`-deref panic on startup. We'll now start up gracefully if the configuration is missing, and just not register the new endpoints.

## Test plan

NA.


daxmc99 (approved):
Thanks!

Issue #52118, opened 2023-05-17
UI is hung when search results include a large single line Javascript file


mike-r-mclaughlin:
- **Sourcegraph version:** 5.0.3
- **Reported by:** https://github.com/sourcegraph/accounts/issues/6938

#### Steps to reproduce:

This seems to be caused by the client-side syntax highlighting when there is a very large single-line Javascript file.

1. Search for something that will have a result in a large single line Javascript file.  For example: `context:global lang:javascript todo file:\.min\.js select:content` [[example](https://sourcegraph.com/search?q=context:global+lang:javascript+todo+file:%5C.min%5C.js+select:content&patternType=standard&sm=1&groupBy=repo)]
    * This will usually cause the the issues shown below.  If you don't spot it on the first try, you can add a large `count:` value to see it as well

#### Expected behavior:

Search results do not cause the UI to hang or crash regardless of their size

#### Actual behavior:

![image](https://github.com/sourcegraph/sourcegraph/assets/23640224/e86429eb-69b1-46ba-9565-81d231e22109)
![image](https://github.com/sourcegraph/sourcegraph/assets/23640224/29835f2e-5b03-4b6c-a1f8-8877a0d93671)

If you would like immediate help on this, please email support@sourcegraph.com (you can still create the issue, but there are no [SLAs](https://about.sourcegraph.com/support/) on issues like there are for support requests).

camdencheek:
Came across another instance of this organically. We should probably set a limit on the max length of lines that we request highlighting for since the highlight ranges can add a ton of complexity to the DOM

PR #60228, opened 2024-02-06
search: introduce max-line-len streaming parameter

keegancsmith:
max-line-len is a new parameter which when set will truncate the Content field of ChunkMatch such that no line is longer than max-line-len. Currently we only set it for interactive queries in the webapp with the conservative number 5kb.

A new field is introduced to ChunkMatch when its content has been truncated. The webapp detects this and will skip syntax highlighting files which contain truncated ChunkMatches. The webapp will still highlight the non-truncated ranges.

We have noticed that some documents are close to our max index size (about 1mb), and are all 1 line. This can lead to response sizes in the 100s of megabytes. This both is slow for the browser to receive and parse, as well as causing locking up/freezing of the browser.

The name max-line-len was chosen as a trade-off between brevity and exactness. I like short key names for URL params that are nearly always set.

Two alternative approaches were considered. The first is something along the lines of display limit which omits large events. This had the downside of potentially displaying nothing or having hard to understand semantics to avoid that. The second approach was naive truncation. The downside here is you miss a lot of potential ranges you could highlight and not having per line information lead to results looking broken.

Test Plan: In my dev environment I:
- lowered the max-line-len to 8
- introduced console logs for skipping syntax highlighting
- introduced an artifical delay to syntax highlighting

The delay was not as important with the final approach I ended up with, but was kept in testing due to its usefulness to notice how the app behaves when that is slow.

I then observed that highlighting on the non-truncated parts still worked. There is one minor artifact of highlighting the text "...truncated", but that seems acceptable.

Screenshot 2024-02-06 at 15 14 26

https://github.com/sourcegraph/sourcegraph/assets/187831/a35b569a-5867-40d3-898e-4c4af5686e6b

Part of https://github.com/sourcegraph/sourcegraph/issues/58815

### DIFF
--- a/client/shared/src/search/stream.ts
+++ b/client/shared/src/search/stream.ts
@@ -100,6 +100,13 @@ export interface ChunkMatch {
     content: string
     contentStart: Location
     ranges: Range[]
+
+    /**
+     * Indicates that content has been truncated.
+     *
+     * This can only be true when maxLineLength search option is non-zero.
+     */
+    contentTruncated?: boolean
 }
 
 export interface SymbolMatch {
@@ -487,10 +494,27 @@ export interface StreamSearchOptions {
     featureOverrides?: string[]
     searchMode?: SearchMode
     sourcegraphURL?: string
-    displayLimit?: number
     chunkMatches?: boolean
     enableRepositoryMetadata?: boolean
     zoektSearchOptions?: string
+
+    /**
+     * Limits the number of matches sent down. Note: this is different to the
+     * count: in the query. The search will continue once we hit displayLimit
+     * and updated filters and statistics will continue to stream down.
+     *
+     * If unset all results are streamed down.
+     */
+    displayLimit?: number
+
+    /**
+     * Truncates content strings such that no line is longer than
+     * maxLineLength. This is used to prevent sending large previews down to
+     * the browser which can cause high CPU and network usage.
+     *
+     * If unset full Content strings are sent.
+     */
+    maxLineLen?: number
 }
 
 function initiateSearchStream(
@@ -504,6 +528,7 @@ function initiateSearchStream(
         featureOverrides,
         searchMode = SearchMode.Precise,
         displayLimit = 1500,
+        maxLineLen,
         sourcegraphURL = '',
         chunkMatches = false,
     }: StreamSearchOptions,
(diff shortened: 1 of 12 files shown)


client/shared/src/search/stream.ts
```diff
@@ -504,6 +528,7 @@ function initiateSearchStream(
         featureOverrides,
         searchMode = SearchMode.Precise,
         displayLimit = 1500,
+        maxLineLen,
```
camdencheek, line 531: Should we add a default here?

camdencheek (approved):
In the client, what happens if we have matched ranges after the truncation? I assume we just don't show them, but I want to make sure we don't have any logic that will behave poorly with that

client/shared/src/search/stream.ts
keegancsmith: hah, I have pending comment I forgot to submit when I created this. I intentionally did this and asked your opinion:

@camdencheek unrelated to this PR, but in stream call we have a default set for displayLimit. It seems like we should be default have no limit set, and instead each call site decides its limit. For example that is what I did here for this param. WDYT?

IE other clients use the stream search functions (eg jetbrains), setting defaults that make sense for the webapp seems wrong? But I could be totally off base here.

keegancsmith:
@fkling I saw a few bits of copy-pasta in the sveltekit for chunkmatches into groupmatches. I didn't update the sveltekit stuff to use this logic, but I suspect we want the same stuff there?

fkling:
@keegancsmith Probably. We'll need to sync other changes to the React to SvelteKit anyway. I'll create a doc/issue to keep track of that. Thanks for pointing that out!

client/shared/src/search/stream.ts
camdencheek: Okay, yeah, I think I agree. We probably shouldn't set defaults unless they can be defaults for most clients.

camdencheek:
Okay, I just tested the "what happens if there are highlights past the truncation limit" and it just doesn't highlight, which seems fine. This change makes it way snappier! Thanks for cranking this out!

camdencheek:
This also fixes https://github.com/sourcegraph/sourcegraph/issues/52118

keegancsmith:
> In the client, what happens if we have matched ranges after the truncation? I assume we just don't show them, but I want to make sure we don't have any logic that will behave poorly with that

Here is the behaviour at different max-line-len. I think this is acceptable behaviour?

#### 5k
image

#### 50

image

#### 20
image

PR #60238, opened 2024-02-06
gitserver: Fix goroutine leak when reader is not drained

eseliger:
When not all messages are read from a reader (for example, if an archive operation times out, or if a ReadFile request is canceled, or the reader is only interested in the first X lines), we didn't properly Wait() on the command.

The added test case proves that before the code changes to the reader, there is 1 goroutine too much still active in the runtime. After this change, it's equal, across a few hundred runs.

## Test plan

Added test case.

### DIFF
--- a/cmd/gitserver/internal/git/gitcli/clibackend.go
+++ b/cmd/gitserver/internal/git/gitcli/clibackend.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os/exec"
 	"strconv"
+	"sync"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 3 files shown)


ggilmore (approved):
LGTM, I do wonder how / when we call close if a context cancellation occurs

eseliger:
> LGTM, I do wonder how / when we call close if a context cancellation occurs

I think the following: 

```
	r, err := backend.ReadFile(ctx, api.CommitID(req.GetCommit()), req.GetPath())
	if err != nil {
		if os.IsNotExist(err) {
			s, err := status.New(codes.NotFound, "file not found").WithDetails(&proto.FileNotFoundPayload{
				Repo:   req.GetRepoName(),
				Commit: req.GetCommit(),
				Path:   req.GetPath(),
			})
			if err != nil {
				return err
			}
			return s.Err()
		}
		var e *gitdomain.RevisionNotFoundError
		if errors.As(err, &e) {
			s, err := status.New(codes.NotFound, "revision not found").WithDetails(&proto.RevisionNotFoundPayload{
				Repo: req.GetRepoName(),
				Spec: e.Spec,
			})
			if err != nil {
				return err
			}
			return s.Err()
		}
		gs.svc.LogIfCorrupt(ctx, repoName, err)
		// TODO: Better error checking.
		return err
	}
	defer r.Close()

	w := streamio.NewWriter(func(p []byte) error {
		return ss.Send(&proto.ReadFileResponse{Data: p})
	})

	_, err = io.Copy(w, r)
	return err
```

Here, the `r` is `io.Copy`-ed into the streamio.Writer. When the client cancels the request, the `ss.Send` function will ultimately error.
Then, the io.Copy returns the err, the function returns, and in defer we call r.Close - which at that point is a "premature" close, because we mightn't have read all content from the reader.

Issue #60172, opened 2024-02-05
Update empty repository filter display


rrhyne:
The empty repository filter is an improvement, but the copy could leave users feeling the repository was not found in the Sourcegraph instance. We need to make it much more clear that it was not found in the results only, and provide users a simple route to getting the repository they need:

## Before

![image](https://github.com/sourcegraph/sourcegraph/assets/539268/37f78389-8ad8-42b3-b3db-1da4e6d0aacb)

## Improved 

![image](https://github.com/sourcegraph/sourcegraph/assets/539268/971e66e2-c2af-41af-8a9a-2744fd35c9eb)

Figma: https://www.figma.com/file/EPgl0qVbqGSumvf52brhda/Code-Search?type=design&node-id=1493-16090&mode=design&t=kZi8GVeAPRfiyFv7-4

taiyab:
Tweaked this a bit more (same Figma link): 

<img width="318" alt="image" src="https://github.com/sourcegraph/sourcegraph/assets/5337876/9404097a-bc7b-4872-959b-b730c4f9ef44">

PR #60239, opened 2024-02-07
Search filters: update empty state for dynamic filter searches

### LINKED ISSUE #60195: Search filters: refine input empty state
cc @rrhyne 

Updates to https://github.com/sourcegraph/sourcegraph/pull/60179

image

https://www.figma.com/file/EPgl0qVbqGSumvf52brhda/Code-Search?type=design&node-id=1493%3A16077&mode=design&t=AuvsLW65e35cRgRT-1

It should:
- only give the filter recommendation for repository and author filters
- only give the filter recommendation when a count limit is hit

I also feel this confusion from users seeing the repo filter as a search is somewhat exacerbated by:
- currently not having the "filter by" title at the top (which is coming)
- having repository as the first filter option in the list

Fallback alternative for non-repo/author filters, or when the reqs. for recommendation aren't hit:
image

@camdencheek Would be great if you could review this to see if it makes sense?

camdencheek:
This implements the changes in https://github.com/sourcegraph/sourcegraph/issues/60195, with a couple of small tweaks described below.

First, I modified the copy because "all repositories" is not accurate since we will be limiting to only the repository they entered.

Before: "Try searching all repositories explicitly by adding `repo:hand` to your original search query" 
After: "Try adding `repo:hand` to your original search query narrow results to that repo."

Second, I modified the text in the button to only link `search bar` because text cannot wrap within a button, so it was making the alignment weird if the linked text was larger.

[Loom walkthrough](https://www.loom.com/share/fce0b1d138774f2b8b305f3419bbba75?sid=e6a4865c-e394-4bb6-845a-118abadc45f5)

Fixes #60195 
Fixes https://github.com/sourcegraph/sourcegraph/issues/60172

## Test plan

Manual testing.

### DIFF
--- a/client/branded/src/search-ui/results/filters/components/dynamic-filter/SearchDynamicFilter.module.scss
+++ b/client/branded/src/search-ui/results/filters/components/dynamic-filter/SearchDynamicFilter.module.scss
@@ -67,18 +67,37 @@
 }
 
 .description {
+    &-header {
+        font-weight: 600;
+        margin-bottom: 0.5rem;
+    }
+
     // margin: 0 0.65rem -0.25rem 0.65rem;
-    background-color: var(--secondary);
+    background-color: var(--secondary-2);
     color: var(--text-muted);
     padding: 0.75rem 1rem;
     border-radius: 0.5rem;
+}
+
+.zero-state-query-button {
+    display: inline;
+    padding: 0 2px;
+    border-radius: 3px;
+    background-color: var(--primary-4);
+    font-size: inherit;
+    text-align: unset;
+    color: var(--text-muted);
+    font-weight: normal;
+    margin: 0;
+    border: none;
+    vertical-align: unset;
 }
 
 .zero-state-search-button {
     text-decoration: underline;
     display: inline;
     padding: 0;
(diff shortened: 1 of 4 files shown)


camdencheek:
> Maybe change to:

Do'h. Thanks 🤦

client/branded/src/search-ui/results/filters/components/dynamic-filter/SearchDynamicFilter.module.scss
```diff
@@ -67,18 +67,37 @@
...
-    background-color: var(--secondary);
+    background-color: var(--secondary-2);
     color: var(--text-muted);
     padding: 0.75rem 1rem;
     border-radius: 0.5rem;
 }
 
+.zero-state-query-button {
+    display: inline;
+    padding: 0 2px;
```
vovakulikov, line 84: Probably our linter won't like px units here, should be 0.125rem instead

vovakulikov (approved):
Thanks, nice UI and UX change

rrhyne (approved):
Looks great, thanks Camden.

PR #60248, opened 2024-02-07
search: mention "..." in alert for unbalanced parens

stefanhengl:
For keyword search we support quoting literals. This makes the content filter unnecessary. Here we change the copy of the alert to recommend using quotes instead of the content: filter.

**standard search**
image

**keyword search**
image

Test plan:
manual testing


jtibshirani (approved):
Good catch.

Issue #59873, opened 2024-01-25
Chat UI mentions that <10 lines is not subject to attribution search


cbart:

cbart:
cc @toolmantim

toolmantim:
First past suggestion: when it passes due to being a small number of lines we change the pass title tooltip to include a note in parentheses: `${passedLabel} (< 10 line threshold)`

The word “threshold” here should be whatever we decide to call that setting.

toolmantim:
I don’t entirely love “threshold” or that title tho — open to suggestions or wordsmithing!

toolmantim:
I think it’s important we just show it passed as normal though, as the admin/org has decided it’s fine. No need for any difference aside from the tooltip.

PR #60250, opened 2024-02-07
snippetAttribution responds with search lower bounds, and whether it ran at all

cbart:
Part of #59873

This pull request adds two parameters that present to the client whether attribution search was run and what the lower bound of lines is (if any).

## Test plan

- [x] tested locally with GraphQL browser
- [x] graphQL API unit tests for both dotcom and enterprise


chwarwick (approved):
This looks good.  Remember the 5.3 branch is already cut I assume this is intended to be a part of that so it needs the backport label.

PR #60252, opened 2024-02-07
Make perforce email matching case insensitive

### LINKED ISSUE #58757: Perforce permissions email matching is case sensitive
- **Sourcegraph version:** v4+
- **Platform information:** Any Sourcegraph deployment when syncing with Perforce code hosts

#### Steps to reproduce:
1. Configure a Perforce code host, with `"authorization": {}` configured to sync and enforce user permissions
2. Configure a user in Perforce and Sourcegraph, with the same email address, but different capitalization (ex. `Firstname.Lastname@domain.tld` vs `firstname.lastname@domain.tld`)
3. Notice that this Sourcegraph user won't be matched with their permissions in Perforce, so they won't see code from the Perforce code host
4. Change the capitalization of their email address on either Sourcegraph or Perforce to match the other
5. Notice that this Sourcegraph user will now be matched with their permissions in Perforce, so they will now see code from the Perforce code host
6. Note that many large enterprises have long histories of changing email address conventions (i.e. their email addresses are a hot mess, and longstanding staff have many email addresses on their LDAP accounts), and Perforce customers can't use a centralized identity provider to batch update the email address on Perforce user accounts.

#### Expected behavior:
1. Email addresses are not case sensitive anywhere else, so they shouldn't be in Sourcegraph
2. Email addresses from both the Perforce users table and the Sourcegraph user should be normalized to lowercase before permissions recording / searching

#### Actual behavior:
1. Capitalization breaks Perforce permissions

/cc @mrnugget

pjlast:
Closes #58757 

Makes perforce email matching case insensitive.

## Test plan

Adjust and add unit test.


eseliger (commented):
Does that also cover the definitions in the p4 protects file getting matched against a users emails?

pjlast:
@eseliger yeah the protects file uses a username if I'm not mistaken. The email is to set up the external account

eseliger (approved):
gotcha, alright! (I know very little about how these things link up, but I trust you here and code and test LGTM) :)

Issue #60218, opened 2024-02-06
Update SAMS config for dotcom prod


thenamankumar:

chrsmith:
Sorry Naman, I was not able to get to this today.

thenamankumar:
Update: https://sourcegraph.slack.com/archives/C05GJPTSZCZ/p1707300512404019

PR #60254, opened 2024-02-07
Add default value for ssc.apiBaseURL

thenamankumar:
Add default value for ssc.apiBaseURL site config param. The default value was mentioned in the JSON schema but not applied. 

fixing this under https://github.com/sourcegraph/sourcegraph/issues/60218

## Test plan

verified locally.

### DIFF
--- a/internal/ssc/ssc.go
+++ b/internal/ssc/ssc.go
@@ -152,12 +152,17 @@ func NewClient() (Client, error) {
 		return &client{}, errors.New("no SAMS authorization provider configured")
 	}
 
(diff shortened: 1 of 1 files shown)


vdavid (approved):
Approving to unblock, but is this the correct place for setting the default? Can't we set one for sgconf.SscApiBaseUrl?

thenamankumar:
@vdavid I wish we could directly set the default in the config struct but all that go code is generated. 

We only update the json schema file and can control it.

PR #60255, opened 2024-02-07
Hide free trial tile if cc added

thenamankumar:
closes: https://github.com/sourcegraph/accounts.sourcegraph.com/issues/444
image

Right now, the tile should go away once the trial period ends. 

But this PR also hides it for the users who put in their CC even before the trial ends.

## Test plan
- sign up for pro trial
- add credit card
- the free trial tile should not be visible


vdavid (approved):
Oh yeah, thank you! <3

PR #60258, opened 2024-02-07
gitserver: Return RevisionNotFoundError early from NewFileReader

eseliger:
This feels more aligned with what other endpoints do. I checked that callers should not depend on the previous behavior.

Also added some additional handling for empty streams, as I noticed along the way that io.EOF errors would sometimes be erroneously returned.

## Test plan

Added a test, existing CI for integration tests.


internal/gitserver/commands.go
```diff
@@ -830,39 +830,40 @@ func (c *clientImplementor) StreamBlameFile(ctx context.Context, repo api.RepoNa
...
-
 		s, ok := status.FromError(err)
-		if !ok {
-			cancel()
-			endObservation(1, observation.Args{})
+		if ok {
+			if s.Code() == codes.PermissionDenied {
+				cancel()
+				endObservation(1, observation.Args{})
+				return nil, &os.PathError{Op: "open", Path: path, Err: os.ErrNotExist}
```
pjlast, line 838: Just double checking that this is the intended error to return for `PermissionDenied` ? If so, should we perhaps add that to the comment as well? "If the file doesn't exist, or if the provided actor is not allowed to access the file, the returned error will pass the os.IsNotExist() check."

Something like that?

pjlast (approved):

PR #60262, opened 2024-02-07
Autocomplete attribution requires site config setting but can be individually disabled with feature flag

### LINKED ISSUE #59824: Autocomplete can be disabled with feature flag, but requires site config setting.
This is for safety in case guardrails breaks autocomplete. We should be able to disable it on the back-end.

cbart:
Closes #59824

Attribution for autocomplete is implemented in the back-end and can currently be enabled by using `attribution.enabled: true` site config. However for safety we allow disabling attribution just for autocomoplete by setting `autocomplete-attribution` feature flag to `false`.

## Test plan

- [ ] Unit tests
- [x] Manual check

### DIFF
--- a/internal/completions/httpapi/handler.go
+++ b/internal/completions/httpapi/handler.go
@@ -256,7 +256,7 @@ func newStreamingResponseHandler(logger log.Logger, db database.DB, feature type
 		}
 		f := guardrails.NoopCompletionsFilter(eventSink)
 		if conf.GetConfigFeatures(conf.SiteConfig()).Attribution &&
-			featureflag.FromContext(ctx).GetBoolOr("autocomplete-attribution", false) {
+			featureflag.FromContext(ctx).GetBoolOr("autocomplete-attribution", true) {
 			ff, err := guardrails.NewCompletionsFilter(guardrails.CompletionsFilterConfig{
 				Sink:             eventSink,
 				Test:             test,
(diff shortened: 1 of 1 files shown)


internal/completions/httpapi/handler.go
```diff
@@ -255,7 +255,8 @@ func newStreamingResponseHandler(logger log.Logger, db database.DB, feature type
 			l.Error("attribution error", log.Error(err))
 		}
 		f := guardrails.NoopCompletionsFilter(eventSink)
-		if featureflag.FromContext(ctx).GetBoolOr("autocomplete-attribution", false) {
+		if conf.GetConfigFeatures(conf.SiteConfig()).Attribution &&
+			featureflag.FromContext(ctx).GetBoolOr("autocomplete-attribution", false) {
```
cbart, line 259:
```suggestion
			featureflag.FromContext(ctx).GetBoolOr("autocomplete-attribution", true) {
```

internal/completions/httpapi/handler.go
```diff
@@ -255,7 +255,8 @@ func newStreamingResponseHandler(logger log.Logger, db database.DB, feature type
 			l.Error("attribution error", log.Error(err))
 		}
 		f := guardrails.NoopCompletionsFilter(eventSink)
-		if featureflag.FromContext(ctx).GetBoolOr("autocomplete-attribution", false) {
+		if conf.GetConfigFeatures(conf.SiteConfig()).Attribution &&
```
keegancsmith, line 258: it looks like GetConfigFeatures is a bit of a naughty API and can return nil. Let me make this more robust for ya so we can get this in before our release branch shutsdown a bit more.
```suggestion
		if cf := conf.GetConfigFeatures(conf.SiteConfig()); cf != nil && cf.Attribution &&
```

Issue #60270, opened 2024-02-07
Separate feature flags for SSC release on web and clients


thenamankumar:
Update feature flag on `use-ssc-for-cody-subscription` to something else on web, so that it is different from the one used on clients and we can control the two releases separately.

PR #60271, opened 2024-02-07
Update SSC release feature flag name

thenamankumar:
closes: https://github.com/sourcegraph/sourcegraph/issues/60270

Change feature flag name `use-ssc-for-cody-subscription` to `use-ssc-for-cody-subscription-on-web`, so it is different from the one used on clients and we can control the two releases separately.

## Test plan
 unit tests covers it.

### DIFF
--- a/client/web/src/cody/featureFlags.ts
+++ b/client/web/src/cody/featureFlags.ts
@@ -1,7 +1,7 @@
 import { useFeatureFlag } from '../featureFlags/useFeatureFlag'
 
 export const useArePaymentsEnabled = (): boolean => {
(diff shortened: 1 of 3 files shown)


vdavid (approved):
@thenamankumar, can you confirm that you went through all mentions of the old flag and updated all places where it made sense? I assume you've done that so approving, just looking for a confirmation.

thenamankumar:
Yes @vdavid I did confirm by doing a grep and search: https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/sourcegraph$+use-ssc-for-cody-subscription&patternType=keyword&sm=0

PR #60279, opened 2024-02-07
Jac/otel

jac:
Upgrades the otel version to fix a version mismatch caused by v1.22.0 being used from the root go.mod (upgraded in this [commit](https://github.com/sourcegraph/sourcegraph/commit/0a038b15932a73057616efc604ca7bec0675cafd)) which conflicts with the v1.21.0 gcp detector in the MSP runtime.

As long as the major version is the same Go will use the highest minor version of a package if used by multiple dependencies. As such because the root go.mod had v1.22 it was being used for the MSP runtime lib instead of v1.21 but the gcp detector was still on v1.21 causing a schema mismatch
## Test plan

- CI
- Built and uploaded a docker image to Cloudrun and verified both the gcp detector and default detector use schema url 1.24


unknwon (approved):

unknwon:
Nice find!

PR #60286, opened 2024-02-07
Search Result: Fix search result padding problem

vovakulikov:
## Test plan
- Check the search result matches UI got no regressions in the layout 
- Check that the search aggregation UI has proper paddings 
- Check that zero state UI has proper paddings

### DIFF
--- a/client/web/src/search/results/components/new-search-content/NewSearchContent.module.scss
+++ b/client/web/src/search/results/components/new-search-content/NewSearchContent.module.scss
@@ -126,6 +126,7 @@
 .content {
     grid-area: contents;
     overflow: auto;
+    padding: 1rem;
 
     &::-webkit-scrollbar {
         width: 0.5rem;
@@ -158,6 +159,12 @@
             scrollbar-width: thin;
             scrollbar-color: var(--oc-gray-6);
         }
+    }
+
+    &--list {
+        // Compensate content parent block paddings in order
+        // to have paddings anywhere but content list with search matches
+        margin: -1rem;
     }
 }
 
--- a/client/web/src/search/results/components/new-search-content/NewSearchContent.module.scss
+++ b/client/web/src/search/results/components/new-search-content/NewSearchContent.module.scss
@@ -165,6 +165,13 @@
         // Compensate content parent block paddings in order
         // to have paddings anywhere but content list with search matches
         margin: -1rem;
+
+        // This is needed because search list may have search result matches
+        // block which have sticky header elements, sticky top position doesn't
+        // take into account negative margins so we
+        [data-result-header] {
+            top: -1rem;
+        }
     }
 }
 
--- a/client/branded/src/search-ui/components/ResultContainer.module.scss
+++ b/client/branded/src/search-ui/components/ResultContainer.module.scss
@@ -12,7 +12,9 @@
     flex-wrap: wrap;
 
     position: sticky;
-    top: 0;
+    // Compensate content parent block paddings in order
+    // to have paddings anywhere but content list with search matches
+    top: -1rem;
 
     z-index: 1; // Show on top of search result contents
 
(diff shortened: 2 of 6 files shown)


client/web/src/search/results/components/new-search-content/NewSearchContent.module.scss
```diff
@@ -159,6 +160,12 @@
             scrollbar-color: var(--oc-gray-6);
         }
     }
+
+    &--list {
+        // Compensate content parent block paddings in order
+        // to have paddings anywhere but content list with search matches
+        margin: -1rem;
+    }
```
camdencheek, line 168: I don't understand what's going on here. Why are we padding the content but unpadding the list? Why don't we just wrap the zero-result state with padding? Also, any idea why this seems to be nondeterministic in the builds?

vovakulikov: @camdencheek adding padding to all elements within the content parent block is problematic because you have to keep in mind to add them. The layout usually should take into account global layout rules. In this case only block that doesn't fit in padding is the search matches list, this is why I added it here. 

>Also, any idea why this seems to be nondeterministic in the builds?

To be honest not sure what happened here, I bet that there was a missing fix in prev rebase conflicts, haven't seen in the search result layout that looks suspicious in this way.

camdencheek: > adding padding to all elements within the content parent block is problematic because you have to keep in mind to add them

But isn't the purpose of having padding in the parent block specifically because it applies to _all_ children? In this case, it _shouldn't_ apply to all children because we don't want it to apply to the search results list. This feels brittle because now the negative margin of the list always has to exactly match the positive margin of the container

vovakulikov: @camdencheek true, but since these margin/padding values are set in the same place (in the same CSS file) I think it's better rather than setting padding for all children manually. They also would have had the same padding values. 

From my side having padding for all and making exceptions with negative margins for one block makes sense rather than having multiple paddings on children's level.

vovakulikov: Having an inner wrapper with paddings also won't work properly because we have different types of content in the search result area on different React component levels. Again multiple padding set isn't better than one fix for search matches list.

camdencheek:
Are you able to reproduce the padding issues locally? Right now, I'm not seeing any issues on s2.

client/web/src/search/results/components/new-search-content/NewSearchContent.module.scss
camdencheek: Then shouldn't we just have a single DOM container that includes the padding for "everything else"? 

In the short time I've been working on the webapp, negative margins have bitten me more than once. In fact, I just ran into another spot where a compensating negative margin was causing really weird behavior because the margin it was compensating for changed

vovakulikov: Also, we can't add paddings directly inside `StreamingSearchResultsList` because it's used in different places and may conflict with styles in other consumers so again all spacing rules should live on the consumer level.

camdencheek:
Okay. I think I can at least explain away the "nondeterminism." I think maybe it was just that, on some widths, the no results page looks fine because there is a max width to the element.

https://github.com/sourcegraph/sourcegraph/assets/12631702/6184a323-2211-4bbf-b4ba-f193f9b347c9

vovakulikov:
Yeah, this happens to me all the time since I use small/mid-size windows, the screenshot from s2 in the latest build. 

Screenshot 2024-02-07 at 22 10 15

camdencheek (approved):

vovakulikov:
I'll merge this but I will keep in mind if anyone gets any confusion about the search result layout I will revert this and set paddings rules manually.

PR #60288, opened 2024-02-07
Search filters: final design tweaks

camdencheek:
This is my last planned set of design fixes for the filters panel. It's somewhat of a monolithic PR since it's the result of pairing with @taiyab and we just fixed things while we were looking at it together.

In summary:
- Fixes the issue where we would show `0` count even though a limit was hit
- Hides the entire "By type" section when a `type:` filter is entered explicitly by the user
- Fixes various spacing issues that became more apparent when the "By type" section is gone
- Fixes various small consistency issues where the "By type" section is rendered differently than the other sections
- Fixes some overlap issues that were caused by negative margins

Depends on and includes https://github.com/sourcegraph/sourcegraph/pull/60216
Stacked on https://github.com/sourcegraph/sourcegraph/pull/60239

Video demo:

https://github.com/sourcegraph/sourcegraph/assets/12631702/35c2357e-263c-4462-8eeb-c7959cafc9ca

## Test plan

Visual testing with Taiyab

### DIFF
--- a/client/branded/src/search-ui/results/filters/NewSearchFilters.module.scss
+++ b/client/branded/src/search-ui/results/filters/NewSearchFilters.module.scss
@@ -11,32 +11,8 @@
         height: 0.5rem;
     }
 
-    &::-webkit-scrollbar-thumb {
-        background-color: var(--oc-gray-4);
-    }
-
-    &::-webkit-scrollbar-track {
-        background-color: var(--oc-gray-1);
-    }
-
     @-moz-document url-prefix('') {
         scrollbar-width: thin;
-        scrollbar-color: var(--oc-gray-1);
-    }
-
-    :global(.theme-dark) & {
-        &::-webkit-scrollbar-thumb {
-            background-color: var(--oc-gray-8);
-        }
-
-        &::-webkit-scrollbar-track {
-            background-color: var(--oc-gray-7);
-        }
-
-        @-moz-document url-prefix('') {
-            scrollbar-width: thin;
-            scrollbar-color: var(--oc-gray-6);
-        }
     }
 }
 
--- a/client/branded/src/search-ui/results/filters/NewSearchFilters.module.scss
+++ b/client/branded/src/search-ui/results/filters/NewSearchFilters.module.scss
@@ -19,14 +19,12 @@
 .filters {
     display: flex;
     flex-direction: column;
-    margin-bottom: 1rem;
     gap: 0.5rem;
 }
 
 .actions {
     position: sticky;
     bottom: 0;
-    margin-top: -0.5rem;
     padding: 1rem 0.75rem;
     border-top: 1px solid var(--border-color);
     background-color: var(--code-bg);
--- a/client/branded/src/search-ui/results/filters/NewSearchFilters.module.scss
+++ b/client/branded/src/search-ui/results/filters/NewSearchFilters.module.scss
@@ -20,6 +20,7 @@
     display: flex;
     flex-direction: column;
     gap: 0.5rem;
+    margin-bottom: auto;
 }
 
 .actions {
--- a/client/branded/src/search-ui/results/filters/NewSearchFilters.module.scss
+++ b/client/branded/src/search-ui/results/filters/NewSearchFilters.module.scss
@@ -4,7 +4,6 @@
     overflow: auto;
     display: flex;
     flex-direction: column;
-    gap: 0.5rem;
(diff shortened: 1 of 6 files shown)


client/branded/src/search-ui/results/filters/NewSearchFilters.module.scss
```diff
@@ -4,52 +4,28 @@
...
-        }
-
-        &::-webkit-scrollbar-track {
-            background-color: var(--oc-gray-7);
-        }
-
-        @-moz-document url-prefix('') {
-            scrollbar-width: thin;
-            scrollbar-color: var(--oc-gray-6);
-        }
```
camdencheek, line 40: These colors weren't being applied, so we got rid of them.

client/branded/src/search-ui/results/filters/NewSearchFilters.module.scss
```diff
@@ -4,52 +4,29 @@
...
+    display: flex;
+    flex-direction: column;
+    gap: 0.5rem;
     margin-bottom: auto;
 }
 
 .actions {
     position: sticky;
     bottom: 0;
-    margin-top: -0.5rem;
```
camdencheek, line 51: This negative margin was causing problems when we updated the spacing on the filters

client/branded/src/search-ui/results/filters/NewSearchFilters.module.scss
vovakulikov: @camdencheek I don't think this is something that we should remove, these rules set a custom style for the filters panel scrollbar. The reason we have it is because we want to show the scrollbar unconditionally (when we have actual scroll) because on MacOS this is a common case when you can't see the scrollbar unless you're performing an actual scroll action. Without a scrollbar it is easy to miss that there is additional content beyond the visible viewport area.

client/branded/src/search-ui/results/filters/components/dynamic-filter/SearchDynamicFilter.module.scss
```diff
@@ -1,5 +1,5 @@
 .root {
-    padding: 0.5rem 0.5rem 0.5rem 0.75rem;
+    padding: 0.75rem 1rem;
```
vovakulikov, line 2: I know that this comes from the original design, but not sure if is it because in Figma all looks a bit downscaled but on my screen (mid-size browser window) this spacing feels off. 

cc @taiyab

client/branded/src/search-ui/results/filters/components/filter-type-list/FilterTypeList.tsx
```diff
@@ -110,32 +110,28 @@ interface FilterTypeButtonProps {
...
-            </Button>
-        </Tooltip>
+        <Button
+            variant={selected ? 'primary' : 'secondary'}
+            outline={!selected}
+            className={classNames(styles.typeListItem, {
+                [styles.typeListItemSelected]: selected,
+            })}
+            onClick={() => onClick(filter, selected)}
+            ref={null}
```
vovakulikov, line 127: I don't belive we need this null ref here

vovakulikov (commented):
Left some  minor comments about spacing and as I think important one about scroll bar

client/branded/src/search-ui/results/filters/components/dynamic-filter/SearchDynamicFilter.module.scss
taiyab: Tested it locally and it seems fine.

taiyab (commented):
Couple of comments. LGTM! 👍

client/branded/src/search-ui/results/filters/NewSearchFilters.module.scss
camdencheek: Okay, I ended up reverting the color changes to the scrollbar because we figured out why the color stuff wasn't working on firefox. We're currently using the default browser behavior for scrollbars and just overriding the colors to be less stark. I apparently have a firefox setting enabled that shows the scrollbar only on scroll, and I don't want to go messing with behavior like that right before code freeze. 

After talking with Vova, it turns out that scrollbar colors and behavior are all over the place in the webapp, so this would be a great thing to follow up on and do a more comprehensive look at "how do we handle scrolling in the webapp."

taiyab (commented):
Fine with the compromises. Looks good.

vovakulikov (approved):

Issue #56915, opened 2023-09-21
Improve error messages displayed to admin when a repository fails


mike-r-mclaughlin:
**Reported by:** https://github.com/sourcegraph/accounts/issues/6938

#### Feature request description

Often exception messages like `Context deadline exceeded` are displayed to the user.   It would be great if the error message indicated what caused this and what to do to resolve it.  

![image](https://github.com/sourcegraph/sourcegraph/assets/23640224/1226e773-64fc-4b81-aa39-f99810977fb3)

#### Is your feature request related to a problem? If so, please describe.

When troubleshooting errors with repository clones, it's difficult for admins to know what to do with a generic error message like `context deadline exceeded`.  Often the log is not accessible or has no additional information.

/cc @sourcegraph/source

PR #60294, opened 2024-02-07
gitserver: Default to unimplemented handler of P4Exec

eseliger:
Just a simple change, we don't need to explicitly reimplement this.

## Test plan

This endpoint is unused, so as long as Go is happy, it will be fine.

PR #60292, opened 2024-02-07
gitserver: Make patch reader for CreateCommitFromPatch streaming

eseliger:
This makes it so that we don't have to buffer the entire patch in memory, which has been seen to be 100MB+ in the past.

## Test plan

Existing test suites cover that this didn't break anything.

PR #60291, opened 2024-02-07
gitserver: Make helper function for maybeStartClone

eseliger:
Now that we know how often this is going to be copied, I decided to make a helper function.

## Test plan

Existing tests verify that the behavior hasn't been changed (except for the error string)

PR #60298, opened 2024-02-08
gitserver: Improve error reporting for context deadlines for clones and updates

eseliger:
This PR started as a "oh let me wrap the context error in a nice message".

Then, as I dug in, I noticed that we did very messy things with context along the way, ultimately causing clones to be canceled while a request loses interest, making us have to start over, and those nicely wrapped errors just not bubbling up. Instead, we would get cryptic "signal: killed" errors now.

I want to come back at this again and check if we could rewrite these errors in the presentation layer only, but for now I think this is a fix that's workable. We should be coming back at this when we look at scheduler.

I'll include inline comments to explain the changes in detail.

Closes https://github.com/sourcegraph/sourcegraph/issues/56915

## Test plan

Test suites still pass, the integration test suite will cover it E2e, and manually verified using a 1 second long command timeout that these errors render nicely now.

### DIFF
--- a/cmd/gitserver/internal/ensurerevision.go
+++ b/cmd/gitserver/internal/ensurerevision.go
@@ -23,6 +23,7 @@ func (s *Server) EnsureRevision(ctx context.Context, repo api.RepoName, rev stri
 		ensureRevisionCounter.WithLabelValues("HEAD").Inc()
 		return false
 	}
+
 	if conf.Get().DisableAutoGitUpdates {
 		ensureRevisionCounter.WithLabelValues("disabled").Inc()
 		// ensureRevision may kick off a git fetch operation which we don't want if we've
@@ -33,7 +34,9 @@ func (s *Server) EnsureRevision(ctx context.Context, repo api.RepoName, rev stri
 	// Revision not found, update before returning.
 	err := s.doRepoUpdate(ctx, repo, rev)
 	if err != nil {
-		ensureRevisionCounter.WithLabelValues("update_failed").Inc()
+		if ctx.Err() == nil {
+			ensureRevisionCounter.WithLabelValues("update_failed").Inc()
+		}
 		s.logger.Warn("failed to perform background repo update", log.Error(err), log.String("repo", string(repo)), log.String("rev", rev))
 		// TODO: Shouldn't we return false here?
 	} else {
--- a/cmd/gitserver/internal/server.go
+++ b/cmd/gitserver/internal/server.go
@@ -162,7 +162,7 @@ type ServerOpts struct {
 }
 
 func NewServer(opt *ServerOpts) *Server {
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithCancelCause(context.Background())
 
 	// GitMaxConcurrentClones controls the maximum number of clones that
 	// can happen at once on a single gitserver.
@@ -247,8 +247,8 @@ type Server struct {
 	// server is stopped. Do not directly call this, rather call
 	// Server.context()
 	ctx      context.Context
-	cancel   context.CancelFunc // used to shutdown background jobs
-	cancelMu sync.Mutex         // protects canceled
+	cancel   context.CancelCauseFunc // used to shutdown background jobs
+	cancelMu sync.Mutex              // protects canceled
 	canceled bool
 	wg       sync.WaitGroup // tracks running background jobs
 
@@ -321,11 +321,12 @@ type clonePipelineRoutine struct {
 	// TODO: Get rid of this dependency.
 	s      *Server
 	queue  *common.Queue[*cloneJob]
-	cancel context.CancelFunc
+	cancel context.CancelCauseFunc
 }
 
 func (p *clonePipelineRoutine) Start() {
-	ctx, cancel := context.WithCancel(context.Background())
+	// TODO: This should probably use serverContext.
+	ctx, cancel := context.WithCancelCause(context.Background())
 	p.cancel = cancel
 	// Start a go routine for each the producer and the consumer.
 	go p.cloneJobConsumer(ctx, p.tasks)
@@ -334,7 +335,7 @@ func (p *clonePipelineRoutine) Start() {
 
 func (p *clonePipelineRoutine) Stop() {
 	if p.cancel != nil {
-		p.cancel()
+		p.cancel(errors.New("clone pipeline routine stopped"))
 	}
 }
 
@@ -416,7 +417,9 @@ var repoCloned = func(dir common.GitDir) bool {
 // Stop cancels the running background jobs and returns when done.
 func (s *Server) Stop() {
 	// idempotent so we can just always set and cancel
-	s.cancel()
+	// Provide a little bit of context of where this context cancellation
+	// is coming from.
+	s.cancel(errors.New("gitserver is shutting down"))
 	s.cancelMu.Lock()
 	s.canceled = true
 	s.cancelMu.Unlock()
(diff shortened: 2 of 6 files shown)


cmd/gitserver/internal/ensurerevision.go
```diff
@@ -52,10 +53,13 @@ func (s *Server) ensureRevision(ctx context.Context, repo api.RepoName, rev stri
...
 		return false
 	}
+
 	// Revision not found, update before returning.
 	err := s.doRepoUpdate(ctx, repo, rev)
 	if err != nil {
-		ensureRevisionCounter.WithLabelValues("update_failed").Inc()
+		if ctx.Err() == nil {
+			ensureRevisionCounter.WithLabelValues("update_failed").Inc()
+		}
```
eseliger, line 62: Make sure we don't miscount here, those aren't actually failed updates, just requests that lost interest in the update result.

cmd/gitserver/internal/server.go
```diff
@@ -377,7 +377,9 @@ var repoCloned = func(dir common.GitDir) bool {
 // Stop cancels the running background jobs and returns when done.
 func (s *Server) Stop() {
 	// idempotent so we can just always set and cancel
-	s.cancel()
+	// Provide a little bit of context of where this context cancellation
+	// is coming from.
+	s.cancel(errors.New("gitserver is shutting down"))
```
eseliger, line 382: gives a little bit more context as to why the context was canceled.

cmd/gitserver/internal/server.go
```diff
@@ -286,7 +286,7 @@ type clonePipelineRoutine struct {
 }
 
 func (p *clonePipelineRoutine) Start() {
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := p.s.serverContext()
```
eseliger, line 289: using serverContext here vs a background context makes it so that gitserver properly waits for the pipeline to shut down current clones. 

TODO: Make sure that gitserver can still shut down properly, as gitserver.Stop is only called AFTER the MonitorBackgroundRoutines returns.

cmd/gitserver/internal/server.go
```diff
@@ -456,19 +458,18 @@ func (s *Server) IsRepoCloneable(ctx context.Context, repo api.RepoName) (protoc
...
+
 	req.Repo = protocol.NormalizeRepo(req.Repo)
 	dir := gitserverfs.RepoDirFromName(s.ReposDir, req.Repo)
 
-	// despite the existence of a context on the request, we don't want to
-	// cancel the git commands partway through if the request terminates.
-	ctx, cancel1 := s.serverContext()
-	defer cancel1()
-	ctx, cancel2 := context.WithTimeout(ctx, conf.GitLongCommandTimeout())
-	defer cancel2()
```
eseliger, line 470: Using the serverContext here already makes it so that the request has no control about whether it can return early from here, even though the work is running in the background anyways. This fixes that, and fixes that we double-register a timeout and serverContext in the case that !repoCloned

cmd/gitserver/internal/server.go
```diff
@@ -744,32 +748,40 @@ func (s *Server) CloneRepo(ctx context.Context, repo api.RepoName, opts CloneOpt
...
-	syncer, err := s.GetVCSSyncer(ctx, repo)
-	if err != nil {
-		return "", errors.Wrap(err, "get VCS syncer")
-	}
+	syncer, remoteURL, err := func() (_ vcssyncer.VCSSyncer, _ *vcs.URL, err error) {
+		defer func() {
+			if err != nil {
+				s.setLastErrorNonFatal(serverCtx, repo, err)
+			}
+		}()
```
eseliger, line 760: wrapped to not repeat setLastErrorNonFatal multiple times in all error branches

cmd/gitserver/internal/server.go
```diff
@@ -788,16 +800,31 @@ func (s *Server) CloneRepo(ctx context.Context, repo api.RepoName, opts CloneOpt
...
+			s.setLastErrorNonFatal(serverCtx, repo, err)
+		}()
+
+		select {
+		case <-done:
+			return "", err
+		case <-ctx.Done():
+			return "", ctx.Err()
+		}
 	}
```
eseliger, line 828: we wrap in a goroutine here, so that beyond context cancellation from the caller the clone continues. There's never a case where only the specific caller is interested in a repo. Gitserver will always have to clone a repo anyways, so we should not abort after making 95% progress here. Instead, we allow the calling context to early-exit from this function.

cmd/gitserver/internal/server.go
```diff
@@ -898,6 +927,9 @@ func (s *Server) doClone(
...
 
 	if cloneErr != nil {
+		if errors.Is(cloneCtx.Err(), context.DeadlineExceeded) {
+			return errors.Newf("failed to clone repo within deadline of %s", cloneTimeout)
+		}
```
eseliger, line 932: Magic new error, to add clarity of what happened here to the user.

cmd/gitserver/internal/server.go
```diff
@@ -1246,64 +1283,76 @@ func (s *Server) doBackgroundRepoUpdate(repo api.RepoName, revspec string) error
...
-	if err != nil {
-		return err
-	}
-	defer cancel2()
-
-	if err = s.RPSLimiter.Wait(ctx); err != nil {
-		return err
-	}
+	err := func(ctx context.Context) error {
+		ctx, cancelLimiter, err := s.acquireCloneLimiter(ctx)
```
eseliger, line 1298: strongly recommend to hide whitespace changes here, to make reviewing easier

cmd/gitserver/internal/server.go
```diff
@@ -1246,64 +1283,76 @@ func (s *Server) doBackgroundRepoUpdate(repo api.RepoName, revspec string) error
...
+				return errors.Wrapf(err, "failed to fetch repo %q", repo)
+			}
 		}
+
+		return postRepoFetchActions(ctx, logger, s.DB, s.Hostname, s.RecordingCommandFactory, repo, dir, remoteURL, syncer)
+	}(ctx)
+
+	if errors.Is(err, context.DeadlineExceeded) {
+		return errors.Newf("failed to update repo within deadline of %s", fetchTimeout)
 	}
```
eseliger, line 1353: wrapped in a function to make it easy to always wrap the error with this pretty message.

eseliger:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/60298?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `main`
  * **PR #60291** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/60291?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #60292** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/60292?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #60294** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/60294?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #60295** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/60295?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
          * **PR #60298** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/60298?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/60298?utm_source=stack-comment).

cmd/gitserver/internal/server.go
```diff
@@ -286,6 +286,7 @@ type clonePipelineRoutine struct {
 }
 
 func (p *clonePipelineRoutine) Start() {
+	// TODO: This should probably use serverContext.
```
eseliger, line 289: But for that I need to ensure that gitserver can still shut down properly, since the clone pipeline needs to stop before the context is cancelled currently.
So that's for later.

ggilmore (approved):

ggilmore:
Will be nice to see if we can see the codes.Aborted error rates in our charts go down as a result of this.

PR #60303, opened 2024-02-08
svelte: Use parallel data loading

fkling:
The way how data loaders are currently depending on each other results in unnecessary waterfall data loading. I originally thought that it would make the data flow clearer, but I think the benefit of parallel data loading outweighs this.

Before:
- all data loaders depend on the top level data loader to provide the GraphQL client, but the top level data loader also fetches initial data for the app which can take time
- any repo data loaders depend on the repo root data loader to validate/resolve the repo and commit

Now:
- data loaders fetch a reference to the global graphql client
- almost all repo loaders use the repo name and revision from the URL
  - this required to change GraphQL queries from `node(id)` to `repository(name)`
  - before the loader could assume that the repo and revision are valid. That's not the case anymore, so each of these queries could "fail". However the `(validrev)` data loader will still validate the revision and render an appropriate error page. So even though the data loaders run and try to fetch the data for an invalid repo/revision, the corresponding page won't be rendered.

If a loader needs access to user data they will of course need to wait for the parent loader to resolve. But so far that hasn't been the case.

## Test plan

Manual testing, `pnpm test`

### DIFF
--- a/client/web-sveltekit/src/routes/[...repo=reporev]/+layout.ts
+++ b/client/web-sveltekit/src/routes/[...repo=reporev]/+layout.ts
@@ -14,7 +14,7 @@ import {
 } from '$lib/shared'
 
 import type { LayoutLoad } from './$types'
-import { ResolveRepoRevison, ResolvedRepository } from './layout.gql'
+import { ResolveRepoRevision, ResolvedRepository, type ResolveRepoRevisionResult } from './layout.gql'
 
 export interface ResolvedRevision {
     repo: ResolvedRepository
@@ -83,21 +83,18 @@ async function resolveRepoRevision({
     repoName: string
     revision?: string
 }): Promise<ResolvedRevision> {
+    // See if we have a cached response
     let data = client.readQuery({
-        query: ResolveRepoRevison,
+        query: ResolveRepoRevision,
         variables: {
             repoName,
             revision,
         },
     })
-    if (
-        !data ||
-        (data.repositoryRedirect?.__typename === 'Repository' && data.repositoryRedirect.commit?.oid !== revision)
-    ) {
-        // We always refetch data when 'revision' is a "symbolic" revision (e.g. a tag or branch name)
+    if (shouldResolveRepositoryInformation(data)) {
         data = await client
             .query({
-                query: ResolveRepoRevison,
+                query: ResolveRepoRevision,
                 variables: {
                     repoName,
                     revision,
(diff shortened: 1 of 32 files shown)


client/web-sveltekit/src/routes/[...repo=reporev]/+layout.ts
```diff
@@ -65,13 +65,13 @@ export const load: LayoutLoad = async ({ parent, params, url, depends }) => {
     }
 }
 
-function redirectToExternalHost(externalRedirectURL: string, currentURL: URL): Redirect {
+function redirectToExternalHost(externalRedirectURL: string, currentURL: URL): never {
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/+layout.ts:68]: Huh, did not expect redirects to be implemented as exceptions

camdencheek (approved):
Some nice simplification to the GraphQL queries too!

client/web-sveltekit/src/routes/[...repo=reporev]/+layout.ts
Comment (fkling) [reply client/web-sveltekit/src/routes/[...repo=reporev]/+layout.ts:68]: I missed that when updating to v2.0. From the docs:

> Like `error(...)`, calling `redirect(...)` will throw an exception, making it easy to stop execution from inside helper functions.

fkling:
I realized why I implemented it the way it was in the first place (should have left a comment): It works better with caching and ensures that all data is consistent when browsing different repository pages. With the changes in this PR you could run into the following scenario:

- User navigate to a file X at revision A
- In the meantime a new commit becomes available (B) that changes file X and Y
- User navigates to file Y from the sidebar, sees content from commit B
- User navigates back to file X (either back button or file sidebar) and would see file X at revision A again

Before writing this I thought of a couple of things we could do but there are only two that are really feasible:

- Never cache those queries -> delay when navigating to pages you've already visited
- Wait for revision to resolve and use specific commit ID for all queries -> waterfall for initial page load but consistent data throughout (the way it is right now without this PR).

Of those two I'd prefer the second one. There are two improvements we can still make wrt to avoiding waterfalls:

- Don't block on the initial GraphQL request which fetches user information (already implemented in this PR)
- Do not "re-resolve" symbolic revisions when navigating to repo that was already visited. That would mean a symbolic revision will always resolve to the commit that the symbolic name pointed at the time the web app was loaded (not yet done in this PR). I think that would be even better wrt back and forward navigation in the browser, because it would mean that you still see the same content when navigating back to a repo page after leaving the repo. (and we can still implement a "side check" to notify the user when a new revision is available).


What are your thoughts about this @camdencheek ?

fkling:
I've made the following changes in a follow up commit:

- Where necessary, the data loader will wait for the parent loader to resolve the revision to ensure consistent data while browsing repo pages. Some pages don't use the current revision. Maybe we'll have to follow up on that and change them to use the revision.
- If possible, the revision from the URL is used. If the revision in the URL is a full SHA1, then it's used as is and the loader doesn't have to wait for the parent. This is a slight optimization and it's reasonable to assume that shared URLs often contain the commit ID.
- We are not re-resolving a symbolic revspec every time when navigating to a repository page. That means the revspec is resolved only the first time (during a single session). This slightly speeds up page navigation but more importantly ensures that the data is consistent when navigating the pages. Eventually we should add a way to information the user when new commits are available.

PR #60307, opened 2024-02-08
svelte: Improve scroll behavior on repo pages

fkling:
This fixes an issue that had been bugging for quite a while: When you scrolled within a file and navigated to a new file, we would open the new file at the same scroll position instead of at the top of the file.

SvelteKit has native support for restoring and resetting the scroll position, but that only works when the whole page is scrollable. In our case we made an "app" wrapper `<div>` scrollable, so that wouldn't work with SvelteKit.
I made the necessary HTML changes so that we can leverage SvelteKit's default behavior but that was problematic too: After clicking on an entry in the file tree the whole file tree would be moved down because now the whole page would be scrolled to the top, revealing the page navigation bar and repo header again.

So what we really want is to scroll to the top of what I consider the *content* of the repo page, excluding page navigation and repo header. That's exactly what the changes in the repo layout file do: After navigation we check the scroll position of the page (easy to do now that truly the whole page is scrollable), and scroll to the beginning of the page content. We need to take some additional steps to restore the correct scroll position when navigating back and forth via the built-in snapshot feature.

https://github.com/sourcegraph/sourcegraph/assets/179026/bd882831-2ee4-449f-89ef-8dc3e1a7cc84

## Test plan

Manual testing

### DIFF
--- a/client/web-sveltekit/src/lib/dom.ts
+++ b/client/web-sveltekit/src/lib/dom.ts
@@ -168,3 +168,25 @@ export const computeFit: Action<HTMLElement, void, ComputeFitAttributes> = node 
         },
     }
 }
+
+/**
+ * Helper action to manage CSS classes on an element. This is used on svelte:body because
+ * it doesn't support the class directive.
+ * See https://github.com/sveltejs/svelte/issues/3105
+ */
+export const classNames: Action<HTMLElement, string | string[]> = (node, classes) => {
+    classes = (Array.isArray(classes) ? classes : [classes]).filter(Boolean)
+    node.classList.add(...classes)
+
+    return {
+        update(newClasses) {
+            newClasses = (Array.isArray(newClasses) ? newClasses : [newClasses]).filter(Boolean)
+            node.classList.remove(...classes)
+            node.classList.add(...newClasses)
+            classes = newClasses
+        },
+        destroy() {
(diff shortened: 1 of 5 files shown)


client/web-sveltekit/src/lib/dom.ts
```diff
@@ -168,3 +168,25 @@ export const computeFit: Action<HTMLElement, void, ComputeFitAttributes> = node
...
     }
 }
+
+/**
+ * Helper action to manage CSS classes on an element. This is used on svelte:body because
+ * it doesn't support the class directive.
+ * See https://github.com/sveltejs/svelte/issues/3105
+ */
+export const classNames: Action<HTMLElement, string | string[]> = (node, classes) => {
+    classes = (Array.isArray(classes) ? classes : [classes]).filter(Boolean)
```
camdencheek, line 178: `.filter(Boolean)`

Is this saying "only retain truthy classes"? So, things like an empty string or undefined would be filtered out?

camdencheek (approved):

client/web-sveltekit/src/lib/dom.ts
fkling: Yes. The DOM API throws an error when an empty string is passed (which I do in the callsite for simplicity). I'll add a comment.

PR #60316, opened 2024-02-08
search: fix hoisting heuristic for nested queries

stefanhengl:
With the recent changes in the client, the client now sends us queries wrapped in parentheses which our hoisting heuristics don't handle well.

There are 2 issues.

1. The leaf parser incorrectly marks some queries as "disambiguated"
2. `Hoist` doesn't deal well with common nested queries such as `(and parameter (or (and parameter pattern ) pattern))`

This PR addresses both issues like follows:

1. We don't mark queries as disambiguated if we encounter a boolean operator inside parenthesis
2. We add a recursion to `Hoist` that deals with the simplest case of `(and parameter expression)`

Note:
This is also fixes a bug we recently documented with a test case, see `job_test.go`

The question is, are we hoisting anything now that we shouldn't but we haven't documented it in tests?

Test plan:
- added new tests to cover the changes

### DIFF
--- a/internal/search/query/parser_test.go
+++ b/internal/search/query/parser_test.go
@@ -343,15 +343,45 @@ func TestParse(t *testing.T) {
 		}
 	}
 
-	// Queries with repo
+	// Queries with top-level parentheses, to mirror what we receive from client
+	autogold.Expect(value{
+		Grammar:   `(or (and "repo:foo" "count:2000" "internal") "src")`,
+		Heuristic: `(and "repo:foo" "count:2000" (or "internal" "src"))`,
+	}).Equal(t, test("(repo:foo count:2000 internal or src)"))
+	autogold.Expect(value{
+		Grammar:   `(and "lang:C++" "type:path" (or (and "repo:foo" "count:2000" "internal") "src"))`,
+		Heuristic: `(and "lang:C++" "type:path" "repo:foo" "count:2000" (or "internal" "src"))`,
+	}).Equal(t, test("(repo:foo count:2000 internal or src) lang:C++ type:path"))
+	autogold.Expect(value{
+		Grammar:   `(or (and "repo:foo" "count:2000" "internal" "limit") "src")`,
+		Heuristic: "Same",
+	}).Equal(t, test("((repo:foo count:2000 internal and limit) or src)"))
+	autogold.Expect(value{
+		Grammar:   `(and "lang:C++" "type:path" (or (and "repo:foo" "count:2000" "internal" "limit") "src"))`,
+		Heuristic: "Same",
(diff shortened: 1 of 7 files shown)


internal/search/query/parser_test.go
```diff
@@ -343,6 +343,68 @@ func TestParse(t *testing.T) {
...
+	}).Equal(t, test("repo:foo count:2000 internal or src"))
+	autogold.Expect(value{
+		Grammar:   `(or (and "repo:foo" "count:2000" "internal") "src")`,
+		Heuristic: "Same",
+	}).Equal(t, test("(repo:foo count:2000 internal) or src"))
+	autogold.Expect(value{
+		Grammar:   `(or (and "repo:foo" "count:2000" (concat "internal" "limit")) "src")`,
+		Heuristic: "Same",
+	}).Equal(t, test("(repo:foo count:2000 internal limit) or src"))
+	autogold.Expect(value{
```
jtibshirani, line 377: Here's an example of a query that was not working using the old heuristic:

```
(repo:foo count:2000 internal and limit) or src
```

We decided that this was ambiguous, because it contains an "and" operator within parentheses. With the new heuristic, we (correctly) leave this alone.

jtibshirani:
> 1. We don't mark queries as disambiguated if we encounter a boolean operator inside parenthesis

This heuristic was a bit too broad, and led to us hoisting too aggressively in some cases.

I updated the heuristic to instead be: a query is ambiguous if and only if there's an `or` expression whose children have parentheses. This covers cases like this, which we don't want to hoist:

```(repo:foo count:2000 internal and limit) or src```

I also tested end-to-end with the new filters panel, and things seem to work correctly.

camdencheek (approved):
Did some local testing, and everything worked as expected. I didn't really give the code a good review because I don't think I can add anything useful here without some deep review of the parsing code. 

One example I still had a question about:


The query `repo:^github\.com/sourcegraph/code-intel-extensions$ (LANG or ID) or camden` is interpreted as `(repo:^github\.com/sourcegraph/code-intel-extensions$ (LANG or ID)) or (camden)`. I _think_ the ideal rule here would be "hoist any nodes that contain no parameters, only patterns and operators". I do not think this is worth blocking this PR on given the looming code freeze, but might be a good followup if you revisit the parser.

jtibshirani:
> repo:^github\.com/sourcegraph/code-intel-extensions$ (LANG or ID) or camden

Indeed this would be nice to fix. However, we had the same unfortunate semantics in 5.2 -- since there are parentheses, we consider the whole query to be "disambiguated" and don't attempt hoisting. So as you mentioned, we'll file it away for later and keep this PR simple.

PR #60319, opened 2024-02-08
database: remove telemetry-export feature flag

keegancsmith:
According to the comment the feature flag should be removed by now. I double checked in slack and we can.

Test Plan: CI


eseliger (approved):

bobheadxi:
Thank you!

PR #60323, opened 2024-02-08
gitserver: Serve Refs via gRPC method

eseliger:
This PR migrates the ref specific calls to a gRPC endpoint.

While looking at this, I realized that we have many methods that want the same thing in the end: refs.

So instead of adding various endpoints (that currently use porcelain commands in the client implementation), we now use the plumbing command `for-each-ref` for all of them.

By that, we can combine the following APIs:

- `ListRefs` -> `ListRefs` (stays the same)
- `ListBranches` -> `ListBranches(headsOnly: true)`
- `ListTags` -> `ListBranches(tagsOnly: true)`
- For endpoints that were interested in both branches and tags, we allow the combination of the two `only` flags
- `BranchesContaining` -> `ListRefs(contains: sha)`
- `RefDescriptions` -> `ListRefs(headsOnly: true, tagsOnly: true)`

Through one simple call to `for-each-ref`, we also get to enjoy consistent ordering throughout. This also fixes an issue and improves perf for the branch picker where ordering would be random once the count hits 1000. 

Closes https://github.com/sourcegraph/sourcegraph/issues/55074

Closes https://github.com/sourcegraph/sourcegraph/issues/60415

## Test plan

Added tests at various layers, for gitcli, for the gRPC server, for the new client.
Also, existing tests still pass after this migration, including integration and E2E tests.
Manually ran my instance and clicked around as well, to verify.


cmd/gitserver/internal/git/iface.go
```diff
@@ -116,3 +126,24 @@ const (
 	// ArchiveFormatTar indicates a tar archive is desired.
 	ArchiveFormatTar ArchiveFormat = "tar"
 )
+
+// ListRefsOpts are additional options passed to ListRefs.
+type ListRefsOpts struct {
+	// If true, only heads are returned. Can be combined with tags_only.
+	HeadsOnly bool
+	// If true, only tags are returned. Can be combined with heads_only.
+	TagsOnly bool
```
pjlast, line 135:
```suggestion
	// If true, only heads are returned. Can be combined with HeadsOnly.
	HeadsOnly bool
	// If true, only tags are returned. Can be combined with TagsOnly.
	TagsOnly bool
```

pjlast (approved):

PR #60330, opened 2024-02-08
packages: Temporarily disable failing for invalid versions on dotcom

eseliger:
Since we don't validate versions, we have about 170k packages on dotcom that constantly fail cloning. This is putting unnecessary load at it, so until we fix this, let's accept that some packages will have a few missing versions, in favor of having the package cloned at all, and a calmer system.

## Test plan

Will monitor dotcom.


pjlast (approved):
🤞

PR #60332, opened 2024-02-08
Add Llama Code 13b to all PLG users and remove unused model conversation

philipp-spiess:
Two things going on here:

1. We add Llama Code 13b to all PLG users. c.f. https://sourcegraph.slack.com/archives/C05AGQYD528/p1707411649398769?thread_ts=1706882079.259169&cid=C05AGQYD528
2. Removed some no longer needed model translation. CG can handle the virtual `fireworks/starcoder-16b` model names for a while now.

## Test plan

- Trivial change, CI will test it

### DIFF
--- a/cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go
+++ b/cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go
@@ -374,6 +374,7 @@ func allowedModels(scope types.CompletionsFeature, isProUser bool) []string {
 			"anthropic/claude-instant-1.2-cyan",
 			"anthropic/claude-instant-1.2",
 			"fireworks/starcoder",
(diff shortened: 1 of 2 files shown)


valerybugakov (approved):

abeatrix (approved):

philipp-spiess:
I wonder if this is still the right place to add this with the addition of SSC/SAMS. @vdavid Do you know by any chance? I am looking for the place where the default list of allowed models for PLG users is define and if it's still on the dotcom side.

vdavid:
@philipp-spiess Yes, this should be enough. We haven’t added a new source for rate limits or allowed models for Cody Gateway yet.
Gateway still goes to dotcom to get this info.
Dotcom optionally (controlled by a feature flag) goes to the new SSC (Self-Serve Cody) service to decide whether the user is Free or Pro, but then dotcom sets the actual values and model lists.

Issue #46956, opened 2023-01-25
Add rate limiting for ADO


varsanojidan:
Now that ADO repository syncing is complete, we should add the ability to configure the rate limit for ADO code hosts.

/cc @ryphil

PR #60333, opened 2024-02-08
ratelimiter: Add configuration options for ADO

eseliger:
We already had a rate limiter in the code, it was just not configurable, so we used Inf always. This was a quick win.

Closes https://github.com/sourcegraph/sourcegraph/issues/46956

## Test plan

Manually set a very low limit locally, ran again, saw requests getting queued. Also the rate limiter state in the UI correctly reflects the configuration.


pjlast (approved):
ez :shipit:

PR #60339, opened 2024-02-08
auth: append random suffix when derive username from email

unknwon:
We are seeing many email-derived usernames to be rejected as suspicious, or high chance of collisions. This would be a common theme when using custom domain emails to sign in to Sourcegraph.com. Thus this PR updates the logic to append random 5-char suffix when the username is derived from email.

image

Not sure if this is something we want to do only in dotcom mode (i.e. hurts Enterprise users, tho SCIM is doing the same thing when turned on), up to the Source team.

## Test plan

CI and tested locally e2e

### DIFF
--- a/cmd/frontend/auth/BUILD.bazel
+++ b/cmd/frontend/auth/BUILD.bazel
@@ -58,6 +58,7 @@ go_test(
         "//cmd/frontend/internal/app/ui",
         "//internal/actor",
         "//internal/auth",
+        "//internal/auth/userpasswd",
         "//internal/conf",
         "//internal/database",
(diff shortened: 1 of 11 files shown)


kalanchan (approved):

chrsmith (approved):
LGTM. It seems reasonable, and presumably we can support users changing their `username`s later, right? (As long as the new one is unique, things should be keyed off of their user ID so the username is just to be human-friendly?)

unknwon:
> things should be keyed off of their user ID so the username is just to be human-friendly?

Yes

PR #60343, opened 2024-02-08
v2 telemetry: add v2 telemetry to src/cody

dadlerj:
Context:

## Test plan

`sg start`
visit page

### DIFF
--- a/client/web/src/cody/useCodyChat.tsx
+++ b/client/web/src/cody/useCodyChat.tsx
@@ -416,7 +416,6 @@ export const useCodyChat = ({
 
     const toggleIncludeInferredRepository = useCallback<CodyClient['toggleIncludeInferredRepository']>(() => {
         const action = scope.includeInferredRepository ? 'disabled' : 'enabled'
-        // TODO CHECK THIS AGAINST SLACK THREAD
         logTranscriptEvent(
             scope.includeInferredRepository
                 ? EventName.CODY_CHAT_SCOPE_INFERRED_REPO_DISABLED
--- a/client/web/src/cody/useCodyChat.tsx
+++ b/client/web/src/cody/useCodyChat.tsx
@@ -204,13 +204,13 @@ export const useCodyChat = ({
             }
             eventLogger.log(v1EventLabel, { transcriptId: transcript.id, ...eventProperties })
 
-            var numericID = new Date(transcript.id).getTime()
+            let numericID = new Date(transcript.id).getTime()
             if (isNaN(numericID)) {
                 numericID = 0
             }
             telemetryRecorder.recordEvent(feature, action, { metadata: { transcriptId: numericID / 1000 } })
         },
-        [transcript]
+        [transcript, telemetryRecorder]
     )
 
     const loadTranscriptFromHistory = useCallback(
--- a/client/web/src/cody/useCodyChat.tsx
+++ b/client/web/src/cody/useCodyChat.tsx
@@ -51,8 +51,8 @@ export interface CodyChatStore
     loadTranscriptFromHistory: (id: string) => Promise<void>
     logTranscriptEvent: (
         v1EventLabel: string,
-        feature: codyTranscriptEventFeatures,
-        action: codyTranscriptEventActions,
+        feature: CodyTranscriptEventFeatures,
+        action: CodyTranscriptEventActions,
         eventProperties?: { [key: string]: any }
     ) => void
     initializeNewChat: () => void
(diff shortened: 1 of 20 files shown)


client/web/src/cody/useCodyChat.tsx
```diff
@@ -164,11 +171,12 @@ export const useCodyChat = ({
...
     /** Event logger for transcript specific events to capture the transcriptId */
     const logTranscriptEvent = useCallback(
-        (eventLabel: string, eventProperties?: { [key: string]: any }) => {
+        (v1EventLabel: string, feature: string, action: string, eventProperties?: { [key: string]: any }) => {
             if (!transcript) {
                 return
             }
-            eventLogger.log(eventLabel, { transcriptId: transcript.id, ...eventProperties })
+            eventLogger.log(v1EventLabel, { transcriptId: transcript.id, ...eventProperties })
+            telemetryRecorder.recordEvent(feature, action, { metadata: { transcriptId: transcript.id } })
```
dadlerj, line 179: @vovakulikov the `feature` and `action` arguments on this line are giving me a predictable type errors:

`Argument of type 'string' is not assignable to parameter of type '"INPUT TYPE ERROR: string type is too broad, should be a known value"'.ts(2345)`

The issue here is that the `TelemetryRecorder.recordEvent` function requires these arguments to be `KnownString`s: https://sourcegraph.com/github.com/sourcegraph/telemetry@65f0b74a5b00f0670420eac4b3367e6ade651cbe/-/blob/src/index.ts?L15-26

I've messed around a bit with ways to...
* Scope down the argument to this `logTranscriptEvent` function. E.g. import the `KnownString` type here from the @sourcegraph/telemetry package? Or create an enum with all of the event features that get passed in here and use that to scope the list of feature types?
* Bring the `telemetryRecorder.recordEvent`s back to the callers? But then they lack knowledge of the transcript. I suppose I could double down on closures and pass a function _into_ logTranscriptEvent that itself accepts the transcript ID as an argument and calls the recordEvent call from there? E.g. the call sites themselves could look like the following:

```
logTranscriptEvent('event name', (transcriptId: number) => telemetryRecorder.recordEvent('feature', 'action', { metadata: { transcriptId } }, { ... })
```

Any suggestions? I feel like no matter how I do this it'll get messy.

vovakulikov: @dadlerj I see, I would go with the simplest way here and create an enum for features events, and actions since this is what the original telemetry recorder API says. They created these restricted types for the reason that people don't use common string keys. So it would be something simular to the event names that cody logic already has https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@dcd306416671b51b7454bb0ca3b7b8c6b11f4844/-/blob/client/web/src/util/constants.ts?L4

Bringing `recordEvent` back to the callers is possible too, but I think adding another function and passing transcript data through closure would be too complex here.

Let me know if enum won't work for you.

dadlerj:
I made a bunch of updates to this one based on what I've learned/done throughout the rest of these PRs. This one has been extensively tested as well, and I'm feeling good about it.

client/web/src/cody/useCodyChat.tsx
```diff
@@ -97,13 +103,36 @@ interface CodyChatProps {
 const CODY_TRANSCRIPT_HISTORY_KEY = 'cody.chat.history'
 const SAVE_MAX_TRANSCRIPT_HISTORY = 20
 
+export type CodyTranscriptEventFeatures =
```
vovakulikov, line 106: Not a blocker at all. My personal opinion about string literal types is that in case you want to change some name of the event you have to go through all consumers of these (the places where we use this event name) and change it manually. At the same time if it was done with enum you could change it just in one place (in the enum declarations)

```ts
enum CodyTranscriptEventFeatures {
   Chat = 'cody.chat'
   ChatItem = 'cody.chat.item'
   ...
   AskCody = 'repo.askCody'
}
```

client/web/src/cody/useCodyChat.tsx
```diff
@@ -164,13 +193,24 @@ export const useCodyChat = ({
 
     /** Event logger for transcript specific events to capture the transcriptId */
     const logTranscriptEvent = useCallback(
-        (eventLabel: string, eventProperties?: { [key: string]: any }) => {
+        (
+            v1EventLabel: string,
+            feature: CodyTranscriptEventFeatures,
+            action: CodyTranscriptEventActions,
```
vovakulikov, line 199: Nice

vovakulikov (approved):

client/web/src/cody/useCodyChat.tsx
dadlerj: Thanks! I think I'm okay with this as-is—having the event names directly in the code (versus an enum in another file) helps ensure naming is relevant to the specific place where it's used (at least, in my mind...).

PR #60344, opened 2024-02-08
Blob view: bandaid selection coloring

camdencheek:
Codemirror applies its own styling for selections, and when the blob view is focused, that styling is higher precedence than the overrides we provide. More confusingly, sometimes (it's not deterministic) the codemirror styling does not respect the light/dark theme settings, which makes selection look very broken. It's still not clear to me why the theme variable is not being respected, but given our proximity to code freeze, this just adds an `!important` to our theme overrides to make it not feel super broken.

[Slack thread](https://sourcegraph.slack.com/archives/C05MHAP318B/p1707329576741979)

Before:
image

A capture of the CSS selector that was problematic:
image

## Test plan

Visually checked that there are no more dark backgrounds and opened the console to double check that the `!important` was showing up and causing that style to apply at higher precedence than the built-in themes

### DIFF
--- a/client/branded/src/search-ui/input/BaseCodeMirrorQueryInput.tsx
+++ b/client/branded/src/search-ui/input/BaseCodeMirrorQueryInput.tsx
@@ -54,10 +54,10 @@ const staticExtensions: Extension = [
             padding: 0,
         },
         '&.cm-focused .cm-selectionLayer .cm-selectionBackground': {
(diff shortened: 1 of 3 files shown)


keegancsmith (approved):
gonna approve since the description plus simplicity of the change seem legit and technically code freeze has started (but I will let this sneak in)

PR #60347, opened 2024-02-08
Add external URL to pings

dadlerj:
## Test plan

tested locally, validated ping data was correct in `event_logs` table

## Preview 🤩
[Preview Link](https://docs.sourcegraph.com/@server-url-pings)


varungandhi-src (approved):

eseliger (approved):
Im sure you thought about it, but could there be customers that get upset if we expose their company internal domains?

dadlerj:
@eseliger it was definitely something we discussed internally, this has come up before.

The deciding factor here is that our V2 telemetry already includes this property: https://docs.sourcegraph.com/dev/background-information/telemetry/protocol#identifier-licensedinstanceidentifier. So the work to ensure that customers and contracts were updated for this change has already happened.

eseliger:
Great, thanks!

Issue #60341, opened 2024-02-08
☂️ Search: solidify content-based language filtering


jtibshirani:
Currently, we translate `lang` filters to filters on file extensions. Because multiple languages may share the same extension, this can lead to errors. For example, `lang:matlab` also matches Objective C files, since they also end in `.m`.

We have a feature `search-content-based-lang-detection` that instead matches against the actual language of the file, as determined by go-enry. This issue tracks work to solidify the feature so we're comfortable recommending it to customers.

- [x] https://github.com/sourcegraph/sourcegraph/pull/60349
    - Simplify the Zoekt queries we create
    - Make sure exclude filters like `-lang` work correctly
    - Combine multiple filters through 'and' instead of 'or'
- [x] https://github.com/sourcegraph/zoekt/pull/740
- [x] https://github.com/sourcegraph/sourcegraph/pull/60496
- [x] https://github.com/sourcegraph/sourcegraph/pull/60625
- [x] https://github.com/sourcegraph/sourcegraph/pull/60626

/cc @sourcegraph/search-platform

jtibshirani:
I'd appreciate your thoughts on the new behavior for `lang:cpp`.
* Before: we always include files ending in `.cpp`, `.hpp`, and `.h`. This includes C files that end in `.h`, which can be confusing (since you see both "C++" and "C" listed in the filters sidebar).
* Now we always return `.cpp` and `.hpp` because these are unambiguous. We also return `.h` files if they resemble C++, like import something from the standard library or use a namespace.

Downsides of new behavior:
* It's not super common, but you can have `.h` files in a C++ repo that just happen to resemble C. See [this example](https://sourcegraph.com/search?q=context%3Aglobal+repo%3A%5Egithub%5C.com%2Fgrf-labs%2Fgrf%24+file%3A%5C.h+lang%3Ac&patternType=keyword&sm=0&filters=%5B%5B%22lang%22%2C%22C%22%2C%22lang%3Ac%22%5D%5D&feat=search-content-based-lang-detection) from the open source repo `grf-labs/grf`.
* go-enry has some classification errors, where something is clearly C++ but it labels it as C (see [this example](https://sourcegraph.test:3443/github.com/grf-labs/grf@25f7756acaea2b926a6dd449060c2d3d53f71fae/-/blob/core/third_party/Eigen/src/plugins/MatrixCwiseUnaryOps.h) from same repo).

[...]

PR #60349, opened 2024-02-09
Search: solidify conversion to Zoekt queries

jtibshirani:
For the 'search-content-based-lang-detection' feature, this PR improves how filters are converted to Zoekt queries:
* Only convert filters to Zoekt lang filters, instead of also including file filters
* Handle excluded filters as well, for example `-lang:go`
* Use 'and' instead of 'or' to combine multiple filters

Relates to https://github.com/sourcegraph/sourcegraph/issues/60341

## Test plan

Added new unit tests, manual testing

### DIFF
--- a/internal/search/zoekt/query.go
+++ b/internal/search/zoekt/query.go
@@ -38,9 +38,9 @@ func QueryToZoektQuery(b query.Basic, resultTypes result.Types, feat *search.Fea
 	// Handle file: and -file: filters.
 	filesInclude, filesExclude := b.IncludeExcludeValues(query.FieldFile)
 	// Handle lang: and -lang: filters.
-	// Languages are already partially expressed with IncludePatterns, but Zoekt creates
-	// more precise language metadata based on file contents analyzed by go-enry, so it's
-	// useful to pass lang: queries down.
+	// By default, languages are converted to file filters. When the 'search-content-based-lang-detection'
+	// feature is enabled, we use Zoekt's native language filters, which are based on the actual language
+	// of the file (as determined by go-enry).
 	langInclude, langExclude := b.IncludeExcludeValues(query.FieldLang)
 	if feat.ContentBasedLangFilters {
 		if len(langInclude) > 0 {
--- a/internal/search/zoekt/query.go
+++ b/internal/search/zoekt/query.go
@@ -47,11 +47,9 @@ func QueryToZoektQuery(b query.Basic, resultTypes result.Types, feat *search.Fea
 		for _, lang := range langInclude {
 			and = append(and, toLangFilter(lang))
 		}
-		if len(langExclude) > 0 {
-			for _, lang := range langExclude {
-				filter := toLangFilter(lang)
-				and = append(and, &zoekt.Not{Child: filter})
-			}
+		for _, lang := range langExclude {
+			filter := toLangFilter(lang)
+			and = append(and, &zoekt.Not{Child: filter})
 		}
 	} else {
 		filesInclude = append(filesInclude, mapSlice(langInclude, query.LangToFileRegexp)...)
--- a/internal/search/zoekt/query.go
+++ b/internal/search/zoekt/query.go
@@ -37,19 +37,21 @@ func QueryToZoektQuery(b query.Basic, resultTypes result.Types, feat *search.Fea
 
 	// Handle file: and -file: filters.
 	filesInclude, filesExclude := b.IncludeExcludeValues(query.FieldFile)
+
 	// Handle lang: and -lang: filters.
 	// By default, languages are converted to file filters. When the 'search-content-based-lang-detection'
 	// feature is enabled, we use Zoekt's native language filters, which are based on the actual language
 	// of the file (as determined by go-enry).
 	langInclude, langExclude := b.IncludeExcludeValues(query.FieldLang)
 	if feat.ContentBasedLangFilters {
-		if len(langInclude) > 0 {
-			filters := toLangFilters(langInclude)
-			and = append(and, &zoekt.Or{Children: filters})
+		for _, lang := range langInclude {
+			and = append(and, toLangFilter(lang))
 		}
 		if len(langExclude) > 0 {
-			filters := toLangFilters(langExclude)
-			and = append(and, &zoekt.Not{Child: &zoekt.Or{Children: filters}})
+			for _, lang := range langExclude {
+				filter := toLangFilter(lang)
+				and = append(and, &zoekt.Not{Child: filter})
+			}
 		}
 	} else {
 		filesInclude = append(filesInclude, mapSlice(langInclude, query.LangToFileRegexp)...)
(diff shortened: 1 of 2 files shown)


internal/search/zoekt/query.go
```diff
@@ -68,26 +82,18 @@ func QueryToZoektQuery(b query.Basic, resultTypes result.Types, feat *search.Fea
...
 	}
 
-	// Languages are already partially expressed with IncludePatterns, but Zoekt creates
-	// more precise language metadata based on file contents analyzed by go-enry, so it's
-	// useful to pass lang: queries down.
-	//
-	// Currently, negated lang queries create filename-based ExcludePatterns that cannot be
-	// corrected by the more precise language metadata. If this is a problem, indexed search
-	// queries should have a special query converter that produces *only* Language predicates
-	// instead of filepatterns.
```
varungandhi-src, line 78: This is no longer applicable because we're not creating "filename-based ExcludePatterns" based on negated lang queries, correct?

internal/search/zoekt/query.go
```diff
@@ -68,26 +82,18 @@ func QueryToZoektQuery(b query.Basic, resultTypes result.Types, feat *search.Fea
...
-		and = append(and, or)
-	}
-
 	return zoekt.Simplify(zoekt.NewAnd(and...)), nil
 }
 
+func toLangFilters(langs []string) []zoekt.Q {
+	var filters []zoekt.Q
+	for _, lang := range langs {
+		lang, _ = enry.GetLanguageByAlias(lang) // Invariant: lang is valid.
```
varungandhi-src, line 91: I know Go doesn't have built-in assertions and it's [documented why](https://go.dev/doc/faq#assertions), but I think it would be helpful to assert that the second return value is true to catch potential bugs.

internal/search/zoekt/query_test.go
```diff
@@ -83,19 +83,31 @@ func TestQueryToZoektQuery(t *testing.T) {
...
+			Query: `lang:Go or lang:Typescript`,
+		},
+		{
+			Name:    "Excluded languages get passed as lang: query",
+			Type:    search.TextRequest,
+			Pattern: `lang:go -lang:typescript -lang:markdown`,
+			Features: search.Features{
+				ContentBasedLangFilters: true,
+			},
+			Query: `lang:Go -(lang:Typescript or lang:markdown)`,
```
varungandhi-src, line 101: Is the test-case doing a case-insensitive comparison (specifically for the `lang:` filter values)?

It seems odd that `markdown` is not getting translated to `Markdown`. 🤔 

https://sourcegraph.com/github.com/go-enry/go-enry/-/blob/data/alias.go?L554

Similarly, I would expect the spelling to be `TypeScript` instead of `typescript`. https://sourcegraph.com/github.com/go-enry/go-enry/-/blob/data/alias.go?L958

internal/search/zoekt/query.go
```diff
@@ -30,16 +30,30 @@ func QueryToZoektQuery(b query.Basic, resultTypes result.Types, feat *search.Fea
...
 
+	var and []zoekt.Q
+	if q != nil {
+		and = append(and, q)
+	}
+
 	// Handle file: and -file: filters.
 	filesInclude, filesExclude := b.IncludeExcludeValues(query.FieldFile)
 	// Handle lang: and -lang: filters.
+	// Languages are already partially expressed with IncludePatterns, but Zoekt creates
```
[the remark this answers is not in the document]
varungandhi-src: I too would like to know this 😄 I was confused when I read this code before.

varungandhi-src (approved):
LGTM, left some minor suggestions and questions. 👍🏽

internal/search/zoekt/query.go
jtibshirani: That's right.

internal/search/zoekt/query.go
```diff
@@ -68,26 +82,18 @@ func QueryToZoektQuery(b query.Basic, resultTypes result.Types, feat *search.Fea
...
-	// Languages are already partially expressed with IncludePatterns, but Zoekt creates
-	// more precise language metadata based on file contents analyzed by go-enry, so it's
-	// useful to pass lang: queries down.
-	//
-	// Currently, negated lang queries create filename-based ExcludePatterns that cannot be
-	// corrected by the more precise language metadata. If this is a problem, indexed search
-	// queries should have a special query converter that produces *only* Language predicates
-	// instead of filepatterns.
-	if len(langInclude) > 0 && feat.ContentBasedLangFilters {
-		or := &zoekt.Or{}
```
jtibshirani, line 80: I believe this was incorrect -- we should be combining these filters through an 'and', not an 'or'. The new logic makes sure to use 'and'.

internal/search/zoekt/query_test.go
```diff
@@ -83,19 +83,46 @@ func TestQueryToZoektQuery(t *testing.T) {
...
 			Type:    search.TextRequest,
-			Pattern: `file:\.go$ lang:go`,
+			Pattern: `lang:go`,
+			Features: search.Features{
+				ContentBasedLangFilters: true,
+			},
+			Query: `lang:Go`,
+		},
+		{
+			Name:    "Multiple languages get passed as lang queries",
```
stefanhengl, line 101: Not related to this PR, but these tests are really hard to read to for me. It is not obvious what exactly is being compared. I think it would be much easier to understand if we had something similar to
```Go
cases : = []struct{
    Query string
    WantZoektQuery string
}{...}
```

where `WantZoektQuery` is the output of `zoekt.Q.String()`

stefanhengl (approved):

internal/search/zoekt/query.go
keegancsmith: agreed. Maybe the person who originally implemented this thought it was convenient to use or here because `(and lang:go lang:c)` is an empty set. I think when we worked on this originally the sg query language didn't support or queries? Either way I agree with using and.

keegancsmith (approved):
LGTM

internal/search/zoekt/query.go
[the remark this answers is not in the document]
keegancsmith: If I where to guess to either had to do with rollout (concerns around old lang data in indexes? I think we switched to enry at the same time in zoekt) or some sembelance of backwards compatibility. The original author of this feature was really concerned about things like breaking lang filters for c/c++ code.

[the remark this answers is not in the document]
jtibshirani: Thanks, this makes sense out of concern for back-compat. I feel good about fully cutting over now, since it's now been a while since we added 'Language' metadata to Zoekt.

internal/search/zoekt/query_test.go
jtibshirani: I think the translation process will normalize the language values. I've converted everything to lowercase to make it clear that we're not doing a case-sensitive comparison.

internal/search/zoekt/query_test.go
jtibshirani: Agreed! For now I just improved the naming:
* `Query` -> `WantZoektQuery`
* `Pattern` -> `Query`

PR #60355, opened 2024-02-09
bazel: remove external tag

burmudar:
> external keyword will force test to be unconditionally executed (regardless of --cache_test_results value).

@rafax double checking whether this was intentional? ie. To always run the test and never use cached results

## Test plan
Test should be cached


rafax:
This test runs against an external environment, so results can change even if the test did not (so, we don't want caching)

burmudar:
> This test runs against an external environment, so results can change even if the test did not (so, we don't want caching)

Ah gotcha. I've added a comment to explain this

rafax (approved):

Issue #60119, opened 2024-02-03
Search preview: can get too big and consume search results


camdencheek:
https://github.com/sourcegraph/sourcegraph/assets/12631702/2af6bfdd-5915-446e-b975-8a79c9edcd3a

PR #60368, opened 2024-02-09
Search-results page: Fix search result layout block spacings

vovakulikov:
[Original Slack report message](https://sourcegraph.slack.com/archives/C05D629GWJK/p1707480193201689) 
Fixes https://github.com/sourcegraph/sourcegraph/issues/60119

Prior to this PR we used negative margins to have paddings for all child content in the content area but made an exception for actual search result matches block (which should have 0 spacing between parent border). This layout doesn't work in case we render content and some other elements on the same level, because of exactly negative margins of the content block. See the image below. 

image

So this PR adds manual wrappings for spacing and handles content children spacing manually as well.  

Also, this PR fixes some regression after #60220, now we render no zero-state content again. This is probably was just a mistake there but if it wasn't @camdencheek let me know

## Test plan
- Check that the layout looks okay with different combinations of states 
   - Just content 
   - Content with limit hit message
   - No content state
   - Content with some smart suggestions

### DIFF
--- a/client/branded/src/search-ui/results/StreamingSearchResultsFooter.tsx
+++ b/client/branded/src/search-ui/results/StreamingSearchResultsFooter.tsx
@@ -7,29 +7,30 @@ import { Alert, LoadingSpinner, Code, Text, H2, H3, ErrorAlert } from '@sourcegr
 
 import { StreamingProgressCount } from './progress/StreamingProgressCount'
 
-import styles from './StreamingSearchResultsList.module.scss'
+import styles from './StreamingSearchResultsFooter.module.scss'
 
 export const StreamingSearchResultFooter: React.FunctionComponent<
     React.PropsWithChildren<{
         results?: AggregateStreamingSearchResults
         children?: React.ReactChild | React.ReactChild[]
+        className?: string
     }>
-> = ({ results, children }) => {
+> = ({ results, children, className }) => {
     const skippedDisplay =
         results?.state === 'complete' && results.progress.skipped.find(skipped => skipped.reason.includes('display'))
     const resultLimitHit =
         results?.state === 'complete' && results.progress.skipped.some(skipped => skipped.reason.includes('-limit'))
 
     return (
-        <div className={classNames(styles.contentCentered, 'd-flex flex-column align-items-center')}>
+        <div className={classNames(className, styles.root)}>
             {(!results || results?.state === 'loading') && (
-                <div className="text-center my-4" data-testid="loading-container">
+                <div className="text-center" data-testid="loading-container">
                     <LoadingSpinner />
                 </div>
             )}
 
             {results?.state === 'complete' && results?.results.length > 0 && (
-                <StreamingProgressCount progress={results.progress} state={results.state} className="mt-4 mb-2" />
+                <StreamingProgressCount progress={results.progress} state={results.state} />
             )}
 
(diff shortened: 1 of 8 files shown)


fkling (approved):

client/web/src/search/results/components/new-search-content/NewSearchContent.tsx
```diff
@@ -403,6 +407,7 @@ const FilePreviewPanel: FC<FilePreviewPanelProps> = props => {
         <Panel
             defaultSize={300}
             minSize={256}
+            maxSize={600}
```
camdencheek, line 410: Might as well sneak this one in 🙂

camdencheek:
> Also, this PR fixes some regression after https://github.com/sourcegraph/sourcegraph/pull/60220, now we render no zero-state content again.

Oh! What was this? Could you post a screenshot?

client/web/src/search/results/components/new-search-content/NewSearchContent.tsx
```diff
@@ -251,80 +251,84 @@ export const NewSearchContent: FC<NewSearchContentProps> = props => {
...
-                        onQuerySubmit={onQuerySubmit}
-                        telemetryService={telemetryService}
-                    />
-                )}
-
-                {aggregationUIMode !== AggregationUIMode.SearchPage && (
-                    <>
-                        <DidYouMean
-                            telemetryService={props.telemetryService}
+                <div className={styles.contentPaddingWrapper}>
```
camdencheek, line 254: Shouldn't this div be inside the conditionals? Otherwise, won't we be adding an extra, padded div even when it's empty?

client/branded/src/search-ui/results/StreamingSearchResultsFooter.tsx
```diff
@@ -61,6 +60,8 @@ export const StreamingSearchResultFooter: React.FunctionComponent<
                     )}
                 </Alert>
             )}
+
+            {children}
```
camdencheek, line 64: Thank you!

vovakulikov:
@camdencheek Basically in that PR in this [line](https://github.com/sourcegraph/sourcegraph/pull/60220/files#diff-3f9b9d5eb2de6f0cef6d9d0385a2ca20ac3310525f4da3019534325ac89d1a35L53) we removed children rendering, these children were some zero-state content such as query examples and docs links.

Screenshot 2024-02-09 at 12 11 43

client/web/src/search/results/components/new-search-content/NewSearchContent.tsx
camdencheek: Confirmed this is an issue

CleanShot 2024-02-09 at 08 13 28@2x

camdencheek: Also, I think this might be the wrong padding for the "Did you mean" element

CleanShot 2024-02-09 at 08 15 20@2x

vovakulikov: Good call, moving this to the condition in jsx won't solve the problem though, because this condition is about content type rendering (aggregation UI chart or actual content) 

I added pseudo-state in CSS to hide this element when we don't have anything in it, to avoid extra spacing

camdencheek (approved):

PR #60370, opened 2024-02-09
Add overhead tracking middleware for Cody Gateway

rafax:
Adds middleware tracking the "overhead" of Cody Gateway - the difference between CG response latency and latency of backend APIs as measured by CG. Currently, we don't have a great way to measure this in a single place:

- API latency is measured by GCP metrics, but only reported in aggregation (embeddings / chat / completions have very different latency characteristics)
- upstream_latency is measured in some BigQuery events, but it's unpractical to make it include response latency (we'd need to delay the event until all response handling is done)
- we should be able to get this from tracing data, but:
    - GCP tracing is unreliable
    - our sampling strategy is a mystery
    - our current tracing tooling doesn't allow us to calculate this

Also changes the way we measure upstream latency not to include the time we spent on parsing responses (but we still include the time we took to read the whole response).

## Test plan

- locally tested 
- e2e tests

### DIFF
--- a/cmd/cody-gateway/internal/httpapi/overhead/overhead.go
+++ b/cmd/cody-gateway/internal/httpapi/overhead/overhead.go
@@ -1,0 +1,45 @@
+package overhead
+
+import (
+	"context"
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/sourcegraph/sourcegraph/internal/codygateway"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+)
+
+// HTTPMiddleware wraps all downstream handlers and exposes a histogram measuring the difference between the Cody Gateway API latency and the backend API latency.
+func HTTPMiddleware(apiHistogram metric.Int64Histogram, handler http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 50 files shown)


philipp-spiess:
> our sampling strategy is a mystery

Hehe

cmd/cody-gateway/internal/httpapi/overhead/overhead.go
```diff
@@ -0,0 +1,45 @@
...
+func HTTPMiddleware(apiHistogram metric.Int64Histogram, handler http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		start := time.Now()
+		r = r.WithContext(context.WithValue(r.Context(), overheadKey, &RequestLatencyData{}))
+
+		handler.ServeHTTP(w, r)
+
+		latency := time.Since(start)
+		o := FromContext(r.Context())
+		overheadMs := latency.Milliseconds() - o.UpstreamLatency.Milliseconds()
```
philipp-spiess, line 24: Nice, makes sense!

philipp-spiess (approved):
Looks good to me :)

valerybugakov (commented):
> our sampling strategy is a mystery

🪄

valerybugakov (approved):

PR #60371, opened 2024-02-09
site admin: add link to Cody Analytics

taras-yemets:
Adds a link to Cody Analytics service.
[Figma](https://www.figma.com/file/pMaNBzIz3to8ujUsRX4f7I/Cody-analytics?type=design&node-id=8-32&mode=design&t=QqFheUIS2zNQYY28-0)

Screenshot 2024-02-09 at 18 45 33

## Test plan
- CI
- Tested manually

### DIFF
--- a/client/web/src/site-admin/routes.tsx
+++ b/client/web/src/site-admin/routes.tsx
@@ -26,6 +26,7 @@ const AnalyticsExtensionsPage = lazyComponent(
     'AnalyticsExtensionsPage'
 )
 const AnalyticsUsersPage = lazyComponent(() => import('./analytics/AnalyticsUsersPage'), 'AnalyticsUsersPage')
+const AnalyticsCodyPage = lazyComponent(() => import('./analytics/AnalyticsCodyPage'), 'AnalyticsCodyPage')
 const AnalyticsCodeInsightsPage = lazyComponent(
     () => import('./analytics/AnalyticsCodeInsightsPage'),
     'AnalyticsCodeInsightsPage'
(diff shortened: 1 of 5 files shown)


chwarwick (commented):
Users will need to request access right? It's probably worth a note about how to do that.

chwarwick (approved):

client/web/src/site-admin/routes.tsx
```diff
@@ -217,6 +218,10 @@ export const otherSiteAdminRoutes: readonly SiteAdminAreaRoute[] = [
         path: '/analytics/users',
         render: () => <AnalyticsUsersPage />,
     },
+    {
+        path: '/analytics/cody',
+        render: () => <AnalyticsCodyPage />,
```
BolajiOlajide, line 223: Hey @taras-yemets , I know it's a bit late but we should be gating this path in the side-admin page.

This should only show up when the license has Cody enabled. I realize you have a backport PR and this is merged but do you think you can push a fix for this to `main` and another to the `backport branch`

CleanShot 2024-02-12 at 01 43 09@2x

taras-yemets: Nice catch, @BolajiOlajide! Thank you! 
I ensured the page is not accessible via sidebar UI (https://github.com/sourcegraph/sourcegraph/pull/60371/commits/c71ffac46155b72378294bf113af234eb189322a), but missed to do the same on the route level.
Fixed in https://github.com/sourcegraph/sourcegraph/pull/60396.
I'll update the backport PR when the one to `main` is merged.

BolajiOlajide: Thank you @taras-yemets

rrhyne:
@taras-yemets here is the original screen I designed for this:

https://www.figma.com/file/pMaNBzIz3to8ujUsRX4f7I/Cody-analytics?type=design&node-id=8-32&mode=design&t=IhCdIeficqxYo6X2-4

I've replaced the image with a current snapshot of the design. Can we get that included so this page is a little less bare? 

Also, we need a way to link back to the instance on the analytics site. The analytics service seems to know about the instance as it reads sourcegraph.sourcegraph.com in the top right, but it doesn't link to the instance, and doesn't contain anything when there is once instance so it appears broken. 

image

PR #60376, opened 2024-02-09
licensing: update doc links to the new doc website

BolajiOlajide:
## Test plan

All links to the old docs page `docs.sourcegraph.com` should be updated to their `sourcegraph.com/docs` equivalent.

### DIFF
--- a/client/web/src/cody/dashboard/CodyDashboardPage.tsx
+++ b/client/web/src/cody/dashboard/CodyDashboardPage.tsx
@@ -38,13 +38,13 @@ const setupOptions: SetupOption[] = [
         icon: <VSCodeIcon className={styles.linkSelectorIcon} />,
         maker: 'Microsoft',
         name: 'VSCode',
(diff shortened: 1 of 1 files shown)


camdencheek (approved):
Oof. Looks like there are more `docs.sourcegraph.com` links around the codebase as well. Should these be pointing to the docs for the current sourcegraph version? Right now, this just directs the user to the latest version. The `/help/...` route will redirect to the docs corresponding to the instance version.

vovakulikov (approved):

BolajiOlajide:
> Oof. Looks like there are more `docs.sourcegraph.com` links around the codebase as well. Should these be pointing to the docs for the current sourcegraph version? Right now, this just directs the user to the latest version. The `/help/...` route will redirect to the docs corresponding to the instance version.

Ideally yes, but a lot of this change aren't easy to migrate without a manual check. Especially because the new docs site url structure isn't the same as the old.

I prioritized the new links added by the licensing work because it felt like the smallest change I could make this close to the release and with fewer stakeholders involved.

PR #60377, opened 2024-02-09
v2 telemetry: add v2 telemetry to src/communitySearchContexts

dadlerj:
Context:

## Test plan

`sg start`
visit page

### DIFF
--- a/client/web/src/communitySearchContexts/CommunitySearchContextPage.tsx
+++ b/client/web/src/communitySearchContexts/CommunitySearchContextPage.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useMemo, useState } from 'react'
+import React, { useCallback, useEffect, useMemo, useState } from 'react'
 
 import { mdiSourceRepositoryMultiple, mdiGithub, mdiGitlab, mdiBitbucket } from '@mdi/js'
 import classNames from 'classnames'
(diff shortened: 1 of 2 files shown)


dadlerj:
This is small/simple PR that is in good shape for review. I'm honestly struggling to know how to test these pages though. Can you provide any guidance on, like, what settings to set and what URLs to visit to actually test locally @vovakulikov ?

vovakulikov:
@dadlerj good questions, I remember that these pages were created for setting `sourcegraph.com` instance contexts with different repositories. I'm not sure this is still in use actually, I will double-check and will be back with updates, thanks for asking.

client/web/src/communitySearchContexts/CommunitySearchContextPage.tsx
```diff
@@ -73,22 +89,37 @@ export const CommunitySearchContextPage: React.FunctionComponent<
...
+    )
+
+    const onRepoLinkClicked = useCallback(
+        (repoName: string) => (): void => {
+            eventLogger.log(
+                'CommunitySearchContextPageRepoLinkClicked',
+                { repo_name: repoName },
+                { repo_name: repoName }
+            )
+            telemetryRecorder.recordEvent('communitySearchContext.repo-link', 'click')
```
vovakulikov, line 119: Is this right and we don't need to collect `repoName` anymore or we should add repo name to the metadata here?

client/web/src/communitySearchContexts/CommunitySearchContextPage.tsx
```diff
@@ -211,14 +244,21 @@ export const CommunitySearchContextPage: React.FunctionComponent<
...
+> = ({ repo, onRepoLinkClicked }) => (
     <li className={classNames('list-unstyled mb-3', styles.repoItem)} key={repo}>
         {repo.startsWith('github.com') && (
             <>
-                <Link to={`https://${repo}`} target="_blank" rel="noopener noreferrer" onClick={RepoLinkClicked(repo)}>
+                <Link
+                    to={`https://${repo}`}
+                    target="_blank"
+                    rel="noopener noreferrer"
+                    onClick={onRepoLinkClicked(repo)}
```
vovakulikov, line 260: Not a blocker but just a small suggestion: Usually we try to avoid explicit closure calls like here
since you have to type these functions as nested callbacks like `onRepoLinkClicked: (repo: string) => (_: any) => void`. 

We could just use the plain/flat function onRepoLinkClicked which takes the repo data as an argument and then on the link click in the `onClick` handler we could do a closure wrapping `onClick={() => onRepoLinkClicked(repo)}`

client/web/src/communitySearchContexts/types.ts
```diff
@@ -15,7 +26,7 @@ export interface CommunitySearchContextMetadata {
     /**
      * The name of the community search context, must match the community search context name as configured in settings. E.g. python2-to-3.
      */
-    spec: string
+    spec: CommunitySearchContextSpecs
```
vovakulikov, line 29: Cool

vovakulikov (approved):
Left a few comments. I almost forgot that we have these pages)

vovakulikov (approved):

PR #60380, opened 2024-02-09
Bug fix: add label to OpenCodeGraph icon

jasonhawkharris:
Before, the OpenCodeGraph icon was the only action without a label: 
Screenshot 2024-02-09 at 2 00 46 PM

Now, the icon has a label, along with a more descriptive tooltip.
Screenshot 2024-02-09 at 2 46 59 PM

A few questions for the team:
1. @taiyab from a design standpoint, is "graph" enough here? @sqs I imagine you might have thoughts here too? I tried `OpenCodeGraph` and `CodeGraph` and felt like they both looked a little clunky next to the rest of the labels. Open to suggestions here. 
2. Is this the best place for this code to live? The rest of the action toggles seem to live in either `client/web/repo/blob/actions` or `client/web/repo/actions`. Should we move this toggle there as well? It wasn't a big deal to find it, but it did take a bit more digging than expected. 
 
# Test plan
Manual/visual testing. CI/CD passing.


sqs:
It might be renamed to OpenCtx anyway. Graph is fine for now. This is experimental anyway. :) Thanks!

camdencheek:
Of note: AFAIK, OpenCodeGraph is an experimental feature that is only enabled on dotcom and s2. 

Given the tight space budget we have here (since we are sharing the bar with the path), I'd be inclined to pin this to the overflow menu with the full "OpenCodeGraph" name. But, since this is an experimental feature, the non-optimality of the design isn't a huge deal.

camdencheek (approved):

BolajiOlajide (approved):
Thank you for the fix @jasonhawkharris .

From a technical standpoint, it looks solid. 
I did the initial work on the actions workflow so I'll try to answer the questions you asked.

> The rest of the action toggles seem to live in client/web/src/repo. Whereas, the OpenCodeGraph toggle lives in client/web/src/opencodegraph/global. It wasn't a big deal to find it, but it did take a bit more digging than expected. Seems a bit out of place to me.

You're right, it currently is that way because of the way we're rendering the repo actions. We make use the concept of Portals to render these actions from different parts of the UI. It's not the easiest to work with but it worked for the use-case of rendering different elements from outside the context of the current DOM component.

> I imagine you might have thoughts here too? I tried OpenCodeGraph and CodeGraph and felt like they both looked a little clunky next to the rest of the labels.

Yes, this was the reason why I didn't push up a PR. @taiyab and I synced on it but couldn't come to a conclusion on this before licensing took most of our focus. `OpenGraph` feels a bit clunkty to me also. However, it's an experimental feature as of today so I don't think it should cause a lot of problems if we leave it as `Graph`

PR #60383, opened 2024-02-09
Batch changes: paginate when listing github apps

camdencheek:
Because we were not paginating, we would end up deleting all installations when there were more than 30 (default page size).

## Test plan

Added a unit test and @BolajiOlajide manually tested the fix.

[Walkthrough of Bolaji running a dead simple test](https://www.loom.com/share/4d1254e072db4973b1f17645b13b5654?sid=d392589a-bd31-48e2-a3a1-071a1d3bf678)

### DIFF
--- a/internal/extsvc/github/v3.go
+++ b/internal/extsvc/github/v3.go
@@ -894,7 +894,7 @@ func (c *V3Client) GetAppInstallation(ctx context.Context, installationID int64)
 // GetAppInstallations fetches a list of GitHub App instalaltions for the
 // authenticated GitHub App.
 //
-// API docs: https://docs.github.com/en/rest/reference/apps#get-an-installation-for-the-authenticated-app
+// API docs: https://docs.github.com/en/rest/apps/apps#list-installations-for-the-authenticated-app
 func (c *V3Client) GetAppInstallations(ctx context.Context, page int) (ins []*github.Installation, hasNextPage bool, _ error) {
 	respState, err := c.get(ctx, fmt.Sprintf("app/installations?page=%d", page), &ins)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


internal/extsvc/github/v3.go
```diff
@@ -895,12 +895,12 @@ func (c *V3Client) GetAppInstallation(ctx context.Context, installationID int64)
 // authenticated GitHub App.
 //
 // API docs: https://docs.github.com/en/rest/reference/apps#get-an-installation-for-the-authenticated-app
```
peterguy, line 897:
```suggestion
// API docs: https://docs.github.com/en/rest/apps/apps#list-installations-for-the-authenticated-app
```

camdencheek: Thanks!

internal/github_apps/store/store.go
```diff
@@ -458,7 +459,22 @@ func (s *gitHubAppsStore) SyncInstallations(ctx context.Context, app ghtypes.Git
 		return errors.Append(errs, err)
 	}
 
-	remoteInstallations, err := client.GetAppInstallations(ctx)
+	getAllAppInstallations := func() ([]*github.Installation, error) {
+		var allInstallations []*github.Installation
+		for page := 0; ; page++ {
```
peterguy, line 464: The page number is zero-based? The [documentation](https://docs.github.com/en/rest/using-the-rest-api/using-pagination-in-the-rest-api?apiVersion=2022-11-28) has as an example, `?page=1>; rel="first"`, suggesting that it's one-based.

camdencheek: good catch! Definitely 1-based. Interpret "draft" as "entirely untested" here 😄

BolajiOlajide (approved):

PR #60388, opened 2024-02-10
Get Stevey a bigger bubble

thenamankumar:
Issue: https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1707499933575329

As reported, the text comes out of the bubble for Cody Web.
image

After testing, I found that the bug is on Firefox because the browsers treat Grids differently. 

I have replaced grid-based CSS with a basic flex model with position absolute to fit text nicely into the bubble. 

Tested on all browsers at various zoom levels.

## Test plan
- Open Cody Web in the sidebar on file view. 
- Create new chat
- The bubble should cover the whole "Hi I'm Cody" text.
- Test on all the browsers.

### DIFF
--- a/client/web/src/cody/components/GettingStarted.module.scss
+++ b/client/web/src/cody/components/GettingStarted.module.scss
@@ -11,6 +11,9 @@
 
 .icon-section {
     margin-bottom: 3rem !important;
+    display: flex;
+    align-items: flex-end;
+    justify-content: center;
 }
 
 .section {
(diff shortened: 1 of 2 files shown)


chrsmith (approved):
I'll take your word for it that the updated markup renders as expected across browsers.

I saw Percy was wired up to the sourcegraph/sourcegraph CI/CD. Is that kinda sorta the best option we have to ensure things like this don't regress and/or for cross-browser testing?

PR #60389, opened 2024-02-10
Make dotcom actor refresh cooldown configurable

rafax:
Make dotcom actor refresh cooldown configurable - see [Slack](https://sourcegraph.slack.com/archives/C05SZB829D0/p1707552838650569)
## Test plan

- tested locally
- e2e tests


thenamankumar (approved):
LGTM, I see the default is still 5m (300s). Are we going to change the config to a lower interval when we deploy this?

rafax:
> LGTM, I see the default is still 5m (300s). Are we going to change the config to a lower interval when we deploy this?

Yes, my intention is for this to be a no-op until we explicitly change the configuration

PR #60397, opened 2024-02-12
Redacting attribution Gateway access token

arafatkatze:
Solves https://github.com/sourcegraph/sourcegraph/issues/60045 

## Test plan
Run in admin and try to save access token and you shall get this. 
image

### DIFF
--- a/internal/conf/validate.go
+++ b/internal/conf/validate.go
@@ -221,6 +221,7 @@ var siteConfigSecrets = []struct {
 	{readPath: `embeddings.accessToken`, editPaths: []string{"embeddings", "accessToken"}},
 	{readPath: `completions.accessToken`, editPaths: []string{"completions", "accessToken"}},
 	{readPath: `app.dotcomAuthToken`, editPaths: []string{"app", "dotcomAuthToken"}},
+	{readPath: `attribution\.gateway.accessToken`, editPaths: []string{"attribution.gateway", "accessToken"}},
 }
 
 // UnredactSecrets unredacts unchanged secrets back to their original value for


chwarwick (approved):
👍 looks like something that would be appropriate to backport

internal/conf/validate.go
```diff
@@ -221,6 +221,7 @@ var siteConfigSecrets = []struct {
 	{readPath: `embeddings.accessToken`, editPaths: []string{"embeddings", "accessToken"}},
 	{readPath: `completions.accessToken`, editPaths: []string{"completions", "accessToken"}},
 	{readPath: `app.dotcomAuthToken`, editPaths: []string{"app", "dotcomAuthToken"}},
+	{readPath: `attribution\.gateway.accessToken`, editPaths: []string{"attribution.gateway", "accessToken"}},
```
keegancsmith, line 224: @arafatkatze @chwarwick is this correct? I see you escape the `.` (`\.`). I don't see the other readPaths doing that, nor do you consistently do that in this string.

keegancsmith:
@arafatkatze FYI to create a backport to the 5.3 branch you can just add the label `backport 5.3`. I see you manually did that, but gonna rely on automation here since it will create a more traceable commit message/etc.

internal/conf/validate.go
arafatkatze: This is correct. 

`	{readPath: `email\.smtp.username`, editPaths: []string{"email.smtp", "username"}},`

I copied it from here

Basically it supports
```

{
 "attribution.gateway": { 
    "accesstoken": "REDACTED"
   }
}
```

arafatkatze:
@keegancsmith Thanks a lot for the help here and sorry for all the PR clutter.

PR #60398, opened 2024-02-12
all: update code paths as if qdrant is disabled

keegancsmith:
Fully removing our use of qdrant is quite a large change. Instead this does a minimal change by just removing the part of our code which optionally searches via qdrant, and additionally the part which uses a real qdrant client. This effectively makes it so qdrant is always disabled.

In practice, the "GetEmbeddingsConfig" will only return a configuration for dotcom. Currently qdrant is disabled on dotcom, so operationally this is a noop change.

Test Plan: CI and investigating that live environments this is a noop.

### DIFF
--- a/cmd/worker/internal/embeddings/repo/worker.go
+++ b/cmd/worker/internal/embeddings/repo/worker.go
@@ -54,8 +54,9 @@ func (s *repoEmbeddingJob) Routines(_ context.Context, observationCtx *observati
 		return nil, err
 	}
 
-	getQdrantDB := vdb.NewDBFromConfFunc(observationCtx.Logger, vdb.NewNoopDB())
-	getQdrantInserter := func() (vdb.VectorInserter, error) { return getQdrantDB() }
+	// qdrant is going to be removed. For now we only ever set the noop db.
+	qdrantDB := vdb.NewNoopDB()
+	getQdrantInserter := func() (vdb.VectorInserter, error) { return qdrantDB, nil }
 
 	workCtx := actor.WithInternalActor(context.Background())
 	return []goroutine.BackgroundRoutine{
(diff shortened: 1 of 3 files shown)


keegancsmith:
This is a very minor change enforcing that we are not actually using qdrant right now. We can follow up with the more intensive code removal, but this seems like a good idea to first land this and then just socialize the idea further that we are not planning on investing in more scalable embeddings search yet.

keegancsmith:
TBH the `VectorDB` interface is quite nice. Even though we plan on removing qdrant, its likely worthwhile as a bit of tech debt cleanup to make the embeddings search use the same interface.

stefanhengl (approved):

chwarwick (approved):

keegancsmith:
@camdencheek I am seeing references to qdrant in the configuration of s2. However, reading the code it looks like unless you are dotcom we don't setup the embeddings store... so are we currently in the place that embedding searches on s2 use qdrant but nothing updates it? Or more likely, embedding code paths are just unused on s2 and that is how we got in that state.

chwarwick:
> @camdencheek I am seeing references to qdrant in the configuration of s2. However, reading the code it looks like unless you are dotcom we don't setup the embeddings store... so are we currently in the place that embedding searches on s2 use qdrant but nothing updates it? Or more likely, embedding code paths are just unused on s2 and that is how we got in that state.

This is correct on any enterprise install embeddings are forcably disabled, I didn't fully remove things because embeddings were still in use on dotcom.  I'll be following up to finish the clean up post 5.3 launch.

camdencheek:
Deferring to Chris on this 🙂 I have no idea what the current state is

PR #60400, opened 2024-02-12
Search results: Modified filter panel background color

taiyab:
Using a more appropriate color variable for the background of the filter panel.

## Before

image

## After

image

## Test plan
Manually tested.


fkling:
What does it look like in light mode?

vovakulikov (approved):

PR #60405, opened 2024-02-12
code-monitors: handle soft-deleted users

stefanhengl:
Code monitors from soft-deleted users cause errors in the client. For example, if an admin opens the code monitors page, they see a "user not found" error.

image

With this change we filter out code monitors of soft-deleted users. This means, the system effectively behaves as if the user was hard-deleted.

For users that were actually hard-deleted, we already delete their code monitors, which avoids this issue.

I also confirmed that we don't run code monitors of soft-deleted users.

Alternatives considered:

I first went down the path to fix this on the level of the resolver and in the client. However it turned out that this would require a lot of small changes of the GraphQL API paired with some fallback logic in the client. Overall the approach taken in this PR is much smaller and easier to understand.

Test plan:
- new unit test
- manual testing:
  - I created a code monitor as user Alice and then, as admin, soft-deleted Alice
  - I confirmed that there is no error and that, as admin, I don't see the code monitor
  - Undeleting Alice makes the code monitor reappear in the list


jtibshirani (approved):
This approach makes sense to me.

BolajiOlajide (approved):

PR #60423, opened 2024-02-12
Log when refresh is blocked by cooldown

rafax:
Log when rate-limit refresh is blocked (so we can create a log-based metric).

## Test plan

- tested locally
- e2e tests pass

### DIFF
--- a/cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
+++ b/cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
@@ -68,7 +68,7 @@ func (s *Source) Get(ctx context.Context, token string) (*actor.Actor, error) {
 
 func (s *Source) Update(ctx context.Context, act *actor.Actor) error {
 	if act.LastUpdated != nil && time.Since(*act.LastUpdated) < s.coolDownInterval {
-		s.log.Debug("actor recently updated, skipping update", log.Duration("secondsSinceUpdate", time.Since(*act.LastUpdated).Seconds()))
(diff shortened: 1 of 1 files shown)


chrsmith (approved):
Do we have DEBUG level logs enabled anywhere? Will this actually get wired up into GCP's log system?

chrsmith:
@rafax , I merged since this certainly wouldn't hurt. But I'd love if you could answer my question:

> Do we have DEBUG level logs enabled anywhere? Will this actually get wired up into GCP's log system?

My experience with AWS is that the application usually won't emit DEBUG logs to STDERR, so you'd never be able to search them or create log-based metrics.

For Cody Gateway, do we emit _all_ logs? Or perhaps there isn't the ability to prune/filter DEBUG logs in GCP like you can in AWS/CloudWatch?

rafax:
> @rafax , I merged since this certainly wouldn't hurt. But I'd love if you could answer my question:
> 
> > Do we have DEBUG level logs enabled anywhere? Will this actually get wired up into GCP's log system?
> 
> My experience with AWS is that the application usually won't emit DEBUG logs to STDERR, so you'd never be able to search them or create log-based metrics.
> 
> For Cody Gateway, do we emit _all_ logs? Or perhaps there isn't the ability to prune/filter DEBUG logs in GCP like you can in AWS/CloudWatch?

By default, we filter anything below INFO, but there is a way to enable DEBUG logging for a single logger (ex. single class) without requiring a code change (just add an envvar). Ex. `SRC_LOG_SCOPE_LEVEL=cody-gateway.events="debug",worker="error"` will allow debug logs for cody-gateway.events logger, but will only allow error-level logs for worker.

PR #60432, opened 2024-02-12
Update Manage Subscription buttons

thenamankumar:
closes: https://github.com/sourcegraph/accounts.sourcegraph.com/issues/491

Change the colour of the Manage Subscription Button on `/cody/manage` to blue and add cc icon.
CleanShot 2024-02-12 at 23 44 04@2x

For Pro users add a primary Manage Subscription button on the right side of the header on `cody/subscription` and add a link back to the dashboard below the header. 

CleanShot 2024-02-12 at 23 45 26@2x

Take care of all the feature flags and states while doing the above changes.

## Test plan

- Turn on use-ssc-for-subscription-on-web feature flag
- Upgrade to pro and add credit card
- Verify that the above buttons and links are working as expected.

### DIFF
--- a/client/web/src/cody/management/CodyManagementPage.tsx
+++ b/client/web/src/cody/management/CodyManagementPage.tsx
@@ -1,6 +1,6 @@
 import React, { useEffect } from 'react'
 
-import { mdiHelpCircleOutline, mdiInformationOutline, mdiOpenInNew } from '@mdi/js'
+import { mdiHelpCircleOutline, mdiInformationOutline, mdiOpenInNew, mdiCreditCardOutline } from '@mdi/js'
 import classNames from 'classnames'
 import { useNavigate } from 'react-router-dom'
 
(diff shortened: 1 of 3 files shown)


chrsmith (approved):
:shipit: ! Thanks for the quick turnaround.

chenkc805:
Thanks Naman! Some changes for these Pro users:

- on /cody/manage, remove the CC icon and change the CTA text to "See current plan"
- on the plans page (sourcegraph.com/cody/subscription), can we replace the text CTA on the Premium plan and move the button there instead? See screenshot: 
304185120-4a695c77-8b2b-417e-b997-cb058fcc0845

danielmarquespt:
> on /cody/manage, remove the CC icon and change the CTA text to "See current plan"

Removing the CC icon is fine. But I think "Manage Subscription" is a better label for that button.

On the tasks related with a subscription, very rarely a user wants to "See the current plan" and more often wants to "Manage (their) subscription". I think the labels should describe the most useful and common scenario for the user.


> on the plans page (sourcegraph.com/cody/subscription), can we replace the text CTA on the Premium plan and move the button there instead? See screenshot:

That makes the task of managing the subscription harder for the user

chenkc805:
> Removing the CC icon is fine. But I think "Manage Subscription" is a better label for that button.
Sounds good to me

> > on the plans page (sourcegraph.com/cody/subscription), can we replace the text CTA on the Premium plan and move the button there instead? See screenshot:
> That makes the task of managing the subscription harder for the user
I don't follow this, how does this make managing the subscription harder for the user? IMO the UI is clearer (big blue button in the middle of the page instead of the top right)

chenkc805:
I'm trying to optimize for #1, and I don't think the button should be hard to find amidst all the rest of the text, since it'll be the only big blue button on the page. So I think we get both #1 and #2 together. How's that sound?

danielmarquespt:
@chenkc805  If you want to optimize for #1 we can do that! I'll test the flow again and see if users miss it

PR #60443, opened 2024-02-12
Svelte webapp: extract dynamic filter sidebar

camdencheek:
While starting to port the sidebar changes from the react app, I found it difficult to organize and understand the dependencies with everything in one file. This just extracts the sidebar into its own package in `lib/search` with as little refactoring as possible to make that happen.

## Test plan

Manually clicked around. Will add playwright tests once I finish porting the changes from the react webapp.


client/web-sveltekit/src/lib/search/dynamicFilters/Sidebar.svelte
```diff
@@ -0,0 +1,156 @@
...
+    import type { QueryStateStore } from '$lib/search/state'
+    import SymbolKind from '$lib/search/SymbolKind.svelte'
+    import { groupFilters } from '$lib/search/utils'
+    import { displayRepoName, type Filter } from '$lib/shared'
+
+    export let queryFromURL: string
+    export let streamFilters: Filter[]
+    export let queryFilters: string
+    export let queryState: QueryStateStore
+    export let size: number
```
fkling, line 17: I don't find this prop very intuitive but I think we'll have to change how we do resizable panels anyway, so I guess it's OK for now.

fkling (approved):

client/web-sveltekit/src/lib/search/dynamicFilters/Sidebar.svelte
camdencheek: Agreed. I think I'll probably actually pull the size out into the container as a followup since that's where the separator lives. Just didn't want to add extra DOM elements in this "move things around" PR

camdencheek:
> Interesting that you got so many import reorders

I wondered about that too! I assume it's prettier that's doing that?

client/web-sveltekit/src/lib/search/dynamicFilters/Sidebar.svelte
fkling: OK. Alternatively you could move the separator into the sidebar component. You don't have to have a single top level element in a component.

camdencheek: Yeah, that could make sense too. I was mostly thinking along the lines of "the size of a component's container should be determined by its parent". Like, `ResizeablePanel` could probably be its own component. It looks like we have a `Panel` in wildcard that serves this purpose.

PR #60451, opened 2024-02-13
pings, telemetry-gateway: stop pubsub client before stopping server

bobheadxi:
Possible alleviation to recent errors like https://sourcegraph.slack.com/archives/C06CCJR4K9R/p1707434020627279 and https://sourcegraph.slack.com/archives/C06CCJR4K9R/p1707772710344069 - it's possible that server stops before the pub/sub client, cancelling the context of ongoing requests and causing them to be interrupted. This could possibly be fixed by ensuring the pub/sub client clears its backlog before we stop the server.

## Test plan

Unit tests on the new LIFOStopRoutine:

```
go test -race -count=100 -run ^TestLIFOStopRoutine$ github.com/sourcegraph/sourcegraph/lib/background
```

### DIFF
--- a/cmd/msp-example/internal/example/example.go
+++ b/cmd/msp-example/internal/example/example.go
@@ -35,7 +35,7 @@ func (s Service) Initialize(
 	logger log.Logger,
 	contract runtime.Contract,
 	config Config,
-) (background.CombinedRoutine, error) {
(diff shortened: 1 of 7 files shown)


jac (approved):

unknwon (approved):
Looks nice!

PR #60459, opened 2024-02-13
Support compression for embedding endpoint

rafax:
Support compression for `/v1/embeddings` endpoint - reasons:

- single batch (n=512) embedding response is ~9.6MB uncompressed, compresses to ~3.5MB
- Egress is now the largest component of Cody Gateway cost
- compressing traffic should make embeddings cheaper / faster for PLG users

This PR only enables compression for `/v1/embeddings` endpoint - other endpoints are primarily used for streaming responses, and compression increases time-to-first-event.

## Test plan

- locally tested with a Sourcegraph instance 
- HTTP Content-encoding negotiation makes this a no-op for clients that don't use compression


cmd/cody-gateway/internal/httpapi/handler.go
```diff
@@ -144,7 +145,12 @@ func NewHandler(
 				otelhttp.WithPublicEndpoint(),
 			),
 		)
-
+		compress, err := httpcompression.DefaultAdapter() // Use the default configuration
+		if err != nil {
+			if err != nil {
+				return nil, errors.Wrap(err, "init compression for Embeddings")
+			}
+		}
```
bobheadxi, line 153:
```suggestion
		if err != nil {
			return nil, errors.Wrap(err, "init compression for Embeddings")
		}
```

bobheadxi (approved):

cmd/cody-gateway/internal/httpapi/handler.go
rafax: Damn, IntelliJ auto-wrapping got me (this looks like an obvious issue here, I didn't see the nested `if` when collapsed). Thank you!

vdavid (approved):
Wow, you make this look so easy. _Just a nine-character change_.

PR #60460, opened 2024-02-13
batches: change criteria for deleting GitHub apps from the database

### LINKED ISSUE #54066: Enhancement: Worker deletes empty GitHub Apps on execution
Enhancement to [PR#52897](https://github.com/sourcegraph/sourcegraph/pull/52897)

When checking & syncing up GitHub App installations, also check if a GitHub App status returns a `404` (from the [Get App endpoint](https://docs.github.com/en/rest/apps/apps?apiVersion=2022-11-28#get-an-app)). Record that the App looks deleted, and surface that as a warning in the admin UI.

**Proposal**: Maybe the better user experience is that rather than deleting an App if we got a 404, we record that it looks deleted, and surface that as a warning from the admin UI...

BolajiOlajide:
Closes #54066

In the past, when we get an error fetching a GitHub app, we assume the app has been deleted and delete all it's installations from the database. This caused issues for one of our customers.

## Test plan

Manual testing. 

I added a GitHub app for Commit Signing, manually triggered the worker to ensure it refreshes it's installations. Then I faked an error in the code, to see if the installations got deleted when a refresh occurred (they didn't). They only got deleted when the client returns a 404.

CleanShot 2024-02-12 at 21 30 54@2x

### DIFF
--- a/internal/github_apps/store/store.go
+++ b/internal/github_apps/store/store.go
@@ -3,6 +3,7 @@ package store
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"net/url"
 	"strings"
 
@@ -16,6 +17,7 @@ import (
 	encryption "github.com/sourcegraph/sourcegraph/internal/encryption"
 	"github.com/sourcegraph/sourcegraph/internal/encryption/keyring"
 	"github.com/sourcegraph/sourcegraph/internal/extsvc"
+	gh "github.com/sourcegraph/sourcegraph/internal/extsvc/github"
 	ghtypes "github.com/sourcegraph/sourcegraph/internal/github_apps/types"
 	itypes "github.com/sourcegraph/sourcegraph/internal/types"
 	"github.com/sourcegraph/sourcegraph/lib/errors"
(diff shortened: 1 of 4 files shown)


internal/github_apps/store/store.go
```diff
@@ -477,9 +477,11 @@ func (s *gitHubAppsStore) SyncInstallations(ctx context.Context, app ghtypes.Git
...
 		errs = errors.Append(errs, err)
 
-		// This likely means the App has been deleted from GitHub, so we should remove all
-		// installations of it from our database, if we have any.
-		if len(dbInstallations) == 0 {
+		// if the error from github is 404 then it means the GitHub app has been deleted. If not, we return
+		// the error and preserve whatever installations we have in the database.
+		// Sample:
+		// {"error": "request to https://ghe.sgdev.org/api/v3/app/installations?page=1 returned status 404: Integration not found", "appName": "BolJHI COMMIT SIgn", "id": 1}
+		if !strings.Contains(err.Error(), "returned status 404: Integration not found") {
```
BolajiOlajide, line 484: I wonder if there's a better way to do this. I can't hardcode the error because the URL might change, for example, if while getting the 3rd page , a GitHub app gets deleted, then the error will look like this:

```
request to https://ghe.sgdev.org/api/v3/app/installations?page=3 returned status 404: Integration not found
```

camdencheek: It looks like we return an [APIError](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@6ad8833d2cd2814b8b3685c10b9bab9f5d7e0524/-/blob/internal/extsvc/github/common.go?L1604-1604), so we should just be able to use `errors.As(` to check the error type and extract the status code

BolajiOlajide: TIL. 

I've updated this check, can you take a look @camdencheek ?

camdencheek (approved):

internal/github_apps/store/store.go
camdencheek: LGTM!

Issue #60439, opened 2024-02-12
Cody Attributions panic


daxmc99:
Several panics were observed in the Attributions code of Cody

```
runtime error: invalid memory address or nil pointer dereference [signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x24d60b1]

at github.com/sourcegraph/sourcegraph/internal/codygateway.(*client).Attribution ( internal/codygateway/client.go:144 )
at github.com/sourcegraph/sourcegraph/cmd/frontend/internal/guardrails/attribution.(*gatewayProxy).SnippetAttribution ( cmd/frontend/internal/guardrails/attribution/attribution.go:85 )
at github.com/sourcegraph/sourcegraph/cmd/frontend/internal/completions.Init.func2.NewAttributionTest.func1 ( cmd/frontend/internal/guardrails/init.go:104 )
at github.com/sourcegraph/sourcegraph/internal/guardrails.(*completionsFilter).runAttribution ( internal/guardrails/attribution_filter.go:178 )
```
https://console.cloud.google.com/errors/detail/CMG3u6v_yZqUtgE;time=P1D?project=src-747bc765eb31a4873e4b

And

```
runtime error: index out of range [43] with length 30

at compress/flate.(*huffmanBitWriter).indexTokens ( GOROOT/src/compress/flate/huffman_bit_writer.go:548 )
at compress/flate.(*huffmanBitWriter).writeBlock ( GOROOT/src/compress/flate/huffman_bit_writer.go:437 )
at compress/flate.(*compressor).writeBlock ( GOROOT/src/compress/flate/deflate.go:170 )
at compress/flate.(*compressor).deflate ( GOROOT/src/compress/flate/deflate.go:408 )
at compress/flate.(*compressor).syncFlush ( GOROOT/src/compress/flate/deflate.go:558 )
at compress/flate.(*Writer).Flush ( GOROOT/src/compress/flate/deflate.go:725 )
at compress/gzip.(*Writer).Flush ( GOROOT/src/compress/gzip/gzip.go:221 )
at github.com/NYTimes/gziphandler.(*GzipResponseWriter).Flush ( external/com_github_nytimes_gziphandler/gzip.go:260 )
at github.com/sourcegraph/sourcegraph/internal/search/streaming/http.(*Writer).EventBytes ( internal/search/streaming/http/writer.go:109 )
[...]

PR #60465, opened 2024-02-13
Quickfix for attribution panics

cbart:
Part of https://github.com/sourcegraph/sourcegraph/issues/60439

- Return error in case `http.Doer` is not set. Not sure how that can be but the error indicates it's nil sometimes.
- Return uninitialized attribution client, if HTTP client was not created.

## Test plan

- [x] Manual check

### DIFF
--- a/cmd/frontend/internal/guardrails/init.go
+++ b/cmd/frontend/internal/guardrails/init.go
@@ -81,6 +81,9 @@ func (e *enterpriseInitialization) Service() attribution.Service {
 	if e.endpoint == "" || e.token == "" {
 		return attribution.Uninitialized{}
 	}
+	if e.client == nil {
+		return attribution.Uninitialized{}
+	}
 	return attribution.NewGatewayProxy(e.observationCtx, e.client)
 }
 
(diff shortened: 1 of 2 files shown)


taras-yemets (approved):

eseliger (commented):
I'm slightly concerned that we don't know why it's nil - can we figure that out before we just plug in some if nil return? :) 

If this is a quick fix so that it can still hit the release, ignore me and we can follow up here.

cbart:
It's here: https://github.com/sourcegraph/sourcegraph/pull/60465/files#diff-0380878f2392758d1876992ca20c56203feaa5e923068ca9855db918dddfbe3fR76

I just don't know exactly what the config condition is that results in uninitialized http doer.

keegancsmith:
> I just don't know exactly what the config condition is that results in uninitialized http doer.

@cbart You are not setting the http client on this line https://github.com/sourcegraph/sourcegraph/blob/bc599bb8f0e40957d5f711e10636bf3ef5c030d0/cmd/frontend/internal/guardrails/init.go#L99

keegancsmith:
Created a PR with a fix https://github.com/sourcegraph/sourcegraph/pull/60465

PR #60471, opened 2024-02-13
guardrails: move httpclient logic into initializer

keegancsmith:
This avoids the footgun of forgetting to set it. It didn't need to be part of the struct since its only ever read in one spot.

This fixes the root cause of the quick fix in https://github.com/sourcegraph/sourcegraph/pull/60465

Test Plan: go test

### DIFF
--- a/cmd/frontend/internal/guardrails/init.go
+++ b/cmd/frontend/internal/guardrails/init.go
@@ -40,11 +40,7 @@ func Init(
 	} else {
 		// On an Enterprise instance endpoint proxies to gateway, and is re-initialized
 		// in case site-config changes.
-		client := httpcli.ExternalDoer
-		if MockHttpClient != nil {
-			client = MockHttpClient
-		}
-		initLogic := &enterpriseInitialization{observationCtx: observationCtx, httpClient: client}
+		initLogic := &enterpriseInitialization{observationCtx: observationCtx}
 		resolver = resolvers.NewGuardrailsResolver(initLogic.Service())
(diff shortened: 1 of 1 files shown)


cmd/frontend/internal/guardrails/init.go
```diff
@@ -76,7 +71,14 @@ func (e *enterpriseInitialization) Service() attribution.Service {
 	if e.endpoint != endpoint || e.token != token {
 		e.endpoint = endpoint
 		e.token = token
-		e.client = codygateway.NewClient(e.httpClient, endpoint, token)
+
+		// We are reachout out of the cluster so we need to use ExternalDoer.
```
keegancsmith, line 75:
```suggestion
		// We are communicate out of the cluster so we need to use ExternalDoer.
```

cmd/frontend/internal/guardrails/init.go
```diff
@@ -76,7 +71,14 @@ func (e *enterpriseInitialization) Service() attribution.Service {
 	if e.endpoint != endpoint || e.token != token {
 		e.endpoint = endpoint
 		e.token = token
-		e.client = codygateway.NewClient(e.httpClient, endpoint, token)
+
+		// We are communicate out of the cluster so we need to use ExternalDoer.
```
keegancsmith, line 75:
```suggestion
		// We communicate out of the cluster so we need to use ExternalDoer.
```

taras-yemets (approved):

PR #60477, opened 2024-02-13
search: better handling of keyword boundaries

stefanhengl:
This fixes a bug in the backend parser to recognise "AND" and "OR" even if they aren't preceded by whitespace. Previously, a query such as "(a or b)and c" was not parsed correctly, because "and" was preceded by ")", which resulted in surprising results.

They key idea of the fix is to recognise parenthesis as proper boundaries and to not advance the parser immediately once we encounter a right parenthesis, but instead to let the caller advance. This way, the caller can construct an operator (`(or a b)` in this case) before continuing with the parsing. This mirrors the behavior of the other keywords that can break the loop in the leaf parser.

The handling of whitespace in our parser is very fragile. A more thorough fix would be to introduce a lexer to eliminate the significance of whitespace from the parser logic. However, we are planning to replace the parser soon. Right now it makes sense to fix this bug within the current framework.

## Test plan
- new units tests
- all existing test pass

### DIFF
--- a/internal/search/query/parser.go
+++ b/internal/search/query/parser.go
@@ -254,8 +254,7 @@ func (p *parser) matchKeyword(keyword keyword) bool {
 
 // matchUnaryKeyword is like match but expects the keyword to be followed by whitespace.
 func (p *parser) matchUnaryKeyword(keyword keyword) bool {
-	if p.pos != 0 && !(isSpace(p.buf[p.pos-1:p.pos]) || p.buf[p.pos-1] == '(') {
-		// "not" must be preceded by a space or ( anywhere except the beginning of the string
+	if p.pos != 0 && !(isSpace(p.buf[p.pos-1:p.pos]) || isRightParen(p.buf[p.pos-1:p.pos]) || isLeftParen(p.buf[p.pos-1:p.pos])) {
 		return false
 	}
 	v, err := p.peek(len(string(keyword)))
@@ -1139,7 +1138,7 @@ func (p *parser) parseAnd() ([]Node, error) {
 	if left == nil {
 		return nil, &ExpectedOperand{Msg: fmt.Sprintf("expected operand at %d", p.pos)}
 	}
-	if !p.expect(AND) {
+	if !(p.expect(AND) || p.expect(NOT)) {
 		return left, nil
 	}
 	right, err := p.parseAnd()
--- a/internal/search/query/parser.go
+++ b/internal/search/query/parser.go
@@ -123,6 +123,16 @@ func isSpace(buf []byte) bool {
 	return unicode.IsSpace(r)
 }
 
+func isRightParen(buf []byte) bool {
+	r, _ := utf8.DecodeRune(buf)
+	return r == ')'
+}
+
+func isLeftParen(buf []byte) bool {
+	r, _ := utf8.DecodeRune(buf)
+	return r == '('
+}
+
 // skipSpace returns the number of whitespace bytes skipped from the beginning of a buffer buf.
 func skipSpace(buf []byte) int {
 	count := 0
@@ -228,7 +238,7 @@ func (p *parser) matchKeyword(keyword keyword) bool {
 	if p.pos == 0 {
 		return false
 	}
-	if !isSpace(p.buf[p.pos-1 : p.pos]) {
+	if !(isSpace(p.buf[p.pos-1:p.pos]) || isRightParen(p.buf[p.pos-1:p.pos])) {
 		return false
 	}
 	v, err := p.peek(len(string(keyword)))
@@ -236,7 +246,7 @@ func (p *parser) matchKeyword(keyword keyword) bool {
 		return false
 	}
 	after := p.pos + len(string(keyword))
-	if after >= len(p.buf) || !isSpace(p.buf[after:after+1]) {
+	if after >= len(p.buf) || !(isSpace(p.buf[after:after+1]) || isLeftParen(p.buf[after:after+1])) {
 		return false
 	}
 	return strings.EqualFold(v, string(keyword))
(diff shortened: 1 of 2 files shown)


internal/search/query/parser_test.go
```diff
@@ -767,6 +767,8 @@ func TestParseParensKeyword(t *testing.T) {
 
 	// parentheses
 	autogold.Expect(`"()"`).Equal(t, test("()"))
+	autogold.Expect(`"()"`).Equal(t, test("(())"))
+	autogold.Expect(`"()"`).Equal(t, test("(     )"))
```
stefanhengl, line 771: I forgot to add these earlier, so I am smuggling them in here to document the behavior.

internal/search/query/parser_test.go
```diff
@@ -785,6 +787,36 @@ func TestParseParensKeyword(t *testing.T) {
...
+	autogold.Expect(`unsupported expression. The combination of parentheses in the query has an unclear meaning. Use "..." to quote patterns that contain parentheses`).Equal(t, test("a )and b"))
+
+	autogold.Expect(`(or "a" "b" "c")`).Equal(t, test("(a or b)or c"))
+	autogold.Expect(`(or "a" "b" "c")`).Equal(t, test("(a or b) or c"))
+	autogold.Expect(`(or (and "a" "b") "c" "d")`).Equal(t, test("(a and b or c) or d"))
+	autogold.Expect(`(or "a" (and "b" "c") "d")`).Equal(t, test("(a or b and c)or d"))
+
+	// first token
+	autogold.Expect(`(and "and" "b")`).Equal(t, test("  and b"))
+	autogold.Expect(`(and "and" "b")`).Equal(t, test("and b"))
```
stefanhengl, line 819: These shouldn't really be autogold tests. Actually, autogold makes these tests harder to work with during debugging and this test doesn't really fit the use case for autogold either. I am planning to convert most if not all of the tests in parser_test.go to simple table-based tests, but I don't want to pollute this PR with the diff.

internal/search/query/parser.go
```diff
@@ -119,8 +119,22 @@ const (
 )
 
 func isSpace(buf []byte) bool {
+	for i := 0; i < len(buf); i++ {
```
jtibshirani, line 122: Could you explain this part? What did we change in terms of handling spaces?

internal/search/query/parser.go
```diff
@@ -1128,7 +1143,7 @@ func (p *parser) parseAnd() ([]Node, error) {
 	if left == nil {
 		return nil, &ExpectedOperand{Msg: fmt.Sprintf("expected operand at %d", p.pos)}
 	}
-	if !p.expect(AND) {
+	if !(p.expect(AND) || p.expect(NOT)) {
```
jtibshirani, line 1146: Could you explain why we check `NOT` here too?

jtibshirani (approved):
Wow this logic is tricky! I don't understand deeply and asked a couple questions.

I don't want to hold this up though, since the tests look great and your overall strategy makes sense.

internal/search/query/parser_test.go
jtibshirani: Nice, looking forward to that change.

Side note: I just learned we parse " and b" as `(and "and" "b")`! hmm!

internal/search/query/parser.go
stefanhengl: Thanks for this comment. It made me reevaluate this change and I now think this is wrong. I believe it fixed some of the issues I saw early on while I was debugging errors around parsing "not", but my other changes have made this unnecessary. I added more test cases for "not" just to be sure.

internal/search/query/parser.go
stefanhengl: All except one call site call this with a single byte. For those call sites this is a NOP change.

In `matchKeyword` we use this to check whether "AND" or "OR" are the first tokens in the query. Before, we just checked whether "AND" or "OR" are at position 0. This means adding space in front of a keyword turned it from a pattern into a keyword, which is odd. 

The change I made just makes the treatment of keywords as first token consistent. However, I think we should rather return a error for this in the future.

internal/search/query/parser_test.go
stefanhengl: > Side note: I just learned we parse " and b" as (and "and" "b")! hmm!

I was surprised too, but at least now it is consistent. This should definitely return an error in the future.

internal/search/query/parser.go
```diff
@@ -223,29 +241,29 @@ func (p *parser) expect(keyword keyword) bool {
...
-// matchKeyword is like match but expects the keyword to be preceded and followed by whitespace.
+// matchKeyword is like match but checks whether the keyword has a valid prefix
+// and suffix.
 func (p *parser) matchKeyword(keyword keyword) bool {
-	if p.pos == 0 {
+	if isSpace(p.buf[:p.pos]) {
 		return false
 	}
-	if !isSpace(p.buf[p.pos-1 : p.pos]) {
+	if !(isSpace(p.buf[p.pos-1:p.pos]) || isRightParen(p.buf[p.pos-1:p.pos])) {
```
keegancsmith, line 250: minor: your usage of isRightParen and isLeftParen always pass in a slice of length 1. Can't you just change these to a comparison? eg `isRightParen(p.buf[p.pos-1:p.pos])` becomes `p.buf[p.pos-1] == ')'`?

internal/search/query/parser.go
```diff
@@ -223,29 +241,29 @@ func (p *parser) expect(keyword keyword) bool {
 	return true
 }
 
-// matchKeyword is like match but expects the keyword to be preceded and followed by whitespace.
+// matchKeyword is like match but checks whether the keyword has a valid prefix
+// and suffix.
 func (p *parser) matchKeyword(keyword keyword) bool {
-	if p.pos == 0 {
```
keegancsmith, line 228: you removed this check and now do things like `p.pos-1`. Is it now possible to go out of bounds?

stefanhengl: No, it's no possible because `isSpace` will return true in that case and matchKeyword returns. Probably a better name for `isSpace` would have been `isFirstToken`.

internal/search/query/parser.go
stefanhengl: Yes, I wanted to mirror the pattern of isSpace to make it more boring to read. Probably overdid a bit here ;-)

PR #60480, opened 2024-02-13
release: skip updating src-cli docs reference

keegancsmith:
I am running the 5.3 release but this is all borked. We used to run a src-cli reference doc generator, but we now have a docs repo which uses mdx files. So the reference generator needs to be updated. So for now I am just skipping this and will follow up later.

Test Plan: progress in running bake-content for release


keegancsmith:
@MaedahBatool we need to follow up on how to get our auto generated references for src-cli.

BolajiOlajide (approved):

PR #60494, opened 2024-02-13
Svelte: port most of the filter panel behavior

camdencheek:
This implements almost all the behavior of the filters panel in the svelte prototype. The only missing behavior I know of is the logic around preserving or resetting filters when the type changes or the query changes. There is obviously still some missing styling, but I think we need to do a more comprehensive color and spacing audit of the prototype, so I skipped most of that.

Apologies for the large PR. I figured this was better than 10 small PRs, and the functionality need to change enough that I needed to do full manual testing anyways.

[Loom walkthrough](https://www.loom.com/share/fcc389aa1dbb44ae81afa2ba7afc7280?sid=7fe2515a-56e8-4ceb-9d0d-59ec9efec65a). Happy to do a live walkthrough as well if that'd be helpful.

## Test plan

So far, manual testing. I would like to implement playwright tests for this as well, but given that I can't currently log in, that's been kinda difficult.


client/web-sveltekit/src/routes/search/SearchResults.svelte
```diff
@@ -109,13 +113,9 @@
...
 <div class="search-results">
-    <DynamicFiltersSidebar
-        size={$sidebarSize}
-        {queryFromURL}
-        {queryFilters}
-        {queryState}
-        streamFilters={$stream?.filters ?? []}
-    />
+    <div style:width={sidebarWidth}>
+        <DynamicFiltersSidebar {selectedFilters} {streamFilters} />
```
camdencheek, line 117: I drastically simplified the sidebar to have only two dependencies: the list of selected filters (from the URL) and the list of filters returned from the backend

client/web-sveltekit/src/lib/LanguageIcon.svelte
```diff
@@ -0,0 +1,14 @@
...
+    import { mdiFileCodeOutline } from '@mdi/js'
+
+    import { getFileIconInfo } from '$lib/wildcard'
+
+    import Icon from './Icon.svelte'
+
+    export let language: string
+    const icon = getFileIconInfo('', language)?.svg
+    const svgPath = icon?.path ?? mdiFileCodeOutline
+    const svgColor = icon?.color
```
fkling, line 11: I think these need to be reactive:
```suggestion
    $: icon = getFileIconInfo('', language)?.svg
    $: svgPath = icon?.path ?? mdiFileCodeOutline
    $: svgColor = icon?.color
```

client/web-sveltekit/src/lib/search/dynamicFilters/CountBadge.svelte
```diff
@@ -0,0 +1,30 @@
+<script lang="ts">
+    import { pluralize } from '$lib/common'
+    import Tooltip from '$lib/Tooltip.svelte'
+    import { Badge } from '$lib/wildcard'
+
+    export let count: number | undefined
+    export let exhaustive: boolean
+
+    function roundCount(count: number): number {
+        const roundNumbers = [10000, 5000, 1000, 500, 100, 50, 10, 5, 1]
```
fkling, line 10: Nit: This could be defined outside, maybe in `<script context="module">`

client/web-sveltekit/src/lib/search/dynamicFilters/HelpFooter.svelte
```diff
@@ -0,0 +1,63 @@
+<script>
+    import LinkShareIcon from '$lib/icons/LinkShare.svelte'
+    import OpenBookIcon from '$lib/icons/OpenBook.svelte'
+</script>
+
+<footer class="help">
```
fkling, line 6: Note: If elements are unique in the template, you don't have to use classes. E.g. you could reference the element directly in the CSS below ( `footer { ... }` ).

client/web-sveltekit/src/lib/search/dynamicFilters/Section.svelte
```diff
@@ -122,20 +100,38 @@
...
+            vertical-align: inherit;
+        }
+    }
+
+    .show-more {
         text-align: center;
     }
 
-    a {
+    li a {
```
fkling, line 134: Doesn't shouldn't be necessary. Did you add as precaution? Or am I not seeing a link somewhere?

client/web-sveltekit/src/lib/search/dynamicFilters/Sidebar.svelte
```diff
@@ -1,151 +1,199 @@
...
-            {/if}
-            {#if filters.utility.length > 0}
-                <Section items={filters.utility} title="Utility" {queryFilters} />
+<aside class="sidebar">
+    <div class="scroll-container">
+        <div class="header">
+            <h3>Filter results</h3>
+            {#if selectedFilters.length > 0}
+                <a href={resetFilters($page.url).toString()}>
+                    <small>Reset all <kbd>{resetModifier} ⌫</kbd></small>
```
fkling, line 82: I assume we don't handle keyboard events yet... should we still show they keyboard shortcut?

fkling (approved):
> I would like to implement playwright tests for this as well, but given that I can't currently log in, that's been kinda difficult.

Why do you need to log in for that? For testing you can simulate a logged in user if that's what you need.

Anyways, looks great overall, I left some comments. Thank you!

client/web-sveltekit/src/lib/search/dynamicFilters/Sidebar.svelte
camdencheek: Okay, I pushed an update that adds the keyboard shortcut on mount. I didn't see any "global shortcut registry" for svelte like I think we have for the react webapp, so I just did it manually for now. It looks like there is already a placeholder issue for "Keyboard shortcuts", so we should take a look at how we want to handle this moving forward

camdencheek:
> Why do you need to log in for that? For testing you can simulate a logged in user if that's what you need.

Oh, yeah, that makes sense. I was thinking I'd be recording a response, but if we're just mocking everything out, then that should just work I guess. I'll take a stab at implementing some tests before merging

client/web-sveltekit/src/lib/search/dynamicFilters/CountBadge.svelte
fkling: Oh actually I meant that the array itself could be defined outside. There is no need to allocate a new array every time the function runs. But maybe engines are smart enough to optimize this? idk It's not a big array either so it probably doesn't matter.

client/web-sveltekit/src/lib/search/dynamicFilters/Sidebar.svelte
```diff
@@ -1,151 +1,210 @@
...
+    $: resetModifier = inferOperatingSystem(navigator.userAgent) === 'MacOS' ? '⌥' : 'Alt'
+    $: resetURL = resetFilters($page.url).toString()
+    $: enableReset = selectedFilters.length > 0
+    $: handleResetKeydown = (event: KeyboardEvent) => {
+        if (enableReset && event.altKey && event.key === 'Backspace') {
+            goto(resetURL)
+        }
+    }
+    onMount(() => window.addEventListener('keydown', handleResetKeydown))
+    onDestroy(() => window.removeEventListener('keydown', handleResetKeydown))
```
fkling, line 84: You can return a function from `onMount` which will be called when the component gets destroyed.

```suggestion
    onMount(() => {
        window.addEventListener('keydown', handleResetKeydown)
        return () => window.removeEventListener('keydown', handleResetKeydown)
    })
```

client/web-sveltekit/src/lib/search/dynamicFilters/Sidebar.svelte
```diff
@@ -1,151 +1,210 @@
...
-    $: filters = groupFilters(streamFilters)
-    $: hasFilters = filters.lang.length > 0 || filters.repo.length > 0 || filters.file.length > 0
+    $: resetModifier = inferOperatingSystem(navigator.userAgent) === 'MacOS' ? '⌥' : 'Alt'
+    $: resetURL = resetFilters($page.url).toString()
+    $: enableReset = selectedFilters.length > 0
+    $: handleResetKeydown = (event: KeyboardEvent) => {
+        if (enableReset && event.altKey && event.key === 'Backspace') {
+            goto(resetURL)
+        }
+    }
```
fkling, line 82: This doesn't have to be reactive. The function will close over `enableReset` and `resetURL` and use whatever values those variables have when the function is called.

```suggestion
    function handleResetKeydown(event: KeyboardEvent) {
        if (enableReset && event.altKey && event.key === 'Backspace') {
            goto(resetURL)
        }
    }
```

client/web-sveltekit/src/lib/search/dynamicFilters/Sidebar.svelte
camdencheek: oh nice, thanks!

camdencheek:
Okay, merging without tests because this is starting to get stale and we don't have any mocking infra for search results yet, so that'll be a larger add than I realized

PR #60498, opened 2024-02-14
Revert "oidc: add prompt login option (#60036)"

unknwon:
This reverts commit 4c5a558d5c4cc234f8d0e4edd448b09e892c6554.

After [some discussions](https://sourcegraph.slack.com/archives/C06HHK23QR4/p1707419775776459), asking users to "sign out twice" may end up with less trouble than "always sign in".

## Test plan

CI


kalanchan (approved):

chenkc805:
Thank you! I'm confident this will lead to a better experience for the ~90% of SG users who only have 1 SG account (note: there's no way to verify this at the moment since email is the user's unique identifier. We don't collect things like phone number, address, or SSN to current verify this)

chenkc805 (approved):

PR #60502, opened 2024-02-14
msp/loadbalancer: add config to enable logging

bobheadxi:
These can be pretty high-volume, so I think it might be best to leave it opt-in for now

## Test plan

click-ops the equivalent change


jac (approved):
Nice!

PR #60503, opened 2024-02-14
Don't verify license on free plan

### LINKED ISSUE #60359: Instance on free tier attempts license checks and fails
https://sourcegraph.slack.com/archives/C05EMJM2SLR/p1707475379851689

The license validity checker makes requests to Sourcegraph to verify the license key. However, since there's no license key, the generated access token is not valid.

This causes logspam and potential customer confusion. Without a key, we should not be sending these requests at all.

We should also triple-check that there's no negative impact from those failing.

pjlast:
Closes #60359 

The free license should not be doing license verification checks.

## Test plan

Add unit test

### DIFF
--- a/internal/licensing/plans.go
+++ b/internal/licensing/plans.go
@@ -58,10 +58,6 @@ func (p Plan) isKnown() bool {
 	return false
 }
 
-func (p Plan) IsFree() bool {
-	return p == PlanFree0 || p == PlanFree1
-}
-
 // Plan is the pricing plan of the license.
 func (info *Info) Plan() Plan {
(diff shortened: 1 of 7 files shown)


internal/licensing/plans.go
```diff
@@ -58,10 +58,6 @@ func (p Plan) isKnown() bool {
 	return false
 }
 
-func (p Plan) IsFree() bool {
-	return p == PlanFree0 || p == PlanFree1
-}
```
eseliger, line 63: oops, that's on me 😬

eseliger (approved):

PR #60506, opened 2024-02-14
msp/alertpolicy: unify on generalized policy builder, with custom condition builders

bobheadxi:
Previously, different conditions would have completely different code paths for generating the policy, which lead to some drift in the ones we pay less attention to (like response code) and a lot of copy-pasta.

This change adjusts our internal abstraction to center around condition builders, and switch on the type of the alert to generate the condition that then goes into the a unified alert policy format. This will make it easier for us to build more alerts on different condition types.

In doing this I notice that `ResourceKind`, `ResourceName` are threshold-aggregation-specific, so I've moved those downwards as well.

## Test plan

`sg msp generate -all` changes only the response code alerts to have additional labels, and also updates the severity level for the external uptime alert to match the config (previous oversight)

### DIFF
--- a/dev/managedservicesplatform/internal/resource/alertpolicy/BUILD.bazel
+++ b/dev/managedservicesplatform/internal/resource/alertpolicy/BUILD.bazel
@@ -3,7 +3,11 @@ load("@io_bazel_rules_go//go:def.bzl", "go_library")
 
 go_library(
     name = "alertpolicy",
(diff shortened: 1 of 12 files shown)


jac (approved):
Nice 🚀 
There was definitely a bit of spaghetti code and this really clears it up making the logic much easier to follow

Issue #59297, opened 2024-01-03
OpenAI tokenizer


cbart:
Add support for OpenAI token consumption tracking for completions api calls.  This issue is add token counting to the Cody Gateway.

chwarwick:
Request has been made to OpenAI to enable usage data for streaming requests.

PR #60507, opened 2024-02-14
Only log when none of the messages contained usage

rafax:
Right now, we log for every message that doesn't contain usage ([logs](https://sifted.eu/articles/pitch-deck-mistral)), but we only want to log when none of the messages for a single request contain usage.

## Test plan

- tested locally


rafax:
Related to https://github.com/sourcegraph/sourcegraph/issues/59297

chwarwick (approved):

PR #60508, opened 2024-02-14
Set returnTo when redirecting to a third-party auth provider on dotcom

vdavid:
- Fixes https://github.com/sourcegraph/accounts.sourcegraph.com/issues/520

Add the `returnTo` URL param if we have a return URL.

## Test plan

- Tested the logic locally with several test URLs with their query string empty and not.
- Couldn't test it end-to-end because I have issues with my local env I haven't resolved.


unknwon (approved):

chrsmith (approved):

chrsmith:
Seems low risk, so I went ahead and merged it @vdavid

PR #60511, opened 2024-02-14
sg/msp: make nobl9 key optional in spec

jac:
Change Nobl9 to a pointer type to make it optional
## Test plan

- locally tested


bobheadxi (approved):
Thanks!

PR #60512, opened 2024-02-14
msp/monitoring: add MetricAbsence condition type, add missing executions alert

bobheadxi:
Prevents the issue we ran into in https://sourcegraph.slack.com/archives/C05GJPTSZCZ/p1707907271464879 from going unnoticed again:

1. Adds a `MetricAbsence` condition type for us to use internally
  a. In doing so, I realized we can shape our abstraction around the GCP UI with the new `ConditionBuilder` type. This is basically what you see in the UI when you create a policy by hand. This is now shared between `MetricAbsence` and `ThresholdAggregation`
2. Adds `FindMaxCronInterval()` on the job schedule spec, which validates cron specs, tries to ensure they are of a certain window, and provides the "maximum interval" it detects.
3. Use `FindMaxCronInterval()` to define an alert where if no executions are detected within that interval + 10 minutes, the alert will fire.

## Test plan

`sg msp generate -all` has no diff except for gatekeeper (our only Job). I put gatekeeper into CLI-apply mode and applied the alert, which looks right:

image

### DIFF
--- a/dev/managedservicesplatform/internal/resource/alertpolicy/alertpolicy.go
+++ b/dev/managedservicesplatform/internal/resource/alertpolicy/alertpolicy.go
@@ -2,6 +2,8 @@ package alertpolicy
 
 import (
 	"fmt"
+	"reflect"
+	"strings"
 
 	"github.com/sourcegraph/sourcegraph/lib/errors"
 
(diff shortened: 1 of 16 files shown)


dev/managedservicesplatform/spec/environment.go
```diff
@@ -561,13 +563,82 @@ type EnvironmentJobSpec struct {
...
+	var errs []error
+	if _, err := s.FindMaxCronInterval(); err != nil {
+		errs = append(errs, errors.Wrap(err, "schedule.cron: invalid schedule"))
+	}
+	return errs
+}
+
+// FindMaxCronInterval tries to find the largest gap between events in the cron
+// schedule. It may return 'nil, nil' if no configuration is available.
+func (s *EnvironmentJobScheduleSpec) FindMaxCronInterval() (*time.Duration, error) {
```
bobheadxi, line 599: @rafax this "detects" the "largest" interval in a corn schedule for the alert we discussed

dev/managedservicesplatform/internal/resource/alertpolicy/alertpolicy.go
```diff
@@ -272,3 +246,21 @@ func New(scope constructs.Construct, id resourceid.ID, config *Config) (*Output,
...
+		t := fmt.Sprintf("%T", o)
+		types = append(types, t)
+		// we must reflect as []any boxes the pointer
+		if !reflect.ValueOf(o).IsNil() {
+			nonNil = append(nonNil, t)
+		}
+	}
+	if len(nonNil) != 1 {
+		return errors.Newf("exactly one of [ %s ] must be specified, found: [ %s ]",
+			strings.Join(types, ", "), strings.Join(nonNil, ", "))
```
bobheadxi, line 263: example:

```
❌ failed to create monitoring stack: failed to create common alerts: cpu: exactly one of [ *alertpolicy.ThresholdAggregation, *alertpolicy.ResponseCodeMetric, *alertpolicy.MetricAbsence ] must be specified, found: [ *alertpolicy.ThresholdAggregation, *alertpolicy.ResponseCodeMetric, *alertpolicy.MetricAbsence ]
```

dev/managedservicesplatform/spec/environment.go
rafax: Sounds good to me.

jac (approved):
Ran locally and everything looks good 👍

unknwon (approved):
PR description LGTM 😁

PR #60525, opened 2024-02-14
search: replace autogold with table based test

stefanhengl:
This is the first PR in a series to replace autogold with simpler table based tests where it makes sense.

Why?

Autogold is great for regression tests or when the expected result is tedious to write out, but it is more difficult to parse visually and it is not as convenient during debugging.

Test plan:
all tests pass

### DIFF
--- a/internal/search/query/parser_test.go
+++ b/internal/search/query/parser_test.go
@@ -714,7 +714,7 @@ func TestMatchUnaryKeyword(t *testing.T) {
 	autogold.Expect("true").Equal(t, test("(not bar)", 1))
 }
 
-func TestParseParensKeyword(t *testing.T) {
+func TestParseSearchTypeKeyword(t *testing.T) {
 	test := func(input string) string {
 		plan, err := Pipeline(
 			Init(input, SearchTypeKeyword),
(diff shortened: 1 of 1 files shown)


internal/search/query/parser_test.go
```diff
@@ -726,101 +726,115 @@ func TestParseParensKeyword(t *testing.T) {
...
+		{input: `(a and b or c) or d`, want: `(or (and "a" "b") "c" "d")`},
+		{input: `(a or b and c)or d`, want: `(or "a" (and "b" "c") "d")`},
+
+		// first token
+		{input: `  and b`, want: `(and "and" "b")`},
+		{input: `and b`, want: `(and "and" "b")`},
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.input, func(t *testing.T) {
```
keegancsmith, line 831: minor, these test case names will be pretty poor. my personal preference which you are welcome to ignore is to just not use subtests. I would just include the input in t.Errorf.

keegancsmith (approved):

internal/search/query/parser_test.go
stefanhengl: :-) You won't like the answer to that. I want subtests because Goland supports running them individually which is very nice for debugging. 

Test names could be prettier, but I think it is still readable enough, no?

Screenshot 2024-02-15 at 08 50 32

internal/search/query/parser_test.go
```diff
@@ -726,101 +726,115 @@ func TestParseParensKeyword(t *testing.T) {
...
-	// first token
-	autogold.Expect(`(and "and" "b")`).Equal(t, test("  and b"))
-	autogold.Expect(`(and "and" "b")`).Equal(t, test("and b"))
+	testcases := []struct {
+		input string
+		want  string
+	}{
+		// parens as grouping
+		{input: `foo and bar and bas`, want: `(and "foo" "bar" "bas")`},
+		{input: `(foo and bar) and bas`, want: `(and "foo" "bar" "bas")`},
```
bobheadxi, line 735: @stefanhengl Might not be what you're looking for but FYI, you can use table driven tests _and_ have `autogold` by using `autogold.Value` in `want`, like this: https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/internal/audit/audit_test.go?L25-36

internal/search/query/parser_test.go
keegancsmith: Do you use your mouse to click on that button?

keegancsmith: TBH the names are better than I expected. I thought go would butcher the special chars.

PR #60535, opened 2024-02-15
Revert #59723 (login with SAMS-dev)

chrsmith:
So this PR just reverts what was added in #59723 😞 

A while back when trying to figure out how to enable testing the integration of SAMS, dotcom, and the SSC backend, there was a solid week where it seemed like supporting both SAMS-dev and SAMS-prod in Sourcegraph.com was the way to go...

However, that time has since passed. And now we just have a dedicated Sourcegraph instance running in dotcom mode for that type of testing. And so we can remove the escape hatch to run the SAMS-dev identity provider.

To complete this cleanup however, we'll also need to (1) remove SAMS-dev identity provider from Sourcegraph.com, as well as (2) remove the SAMS-dev client secret from the SITE_JSON configuration data.

I'll take care of removing the SAMS-dev identity provider from Sourcegraph.com before merging.

## Test plan

Um, I removed all the relevant code? The code that runs quickest is that which doesn't run at all?


unknwon (approved):
It has served us well

chrsmith:
Note to self: When ready, remove the SAMS-dev IdP config by following the steps here:
https://handbook.sourcegraph.com/departments/engineering/teams/devops/update_sg_website_config/#sensitive-configurations

chrsmith:
Navigating to https://sourcegraph.com/sign-in?sourcegraph-accounts-dev=true just redirects to SAMS-prod now. So I'll just go ahead and merge this.

PR #60536, opened 2024-02-15
Allow Bedrock completions provider to accept a full endpoint url

chwarwick:
This allows a full endpoint to provided to the completions configuration instead of a aws region.  The reason for this change is to support scenarios where the default bedrock endpoints are not reachable from the frontend deployment.  When specifying the endpoint it is necessary to set the AWS_REGION environment variable to the region that the Bedrock models are in.
## Test plan
These commits manually were tested in a reference environment with a tunnel between GCP and AWS that approximated the type of environment this is intended to support


arafatkatze (approved):
🚢

eseliger (approved):
A test case to illustrate the exact behavior here would be nice, to make sure we don't accidentally regress here. It took me a bit to (I think) understand what this does.
Approve to unblock

PR #60544, opened 2024-02-15
search: refactor matchKeyword

stefanhengl:
This is a follow-up refactor based on the post merge comments in #60477. 

- `isSpace` now returns false for empty buffers, which makes it a generalisation of the previous implementation. This means we have to check for `p.pos == 0` again, which makes the code more explicit.
- replace the helper functions `isLeftParen`, `isRightParent` with simple rune comparisons.  

## Test plan
All tests pass

### DIFF
--- a/internal/search/query/parser.go
+++ b/internal/search/query/parser.go
@@ -134,16 +134,6 @@ func isSpace(buf []byte) bool {
 	return true
 }
 
-func isLeftParen(buf []byte) bool {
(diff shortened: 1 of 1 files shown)


keegancsmith (approved):
is there a test case which would of caught us not checking for pos == 0?

stefanhengl:
> is there a test case which would of caught us not checking for pos == 0?

yup, apart from some edge cases, all parser tests fail if we don't check properly.

PR #60546, opened 2024-02-15
telemetry-gateway: fast-error on oversized event, improve logging

bobheadxi:
Telemetry Gateway recently received an event that, when composed as JSON with license key and everything, seems to be exceeding 10MB in size, which is the Pub/Sub limit: https://sourcegraph.slack.com/archives/C06CCJR4K9R/p1707986358501489?thread_ts=1707984656.886109&cid=C06CCJR4K9R

Pending a decision from the team (https://sourcegraph.slack.com/archives/CN4FC7XT4/p1707986514302069), let's record some details here for now:

1. Return special annotate error with additional details on clearly oversized events, and don't both trying to send it to pub/sub, as that may cause problems if this continues
2. Improve our logger such that all events in a stream are annotated with the instance ID

Note the TODO:

```go
// TODO: We can't error forever, as the instance will keep trying to
// deliver this event - eventually we may just need to take an action,
// then pretend the event succeeded.
```

## Test plan

CI

### DIFF
--- a/cmd/telemetry-gateway/internal/events/BUILD.bazel
+++ b/cmd/telemetry-gateway/internal/events/BUILD.bazel
@@ -13,7 +13,9 @@ go_library(
         "//internal/pubsub",
         "//internal/telemetrygateway/v1:telemetrygateway",
         "//lib/errors",
+        "@com_github_cockroachdb_redact//:redact",
         "@com_github_sourcegraph_conc//pool",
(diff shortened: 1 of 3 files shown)


bobheadxi:
Hoping to get an expedited merge here to collect some more data on the problematic events

jac (approved):

PR #60547, opened 2024-02-15
Detect patterns in prompt

rafax:
Quick PR to give us some insight into patterns we want to detect. 

Will follow-up with refactoring / per-phrase counting / using this in flagging / generalising flagging to all providers and features.

## Test plan

- tested locally
- E2E tests pass


cmd/cody-gateway/internal/httpapi/completions/anthropic.go
```diff
@@ -135,6 +135,7 @@ func NewAnthropicHandler(
 		// user.
 		2, // seconds
 		autoFlushStreamingResponses,
+		config.DetectedPromptPatterns,
```
bobheadxi, line 138: (is this used? I can't seem to find any references to it)

bobheadxi (approved):

cmd/cody-gateway/internal/httpapi/completions/anthropic.go
rafax: It's used here https://github.com/sourcegraph/sourcegraph/pull/60547/files#diff-d33acbb01d160f4bc5b97ddd44ee7f3dd470d61e4da4914d783aaa9241fc87feR260

rafax: I want to refactor the whole HttpHandler -> PerProviderHandler -> abstract UpstreamHandler config and state passing not to drop a bunch of random parameters, but that's a bigger refactor, so I will do it after the launch tomorrow.

bobheadxi: Ah, missed it - thank you!

PR #60549, opened 2024-02-15
ci: generate different steps depending on what pipeline we're in

burmudar:
This change mostly affects the Aspect Pipeline, moving the initial steps being defined in the UI to having them in the `buildkite.yml` file.

Once this is merged the UI steps will be updated to prefer uploading the pipeline rather than raw steps
## Test plan
- [sourcegraph pipeline](https://buildkite.com/sourcegraph/sourcegraph/builds/262493)
- [aspect pipeline](https://buildkite.com/sourcegraph/aspect-experimental/builds/8813) on this branch - notice that the pipeline is uploaded
- [aspect pipeline](https://buildkite.com/sourcegraph/aspect-experimental/builds/8814) not on this branch - notice different steps being used

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -25,7 +25,7 @@ steps:
         label: ":oncoming_police_car: commit check"
         agents: { queue: aspect-small }
         command: |
-          echo "--- :git: check we're after commit
+          echo "--- :git: check we're after commit"
           if git merge-base --is-ancestor 20a3c0836eaa300beecf88e21f78dcd708b047d3 HEAD; then
             echo "✅ We are after the commit."
           else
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -17,3 +18,34 @@ steps:
...
+      - key: aspect-workflows-upload
+        depends_on: "after-commit-check"
+        label: ":aspect: Setup Aspect Workflows"
+        commands:
+          - "rosetta steps | buildkite-agent pipeline upload"
+        agents:
+          queue: aspect-small
+      - label: ":pipeline: Generate pipeline"
+        depends_on: "after-commit-check"
+        if: build.branch !~ /^main(-dry-run\/)?.*/
```
Strum355, line 45: Do we still need dry-run stuff? I thought we got rid of that

burmudar: Yeah we do. Not as frequent as people use to but it's still used

Strum355 (approved):

PR #60550, opened 2024-02-15
database: return not found error for repos.GetBy{ID,Name}

keegancsmith:
I was looking at some logs and noticed the gitserver janitor job constantly logging that it failed to lookup a repo. I checked and it correctly said these repos no longer existed, but this was logging when things go wrong. Turns out in a refactor from August 2022 (cd1131217d73) we wrapped the error type in scanGitserverRepos, but didn't update the call sites of it to take into account wrapped errors.

This means since then lookups of missing repos via the normal GetBy methods have not been returning missing repo errors! This likely has a bunch of ramifications that I haven't explored yet.

In the case of gitserver janitor, I believe we would never delete missing repositories now. There is a decent chance this will free up a bunch of disk space for all sourcegraph instances. Atleast for the logs I saw on s2, this is a big chunk of repos.

Minor change: I adjusted the janitor log which was reporting this to actually report the repo name instead of the directory.

Test Plan: updated the unit tests to assert on missing repo errors.

### DIFF
--- a/cmd/gitserver/internal/cleanup.go
+++ b/cmd/gitserver/internal/cleanup.go
@@ -358,7 +358,8 @@ func cleanupRepos(
 			return false, nil
 		}
 
-		_, err := db.GitserverRepos().GetByName(ctx, gitserverfs.RepoNameFromDir(reposDir, dir))
+		repoName := gitserverfs.RepoNameFromDir(reposDir, dir)
+		_, err := db.GitserverRepos().GetByName(ctx, repoName)
 		// Repo still exists, nothing to do.
 		if err == nil {
 			return false, nil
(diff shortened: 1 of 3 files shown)


eseliger (approved):
Nice find, thank you!

keegancsmith:
Backport for the next patch release? Or rather just wait?

eseliger:
This code path is off by default and only in testing for S2 right now, so I think we should be fine here, WDYT?

keegancsmith:
oh I didn't realise it was only for s2! yeah sounds good.

eseliger:
It should eventually be used everywhere but we didn't have enough confidence in it yet 😬

PR #60554, opened 2024-02-15
Reduce info logspam from logging in worker

eseliger:
This seems quite noisy and clearly like a debug signal, so moving it to debug level instead.

```
[         worker] INFO worker.repo-perms-syncer workerutil/worker.go:329 Dequeued record for processing {"TraceId": "fdcacc27c3b381f2746ff76e8573eaed", "SpanId": "b98a6da2a9121465", "name": "user_permissions_sync_job_worker", "id": "208561"}
```

## Test plan

No noise in my terminal anymore.


pjlast (approved):
ez stamp

PR #60564, opened 2024-02-15
search: include query in feedback prompt

stefanhengl:
This is limited to .com

This adds the user's query as initial value to the feedback prompt. The query helps us to put the feedback into context. Users can simply delete the query if they don't want to send it to us.

Screenshot 2024-02-22 at 11 35 30

Test plan:
manual testing


client/web/src/search/results/components/search-results-info-bar/SearchResultsInfoBar.tsx
```diff
@@ -241,6 +241,7 @@ export const SearchResultsInfoBar: FC<SearchResultsInfoBarProps> = props => {
                             : null
                     }
                     onClose={() => setFeedbackModalOpen(false)}
+                    initialValue={query && '(' + patternType + ') ' + query}
```
jtibshirani, line 244: Tiny suggestion: it'd be nice if we printed a template like
```
<Feedback here>

Query: ... auto-fill query
```

jtibshirani (approved):
Sorry for the slow review here! I think this is totally fine for dot com, but I am not sure if we can be encouraging users to send us their queries. Before we ship this, let's double-check with our data team whether this is okay.

stefanhengl:
> Before we ship this, let's double-check with our data team whether this is okay.

I checked with our data team. TLDR: It's complicated for enterprise customers, .com is fine. I updated the PR.

PR #60567, opened 2024-02-15
svelte: Change GraphQL library and add repo pages error handling

fkling:
I'm truly sorry for this monster PR. How did it come to that? It all started out by wanting to fix the following scenario:

- User navigates to file/folder in non-existing folder
- Sidebar is "stuck" in loading state
- User navigates to existing folder (e.g. repo root)
- Sidebar is still stuck in loading state

This happens in both the React and SvelteKit app.

To fix this in the SK app I refactored the logic for managing the file tree sidebar state into it's own store, that would return an Error when data loading failed. That in turn required me to consolidate the data loading in a single place (no more loading the initial set of data in the data loader), which actually also solved another problem.
BUT fetching the file data only once the file tree provider is created is suboptimal. So I thought I could at least fire of the same GraphQL query in the loader so that it will be already in the cache when the component fetches the data.

That's where things started to fall apart.

First, for whatever reason the executing the query in the component would result in a second network request for exactly the same data (i.e. "preloading" didn't work).
Secondly this somehow caused the observable query for the history panel to fire twice. I'm also don't understand why, but I've seen bug reports for Apollo that mention that observable queries are unexpectedly re-executed when other queries are triggered.

I've run into other problems with Apollo before, mainly related to the cache not being properly configured or `id` fields missing from queries. These issues have never been obvious (the network tab shows data is returned from the server but the UI component receives an empty list or `undefined`).

So finally I started looking into other GraphQL libraries and [urql](https://github.com/urql-graphql/urql) seems promising: It's lightweight and has a simpler caching model by default, with the option to upgrade to normalized caching, like Apollo, later on.

Switching the GraphQL library meant I needed to update all GraphQL callsites (maybe we should introduce a wrapper to abstract? not sure)

Once that was done I modified data loaders, layouts and pages to handle errors better, and added unit and playwrite tests to ensure it works as expected.

While the infinity scrolling handling is a bit gnarly, I think overall it's easier to follow the data flow. Where possible I tried to use `{#await ...}` and handle pending, loaded and error state directly in the template. Where this isn't possible I'm converting the promise to a store with `createPromiseStore` which now has a simpler API as well.

Most notable changes:

- Switch to urql library. In it's current form it's a simpler implementation because it uses _document caching_ instead of _normalized caching_. That means that individual requests/responses are cached, there is no "shared cache" that queries can draw from. However so far most of our queries are quite "disjoint" so at least for now the reduced complexity (no need to define interface->type mappings, key fields, cache normalization) seems to be worth the tradeoff.
- Infinity scrolling response consolidation is now handled by a helper function/store and is not in background in the cache. The new API is quite verbose but I didn't want to spent too much time optimizing for this use case yet. It's already more complex than I prefer, but I wanted it to properly handle the case of restoring the right scroll position when navigating back to it, without triggering new network requests.
- Added simple version of `Alert` component to show error and info messages.
- Added error handling to repository pages. When data loading fails for any reason or when a resource is not available we surface that information accordingly. Note that it doesn't look great UI wise yet! I'm just adding `<Alert>`s where needed. My main goal was to not silently ignore errors anymore.
- Fixed various issues that I stumbled upon while working on this (for example an issue with GraphQL mocking)

Example for accessing a non-existing directory:

## Test plan

Manual testing, `pnpm test`, `pnpm vitest`


client/web-sveltekit/src/app.html
```diff
@@ -13,6 +13,6 @@
         %sveltekit.head%
     </head>
     <body data-sveltekit-preload-data data-sveltekit-preload-code="hover">
-        %sveltekit.body%
+        <div style="display: contents">%sveltekit.body%</div>
```
fkling, line 16: `pnpm check` suggested to do this.

client/web-sveltekit/src/lib/Paginator.ts
```diff
@@ -1,7 +1,7 @@
 export enum Param {
-    before = '$before',
-    after = '$after',
-    last = '$last',
+    before = '_before',
+    after = '_after',
+    last = '_last',
```
fkling, line 4: The `$` is encoded in URLs which makes testing for expected URLs more annoying. I don't think it matters what we use here. We can't use `after` directly though because that's already used for the time period (by the current web app).

client/web-sveltekit/src/lib/TreeNode.svelte
```diff
@@ -86,7 +88,7 @@
         margin: 0.25rem 0;
         border-radius: var(--border-radius);
 
-        &[aria-expanded='true'][tabindex='0']:focus {
+        &[tabindex='0']:focus {
```
fkling, line 91: This had the effect that the focus ring would only be shown for open folders, which made it look like that keyboard navigation doesn't work for the file tree. No idea how that got in.

client/web-sveltekit/src/lib/graphql/apollo.ts
fkling, line 1: We still need to use an Apollo client to work with APIs from web and other packages (temporary settings specifically). Long term plan is to phase that out of course.

client/web-sveltekit/src/lib/stores.ts
```diff
@@ -38,13 +35,6 @@ export const settings = {
     },
 }
 
-export const graphqlClient = {
-    subscribe(subscriber: (client: GraphQLClient) => void) {
-        const { client } = getStores()
-        return client.subscribe(subscriber)
-    },
-}
-
```
fkling, line 47: This wasn't used anymore anywhere.

client/web-sveltekit/src/routes/+layout.svelte
```diff
@@ -54,35 +52,6 @@
...
-        const start = Date.now()
-        requestAnimationFrame(function scroll() {
-            if (main) {
-                main.scrollTo(0, y)
-            }
-            if ((!main || main.scrollTop !== y) && Date.now() - start < 3000) {
-                requestAnimationFrame(scroll)
-            }
-        })
-    }
```
fkling, line 85: This isn't needed anymore since we made the whole body scrollable.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.ts
```diff
@@ -1,63 +1,82 @@
...
-    const topTreePath: Record<string, string> = {}
-
-    getRootPath = (repo: string, path: string) => {
-        const treePath = topTreePath[repo]
-        if (treePath && (treePath === REPO_ROOT || path.startsWith(treePath))) {
-            return topTreePath[repo]
-        }
-        return (topTreePath[repo] = path)
-    }
-}
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.ts:29]: This actually caused issues with pre-loading and browser history navigation. When one hovered over a link that would preload information for a "higher level" directory, but then clicked the browser back button the page would load the new file tree.
Now the "top level folder" state is tracked in the component itself and is not affected anymore by data preloading.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/layout.gql
```diff
@@ -4,12 +4,6 @@ query GitHistoryQuery($repoName: String!, $revspec: String!, $first: Int, $after
         commit(rev: $revspec) {
             id
             ancestors(first: $first, path: $filePath, afterCursor: $afterCursor) {
-                # This is a bit hacky, but by fetching all the data needed by both
-                # the history panel and the commits page we ensure that our custom
-                # Apollo infinitiy scroll cache handling for this field works
-                # correctly. Eventually we should revsisit the use of infinity scroll
-                # on the commits page.
-                ...CommitsPage_GitCommitConnection
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/layout.gql:12]: This is not needed anymore because the infinity scrolling logic is not shared anymore.

client/web-sveltekit/src/lib/TreeView.svelte
```diff
@@ -26,8 +26,22 @@
 
     export let treeProvider: TreeProvider<N>
 
+    /**
+     * Scroll the selected item into view.
+     * If the selected item cannot be found it will try again up to 5 times.
```
camdencheek, line 31: Why would this be nondeterministic? (that is, why would retrying help?)

camdencheek (approved):
I did not read every line, but the motivation of the change sounds good to me. Thanks!

client/web-sveltekit/src/lib/TreeView.svelte
fkling: I guess I should add that: It's possible that selected item is in a subfolder that wasn't loaded yet, so there will be a series of requests to open the subfolder first.
I don't think it happens often since we normally only show the contents of the parent folder, but I saw this happening sometimes, I believe when navigating back and forth and having closed previous folders.

fkling: I'm going to remove that change since it's not really related to the rest and will think about a more reliable approach.

PR #60582, opened 2024-02-16
backcompat: make step non-async

burmudar:
Missed a series of failures on https://buildkite.com/sourcegraph/sourcegraph-backcompat/builds/23295
## Test plan
Primary build should wait for this step to complete


Strum355:
Why was this async in the first place? Does the reason still hold?

burmudar:
> Why was this async in the first place? Does the reason still hold?

It is a broken assumption that the build would still fail if a child build fails also that the build would run concurrently while other steps complete ... which happens in any case with steps ...

eseliger (approved):

PR #60596, opened 2024-02-17
Svelte: add display limit and content length limit

camdencheek:
This adds a display limit (currently unlimited) and a line length limit (currently unlimited) to the search requests from the Svelte webapp. Currently, we're trying to stream the entire result set back to the client, which can be quite large. Additionally, rendering and highlighting long lines can be expensive.

## Test plan

Clicked around, testing that the webapp no longer seized up on large result sets.

### DIFF
--- a/client/web-sveltekit/src/routes/search/+page.ts
+++ b/client/web-sveltekit/src/routes/search/+page.ts
@@ -126,6 +126,8 @@ export const load: PageLoad = ({ url, depends }) => {
             featureOverrides: [],
             chunkMatches: true,
             searchMode,
+            displayLimit: 500,
+            maxLineLen: 1000,
         }
 
         // We create a new stream only if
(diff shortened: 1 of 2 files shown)


fkling (approved):
Thank you!

fkling:
Would we need a display limit if we truly virtualize rendering results? (I.e. not render results that are outside the viewport, like CodeMirror does with lines). Or would we still need a display limit for other reasons?

camdencheek:
We still want a display limit just to limit the volume of data we send to the client. Even with a display limit, the stream can still hit 100s of MBs. 

In theory, we might be able to stop streaming the contents after some limit and just stream the ranges so we fetch the contents lazily after N results, but that's getting complex and I honestly don't think there's all that much value to more results past the first few hundred.

PR #60609, opened 2024-02-19
sg: bazelisk check skipped on nix develop

keegancsmith:
If using nix develop we can defer to our nix environment correctly installing bazel. In particular on linux nix directly installs the correct version of nix and we do not use bazelisk.

Test Plan: "sg start enterprise-bazel" on nixos

### DIFF
--- a/dev/sg/internal/check/helpers.go
+++ b/dev/sg/internal/check/helpers.go
@@ -2,6 +2,7 @@ package check
 
 import (
 	"context"
+	"os"
 	"time"
 
 	"github.com/sourcegraph/sourcegraph/lib/errors"
@@ -52,4 +53,16 @@ func WrapErrMessage(check CheckFunc, message string) CheckFunc {
 		}
 		return nil
 	}
+}
+
+// SkipOnNix will not run check if running inside of our nix develop
+// environment. reason is not read, but is used to document why at the
+// callsite.
+func SkipOnNix(reason string, check CheckFunc) CheckFunc {
+	if os.Getenv("SRC_NIX_DEVELOP") == "1" {
+		return func(ctx context.Context) error {
+			return nil
+		}
+	}
+	return check
 }
--- a/dev/sg/internal/check/helpers.go
+++ b/dev/sg/internal/check/helpers.go
@@ -59,7 +59,7 @@ func WrapErrMessage(check CheckFunc, message string) CheckFunc {
 // environment. reason is not read, but is used to document why at the
 // callsite.
 func SkipOnNix(reason string, check CheckFunc) CheckFunc {
-	if os.Getenv("SRC_NIX_DEVELOP") == "1" {
+	if os.Getenv("IN_NIX_SHELL") != "" && os.Getenv("name") == "sourcegraph-dev-env" {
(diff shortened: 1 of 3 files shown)


keegancsmith:
@Strum355 btw does "sg start enterprise-bazel" work for you on NixOS? Currently I am getting a coredumping JVM :O

Strum355:
I thought we had the following specifically for this :thinking: https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/shell.nix?L117

> @Strum355 btw does "sg start enterprise-bazel" work for you on NixOS? Currently I am getting a coredumping JVM :O

Will test :loading:

Strum355:
> @Strum355 btw does "sg start enterprise-bazel" work for you on NixOS? Currently I am getting a coredumping JVM :O

Working fine for me here

dev/sg/internal/check/helpers.go
```diff
@@ -53,3 +54,15 @@ func WrapErrMessage(check CheckFunc, message string) CheckFunc {
 		return nil
 	}
 }
+
+// SkipOnNix will not run check if running inside of our nix develop
+// environment. reason is not read, but is used to document why at the
+// callsite.
+func SkipOnNix(reason string, check CheckFunc) CheckFunc {
+	if os.Getenv("SRC_NIX_DEVELOP") == "1" {
```
Strum355, line 62: We can check for `IN_NIX_SHELL` and/or `name` variable
```
$ env | grep -E 'sourcegraph-dev-env|IN_NIX_SHELL'
IN_NIX_SHELL=impure
name=sourcegraph-dev-env
```

keegancsmith: cool, I'll use that instead.

keegancsmith:
> I thought we had the following specifically for this 🤔 https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/shell.nix?L117

the check is directly running bazel instead of bazelisk

Strum355:
> > I thought we had the following specifically for this 🤔 [sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/shell.nix?L117](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/shell.nix?L117)
> 
> the check is directly running bazel instead of bazelisk

realized that 😦 not great but also cant compromise the check for others just for our sake lol

keegancsmith:
> > the check is directly running bazel instead of bazelisk
> 
> realized that 😦 not great but also cant compromise the check for others just for our sake lol

Are we compromising though? We are just opting out of confirming bazel is bazelisk in nix environments?

Strum355:
> > > the check is directly running bazel instead of bazelisk
> > 
> > 
> > realized that 😦 not great but also cant compromise the check for others just for our sake lol
> 
> Are we compromising though? We are just opting out of confirming bazel is bazelisk in nix environments?

I meant if we were to have a different solution than the one here, which wouldn't require nix-specific checks but would be less comprehensive for everyone else 🙂 sorry for the confusion!

Strum355 (approved):

PR #60610, opened 2024-02-19
executors: fix AMI push script for when aws_regions.json missing

Strum355:
Causing error to not cause the script to fail if the file was missing

## Test plan

https://stackoverflow.com/questions/44014857/function-invocation-in-for-loop-does-not-fail-even-with-set-e


eseliger (approved):
stamp, not a bash mage

PR #60611, opened 2024-02-19
scripts: dont run subshell in for loop

Strum355:
For same reason as https://github.com/sourcegraph/sourcegraph/pull/60610, cleaning up the remaining scripts

## Test plan

https://stackoverflow.com/questions/44014857/function-invocation-in-for-loop-does-not-fail-even-with-set-e


fkling (approved):

Strum355:
For same reason as https://github.com/sourcegraph/sourcegraph/pull/60610, cleaning up the remaining scripts

## Test plan

https://stackoverflow.com/questions/44014857/function-invocation-in-for-loop-does-not-fail-even-with-set-e

PR #60620, opened 2024-02-19
Re-add sign-in redirect with the redirectUrl as absolute URL

vdavid:
## Test plan

@eseliger tests it locally


eseliger (approved):
stamping, talked on slack that this seems to work, but we want to follow up on making sure it keeps working with absolute URLs as well.

PR #60621, opened 2024-02-19
sg: Support only flag in sg ops update-images

eseliger:
This adds a new flag that allows us to run `sg ops update-images --only='gitserver'` to update only the image of a specific service.

## Test plan

Code review.


Strum355 (approved):
LGTM :clueless:

unknwon (approved):

PR #60626, opened 2024-02-19
Symbol search: support content-based lang detection

jtibshirani:
The symbol service already stores each file's detected language. Now, when the
'search-content-based-lang-detection' feature is enabled, the symbols service
filters directly on the language column.

This PR doesn't update Rockskip, since it doesn't store the detected language.

Relates to https://github.com/sourcegraph/sourcegraph/issues/60341

## Test plan

Added unit tests, manual testing


cmd/symbols/internal/database/store/search.go
```diff
@@ -120,6 +127,10 @@ func makeSearchCondition(column string, regex string, isCaseSensitive bool) *sql
 	return sqlf.Sprintf(column+" REGEXP %s", regex)
 }
 
+func makeLangCondition(lang string) *sqlf.Query {
```
jtibshirani, line 130: One thing to be aware of: the languages stored here are produced by ctags, whereas the lang filters are expected to be normalized languages from go-enry. I haven't noticed any difference in practice.

jtibshirani, line 130: One thing to be aware of: the languages stored here are produced by ctags, whereas the lang filters are expected to be normalized languages from go-enry. I looked through `universal-ctags --list-languages` and they seem to match up well, so I don't expect this to be a problem in practice.

varungandhi-src: Strong suggestion: Please let's use different types here `EnryLanguage` and `CtagsLanguage`. These kinds of problems come up in edge cases. For example,
- `CUDA` is all upper-case in ctags but `Cuda` in enry.
- `COBOL` is uppercase in enry but `Cobol` in ctags.
- `MatLab` in ctags is `MATLAB` in enry.
- `Protobuf` in ctags is `Protocol Buffers` in enry
- ctags has both `Perl6` and `Raku` (which are the same language IIUC) whereas enry only has `Perl 6`
- `ObjectiveC` in ctags is `Objective-C` in enry
- `Asm` in ctags is `Assembly` in enry (there's also `Unix Assembly`)
- ctags has `Zsh` and `Sh` whereas enry has `Shell`.

Some of these are fixable with the case conversion below. But some are not.

I'm not expecting this PR to have a perfect conversion function. But it seems conceptually wrong to conflate these two.

internal/rockskip/server_test.go
```diff
@@ -118,7 +119,11 @@ func TestIndex(t *testing.T) {
 	verifyBlobs := func() {
 		repo := "somerepo"
 		commit := getHead()
-		args := search.SymbolsParameters{Repo: api.RepoName(repo), CommitID: api.CommitID(commit), Query: ""}
+		args := search.SymbolsParameters{
+			Repo:         api.RepoName(repo),
+			CommitID:     api.CommitID(commit),
+			Query:        "",
+			IncludeLangs: []string{"Text"}}
```
varungandhi-src, line 126: I thought this patch was additive, so I don't quite understand why this part of the test needs an update, since the `"Text"` string doesn't appear elsewhere in this patch. Is this test going to continue to pass even if `[]string{"Text"}` is omitted/did you add this to make sure we're exercising the 'has language filter' code path?

cmd/symbols/internal/database/store/search.go
jtibshirani: Thanks, there are indeed more exceptions than I realized. To check I understand your suggestion -- we should have an explicit conversion from go-enry language to ctags language, and update naming to make this distinction clear.

internal/rockskip/search.go
```diff
@@ -437,6 +438,16 @@ func convertSearchArgsToSqlQuery(args search.SymbolsParameters) *sqlf.Query {
 	// ExcludePaths
 	conjunctOrNils = append(conjunctOrNils, negate(regexMatch(pathConditions, args.ExcludePattern, args.IsCaseSensitive)))
 
+	// Rockskip doesn't store the file's language, so we convert the language filters into path filters as a
+	// best effort approximation. We ignore the search's case-sensitivity, since it doesn't apply to these filters.
```
varungandhi-src, line 442: Suggestion: Might be useful to file an issue for this gap and link that from this comment.

cmd/symbols/internal/database/store/search.go
varungandhi-src: > To check I understand your suggestion -- we should have an explicit conversion from go-enry language to ctags language, and update naming to make this distinction clear.

Yes, and specifically for naming, I'm emphasizing that we should have different _types_ (not just variable names) for these two, and a conversion function from one type to another.

We have something similar in the syntax highlighter, albeit more complicated, to translate Sublime grammar names to Tree-sitter language names.

https://sourcegraph.com/github.com/sourcegraph/sourcegraph@8379154d11d98bae47f5a7bb17336485e8b93ecb/-/blob/docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting.rs?L163-201

Right now, the `langInclude` and `langExclude` in various places are just strings instead of being domain-specific types, and that's understandable, but changing that to be a more specific type would be valuable. (However, I'm not suggesting you do that in this PR specifically.)

internal/rockskip/search.go
jtibshirani: I plan to file a follow-up issue to #60341 tracking the tasks to fully complete the feature. It will include the inconsistencies we observed in the filters panel too.

internal/rockskip/server_test.go
jtibshirani: Oops, I did this too quickly! I just pushed a correction. Here I added a lang filter on "text" to ensure the new codepath in `internal/rockskip/search.go` is exercised. If we remove that logic, this test will fail.

Overall, the rockskip tests are *really* lacking. If we make further changes, we will invest in new unit testing suites.

cmd/symbols/internal/database/store/search.go
jtibshirani: Great, that matches my understanding -- I will punt on any new types for this PR.

cmd/symbols/internal/api/search_sqlite.go
```diff
@@ -36,6 +36,8 @@ func MakeSqliteSearchFunc(observationCtx *observation.Context, cachedDatabaseWri
 			attribute.Int("numIncludePatterns", len(args.IncludePatterns)),
 			attribute.String("includePatterns", strings.Join(args.IncludePatterns, ":")),
 			attribute.String("excludePattern", args.ExcludePattern),
+			attribute.String("includeLangs", strings.Join(args.IncludeLangs, ":")),
```
keegancsmith, line 39: I know you are just matching the pattern used here, but a better attribute to use is `attribute.StringSlice` which will avoid extra work if we don't capture the trace + the trace will be displayed better in tracing UIs

keegancsmith (approved):

jtibshirani:
Thanks for the reviews! I realize this isn't the most high quality + complete solution, but just putting in "best effort" here until we can really invest in completing the feature.

Issue #56712, opened 2023-09-16
gitserver: Improve mapping of paths to repository


eseliger:
Today, gitserver loosely tracks a repo on disk with a repo in the database by turning the repo name into a dir on disk, and vice-versa.

So for example for a repo called `github.com/sourcegraph/sourcegraph`, the path will be `/data/repos/github.com/sourcegraph/sourcegraph`, and when looking at the path `/data/repos/github.com/sourcegraph/sourcegraph`, the repo will be found by saying `WHERE repo.name = 'github.com/sourcegraph/sourcegraph'`. 

So far so simple, but `name` is not actually a unique identifier in the Sourcegraph `repo` table. Also, it could change over time (repo is renamed).
So what happens when a repo is renamed on the code host? TODO figure this out.
When a repo is deleted on the code host, it's name in the database is turned into `DELETED-<transaction_timestamp>-<name>`, which makes the mapping harder. We can still find the path on disk by stripping the `DELETED-<transaction_timestamp>-` prefix, but we are not able to go the other way around: from the repo path on disk to the deleted version of the name, because we cannot know the transaction_timestamp.
Last, when a repo called `github.com/sourcegraph/sourcegraph` is deleted, and a NEW repo on the code host (different `external_id`) with the same name, the DB looks like the following:
- `name: github.com/sourcegraph/sourcegraph external_id: 234`
- `name: DELETED-123456789-github.com/sourcegraph/sourcegraph external_id: 123`
After our above logic that turns a repo name into a path by stripping the `DELETED-<ts>` prefix, two repo records now point to the same directory.

We should find a stronger mapping that is not based on the name, for all the reasons above.
Ideas that we had before:
- Store a `repo_dir` field on the gitserver_repos entry, so we can reverse-lookup path->repo and can create another directory for the `external_id: 234` repo from the duplicate case above
- Store the `repo.id` field in the clone directory
- Migrate the gitserver filesystem to a new structure that includes repo IDs, perhaps by utilizing symlinks for the debugging benefits, while keeping a large directory that just contains numeric repos, like `/data/repos/by-id/<id>`, which `/data/repos/github.com/sourcegraph/sourcegraph` could symlink to, for simpler debugging

/cc @sourcegraph/source

Issue #35960, opened 2022-05-24
Repos that have .git in the name do not sync properly


billCaplan:
- **Sourcegraph version:** 3.30.4
- **Platform information:** Kubernetes

#### Steps to reproduce:

1. In a code host connection, include a repo that has .git in the name (In the actual repo name, not the file extension Ex: repo_name.git.git
2. Save and try to sync repos

#### Expected behavior:
We are able to properly find and sync the repo.

#### Actual behavior:
It appears we strip the .git from the repo name when we save it, so instead of trying to sync repo_name.git, we just try to sync repo_name

From Gitserver:

> 
> Gitserver log:
> 
> t=2022-04-07T16:00:50+0000 lvl=warn msg="error cloning repo" repo=<repo name redacted, but does not include .git> <repo URL redacted, but does not include .git> err="get VCS syncer: get repository: repo not found:

If would like immediate help on this, please email support@sourcegraph.com (you can still create the issue, but there are not [SLAs](https://about.sourcegraph.com/support/) on issues like there are for support requests).

PR #60627, opened 2024-02-19
gitserver: Introduce FS layer to encapsulate repo name conversions

eseliger:
As posted on slack yesterday:

When talking to gitserver, in various random layers we call a function called `protocol.NormalizeRepo`. This function potentially “normalizes” the repo name, aka. it changes it.

Gitserver, historically (years ago) was a simple git proxy. So you ask it for repo content, and it eventually delivers. It’s been a write-through cache, basically.
Then, we transformed it a bunch of times, first to keep a fixed set of repos always (code host connection syncing), then later to be a stateful service that persists it’s state in the DB.

And here comes the thing: Unlike years ago, gitserver has to map what it has on disk to the DB. The key we use for that mapping is the *repo name*. (Which is not a great key, because we reuse the name when a repo is deleted and created under another external ID, but that’s for later..).

So when we ask gitserver to clone a repo, it first asks the DB: “Hey how do I clone this repo with name X?“. During regular operations, we try to keep the gitserver_repos table up to date, by saying “repo with name X is cloned now”.
So it all depends on the repo name, it _has_ to match something in the db, otherwise a clone cannot work. Otherwise, the repo state cannot be updated in the DB.

When we talk to gitserver, we currently sometimes use this “normalization” function. Sometimes not. You’ll notice that when you add a repo called `https://bitbucket.sgdev.org/SOUR/src-cli.git.git` (notice the double .git, one of them is part of the name, only the latter is part of the git url scheme), it fails cloning initially, but eventually works. Because some request along the way does not do that “normalization”, thus letting gitserver finally clone the repo alright. But it takes a bunch of failed attempts. Recloning the repo from the UI doesn’t work, it complains that the repo was not found (because that request does corrupt the name somewhere in the layers). The same issue applies for `api.UndeletedRepoName`, it tries to do the right thing, but does it at a too high level, thus making that we lose track of the repo in the DB.

In gitserver we do the following:
When we get a repo name, we normalize to get the dir on disk. So `[bitbucket.sgdev.org/SOUR/src-cli.git](http://bitbucket.sgdev.org/SOUR/src-cli.git)` becomes `/data/repos/bitbucket.sgdev.org/SOUR/src-cli/.git`. Some other normalizations make sure that the dir is a valid path and no malicious paths can be used (../../etc/passwd and so forth).
When we traverse over the repos on disk to do maintenance tasks etc, we don’t have a repo name, we have a repo dir instead. So for those, we do this process the other way around. `/data/repos/bitbucket.sgdev.org/SOUR/src-cli/.git` becomes `[bitbucket.sgdev.org/SOUR/src-cli](http://bitbucket.sgdev.org/SOUR/src-cli)`. Notice? The name is NOT the same name as it was initially, because this conversion is not lossless back and forth.

So we are no longer able to map the repository on disk to the corresponding database entry. Bad. Even worse, we cannot clone repos that end in `.git`, and repos called `bitbucket/x` and `bitBucket/x` are suddenly not updating the right repo in the DB.

---

There’s a few things we can do here, what I went for in this PR is to push those normalizations further down into the FS layer, but keep it untouched in the layers where we deal with `api.RepoName` which is often passed to the DB.

Ultimately, as soon as we have to make a conversion or when a repo name changes, this mapping will still break. It’ll require a reclone, or make it so that we cannot update DB state in janitor anymore once a repo was soft-deleted.
This seems like a harder problem to solve, but I think we can get into a much better state by pushing these normalizations into the FS layer for now.

To make this abstraction as obvious as possible, I decided to do a migration I intended for a while anyways: Encapsulate the FS activities in an interface. This also makes it easier to mock things out, for example the check for IsRepoCloned in one of the tests.

**Considerations when reviewing this code:**

- Please triple check my inline comments where I state why I think the normalization in that place is not required
- Please quadruple check my gut that (given normalization broke cloning before) there should be no major disruption changing the AddrForRepo logic

Closes https://github.com/sourcegraph/sourcegraph/issues/35960

Closes https://github.com/sourcegraph/sourcegraph/issues/56712

## Test plan

Pruned gitserver disk locally, let it reclone all the things, made sure clones work as expected, even for repos that end in `.git`. Also verified that deleting a repo now properly finds the correct repo in the DB to update.


internal/gitserver/client.go
```diff
@@ -602,20 +602,18 @@ func (c *RemoteGitCommand) sendExec(ctx context.Context) (_ io.ReadCloser, err e
 		}
 	}()
 
-	repoName := protocol.NormalizeRepo(c.repo)
-
```
eseliger, line 606: Normalizing the repoName here did nothing good. `ClientForRepo` already applied it, so that was a noop, and the `Repo` arg in the ExecRequest basically got corrupted, meaning that gitserver could never find the corresponding DB entry for the repo sent over (because we corrupt the de-facto PK).

Removing it here doesn't actually change anything other than not sending a potentially wrong name over. The ClientForRepo function didn't benefit from this line anyways.

internal/gitserver/client.go
```diff
@@ -657,9 +655,7 @@ func (c *clientImplementor) Search(ctx context.Context, args *protocol.SearchReq
 	})
 	defer endObservation(1, observation.Args{})
 
-	repoName := protocol.NormalizeRepo(args.Repo)
```
eseliger, line 660: See the same description as for the Exec request:

Normalizing the repoName here did nothing good. `ClientForRepo` already applied it, so that was a noop, and the `Repo` arg in the ExecRequest basically got corrupted, meaning that gitserver could never find the corresponding DB entry for the repo sent over (because we corrupt the de-facto PK).

Removing it here doesn't actually change anything other than not sending a potentially wrong name over. The ClientForRepo function didn't benefit from this line anyways.

internal/gitserver/client.go
```diff
@@ -874,11 +870,7 @@ func (c *clientImplementor) Remove(ctx context.Context, repo api.RepoName) (err
 	})
 	defer endObservation(1, observation.Args{})
 
-	// In case the repo has already been deleted from the database we need to pass
-	// the old name in order to land on the correct gitserver instance
-	undeletedName := api.UndeletedRepoName(repo)
-
-	client, err := c.ClientForRepo(ctx, undeletedName)
```
eseliger, line 881: We actually _cannot_ do that. It means that gitserver can not correctly set the DB state of this repo to uncloned after it removed the repo.

Scenario:
repo name is `DELETED-123123-github.com/src-cli`, in the DB the `repo.name` column contains exactly that value. The 123123 is the timestamp of the time it was deleted at. It cannot be reconstructed.

Now we tell gitserver to delete this repo, it can do that alright, because it correctly finds the undeleted version `github.com/src-cli` on disk (the new version can still do that, because we now Undelete the repo name at the name->fs path conversion layer).
It now attempts to update `github.com/src-cli`, which fails. We have not correctly set the state in the DB. 

Now, since we don't scramble the name, this works as expected.

internal/gitserver/addrs.go
```diff
@@ -174,8 +173,11 @@ type GitserverAddresses struct {
 func (g *GitserverAddresses) AddrForRepo(ctx context.Context, repoName api.RepoName) string {
 	addrForRepoInvoked.Inc()
 
-	// Normalizing the name in case the caller didn't.
-	name := string(protocol.NormalizeRepo(repoName))
```
eseliger, line 178: Normalizing is an operation to make the reponame safe to map to the filesystem on disk. It is not needed here. It didn't hurt that it was here, but it is certainly confusing. It also means that the addrforrepo logic is more complex than it has to be, making it harder to potentially reimplement in in the DB layer or elsewhere.

What we do need to do though is undeleting the RepoName here. We want to map the requests to the same shard for `DELETED-123123-github.com/src-cli` and `github.com/src-cli`, so we can delete a repo from disk after the latter has been marked as deleted.
Note that this only affects the sharding decision, NOT the repo name sent over.

cmd/gitserver/shared/shared.go
```diff
@@ -67,7 +67,8 @@ func Main(ctx context.Context, observationCtx *observation.Context, ready servic
 	}
 
 	// Prepare the file system.
-	if err := gitserverfs.InitGitserverFileSystem(logger, config.ReposDir); err != nil {
+	fs := gitserverfs.New(observationCtx, config.ReposDir)
```
eseliger, line 70: Feels nice to not have to pass the reposDir around everywhere, and instead have a subsystem that manages it.

cmd/gitserver/internal/servermetrics.go
```diff
@@ -24,44 +20,6 @@ func (s *Server) RegisterMetrics(observationCtx *observation.Context, db dbutil.
 		// See https://github.com/sourcegraph/sourcegraph/issues/54317 for details.
 		s.logger.Warn("Disabling 'echo' metric")
 	}
-
```
eseliger, line 27: These live on the FS implementation now.

cmd/gitserver/internal/search.go
```diff
@@ -84,7 +83,6 @@ func (s *Server) SearchWithObservability(ctx context.Context, tr trace.Trace, ar
 // search handles the core logic of the search. It is passed a matchesBuf so it doesn't need to
 // concern itself with event types, and all instrumentation is handled in the calling function.
 func (s *Server) search(ctx context.Context, args *protocol.SearchRequest, onMatch func(*protocol.CommitMatch) error) (limitHit bool, err error) {
-	args.Repo = protocol.NormalizeRepo(args.Repo)
```
eseliger, line 87: No need to "normalize" here, it just scrambles the repo name so we cannot properly map it back to the DB entity. 
The FS handles getting the path correctly.

cmd/gitserver/internal/patch.go
```diff
@@ -45,21 +44,20 @@ func (s *Server) CreateCommitFromPatch(ctx context.Context, req protocol.CreateC
 
 	var resp protocol.CreateCommitFromPatchResponse
 
-	repo := string(protocol.NormalizeRepo(req.Repo))
-	repoDir := filepath.Join(s.reposDir, repo)
-	repoGitDir := filepath.Join(repoDir, ".git")
```
eseliger, line 50: this actually didn't use the helper method at all, which could've caused an inconsistency. We now properly use RepoDir on the fs here.

cmd/gitserver/internal/patch.go
```diff
@@ -293,7 +291,7 @@ func (s *Server) CreateCommitFromPatch(ctx context.Context, req protocol.CreateC
 			resp.ChangelistId = cid
 		} else {
 			cmd = exec.CommandContext(ctx, "git", "push", "--force", remoteURL.String(), fmt.Sprintf("%s:%s", cmtHash, ref))
-			cmd.Dir = repoGitDir
+			repoGitDir.Set(cmd)
```
eseliger, line 294: This is important as it also sets some addtl env vars on the git command.

cmd/gitserver/internal/patch.go
```diff
@@ -330,7 +328,7 @@ func (s *Server) CreateCommitFromPatch(ctx context.Context, req protocol.CreateC
 
 	if req.PushRef == nil {
 		cmd = exec.CommandContext(ctx, "git", "update-ref", "--", ref, cmtHash)
-		cmd.Dir = repoGitDir
+		repoGitDir.Set(cmd)
```
eseliger, line 331: This is important as it also sets some addtl env vars on the git command.

cmd/gitserver/internal/perforce/changelist.go
```diff
@@ -17,12 +17,6 @@ import (
 
 // GetChangeListByIDArguments are the arguments for GetChangelistByID.
 type GetChangeListByIDArguments struct {
-	// ReposDir is the directory where the repositories are stored.
-	ReposDir string
-	// P4Home is the path to the directory that 'p4' will use as $HOME
-	// and where it will store cache data.
-	P4Home string
-
```
eseliger, line 25: Ideally, we would move to the same structure that we adopted for git recently: Have a backend interface that proxies all these calls. That way, we can feed the FS at backend creation time and don't need to make it a concern of the caller to bring the right FS as an argument of the function.
Anyhow, that's for later. For now, we pass the FS down explicitly.

cmd/gitserver/internal/server.go
```diff
@@ -498,8 +499,7 @@ func (s *Server) IsRepoCloneable(ctx context.Context, repo api.RepoName) (protoc
 func (s *Server) RepoUpdate(req *protocol.RepoUpdateRequest) protocol.RepoUpdateResponse {
 	logger := s.logger.Scoped("handleRepoUpdate")
 	var resp protocol.RepoUpdateResponse
-	req.Repo = protocol.NormalizeRepo(req.Repo)
```
eseliger, line 501: Normalizing the repo here means that the CloneRepo below would not be able to get the VCSSyncer for the repo, failing the clone. We now keep the name as-is.

cmd/gitserver/internal/server.go
```diff
@@ -630,6 +634,7 @@ func (s *Server) CloneRepo(ctx context.Context, repo api.RepoName, opts CloneOpt
 	// the actual running clone for the DB state of last_error.
 	defer func() {
 		// Use a different context in case we failed because the original context failed.
+		// TODO: Should not store anything when tryacquire returned false.
```
eseliger, line 637: Unrelated but wanted to leave a note for myself :D

cmd/gitserver/internal/server.go
```diff
@@ -1136,8 +1168,7 @@ func (s *Server) doBackgroundRepoUpdate(repo api.RepoName, revspec string) error
 		return err
 	}
 
-	repo = protocol.NormalizeRepo(repo)
```
eseliger, line 1139: Normalizing here means that getRemoteURL right below would fail. We don't do that anymore.

cmd/gitserver/internal/statesyncer.go
```diff
@@ -189,11 +188,6 @@ func syncRepoState(
 		for _, repo := range repos {
 			repoSyncStateCounter.WithLabelValues("check").Inc()
 
-			// We may have a deleted repo, we need to extract the original name both to
-			// ensure that the shard check is correct and also so that we can find the
-			// directory.
-			repo.Name = api.UndeletedRepoName(repo.Name)
```
eseliger, line 195: The addrforrepo func handles that. Altering the name means the state syncer will not correctly update the record in the DB.

cmd/gitserver/internal/statesyncer.go
```diff
@@ -202,9 +196,12 @@ func syncRepoState(
 			}
 			repoSyncStateCounter.WithLabelValues("this_shard").Inc()
 
-			dir := gitserverfs.RepoDirFromName(reposDir, repo.Name)
-			cloned := repoCloned(dir)
-			_, cloning := locker.Status(dir)
+			cloned, err := fs.RepoCloned(repo.Name)
+			if err != nil {
+				// Failed to determine cloned state, we have to skip this record for now.
```
eseliger, line 201: q: does this make sense as the action to take, skip over it?

internal/gitserver/protocol/util.go
```diff
@@ -14,33 +14,5 @@ func NormalizeRepo(input api.RepoName) api.RepoName {
 	repo = path.Clean("/" + repo)
```
eseliger, line 14: Ideally, this function doesn't need to exist at all. Whenever it makes a transformation, it will break the reponame->dir->reponame mapping. 
For now, until we spend more time on a better mapping table, for security reasons we will keep this simplified version, but the goal really is to get rid of it.

cmd/gitserver/internal/cleanup.go
```diff
@@ -383,26 +400,27 @@ func cleanupRepos(
 		}
 
 		// The repo does not exist in the DB (or is soft-deleted), continue deleting it.
-		err = gitserverfs.RemoveRepoDirectory(ctx, logger, db, shardID, reposDir, dir, false)
+		// TODO: For soft-deleted, it might be nice to attempt to update the clone status,
+		// but that can only work when we can map a repo on disk back to a repo in DB
+		// when the name has been modified to have the DELETED- prefix.
+		err = fs.RemoveRepo(repoName)
```
eseliger, line 406: For later.

cmd/gitserver/internal/server_test.go
```diff
@@ -74,32 +74,6 @@ func TestExecRequest(t *testing.T) {
...
-				Repo: "github.com/nicksnyder/go-i18n",
-				Args: [][]byte{[]byte("diff")},
-			},
-			ExpectedCode:  codes.NotFound,
-			ExpectedError: "repo not found",
-			ExpectedDetails: []any{&v1.RepoNotFoundPayload{
-				Repo:            "github.com/nicksnyder/go-i18n",
-				CloneInProgress: true,
-			}},
-		},
```
eseliger, line 102: These are gRPC layer concerns and with the new FS mocks needed reimplementing but didn't actually test anything important, so skipped over.

cmd/gitserver/internal/gitserverfs/fs.go
```diff
@@ -48,39 +255,39 @@ func RepoNameFromDir(reposDir string, dir common.GitDir) api.RepoName {
 	return protocol.NormalizeRepo(api.RepoName(name))
```
eseliger, line 255: why normalize here?

cmd/gitserver/internal/gitserverfs/fs.go
```diff
@@ -1,45 +1,252 @@
...
+// and where it will store cache data.
+const p4HomeName = ".p4home"
+
+func repoDirFromName(reposDir string, name api.RepoName) common.GitDir {
 	p := string(protocol.NormalizeRepo(name))
 	return common.GitDir(filepath.Join(reposDir, filepath.FromSlash(p), ".git"))
 }
 
-func RepoNameFromDir(reposDir string, dir common.GitDir) api.RepoName {
+func repoNameFromDir(reposDir string, dir common.GitDir) api.RepoName {
```
eseliger, line 249: Note: This reverse lookup is still flawed and needs more work in another PR, because when we need to normalize a repo name to disk, we can't retrieve the original name ever.

We'll need to introduce a map or something that keeps the normalizations stored.

cmd/gitserver/internal/gitserverfs/fs.go
```diff
@@ -1,45 +1,252 @@
...
-const P4HomeName = ".p4home"
+func New(observationCtx *observation.Context, reposDir string) FS {
+	return &realGitserverFS{
+		logger:              observationCtx.Logger.Scoped("gitserverfs"),
+		observationCtx:      observationCtx,
+		reposDir:            reposDir,
+		clonedState:         make(map[api.RepoName]struct{}),
+		lastCloneStateReset: time.Now(),
+	}
+}
```
eseliger, line 50: run initialize in here instead of exposing it

pjlast (approved):
Went over this in a pairing session. Spotted one possible out of date comment?

cmd/gitserver/internal/clone.go
```diff
@@ -5,37 +5,43 @@ import (
...
-// clone operation is running for that repo yet. If it is already cloning, a NotFound
-// error with CloneInProgress: true is returned.
+// cloning the repository in the background and returns a CloneStatus.
 // Note: If disableAutoGitUpdates is set in the site config, no operation is taken and
 // a NotFound error is returned.
-func (s *Server) MaybeStartClone(ctx context.Context, repo api.RepoName) (notFound *protocol.NotFoundPayload, cloned bool) {
-	dir := gitserverfs.RepoDirFromName(s.reposDir, repo)
-	if repoCloned(dir) {
-		return nil, true
+func (s *Server) MaybeStartClone(ctx context.Context, repo api.RepoName) (cloned bool, status CloneStatus, _ error) {
```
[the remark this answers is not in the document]
eseliger: It's still true, isn't it? maybe during merge conflicts this diff was scrambled though 🙈 https://github.com/sourcegraph/sourcegraph/pull/60627/files#diff-fbe02db9cf392abbd5452e8c6dee945bdee1eea7044a5264485b6a4ffe704e08R32-R34

PR #60635, opened 2024-02-20
Return better error and status code from dotcom refresh limits api

thenamankumar:
The refresh rate limits API on dotcom always returns 500 when it fails to update rate limits on dotcom.

Related Issue: https://github.com/sourcegraph/accounts.sourcegraph.com/issues/562

This PR updates the API to return a better error message and status.

This will also propagate the 429 error it receives from the gateway to the caller SSC. 
This will allow SSC to retry at a later time and do better error logging in sentry. 

SSC side of PR: https://github.com/sourcegraph/accounts.sourcegraph.com/pull/579

## Test plan

ran locally

### DIFF
--- a/cmd/frontend/graphqlbackend/user.go
+++ b/cmd/frontend/graphqlbackend/user.go
@@ -531,7 +531,7 @@ func (r *schemaResolver) ChangeCodyPlan(ctx context.Context, args *changeCodyPla
 		return nil, err
 	}
 
-	if err := cody.RefreshGatewayRateLimits(ctx, userID, r.db); err != nil {
+	if err, _ := cody.RefreshGatewayRateLimits(ctx, userID, r.db); err != nil {
 		// We intentionally don't fail the upgrade flow here, Gateway will pickup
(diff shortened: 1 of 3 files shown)


cmd/frontend/graphqlbackend/user.go
```diff
@@ -531,7 +531,7 @@ func (r *schemaResolver) ChangeCodyPlan(ctx context.Context, args *changeCodyPla
 		return nil, err
 	}
 
-	if err := cody.RefreshGatewayRateLimits(ctx, userID, r.db); err != nil {
+	if err, _ := cody.RefreshGatewayRateLimits(ctx, userID, r.db); err != nil {
```
vdavid, line 534: Do we want to log the status code here, too?

vdavid (approved):
Nice. Left a question. Approving to unblock.

cmd/frontend/graphqlbackend/user.go
thenamankumar: This mutation isn't used at all now. Will remove the December GA stuff later.

PR #60636, opened 2024-02-20
Replace /get-cody links with /cody/manage

vdavid:
`/get-cody` is just a redirect to `/cody/manage`.
Cody clients probably contain a lot of links to `/get-cody` so we still need that redirect (and will need it for a year plus) but within the web app, we can go to the management page directly.

## Test plan

Not tested yet!


chrsmith (approved):
I assume you know what you are doing. But what happens if you navigate to `/cody/manage` and you aren't signed in? Are there any other "weird corner cases" to think about that `/get-cody` might be handling that `/cody/manage` does not?

PR #60638, opened 2024-02-20
lib/output: detectCapabilities on output writer

keegancsmith:
I had go tests failing inside of the emacs compilation-mode buffer. It was due to detecting the capabilities of the terminal incorrectly, leading to us capturing output in tests with extra escape codes. This lead to tests failing since we would assert on the output captured.

From reading the logic of when we pass in a different std.Output, we should in fact be detecting capabilities on it. It is possible now that we incorrectly detect when wrapping the writer. However, from looking at call sites we don't do that outside of tests.

Test Plan: go test

### DIFF
--- a/lib/output/capabilities.go
+++ b/lib/output/capabilities.go
@@ -1,6 +1,7 @@
 package output
 
 import (
+	"io"
 	"os"
 	"strconv"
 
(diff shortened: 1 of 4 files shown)


eseliger (approved):
> It is possible now that we incorrectly detect when wrapping the writer. However, from looking at call sites we don't do that outside of tests.

Could that be a future footgun?

PR #60643, opened 2024-02-20
gitserver: Fix infinite clone loop for repo duplicates

### LINKED ISSUE #55545: gitserver: reports error when cloning repository with same name of deleted-but-not-purged repository
Reported in in [Slack](https://sourcegraph.slack.com/archives/C05EMJM2SLR/p1691063941309499?thread_ts=1690911929.734609&cid=C05EMJM2SLR).

When repo is marked as deleted in `repo` table, but not yet [purged](https://github.com/sourcegraph/sourcegraph/blob/eabdecd94105af576df2d74bf5bc15e9e26d4cd0/internal/repos/purge.go#L91) from the disk space, there might be a case when another repo with the same name is added and we [error](https://github.com/sourcegraph/sourcegraph/blob/419272d4acd9669ddacd1fa76a030a3b4b0a9e7b/cmd/gitserver/server/server.go#L2272) while trying to clone, because the git dir is busy on the disk space.

eseliger:
This is admittedly a bit of a hack, but this cost us thousands of dollars on cloud before so I don't feel too terrible about this. The real solution will require to track repos by an actually unique identifier that is immutable, but that's much more work.

After the various fixes made to repo name normalization, there's less instances of "accidentally updating the wrong DB entry". In fact, after the previous PR everything already works alright. Except on a restart, where gitserver scans all repos and updates their clone status.

To work around this, we now exclude repos from state updates in the initial run that have another undeleted repo with the same name.

This is not a complete solution, but addresses all the happy path cases I could find where a repo was incorrectly recloned.

Closes #55545 

## Test plan

Recreated the scenario where repo A is deleted, and repo B has taken the name of repo A in the database. On main, we regularly delete this repo and reclone it. On this branch, this doesn't happen. Also adjusted tests for the store method.

### DIFF
--- a/internal/database/gitserver_repos.go
+++ b/internal/database/gitserver_repos.go
@@ -31,6 +31,16 @@ type GitserverRepoStore interface {
 	// our repo and gitserver_repos table. It is impossible for us not to have a
 	// corresponding row in gitserver_repos because of the trigger on repos table.
 	// Use cursors and limit batch size to paginate through the full set.
+	// Note that this iterator SKIPS any repos that have been soft deleted and another
+	// undeleted row has taken their name.
+	// This usually happens when a repo is deleted on the code host and another repo
+	// with the same name is created, the name will be the same, the external ID will
+	// not.
+	// This is a patch to the fact that we do not have a unique mapping from repo dir
+	// to the repo table. Skipping helps us avoid marking the repo as cloned, which
+	// will cause it to be pruned, and then the new repo will be cloned again,
+	// and eventually the statesyncer will remove the repo again; causing an infinite
+	// clone loop.
 	IterateRepoGitserverStatus(ctx context.Context, options IterateRepoGitserverStatusOptions) (rs []types.RepoGitserverStatus, nextCursor int, err error)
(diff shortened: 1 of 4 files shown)


internal/database/gitserver_repos.go
```diff
@@ -318,7 +328,12 @@ SELECT
 	gr.corruption_logs
 FROM gitserver_repos gr
 JOIN repo ON gr.repo_id = repo.id
-WHERE %s
+WHERE
+	%s
+	-- If the repo is deleted, we need to check if there are any undeleted repos with the same name,
+	-- if so we skip over it. This is not ideal, but it fixes an issue in the state syncer where
+	-- this setup of repos causes an infinite reclone loop.
+	AND NOT (repo.deleted_at IS NOT NULL AND EXISTS (SELECT 1 FROM repo other WHERE regexp_replace(repo.name, '^DELETED-\d+\.\d+-', '') = other.name AND other.deleted_at IS NULL))
```
pjlast, line 336: Well I'm going to trust this regexp 😄

eseliger: we don't have anything better right now, until we track repos on disk by ID :( This is just statesyncer, nothing janitor wouldn't eventually fix I guess.

PR #60650, opened 2024-02-20
ensure utm param is only included in links to dotcom from private ins…

dadlerj:
…tances; update the name

## Test plan

TBD


dadlerj:
Brittani will share more context! I haven't actually tested this code yet, so we should ensure that it doesn't break the site.

PR #60652, opened 2024-02-20
REALLY ready really means REALLY ready

jamesmcnamara:
As discussed in [this thread](https://sourcegraph.slack.com/archives/C04MYFW01NV/p1707144816301279), esbuild is not the slowest link anymore, so it instead pings the `HTTPS_WEB_SERVER_URL` until we get a 2xx response before announcing that we're open for business.

## Test plan
Ran locally variously disabling/delaying caddy and frontend to see that it logs errors and backs-off its ping


client/web/dev/esbuild/server.ts
```diff
@@ -66,14 +66,49 @@ export const esbuildDevelopmentServer = async (
...
+interface PingOptions {
+    url: string
+    backoffMillis?: number
+    factor?: number
+    maxBackoffMillis?: number
+}
+
+function pingUntilReady({
+    url,
+    backoffMillis = 500,
```
eseliger, line 87: do we need a backoff here? that'll potentially defer the ready message for perceived lower performance

eseliger (approved):
nice idea!

client/web/dev/esbuild/server.ts
jamesmcnamara: It's a butt load of output otherwise. And it caps at 10s long so at most you'll have a 10s pause on seeing the message assuming it's already taken >24s.

jhchabran (approved):

PR #60668, opened 2024-02-21
Revert "syntax-highlighter: Turn off Rocket logging with QUIET=true (#60660)"

varungandhi-src:
This reverts commit c19a432ab777335573aa4e6ffcf4668f1cb7be8a.

## Test plan

Run the following locally:

```
bazel build //docker-images/syntax-highlighter:image_tarball
docker run --detach --name=syntax-highlighter -p 9238:9238 syntect-server:candidate
curl http://localhost:9238/health
```

This prints OK after this revert, but not before.

### DIFF
--- a/docker-images/syntax-highlighter/src/main.rs
+++ b/docker-images/syntax-highlighter/src/main.rs
@@ -72,14 +72,12 @@ fn rocket() -> _ {
         .get(&tree_sitter_all_languages::ParserId::Go);
 
     // Only list features if QUIET != "true"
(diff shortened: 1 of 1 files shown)


keegancsmith (approved):

keegancsmith:
So why did this break the health check? Or was it just totally broken?

varungandhi-src:
Right, the process simply died for some reason inside the container (didn't happen when running the binary directly). Will look into this tomorrow so we can turn off the logging properly.

PR #60670, opened 2024-02-21
honey: HONEYCOMB_LOCAL which logs events to stderr

keegancsmith:
I found this very useful in my dev environment to debug some searches. Initially I added in a new event type. However, we do some transformations in our eventWrapper which I wanted to preserve. So instead at time of Send we decide if we should send to honeycomb or just log to stderr.

Note: this produces a lot of log output due to background processes running commands against git/etc.

Test Plan: "HONEYCOMB_LOCAL=true sg start enterprise" and see events in console when I do searches.

### DIFF
--- a/internal/honey/event.go
+++ b/internal/honey/event.go
@@ -1,6 +1,11 @@
 package honey
 
 import (
+	"fmt"
+	"os"
+	"slices"
+	"strings"
+
 	"github.com/honeycombio/libhoney-go"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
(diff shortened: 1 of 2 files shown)


keegancsmith:
@eseliger if you try this out, you will notice gitserver does a _lot_ of logging. It seems we now log when the janitor job does things like get git configuration. I am guessing these events are just unnecessary to log to honeycomb in general?

eseliger:
> I am guessing these events are just unnecessary to log to honeycomb in general?

We log all git latencies right now. I found that might be useful when we get to try to improve some overhead of certain calls, but I have no idea how much it costs on the honey side.

eseliger (approved):

Issue #60379, opened 2024-02-09
Symbol search results duplicated for certain query structures


jasonhawkharris:
When keyword search is enabled, certain query structures and filter combinations incorrectly return duplicate results. For example, when searching for:

`context:global repo:^github\.com/sourcegraph/sourcegraph$ repo link picker`

with the following filters selected:
![Screenshot 2024-02-09 at 1 41 09 PM](https://github.com/sourcegraph/sourcegraph/assets/62355966/9fcc5289-de15-4baf-9f6c-96d256846c6f)

This returns the same result listed three times. 
![Screenshot 2024-02-09 at 1 42 03 PM](https://github.com/sourcegraph/sourcegraph/assets/62355966/773fb587-da7f-43cc-a25c-7d84b83a29f2)

There are two issues:
1. Currently, symbol results are grouped into chunks. So repo, link, and picker each return as separate matches, even though they are from the same result. This chunking behavior is unintended in this case. 
2. The matched keyword ranges are not highlighted, making the duplication less obvious to the user. This is likely the bigger issue here. 

Potential solutions:
1. Keep the results as-is, but fix highlighting to show which tokens match the search keywords (i.e. highlight repo or link or picker). 
2. Modify the results to remove duplicate listings of the same result.

I recommend option one from a product standpoint, as it maintains completeness while improving visibility into why duplication occurs. But I welcome input from the platform team on feasibility or alternate suggestions.

For more context, here is [the original slack thread](https://sourcegraph.slack.com/archives/C05D629GWJK/p1707500762012769)

/cc @sourcegraph/search-platform

jasonhawkharris:
CC: @camdencheek @jtibshirani

PR #60697, opened 2024-02-22
search: deduplicate symbol results from zoekt

keegancsmith:
A query like `type:symbol repo link picker` is translated into zoekt as

```
(and sym:substr:"repo" sym:substr:"link" sym:substr:"picker")
```

If you have a symbol `RepoLinkPicker` Zoekt's result type will model this as 3 matches (which is correct) but all with the same `SymbolInfo.` When we convert this into Sourcegraph's symbol information we only want to have a match per symbol. Before this change we would translate this into 3 results, internally each having a different Column the match appears at. By the time it would reach the user via the streaming API the column is gone and we just have symbol information, which means effectively we had 3 duplicate results.

This commit makes what Zoekt returns match the expectations of Sourcegraph's symbol results. Note: we have to use a slight heuristic here where we say two symbols are identical if they have the same line number and symbol info.

Also note the TODO I added. The Column returned by Zoekt is where inside of the symbol we found a match. But in downstream parts of the code we seem to treat the Start Column as the start of the symbol. This is hard to fix now, so instead we make a note and just tackle this.

Test Plan: added a unit test. Additionally on my local devserver searching `repo link picker type:symbol` stopped returning duplicate results.

Fixes https://github.com/sourcegraph/sourcegraph/issues/60379

### DIFF
--- a/internal/search/zoekt/indexed_search.go
+++ b/internal/search/zoekt/indexed_search.go
@@ -591,6 +591,29 @@ func zoektFileMatchToSymbolResults(repoName types.MinimalRepo, inputRev string, 
 		}
 	}
 
+	// We deduplicate symbol matches. For example searching for "foo AND bar"
+	// will return the symbol "foobar" twice. However, sourcegraph's result
+	// type for symbol is modeled as a result per symbol. Additionally we use
+	// a heuristic to determine duplicate matches. We regard a match as the
+	// same if they have the same symbol info and appear on the same line. I
+	// am unaware of a language where you can break that assumption.
+	if len(symbols) <= 1 {
+		return symbols
+	}
+	dedup := symbols[:1]
+	for _, sym := range symbols[1:] {
+		last := dedup[len(dedup)-1]
+		if (sym.Symbol.Line == last.Symbol.Line) &&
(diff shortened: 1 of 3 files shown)


stefanhengl (approved):
LGTM!

internal/search/zoekt/indexed_search.go
```diff
@@ -591,6 +591,29 @@ func zoektFileMatchToSymbolResults(repoName types.MinimalRepo, inputRev string,
 		}
 	}
 
+	// We deduplicate symbol matches. For example searching for "foo AND bar"
+	// will return the symbol "foobar" twice. However, sourcegraph's result
+	// type for symbol is modeled as a result per symbol. Additionally we use
+	// a heuristic to determine duplicate matches. We regard a match as the
+	// same if they have the same symbol info and appear on the same line. I
+	// am unaware of a language where you can break that assumption.
```
camdencheek, line 599: Huh, surprising to me that we don't retain the ranges that matched within symbol. That would be a nice improvement at some point.

camdencheek (commented):
Thanks Keegan!

PR #60703, opened 2024-02-22
Adds a little program to check potential space savings in symbol trie

kritzcreek:
Mostly meant as a way to share the script and collect some data for different indexes. Check out the README for how to run (remember to cd into `dev/codeintel-merge-tries/` first).

/cc @varungandhi-src @keynmol

## Test plan

Just a script


varungandhi-src:
Thanks for submitting this as a PR (and in Rust instead of Purescript 😛), instead of taking the lazy route by sticking a script in a Slack thread like me, or in a GitHub issue.

Minor preference for having this under `lib/codeintel/tools` and not under `dev/`.

varungandhi-src:
Found the source of the discrepancy, it is because the input data is not sorted by ID. 

```
9362,25,NYTimes/gziphandler v1.1.1 `github.com/NYTimes/gziphandler`/,9355
9363,25,GzipHandler().,9362
9505,25,Must().,9502
9364,25,aws/smithy-go v1.13.5 `github.com/aws/smithy-go/middleware`/stringer#,9355
9365,25,String.,9364
```

Would prefer to port the Python algorithm to Rust since it was much smaller (no fixed point), with a sorting step at the start. It would also avoid the need to iterate over a hashmap in the current algorithm, which introduces non-determinism.

varungandhi-src (approved):
Apart from the point about the simplification PR and moving it to a different directory, I think this is OK. Thanks for documenting the Postgres query.

dev/codeintel-merge-tries/README.md
```diff
@@ -0,0 +1,17 @@
+A little program to check how many rows we could save by merging tries inside our
+`codeintel_scip_symbol_names` table.
+
+After uploading an index to your local sourcegraph instance, run like so:
+`psql -d sourcegraph -c "COPY (select * from codeintel_scip_symbol_names where upload_id = (select max(upload_id) from codeintel_scip_symbol_names)) TO STDOUT DELIMITER ',' CSV HEADER;" | cargo run --release`
+
+For example when running after uploading an index for sourcegraph/sourcegraph:
+
+```
+$ psql -d sourcegraph -c "COPY (select * from codeintel_scip_symbol_names where upload_id = (select max(upload_id) from codeintel_scip_symbol_names)) TO STDOUT DELIMITER ',' CSV HEADER;" | cargo run --release
```
varungandhi-src, line 10:
```suggestion
$ psql --no-psqlrc --csv --dbname=sourcegraph --command="select * from codeintel_scip_symbol_names where upload_id = (select max(upload_id) from codeintel_scip_symbol_names)" | cargo run --release
```

Slightly cleaner with longer flags for documentation

dev/codeintel-merge-tries/README.md
```diff
@@ -0,0 +1,17 @@
+A little program to check how many rows we could save by merging tries inside our
+`codeintel_scip_symbol_names` table.
+
+After uploading an index to your local sourcegraph instance, run like so:
+`psql -d sourcegraph -c "COPY (select * from codeintel_scip_symbol_names where upload_id = (select max(upload_id) from codeintel_scip_symbol_names)) TO STDOUT DELIMITER ',' CSV HEADER;" | cargo run --release`
+
+For example when running after uploading an index for sourcegraph/sourcegraph:
```
varungandhi-src, line 7:
```suggestion
Upload an index to your local Sourcegraph instance and run:
```

(Since the command is redundant with what is shown below.)

kritzcreek: Thanks, I split out the command itself into its own block so it's easier to copy (GitHub has this copy button for code blocks)

PR #60707, opened 2024-02-22
ci: debug logging for sg lint changed files

Strum355:
I've been unable to reproduce why `sg lint` appears to be running more lints than expected for a given diff, so lets add some friendly debug logging so we can investigate 😊 

## Test plan

:eye: CI :eye:

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -229,8 +229,6 @@ gazelle(
 sh_binary(
     name = "configure",
     srcs = ["//dev/ci:bazel-configure.sh"],
-    data = ["@go_sdk//:bin/go"],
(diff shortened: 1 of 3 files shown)


BUILD.bazel
```diff
@@ -229,8 +229,6 @@ gazelle(
 sh_binary(
     name = "configure",
     srcs = ["//dev/ci:bazel-configure.sh"],
-    data = ["@go_sdk//:bin/go"],
-    env = {"GO": "$(rootpath @go_sdk//:bin/go)"},
```
Strum355, line 233: This isnt needed by `//:gazelle`

jamesmcnamara (approved):

PR #60712, opened 2024-02-23
Simplify logic for trie merging

varungandhi-src:


kritzcreek:
Very nice, I stared at the Python for a while and couldn't figure out what what wrong. Good reminder to check your assumptions about the data first...

PR #60713, opened 2024-02-23
search: set options for NOT-INDEXED search

keegancsmith:
Came across this since I was checking we always set ChunkMatches: true. I additionally listed a few more options to ensure we don't bother sending back all the matches since we only rely on the stat. Additionally I raise the limits just to make sure we give accurate counts.

Test Plan: check not indexed count didn't change in dev server


jtibshirani (approved):
Looks good to me! Raising the limits super high seems fine, as it's a simple query and we don't perform this search very often.

PR #60714, opened 2024-02-23
search: remove support for reading LineMatches

keegancsmith:
We only ever ask for ChunkMatches and we more than likely will evolve the Zoekt API so it is the default and possible remove support for LineMatches.

Test Plan: go test


camdencheek (approved):

jtibshirani (approved):
Nice we're working to simplify this!

PR #60720, opened 2024-02-23
insights: remove config watcher

keegancsmith:
The computed state is so cheap that we might as well just compute it each time we want to access it. The main difference is we won't log insights backfiller changing. However, the old code just always logged at startup and whenever any site config changed. It also feels somewhat low value, so I think its fine just not to log.

Additionally the old code protected writes to task.config via a mutex, but never used that mutex for reads. So it was in fact racey.

FYI I came across this because I was looking into reducing the logspam in our product.

Test Plan: go test

### DIFF
--- a/internal/insights/scheduler/backfill_state_inprogress_handler.go
+++ b/internal/insights/scheduler/backfill_state_inprogress_handler.go
@@ -56,8 +56,6 @@ func makeInProgressWorker(ctx context.Context, config JobMonitorConfig) (*worker
 		MaxNumRetries:     3,
 	})
 
-	handlerConfig := newHandlerConfig()
-
 	task := &inProgressHandler{
 		workerStore:        workerStore,
 		backfillStore:      backfillStore,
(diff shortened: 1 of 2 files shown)


internal/insights/scheduler/backfill_state_inprogress_handler.go
```diff
@@ -255,18 +233,18 @@ func (h *inProgressHandler) doExecution(ctx context.Context, execution *backfill
 	}
 
 	execution.logger.Debug("starting primary loop", log.Int("seriesId", execution.series.ID), log.Int("backfillId", execution.backfill.Id))
-	if interrupted, err := itrLoop(h.config.pageSize, h.config.repoConcurrency, execution.itr.NextPageWithFinish); err != nil {
+	if interrupted, err := itrLoop(config.pageSize, config.repoConcurrency, execution.itr.NextPageWithFinish); err != nil {
```
keegancsmith, line 236: @chwarwick is this a safe change? Or where you previously relying on these values changing while doExecution ran?

camdencheek (approved):
LGTM. Thanks for the simplification!

PR #60721, opened 2024-02-23
Increase Enteprise Code Completions limit to 1000

rafax:
Increase default Code Completions limit for Enterprise Cody customers to 1000/user/day.

## Test plan

- existing unit tests were updated
- tested manually

### DIFF
--- a/internal/licensing/codygateway.go
+++ b/internal/licensing/codygateway.go
@@ -91,7 +91,7 @@ func NewCodyGatewayCodeRateLimit(plan Plan, userCount *int, licenseTags []string
 	default:
 		return CodyGatewayRateLimit{
 			AllowedModels:   models,
-			Limit:           int64(100 * uc),
+			Limit:           int64(1000 * uc),
 			IntervalSeconds: 60 * 60 * 24, // day
 		}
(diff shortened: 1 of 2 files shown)


internal/licensing/codygateway.go
```diff
@@ -91,7 +91,7 @@ func NewCodyGatewayCodeRateLimit(plan Plan, userCount *int, licenseTags []string
 	default:
 		return CodyGatewayRateLimit{
 			AllowedModels:   models,
-			Limit:           int64(100 * uc),
+			Limit:           int64(1000 * uc),
```
eseliger, line 94: we should rethink this, I think the default for enterprise-1 and enterprise-0 should be off now. But this change LGTM regardless and can be merged.

eseliger (approved):

internal/licensing/codygateway.go
rafax: Yeah, I think we need to update this to new licenceIDs and (ideally) simplify so it's future-proof, but that can likely happen next week and be owned by CPE team (@cbart @chwarwick are added as reviewers).

PR #60722, opened 2024-02-23
Allow sign-in providers for permissions use only

### LINKED ISSUE #60615: Allow that identity providers cannot be used for sign in
A customer would like to use an identity provider for connecting an external identity to Sourcegraph for repo syncing purposes, but they don't want users to be able to authenticate using that provider. We should provide an option for that.

pjlast:
Closes #60615 

The check that determines whether or not to auto-redirect to the auth provider does not take into account whether or not the auth provider is marked as hidden or not.

This PR fixes that so that if an auth provider his hidden, it is not factored into the decision of whether or not the user should be auto-redirected to the auth provider.

So now, if you have two auth providers configured:

```json
{
  "auth.providers": [
    {
      "type": "github",
      "hidden": true
    },
    {
      "type": "gitlab"
    }
  ]
}
```

you would be auto-redirected to the GitLab sign-in page when visting `/sign-in` on Sourcegraph

Furthermore, this PR also adds a new config option, `noSignIn`, that allows an auth provider to be only hidden from the sign-in page (`hidden` hides it from the sign-in and account security page).

## Test plan

Existing tests pass.


eseliger (commented):
This should probably apply to all provider types, not just OAuth

eseliger:
aha! that's interesting(read: a little confusing), but also probably fine :D

cmd/frontend/internal/app/jscontext/jscontext.go
```diff
@@ -41,6 +41,7 @@ var BillingPublishableKey string
 
 type authProviderInfo struct {
 	IsBuiltin         bool    `json:"isBuiltin"`
+	NoSignIn          bool    `json:"noSignIn"`
```
eseliger, line 44: I think we'd want to add that to the jscontext type definition on the JS side, to stay consistent.

eseliger (approved):

cmd/frontend/internal/app/jscontext/jscontext.go
pjlast: Ah sorry, this was WIP, should have marked it as draft again

PR #60752, opened 2024-02-26
search: do not record full error for cancel or timeout

keegancsmith:
When inspecting honeycomb there are lots of very noisy errors which are due to timeouts or cancellations. It is noisy since it is a combined error of all the backends that didn't finish. Additionally these are not errors, but rather expected behaviour due to how we use context to implement match limits and timeouts.

This commit instead does not record the error field in that case, but instead expands the "code" field to do record this. So we now have a more useful prometheus metric as well as the code field in honeycomb.

Test Plan: go test

### DIFF
--- a/internal/search/backend/metered_searcher.go
+++ b/internal/search/backend/metered_searcher.go
@@ -16,6 +16,7 @@ import (
 	"github.com/sourcegraph/sourcegraph/internal/actor"
 	"github.com/sourcegraph/sourcegraph/internal/honey"
 	"github.com/sourcegraph/sourcegraph/internal/trace"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
 )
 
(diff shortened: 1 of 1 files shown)


internal/search/backend/metered_searcher.go
```diff
@@ -168,14 +166,13 @@ func (m *meteredSearcher) StreamSearch(ctx context.Context, q query.Q, opts *zoe
 		}
 	}))
 
-	if err != nil {
-		code = "error"
-	}
+	code, maybeErrorStr := codeAndErrorStr(err)
```
jtibshirani, line 169: It's often nice to compute these lazily, but doesn't look easy to do here.

jtibshirani (approved):
Makes sense to me!

PR #60764, opened 2024-02-27
v2 telemetry: add v2 telemetry to src/components/fuzzyFinder

dadlerj:
Context:

## Test plan

`sg start`
visit page


client/web/src/components/fuzzyFinder/FuzzyModal.tsx
```diff
@@ -287,7 +288,7 @@ export const FuzzyModal: React.FunctionComponent<React.PropsWithChildren<FuzzyMo
             maxResults,
```
vovakulikov, line 288: We can avoid changing it in this PR but just noticed for myself that neither maxResults, initialMaxResults, or setMaxResults is used in the `renderFuzzyResults`

PR #60766, opened 2024-02-27
tracer: enable SDK logging in debug mode, remove sync exporter

bobheadxi:
Adds OTEL Go SDK logging enablement as part of the `tracing.debug` option. In particular, this is the only way to get the spans dropped by the batch span processor in the SDK: https://github.com/open-telemetry/opentelemetry-go/blob/1d1ecbc5f936208a91521ede9d0b2f557170425e/sdk/trace/batch_span_processor.go#L287C1-L288C1

To make sure this is usable, I've also removed the use of the sync span exporter on debug mode. This is rarely used and shouldn't be used in production, and knobs are available for making tracing "faster" (if needed) in local dev via `OTEL_BSP_*` env vars instead.

Part of https://sourcegraph.slack.com/archives/C04MYFW01NV/p1706107280411819

## Test plan

```
sg run otel-collector jaeger
sg start
```

Logs are available under `tracer.otel` scope:

```
[     embeddings] INFO tracer.otel global/internal_logging.go:52 "level"=4 "msg"="Tracer created" "name"="github.com/XSAM/otelsql" "version"="0.27.0" "schemaURL"=""
```

When debug mode is disabled, these logs don't show up


internal/tracer/tracer.go
```diff
@@ -130,3 +138,13 @@ func Init(logger log.Logger, c WatchableConfigurationSource) {
...
+
+type toggledLogrSink struct {
+	logr.LogSink
+	// debugMode is returned when Enabled() is called on this sink, instead of
+	// the underlying LogSink's implementation. In other words, if debug mode
+	// is enabled, all logs using this sink are enabled.
+	debugMode *atomic.Bool
+}
+
+func (s toggledLogrSink) Enabled(_ int) bool { return s.debugMode.Load() }
```
camdencheek, line 150: Oh, neat. I didn't realize it was _quite_ that straightforward. The `logr` interfaces seemed kinda gnarly.

camdencheek: If debug mode is _not_ enabled, should we fall back to the value of `s.LogSink.Enabled()`?

bobheadxi: Yeah I'm not a fan of the interface - this PR hacks around by using our  `sourcegraph/log` -> `std` adapter and providing that to `logr`. We lose the granularity of level, but levels are just numbers in `logr` with no fixed scale (????) and this should only be used for debugging, so oh well

internal/tracer/logged_otel.go
```diff
@@ -41,7 +41,6 @@ func (s *loggedOtelTracerProvider) concreteTracer(instrumentationName string, op
 		logger = s.logger.With(
 			log.String("tracerName", instrumentationName),
 			log.String("provider", fmt.Sprintf("%T", s.provider)))
-		logger.Info("Tracer")
```
bobheadxi, line 44: SDK generates this same message, and it's not useful anyway

camdencheek (approved):
LGTM! I never felt comfortable turning on debug mode anywhere other than locally anyways, so I think just always using the batch processor makes sense

internal/tracer/tracer.go
bobheadxi: > If debug mode is not enabled, should we fall back to the value of s.LogSink.Enabled()?

No, because I think the default sink has a level that will include many SDK logs. In most healthy scenarios, nobody needs to see these logs - OTEL has an error handler callback (L104) which reports things that export levels

camdencheek: got it, thanks!

bobheadxi: (IMO dropping large numbers of spans should be an error, but ¯\_(ツ)_/¯

PR #60767, opened 2024-02-27
v2t: add v2t to the rest of src/components

dadlerj:
Context:

## Test plan

`sg start`
visit page

### DIFF
--- a/client/web/src/components/SelfHostedCta/SelfHostedCta.tsx
+++ b/client/web/src/components/SelfHostedCta/SelfHostedCta.tsx
@@ -3,24 +3,33 @@ import React from 'react'
 import { mdiArrowRight } from '@mdi/js'
 import classNames from 'classnames'
 
+import { TelemetryV2Props } from '@sourcegraph/shared/src/telemetry'
 import type { TelemetryProps } from '@sourcegraph/shared/src/telemetry/telemetryService'
 import { Link, Icon, H3 } from '@sourcegraph/wildcard'
 
 import { MarketingBlock } from '../MarketingBlock'
 
-export interface SelfHostedCtaProps extends TelemetryProps {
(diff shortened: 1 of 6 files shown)


client/web/src/components/SelfHostedCta/SelfHostedCta.tsx
```diff
@@ -3,35 +3,46 @@ import React from 'react'
...
-    page: string
+    // The name of the page the CTA will be posted. Add more as needed. DO NOT
+    // include full URLs here, because this will be logged to our analytics
+    // systems. We do not want to expose private repo names or search queries
+    // to our analytics.
+    page: 'organizations' | 'storybook'
+}
+
+// pages as numbers for v2 telemetry
+const V2Pages: { [k in SelfHostedCtaProps['page']]: number } = {
```
vovakulikov, line 23: Not a blocker, just out of curiosity I see that in the new telemetry, we use numbers more than some enums or string IDs, is there a tech reason for this? Do we encode somehow this info later? Or it was just our decision (I have nothing against numeric IDs but maybe human readability probably is worse compared to some enum)

vovakulikov (approved):

client/web/src/components/SelfHostedCta/SelfHostedCta.tsx
dadlerj: Yes, this was intentional—only logging numeric values helps avoid any private data being captured when developers add new logging and don't look at the rules very carefully. It's unpleasant from a usability standpoint, yes, but worth it from a data privacy one.

vovakulikov: Makes sense, thanks

PR #60768, opened 2024-02-28
v2t: add src/enterprise/batches

dadlerj:
Context:

* https://github.com/sourcegraph/sourcegraph/pull/60767

## Test plan

`sg start`
visit page

### DIFF
--- a/client/web/src/enterprise/batches/batch-spec/execute/workspaces/WorkspaceDetails.tsx
+++ b/client/web/src/enterprise/batches/batch-spec/execute/workspaces/WorkspaceDetails.tsx
@@ -18,6 +18,7 @@ import indicator from 'ordinal/indicator'
 
 import { dataOrThrowErrors } from '@sourcegraph/http-client'
 import type { Maybe } from '@sourcegraph/shared/src/graphql-operations'
+import { TelemetryV2Props } from '@sourcegraph/shared/src/telemetry'
 import {
     Badge,
     LoadingSpinner,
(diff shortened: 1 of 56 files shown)


dadlerj:
This is a really big one, sorry! I've tested it more extensively than most others.

client/web/src/enterprise/batches/batch-spec/execute/workspaces/WorkspaceDetails.tsx
```diff
@@ -605,9 +639,12 @@ const WorkspaceStep: React.FunctionComponent<React.PropsWithChildren<WorkspaceSt
                             <Tabs
                                 size="medium"
                                 behavior="forceRender"
-                                onChange={index =>
+                                onChange={index => {
                                     eventLogger.log(`batch_change_execution:workspace_tab_${tabsNames[index]}:clicked`)
-                                }
+                                    telemetryRecorder.recordEvent('batchChange.execution.tab', 'click', {
+                                        metadata: { tab: index },
```
vovakulikov, line 645: Not a blocker, but this is why I'm a bit concerned about using index in arguments here, if we add more tabs before some existing tab or at the beginning it will shift all further tab's indexes and hence it will mess up the telemetry data that we collected before. 

I would leave a comment about this here, just to make sure the next person who touches this knows they should be careful here.

vovakulikov: Maybe this is ok cause the previous logic also has this flaw

vovakulikov (approved):
Approved, no worries about the size, changes are all common and straightforward, left one minor comment. Thanks!

client/web/src/enterprise/batches/batch-spec/execute/workspaces/WorkspaceDetails.tsx
dadlerj: Yeah, this is a good concern, but it's an acceptable risk here

PR #60774, opened 2024-02-28
Fix Cody PLG "Manage subscription" button

vdavid:
Fixes [this](https://sourcegraph.slack.com/archives/C05PC7AKFQV/p1709108448338019) problem where the "manage subscription" button on `/cody/manage` didn't link to the right page.

The problem was a missing `event.preventDefault()` call but I also noticed that on hovering the button, the wrong URL was displayed as the target, so fixed that, too.

## Test plan

Tested it locally.

### DIFF
--- a/client/web/src/cody/management/CodyManagementPage.tsx
+++ b/client/web/src/cody/management/CodyManagementPage.tsx
@@ -51,10 +51,6 @@ import { CHANGE_CODY_PLAN, USER_CODY_PLAN, USER_CODY_USAGE } from '../subscripti
 
 import styles from './CodyManagementPage.module.scss'
 
-
-
-
-
 interface CodyManagementPageProps {
     isSourcegraphDotCom: boolean
     authenticatedUser: AuthenticatedUser | null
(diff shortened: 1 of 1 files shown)


client/web/src/cody/management/CodyManagementPage.tsx
```diff
@@ -200,7 +200,9 @@ export const CodyManagementPage: React.FunctionComponent<CodyManagementPageProps
                                 <ButtonLink
                                     variant="primary"
                                     size="sm"
-                                    onClick={() => {
+                                    href={manageSubscriptionRedirectURL}
+                                    onClick={event => {
+                                        event.preventDefault()
                                         eventLogger.log(EventName.CODY_MANAGE_SUBSCRIPTION_CLICKED)
                                         window.location.href = manageSubscriptionRedirectURL
```
thenamankumar, line 207: We are also making a redirection here: `window.location.href = manageSubscriptionRedirectURL` may be we can remove this, given we are now using the href arg.

thenamankumar (approved):

client/web/src/cody/management/CodyManagementPage.tsx
vdavid: We surely want to do the logging, and I wanted to avoid a race condition by manually redirecting.

Also, for some weird reason, just having the `href` there wasn't enough, the button didn't do anything without the manual redirect. I didn't investigate it further.

vdavid:
@chrsmith Yes, never worked.

> Why is event.preventDefault() required? 
The event logging action and redirect were triggered, but then the browser-native behavior kicked in and a "redirect in-place" happened before our redirect could've taken place, so in effect, the button didn't redirect.

> Why are we specifying BOTH an href AND onclick field? 
Because if we don't have the href, the link in the status bar on hover is wrong. The href is only there to fix the status bar text. (I've tried to indicate this in the PR desc, apparently I could've been clearer.)

PR #60779, opened 2024-02-28
search: remove unused toZoektPatternNew

keegancsmith:
I think this was a leftover when we where experimenting with keyword search.

Test Plan: go test


stefanhengl (approved):
Thanks!

PR #60787, opened 2024-02-28
search: validate "content:" queries

keegancsmith:
We don't actually support "content:" queries. Instead how it works is that if "content:" is specified we adjust the type of the query to type:file. This commit adds some validation that we don't use this outside of type:file or the default type.

Test Plan: TODO

### DIFF
--- a/internal/search/query/BUILD.bazel
+++ b/internal/search/query/BUILD.bazel
@@ -27,6 +27,7 @@ go_library(
         "//internal/lazyregexp",
         "//internal/search/filter",
         "//internal/search/limits",
+        "//internal/search/result",
         "//lib/errors",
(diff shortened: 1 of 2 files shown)


jtibshirani:
Exploring the alternative option: how hard would it be to support `content` filters correctly? As a user I'd be really confused by the limitation "the query contains a content: field and the query is not type:file".

keegancsmith:
I agree we should do that. I think the easiest way forward for us in how we represent queries internally is to transform `foo` into `(content:foo or file:foo)`.

stefanhengl:
Going through my backlog. Should we close this?

keegancsmith:
Yeah lets close this out, I'll mention in the fixing content: filter issue a link to this PR.

PR #60793, opened 2024-02-29
sg: add flag to allow running sg ci bazel with current staged files

Strum355:
Instead of having to make a commit, just take the staged files (if using the --staged flag)

## Test plan

I ran it locally! : ) https://buildkite.com/sourcegraph/sourcegraph/builds/263512#018df461-13a2-4dd3-8c2c-164024720745


jamesmcnamara (approved):
This always gets me, I always end up just kicking off my task on main

PR #60795, opened 2024-02-29
rbac: add dotcom-only coles and productsubscription roles

bobheadxi:


bobheadxi:
* **#60796** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/60796?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#60795** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/60795?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/60795?utm_source=stack-comment).

BolajiOlajide (approved):
LGTM!

eseliger (approved):
Could we agree on the naming for the role here, so that I can make https://github.com/sourcegraph/sourcegraph/pull/60014 stack on top of this? 

The goal of that PR is to decouple license mgmt from site admin for UI access, but it seems similar enough to what you're doing here :)

bobheadxi:
> The goal of that PR is to decouple license mgmt from site admin for UI access, but it seems similar enough to what you're doing here :)

I would vote for "product subscriptions" and write == "can manage licenses". I think I'd like to make sure we help people remember that subscriptions as an entity exists, especially with longer-lived "subscription" traits like CG rate limits. WDYT?

eseliger:
Makes sense to me, resembles the entity name as well. Thanks for entertaining me :)

unknwon (approved):

PR #56866, opened 2023-09-20
worker/soap-cleaner: demote users after SOAP expires

bobheadxi:
- Expiry is now based on the age of the SOAP connection, not the user itself
- If a user has more than 1 external service account, we now demote that user so that they do not get site admin, and remove the external account connection

## Test plan

Integration tests

PR #60798, opened 2024-02-29
soap cleaner: guard all delete ops on whether the set is non-empty

bobheadxi:
This is the fix for an issue @michaellzc reported in https://sourcegraph.slack.com/archives/C05DWT4ANHH/p1709240944860449 - when `db.UserExternalAccounts().Delete(...)` is called with an empty ID set (i.e. the cleanup heuristic finds nobody that needs to be demoted/deleted), the method interprets that as a request to delete _all_ SOAP users, which is definitely not what we want. This regression was introduced in https://github.com/sourcegraph/sourcegraph/pull/56866

## Test plan

A new test case courtesy of @michaellzc


cmd/worker/internal/auth/sourcegraph_operator_cleaner.go
```diff
@@ -157,27 +157,33 @@ GROUP BY user_id
...
-	}
-	if err := h.db.UserExternalAccounts().Delete(ctx, database.ExternalAccountsDeleteOptions{
-		IDs:         deleteExternalAccountIDs,
-		ServiceType: auth.SourcegraphOperatorProviderType,
-		HardDelete:  true,
-	}); err != nil && !errcode.IsNotFound(err) {
-		return errors.Wrap(err, "remove SOAP accounts")
+	if len(deleteExternalAccountIDs) > 0 {
+		if err := h.db.UserExternalAccounts().Delete(ctx, database.ExternalAccountsDeleteOptions{
+			IDs:         deleteExternalAccountIDs,
```
bobheadxi, line 181: only this one is a likely bug, but I guarded the others as well just in case

unknwon (approved):

michaellzc (approved):

PR #60805, opened 2024-03-01
scim: Fix Entra integration

eseliger:
There seems to be some issue with `Resources: Null` in the test connection request. This fixes the flow.

## Test plan

Manually tried it out.


vdavid (approved):
Thanks, I've started creating this PR yesterday but than got side-tracked. The comment is pretty helpful as well.

Note: We should probably update "Azure AD" to "Entra ID" in our docs. I'll message @MaedahBatool about this.

PR #60808, opened 2024-03-02
web: Fix monaco completion/validation

fkling:


client/build-config/src/monaco-editor.ts
```diff
@@ -24,6 +24,7 @@ export const MONACO_LANGUAGES_AND_FEATURES: Required<
         'cursorUndo',
         'find',
         'format',
+        'gotoError', // needed for warning and error squiggles
```
fkling, line 27: Without this there would be a client side error. This is needed to show a tooltip when hovering over warnings and errors.

client/web/dev/esbuild/webmanifest.test.ts
fkling, line 1: This was moved and adapted from `manifest.test.ts`.

jamesmcnamara:
@fkling This looks great! Unfortunately `sg start enterprise-bazel` does not use the bazel web bundle (a good way to check is to look at the [command set definition](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@d613008c3f2b344ceb7066906e0ebee9cf5c55c1/-/blob/sg.config.yaml?L1105) for what is in `bazelCommands` and what is in `commands`).

I'll do a quick build with this to try and see if the bazel bundle works. What page / route can I nav to to test the editor?

fkling:
@jamesmcnamara You can go to `/settings`, which should redirect you to `/users/<user>/settings` and load the settings editor. Hovering over JSON keys there should show tooltips and there shouldn't be any errors in the browser console.

> Unfortunately `sg start enterprise-bazel` does not use the bazel web bundle

That's what I feared... as you can see there was some bazel specific logic in the manifest plugin that seemed unnecessary, it would be great to verify somehow that the files can actually be loaded in a complete bazel build.

jamesmcnamara:
I'm gonna try adding a target to run an esbuild server out of the bazel output tree; could be useful for other local flows.

jamesmcnamara:
@fkling I sank most of the day into trying to reproduce the bazel issue without any real luck. I even built a frontend image off of main and then used the https://github.com/sourcegraph/deploy-sourcegraph-docker config to bring a whole sourcegraph in docker and they all had code completion in the settings editor.

So potentially this is a dotcom/S2 specific issue? It's kind of strange

jamesmcnamara:
@fkling When I go to S2 or dotcom I see that it hits an error on `loadForeignModule` which is presumably the call from monaco to async-load the workers which don't exist. If I type anything, I get no completion.

When I build `main` either for dev or via bazel locally it, it still gives me an error on the Monaco file (there is an export token in that file that it doesn't know how to parse) but it does provide autocomplete.

I believe what we ship in production is the `frontend` [image](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@ad97b8b4ea473349fe47bc072e12b26da41d1fb7/-/blob/cmd/frontend/BUILD.bazel?L6-20) which is just a tiny static web server which depends on the client bundle.

I'd be happy to pair on this if you have any ideas but I'm unfortunately a little uninformed of the dark depths of how dotcom/s2 are released.

jamesmcnamara:
Good news is that it wasn't a weird S2 issue! Somehow along the way I managed to merge your changes into my local main so when I was building it through bazel it was actually building your code and not main. Thus it was actually confirming that this code works.

I added a target to sg.config `web-bazel` that just runs the UI through bazel so things like this can be confirmed in the future.

One thing that is still weird is that the working code still throws that above error from the monaco code (I assume because that file didn't get bundled into CJS?). It doesn't seem to be breaking anything though.

Here's a video JH took proving it works locally through Bazel.
https://github.com/sourcegraph/sourcegraph/assets/5115807/e250e28a-9673-443b-ba14-62e6229afadb

fkling:
@jamesmcnamara yay! Thank you for confirming 😃  I'll try to find out what's up with the foreign module error.

fkling:
@jamesmcnamara FWIW I don't see that foreign module error when running it locally (non-bazel). Do you see that only with bazel?
To avoid letting this get stale, maybe we can still merge it and continue investigate the other issue?

sg.config.yaml
eseliger, line 1: Are these changes related?

client/build-config/src/esbuild/workerPlugin.ts
```diff
@@ -18,7 +24,8 @@ async function buildWorker(
                 ...RXJS_RESOLUTIONS,
             }),
         ],
-        sourcemap: true,
+        // (inline) Sourcemaps drastically increase the size of the generated output
```
eseliger, line 27: could we enable them for dev?

eseliger (approved):
wow! thank you so much, that's definitely not been a simple fix. ty for treading through it :)

sg.config.yaml
fkling: @jamesmcnamara Added them for testing the bazel frontend build. I can remove them again if we just want to focus on fixing the issue.

PR #60810, opened 2024-03-02
Move dotcom check out of cmd/frontend

eseliger:
Simple refactor, moved the methods into internal/dotcom so that we don't import across package boundaries.

## Test plan

Simply moved code, existing test suites will find issues.


cmd/frontend/auth/user.go
```diff
@@ -178,7 +178,7 @@ func GetAndSaveUser(ctx context.Context, db database.DB, op GetAndSaveUserOp) (n
 			SourcegraphOperator: acct.AccountSpec.ServiceType == auth.SourcegraphOperatorProviderType,
 		}
 
-		if envvar.SourcegraphDotComMode() {
+		if dotcom.SourcegraphDotComMode() {
```
burmudar, line 181: This 🥇

burmudar (approved):

unknwon (approved):
Other than the name looks a bit redundant in terms of "dotcom/DotCom", but can live with it in the context of this repo 😂

PR #60819, opened 2024-03-04
syntax-highlighter: Add snapshot test for Groovy

varungandhi-src:
There was a Groovy issue reported recently.

https://github.com/sourcegraph/customer/issues/2858

The client code attempts to trigger code navigation based on syntax kinds set
by the syntax highlighter, but I noticed we didn't have any tests for the
syntax kinds for Groovy.

So let's at least add a test before fixing this. The fix is tracked in:
- https://github.com/sourcegraph/sourcegraph/issues/60820

This PR just adds a test to make the issue clear.

## Test plan

Snapshot tests


kritzcreek (approved):
LGTM

PR #60828, opened 2024-03-04
feat: extend graphql User with evaluateFeatureFlag operation

### LINKED ISSUE #59169: Use GraphQl to collect users and feature flags set

https://github.com/sourcegraph/accounts/issues/579 would like to use GraphQL to gather user information for all users that have a specific feature flag set, this functionality will be useful for them to help keep up with user lists.

#### Is your feature request related to a problem? If so, please describe.

Currently, GraphQL doesn't have a constructible query to return users with feature flags set.
#### Additional context

https://github.com/sourcegraph/accounts/issues/579 believe that getting user information such as `username`, `email`, `display_name`, `created_at`, `updated_at`, `deleted_at` from all the users that have a specific feature flag set would be advantageous information to monitor and maintain users.

bahrmichael:
This PR closes #59169 by extending the GraphQL `User` schema with the `evaluateFeatureFlag` function.

This function takes the name of a feature flag, and returns if this feature flag is active for the given user.

It's different to the existing root-level `evaluateFeatureFlag` as it does not evaluate for the calling user, but for any users that the user requests data for.

Example usage:

```
{
  users {
    nodes {
      username
      flagWithOverrides:evaluateFeatureFlag(flagName: "test-flag-secret")
      nonExistent:evaluateFeatureFlag(flagName: "non-existent-flag")
    }
  }
  currentUser {
    username
    evaluateFeatureFlag(flagName: "test-flag")
  }
}
```

This may yield (depending on your flags and overrides)

```
{
  "data": {
    "users": {
      "nodes": [
        {
          "username": "test",
          "flagWithOverrides": false,
          "nonExistent": null
        },
        {
          "username": "bahrmichael",
          "flagWithOverrides": true,
          "nonExistent": null
        },
        {
          "username": "sourcegraph",
          "flagWithOverrides": true,
          "nonExistent": null
        },
        {
          "username": "michael-2",
          "flagWithOverrides": false,
          "nonExistent": null
        }
      ]
    },
    "currentUser": {
      "username": "sourcegraph",
      "evaluateFeatureFlag": true
    }
  }
}
```

## Test plan

- Added new tests in `user_tests.go` to verify if the new field works from the `User` and `Users` schema.
- The SQL query function is covered by `feature_flags_test.go#testUserFlags`.
- Manual testing in the UI to verify if overrides work as expected (see example result pasted above).


bahrmichael:
The current exploration allows me to run a query like this

```
  organization(name:"michael-test") {
    name
    members {
      nodes {
        username
        featureFlagOverrides {
      		targetFlag {
        		... on FeatureFlagBoolean {
          		name
          		value
        		}
      		}
      	value
    		}
      }
    }
  }
```

which yields a response like this

```
    "organization": {
      "name": "michael-test",
      "members": {
        "nodes": [
          {
            "username": "sourcegraph",
            "featureFlagOverrides": [
              {
                "targetFlag": {
                  "name": "test-flag",
                  "value": true
                },
                "value": true
              }
            ]
          }
        ]
      }
    },
```

if the user belongs to an org.

Open questions:

- Do we need to consider cases where a user is not part of an organization?
- Are the requests mentioned in the target issue only intended for organizations (and therefore we can ignore the question above)?
- What's the difference between evaluating feature flags and reading feature flag overrides? I guess that evaluating takes the org default, org override, and user override into account, while the override just yields e.g. if a user has set an override.

bahrmichael:
Open questions:
- Documentation?
- Changelog?

camdencheek (approved):
Nice! Wasn't expecting this one to be quite so straightforward. Looks great to me, and thanks for the tests!

airamare01 (approved):
This is awesome! Thanks Michael

PR #60830, opened 2024-03-04
Add support for Claude 3 to Cody Gateway

philipp-spiess:
This PR adds a new `/claude-messages` endpoint on Cody Gateway. The idea is that ultimately all completions should go through the same API so that:

- Clients can bypass the SG instance for Chat too without having to implement provider specific code
- Our customers can update to new providers without having to update their SG instance 
- We migrate to a messages style API that we see all providers supporting now (avoiding the awkward prompt construction we were doing for Anthropic chats before)

For now this new API is implemented similarly to the other Cody Gateway providers and _only supports Claude models_ using the "new" `/messages` API. I will follow up with more features here shortly but want to get the first version out ASAP so we can enable Claude 3 for PLG soon.

In a later PR, I will also the same API interface to the SG instance. For enterprise, the plan is to have the SG instance be the proxy to Cody Gateway. BYOLLM can run the same provider-specific resolvers that Cody Gateway can run (I will follow up with an abstraction for this similar).

We also do not rewrite the result of the API right now (so the `unified` API returns Anthropic specific SSE events). This can be changed later by having the Clients already implement the future API so we can change the server without a breaking change: https://github.com/sourcegraph/cody/pull/3301#discussion_r1513198814

## Test plan

```bash
# For streaming requests
curl -vS -X POST http://localhost:9992/v1/completions/anthropic-messages -H 'Authorization: bearer <token>' -d '{"stream":true,"max_tokens":50, "model": "anthropic/claude-3-sonnet-20240229", "stop_sequences": [], "messages":[{"role":"system","content":[{"type": "text", "text": "You are Cody, an AI chat bot created by Sourcegraph."}]},{"role":"user","content":[{"type": "text", "text": "Who are you?"}]}]}' -H 'X-sourcegraph-feature: chat_completions'

# For non-streaming requests
curl -vS -X POST http://localhost:9992/v1/completions/anthropic-messages -H 'Authorization: bearer <token>' -d '{"stream":false,"max_tokens":50, "model": "anthropic/claude-3-sonnet-20240229", "stop_sequences": [], "messages":[{"role":"system","content":[{"type": "text", "text": "You are Cody, an AI chat bot created by Sourcegraph."}]},{"role":"user","content":[{"type": "text", "text": "Who are you?"}]}]}' -H 'X-sourcegraph-feature: chat_completions'
```


cmd/cody-gateway/internal/httpapi/completions/unified.go
```diff
@@ -0,0 +1,187 @@
...
+type unifiedRequest struct {
+	Messages      []unifiedMessage        `json:"messages,omitempty"`
+	Model         string                  `json:"model"`
+	MaxTokens     int32                   `json:"max_tokens,omitempty"`
+	Temperature   float32                 `json:"temperature,omitempty"`
+	TopP          float32                 `json:"top_p,omitempty"`
+	TopK          int32                   `json:"top_k,omitempty"`
+	Stream        bool                    `json:"stream,omitempty"`
+	StopSequences []string                `json:"stop_sequences,omitempty"`
+	Metadata      *unifiedRequestMetadata `json:"metadata,omitempty"`
```
philipp-spiess, line 69: This shouldn't be set by the client. Instead we should be able to get this from the access token somehow. This is also provider specific and shouldn't be in the unified request API

cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go
```diff
@@ -348,6 +348,10 @@ func allowedModels(scope types.CompletionsFeature, isProUser bool) []string {
 
 		if !isProUser {
 			return []string{
+				"anthropic/claude-3-sonnet-20240229",
+				"fireworks/" + fireworks.Mixtral8x7bInstruct,
```
philipp-spiess, line 352: This is TBD https://sourcegraph.slack.com/archives/C05AGQYD528/p1709638080344719 

might add this later (in which case free user wont have a way to select a different LLM)

cmd/cody-gateway/internal/httpapi/completions/upstream.go
```diff
@@ -130,8 +130,10 @@ func makeUpstreamHandler[ReqT UpstreamRequest](
 	// Convert allowedModels to the Cody Gateway configuration format with the
 	// provider as a prefix. This aligns with the models returned when we query
 	// for rate limits from actor sources.
-	for i := range allowedModels {
-		allowedModels[i] = fmt.Sprintf("%s/%s", upstreamName, allowedModels[i])
+	clonedAllowedModels := make([]string, len(allowedModels))
```
chwarwick, line 133: This looks unnecessary why don't we just pass in the list of models in the `{upstream}/{model}` format?

cmd/cody-gateway/internal/httpapi/completions/unified.go
```diff
@@ -0,0 +1,339 @@
...
+const anthropicMessagesAPIURL = "https://api.anthropic.com/v1/messages"
+
+// The unified API endpoint is a new general-purpose AI inference API inspired
+// by the OpenAI API. The idea is is that regardless of which model you want to
+// use, there's one _unified_ API to use for all Sourcegraph products.
+//
+// Sharing the same API interface between the SG instance and Cody Gateway also
+// allows clients to implement RFC888 and connect to Cody Gateway directly.
+//
+// Right now, unified API is only available for the Claude 3 model family and
```
chwarwick, line 36: This looks like a copy of the anthropic handler which is ok for now.  Do you plan on setting up the patterns for supporting multiple providers? Such as different decoders and response objects? Are those the same for Anthropic & OpenAI?

chwarwick (commented):
It may just be me but I typically try to separate dotcom changes and CodyGateway changes because they are deployed separately and to me that makes it clearer about how it needs to roll out and also revert if needed. 

I like the idea of a unified endpoint is that work you plan to follow up with shortly? If not we should make this an Anthropic v2 endpoint since it only supports that.

cmd/cody-gateway/internal/httpapi/completions/unified.go
```diff
@@ -0,0 +1,339 @@
...
+	eventLogger events.Logger,
+	rs limiter.RedisStore,
+	rateLimitNotifier notify.RateLimitNotifier,
+	httpClient httpcli.Doer,
+	config config.AnthropicConfig,
+	promptRecorder PromptRecorder,
+	autoFlushStreamingResponses bool,
+) (http.Handler, error) {
+	// Tokenizer only needs to be initialized once, and can be shared globally.
+	tokenizer, err := tokenizer.NewAnthropicClaudeTokenizer()
```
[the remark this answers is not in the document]
philipp-spiess: Okay I really need to step through this in the debugger then (had this planned anyways). 🤔 🤔  For some reason the io stream is already empty when we try to read it. I will compare it to the default anthropic endpoint.

cmd/cody-gateway/internal/httpapi/completions/unified.go
philipp-spiess: Yeah I want to basically copy the pattern we have on the Sourcegraph instance side for this (so that we can reuse the code between the BYOB implementation and Cody Gateway).

I want to look into this this week but the timeline I have planned is:

- Roll out this change on CG
- Roll out the changes on the SG instance side for dotcom so we give dotcom user access to the new models
- Roll out the client changes as part of the release tomorrow
- Follow up by implementing the unified API on the SG instance side
- Think about the abstraction for how to handle different providers so that we can reuse the code between CG and the SG instance

What do you think? :)

chwarwick: That makes sense to me, I just didn't want to leave a "unified" endpoint out there if it's only going to support anthropic for some time.  I think once you get the token usage working then it should be good to move forward.

philipp-spiess: @chwarwick Yeah this makes sense and I do worry about this too. I don't think it's much work to follow through but I’m not as productive with Go yet.

Maybe it s safer to call this `/anthropic2` for now (or `/anthropic-messages`?) and then I start introducing the "unified" name on the SG instance side first (where we need a new endpoint to `/complete` and `/code` anyways (because of the messages API) and we can unify both concepts there. 

Might feel more natural this way and it reduced the risk of us having a stale "unified" API lying around 😬

cmd/cody-gateway/internal/httpapi/completions/upstream.go
philipp-spiess: Agree this is odd but I'll likely not going to change this in this PR (want to avoid more risky changes)

cmd/cody-gateway/internal/httpapi/completions/unified.go
chwarwick: I don't see any harm in doing it and the routing is [setup](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@d31e1a7682ac9c620e5678bccf7a0d7e7390e88b/-/blob/cmd/cody-gateway/internal/httpapi/handler.go?L75:2-75:45) expecting versioning so I don't think having a v2 anthropic completions endpoint which exposes their new api would be harmful or stop us from later making a uniform api.

philipp-spiess: Hmm I think switching the top level version would be indicative of a greater change. I called this `/anthropic-messages` for now but still under the `v1` namespace since "it follows the same pattern" as the one before. Not sure if this makes sense, I can change that again but I think `v2` would be better for the unified endpoint later (where we want to migrate all providers to use this API instead)

cmd/cody-gateway/internal/httpapi/completions/anthropicmessages.go
```diff
@@ -0,0 +1,325 @@
...
+	"github.com/sourcegraph/sourcegraph/cmd/cody-gateway/internal/notify"
+	"github.com/sourcegraph/sourcegraph/cmd/cody-gateway/internal/tokenizer"
+	"github.com/sourcegraph/sourcegraph/internal/codygateway"
+	"github.com/sourcegraph/sourcegraph/internal/conf/conftypes"
+	"github.com/sourcegraph/sourcegraph/internal/httpcli"
+)
+
+const anthropicMessagesAPIURL = "https://api.anthropic.com/v1/messages"
+
+// This implements the newer `/messages` API by Anthropic
```
rafax, line 28: link to doc?

cmd/cody-gateway/internal/httpapi/completions/anthropicmessages.go
```diff
@@ -0,0 +1,325 @@
...
+func (a *AnthropicMessagesHandlerMethods) getRequestMetadata(body anthropicMessagesRequest) (model string, additionalMetadata map[string]any) {
+	return body.Model, map[string]any{
+		"stream":     body.Stream,
+		"max_tokens": body.MaxTokens,
+	}
+}
+func (a *AnthropicMessagesHandlerMethods) transformRequest(r *http.Request) {
+	r.Header.Set("Content-Type", "application/json")
+	r.Header.Set("X-API-Key", a.config.AccessToken)
+	r.Header.Set("anthropic-version", "2023-06-01")
```
rafax, line 216: If we're moving to messages, should we use a newer version?

cmd/cody-gateway/internal/httpapi/completions/anthropicmessages.go
```diff
@@ -0,0 +1,325 @@
...
+		}
+	}
+	if err := dec.Err(); err != nil {
+		logger.Error("failed to decode Anthropic streaming response", log.Error(err))
+	}
+
+	return promptUsage, completionUsage
+}
+
+func isFlaggedAnthropicMessagesRequest(tk *tokenizer.Tokenizer, r anthropicMessagesRequest, cfg config.AnthropicConfig) (*flaggingResult, error) {
```
rafax, line 284: Can you make the existing version of this generic+make it accept both anthropicRequest and anthropicMessagesRequest? Trying to synchronize flagging / blocking across 2 copies will be a maintenance nightmare

cmd/cody-gateway/internal/httpapi/completions/anthropicmessages.go
```diff
@@ -0,0 +1,325 @@
...
+		}
+
+		for _, c := range m.Content {
+			if c.Type == "text" {
+				sb.WriteString(c.Text)
+			}
+		}
+		sb.WriteString("\n\n")
+	}
+	return sb.String()
```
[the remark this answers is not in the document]
philipp-spiess: You mean as a way to avoid concatenating the string multiple times? We don't want to send this to the downstream API so I avoided adding it to `r`.

cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go
rafax: Now that this PR is Anthropic specific, I think we can remove this.

cmd/cody-gateway/internal/httpapi/completions/anthropicmessages.go
philipp-spiess: This version is the one they also mention in their API docs so I assume it's the most recent one? https://docs.anthropic.com/claude/reference/messages_post

rafax: Right, I went there to get the exact version and realized this is the last one, my bad, resolving

philipp-spiess:
Fixed the SSE stream parsing for usage data now. @rafax @chwarwick anyone willing to give this a final review?

Screenshot 2024-03-06 at 17 15 22

cmd/cody-gateway/internal/httpapi/completions/anthropicmessages.go
```diff
@@ -0,0 +1,304 @@
...
+			continue
+		}
+
+		var event anthropicMessagesStreamingResponse
+		if err := json.Unmarshal(data, &event); err != nil {
+			logger.Error("failed to decode event payload", log.Error(err), log.String("body", string(data)))
+			continue
+		}
+
+		fmt.Printf("%+v\n\n", event)
```
chwarwick, line 262:
```suggestion

```

cmd/cody-gateway/internal/httpapi/completions/anthropicmessages.go
```diff
@@ -0,0 +1,304 @@
...
+				completionUsage.tokens = event.Usage.OutputTokens
+			}
+		}
+	}
+	if err := dec.Err(); err != nil {
+		logger.Error("failed to decode Anthropic streaming response", log.Error(err))
+	}
+
+	fmt.Printf("promptUsage: %+v\n", promptUsage)
+	fmt.Printf("completionUsage: %+v\n", completionUsage)
```
chwarwick, line 284:
```suggestion	
```

cmd/cody-gateway/internal/httpapi/completions/unified.go
chwarwick: I'm ok leaving it as v1 since anthropic also refers to it as their v1 messages api.

chwarwick (approved):
Looks like you've got it all covered, just a few remaining fmt.Prints that need to be removed.

PR #60833, opened 2024-03-04
search: use repoPagerJob for diff/commit search

stefanhengl:
This is a refactor. Instead of resolving repos inside a diff/commit job, we pull the logic out and wrap the `commit.SearchJob` inside a `repoPagerJob`, just like we do for text search or symbol search.

Why?

To support diff/commit search in Search Jobs, the job must be wrapped in a top-level `repoPagerJob`.

Test plan:
- CI
- Manual testing: I tried a couple of diff/commit searches and compared results to s2.


internal/search/commit/commit.go
```diff
@@ -19,16 +19,14 @@ import (
...
 	"github.com/sourcegraph/sourcegraph/internal/search/result"
 	"github.com/sourcegraph/sourcegraph/internal/search/streaming"
-	"github.com/sourcegraph/sourcegraph/internal/trace"
 	"github.com/sourcegraph/sourcegraph/internal/types"
 )
 
 type SearchJob struct {
 	Query                gitprotocol.Node
-	RepoOpts             search.RepoOptions
+	Repos                []*search.RepositoryRevisions
```
stefanhengl, line 29: RepoRev might be a better a name here, but I am following the convention in `setRepos` in "repo_pager_job.go".

jtibshirani (approved):
Looks good to me! Just double-checking, is this covered by existing integration tests for diff/ commit search?

stefanhengl:
> Looks good to me! Just double-checking, is this covered by existing integration tests for diff/ commit search?

Yes, we have a few integration tests [here](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@33a202691ada7f3932dce973634e8a88d1bea9ae/-/blob/dev/gqltest/search_test.go?L627-654) and [here](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@33a202691ada7f3932dce973634e8a88d1bea9ae/-/blob/dev/gqltest/search_test.go?L1421-1440).

PR #60835, opened 2024-03-04
Block bad phrases for flagged requests

rafax:
Block bad phrases for flagged requests - requests that are not flagged are unaffected.
Also refactor passing different types of phrases to the flagging/blocking logic so that:

- we always pass and look for phrases, not regular expressions
- phrases are always passed lower-case
- minor clean-up of duplicate arguments

## Test plan

- unit tests
- e2e tests
- tested locally


cmd/cody-gateway/internal/httpapi/completions/anthropic.go
```diff
@@ -54,9 +55,9 @@ func isFlaggedAnthropicRequest(tk *tokenizer.Tokenizer, ar anthropicRequest, pro
 		reasons = append(reasons, "high_prompt_token_count")
 	}
 
-	if len(reasons) > 0 {
+	if len(reasons) > 0 { // request is flagged
 		blocked := false
-		if tokenCount > cfg.PromptTokenBlockingLimit || ar.MaxTokensToSample > int32(cfg.ResponseTokenBlockingLimit) {
+		if tokenCount > cfg.PromptTokenBlockingLimit || ar.MaxTokensToSample > int32(cfg.ResponseTokenBlockingLimit) || containsAny(strings.ToLower(ar.Prompt), cfg.BlockedPromptPatterns) {
```
chrsmith, line 60: Nit: This `if`-statement is long enough I'd suggest we try to find a way to make it look "shorter", or at least be more legible.

Perhaps we rename `containsAny` to `containsBlockedPatterns(cfg, ar.Prompt)`? That would reduce the "number of moving parts" here, and make the function easier to understand/test. What do you think?

cmd/cody-gateway/internal/httpapi/completions/anthropic.go
```diff
@@ -106,10 +116,15 @@ func NewAnthropicHandler(
...
 		return nil, err
 	}
-	promptRegexps := []*regexp.Regexp{}
+	var promptRegexps []*regexp.Regexp
+	var blockedPhrases []string
 	for _, pattern := range config.AllowedPromptPatterns {
 		promptRegexps = append(promptRegexps, regexp.MustCompile(pattern))
 	}
+	for i, pattern := range config.BlockedPromptPatterns {
+		config.BlockedPromptPatterns[i] = strings.ToLower(pattern)
```
chrsmith, line 125: Maybe this is worth calling out with a comment in the code, e.g. on line 123?

That we do the prompt check in a case insensitive fashion, and as an optimization just force the prompt and block patterns to be done _after_ calling `ToLower()` on the strings?

cmd/cody-gateway/internal/httpapi/completions/anthropic.go
```diff
@@ -194,6 +209,7 @@ type AnthropicHandlerMethods struct {
 	promptRegexps      []*regexp.Regexp
 	promptRecorder     PromptRecorder
 	config             config.AnthropicConfig
+	blockedPhrases     []string
```
chrsmith, line 212: Could you add a doc comment here explaining what `blockedPhrases` controls, and with an example of what a typical value would be? e.g. is this supposed to be "plain text" or will it accept a regular expression? If a search is "blocked", how does that surface to users? (400? 403? Does it count against usage quotas? Does it auto-ban the account? etc.)

cmd/cody-gateway/internal/httpapi/completions/anthropic_test.go
```diff
@@ -67,7 +67,29 @@ func TestIsFlaggedAnthropicRequest(t *testing.T) {
...
+
+	t.Run("low prompt token count and bad phrase", func(t *testing.T) {
+		cfgWithBadPhrase := &cfg
+		cfgWithBadPhrase.BlockedPromptPatterns = []string{"bad phrase"}
+		longPrompt := strings.Repeat("word ", 5)
+		ar := anthropicRequest{Model: "claude-2", Prompt: validPreamble + " " + longPrompt + "bad phrase"}
+		result, err := isFlaggedAnthropicRequest(tk, ar, []*regexp.Regexp{regexp.MustCompile(validPreamble)}, *cfgWithBadPhrase)
+		require.NoError(t, err)
+		// for now, we should not flag requests purely because of bad phrases
+		require.False(t, result.IsFlagged())
```
chrsmith, line 89: How is this unit test different from the previous one? Why would we _not_ flagged this, despite using the same "bad phrase" from the `BlockedPromptPatterns`?

Would it be correct to say:

```go
// Currently, we do not flag requests only because of bad phrases.
// It must _also_ be blocked for another reason. So in this case,
// since `longPrompt` isn't greater than `PromptTokenFlaggingLimit`,
// the request isn't flagged despite containing a bad phrase.
```

☝️ is too long, but is that the rational?

chrsmith (approved):
Some minor nit picks. Mostly around just adding doc comments to make it easier for people unfamiliar with the codebase to know how to use this new capability correctly.

cmd/cody-gateway/internal/httpapi/completions/anthropic.go
rafax: Fair, I plan to follow-up (tomorrow) with a refactor that will take all lower-casing / pre-processing to happen at Config parse time (instead of being scattered here and in other places).

chrsmith: > instead of being scattered here and in other places.

To be honest, I'm not so worried about when or where the processing happens. Just that from the code it's clear what the expectations are. So keeping it here when we create the new completions object seems fine, as long as there is a comment on the code and `containsAny`.

... but having this _modify_ the configuration object as a side effect is kinda sketchy. So moving that logic closer to Config parse time probably would make sense. (And have the doc comment on the `BlockedPromptPatterns` field contain something like "will always be lower-case" as that is an invariant we can now rely on.)

cmd/cody-gateway/internal/httpapi/completions/anthropic.go
rafax: Changed name, moved ToLower()

cmd/cody-gateway/internal/httpapi/completions/anthropic_test.go
rafax: This is rational - initially (in this PR), we only perform blocking if a given request was flagged for another reason to minimize false positives (the only requests affected by blocking are ones that are already affected by flagging and downstream mechanisms).

cmd/cody-gateway/internal/httpapi/completions/anthropic.go
rafax: Added a doc comment to AnthropicConfig

PR #60837, opened 2024-03-04
Remove dotcom specific gmail abuse check

evict:
This is being handled by accounts now so we can remove the code from Dotcom.

## Test plan
CI tests


evict:
> Thanks!
> 
> For the same reasoning,
> 
> https://github.com/sourcegraph/sourcegraph/blob/58971151348dde1365e1f23fd65f62e9d40f5727/internal/security/security.go#L67
> 
> can be deleted too?

Done! Thanks. Technically customers could be using it, but I don't see why they should.

PR #60838, opened 2024-03-04
Allow site-admins to view user events on Dotcom

evict:
This feature was introduced in 2021 when we still had private code on Dotcom. We no longer have private code on Dotcom and it can be useful to view user events when investigating an incident.

## Test plan
CI tests.


eseliger (approved):

unknwon (approved):
👍 

Would be good to also open up username change by site admin on dotcom too 😂 right now we need to access DB for that type of support requests.

eseliger:
@unknwon that makes me wonder, can we remove a bunch of dotcom checks as a result of this? The divergence of dotcom from other instances is often quite scary

unknwon:
> @unknwon that makes me wonder, can we remove a bunch of dotcom checks as a result of this? The divergence of dotcom from other instances is often quite scary 

Yes I believe we can. We never did clean up site-admin-related divergence.

PR #60840, opened 2024-03-04
tracer/exporters: use internalerrs loggers on OTEL gRPC exporter

bobheadxi:
In dotcom, the OTEL SDK reports:

```
traces export: rpc error: code = Internal desc = grpc: error while marshaling: string field contains invalid UTF-8
```

This isn't super helpful for narrowing down where these invalid attributes are coming from.

Luckily, we have some middleware providing some slightly more detailed diagnostics in `grpc/internalerrs`, including detailed logging for this exact problem - this hooks that middleware into the OTEL SDK to help us pin down the source of this issue.

## Test plan

```
sg start
```

Tried a `trace=1` search, nothing bad happened. This middleware is already used extensively elsewhere, and we have an escape hatch via `SRC_GRPC_INTERNAL_ERROR_LOGGING_ENABLED`. Confirmed there's no logspam from failed export due to lack of an active collector in local dev.

### DIFF
--- a/internal/tracer/oteldefaults/exporters/BUILD.bazel
+++ b/internal/tracer/oteldefaults/exporters/BUILD.bazel
@@ -10,6 +10,7 @@ go_library(
     importpath = "github.com/sourcegraph/sourcegraph/internal/tracer/oteldefaults/exporters",
     visibility = ["//:__subpackages__"],
     deps = [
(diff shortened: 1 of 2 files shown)


camdencheek (approved):
LGTM, thanks!

PR #4981, opened 2019-07-26
UI for JSON Schema config/settings

sqs:
WIP, see https://sourcegraph.slack.com/archives/C0C324C91/p1564135367019500 for video

PR #60842, opened 2024-03-04
dbconn/otelsql: run all string attributes through strings.ToValidUTF8

bobheadxi:
Combat issues we're seeing with non-UTF8 values in trace attributes breaking trace exports - at least [one sample](https://sourcegraph.slack.com/archives/C04MYFW01NV/p1709584216722099?thread_ts=1706107280.411819&cid=C04MYFW01NV) indicates that otelsql attributes are the likely issue. This _may_ be a regression from https://github.com/sourcegraph/sourcegraph/pull/4981. This change:

1. Runs all SQL args that get converted to strings through `strings.ToValidUTF8`
2. Customize how the query gets added as an attribute (to make sure it's also valid UTF8, and doesn't have a crazy length)
   1. Because queries now get truncated, this change also ensures that the inferred caller context (visible in Cloud SQL) also gets added to the span
3. Makes the truncation thresholds configurable and bumps the default to 240 - the previous threshold of 100 was a bit strict, and made the query attribute pretty useless

I also noticed that some types are not handled (at least one instance of `int32`) - I'm not sure where my docstring of valid types comes from, but I added the other basic numeric types just in case.

## Test plan

image

### DIFF
--- a/internal/database/dbconn/open.go
+++ b/internal/database/dbconn/open.go
@@ -262,6 +262,10 @@ func open(cfg *pgx.ConnConfig) (*sql.DB, error) {
 			OmitConnPrepare:      true,
 			OmitRows:             true,
 			OmitConnectorConnect: true,
+			// Do not include the otesql-generated 'db.statement' attribute so
+			// that we can add our own processed, truncated version of it in
+			// instrumentQuery(...).
+			DisableQuery: true,
 		}),
 		otelsql.WithAttributesGetter(argsAsAttributes),
 	)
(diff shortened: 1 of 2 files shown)


Strum355 (approved):
🙈

camdencheek:
> Makes the truncation thresholds configurable and bumps the default to 240

Doesn't google cloud limit it to 128 anyways?

bobheadxi:
> Doesn't google cloud limit it to 128 anyways?

Ah, good catch - let me use that as the default limit instead (it's still important we do the truncation ourselves to some degree though)

bobheadxi:
Oh actually @camdencheek only the key is 128, the _value_ of attributes is capped at 256: https://cloud.google.com/trace/docs/quotas, so the new default should be good. I'll add a docstring indicating this

internal/database/dbconn/open.go
```diff
@@ -366,29 +385,48 @@ func argsAsAttributes(ctx context.Context, _ otelsql.Method, _ string, args []dr
...
+// truncateStringValue should be used on all string attributes in the otelsql
+// instrumentation. It ensures the length of v is within argsAttributesValueLimit,
+// and ensures v is valid UTF8.
 func truncateStringValue(v string) string {
 	if len(v) > argsAttributesValueLimit {
-		return v[:argsAttributesValueLimit]
+		return strings.ToValidUTF8(v[:argsAttributesValueLimit], utf8Replace)
 	}
-	return v
+	return strings.ToValidUTF8(v, utf8Replace)
```
camdencheek, line 420: note that `utf8Replace` is 3 bytes, so it's possible to exceed the limit after replacement.

bobheadxi: Hm, I think in general the configurable cap has nothing to do with actual byte size

bobheadxi: Added a docstring indicating this is only a heuristic for now. Let's see what happens in dotcom with this change for now - I think Cloud Trace automatically truncates for us, and we can change the default/heuristic for truncation later

Issue #60605, opened 2024-02-19
search: panic in ChunkMatch.MatchedContent for fileContainsFilterJob


keegancsmith:
An interesting panic which has an off by 1. I am unsure how this can happen, but just clamping the offsets we use feels like a hack which hides the true bug.

Possible ideas:
- some sort of trailing newline we remove
- incorrect offset calculation at end of document?
- multibyte runes messing with calculations?

I haven't dived deep into any of these theories, posting this issue to track this likely rare occurance and hoping something pops out at someone.

https://console.cloud.google.com/errors/detail/CJiyg6P9hJmlZw;time=P30D?project=sourcegraph-dev

```
runtime error: slice bounds out of range [:10048] with length 10047

at github.com/sourcegraph/sourcegraph/internal/search/result.ChunkMatch.MatchedContent ( internal/search/result/file.go:261 )
at github.com/sourcegraph/sourcegraph/internal/search/job/jobutil.(*fileContainsFilterJob).filterChunk ( internal/search/job/jobutil/filter_file_contains.go:144 )
at github.com/sourcegraph/sourcegraph/internal/search/job/jobutil.(*fileContainsFilterJob).filterFileMatch ( internal/search/job/jobutil/filter_file_contains.go:129 )
at github.com/sourcegraph/sourcegraph/internal/search/job/jobutil.(*fileContainsFilterJob).filterEvent ( internal/search/job/jobutil/filter_file_contains.go:112 )
at github.com/sourcegraph/sourcegraph/internal/search/job/jobutil.(*fileContainsFilterJob).Run.func2 ( internal/search/job/jobutil/filter_file_contains.go:97 )
at github.com/sourcegraph/sourcegraph/internal/search/streaming.StreamFunc.Send ( internal/search/streaming/stream.go:24 )
at github.com/sourcegraph/sourcegraph/internal/search/job.(*observingStream).Send ( internal/search/job/observe.go:51 )
at github.com/sourcegraph/sourcegraph/internal/search/job.(*observingStream).Send ( internal/search/job/observe.go:51 )
at github.com/sourcegraph/sourcegraph/internal/search/zoekt.sendMatches ( internal/search/zoekt/indexed_search.go:432 )
at github.com/sourcegraph/sourcegraph/internal/search/zoekt.DoZoektSearchGlobal.func1 ( internal/search/zoekt/indexed_search.go:294 )
[...]

PR #60844, opened 2024-03-04
Search: protect against panic

camdencheek:
This is a band-aid in response to https://github.com/sourcegraph/sourcegraph/issues/60605. I still plan to follow up with a more complete fix, but wanted to fix the immediate problem so we 1) have a solution for customers who hit this, and 2) can get a patch release out since there is not another planned for 2 more weeks. 

## Test plan

Added unit test.

### DIFF
--- a/internal/search/result/file.go
+++ b/internal/search/result/file.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"unicode/utf8"
 
+	"github.com/sourcegraph/log"
 	"github.com/sourcegraph/sourcegraph/internal/api"
 	"github.com/sourcegraph/sourcegraph/internal/search/filter"
 	"github.com/sourcegraph/sourcegraph/internal/types"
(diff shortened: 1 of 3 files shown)


internal/search/result/file.go
```diff
@@ -247,13 +248,22 @@ type ChunkMatch struct {
...
+		// TODO(camdencheek): fix root cause(s).
+		// See https://github.com/sourcegraph/sourcegraph/issues/60605
+		start := rr.Start.Offset
+		end := rr.End.Offset
+		if start > len(cm.Content) || end > len(cm.Content) {
+			log.Scoped("search").Error("matched range out of bounds", log.String("content", cm.Content), log.Int("start", start), log.Int("end", end))
+			start = min(len(cm.Content), start)
+			end = min(len(cm.Content), end)
+		}
+		res = append(res, cm.Content[start:end])
```
jtibshirani, line 266: Just checking, is the calling code okay if the matched content ends up being empty?

jtibshirani (approved):

internal/search/result/file.go
camdencheek: Yep! We generally expect zero-length matches to be accepted across the stack. The places this is used are just using it to do a followup match, so it would be a match on an empty string, which is fine.

PR #60857, opened 2024-03-05
Log if one of main LLM providers is not enabled

rafax:
Adds a log message when an access token for one of our main providers (that we expect to be configured) is not present.

## Test plan

- tested locally
- will test in dev (where we face this issue)


arafatkatze (approved):
Thank You!!

PR #60858, opened 2024-03-05
svelte: Update top navigation to match S2 navigation

### LINKED ISSUE #60736: Adjust top navigation for S2
As part of enabling search pages by default on S2 we should update the top navigation to reflect the navigation as it is present in the React app on S2.

Optional, if it's easy to do:
- Take enterprise vs dotcom differences into account
- Take licensing information into account, if available

fkling:
Closes #60736 

This commit updates the top navigation to match what's available on S2. That means it doesn't take into account license differences or the dotcom environment, but I tried to implement in such a way that we can add support for this easily, by making the navigation derived from data.

To make styles consolidated I didn't reuse the existing Menu.svelte component but used the dropdown builder from @melt-ui directly.

NOTE: I kept the logic for marking external links, but I think we should remove that before Milestone 1.

## Test plan

Manual testing


client/web-sveltekit/src/routes/MainNavigationEntry.svelte
```diff
@@ -0,0 +1,111 @@
...
+    import { isCurrent, type NavigationEntry, type NavigationMenu } from './mainNavigation'
+    import { mdiChevronDown, mdiChevronUp } from '@mdi/js'
+    import { page } from '$app/stores'
+    import MainNavigationLink from './MainNavigationLink.svelte'
+
+    export let entry: NavigationEntry | NavigationMenu
+
+    function isNavigationEntry(entry: NavigationEntry | NavigationMenu): entry is NavigationEntry {
+        return entry && 'href' in entry
+    }
```
camdencheek, line 13: Super unrealted, but do you have any intuition for how to think about these type-narrowing function signatures? Every time I see one, I really struggle to understand what it is actually doing. Like, is `entry is NavigationEntry` a return type?

client/web-sveltekit/src/routes/MainNavigationEntry.svelte
```diff
@@ -0,0 +1,111 @@
+<script lang="ts">
+    import Icon from '$lib/Icon.svelte'
+    import { createDropdownMenu } from '@melt-ui/svelte'
```
camdencheek, line 3: Have we made any real decision `melt-ui`, or is this just "ease in absence of a decision"?

camdencheek (approved):

client/web-sveltekit/src/routes/MainNavigationEntry.svelte
fkling: Yeah, it basically is a return type. I guess it can be interpreted as `entry extends NavigationEntry ? true : false`. I don't know if that helps though 😅 Or you can read it as `x is a Y` and you are asserting that the value is one of the subtypes of the union. I don't know if there is a more intuitive way to think about that.

client/web-sveltekit/src/routes/MainNavigationEntry.svelte
fkling: I don't think we have yet. At the moment, in the absence of any other decision, it seems to be the best solution: It provides the necessary functionality while leaving complete control over styling. And I'm already using it elsewhere.

Issue #60732, opened 2024-02-23
Properly populate all parameters for streaming search


fkling:
I originally only set the minimal parameters to make streaming search work. We should audit them and pass the right values from URLs, local storage, feature flags, etc.
```[tasklist]
### Tasks
- [x] trace
- [ ] featureOverrides
```

PR #60871, opened 2024-03-05
Svelte: respect trace=1 query param

camdencheek:
This propagates the `trace=1` from the URL query to the search stream request.

I also looked into maintaining the `trace=1` when submitting a query (new navigation), but I wasn't sure what the best way to do that is. The React webapp maintains that query param across navigations, but it's not clear to me whether that's actually a good pattern. If I understand correctly, we would need to check and forward the `&trace=1` at every place we navigate, which would be pretty overwhelming. 

The trace setting is independent of search query params. Is there a good pattern for storing/fetching/forwarding pieces of state in the URL?

Contributes to [#60732](https://github.com/sourcegraph/sourcegraph/issues/60732)

## Test plan

Added `&trace=1` to a search URL and checked that the returned trace URL works.

### DIFF
--- a/client/web-sveltekit/src/lib/search/state.ts
+++ b/client/web-sveltekit/src/lib/search/state.ts
@@ -37,7 +37,7 @@ export class QueryState {
     private defaultQuery = ''
     private defaultSearchContext = 'global'
 
-    private constructor(public readonly options: Partial<Options>, public settings: QuerySettings) {}
+    private constructor(public readonly options: Partial<Options>, public settings: QuerySettings) { }
 
     public static init(options: Partial<Options>, settings: QuerySettings): QueryState {
         return new QueryState(options, settings)
@@ -151,18 +151,17 @@ export function queryStateStore(initial: Partial<Options> = {}, settings: QueryS
 export function getQueryURL(
     queryState: Pick<QueryState, 'searchMode' | 'query' | 'caseSensitive' | 'patternType' | 'searchContext'>,
     enforceCache = false
-): string {
-    const searchQueryParameter = buildSearchURLQuery(
+): URL {
+    let url = new URL('/search')
+    url.search = buildSearchURLQuery(
         queryState.query,
         queryState.patternType,
         queryState.caseSensitive,
         queryState.searchContext,
         queryState.searchMode
     )
-
-    let url = '/search?' + searchQueryParameter
     if (enforceCache) {
-        url += `&${USE_CLIENT_CACHE_QUERY_PARAMETER}`
+        url.searchParams.append(USE_CLIENT_CACHE_QUERY_PARAMETER, '1')
     }
     return url
 }
--- a/client/web-sveltekit/src/lib/search/state.ts
+++ b/client/web-sveltekit/src/lib/search/state.ts
@@ -37,7 +37,7 @@ export class QueryState {
     private defaultQuery = ''
     private defaultSearchContext = 'global'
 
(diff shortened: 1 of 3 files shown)


client/web-sveltekit/src/lib/search/state.ts
```diff
@@ -151,18 +151,17 @@ export function queryStateStore(initial: Partial<Options> = {}, settings: QueryS
 export function getQueryURL(
     queryState: Pick<QueryState, 'searchMode' | 'query' | 'caseSensitive' | 'patternType' | 'searchContext'>,
     enforceCache = false
-): string {
-    const searchQueryParameter = buildSearchURLQuery(
+): URL {
```
camdencheek, line 154: Tiny refactor: return a URL instead of a string so it's more strongly typed and easier to add params to correctly. `goto()` can take either a URL or a string, so callers do not need to be updated.

client/web-sveltekit/src/routes/search/+page.ts
```diff
@@ -122,7 +123,8 @@ export const load: PageLoad = ({ url, depends }) => {
             version: LATEST_VERSION,
             patternType,
             caseSensitive,
-            trace: '',
+            trace,
+            // TODO(@camdencheek): populate these from local storage
```
camdencheek, line 127: The primary way to set feature flags is with the dev popover, but that's not implemented in svelte yet. Setting feature flags from the UI is not critical functionality, so I think it's fine to skip this for now and come back to it after Milestone 1

vovakulikov (approved):

fkling:
@camdencheek

> The trace setting is independent of search query params. Is there a good pattern for storing/fetching/forwarding pieces of state in the URL?

I'm curious how we actually process `trace=1` in the app. I assume when it's present in the URL it's the server that does something with it. On the client we probably read it out and (eventually) pass it to requests made by `fetch`?
So if you where to click a link in the app and we do client side routing we wouldn't actually send any information to the server, and we'd just use the parameter again to pass it along with `fetch` requests.

In that case _maybe_ it's good enough to store that flag in memory and read it out where needed, although I admit that it wouldn't be obvious that tracing is still enabled after you navigated to a different page.
However, maybe we can just show something in the UI for that (e.g. a badge/button that indicates whether tracing is enabled and maybe even allows you to enable tracing).

---

But coming back to the original question, I guess one drawback of using `<a />` elements directly is that we can't add additional logic easily for "preserving" common parameters. However, maybe that's a good thing. We could still have a helper function `preserveParemeters` that you could pass a partial URL to and that would add the parameters to preserve, e.g. `<a href={preserveParameters('/...')}>...</a>`. It's not ideal because it's easy to forget, but on the other hand it's explicit and you know exactly what's going on.
Alternatively, but I don't really like that solution, we could introduce a custom component for rendering links, just like we have in React. But then that's another API to learn...

I assume we'll come up and refine our strategy as there is more need for doing this.

camdencheek:
> I'm curious how we actually process trace=1 in the app

Yes, ultimately it's the server that I really care about. We do also do some stuff with tracing in the webapp itself, but I'm honestly pretty unfamiliar how that works. 

In order to enable tracing on the server, we need to pass either `?trace=1` as a query param or set the `X-Sourcegraph-Should-Trace=1` header. Both have the same effect. This chooses to match the behavior of the existing webapp for search and add the `?trace=1` query param.

Also, I'm not at all tied to the `&trace=1` convention. We just need a simple way to enable or disable tracing when debugging in the webapp, and that's worked pretty well so far. We could also do something like add a temporary setting or something.

PR #60872, opened 2024-03-05
web: Trigger page refresh to load SvelteKit for supported/enabled routes

### LINKED ISSUE #60853: Enable seamless navigation from React to SvelteKit app
Currently it's not possible to "naturally" navigate from the React app to the SvelteKit app. I.e. if SvelteKit is enabled for the search page, then clicking on the 'Search' menu item in the React app will stay in the React app.

The navigation logic in the React app needs to be intercepted/extended to make a page refresh when a SvelteKit supported (and enabled) route is navigated to. Due to repository pages being kind of a "catch all" route, this likely requires comparing against all known top-level routes.

fkling:
Closes #60853 

This commit adds a new Link component that is aware of the `web-next` and `web-next-rollout` (renamed from `web-next-enabled` to make the purpose clearer) feature flags, and forces a page refresh when those feature flags are set and the path is supported/enabled.

The page refresh is done by simply rendering an `<AnchorLink />`, which bypasses React Router.

My only (unfounded?) concern is the impact on memory/performance of using `useFeatureFlag` inside the link component.

## Test plan

- Create Svelte build via `cd client/web-sveltekit/ && DEPLOY_TYPE=dev pnpm build`
- Start devserver with `sg start enterprise`
- Navigate to `/search` via top navigation -> stays in React app
- Enable `web-next-rollout` feature flag
- Navigate to `/search via top navigation -> loads Svelte app
- Commit search, navigate to repo page -> loads Svelte app
- Enable `web-next` feature flag
- Navigate to `/search via top navigation -> loads Svelte app
- Commit search, navigate to repo page -> stays in Svelte app

Note: For some reason loading files from the devserver is slow, which is why switching between the apps is slow.

https://github.com/sourcegraph/sourcegraph/assets/179026/b329ffae-d8a9-4bdd-b371-670b5481d7b1


client/web/src/sveltekit/SvelteKitAwareLink.tsx
```diff
@@ -0,0 +1,32 @@
...
+import isAbsoluteUrl from 'is-absolute-url'
+import { useFeatureFlag } from 'src/featureFlags/useFeatureFlag'
+
+import { RouterLink, type Link, AnchorLink } from '@sourcegraph/wildcard'
+
+import { isRolledOutRoute, isSupportedRoute } from './util'
+
+export const SvelteKitAwareLink = React.forwardRef(({ to, children, ...rest }, reference) => {
+    const [webNextEnabled] = useFeatureFlag('web-next-rollout', false)
+    const [webNext] = useFeatureFlag('web-next', false)
```
camdencheek, line 12: This should be heavily cached, yes?

camdencheek (approved):

client/web/src/sveltekit/SvelteKitAwareLink.tsx
fkling: Yes.

client/web/src/sveltekit/util.ts
```diff
@@ -1,26 +1,72 @@
...
+    route =>
+        // Remove the repository catch-all route because it matches everything
+        route !== PageRoutes.RepoContainer &&
+        // Remove index route because it will be a prefix of every pathname
+        route !== PageRoutes.Index
+)
+
+// All the routes that are supported by SvelteKit.
+// Also see `SVELTEKIT_SUPPORTED_REPO_PATHS` for the list of supported repo sub-pages.
+const supportedRoutes = new Set<string>([PageRoutes.Search, PageRoutes.RepoContainer])
```
vovakulikov, line 16: I will remember this but we probably should add some contribution doc about adding a new route to the svelte version, because as I understand we should keep this file with svelte routes in sync.

vovakulikov (approved):

client/web/src/sveltekit/util.ts
fkling: Yeah, I'll update the README. I'll probably have other things to add anyway.

PR #60877, opened 2024-03-05
Allow to get Blame information for a whole file

eseliger:
This allows to get blame for the entire file without specifying `1` and `100000` as the range.

## Test plan

Verified manually in my local instance.

### DIFF
--- a/cmd/frontend/graphqlbackend/git_blob.go
+++ b/cmd/frontend/graphqlbackend/git_blob.go
@@ -10,8 +10,8 @@ import (
 )
 
 type GitTreeEntryBlameArgs struct {
-	StartLine        int32
-	EndLine          int32
+	StartLine        *int32
+	EndLine          *int32
 	IgnoreWhitespace bool
 }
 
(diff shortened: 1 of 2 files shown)


cmd/frontend/graphqlbackend/git_blob.go
```diff
@@ -10,19 +10,26 @@ import (
...
-		},
+	}
+
+	if (args.StartLine == nil) != (args.EndLine == nil) {
+		return nil, errors.New("both startLine and endLine must be specified or neither")
+	}
+
+	if args.StartLine != nil && args.EndLine != nil {
+		opts.Range = &gitserver.BlameRange{
+			StartLine: int(*args.StartLine),
```
mmanela, line 30: Why do you need the `*args` here, doesn't go dereference automatically? (Like on line 21)

cmd/frontend/graphqlbackend/git_blob.go
```diff
@@ -10,19 +10,26 @@ import (
...
-			StartLine: int(args.StartLine),
-			EndLine:   int(args.EndLine),
-		},
+	}
+
+	if (args.StartLine == nil) != (args.EndLine == nil) {
+		return nil, errors.New("both startLine and endLine must be specified or neither")
+	}
+
+	if args.StartLine != nil && args.EndLine != nil {
```
mmanela, line 28: When they are both nil, the blame code just works correctly and takes the whole file?

cmd/frontend/graphqlbackend/git_blob.go
eseliger: `IgnoreWhitespace` in line 21 is of type `bool`, StartLine here is of type `*int32`, which is required by our GraphQL library to indicate something is nullable.

The syntax might be a bit confusing at first, we don't actually dereference `args` here, but instead the last object in the chain is dereferenced, so `StartLine`.

cmd/frontend/graphqlbackend/git_blob.go
eseliger: Yes, that is already the case for gitserver.Client.Blame, it was just not exposed in the GraphQL API layer 👍 The gitcli package has a test for empty input returns whole file blame: https://sourcegraph.com/-/editor?remote_url=github.com%2Fsourcegraph%2Fsourcegraph&branch=es%2Fgrpc-refs&file=cmd%2Fgitserver%2Finternal%2Fgit%2Fgitcli%2Fblame_test.go&editor=VSCode&version=2.2.16&start_row=44&start_col=0&end_row=48&end_col=0

mmanela (approved):

PR #60878, opened 2024-03-05
Search: avoid crashing on frontend panic

jtibshirani:
Because of a recent regression, frontend could panic when streaming back search
matches. These panics occur in a raw goroutine, which we didn't manually
propagate. So the panics caused frontend to crash.

This PR updates the relevant places to use `conc` objects, which properly
propagate panics. I did a "best effort" based on the original issue and didn't
audit all frontend search code.

Related to INC-258
Related to #60605

## Test plan

Happy path is covered by existing tests like `aggregate_test.go` and
`horizontal_test.go`. Manually repro'd the original issue #60605 with the
search `repo:sourcegraph/sourcegraph camden cheek file:contains.content(>\n)`
and checked frontend no longer crashes. Also manually tested code insights.

### DIFF
--- a/cmd/frontend/internal/compute/streaming/compute.go
+++ b/cmd/frontend/internal/compute/streaming/compute.go
@@ -3,6 +3,7 @@ package streaming
 import (
 	"context"
 
+	"github.com/sourcegraph/conc/pool"
 	"github.com/sourcegraph/conc/stream"
 	"github.com/sourcegraph/log"
 
@@ -90,23 +91,24 @@ func NewComputeStream(ctx context.Context, logger log.Logger, db database.DB, se
 		alert *search.Alert
 		err   error
 	}
-	final := make(chan finalResult, 1)
-	go func() {
-		defer close(final)
+
+	pl := pool.NewWithResults[finalResult]()
+	pl.Go(func() finalResult {
 		defer close(eventsC)
 		defer close(errorC)
 		defer s.Wait()
 
 		alert, err := searchClient.Execute(ctx, stream, inputs)
-		final <- finalResult{alert: alert, err: err}
-	}()
+		return finalResult{alert: alert, err: err}
+	})
 
 	return eventsC, func() (*search.Alert, error) {
 		computeErr := <-errorC
 		if computeErr != nil {
 			return nil, computeErr
 		}
-		f := <-final
-		return f.alert, f.err
+		results := pl.Wait()
+		r := results[0]
+		return r.alert, r.err
 	}
 }
--- a/internal/search/backend/aggregate.go
+++ b/internal/search/backend/aggregate.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
-	"github.com/sourcegraph/conc"
 	"github.com/sourcegraph/zoekt"
 )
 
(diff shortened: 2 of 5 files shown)


cmd/frontend/internal/compute/streaming/compute.go
```diff
@@ -90,23 +91,24 @@ func NewComputeStream(ctx context.Context, logger log.Logger, db database.DB, se
 		alert *search.Alert
 		err   error
 	}
-	final := make(chan finalResult, 1)
-	go func() {
-		defer close(final)
+
+	pl := pool.NewWithResults[finalResult]()
```
jtibshirani, line 95: Looking for feedback here -- is this an abuse of conc concepts to use pools and wait groups just for better panic handling?

internal/search/backend/horizontal.go
```diff
@@ -60,38 +61,40 @@ func (s *HorizontalSearcher) StreamSearch(ctx context.Context, q query.Q, opts *
 	var mu sync.Mutex
 	dedupper := dedupper{}
 
-	ch := make(chan error, len(clients))
-	for endpoint, c := range clients {
-		go func(endpoint string, c zoekt.Streamer) {
+	pl := pool.NewWithResults[error]()
+	for endpoint, client := range clients {
+		e := endpoint
```
jtibshirani, line 66: I believe we can remove this once we update to Go 1.22, which fixes this common "gotcha".

cmd/frontend/internal/compute/streaming/compute.go
camdencheek: Seems fine to me. `conc` also exposes a lower-level primitive [panics.Catcher](https://pkg.go.dev/github.com/sourcegraph/conc/panics#Catcher), but I think that would be a bit awkward to use directly here. There isn't currently a "oneshot" version of a pool (though maybe there should be).

internal/search/backend/aggregate.go
```diff
@@ -147,8 +148,10 @@ func newFlushCollectSender(opts *zoekt.SearchOptions, endpoints []string, maxSiz
 		timerCancel:  timerCancel}
 
 	// As an escape hatch, stop collecting after twice the FlushWallTime. This protects against
-	// cases where an endpoint stops being responsive so we never receive its results.
-	go func() {
+	// cases where an endpoint stops being responsive so we never receive its results. Even though
+	// it's a non-blocking goroutine, we use a WaitGroup to properly propagate panics.
```
camdencheek, line 152: `wg.Wait()` is never called, so this will not propagate any panics -- it will just ignore them. This goroutine is not well-scoped to its parent, so it's not clear to me how we would propagate the panic since there is no parent goroutine waiting for it to complete.

internal/search/backend/horizontal.go
```diff
@@ -60,38 +61,40 @@ func (s *HorizontalSearcher) StreamSearch(ctx context.Context, q query.Q, opts *
...
 	}
 
-	var errs errors.MultiError
-	for i := 0; i < cap(ch); i++ {
-		errs = errors.Append(errs, <-ch)
+	errs := pl.Wait()
+	var multiError errors.MultiError
+	for _, err := range errs {
+		multiError = errors.Append(multiError, err)
 	}
```
camdencheek, line 96: If you use `pool.New().WithErrors()` instead of `pool.NewWithResults[error]()`, `conc` will do this for you.

internal/search/backend/aggregate.go
jtibshirani: Oops, my mistake! I will refactor this (or maybe back it out as this change is non-exhaustive).

camdencheek: Sounds good! Approved pending a fix here. Feel free to re-request review if you think the refactor merits it

camdencheek (approved):

internal/search/backend/aggregate.go
jtibshirani: This would be a big refactor in sensitive code, so I just backed this out for now.

This PR is really just ... I learned about panic propagation and fixed some low-hanging issues, but unfortunately doesn't make a big dent in the overall problem.

camdencheek: Yep, sounds good. Panic propagation is a mess, and we mostly just didn't even attempt it before `conc`. _Sometimes_ we would log a message and ignore the panic...but that is also problematic.

internal/search/backend/horizontal.go
jtibshirani: Opened https://github.com/sourcegraph/sourcegraph/pull/60944

PR #60883, opened 2024-03-06
search jobs: support diff, commit and path search

stefanhengl:
This adds support for diff, commit and path results types to Search Jobs. We also improve the error messages to be more helpful.

Here are a couple of examples of search queries which are now supported in addition to content search

- diff search: `author:alice after:"1 year ago" foo type:diff`
- commit search: `bas type:commit`
- path search: `file:search.go type:path`

Notes:
- For interactive search, the default types are path, file and repo. For Search Jobs, the default types are path and file, because we don't support repo search yet.
- This change is fully backward compatible. The API does not change.

Test plan:
- New and updated unit tests
- I ran a couple diff/commit search jobs locally and compared the result counts with the counts returned by interactive search.


internal/search/job/jobutil/exhaustive_job.go
```diff
@@ -59,8 +52,49 @@ func NewExhaustive(inputs *search.Inputs) (Exhaustive, error) {
...
+
+	var planJob job.Job
+
+	if resultTypes.Has(result.TypeCommit | result.TypeDiff) {
+		_, _, own := isOwnershipSearch(b)
+		diff := resultTypes.Has(result.TypeDiff)
+		// Follows the logic of interactive search, see
+		// https://github.com/sourcegraph/sourcegraph/pull/35741
+		repoOptionsCopy := repoOptions
+		repoOptionsCopy.OnlyCloned = true
```
mmanela, line 70: Is a repo not being cloned just a point in time situation? How will a user know why repos were skipped?

stefanhengl: Yes, that should be a temporary state.

We could go either way here TBH, but it feels ok to mirror what interactive search does for consistency. The user can find out which exact repos-revision pairs were searched by checking the logs (downloadable from the UI). This fits well with the compliance use-case because we state exactly what we searched.

If we set `OnlyCloned = false`, search jobs on large instances might fail more often and the user has to look at the logs to find out why. I would feel better about this option if we had a way of reporting partial success to the user.

mmanela: What happens if onlyCloned is false? Will we use searcher or something like that to pull the info?

jtibshirani: It seems Search Jobs has a different target use case than interactive search, as it's much more focused on compliance and it'd be really bad if a user missed the fact we skipped over certain repos. So ideally our defaults would reflect that, and we'd loudly surface errors/ missing data instead of focusing on "best effort" as we usually do.

Am I understanding right that we'd currently fail the entire job if one repo search errors? Then your current approach makes sense to me. But big +1 to adding support for partial results so we can clearly surface errors to users.

internal/search/job/jobutil/exhaustive_job.go
```diff
@@ -59,8 +52,49 @@ func NewExhaustive(inputs *search.Inputs) (Exhaustive, error) {
...
+
+		// We can probably support higher limits here if we need to. This is the same
+		// limit we use for interactive search. The assumption is that the limiting
+		// factor for diff/commit search is usually time and not the number of results.
+		// We should revisit this if we see that we are hitting the limit often.
+		fileMatchLimit := int32(computeFileMatchLimit(b, inputs.DefaultLimit()))
```
jtibshirani, line 76: Would it make sense to use the super high limit `DefaultMaxSearchResultsExhaustive` here for consistency?

jtibshirani (commented):
This looks good to me, I left some small comments/ questions as I learn how this works more deeply.

internal/search/job/jobutil/exhaustive_job.go
mmanela: But is that what this setting will do? will onlyCloned just silently filter out non-indexed items? 
And oppositely, if we did onlyCloned=false, would it just work but be slow?

jtibshirani: Ah sorry, missed your question. My understanding: `onlyCloned=true` will silently skip over non-indexed repos. If `onlyCloned=false`, the search will fail. Searcher is not able to help here, since we don't even have the repo available within Sourcegraph.

stefanhengl:
> Am I understanding right that we'd currently fail the entire job if one repo search errors?

Yes

internal/search/job/jobutil/exhaustive_job.go
stefanhengl: Great pointer. Turns out we already use `DefaultMaxSearchResultsExhaustive` because the protocol is `exhaustive`. I removed the wrong comment and simplified the limit calculation: In our case, because we don't support file predicates or "own" searches, `computeFileMatchLimit` always ends up calling `b.MaxResults` anyway.

While reading the code I noticed we don't catch all "own" search during validation. I will fix that in another PR.

jtibshirani (approved):

PR #60889, opened 2024-03-06
search: update patternType suggestions

stefanhengl:
With this change we only show "structural" as suggestion for pattern type if structural search is enabled.

structural search **enabled**
Screenshot 2024-03-06 at 14 40 59

structural search **disabled**
Screenshot 2024-03-06 at 14 39 51

Test plan:
- updated unit tests
- manual testing


client/shared/src/search/query/filters.ts
```diff
@@ -271,9 +271,19 @@ export const FILTERS: Record<NegatableFilter, NegatableFilterDefinition> &
         placeholder: '"content"',
     },
     [FilterType.patterntype]: {
-        discreteValues: () =>
-            ['keyword', 'literal', 'regexp', 'standard', 'structural'].map(value => ({ label: value })),
-        description: 'The pattern type (standard, keyword, regexp, literal) in use',
+        discreteValues: () => {
+            const patternTypes = ['keyword', 'literal', 'regexp', 'standard']
+            if (typeof window === 'undefined' || window.context?.experimentalFeatures?.structuralSearch === 'enabled') {
```
stefanhengl, line 276: `window` is not available for some of tests hitting this code path, hence the check for `typeof window === 'undefined'`

client/shared/src/search/query/diagnostics.test.ts
```diff
@@ -608,6 +608,15 @@ describe('getDiagnostics()', () => {
     })
 
     describe('structural search and type: filter', () => {
```
stefanhengl, line 610: These tests assume structural search to be enabled

mmanela (approved):

jtibshirani (approved):

PR #60891, opened 2024-03-06
Record pattern that was blocked

rafax:
Follow-up to https://github.com/sourcegraph/sourcegraph/pull/60835 recording the pattern that we blocked

## Test plan

- tested locally
- e2e tests


vdavid (approved):
Nice, it all looks good.

PR #60897, opened 2024-03-06
Bahrmichael/playwright playground

bahrmichael:


bahrmichael:
Progress update:

- Achieved a proof of concept on Thursday last week: https://sourcegraph.slack.com/archives/C05D629GWJK/p1710422190795459
- No progress on Friday and Monday due to PTO
- Had an issue with ordered execution of tasks, but that doesn't seem to be a problem at all now. Instead I had to make sure both tasks escape the bazel sandbox so they can use the same browser.
- I have been able to run a playwright test against my local instance of sourcegraph, and produce errors and screenshots. Bazel and the tasks used make it a bit difficult to debug tests, so I'm now putting more effort into making sure we can run playwright without problems locally. There are dependency problems from trying out a lot of things when getting Playwright to run in bazel that I now need to resolve.

bahrmichael:
With [1f5f2de](https://github.com/sourcegraph/sourcegraph/pull/60897/commits/1f5f2deb227bc38660569e57aa6d59a549de7d73) I'm now able to run tests locally and get the expected results. The tests are however failing with playwright, because it seems that the bazel build doesn't respect the playwright config. This is indicated by bazel based tests taking 30s (default timeout; configure timeout is 5s) and complaining about an invalid URL (tests without bazel properly pick up the baseURL from the playwright config).

bahrmichael:
More progress! I've been able to run use the right config in playwright, and also learned about how environment variables are used in bazel. When running a command like the one below, we get an error locally saying that this URL is not accessible (if we don't have a server running there):

```
bazel test "//client/playwright:*"  --action_env=SOURCEGRAPH_BASE_URL="http://localhost:7080"
```

This variable is set in sg.config.yml, and the next commit here should show if that variable is populated for our test.

---

An additional topic that came up when talking with @fkling is that for svelte we want to start our own dev server (see playwright config in client/web-sveltekit) instead of using a server that's already running at another port (see approach above). I'll start working on that tomorrow.

client/web-sveltekit/src/routes/search/page.spec.ts
```diff
@@ -1,49 +1,133 @@
...
+test.describe('page.spec.ts', () => {
+
+    let browser: Browser;
+    let page: Page;
+    let polly: Polly;
+    let server: PollyServer;
+
+    test.beforeEach(async () => {
+
+        browser = await chromium.launch();
```
bahrmichael, line 18: Your IDE may complain that the typing is incompatible with playwright/test versions 1.40.0 and 1.40.2 being incompatible. I had no problems running this, and can look into dependency conflicts later.

PR #60900, opened 2024-03-06
Add RealIP field to request client

evict:
This PR introduces the `RealIP` field for accurate client IP usage behind Cloudflare proxies. This value can be used in scenarios where real client IPs are required, such as rate limiting.

## Test plan
CI tests, added new tests.


internal/completions/httpapi/limiter.go
```diff
@@ -63,20 +64,10 @@ func (r *rateLimiter) TryAcquire(ctx context.Context) (err error) {
...
-			ip = req.IP
-			// Note: ForwardedFor header in general can be spoofed. For
-			// Sourcegraph.com we use a trusted value for this so this is a
-			// reliable value to rate limit with.
-			if req.ForwardedFor != "" {
-				ip = req.ForwardedFor
-			}
-		}
-		if ip == "" {
+		if req.IP == "" {
```
eseliger, line 67: previously we checked that `req` is not nil, can this cause a nil panic now?

internal/requestclient/http.go
```diff
@@ -106,10 +110,16 @@ func httpMiddleware(next http.Handler, external bool) http.Handler {
...
+			// Set the real IP of the client to the connecting IP, if using Cloudflare.
+			if forwardedFor != "" {
+				realIp = forwardedFor
+			}
 		}
 
 		ctxWithClient := WithClient(req.Context(), &Client{
-			IP:           strings.Split(req.RemoteAddr, ":")[0],
+			IP:           remoteAddr,
+			RealIP:       realIp,
```
bobheadxi, line 122: Calling this `realIP` seems strange when the heuristic is only "real" in CF - why not just use `forwardedFor` in dotcom?

bobheadxi: This is simlar to `CF-IPCountry` - instead of a new `realCountry`, we just know that the inferred country is "real" in dotcom/cloud, and "not necessarily real" everywhere else. In this case, we know that `forwardedFor` is "real" in dotcom/cloud, and "not necessarily real" everywhere else

bobheadxi, line 122: Assuming there's a strong case for a new field `RealIP`, we should make this always empty outside of the CF case and omitted by default IMO, to avoid providing a misleading value

internal/requestclient/http.go
```diff
@@ -106,10 +110,16 @@ func httpMiddleware(next http.Handler, external bool) http.Handler {
...
 			ForwardedFor: req.Header.Get(headerKeyForwardedFor),
```
bobheadxi, line 123: Hm I think there's a bug here, should this use the parsed `forwardedFor` variable? 🤔

internal/requestclient/http.go
keegancsmith: Yeah I agree with this point of view.

If we really need a new field, I would call it something like `TrustedIP`. However, I suspect this work is mostly for dotcom so we should probably use similiar logic that we did for forwardedFor. On dotcom we use the CF fields, for other instances we have more trust in the clients w.r.t. ratelimiting/etc.

eseliger (commented):
is this still relevant?

evict:
> is this still relevant?

I will close the issue. It's not a hard requirement, nor a priority right now. We can revisit later.

Issue #60867, opened 2024-03-05
Directory Commits Page Not Accessible


gabtorre:
- **Sourcegraph version:**  5.3

The commits panel, which was previously available in the directory view, has been removed in version 5.3. This change has resulted in users being unable to navigate to the commits page through the UI. The only workaround currently available is manually editing the URL by replacing `tree` with `commits`.

#### Steps to reproduce:

1. Visit a directory page for a repo.

https://sourcegraph.com/github.com/sourcegraph/sourcegraph@87a310bc060c3ca8de42be579b8c53a5ad8a9b71/-/tree/cmd/cody-gateway/internal/httpapi/requestlogger

#### Expected behavior:
`Show all` button to access the directory commits.

<img width="1237" alt="Screenshot 2024-03-05 at 6 00 26 PM" src="https://github.com/sourcegraph/sourcegraph/assets/69164745/4dd5be98-5404-4aa9-a1b1-d08c45b9c1dd">

#### Actual behavior:
No direct way to view all the commits.
<img width="1231" alt="Screenshot 2024-03-05 at 6 00 36 PM" src="https://github.com/sourcegraph/sourcegraph/assets/69164745/cdd969e6-30b5-44e4-afab-3de52703d2db">

Reported by: https://github.com/sourcegraph/accounts/issues/564

If you would like immediate help on this, please email support@sourcegraph.com (you can still create the issue, but there are no [SLAs](https://sourcegraph.com/support/) on issues like there are for support requests).

PR #60909, opened 2024-03-07
Add link to all commits from tree page

gabtorre:
Closes https://github.com/sourcegraph/sourcegraph/issues/60867

This PR adds a link to the page showing all commits like the repo root page.

## Test plan

Manually tested

Screenshot 2024-03-13 at 6 38 27 PM


jdm-twtr:
I think this is a reasonable stop-gap to the issue! Putting my UX hat on now :) the placement of "show all commits" does not feel like it belongs in the directory table. I'd imagine a standalone button centered below the file list, in place of where the previous commit history table was. Users familiar with Sourcegraph will be looking there for the commit history, and find the button instead, which is less drastic of a change than placing a text link in an adjacent, arguably unrelated table.

Just my 2 cents.

fkling:
Sorry, this slipped through the cracks... there seem to be a lot of unrelated changes in this PR, can you rebase?

My gut reaction is that the link feels a bit out of place in this view. Although I'm not a big fan of this either, what do you think about adding a link/button similar to one the one the repo homepage:



That would at least make things consistent.

gabtorre:
That makes sense. Would something like this work better? @fkling 

Screenshot 2024-03-11 at 4 06 39 PM

fkling:
@gabtorre Yes!

peterguy:
@gabtorre I don't see any `Commits` button when running it locally. Maybe I did something wrong; you can audit my steps to make sure:
- `git status` to ensure no outstanding changes
- `git switch main` and `git pull` to update my local copy
- `sg update` and `sg setup` to ensure I have the latest `sg`
- `sg start`
- Launch [https://sourcegraph.test:3443/github.com/sourcegraph/sourcegraph/-/tree/client/eslint-plugin-wildcard](https://sourcegraph.test:3443/github.com/sourcegraph/sourcegraph/-/tree/client/eslint-plugin-wildcard) in Chrome

I also don't see any ownership info; not sure if that's an indicator of something broken on my end or not.

screen capture of the folder view

gabtorre:
I just pushed the changes and updated the description, sorry about that! @peterguy

client/web/src/repo/tree/TreePageContent.tsx
```diff
@@ -14,7 +14,18 @@ import { SearchPatternType, type TreeFields } from '@sourcegraph/shared/src/grap
 import type { PlatformContextProps } from '@sourcegraph/shared/src/platform/context'
 import type { TelemetryProps } from '@sourcegraph/shared/src/telemetry/telemetryService'
 import { buildSearchURLQuery } from '@sourcegraph/shared/src/util/url'
-import { Badge, ButtonLink, Card, CardHeader, Icon, Link, Text, Tooltip } from '@sourcegraph/wildcard'
+import {
+    Badge,
+    Button,
+    ButtonLink,
+    ButtonGroup,
```
peterguy, line 21: Don't need `ButtonGroup` if it's only one button.

client/web/src/repo/tree/TreePageContent.tsx
```diff
@@ -196,7 +227,14 @@ export const TreePageContent: React.FunctionComponent<React.PropsWithChildren<Tr
                     )}
                 </section>
             )}
-            <section className={classNames('test-tree-entries container mb-3 px-0', styles.section)}>
+
+            <section
+                className={classNames(
+                    'test-tree-entries container mb-3 px-0',
+                    styles.section,
+                    !readmeEntry ? 'mt-3' : undefined
```
peterguy, line 235: What's this do? I can't tell any difference with it present or absent.

gabtorre: It adds a top margin if there's no readme section

Screenshot 2024-03-13 at 8 03 28 PM

peterguy:
Looks great, thanks, @gabtorre!

peterguy (approved):
Looks good!

client/web/src/repo/tree/TreePageContent.tsx
```diff
@@ -177,6 +178,23 @@ export const TreePageContent: React.FunctionComponent<React.PropsWithChildren<Tr
 
     return (
         <>
+            {!isRoot && (
+                <div className={menuStyles.menu}>
+                    <Tooltip content="Git commits">
+                        <Button
+                            className="flex-shrink-0"
+                            to={`/${encodeURIPathComponent(repo.name)}/-/commits/${filePath}`}
```
fkling, line 186: The repo homepage has a bug that the commits link doesn't respect the currently selected revision. We couldn't make the same mistake here. I.e. the link should preserve the revision that's present in the current URL.

gabtorre: I'll follow up on the repo homepage

peterguy:
@gabtorre I double-checked your modification to include the revision in the link to the commits. Works well for `branches`, `tags`, and `commits`. Looks good!

Issue #60730, opened 2024-02-23
Add support for global alerts


fkling:
The alerts shown above the top navigation.

PR #60910, opened 2024-03-07
Add basic global notification UI

vovakulikov:
Closes https://github.com/sourcegraph/sourcegraph/issues/60730

## Todo
- [x] Add common/basic global notification UI
- [x] Fix top-down class passing (`:global` selector problem)
- [x] Add site flags fetching to the global notification 
 
 
Screenshot 2024-03-11 at 14 44 34

## Test plan
- Check that global notification renders common notification (dev-related about staging, notification from site admin)
- Check that dismissible notifications hide when you close them with the hide button
- Check markdown notification with arbitrary content (check that the link within markdown leads to the right URL)


client/web-sveltekit/src/lib/global-notifications/Markdown.svelte
```diff
@@ -0,0 +1,519 @@
+<script lang="ts">
+    export let dangerousInnerHTML: string = ''
+</script>
+
+{#if dangerousInnerHTML}
+    <div class='markdown'>
+        {@html dangerousInnerHTML}
+    </div>
+{/if}
```
fkling, line 9: I've never understood why our Markdown components don't actually parse the markdown. IMO that makes more sense.

I'd do

```suggestion
<script lang="ts">
    import {renderMarkdown} from '$lib/common'
    export let content: string
</script>

{#if content}
    <div class='markdown'>
        {@html renderMarkdown(content)}
    </div>
{/if}
```

`content` is up for debate but I wouldn't call it `dangerousInnerHTML`.

client/web-sveltekit/src/lib/global-notifications/GlobalNotifications.svelte
```diff
@@ -0,0 +1,78 @@
+<script lang="ts">
+    import {renderMarkdown} from '@sourcegraph/common'
```
fkling, line 2:
```suggestion
    import {renderMarkdown} from '$lib/common'
```

`renderMarkdown` is already exported from `$lib/common`.

client/web-sveltekit/src/lib/global-notifications/DismissibleAlert.svelte
```diff
@@ -0,0 +1,75 @@
...
+<script lang="ts">
+    import {mdiClose} from '@mdi/js'
+    import Icon from '$lib/Icon.svelte'
+    import {Alert, Button} from '$lib/wildcard'
+
+    export let variant: 'info' | 'danger'
+    export let partialStorageKey: string
+
+    let className = '';
+    export {className as class};
```
fkling, line 26: See my other comment. I'd prefer it if we can avoid this.

client/web-sveltekit/src/lib/global-notifications/GlobalNotifications.svelte
```diff
@@ -0,0 +1,161 @@
+<script lang="ts" context="module">
+    import {AlertType} from '$root/client/shared/src/graphql-operations';
+    import {formatDistanceStrict, isAfter} from 'date-fns';
+
+    type PossibleAlertVariation = 'info' | 'warning' | 'danger'
```
vovakulikov, line 5: I could have this union string in the alert (or even have an enum for this) but not sure why we don't do it in other places, I like enums but for some reason, we don't use them much.

client/web-sveltekit/src/lib/global-notifications/components/DismissibleAlert.svelte
```diff
@@ -0,0 +1,84 @@
...
+    </Alert>
+{/if}
+
+<style lang="scss">
+  .root {
+    display: flex;
+    align-items: center;
+    padding-right: 0.5rem;
+
+      &:last-child {
```
vovakulikov, line 69: I removed class overriding over the alert component but I don't think even this style works here, probably because root binds with non-HTML element Alert. Any ideas (except an additional wrapper) on how to solve this without classes overriding

client/web-sveltekit/src/routes/+layout.svelte
```diff
@@ -90,6 +91,7 @@
 <svelte:body use:classNames={$scrollAll ? '' : 'overflowHidden'} />
 
 <InfoBanner />
+<GlobalNotification/>
```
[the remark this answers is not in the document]
vovakulikov: Since notifications component takes not one but multiple fields from settings I read settings from the store, but if you strongly feel that props are better I can switch.

client/web-sveltekit/src/lib/global-notifications/GlobalNotifications.svelte
fkling: I think for an input/prop, a string union is more ergonomic to use. E.g. `type="info"` is easier to write (especially with autocompletion) than e.g. `type={AlertType.Info}`.

client/web-sveltekit/src/lib/global-notifications/components/DismissibleAlert.svelte
fkling: What should actually happen here? The last `<DismissableAlert />` should not have a border?

fkling: To me it seems to be the responsibility of the caller (i.e. the component that renders the dismissible alerts) to fit them properly (UI-wise) into the context. Without any context there doesn't seem to be an obvious reason why a dismissable alert that happens to be the last child of an element shouldn't have a border.

vovakulikov: Correct, this is what I had in the first version but there was a problem passing classes too, to control this style (last element border) you have to pass a CSS class to the root element (something that we're fighting against at the moment) at the same time I'm trying to avoid wrappers as much as possible. 

This made me think that okay maybe in this case where dismissible alerts are used only within global notification usage where we know it shouldn't have a border if it's the last element in the list it's okay. But even this doesn't work because we're passing classes to Alert component.

vovakulikov: I'm gonna skip this problem for a while because the last element border is actually ok and it's not a blocker. but this problem with composition styles between consumer and reusable components bothers me more and more. Need to think and maybe make a list of exceptions when CSS modules make more sense

client/web-sveltekit/src/lib/global-notifications/GlobalNotifications.svelte
```diff
@@ -0,0 +1,136 @@
...
+                    ? 'expired ' + formatRelativeExpirationDate(expiresAt) // 'Expired two months ago'
+                    : 'will expire in ' + formatDistanceStrict(expiresAt, Date.now())
+            }.&nbsp;
+            <a href="/site-admin/license">Renew now</a>
+            &nbsp;or&nbsp;
+            <a href="https://sourcegraph.com/contact">contact Sourcegraph</a>
+        </DismissibleAlert>
+    {/if}
+
+    {#if process.env.SOURCEGRAPH_API_URL}
```
vovakulikov, line 113: @fkling btw, forgot to ask about this, does vite (or maybe it's related to the Svelte compiler) do anything special with process vars? I found that even if I specify this variable this still has no value in a built bundle. In webpack we proxy process variables with plugin I think.

vovakulikov:
@fkling fixed by comments above, check whenever you have a minute, 😉

client/web-sveltekit/src/lib/global-notifications/components/DismissibleAlert.svelte
fkling: I don't think the problem would be solved by CSS modules, unless you mean that different components would use the same CSS module (i.e. DismissableAlert would _not_ render Alert but its own elements and use the same CSS module as Alert).

---

You could add a `data-alert` prop to the element rendered by `Alert` and then in `GlobalNotifications.svelte` you do

```css
.root :global([data-alert]):last-child {
   border-bottom: none;
}
```

or something like that.

client/web-sveltekit/src/lib/global-notifications/GlobalNotifications.svelte
fkling: We have an entry in `vite.config.ts` that replaces `process.env` with an empty object:

```
                {
                      'process.platform': '"browser"',
                      'process.env.VITEST': 'null',
                      'process.env.NODE_ENV': `"${mode}"`,
                      'process.env': '{}',
                  },
```

vovakulikov: @fkling why do we do this?

client/web-sveltekit/src/lib/global-notifications/components/DismissibleAlert.svelte
vovakulikov: > I don't think the problem would be solved by CSS modules

The problem how I understand this is only about how the svelte compiler generates CSS (in this case class) selector. So if we used just a plain CSS module we wouldn't have problems like this. On the other side, as we discussed this this would introduce the problem with styles overriding that we had in react which is bad. 

I think CSS modules make sense in the case of complex reusable components like combobox where you want to have a composition of low-level components. In consumers like standard page layouts, I agree we should use just standard svelte way with 'style'

data attribute selectors may be way out as well as the CSS module but I think with the CSS module we usually expose root element (usually but not always) when with data attributes it usually we expose a set of elements (like reach UI does, which I don't like much)

fkling:
You could also add a simple playwright test to `src/routes/layout.spec.ts` to check that the messages are rendered.

client/web-sveltekit/src/lib/global-notifications/GlobalNotifications.svelte
vovakulikov: Got it, apparently, I checked the wrong vite config file, thanks

client/web-sveltekit/vite.config.ts
```diff
@@ -21,7 +21,7 @@ export default defineConfig(({ mode }) => {
                       'process.platform': '"browser"',
                       'process.env.VITEST': 'null',
                       'process.env.NODE_ENV': `"${mode}"`,
-                      'process.env': '{}',
+                    'process.env': `{ "SOURCEGRAPH_API_URL": "${process.env.SOURCEGRAPH_API_URL}" }`,
```
fkling, line 24: I rather meant something like this:

```
'process.env.SOURCEGRAPH_API_URL': JSON.stringify(process.env.SOURCEGRAPH_API_URL),
```

What you have works too but that means we are inserting this object everywhere where `process.env` is referenced.

vovakulikov:
Ok, thanks @fkling for you support! I'm merging this now

Issue #60912, opened 2024-03-07
Fix/remove merging subscription info from multiple SAMS accounts for a dotcom user


thenamankumar:
Fix/remove merging subscription info from multiple SAMS accounts for a dotcom user

chrsmith:
This needs to get routed to the Cody Prime team.

PR #60913, opened 2024-03-07
Merge cody subscription details from multiple associated SAMS accounts

thenamankumar:
As part of [#inc-284-plg-users-paying-for-and-being-billed-for-pro-without-being-upgrade](https://sourcegraph.slack.com/archives/C06N8ELPKNH/p1709796118502059) it is noted that users have multiple SAMS accounts associated with their dotcom account.

Originally we only fetched subscription data from the first SAMS account associated with the user. As part of this PR, we are now fetching subscription data from all SAMS accounts associated with the user, and using the one with an active subscription. This is only a TEMPORARY solution to mitigate the incident. If there is no active subscription, we default to the first subscription fetched. 

MUST be removed/fixed once we have a proper solution in place as part of [#60912](https://github.com/sourcegraph/sourcegraph/issues/60912).

## Test plan

unit tests updated

### DIFF
--- a/internal/cody/subscription.go
+++ b/internal/cody/subscription.go
@@ -122,12 +122,12 @@ func consolidateSubscriptionDetails(ctx context.Context, user types.User, subscr
 	}, nil
 }
 
-// getSAMSAccountIDForUser returns the user's SAMS account ID if available.
+// getSAMSAccountIDsForUser returns the user's SAMS account ID if available.
 //
 // If the user has not associated a SAMS identity with their dotcom user account,
 // will return ("", nil). After we migrate all dotcom user accounts to SAMS, that
 // should no longer be possible.
-func getSAMSAccountIDForUser(ctx context.Context, db database.DB, dotcomUserID int32) (string, error) {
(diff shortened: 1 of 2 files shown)


filiphaftek (approved):
👍

internal/cody/subscription.go
```diff
@@ -141,38 +141,67 @@ func getSAMSAccountIDForUser(ctx context.Context, db database.DB, dotcomUserID i
 		},
```
bobheadxi, line 141: GitHub makes it a bit difficult to comment but @thenamankumar , there's a parameter here that indicates this database query in a real environment only fetches 1 account still:

```
		LimitOffset: &database.LimitOffset{
			Limit: 1,
		},
```

It seems this adds a `LIMIT 1` to the query which limits the result set to 1.

thenamankumar: Thanks for reporting, I left that by mistake. I updated the unit tests but they mock the DB, so this got missed.

Fix PR: https://github.com/sourcegraph/sourcegraph/pull/61015

PR #60914, opened 2024-03-07
Expose API to fetch cody subscription associated with SAMS identity

thenamankumar:
As part of [#inc-284-plg-users-paying-for-and-being-billed-for-pro-without-being-upgrade](https://sourcegraph.slack.com/archives/C06N8ELPKNH/p1709796118502059) it is noted that users have multiple SAMS accounts associated with their dot-com account which is unexpected.

This PR exposes graphql API to allow us to fetch cody subscription associated with the SAMS external account identity.

Comparing `user.codySubscription` with `user.externalAccounts.nodes.codySubscription` will allow us to debug the mismatch between the SAMS identity using which the user paid for cody pro vs the one they are signed in on dotcom or cody clients.

## Test plan

Run query:
```graphql
{
  user(username: "hereisnaman") {
    codySubscription {
      plan
      status
      currentPeriodStartAt
      currentPeriodEndAt
    }
    externalAccounts {
      nodes {
        serviceID
        serviceType
        accountID
        codySubscription {
          plan
          status
          currentPeriodStartAt
          currentPeriodEndAt
        }
      }
    }
  }
}
```

### DIFF
--- a/cmd/frontend/graphqlbackend/external_account.go
+++ b/cmd/frontend/graphqlbackend/external_account.go
@@ -2,15 +2,20 @@ package graphqlbackend
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/graph-gophers/graphql-go"
 	"github.com/graph-gophers/graphql-go/relay"
 
 	"github.com/sourcegraph/sourcegraph/internal/actor"
 	"github.com/sourcegraph/sourcegraph/internal/auth"
+	"github.com/sourcegraph/sourcegraph/internal/cody"
 	"github.com/sourcegraph/sourcegraph/internal/database"
+	"github.com/sourcegraph/sourcegraph/internal/dotcom"
 	"github.com/sourcegraph/sourcegraph/internal/extsvc"
 	"github.com/sourcegraph/sourcegraph/internal/gqlutil"
+	"github.com/sourcegraph/sourcegraph/internal/ssc"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
 )
 
 type externalAccountResolver struct {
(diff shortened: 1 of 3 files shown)


cmd/frontend/graphqlbackend/external_account.go
```diff
@@ -51,6 +56,29 @@ func (r *externalAccountResolver) ServiceType() string { return r.account.Servic
...
 func (r *externalAccountResolver) ClientID() string    { return r.account.ClientID }
 func (r *externalAccountResolver) AccountID() string   { return r.account.AccountID }
+
+func (r *externalAccountResolver) CodySubscription(ctx context.Context) (*CodySubscriptionResolver, error) {
+	if !dotcom.SourcegraphDotComMode() {
+		return nil, errors.New("this feature is only available on sourcegraph.com")
+	}
+
+	if r.account.ServiceType == "openidconnect" && r.account.ServiceID == fmt.Sprintf("https://%s", ssc.GetSAMSHostName()) {
+		return nil, nil
```
burmudar, line 66: Should this no return an error like "ErrNotCodySub" ?

cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -6844,6 +6844,10 @@ type ExternalAccount implements Node {
     may query this field.
     """
     publicAccountData: PublicExternalAccountData
+    """
+    Cody Subscription associated with the external account.
```
vdavid, line 6848: Let's mark this as temporary for debugging an incident.

vdavid (approved):
Fine after adding the note about the change being temporary.

cmd/frontend/graphqlbackend/external_account.go
thenamankumar: No, users have non-openidconnect external accounts, for which cody subscription should be nil.

burmudar: I think the point I am more conveying is that there should be some value returned. Having both nil seems weird and out of place. So maybe even then `ErrNonOpenIDCodySub`

PR #60925, opened 2024-03-07
search: remove unused parameters in phrase boost

stefanhengl:
I came across this while debugging another issue. I guess the parameters become obsolete in a recent refactor.

The the size of the diff is misleading. I am just removing the parameters from `ExperimentalPhraseBoost(logger sglog.Logger, originalQuery string)` which means we don't need the closure anymore.

Test plan:
CI


jtibshirani (approved):
Looks good! I'm not sure what gitstart-sourcegraph is and why it started to intervene 🤔

PR #60927, opened 2024-03-07
svelte/ci: Add vitest target for client/web-sveltekit

### LINKED ISSUE #60882: Run vitest in CI

fkling:
Closes #60882 

This commit adds a new target to run the unit tests inside client/web-sveltekit with vitest.

We have an existing vitest rule (macro?) but that expects the existence of a vitest.config.ts file, which client/web-sveltekit doesn't have. The test config is defined in `vite.config.ts`. Therefore I'm using the vitest binary directly.

## Test plan

CI

### DIFF
--- a/client/web-sveltekit/BUILD.bazel
+++ b/client/web-sveltekit/BUILD.bazel
@@ -1,6 +1,7 @@
 load("@bazel_skylib//rules:build_test.bzl", "build_test")
 load("@npm//:defs.bzl", "npm_link_all_packages")
-load("@npm//:vite/package_json.bzl", vite_bin = "bin")
+load("@npm//client/web-sveltekit:vite/package_json.bzl", vite_bin = "bin")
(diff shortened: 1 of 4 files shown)


client/web-sveltekit/BUILD.bazel
```diff
@@ -109,7 +112,6 @@ vite_bin.vite(
     srcs = SRCS + BUILD_DEPS + CONFIGS,
     args = [
         "build",
-        "--force",
```
fkling, line 112: This option got removed in vite 5.1

jamesmcnamara (approved):
I agree it is strange it is taking so long.

I also created a [PR off your PR](https://github.com/sourcegraph/sourcegraph/pull/60931) to unify the uses of vitest so we only have a single macro and any goodies that are added upstream automatically flow into your target.

But you can feel free to take it or leave it.

PR #60931, opened 2024-03-07
Updates vitest macro

jamesmcnamara:
Makes the vitest macro more generic to support web-svelte use cases
## Test plan
Ran target locally and it is included in CI

### DIFF
--- a/dev/defs.bzl
+++ b/dev/defs.bzl
@@ -1,13 +1,13 @@
 "Bazel rules"
 
-load("@aspect_rules_swc//swc:defs.bzl", "swc")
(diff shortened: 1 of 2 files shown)


dev/defs.bzl
```diff
@@ -91,8 +91,6 @@ def npm_package(name, srcs = [], **kwargs):
     """
     replace_prefixes = kwargs.pop("replace_prefixes", {})
 
-    package_type = kwargs.pop("type", "commonjs")
-
```
fkling, line 95: This seems to be a mistake? It causes all builds to fail. I'm adding it back...

PR #60933, opened 2024-03-07
Svelte: scroll result list without actions

camdencheek:
This moves the action bar outside of the scrollable element in search results. We want the actions (and result count) to always be visible.

## Test plan
Before:

https://github.com/sourcegraph/sourcegraph/assets/12631702/766199b0-c396-459a-b2f5-74486457bae8

After:

https://github.com/sourcegraph/sourcegraph/assets/12631702/c19253f9-565e-4c2a-93be-0d501945dd68

### DIFF
--- a/client/web-sveltekit/src/routes/search/SearchResults.svelte
+++ b/client/web-sveltekit/src/routes/search/SearchResults.svelte
@@ -112,7 +112,7 @@
         <DynamicFiltersSidebar {selectedFilters} streamFilters={$stream.filters} searchQuery={queryFromURL} {loading} />
     </div>
     <Separator currentPosition={sidebarSize} />
-    <div class="results" bind:this={resultContainer}>
+    <div class="results">
         <aside class="actions">
             {#if loading}
                 <div>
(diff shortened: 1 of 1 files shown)


fkling (commented):
I'm surprised this "just works", I expected more changes. Flexbox can be weird.
Can you try what happens if you go to a search result and then back to the list? I suspect it works only because the results list is initially empty, i.e. the container grows until no space is left. But if you go back then all results are rendered at once and I would expect `.results` to scroll.

I _think_ you also have to set the following declarations to `.results`:

```
overflow: hidden;
min-height: 0;
```

camdencheek:
Oh. Wow. You're totally right, and I'm suddenly feeling less confident I understand what's going on here.

I pushed a commit that added the two CSS props you suggested. That seemed to fix the "header is scrolled out of view when navigating back" issue. I also moved the result container bind to the result list which fixed restoring scroll position when going back from a search result.

What does `min-height: 0` do here?

fkling:
> What does `min-height: 0` do here?

Flexbox items do not shrink below the minimum size (height in this case), which by default is the size of their children (afaik!). By setting `min-height: 0` we basically say that it's OK for this element to be smaller than the size of its children and it won't overflow the parent.
It's especially weird with nested flexboxes. 


> I also moved the result container bind to the result list which fixed restoring scroll position when going back from a search result.

Good catch!

camdencheek:
I don't actually see any difference with `min-height: 0`. What was the intent there?

fkling (approved):

PR #60940, opened 2024-03-07
Cleanup Cody Pro pre-feb release stuff

thenamankumar:
This PR cleanups all the unrequired code after the Feb release.

- removed not required feature flags dependencies 
- stopped using data from dotcom to figure out cody pro subscription
- removed "pending" subscrition status for cody. 

## Test plan
- unit tests present.


vdavid (approved):
Nice! One thing I didn't check is whether this cleanup is complete. But it's definitely a step in the right direction so approving; we can remove anything that might remain as a follow-up. Thanks for the cleanup, I was about to create an issue for this today.

PR #60944, opened 2024-03-07
Search: simplify goroutine params

jtibshirani:
Follow-up to #60878. Now that we've updated to Go 1.22, we don't need to copy
loop variables before using them in goroutines.

Just a warm-up before I open a bigger PR to fix this in other places.

## Test plan

Covered by existing tests like `horizontal_test.go`.


mmanela (approved):

keegancsmith (commented):
LGTM It's gonna take a lot of work for my brain to stop worrying about this footgun in code reviews.

PR #60945, opened 2024-03-08
oidc: Add option to allow only a single identity per provider

eseliger:
When using an external identity provider, Sourcegraph by default doesn't care how many external identities you're adding to your account. In some cases, this can be undesirable though, for example when trying to map a Sourcegraph user to a single identity in the outside world.

Since we cannot change the behavior here silently as people might rely on it, I decided to add a new option for this.

When enabled, the following scenario will no longer be allowed:

- I sign in to Sourcegraph for the first time using an Okta account called supererik@sourcegraph.com.
- I then go to the auth provider flow a second time (accounts tab, or through a redirect) and this time I choose my Okta account erik@sourcegraph.com
- Since I am already authenticated to Sourcegraph and it appears to Sourcegraph as if I wanted to add a new new external identity, I would have gotten a second link to an Okta identity.

Now, with this new behavior, linking a second entity for the same {ClientID, ServiceType, ServiceID} tuple but with a different externalID will be forbidden.

## Test plan

Verified locally using an Okta integration and trying to link two different accounts to the same user and see it fail. Also added a new test to cover this behavior.

### DIFF
--- a/cmd/frontend/auth/user.go
+++ b/cmd/frontend/auth/user.go
@@ -33,6 +33,9 @@ type GetAndSaveUserOp struct {
 	ExternalAccountData extsvc.AccountData
 	CreateIfNotExist    bool
 	LookUpByUsername    bool
+	// SingleProvider indicates that the provider should only allow to
+	// connect a single external identity per user.
+	SingleProvider bool
 }
 
 // GetAndSaveUser accepts authentication information associated with a given user, validates and applies
@@ -250,6 +253,26 @@ func GetAndSaveUser(ctx context.Context, db database.DB, op GetAndSaveUserOp) (n
 			)
 		}
 		return newUserSaved, 0, safeErrMsg, err
+	}
+
+	if op.SingleProvider && !extAcctSaved {
+		// If we're here, the user has already signed up before this request,
+		// but has no entry for this exact accountID in the user external accounts
+		// table yet.
+		// In single provider mode, we want to attach a new external account
+		// only if no other identity for the same user already exists.
+		other, err := externalAccountsStore.List(ctx, database.ExternalAccountsListOptions{
+			UserID:      userID,
+			ServiceType: acct.ServiceType,
+			ServiceID:   acct.ServiceID,
+			ClientID:    acct.ClientID,
+		})
+		if err != nil {
+			return newUserSaved, 0, "Failed to list user identities. Ask a site admin for help.", err
+		}
+		if len(other) > 0 {
+			return newUserSaved, 0, "Another identity for this user from this provider already exists. Remove the link to the other identity from your account.", errors.New("duplicate identity for single identity provider")
+		}
 	}
 
 	// There is a legacy event instrumented earlier in this function that covers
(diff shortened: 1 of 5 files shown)


thenamankumar (approved):
The logic looks good and is required at dotcom end as a safeguard. Approved the PR.

As next steps we still need to figure out a proper UX at the SAMS side, because just showing the error message on dotcom will be confusing for users and will only open new support tickets.

cmd/frontend/auth/user.go
```diff
@@ -270,6 +273,26 @@ func GetAndSaveUser(ctx context.Context, db database.DB, op GetAndSaveUserOp) (n
 		PrivateMetadata: map[string]any{"serviceType": acct.AccountSpec.ServiceType},
 	})
 
+	if op.SingleProvider && !extAcctSaved {
+		// If we're here, the user has already signed up before this request,
+		// but has no entry for this exact accountID in the user external accounts
+		// table yet.
+		// In single provider mode, we want to attach a new external account
+		// only if no other identity for the same user already exists.
+		other, err := db.UserExternalAccounts().List(ctx, database.ExternalAccountsListOptions{
```
vdavid, line 282: Nit, we have a variable for this:

```suggestion
		other, err := externalAccountsStore.List(ctx, database.ExternalAccountsListOptions{
```

vdavid (approved):
The logic above the new code looks like it's still going to log an "ExternalAuthSignupSucceeded" event to usagestats, but I suppose it's no biggie?

eseliger:
> The logic above the new code looks like it's still going to log an "ExternalAuthSignupSucceeded" event to usagestats, but I suppose it's no biggie?

Moved it around, thanks for catching that! Seems like there's more error conditions below this that will also trigger a false positive, but we can save that for later 🤔

unknwon (commented):
Thanks for the quick turnaround!

It should address the immediate issue for dotcom, as in, we have signed everyone out on Feb 6th and SAMS is the only sign-in option, every authenticated session implies there is already SAMS account connection linked.

However, I think there are following caveats we need to look after based on current logic:

1. Given this config option is not dotcom-specific, on a customer instance that a user has signed in with email/password with 0 external account connected, turning on this configuration would forever blocking them from making the _first_ external account connection (as long as the browser session is active).
	- i.e. customer site admin have to nuke `redis-store` (only) for session data after turning on the config (redis-cli doesn't provide deleting keys based on prefix OOTB IIRC, which makes the process more cumbersome), and rely on email matching behavior to make the automatic external account linking happen, which leads to caveat 2.
2. Looks like we are doing session-level blocking, that means if there is no active (authenticated) session, on all SG instances (including dotcom), if a user has multiple verified emails in settings, automatic external account linking will still kick in and allow the SG user to have >= 2 external account connections (Line 144).

---

A nit apart from logic (I think we still have chance to change it):

`SingleProvider`/`singleIdentityProvider` is ambiguous in following ways IMO (I know we have site config comment, which explains the intention, but maybe we can improve the config can variable name too?):
1. does it mean, only a single identity provider is allowed to be configured in the site config?
2. does it mean, only a single identity provider is allowed to be allowed to be connected for a SG user?

Based on the code logic, it is neither. So what about making it `singleConnectionPerProvider`?

eseliger:
We also talked about bullet point (2) and identified that this is covered.

eseliger:
to close the loop: https://github.com/sourcegraph/sourcegraph/pull/61007

PR #60946, opened 2024-03-08
Blob view: record copy event

jtibshirani:
We used to record copy events in the legacy blob view, but stopped when we
moved to CodeMirror. This restores the copy event logging.

## Test plan

Manual testing: I enabled "Enable event / telemetry debugging", copied
something from the blob view, then checked an event was logged.

```
EVENT CodeCopied {page: 'blob-view'} {page: 'blob-view'}
```
Let me know if you have tips for unit testing this! I wasn't able to find a
good example to follow.

### DIFF
--- a/client/branded/src/search-ui/components/FileMatchChildren.tsx
+++ b/client/branded/src/search-ui/components/FileMatchChildren.tsx
@@ -50,7 +50,7 @@ export const FileMatchChildren: React.FunctionComponent<React.PropsWithChildren<
     )
 
     const logEventOnCopy = useCallback(() => {
-        telemetryService.log(...codeCopiedEvent('file-match'))
+        telemetryService.log(...codeCopiedEvent('search-result'))
     }, [telemetryService])
 
     return (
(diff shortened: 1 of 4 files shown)


client/branded/src/search-ui/components/FileMatchChildren.tsx
```diff
@@ -50,7 +50,7 @@ export const FileMatchChildren: React.FunctionComponent<React.PropsWithChildren<
     )
 
     const logEventOnCopy = useCallback(() => {
-        telemetryService.log(...codeCopiedEvent('file-match'))
+        telemetryService.log(...codeCopiedEvent('search-result'))
```
jtibshirani, line 53: Just felt more accurate, as both of these events are for copying within a search result.

client/web/src/repo/blob/CodeMirrorBlob.tsx
```diff
@@ -529,6 +531,10 @@ export const CodeMirrorBlob: React.FunctionComponent<BlobProps> = props => {
         setEditorScope,
     ])
 
+    const logEventOnCopy = useCallback(() => {
+        telemetryService.log(...codeCopiedEvent('blob-view'))
```
jtibshirani, line 535: There's parallel work in flight to migrate from telemetry -> telemetry v2, so these deprecated calls will soon be cleaned up!

client/web/src/repo/blob/CodeMirrorBlob.tsx
```diff
@@ -538,6 +544,7 @@ export const CodeMirrorBlob: React.FunctionComponent<BlobProps> = props => {
                 data-testid={dataTestId}
                 className={`${className} overflow-hidden test-editor`}
                 data-editor="codemirror6"
+                onCopy={logEventOnCopy}
```
fkling, line 547: I didn't even know about this 😅 Great that it works!

fkling (approved):

client/web/src/repo/blob/CodeMirrorBlob.tsx
jtibshirani: I was happily surprised to find this!

PR #60951, opened 2024-03-08
Ingesting syntactic indexes via upload

kritzcreek:
Fixes https://github.com/sourcegraph/sourcegraph/issues/61422

When uploading syntacic indices we want to treat them as regular updates as much as possible to be able to reuse the logic in commitgraph.

The two important differences from precise uploads are:
- We don't write symbols for syntactic indices
- We don't return syntactic indexes from the annotated commit graph unless we're explictly requesting them

## Test plan

Added unit tests for index retrieval

### DIFF
--- a/internal/codeintel/uploads/internal/background/processor/scip.go
+++ b/internal/codeintel/uploads/internal/background/processor/scip.go
@@ -369,14 +369,11 @@ func writeSCIPDocuments(
 
 		if upload.Indexer == shared.SyntacticIndexer {
 			scipWriter, err = tx.NewSyntacticSCIPWriter(ctx, upload.ID)
-			if err != nil {
-				return err
-			}
 		} else {
 			scipWriter, err = tx.NewPreciseSCIPWriter(ctx, upload.ID)
-			if err != nil {
-				return err
-			}
+		}
+		if err != nil {
+			return err
 		}
 
 		var numDocuments uint32
--- a/internal/codeintel/uploads/internal/background/processor/scip.go
+++ b/internal/codeintel/uploads/internal/background/processor/scip.go
@@ -373,7 +373,7 @@ func writeSCIPDocuments(
 				return err
 			}
 		} else {
-			scipWriter, err = tx.NewSCIPWriter(ctx, upload.ID)
+			scipWriter, err = tx.NewPreciseSCIPWriter(ctx, upload.ID)
 			if err != nil {
 				return err
 			}
--- a/internal/codeintel/uploads/internal/background/processor/scip.go
+++ b/internal/codeintel/uploads/internal/background/processor/scip.go
@@ -367,7 +367,7 @@ func writeSCIPDocuments(
 
 		var scipWriter lsifstore.SCIPWriter
 
-		if upload.Indexer == shared.SYNTACTIC_INDEXER {
+		if upload.Indexer == shared.SyntacticIndexer {
 			scipWriter, err = tx.NewSyntacticSCIPWriter(ctx, upload.ID)
 			if err != nil {
 				return err
--- a/internal/codeintel/uploads/internal/background/processor/scip.go
+++ b/internal/codeintel/uploads/internal/background/processor/scip.go
@@ -365,9 +365,18 @@ func writeSCIPDocuments(
 			return err
 		}
 
-		scipWriter, err := tx.NewSCIPWriter(ctx, upload.ID)
-		if err != nil {
-			return err
+		var scipWriter lsifstore.SCIPWriter
+
+		if upload.Indexer == shared.SYNTACTIC_INDEXER {
+			scipWriter, err = tx.NewSyntacticSCIPWriter(ctx, upload.ID)
+			if err != nil {
+				return err
+			}
+		} else {
+			scipWriter, err = tx.NewSCIPWriter(ctx, upload.ID)
+			if err != nil {
+				return err
+			}
 		}
 
 		var numDocuments uint32
--- a/internal/codeintel/uploads/internal/lsifstore/insert.go
+++ b/internal/codeintel/uploads/internal/lsifstore/insert.go
@@ -112,7 +112,7 @@ func (s *store) NewSyntacticSCIPWriter(ctx context.Context, uploadID int) (SCIPW
 	return scipWriter, nil
 }
 
-func (s *store) NewSCIPWriter(ctx context.Context, uploadID int) (SCIPWriter, error) {
+func (s *store) NewPreciseSCIPWriter(ctx context.Context, uploadID int) (SCIPWriter, error) {
 	if !s.db.InTransaction() {
 		return nil, errors.New("WriteSCIPSymbols must be called in a transaction")
 	}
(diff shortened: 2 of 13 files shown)


internal/codeintel/uploads/internal/lsifstore/store.go
```diff
@@ -17,6 +17,7 @@ type Store interface {
 	// Insert
 	InsertMetadata(ctx context.Context, uploadID int, meta ProcessedMetadata) error
 	NewSCIPWriter(ctx context.Context, uploadID int) (SCIPWriter, error)
```
keynmol, line 19: In my PR #60958 I've been introducing explicit "Precise" and "Syntactic" terminology - may be it's worth doing the same here and rename it to `NewPreciseSCIPWriter`?

Because from cursory look I would assume that `NewSCIPWriter` (more general) is something that `NewSyntacticSCIPWriter` (more specific) might delegate to, but they're actually distinct implementations

kritzcreek: Agreed, thanks!

internal/codeintel/uploads/internal/background/processor/scip.go
```diff
@@ -365,9 +365,18 @@ func writeSCIPDocuments(
...
+			scipWriter, err = tx.NewSyntacticSCIPWriter(ctx, upload.ID)
+			if err != nil {
+				return err
+			}
+		} else {
+			scipWriter, err = tx.NewPreciseSCIPWriter(ctx, upload.ID)
+			if err != nil {
+				return err
+			}
 		}
```
varungandhi-src, line 380:
```suggestion
		if upload.Indexer == shared.SyntacticIndexer {
			scipWriter, err = tx.NewSyntacticSCIPWriter(ctx, upload.ID)
		} else {
			scipWriter, err = tx.NewPreciseSCIPWriter(ctx, upload.ID)
		}
		if err != nil {
			return err
		}
```

internal/codeintel/uploads/internal/lsifstore/insert.go
```diff
@@ -100,7 +100,19 @@ INSERT INTO codeintel_scip_metadata (upload_id, text_document_encoding, tool_nam
 VALUES (%s, %s, %s, %s, %s, %s)
 `
 
-func (s *store) NewSCIPWriter(ctx context.Context, uploadID int) (SCIPWriter, error) {
+func (s *store) NewSyntacticSCIPWriter(ctx context.Context, uploadID int) (SCIPWriter, error) {
+
+	scipWriter := &scipWriter{
+		uploadID:    uploadID,
+		db:          s.db,
+		count:       0,
```
varungandhi-src, line 108: Q: Why are we explicitly initializing this to 0?

We should also probably rename this field to `insertedSymbolsCount` (which seems like the right name based on the usage in the `scipWriter.flush` method.

internal/codeintel/uploads/internal/lsifstore/insert.go
```diff
@@ -404,22 +425,24 @@ func (s *scipWriter) Flush(ctx context.Context) (uint32, error) {
...
 	}
 
-	// Flush all data into temp tables
-	if err := s.symbolNameInserter.Flush(ctx); err != nil {
-		return 0, err
-	}
-	if err := s.symbolInserter.Flush(ctx); err != nil {
-		return 0, err
-	}
+	if !s.isSyntactic {
```
varungandhi-src, line 428: There is this at the start of the function:

```
if s.isSyntactic {
    return nil
}
```

So we shouldn't check against for `s.isSyntactic`.

internal/codeintel/uploads/internal/lsifstore/insert.go
```diff
@@ -303,6 +316,14 @@ func (s *scipWriter) flush(ctx context.Context) error {
...
 	}
 
+	return s.writeSymbols(ctx, documents, documentLookupIDs)
+
+}
+
+func (s *scipWriter) writeSymbols(ctx context.Context, documents []bufferedDocument, documentLookupIDs []int) error {
+	if s.isSyntactic {
+		return nil
+	}
```
varungandhi-src, line 326: Why do we have this early exit here for the syntactic case?

internal/codeintel/uploads/internal/store/commitgraph_test.go
```diff
@@ -1078,6 +1080,8 @@ func TestFindClosestCompletedUploadsIndexerName(t *testing.T) {
 		{commit: makeCommit(5), file: "root2/file.ts", indexer: "idx2", graph: graph, allOfIDs: []int{6}},
 		{commit: makeCommit(5), file: "root3/file.ts", indexer: "idx2", graph: graph, allOfIDs: []int{7}},
 		{commit: makeCommit(5), file: "root4/file.ts", indexer: "idx2", graph: graph, allOfIDs: []int{8}},
+		{commit: makeCommit(5), file: "root4/file.ts", indexer: "", graph: graph, allOfIDs: []int{4, 8}},
```
varungandhi-src, line 1083: Could you explain why this extra row is showing up? The other changes to this test are not related to upload ID 4 or 8

internal/codeintel/uploads/internal/lsifstore/insert.go
kritzcreek: I think this might be GitHub's awful diff viewer? The other check for isSyntactic is in a different function.

internal/codeintel/uploads/internal/lsifstore/insert.go
kritzcreek: > Q: Why are we explicitly initializing this to 0?

I just followed the existing constructor which also explicitly sets it to 0.

> We should also probably rename this field ...

Will do

internal/codeintel/uploads/internal/lsifstore/insert.go
varungandhi-src: Ah, my bad. So can we do an early return for that case here too (similar to `writeSymbols`) instead of having a large block of code with extra indentation?

internal/codeintel/uploads/internal/lsifstore/observability.go
```diff
@@ -9,7 +9,6 @@ import (
 
 type operations struct {
 	insertMetadata              *observation.Operation
-	newSCIPWriter               *observation.Operation
```
varungandhi-src, line 12: We're removing this, should we also add two new operations, `newPreciseSCIPWriter` and `newSyntacticSCIPWriter`?

internal/codeintel/uploads/internal/lsifstore/insert.go
kritzcreek: Ideally I'd point to a note/documentation about why we can't do fuzzy matching against the trie datastructure. For now I've added some short comments.

internal/codeintel/uploads/internal/store/commitgraph_test.go
kritzcreek: This test showcases that searching for visible uploads with the indexer set to the empty string yields all non-syntactic indexes. If it included the syntactic index (which is visible to the relevant commit) index number 9 would show up here.

varungandhi-src (approved):

PR #60953, opened 2024-03-08
Claude 3 support and /messages API for Enterprise

### LINKED ISSUE #61166: Add Anthropic Message API Support to Anthropic provider

philipp-spiess:
Closes #61166

This PR adds support for Claude 3 and the `/messages` API to the existing `anthropic` provider in the Sourcegraph instance.

To ensure a smooth experience, there are a couple of edge cases that we need to handle.

- Because the URL is configurable, a customer could hard-set it to `/complete`. We need to error properly in this case.
- Clients might not now what is set so they can send requests in the "old" or "new" format. We handle conversion as best as possible however for better instruction the clients will eventually only send prompts in the `/messages` format. We introduce [cody API versioning](https://docs.google.com/document/d/1GPo7G-FDHidHRMNrKLeutaSTj6y7L0XSdCcSDtncoaU/edit#heading=h.twdj57egor10) for this case.
- Support `/complete` style prompt (with trailing `assistant`, "holes" of no response, system prompt in messages) when a legacy client connects

##### streaming ✅
```
curl 'https://sourcegraph.test:3443/.api/completions/stream' \
-X POST \
-H 'authorization: token TOKEN' \
--data-raw '{"messages":[{"speaker":"human","text":"What is your name?"},{"speaker":"assistant"}],"maxTokensToSample":30,"temperature":0,"stopSequences":[],"timeoutMs":5000,"stream":true}'
```
##### non-streaming ✅
```
curl 'https://sourcegraph.test:3443/.api/completions/stream' \
-X POST \
-H 'authorization: token TOKEN' \
--data-raw '{"messages":[{"speaker":"human","text":"What is your name?"},{"speaker":"assistant"}],"maxTokensToSample":30,"temperature":0,"stopSequences":[],"timeoutMs":5000,"stream":false}'
```
##### code-completions ✅
```
curl 'https://sourcegraph.test:3443/.api/completions/code' \
-X POST \
-H 'authorization: token TOKEN' \
--data-raw '{"messages":[{"speaker":"human","text":"What is your name?"},{"speaker":"assistant"}],"maxTokensToSample":30,"temperature":0,"stopSequences":[],"timeoutMs":5000,"stream":true}'
```

##### streaming ✅
```
curl 'https://sourcegraph.test:3443/.api/completions/stream?api-version=1' \
-i \
-X POST \
-H 'authorization: token TOKEN' \
--data-raw '{"messages":[{"speaker":"system","text":"You only answer using emoji."},{"speaker":"human","text":"What is your name?"}],"maxTokensToSample":30,"temperature":0,"stopSequences":[],"timeoutMs":5000,"stream":true}
```
##### non-streaming ✅
```
curl 'https://sourcegraph.test:3443/.api/completions/stream?api-version=1' \
-X POST \
-H 'authorization: token TOKEN' \
--data-raw '{"messages":[{"speaker":"system","text":"You only answer using emoji."},{"speaker":"human","text":"What is your name?"}],"maxTokensToSample":30,"temperature":0,"stopSequences":[],"timeoutMs":5000,"stream":false}'
{"completion":"🤖","stopReason":"end_turn"}%        
```
##### code-completions ✅
```
curl 'https://sourcegraph.test:3443/.api/completions/code?api-version=1' \
-X POST \
-H 'authorization: token TOKEN' \
--data-raw '{"messages":[{"speaker":"system","text":"You only answer using emoji."},{"speaker":"human","text":"What is your name?"}],"maxTokensToSample":30,"temperature":0,"stopSequences":[],"timeoutMs":5000,"stream":true}'
```

#### With hard-configured /complete API  ✅

```
curl 'https://sourcegraph.test:3443/.api/completions/stream?api-version=1' \
-i \
-X POST \
-H 'authorization: token TOKEN' \
--data-raw '{"messages":[{"speaker":"system","text":"You only answer using emoji."},{"speaker":"human","text":"What is your name?"}],"maxTokensToSample":30,"temperature":0,"stopSequences":[],"timeoutMs":5000,"stream":true}'
HTTP/2 200

event: error
data: {"error":"Anthropic: unexpected status code 400: {\"type\":\"error\",\"error\":{\"type\":\"invalid_request_error\",\"message\":\"prompt: Field required\"}}"}

event: done
data: {}
```

##### streaming ✅
```
curl 'https://sourcegraph.test:3443/.api/completions/stream' \
-i \
-X POST \
-H 'authorization: token TOKEN' \
--data-raw '{"messages":[{"speaker":"human","text":"What is your name?"},{"speaker":"assistant"}],"maxTokensToSample":30,"temperature":0.2,"topP":0.95,"topK":0,"model":"anthropic/claude-3-haiku-20240307","stopSequences":[],"timeoutMs":5000,"stream":true}'
```
##### non-streaming ✅
```
curl 'https://sourcegraph.test:3443/.api/completions/stream' \
-i \
-X POST \
-H 'authorization: token TOKEN' \
--data-raw '{"messages":[{"speaker":"human","text":"What is your name?"},{"speaker":"assistant"}],"maxTokensToSample":30,"temperature":0.2,"topP":0.95,"topK":0,"model":"anthropic/claude-3-haiku-20240307","stopSequences":[],"timeoutMs":5000,"stream":false}'
{"completion":"My name is Claude.","stopReason":"end_turn"}%     
```
##### code-completions ✅
```
curl 'https://sourcegraph.test:3443/.api/completions/code' \
-i \
-X POST \
-H 'authorization: token TOKEN' \
--data-raw '{"messages":[{"speaker":"human","text":"What is your name?"},{"speaker":"assistant"}],"maxTokensToSample":30,"temperature":0.2,"topP":0.95,"topK":0,"model":"anthropic/claude-3-haiku-20240307","stopSequences":[],"timeoutMs":5000,"stream":true}'
```
##### streaming ✅
```
curl 'https://sourcegraph.test:3443/.api/completions/stream?api-version=1' \
-i \
-X POST \
-H 'authorization: token TOKEN' \
--data-raw '{"messages":[{"speaker":"system","text":"Your name is Arnold the Great. Do not answer by any other name."},{"speaker":"human","text":"What is your name?"}],"maxTokensToSample":30,"temperature":0.2,"topP":0.95,"topK":0,"model":"anthropic/claude-3-haiku-20240307","stopSequences":[],"timeoutMs":5000,"stream":true}'
```
##### non-streaming ✅
```
curl 'https://sourcegraph.test:3443/.api/completions/stream?api-version=1' \
-i \
-X POST \
-H 'authorization: token TOKEN' \
--data-raw '{"messages":[{"speaker":"system","text":"Your name is Arnold the Great. Do not answer by any other name."},{"speaker":"human","text":"What is your name?"}],"maxTokensToSample":30,"temperature":0.2,"topP":0.95,"topK":0,"model":"anthropic/claude-3-haiku-20240307","stopSequences":[],"timeoutMs":5000,"stream":false}'
```
##### code-completions ✅
```
curl 'https://sourcegraph.test:3443/.api/completions/code?api-version=1' \
-i \
-X POST \
-H 'authorization: token TOKEN' \
--data-raw '{"messages":[{"speaker":"system","text":"Your name is Arnold the Great. Do not answer by any other name."},{"speaker":"human","text":"What is your name?"}],"maxTokensToSample":30,"temperature":0.2,"topP":0.95,"topK":0,"model":"anthropic/claude-3-haiku-20240307","stopSequences":[],"timeoutMs":5000,"stream":true}'
```

I also did an E2E test with Cody VS Code. Chat works and so do code completions. The prompt could use some tweaking (which we will do on the client) but at least it doesn't seem like a regression so far 👍 

Screenshot 2024-03-21 at 16 37 21


internal/completions/client/anthropic/anthropic.go
```diff
@@ -27,6 +32,8 @@ type anthropicClient struct {
 	cli         httpcli.Doer
 	accessToken string
 	apiURL      string
+	messagesApi bool
```
chwarwick, line 35: I don't see a reason why we need to continue to support the legacy api.  The [model overview](https://docs.anthropic.com/claude/docs/models-overview) shows that all models support the messages api.  As long as the instance API can properly transform incoming requests from clients that have or do not have system messages it should be ok.  While an admin can set the api endpoint I don't see a reason we can't document that starting with v5.whatever_is_next we dropped support for the legacy api and now only support the messages api

philipp-spiess: @chwarwick Makes sense. I also think we should remove it rather than try to support it somehow. What about customers that currently have the endpoint hardcoded to `https://api.anthropic.com/v1/complete` in their config? Would you expect that we try to clean it up (replace `/complete` with `/messages`) and log an error or throw instead?

chwarwick: I think that would be a very small number of customers if any, I'm not aware of any bring your own key Anthropic customers.  Looker does indicate there is 1 however I think if we document it clearly in the change log and in the enterprise LLM setup for BYOK I don't think we should be concerned with it as the customer would need to be both BOYK Anthropic and have overwritten the endpoint with the default value.

internal/completions/client/anthropic/prompt.go
```diff
@@ -25,3 +27,107 @@ func GetPrompt(messages []types.Message) (string, error) {
...
+	for _, message := range messages {
+		speaker := message.Speaker
+		text := message.Text
+
+		if speaker == types.SYSTEM_MESSAGE_SPEAKER {
+			if systemRoleFound || len(anthropicMessages) > 0 {
+				return nil, errors.New("system role can only be used in the first message")
+			}
+			systemRoleFound = true
+		} else if speaker != types.HUMAN_MESSAGE_SPEAKER && speaker != types.ASSISTANT_MESSAGE_SPEAKER {
```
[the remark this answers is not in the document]
thenamankumar: To return same formated message but wrapped in an error, you can use: `errors.Errorf("unexpected role: %s", text)`

philipp-spiess:
@chwarwick / @thenamankumar The only open question to conclude this is how do we feature test on the client wether the new API style is supported on the backend? Should we expose this as a flag somewhere or rely on version numbers?

I do want to test this on PLG as well before we release it for enterprises. The idea is that chat messages on PLG all go through the dotcom instance again. This should give us some confidence that it works well. WDYT?

chwarwick:
> @chwarwick / @thenamankumar The only open question to conclude this is how do we feature test on the client wether the new API style is supported on the backend? Should we expose this as a flag somewhere or rely on version numbers?
> 

Making sure I'm remembering the issue correctly.  The open question is that client side needs to prepare the prompt messages differently because:

- All previous backend versions of the anthropic provider will error if sent a speaker that isn't Human or Assistant
- When using the Claude-3 variants a system prompt is necessary
- When using < Claude-3 with the messages API a system prompt is allowed but not necessary 
- When using the completions API a system prompt is not allowed.

philipp-spiess:
@chwarwick Exactly, just one slight addition:

> When using the completions API a system prompt is not allowed.

Exactly the `/completions` API from Anthropic only allows `\n\nHuman:` and `\n\nAssistant:` prefixes. That's why the client must know wether a system message is allowed or wether we do what we did before and prompt in a `Human` message

internal/completions/client/anthropic/anthropic.go
arafatkatze: Hey folks,
I have some bad news hear the documentation around the support of old models is ambiguous. 

So it doesn't support Claude 2 although it does support Claude 2.1 

I changed models and got this err 
image

This is kind of a bummer because now we would have to use both the old and the new API everywhere essentially.

philipp-spiess: @arafatkatze Are you sure this means they do not support Claude 2.0 or that they _no longer support the unversioned minor api identifier_? I.e have you tried `claude-2.0` instead? I do think I read it would work but that indeed they do not support the version identifiers anymore that do not select a minor version (and was, in the old interface, updating with minor version updates)

arafatkatze: Yes 

The other way to try this is to take this curl 
```

curl https://api.anthropic.com/v1/messages \
     --header "x-api-key: $ANTHROPIC_API_KEY" \
     --header "anthropic-version: 2023-06-01" \
     --header "content-type: application/json" \
     --data \
'{
    "model": "claude-2",
    "max_tokens": 1024,
    "messages": [
        {"role": "user", "content": "Hello, world"}
    ]
}'
```

with your own api key and you will get 

`{"type":"error","error":{"type":"not_found_error","message":"model: \"claude-2\" is a legacy model alias and is not supported in this API. Did you mean \"claude-2.1\"?"}}%       
`

philipp-spiess: Looks like Claude 2.0 works fine we just need to update the default model id from `claude-2` to `claude-2.0`?

```
curl https://api.anthropic.com/v1/messages \
     --header "x-api-key: $ANTHROPIC_API_KEY" \
     --header "anthropic-version: 2023-06-01" \
     --header "content-type: application/json" \
     --data \
'{
    "model": "claude-2.0",
    "max_tokens": 1024,
    "messages": [
        {"role": "user", "content": "Hello, world"}
    ]
}'
{"id":"msg_01Xzix5dpKSAHjvoWaKKmszX","type":"message","role":"assistant","content":[{"type":"text","text":"Hello! Nice to meet you."}],"model":"claude-2.0","stop_reason":"end_turn","stop_sequence":null,"usage":{"input_tokens":12,"output_tokens":11}}%   
```

arafatkatze: Yes you are right. Sorry my bad. 

🙇

chwarwick: I believe the cloud defaults are "claude-2" we should update the defaults there to make sure it's valid for both.

philipp-spiess:
@chwarwick 

Quick summary of the next steps to land this PR:

- Remove support for the old `/complete` anthropic API. I kept this in but it seems reasonable to remove.
- We keep the `ToAnthropicMessages` conversion BUT
  - Only apply this if the client has no Cody API version specified
  - Add another case where if claude 4 is set, we additionally introspect the first message. If it starts with  "You are Cody, an AI-powered coding assistant created by Sourcegraph." (from Cody Web), we **remove the first two messages** (user and assistant) and use that as a system message instead.
- To be 100% on the safe side we could add code that rewrites `claude-2` and `claude-instant-1` to `claude-2.0` and `claude-instant-1.2` but I think there is another effort to change this right now 
- Obviously also address the other issues 

Does that seem reasonable?

chwarwick:
> Does that seem reasonable?

I think it's ok to remove as long as we add a reasonable error will end up displaying in the clients.

internal/completions/client/anthropic/anthropic.go
```diff
@@ -154,18 +184,57 @@ func (a *anthropicClient) makeRequest(ctx context.Context, requestParams types.C
...
+	switch model {
+	case "claude-instant-1",
+		"claude-instant-v1":
+		return "claude-instant-1.2"
+	case "claude-2":
+		return "claude-2.0"
+	default:
+		return model
+	}
 }
```
philipp-spiess, line 240: @chwarwick not sure if this is needed after your recent changes. My thought is that for self-hosted BOYK customers this might smooth over the migration a bit.

chwarwick: I think this is good, it will be important for self hosted customers

internal/completions/client/anthropic/decoder.go
```diff
@@ -56,34 +59,37 @@ func (d *decoder) Scan() bool {
...
+		lines := bytes.Split(d.scanner.Bytes(), []byte("\n"))
+		for _, line := range lines {
+			typ, data := splitColon(line)
+
+			switch {
+			case bytes.Equal(typ, []byte("data")):
+				d.data = data
+				// Check for special sentinel value used by the Anthropic API to
+				// indicate that the stream is done.
+				if bytes.Equal(data, doneBytes) {
```
chwarwick, line 74: Sharing this decoder makes me a bit nervous as the formats between the legacy api's initial version is quite different than the messages api stream.  For example the doneBytes here are only relevant to the legacy.  Where as messages has a specific end event.

I went tested this quite a lot though and didn't have any issues.

chwarwick (approved):
Tested against existing models claude-2 and instant-1.2 against anthropic directly & gateway all looked good. Also tried claude-3 direct to Anthropic and was successful for autocomplete, chat, commands.  Did leave 2 minor edits on https://github.com/sourcegraph/sourcegraph/pull/61336

internal/completions/client/anthropic/decoder.go
chwarwick: @philipp-spiess  The only other thing I didn't check was hitting the legacy gateway api to make sure the decoder still performs as expected. It appears to but the decoder used to skip the ping events but now includes them that might be an issue if they are json formatted, the test passes because it's just a timestamp, but the anthropic docs don't show the format for the original version so I can't 100% confirm.  This will be something to keep an eye on after rollout.

philipp-spiess: >  The only other thing I didn't check was hitting the legacy gateway api to make sure the decoder still performs as expected.

I tested that case, too. I even tried very long generations and it was all fine. 

I am not too worried that this decoder can't handle both cases: Remember in the first version of this PR, I had the SG backend still support the `/complete` API and all tests still passed. I haven't changed the decoder since.

If you want though we can remove all events of type `ping` just in case. That should let us sleep better at night 😃

chwarwick: my concern was more random events like ping or others that are json strings that then end up in the default switch case that errors, but I assume those would be happening already if they existed.

philipp-spiess: @chwarwick Yeah. And the parser we use for CG is already the same one I just removed the fork since it was only the `\r\n\r\n` vs `\n\n` deliminator that I had to change IIRC

PR #60955, opened 2024-03-08
Integrate the new Triton-server hosted Embeddings model

rafax:
This PR integrates the embeddings model hosted at https://embeddings-triton.sgdev.org/ with Cody Gateway, registering it as an additional model for `/v1/embeddings` endpoint.

Current shortcomings:

- no backend auth (upstream doesn't support it)
- uses staging environment of Triton (until we have prod)
- no rate-limiting (for testing we do not worry about usage)
- uses in-line model allowlisting until [allowlist rolls out on sourcegraph.com](https://github.com/sourcegraph/sourcegraph/pull/60956)

## Test plan

- added e2e test
- local testing


cmd/cody-gateway/internal/httpapi/embeddings/sourcegraph.go
```diff
@@ -0,0 +1,122 @@
...
+	var tritonResponse tritonResponse
+	err = s.json.NewDecoder(resp.Body).Decode(&tritonResponse)
+	if err != nil {
+		return nil, 0, err
+	}
+
+	res := codygateway.EmbeddingsResponse{
+		Embeddings:      make([]codygateway.Embedding, items),
+		Model:           request.Model,
+		ModelDimensions: 768,
```
eseliger, line 80: nit:

```suggestion
		ModelDimensions: modelDimensions,
```

eseliger (approved):
Nice!

cmd/cody-gateway/internal/httpapi/embeddings/sourcegraph.go
rafax: Thank you for catching this :) But, I really want a "Cody PR buddy" that will catch similar things

PR #60958, opened 2024-03-08
Extract reposcheduler and add support for syntactic indexes

keynmol:
Part of #59795
Fixes ENG-23423

This PR's main goal is to extract the logic that selects repositories that should be scanned for new commits-candidates for syntactic indexing.

As part of this work, I've extracted this specific functionality into a new component, called reposcheduler, to reduce the area of responsibility assumed by auto-indexing service.

This is a relatively minor refactoring, mostly invisible from the outside.

The implementation of reposcheduling logic was mostly ported over, with some changes to SQL queries.

We are using a new table, instead of a `index_type` extra field on the `lsif_last_index_scan` table, because there is currently no way to introduce a composite uniqueness constraint (repository_id, index_type) in a backwards compatible way.

This PR will be squashed and merged. Old commits are left in PR history to refer to previous approaches.

## Test plan

- Existing tests that cover precise indexing 
- A copy of those tests that cover syntactic testing in isolation
- Table tests that operate on shared tests and prove isolation of precise and syntactic schedulers


internal/codeintel/autoindexing/internal/background/scheduler/job_scheduler.go
```diff
@@ -128,11 +130,11 @@ func (b indexSchedulerJob) handleScheduler(
 		errMu sync.Mutex
 	)
 
-	for _, repositoryID := range repositories {
+	for _, repository := range repositories {
 		if err := sema.Acquire(ctx, 1); err != nil {
 			return err
 		}
-		go func() {
+		go func(repositoryID int) {
```
mmanela, line 137: New to go, out of curiosity, why change to pass repoid into the goroutine versus using the lexical closure of `repository` and just grabbing .ID in the method?

keynmol: I have asked our local Council of Gophers and this _used_ to be necessary (https://pauldigian.com/advanced-go-loops-and-goroutines)

But we very recently upgraded to 1.22 which no longer has this problem: https://tip.golang.org/doc/go1.22

internal/codeintel/reposcheduler/store_observability.go
```diff
@@ -0,0 +1,47 @@
...
+			observationCtx.Registerer,
+			"codeintel_autoindexing_store",
+			metrics.WithLabels("op"),
+			metrics.WithCountHelp("Total number of method invocations."),
+		)
+	})
+
+	op := func(name string) *observation.Operation {
+		return observationCtx.Operation(observation.Op{
+			Name:              fmt.Sprintf("codeintel.reposcheduler.store.%s", name),
```
varungandhi-src, line 34: Should these two names be more similar/do you know what both of these are doing?

internal/codeintel/reposcheduler/store.go
```diff
@@ -0,0 +1,289 @@
...
+// repositories known to the instance.
+func (store *store) GetRepositoriesForIndexScan(
+	ctx context.Context,
+	batchOptions RepositoryBatchOptions,
+	now time.Time,
+) (_ []int, err error) {
+	var repositoryMatchLimitValue int
+	if batchOptions.RepositoryMatchLimit != nil {
+		repositoryMatchLimitValue = *batchOptions.RepositoryMatchLimit
+	}
```
varungandhi-src, line 119: Here, the `0` will be translated to a `LIMIT 0` in the query right? Is that deliberately allowed/possible somehow in some site configuration?

internal/codeintel/reposcheduler/store.go
```diff
@@ -0,0 +1,289 @@
...
+	queries = append(queries, sqlf.Sprintf(getRepositoriesForIndexScanRepositoriesWithPolicyViaPatternQuery(store.dbLayout)))
+
+	for i, query := range queries {
+		queries[i] = sqlf.Sprintf("(%s)", query)
+	}
+
+	query := getRepositoriesForIndexScanQuery(store.dbLayout)
+
+	fmt.Println(query)
+
```
varungandhi-src, line 145:
```suggestion
```

Maybe you can annotated debug lines with some comment `// REVIEW: remove this` as a reminder 😅

varungandhi-src (commented):
I did a first round visiting 20/30 files, will do a second round later since this PR is quite big. 😬

internal/codeintel/reposcheduler/store.go
keynmol: I wonder if we could have a linter that detects that - there's probably very few cases where we want prints instead of logging, and it should save time on the review

internal/codeintel/autoindexing/init.go
```diff
@@ -49,17 +50,18 @@ func NewIndexSchedulers(
 	uploadSvc UploadService,
 	policiesSvc PoliciesService,
 	policyMatcher PolicyMatcher,
-	autoindexingSvc *Service,
+	repoSchedulingSvc reposcheduler.RepositorySchedulingService,
+	autoindexingDeps AutoIndexingService,
```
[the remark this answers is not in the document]
keynmol: > it's unclear why there is Deps here and Svc elsewhere

Ah, this is left over from trying to break this weird AutoIndexingService apart, but IIRC I hit on cyclic dependencies and resolving them would require more changes than this PR could handle.

I will revisit in a separate PR.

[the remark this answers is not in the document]
varungandhi-src: This question is about a local variable name, not sure why there would be cyclic dependencies involved? Did you mean to post this reply to one of my other comments?

internal/codeintel/reposcheduler/store_test.go
```diff
@@ -0,0 +1,323 @@
...
+	assertRepoList(t, store, NewBatchOptions(time.Hour, true, nil, 100), now.Add(time.Minute*30), []int(nil))
+
+	// 90 minutes later, all repositories are visible
+	assertRepoList(t, store, NewBatchOptions(time.Hour, true, nil, 100), now.Add(time.Minute*89), []int{50, 51, 52, 53})
+
+	// Make new invisible repository
+	insertRepo(t, db, 54, "r4")
+
+	// 95 minutes later, new repository is not yet visible
+	assertRepoList(t, store, NewBatchOptions(time.Hour, true, nil, 100), now.Add(time.Minute*95), []int(nil))
```
varungandhi-src, line 81: 1. It looks like the comments are out-of-sync with the actual `assertRepoList` lines.
2. Could you explain the significance of the time instants chosen here? I don't understand what the numbers 20, 30, 90 and 95 are derived based on.

internal/codeintel/reposcheduler/store_test.go
```diff
@@ -0,0 +1,323 @@
...
+
+	// 110 minutes later, updated repositories are ready for re-indexing
+	assertRepoList(t, store, NewBatchOptions(time.Hour, true, nil, 100), now.Add(time.Minute*110), []int{50})
+}
+
+/*
+This set of assertions verifies that repo scheduling logic works with syntactic
+indexing in isolation - i.e. we assume that there is no overlapping precise/syntactic scheduling running.
+For that we have a separate test.
+*/
```
varungandhi-src, line 108: Looks like this comment was copy-pasted from the comment on the above test - should this say `precise` instead of `syntactic`?

Also, it seems like the bulk of the logic between these tests is the same -- thoughts on factoring it out into a common function? (or do we expect that we will need to evolve these tests differently?)

internal/database/schema.md
```diff
@@ -2122,6 +2122,7 @@ Stores data points for a code insight that do not need to be queried directly, b
  repository_patterns         | text[]                   |           |          | 
  last_resolved_at            | timestamp with time zone |           |          | 
  embeddings_enabled          | boolean                  |           | not null | false
+ syntactic_indexing_enabled  | boolean                  |           |          | true
```
varungandhi-src, line 2125: Does making this true by default from the get-go create a risk of spawning a bunch of jobs that aren't processed to the end? During our initial discussion, I believe we agreed on doing a gradual rollout where we enable it for certain repos at the start, see how it works (or doesn't) and then gradually increase the number of repos, before making it on-by-default.

migrations/frontend/squashed.sql
```diff
@@ -1721,7 +1721,8 @@ CREATE TABLE lsif_configuration_policies (
     protected boolean DEFAULT false NOT NULL,
     repository_patterns text[],
     last_resolved_at timestamp with time zone,
-    embeddings_enabled boolean DEFAULT false NOT NULL
+    embeddings_enabled boolean DEFAULT false NOT NULL,
+    syntactic_indexing_enabled boolean DEFAULT true
```
varungandhi-src, line 1725: Is there a reason for this to be nullable/what does the null state mean?

varungandhi-src (changes requested):
Mostly LGTM, would be good to have more clarity on the test logic and the decision to set the boolean to be on by default.

internal/database/schema.md
keynmol: Here's what I was thinking:

1. This particular field will only decide whether repositories will be _ever_ marked as candidates for syntactic scheduling. This current PR will just update and maintain the syntactic indexing last scan table, not spawn any jobs
2. For jobs to actually be spawned, it needs a worker 
3. Before I PR the worker in, I want to introduce a site config setting (disabled by default) which enables/disables syntactic indexing - this setting will also enable/disable UI toggle that will control this particular boolean field

The site config setting will obviously be enabled on demo and S2, where we can then further control the rollout of this. Customers and dotcom won't have any jobs scheduled.

I don't have a strong opinion about exact value of this field in context of this PR, so can happily revert back to false.

keynmol: > This current PR will just update and maintain the syntactic indexing last scan table, not spawn any jobs

Oh and also that will only happen upon the invoking the syntactic version of this code, which the worker does.

Anyways, for sanity reasons I'll revert it to false.

varungandhi-src: 👍🏽 Yeah, I think having it as `false` by default is safer so that when the site config setting is enabled (e.g. we change the default), you still need to have a policy for syntactic indexing (so that we can gradually enable it for specific repos).

internal/codeintel/autoindexing/init.go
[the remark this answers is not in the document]
keynmol: I'm responding to "and it's unclear why there is Deps here and Svc elsewhere" specifically - previously I made an attempt to break apart and refactor the Service + renamed it to Deps, and then reverted that change because I introduced cyclical dependencies.

The variable name will be changed.

migrations/frontend/squashed.sql
keynmol: No reason, it's an oversight - I focused on the existing data, and forgot to add a constraint for data going forward, fixed in https://github.com/sourcegraph/sourcegraph/pull/60958/commits/81a3d53c6f012b5936953e3a1e4859a9d4b67d00

internal/codeintel/reposcheduler/store_test.go
keynmol: @varungandhi-src At this point we don't expect them to evolve differently, so I converted duplicated tests into a table test and added hopefully better comments to all the code I copied over from the code before this PR, see https://github.com/sourcegraph/sourcegraph/pull/60958/commits/6f66b24673e052303875d4b63188c2ecf0b26665 for comments and https://github.com/sourcegraph/sourcegraph/pull/60958/commits/57ccc2072cda6fdcc0b5b7786c97212896abc1cb for refactor

internal/codeintel/reposcheduler/store_test.go
keynmol: Addressed hopefully in https://github.com/sourcegraph/sourcegraph/commit/6f66b24673e052303875d4b63188c2ecf0b26665 - let me know if more clarity is needed

internal/codeintel/reposcheduler/store.go
keynmol: Technically there's nothing wrong with a 0 value for this particular setting - the setting is computed here: https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/internal/conf/computed.go?L315-322 and it only handles the `-1` case to explicitly identify "no limit" setting.

This particular limit only controls the number of repositories returned from matching a global policy - which can match _all_ repos on an instance and therefore should be tightly controlled. Setting limit to 0 would be equivalent to disallowing global policy altogether, but as it stands 0 is not an erroneous value - albeit a useless one.

cmd/worker/internal/codeintel/autoindexing_scheduler.go
```diff
@@ -52,7 +52,8 @@ func (j *autoindexingScheduler) Routines(_ context.Context, observationCtx *obse
 		services.UploadsService,
 		services.PoliciesService,
 		matcher,
-		services.AutoIndexingService,
+		services.PreciseRepoSchedulingService,
+		*services.AutoIndexingService,
```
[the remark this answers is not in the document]
keynmol: I took a closer look, and it's a mess.

The code for various internal services is intermingled with what graphql resolvers need.

For now I reverted the changes here, we can take another pass at refactoring this - I took three attempts as part of this PR and don't feel like it's worth it anymore.

internal/codeintel/reposcheduler/init.go
```diff
@@ -0,0 +1,33 @@
...
+
+type RepositoryToIndex struct {
+	ID int
+}
+
+type RepositorySchedulingService interface {
+	GetRepositoriesForIndexScan(
+		ctx context.Context,
+		_ RepositoryBatchOptions,
+		now time.Time) (_ []RepositoryToIndex, err error)
```
varungandhi-src, line 16:
```suggestion
	GetRepositoriesForIndexScan(ctx context.Context, _ RepositoryBatchOptions, now time.Time) ([]RepositoryToIndex, error)
```

Minor tweak to match typical formatting style

internal/codeintel/reposcheduler/options.go
```diff
@@ -0,0 +1,33 @@
...
+	// If allowGlobalPolicies is false, then configuration policies that define neither a repository id
+	// nor a non-empty set of repository patterns wl be ignored.
+	// When true, such policies apply over all repositories known to the instance.
+	AllowGlobalPolicies bool
+
+	// This optional limit controls how many repositores will be considered by matching
+	// via global policy. As global policy matches large sets of repositories,
+	// this limit allows reducing the number of repositories that will be considered
+	// for scanning.
+	GlobalPolicyRepositoriesMatcLimit *int
```
varungandhi-src, line 20:
```suggestion
	GlobalPolicyRepositoriesMatchLimit *int
```

internal/codeintel/reposcheduler/options.go
```diff
@@ -0,0 +1,33 @@
...
+type RepositoryBatchOptions struct {
+	// Time to wait before making repository visible again
+	// after previous scan. Note that this delay can be superseded if repository
+	// has been changed in the meantime
+	ProcessDelay time.Duration
+
+	// If allowGlobalPolicies is false, then configuration policies that define neither a repository id
+	// nor a non-empty set of repository patterns wl be ignored.
+	// When true, such policies apply over all repositories known to the instance.
+	AllowGlobalPolicies bool
```
varungandhi-src, line 14:
```suggestion
	// If AllowGlobalPolicies is false, then configuration policies that do not specify
	// a repository name or patterns will be ignored.
	// When true, such policies apply over all repositories known to the instance.
	AllowGlobalPolicies bool
```

Do we literally allow site admins to specify a repository id that is an integer? Or does "id" just here mean the base case of a literal pattern? If it's a string, might be clearer to say "repository name".

internal/codeintel/reposcheduler/store.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+var _ RepositorySchedulingStore = &store{}
+
+type StoreType int8
+
+const (
+	Precise   StoreType = 1
+	Syntactic StoreType = 2
+)
```
varungandhi-src, line 42:
```suggestion
type storeType int8

const (
	preciseStore   StoreType = 1
	syntacticStore StoreType = 2
)
```

Suggestion: Make these private, since AFAICT, the values are only used within this same package.

internal/codeintel/reposcheduler/store_test.go
```diff
@@ -0,0 +1,303 @@
...
+			// T + 20 minutes: first two repositories are still on cooldown
+			assertRepoList(t, tt.store, NewBatchOptions(1*time.Hour, true, nil, 100), now.Add(time.Minute*20), []int{52, 53})
+
+			// T + 30 minutes: all repositories are on cooldown
+			assertRepoList(t, tt.store, NewBatchOptions(1*time.Hour, true, nil, 100), now.Add(time.Minute*30), []int(nil))
+
+			// T + 90 minutes: all repositories are visible again
+			// Repos 50, 51 are visible because they were scheduled at (T + 0) - which is more than 1 hour ago
+			// Repos 52, 53 are visible because they were scheduled at (T + 20) - which is more than 1 hour ago
+			assertRepoList(t, tt.store, NewBatchOptions(1*time.Hour, true, nil, 100), now.Add(time.Minute*90), []int{50, 51, 52, 53})
```
varungandhi-src, line 112: Thank you, now this is much more understandable. I think part of what was throwing me off was the names. Here, the function name starts with `assert` and the query this bottoms out is `getRepositoriesForIndexScanQuery`. However, that query modifies state, and consequently, this `assertBlah` function also modifies state, whereas normally you'd expect 'get' and 'assert' to not modify state.

I don't have specific alternate naming suggestions, but the usage of `get` for a query which does inserts seems a bit unfortunate.

varungandhi-src (approved):
Thanks for all the changes. The test cases in particular are much more easily understandable.

internal/codeintel/reposcheduler/store_test.go
keynmol: Yeah, the upsert nature of that operation is a bit confusing - also makes it hard to debug.
The insert is pushed into the query for (I believe) atomicity and to handle possible dotcom batch sizes (though I wonder how much of a problem it is to literally return only integer repository IDs, you can do thousands of them)

We can rename it to Calculate/Compute, and replace this single helper method assert with a combo of assert + a local `computeReposForScheduling` or something, but I think we'd hit diminishing returns wrt readability

PR #60974, opened 2024-03-08
sg/msp: add `bigqueryconnection.googleapis.com` api

jac:
Adds bigquery connection API to allow for connecting big query to cloudsql
## Test plan

CI, local generated TF, applied to testbed


akalia25 (approved):

bobheadxi:
FYI @jac for one-off GCP API enablement we can also use https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_project_service, which is what this var is used to generate

PR #60976, opened 2024-03-08
svelte: Fix main navigation dropdown hidden behind search input

fkling:
Due to z-index setting, the search dropdown of the main navigation is hidden behind the search input on the search results page.

This commit solves this by introducing stacking contexts at the appropriate level:

- In the search input, the actual input needs a higher z-index to appear above the suggestion container. This only needs to work inside the search input so it should create a new stacking context.
- In the search results page, the (full) search input should appear above the search results list. Using `position: sticky` would cause the headsers to appear above the search input suggestions. Giving the search input container a higher z-index fixes that. But again, that only needs to work inside the search results itself. Because the search results page doesn't have its own container I added `isolation: isolate` to the `<main />` element in the top level layout. I think that makes sense conceptually.
- melt-ui appends the dropdown to the bottom of the page so it will now naturally render above all other elements.

I added a test to verify that this works (the test failed without the CSS changes). For that I also added a rudimentary mocking API for search results, otherwise I was getting errors.

Before:

After: 

## Test plan

Manual testing and new playwright test

### DIFF
--- a/client/web-sveltekit/src/lib/search/input/SearchInput.svelte
+++ b/client/web-sveltekit/src/lib/search/input/SearchInput.svelte
@@ -223,11 +223,9 @@
     @use '$lib/breakpoints';
 
     form {
+        isolation: isolate;
         width: 100%;
         position: relative;
-        // Necessary to ensure that the search input (especially the suggestions) are rendered above sticky headers
-        // in the search results page ("position: sticky" creates a new stacking context).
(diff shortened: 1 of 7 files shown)


camdencheek (approved):
Makes sense to me! And yay for rudimentary search stream mocking. I still plan to filters tests at some point, so that'll make my job a bit easier.

PR #60982, opened 2024-03-09
Revert "web: Trigger page refresh to load SvelteKit for supported/enabled routes"

camdencheek:
Reverts sourcegraph/sourcegraph#60872

Git blame view seems to be broken because of this PR, and it spams the console with a bunch of scary looking stuff so I'm not confident there aren't other things broken as well. The revert applies cleanly so I figure it gives us some time to take a closer look on Monday. 

# Test plan

Tested locally that git blame works


DaedalusG (approved):

fkling:
It's the same problem that was fixed in https://github.com/sourcegraph/sourcegraph/pull/60967. The tl;dr is that we create multiple react roots to render React inside codemirror and we haven't made the apollo client available to those. That hasn't been necessary so far but with the new Link component making graphql requests it's necessary now.

PR #60991, opened 2024-03-11
telemetrygateway: update schema to accommodate managed services

bobheadxi:
Expands the Telemetry Gateway service schema to accommodate events from managed services, as proposed in [RFC 897](https://docs.google.com/document/d/1OberM_s04DdkCmo8WUN9mrjn5agNNoGtVKYESQyiuxo/edit?tab=t.0#heading=h.aai9y6ehm647). The idea is that managed services _backends_ will publish events to the same gRPC Telemetry Gateway service that single-tenant Sourcegraph do today. The changes include:

1. A new RPC, `RecordEvent`, for publishing events one at a time. This will be much easier to work with for new services, as the bidirectional streaming RPC is hard to hold right and is overkill for low-volume exporters.
2. A new `Identifier`, `ManagedService`, which managed services can use to identify as e.g. `service_id: ssc, service_environment: prod`.
3. A new `EventUser` identifier type, `optional string sams_external_id`, for associating an event with a SAMS user. We only expect this from managed services.
4. Docstring changes for clearer references to "managed services" and "single-tenant Sourcegraph".

Each service will implement its own mechanisms for how simple events are submitted by its clients, enriched by the backend, and then published to Telemetry Gateway. Pretty much everything else can be treated the same way as in Sourcegraph instances, but specific to the exporting service's context.

**This change does _not_ include any implementation yet** - in particular, the new `RecordEvent` RPC will return an unimplemented error. This is implemented in https://github.com/sourcegraph/sourcegraph/pull/61022

## Test plan

n/a


jac (approved):

dadlerj (approved):
Amazing thanks

bobheadxi:
* **#61022** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/61022?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#60991** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/60991?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

unknwon (approved):

internal/telemetrygateway/v1/telemetrygateway.proto
```diff
@@ -169,15 +209,24 @@ message EventBillingMetadata {
...
+  // often used for unauthenticated users, but can persist to authenticated
+  // users as well.
   optional string anonymous_user_id = 2;
+  // Sourcegraph Accounts Management System (SAMS) account associated with the
+  // user. This is only valid for services leveraging SAMS as an identity
+  // provider - on other words, traditional Sourcegraph instances will not
+  // provide this.
+  //
+  // Learn more about SAMS: https://handbook.sourcegraph.com/departments/engineering/teams/core-services/sams
+  optional string sams_user_id = 3;
```
chrsmith, line 229: For precision, this should be the "SAMS account external ID". (`sams_external_id`?) While it's probably fine to consider this the "user ID", strictly speaking it is not. (Internally, the SAMS database has a user ID field that is something totally different. And all SAMS APIs expose the stable ID expected here as an "external ID".)

chrsmith (approved):
Looks good to me. 🙏 for making a non-streaming version of the service, since support for streaming on gRPC is spotty across platforms.

The only hit pick is just the name of the SAMS ID, and how pedantic we want to be. Though @unknwon 's preference should override any opinions of mine here.

internal/telemetrygateway/v1/telemetrygateway.proto
unknwon: Equally confusing due to historical debt: it is 100% true calling something `user_id` in our system, for the past ten years, we just assumed that it's a numeric ID, or sometimes called them database ID. Technically speaking, `sams_user_id` is correct because that's what SAMS exposes as the user ID after all. Now, due to our history, would it have more clarify to everyone by naming it `sams_user_external_id`, 100%. 😂

bobheadxi: Good call-out - I shall rename this to `sams_external_id`, and update docstrings accordingly 😁

bobheadxi:
Merging so that @chrsmith can get started if he'd like - came down sick this afternoon so will probably wrap up https://github.com/sourcegraph/sourcegraph/pull/61022 tomorrow

PR #60992, opened 2024-03-11
graphqlbackend: allow siteadmins to view roles on dotcom

bobheadxi:
Currently, listing roles associated with a user on dotcom fails with the error `roles are not available on sourcegraph.com`. However, roles _are_ available on dotcom and you can configure them for users (e.g. Entitler, #60795), but there's no way to check what roles a user has with  the current guard.

This relaxes the restriction to allow site admins to view a user's roles.

## Test plan

Test case

### DIFF
--- a/cmd/frontend/graphqlbackend/roles_test.go
+++ b/cmd/frontend/graphqlbackend/roles_test.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/sourcegraph/log/logtest"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/sourcegraph/sourcegraph/cmd/frontend/graphqlbackend/apitest"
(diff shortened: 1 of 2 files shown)


BolajiOlajide (approved):
This works for the backend but I think we have a guard on the frontend if we ever want to display roles in the user profile.

I'm not sure it affects what you're working on but wanted to bring it to your attention.
https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/client/web/src/user/area/UserArea.tsx?L56%3A9-61%3A10

Asides that, the changes look good to me from a code PoV.

evict (approved):

bobheadxi:
> This works for the backend but I think we have a guard on the frontend if we ever want to display roles in the user profile.
>
> I'm not sure it affects what you're working on but wanted to bring it to your attention.
https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/client/web/src/user/area/UserArea.tsx?L56%3A9-61%3A10

TIL `@skip` in GraphQL! But yeah, I don't think I'll change the web app - for administrative purposes in dotcom it's sufficient probably to be able to see it via API directly, and making this work in the frontend probably requires adding another check for siteadmin as well otherwise the query will fail

PR #60996, opened 2024-03-11
Removed unused bazel macro arg

jamesmcnamara:
The `npm_package` macro had an unused argument that it was silently dropping before passing on, making any call site that used very misleading. Also fixed a small bug with running blobstore through bazel.

## Test plan
Ran `bazel test //...`

### DIFF
--- a/client/build-config/BUILD.bazel
+++ b/client/build-config/BUILD.bazel
@@ -55,5 +55,4 @@ npm_package(
         "package.json",
         ":build-config_lib",
     ],
-    type = "commonjs",
 )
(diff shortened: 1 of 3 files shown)


burmudar (approved):
🚗 drive by approve

fkling:
Ah, that's what this was for... I removed that lines from the other PRs that I didn't know what they were for or were causing issues 😬

jamesmcnamara:
@fkling Yeah it was unused but you were correct that it was still causing failures because there was at least one case of a caller trying to use it. This closes that loop.

PR #60999, opened 2024-03-11
svelte: Enable autofocus on search homepage

### LINKED ISSUE #60836: Autofocus search input

fkling:
Closes #60836 

This enables autofocus on the search homepage in a similar way as was done in the React version.

Some additional changes:

- `@melt-ui/svelte` was update to the latest version to bring in a focus bugfix. This bug caused focus to be stolen from other elements when dropdown menus were mounted. This prevent the autofocusing the search input.
- `afterNavigate` is used to focus the element, as suggested in the SvelteKit documentation. We also focus it after the input is created, in case it's used in a situation where the input is shown in response to some other user interaction.
- Some refactoring of the search input component to avoid unnecessarily recreating extensions.

## Test plan

Manual testing and new playwright test

### DIFF
--- a/client/web-sveltekit/package.json
+++ b/client/web-sveltekit/package.json
@@ -65,7 +65,7 @@
   },
   "type": "module",
   "dependencies": {
-    "@melt-ui/svelte": "^0.66.2",
+    "@melt-ui/svelte": "^0.76.0",
     "@popperjs/core": "^2.11.8",
     "@sourcegraph/branded": "workspace:*",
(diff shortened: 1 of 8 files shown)


fkling:
Forgot to add that autofocus doesn't work when clicking on 'search home' in the main nav. I think that's because the focus goes back to the trigger for the dropdown menu (which is the accessibility compliant behavior). So we have conflicting focus behavior here and need to figure out how we want it to work.

Issue #60934, opened 2024-03-07
Cody: Azure OpenAI API version updates


chwarwick:
Update your code to use the latest Azure OpenAI Service preview API by 2 April 2024
You're receiving this email because you've used an older Azure OpenAI Service preview API in the last 30 days.

We recently made Azure OpenAI API [version 2024-02-15-preview](https://learn.microsoft.com/azure/ai-services/openai/api-version-deprecation) available. You'll need to update any code that uses the following older versions before they're [retired](https://learn.microsoft.com/azure/ai-services/openai/api-version-deprecation) on 2 April 2024:

2023-03-15-preview
2023-07-01-preview
2023-08-01-preview
2023-09-01-preview
2023-10-01-preview
2023-12-01-preview
Required action
To avoid potential service disruptions, [update your code](https://learn.microsoft.com/azure/ai-services/openai/reference) that uses any of the retiring Azure OpenAI API versions to use version 2024-02-15-preview or newer by 2 April 2024.

PR #61005, opened 2024-03-11
Cody: Update Azure OpenAI client and force api-version to a stable one

chwarwick:
This updates the Azure OpenAI client and also overrides the round tripper to force it to use the stable 2023-05-15 api version instead of the default which uses preview versions that keep being retired.

closes https://github.com/sourcegraph/sourcegraph/issues/60934
## Test plan
Verify chat works and uses correct api version:
```
[       frontend] [Mar 11 17:37:11.359346] Response: ==> REQUEST/RESPONSE (Try=1/1.327581208s, OpTime=1.327591292s) -- RESPONSE RECEIVED
[       frontend]    POST https://{DEV_ENVIRONMENT}/openai/deployments/gpt-4-test/chat/completions?api-version=2023-05-15
```
Verify autocomplete works and uses correct api version
```
[       frontend] [Mar 11 17:38:03.682053] Response: ==> REQUEST/RESPONSE (Try=1/504.029208ms, OpTime=504.055333ms) -- RESPONSE RECEIVED
[       frontend]    POST https://sourcegraph-test-oai.openai.azure.com/openai/deployments/gpt-35-turbo-instruct-test/completions?api-version=2023-05-15
```


internal/completions/client/azureopenai/openai.go
```diff
@@ -52,10 +56,17 @@ func GetAPIClient(endpoint, accessToken string) (CompletionsClient, error) {
 	apiClient.mu.RUnlock()
 	apiClient.mu.Lock()
 	defer apiClient.mu.Unlock()
+
+	// API Versions and docs https://learn.microsoft.com/en-us/azure/ai-services/openai/reference#completions
+	clientOpts := &azopenai.ClientOptions{
+		ClientOptions: azcore.ClientOptions{
+			Transport: apiVersionClient("2023-05-15"),
+		},
+	}
```
eseliger, line 65: in case we update the client again, could it be that their response parser breaks if it doesn't support that version anymore?

Not a blocker, this is still a good change, just a little scared. 
Maybe we could use our HTTP recorder tests here to ensure that a newer version of the library is still able to parse the response alright?

internal/completions/client/azureopenai/openai.go
```diff
@@ -373,3 +385,52 @@ func toStatusCodeError(err error) error {
...
+
+type apiVersionRoundTripper struct {
+	rt         http.RoundTripper
+	apiVersion string
+}
+
+func (rt *apiVersionRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	// Copy the request and URL since modifying is not permitted
+	newReq := req
+	var newUrl url.URL = *req.URL
```
eseliger, line 397: I think this way of cloning a URL is unsafe because some subfields are pointers.

There's a `Clone` method on `*http.Request` that handles this gracefully.

internal/completions/client/azureopenai/openai.go
eseliger: I see this ClientOptions struct has a field called `APIVersion` - I assume this is not useful for us?

chwarwick: > I see this ClientOptions struct has a field called `APIVersion` - I assume this is not useful for us?

Correct assumption: `Request failed: this client doesn't support overriding its API version`

eseliger: Ah, unfortunate :( if the method looked like this, it would be configurable by us :( 

```go
// NewClientWithKeyCredential creates a new instance of Client that connects to an Azure OpenAI endpoint.
//   - endpoint - Azure OpenAI service endpoint, for example: https://{your-resource-name}.openai.azure.com
//   - credential - used to authorize requests with an API Key credential
//   - options - client options, pass nil to accept the default values.
func NewClientWithKeyCredential(endpoint string, credential *azcore.KeyCredential, options *ClientOptions) (*Client, error) {
	if options == nil {
		options = &ClientOptions{
// NEW:
			ClientOptions: azcore.ClientOptions{
				APIVersion: "2023-09-23",
			},
		}
	}

	authPolicy := runtime.NewKeyCredentialPolicy(credential, "api-key", nil)
	azcoreClient, err := azcore.NewClient(clientName, version, runtime.PipelineOptions{
// NEW:
		APIVersion: runtime.APIVersionOptions{Location: runtime.APIVersionLocationQueryParam, Name: "api-version"},
		PerRetry:   []policy.Policy{authPolicy},
	}, &options.ClientOptions)
	if err != nil {
		return nil, err
	}

	return &Client{
		internal: azcoreClient,
		clientData: clientData{
			endpoint: endpoint,
			azure:    true,
		},
	}, nil
}
```

chwarwick: > in case we update the client again, could it be that their response parser breaks if it doesn't support that version anymore?

Totally this is far from ideal but better then reimplementing service principals, workload identity and the others.  I'll add a backlog item to add a recorder test, it's a bit of a pain to get the format of the events.

eseliger: works! thanks!

internal/completions/client/azureopenai/openai.go
```diff
@@ -373,3 +385,49 @@ func toStatusCodeError(err error) error {
...
+	if http2Transport, err := http2.ConfigureTransports(azureClientDefaultTransport); err == nil {
+		http2Transport.ReadIdleTimeout = 10 * time.Second
+		http2Transport.PingTimeout = 5 * time.Second
+	}
+
+	return &http.Client{
+		Transport: &apiVersionRoundTripper{
+			rt:         azureClientDefaultTransport,
+			apiVersion: apiVersion,
+		},
```
eseliger, line 431: one last thing, since this doesn't wrap our httpcli transport, these requests won't be counted accordingly and cannot appear in the outgoing request logs. Not a blocker for now, but would be nice to wrap that in here.

eseliger (approved):

internal/completions/client/azureopenai/openai.go
chwarwick: Not adding for now since it's not a regression i'll add to the follow up issue along with parsing tests.

PR #61006, opened 2024-03-11
gitserver: Implement ls-tree operations on the server-side

eseliger:


cmd/frontend/graphqlbackend/git_commit.go
```diff
@@ -320,7 +320,19 @@ func (*rootTreeFileInfo) Size() int64        { return 0 }
...
+	}
+
+	names := make([]string, len(fds))
+	for i, fd := range fds {
+		if fd.IsDir() {
+			continue
+		}
+		names[i] = fd.Name()
+	}
+	return names, nil
```
eseliger, line 335: Simplified this to just be a readdir function, it has some additional overhead but the number of dirs should usually be much smaller than the number of files, and this resolver is already quite worrisome, due to the amount of data it'll return for large monorepos.

cmd/gitserver/internal/git/gitcli/odb.go
eseliger, line 1: Note: `lsTree` used to do `fileutil.SortFileInfosByName(fis)`, but this requires to read ALL entries into memory so it doesn't scale. Thus, I removed it.

internal/gitserver/client.go
```diff
@@ -450,9 +449,6 @@ type Client interface {
 	// all commits reachable from HEAD.
 	CommitsUniqueToBranch(ctx context.Context, repo api.RepoName, branchName string, isDefaultBranch bool, maxAge *time.Time) (map[string]time.Time, error)
 
-	// LsFiles returns the output of `git ls-files`.
-	LsFiles(ctx context.Context, repo api.RepoName, commit api.CommitID, pathspecs ...gitdomain.Pathspec) ([]string, error)
```
eseliger, line 454: LsFiles is a command mainly meant for interacting with the index, which doesn't exist in bare repos.

This is now ReadDirPatterns which is a derivative of ReadDir, both using ls-tree under the hood.

internal/gitserver/client.go
```diff
@@ -411,11 +415,6 @@ type Client interface {
 	// ListTags returns a list of all tags in the repository. If commitObjs is non-empty, only all tags pointing at those commits are returned.
 	ListTags(ctx context.Context, repo api.RepoName, commitObjs ...string) ([]*gitdomain.Tag, error)
 
-	// ListDirectoryChildren fetches the list of children under the given directory
-	// names. The result is a map keyed by the directory names with the list of files
-	// under each.
-	ListDirectoryChildren(ctx context.Context, repo api.RepoName, commit api.CommitID, dirnames []string) (map[string][]string, error)
```
eseliger, line 417: This is basically ReadDir with some mapping function at the end, I moved the mapping to the only caller that made use of this, and made it use ReadDir instead to not expose more APIs than needed (and needed to be maintained).

internal/gitserver/commands.go
```diff
@@ -455,293 +450,98 @@ func (c *clientImplementor) ReadDir(ctx context.Context, repo api.RepoName, comm
...
 		if err != nil {
 			return nil, err
 		}
-
-		// It's only worthwhile to cache if the operation took a while and returned a lot of
-		// data. This is a heuristic.
-		if time.Since(start) > 500*time.Millisecond && len(entries) > 5000 {
-			lsTreeRootCacheMu.Lock()
-			lsTreeRootCache.Add(key, entries)
-			lsTreeRootCacheMu.Unlock()
```
eseliger, line 527: This client-side cache is gone, we instead can implement this in the backend. But I don't see huge value in that, instead we should not be requesting 10s of thousands of dirs at once ideally, and instead traverse using individual API calls like reading a real filesystem.

PR #61007, opened 2024-03-11
oidc: Rename singleProvider property to singleIdentityPerUser

eseliger:
This has not been used anywhere yet and I've gotten feedback that the name was confusing. So changing it here.

## Test plan

Existing test suites, code review.


chrsmith (approved):
I agree the new name is clearer. Do we need to document this flag somewhere else as well? As this would be a new configuration knob for site-admins?

eseliger:
The docs page renders the JSON schema including inline comments, I think that's enough for now? I don't really want to encourage anyone to use it if they don't have to, since it's currently only tailored for dotcom 😅

PR #61009, opened 2024-03-11
Simplify goroutine params

jtibshirani:
Now that we've updated to Go 1.22, we don't need to copy loop variables before
using them in goroutines.

I found these using the regex searches `go func\(\w+` and `\.Go(func\(\w+`. I
also simplified some non-loop vars when it made sense. 

## Test plan

Straight refactor, covered by existing tests

### DIFF
--- a/cmd/frontend/graphqlbackend/observability.go
+++ b/cmd/frontend/graphqlbackend/observability.go
@@ -43,10 +43,10 @@ func (r *schemaResolver) TriggerObservabilityTestAlert(ctx context.Context, args
 	metric.With(nil).Set(1)
 
 	// reset the metric after some amount of time
-	go func(m *prometheus.GaugeVec) {
+	go func() {
 		time.Sleep(1 * time.Minute)
-		m.With(nil).Set(0)
-	}(metric)
+		metric.With(nil).Set(0)
+	}()
 
 	return &EmptyResponse{}, nil
 }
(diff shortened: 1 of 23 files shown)


internal/batches/service/workspace_resolver.go
```diff
@@ -206,10 +206,10 @@ func findIgnoredRepositories(ctx context.Context, gitserverClient gitserver.Clie
 	}
 	close(input)
 
-	go func(wg *sync.WaitGroup) {
+	go func() {
 		wg.Wait()
 		close(results)
-	}(&wg)
```
jtibshirani, line 212: Not sure why we did this, hopefully I'm not missing something 🤔

cmd/frontend/graphqlbackend/observability.go
```diff
@@ -43,10 +43,10 @@ func (r *schemaResolver) TriggerObservabilityTestAlert(ctx context.Context, args
 	metric.With(nil).Set(1)
 
 	// reset the metric after some amount of time
-	go func(m *prometheus.GaugeVec) {
+	go func() {
 		time.Sleep(1 * time.Minute)
-		m.With(nil).Set(0)
-	}(metric)
+		metric.With(nil).Set(0)
+	}()
```
keegancsmith, line 49: heh this one wasn't even in a for loop

internal/batches/service/workspace_resolver.go
keegancsmith: yeah this looks fine.

keegancsmith (approved):
I'm slightly nervous than in some cases we may mutate the outer var and depend on the function capture to create a new variable. An example of this would be changing the ctx. IE not all of these changes are to do with loop variables. However, I tried to review this closely to account for this but may have missed a spot.

jtibshirani:
@keegancsmith thanks for reviewing carefully! I also tried to take a close look at each spot to check for issues. In retrospect, scoping the change just to loop vars would've made sense.

PR #61010, opened 2024-03-11
Cleanup Cody Gateway HTTP endpoint registration code

chrsmith:
> ⚠️ This PR is partly a conversation starter, and so feel free to suggest even better ways to clean up this part of the code. Or, if you believe things are fine as-is, I can drop it entirely. This is just in service of making it easier to land "background request sampling" discussed later.

It appears that over time we've added several common layers of HTTP middleware, but in a way that is increasingly difficult to maintain. This PR makes a few minor changes that lead to a net reduction in overall lines of code.

For reviewing, it may make more sense to look at the PR commit-by-commit.

**Minimize the use between `if` and `else` keywords.**

Something that stuck with me from _Code Complete_, was that if you use an `if/else` construct, and one of the code blocks is really short, you should make that one come "first". So that when you read the code, you can at a glance see the "if X, then Y. Otherwise long-winded-Z".

The code in `handler.go` is a good example of this. Where sometimes you would go more than a "page" of code, before seeing `} else { logger.Error("Anthropic access token not set") }`.

So simply inverting the `if`-predicate and putting the "or log an error" code first, IMHO, makes things easier to read.

**Introducing `registerStandardEndpoint` and `registerSimpleGETEndpoint`**

The next change was to wrap the boilerplate for registering HTTP endpoints (and the various layers of instrumentation, authorization, and performance counters) into a single method.

It's one thing to have duplicate code when you are doing the same thing several times. But I'd argue that the way we register our HTTP endpoints involve _a lot_ of moving parts. (e.g. configuration passed to the various layers of middleware.) So having a "standard" registration method will not only cut down on the boilerplate, but also provide us with a way to make it easier to add more logic/customization.

e.g. maybe we can reconcile `registerStandardEndpoint` and `registerSimpleGETEndpoint` by making the `attribute.Set` parameter optional and adding the HTTP method as a new parameter?

## Do we need this PR?

The main reason why I'm sending this out for review, is to make it easier to add the _next_ PR. Which I am thinking would come as yet another HTTP middleware layer.

As part of doing a better job of identifying and rejecting abusive traffic, I would like to add a new hook for our LLM-based endpoints that will periodically trigger an async "LLM API moderation check". e.g. sample N% of our traffic, sending it to a moderation endpoint async.

My thinking is that we'd call this "moderation check middleware" for all of our LLMs, as well as make the design pluggable so we could use some other 3rd party API for moderating LLM requests. Or even rely on our own classifier if desired.

We would then use the moderation endpoint results to track some sort of confidence interval for the non-abuse of our request traffic. (e.g. by sampling X of the past Y requests, over various timeframes, we have Z% confidence that...) And, if it does identify an abuse LLM request, we would then use that to trigger our existing anti-abuse mechanisms.

## Test plan

CI. The code (should 😅) be functionally equivalent. Existing testing (should 😅) be sufficient.

### DIFF
--- a/cmd/cody-gateway/internal/httpapi/handler.go
+++ b/cmd/cody-gateway/internal/httpapi/handler.go
@@ -76,7 +76,9 @@ func NewHandler(
 	// V1 service routes
 	v1router := r.PathPrefix("/v1").Subrouter()
 
-	if config.Anthropic.AccessToken != "" {
+	if config.Anthropic.AccessToken == "" {
+		logger.Error("Anthropic access token not set. Not registering Anthropic-related endpoints.")
+	} else {
 		anthropicHandler, err := completions.NewAnthropicHandler(
 			logger,
 			eventLogger,
@@ -138,10 +140,11 @@ func NewHandler(
 					otelhttp.WithPublicEndpoint(),
 				),
 			))
-	} else {
-		logger.Error("Anthropic access token not set")
 	}
-	if config.OpenAI.AccessToken != "" {
+
+	if config.OpenAI.AccessToken == "" {
+		logger.Error("OpenAI access token not set. Not registering OpenAI-related endpoints.")
+	} else {
 		v1router.Path("/completions/openai").Methods(http.MethodPost).Handler(
 			overhead.HTTPMiddleware(latencyHistogram,
 				instrumentation.HTTPMiddleware("v1.completions.openai",
@@ -209,10 +212,11 @@ func NewHandler(
 					otelhttp.WithPublicEndpoint(),
 				),
 			))
-	} else {
-		logger.Error("OpenAI access token not set")
 	}
-	if config.Fireworks.AccessToken != "" {
+
+	if config.Fireworks.AccessToken == "" {
+		logger.Error("Fireworks access token not set. Not registering Fireworks-related endpoints.")
+	} else {
 		v1router.Path("/completions/fireworks").Methods(http.MethodPost).Handler(
 			overhead.HTTPMiddleware(latencyHistogram,
 				instrumentation.HTTPMiddleware("v1.completions.fireworks",
@@ -237,8 +241,6 @@ func NewHandler(
 					otelhttp.WithPublicEndpoint(),
 				),
 			))
-	} else {
-		logger.Error("Fireworks access token not set")
 	}
 
 	// Register a route where actors can retrieve their current rate limit state.
--- a/cmd/cody-gateway/internal/httpapi/handler.go
+++ b/cmd/cody-gateway/internal/httpapi/handler.go
@@ -99,6 +99,22 @@ func NewHandler(
 		v1router.Path(route).Methods(http.MethodPost).Handler(overheadMiddleware)
 	}
 
+	// registerSimpleGETEndpoint registers a basic HTTP GET endpoint, without the
+	// latency and performance counter middle ware that we register for other endpoints.
+	registerSimpleGETEndpoint := func(name, route string, handler http.Handler) {
+		v1router.Path(route).Methods(http.MethodGet).Handler(
+			instrumentation.HTTPMiddleware(name,
+				authr.Middleware(
+					requestlogger.Middleware(
+						logger,
+						handler,
+					),
+				),
+				otelhttp.WithPublicEndpoint(),
+			),
+		)
+	}
+
 	if config.Anthropic.AccessToken == "" {
 		logger.Error("Anthropic access token not set. Not registering Anthropic-related endpoints.")
 	} else {
@@ -156,17 +172,7 @@ func NewHandler(
 			attributesOpenAICompletions,
 			openAIHandler)
 
-		v1router.Path("/embeddings/models").Methods(http.MethodGet).Handler(
-			instrumentation.HTTPMiddleware("v1.embeddings.models",
-				authr.Middleware(
-					requestlogger.Middleware(
-						logger,
-						embeddings.NewListHandler(),
-					),
-				),
-				otelhttp.WithPublicEndpoint(),
-			),
-		)
+		registerSimpleGETEndpoint("v1.embeddings.models", "/embeddings/models", embeddings.NewListHandler())
 
 		embeddingsHandler := embeddings.NewHandler(
 			logger,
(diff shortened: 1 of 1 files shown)


cmd/cody-gateway/internal/httpapi/handler.go
```diff
@@ -75,6 +75,29 @@ func NewHandler(
...
+		// Wrap that in our instrumentation middleware, adding more logging.
+		instrumentedHandler := instrumentation.HTTPMiddleware(
+			name,
+			gaugedHandler,
+			otelhttp.WithPublicEndpoint())
+		// Finally wrap that again in our overall middleware.
+		overheadMiddleware := overhead.HTTPMiddleware(latencyHistogram, instrumentedHandler)
+
+		v1router.Path(route).Methods(http.MethodPost).Handler(overheadMiddleware)
+	}
```
vdavid, line 100: Just curious, why an inline function rather than a package-level private function?
I understand that this is only locally used, but I saw less Go code structured this way, so making this a package-level private function seems more idiomatic to me.
But then, you've seen a lot more Go code than me overall, so this might be just my observation. I don't see any clear pros for making it a package-level private function.

cmd/cody-gateway/internal/httpapi/handler.go
```diff
@@ -208,17 +214,9 @@ func NewHandler(
...
-				requestlogger.Middleware(
-					logger,
-					featurelimiter.ListLimitsHandler(logger, rs),
-				),
-			),
-			otelhttp.WithPublicEndpoint(),
-		),
-	)
+	limitsHandler := featurelimiter.ListLimitsHandler(logger, rs)
+	registerSimpleGETEndpoint("v1.limits", "/limits", limitsHandler)
```
vdavid, line 218: Again just curious: why the variable? In the change above (line 175), you used no variable. Coming from writing more TypeScript than Go, I'd inline this single-use variable. Just curious about the patterns in your mind for making this decision. Is there a specific line-length cutoff, or more of a gut feeling?

vdavid (approved):
Reviewed commit by commit, LGTM, thanks for the cleanup!

cmd/cody-gateway/internal/httpapi/handler.go
chrsmith: In general, package-level private functions are preferable. (Since that way you have something that you can write unit tests for, reuse, etc.)

However, if the function is super-duper specific, or doesn't really make sense anywhere else, having an inline function like this isn't out of the question.

But the specific reason why this is an inline function instead of a package-level private is because it captures a few values in the closure. e.g. `logger`, `counter`, and `latencyHistogram` are not parameters to this function. (Instead they are magically referencing the local variables in the parent function.)

So if this were a package-level private function it would have a much more complicated looking signature... to the point that it would probably be a net negative to use. (e.g. if a function requires 6+ parameters, maybe that's a good sign something is off.)

cmd/cody-gateway/internal/httpapi/handler.go
chrsmith: This is entirely a matter of style and personal preference. I split it up because having it all on one line would be a little "too long", and to (arguably) improve readability it was split up.

```go
// Method 1, inline.
registerSimpleGETEndpoint("v1.limits", "/limits", featurelimiter.ListLimitsHandler(logger, rs)

// Method 2, dedicated local variable.
limitsHandler := featurelimiter.ListLimitsHandler(logger, rs)
registerSimpleGETEndpoint("v1.limits", "/limits", limitsHandler)
```

Now, this argument is apparently bullshit. Because comparing the two functions side-by-side, using an inline approach here would be even shorter! 🤡😭 

```go
registerSimpleGETEndpoint("v1.embeddings.models", "/embeddings/models", embeddings.NewListHandler())
registerSimpleGETEndpoint("v1.limits", "/limits", featurelimiter.ListLimitsHandler(logger, rs)
```

vdavid: Okay, this was super nit, just wanted to learn some best practices here.
What I heard was "these two are equally fine in Go, don't think too much about it", which is a helpful takeaway.

bobheadxi: My unrequested preference 😉 

```go
registerSimpleGETEndpoint("v1.limits", "/limits",
    featurelimiter.ListLimitsHandler(logger, rs)
```

chrsmith: Well @bobheadxi , I "accept your truth" but let me try to parrot an argument I read when this type of thing came up in the Google Style Guide.

The "problem" with that code snippet, is that by having _some_ parameters continued on the first line, and _some other_ parameters on the second line, you are losing any meaning conveyed by the indentation after the new line.

Moreover, for even more complicated cases, it can be even more confusing. Consider:

```go
foo("arg1", bar(
    "arg2", "arg3"),
    "arg4")
```

So the "rule" would be something like: add the newline at the first { thingey } out of principle. (I don't recall exactly what things was.) But the idea was that way you would _always_ write code like the following. (i.e. if you break up the function arguments to another line, then you do so at the start of the function call.)

```go
foo(
    "arg1",
    bar("arg2", "arg3"),
    "arg4"))
```

Or some other variation...

However, @vdavid , just so you have the "OFFICIAL" and "CRYSTAL CLEAR BEST PRACTICES" when it comes to Golang, the only rule you need to know is: just do what `gofmt` demands. And keep any other linters happy.

But unlike, say, `eslint`, Golang linters _intentionally_ do _not_ provide any configuration knobs. Specifically to avoid random minutia like this in code reviews :)

https://www.youtube.com/watch?v=PAAkCSZUG1c&t=523s

bobheadxi: > But unlike, say, eslint, Golang linters intentionally do not provide any configuration knobs. Specifically to avoid random minutia like this in code reviews :)

For sure, I think any way is fine, but you know how things go... who can turn down some random minutia 😉 

Continuing the minutia for fun, in my mind, and the way I've seen it on occasion, it doesn't hurt readability too much to have newlines _per argument_ after the first few, i.e. I would say

```go
foo("arg1", bar(
    "arg2", "arg3"),
    "arg4")
```

is bad, but

```go
foo("arg1",
  bar("arg2", "arg3"),
  "arg4")
```

is _fine_. The idea being in Go, there's a "scope precedence" for the args, e.g. `context comes first`. These args are just context and higher-level things that can be grouped together on the first line of the call. But, maybe my brain is weird 😁 

I will say, I do find things like:

```go
limitsHandler := featurelimiter.ListLimitsHandler(logger, rs)
registerSimpleGETEndpoint("v1.limits", "/limits", limitsHandler)
```

a bit annoying to read sometimes if the var doesn't get reused, partly because sometimes I've seen copy-pasta go wrong here when there are things like `limitsHandler1`, `limitsHandler2` in the same scope and you use the wrong thing

cmd/cody-gateway/internal/httpapi/handler.go
```diff
@@ -75,8 +75,49 @@ func NewHandler(
...
+			otelhttp.WithPublicEndpoint())
+		// Finally wrap that again in our overall middleware.
+		overheadMiddleware := overhead.HTTPMiddleware(latencyHistogram, instrumentedHandler)
+
+		v1router.Path(route).Methods(http.MethodPost).Handler(overheadMiddleware)
+	}
+
+	// registerSimpleGETEndpoint registers a basic HTTP GET endpoint, without the
+	// latency and performance counter middle ware that we register for other endpoints.
+	registerSimpleGETEndpoint := func(name, route string, handler http.Handler) {
```
thenamankumar, line 104: I would rename `registerSimpleGETEndpoint` to `registerBaseGETEndpoint`.

thenamankumar (approved):

chwarwick (approved):

PR #61013, opened 2024-03-12
Svelte: implement preview panel

camdencheek:
This implements the basics of the preview panel.

Followup tasks:
- Implement the codemirror extension that lets you cycle through matches within a file
- Add preloading on hover
- Make a more general "resizeable split" component

## Test plan

Loom walkthrough with manual testing and added a playwright test (yay!)

https://www.loom.com/share/c5e88700be014a52a6b8e794032a8894

### DIFF
--- a/client/web-sveltekit/src/lib/utils/promises.ts
+++ b/client/web-sveltekit/src/lib/utils/promises.ts
@@ -81,3 +81,19 @@ export function createPromiseStore<D, E = Error>(): PromiseStore<D, E> {
         },
     }
 }
+
+/**
+ * Returns a store that publishes updates when the promise is resolved or rejected.
+ */
+export function toReadable<D, E = Error>(promise: Promise<D>): Readable<Result<D, E>> {
+    const resultStore = writable<Result<D, E>>({ value: null, error: null, pending: true })
+    promise.then(
+        result => {
+            resultStore.set({ value: result, error: null, pending: false })
+        },
+        error => {
+            resultStore.set({ value: null, error: error, pending: false })
+        }
+    )
+    return resultStore
+}
--- a/client/web-sveltekit/src/lib/utils/promises.ts
+++ b/client/web-sveltekit/src/lib/utils/promises.ts
@@ -85,7 +85,7 @@ export function createPromiseStore<D, E = Error>(): PromiseStore<D, E> {
 /**
  * Returns a store that publishes updates when the promise is resolved or rejected.
  */
-export function toReadable<D, E = Error>(promise: Promise<D>): Readable<Result<D, E>> {
+export function toReadable<D, E = Error>(promise: PromiseLike<D>): Readable<Result<D, E>> {
     const resultStore = writable<Result<D, E>>({ value: null, error: null, pending: true })
     promise.then(
         result => {
--- a/client/web-sveltekit/src/routes/search/PreviewButton.svelte
+++ b/client/web-sveltekit/src/routes/search/PreviewButton.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
     import type { ContentMatch, PathMatch, SymbolMatch } from '$lib/shared'
+    import { Button } from '$lib/wildcard'
 
     import { getSearchResultsContext } from './searchResultsContext'
 
@@ -8,16 +9,13 @@
     const searchResultContext = getSearchResultsContext()
 </script>
 
-<button on:click={() => searchResultContext.setPreview(result)}>Preview</button>
+<div class="preview-button">
+    <Button variant="link" on:click={() => searchResultContext.setPreview(result)}>Preview</Button>
+</div>
 
 <style lang="scss">
-    button {
-        border: none;
-        background-color: transparent;
-        font-weight: 500;
-        color: var(--link-color);
-        &:hover {
-            text-decoration: underline;
-        }
+    // Override the padding of the button so it doesn't grow the header.
+    .preview-button :global(button) {
+        padding: 0;
     }
 </style>
--- a/client/web-sveltekit/src/routes/search/PreviewButton.svelte
+++ b/client/web-sveltekit/src/routes/search/PreviewButton.svelte
@@ -9,3 +9,15 @@
 </script>
 
 <button on:click={() => searchResultContext.setPreview(result)}>Preview</button>
+
+<style lang="scss">
+    button {
+        border: none;
+        background-color: transparent;
+        font-weight: 500;
+        color: var(--link-color);
+        &:hover {
+            text-decoration: underline;
+        }
+    }
+</style>
--- a/client/web-sveltekit/src/routes/search/PreviewButton.svelte
+++ b/client/web-sveltekit/src/routes/search/PreviewButton.svelte
@@ -1,0 +1,11 @@
+<script lang="ts">
+    import type { ContentMatch, PathMatch, SymbolMatch } from '$lib/shared'
+
+    import { getSearchResultsContext } from './searchResultsContext'
+
+    export let result: ContentMatch | SymbolMatch | PathMatch
+
+    const searchResultContext = getSearchResultsContext()
+</script>
+
+<button on:click={() => searchResultContext.setPreview(result)}>Preview</button>
--- a/client/web-sveltekit/src/routes/search/PreviewPanel.svelte
+++ b/client/web-sveltekit/src/routes/search/PreviewPanel.svelte
@@ -71,7 +71,7 @@
 <div class="header">
     <h3>File Preview</h3>
     <button on:click={() => searchResultContext.setPreview(null)}>
-        <Icon svgPath={mdiClose} class="close-icon" />
+        <Icon svgPath={mdiClose} class="close-icon" size={16} inline />
     </button>
 </div>
 <div class="file-link">
@@ -127,6 +127,14 @@
             margin: 0;
             margin-right: auto;
             height: fit-content;
+        }
+        button {
+            background-color: transparent;
+            color: var(--text-muted);
+            border: none;
+            &:hover {
+                color: var(--body-color);
+            }
         }
     }
 
--- a/client/web-sveltekit/src/routes/search/PreviewPanel.svelte
+++ b/client/web-sveltekit/src/routes/search/PreviewPanel.svelte
@@ -23,7 +23,10 @@
     import { settings } from '$lib/stores'
 
     // TODO: should I be importing this from a page? Seems funky.
-    import { BlobPageQuery } from '../[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/page.gql'
+    import {
+        BlobPageQuery,
+        BlobSyntaxHighlightQuery,
+    } from '../[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/page.gql'
 
     export let repoName: string
     export let commitID: string
@@ -38,6 +41,8 @@
         },
     })
 
+    console.log({ filePath })
+
     const blob = client
         .query(BlobPageQuery, {
             repoName,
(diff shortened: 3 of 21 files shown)


client/web-sveltekit/src/lib/utils/promises.ts
```diff
@@ -81,3 +81,19 @@ export function createPromiseStore<D, E = Error>(): PromiseStore<D, E> {
         },
     }
 }
+
+/**
+ * Returns a store that publishes updates when the promise is resolved or rejected.
+ */
```
camdencheek, line 87: @fkling I added this as a lighter weight helper for the times I just want to wrap a single promise in a `Readable`. It uses same nice `Result` interface as above. Would appreciate both a "correctness" review here and a "is this actually useful" review

client/web-sveltekit/src/routes/search/PreviewPanel.svelte
```diff
@@ -0,0 +1,154 @@
...
+                }}
+                highlights={$highlightStore.value ?? ''}
+                {codeIntelAPI}
+            />
+        {/if}
+    </div>
+</div>
+
+<style lang="scss">
+    .preview-container {
```
camdencheek, line 114: Used the more verbose `.preview-container` rather than `.container` because we have global styles that apply to `.container`, which was causing a bunch of surprising behavior.

client/web-sveltekit/src/routes/search/PreviewPanel.svelte
```diff
@@ -0,0 +1,154 @@
...
+        start: Location
+        end: Location
+    }
+
+    export interface Location {
+        // A zero-based line number
+        line: number
+        // A zero-based column number
+        column: number
+    }
```
fkling, line 12: (note to self) This is one of these cases where we should have better type checking / validation somehow, to ensure that one cannot pass a one-based location in the future

client/web-sveltekit/src/routes/search/PreviewPanel.svelte
```diff
@@ -0,0 +1,154 @@
...
+
+    const searchResultContext = getSearchResultsContext()
+
+    const client = getGraphQLClient()
+    $: codeIntelAPI = createCodeIntelAPI({
+        settings: setting => (isErrorLike($settings?.final) ? undefined : $settings?.final?.[setting]),
+        requestGraphQL(options) {
+            return from(client.query(options.request, options.variables).then(toGraphQLResult))
+        },
+    })
```
fkling, line 53: (note to self): Maybe it makes sense to make this a reusable function, or even make this something that is made available via context.

client/web-sveltekit/src/routes/search/PreviewPanel.svelte
```diff
@@ -0,0 +1,154 @@
...
+
+    const blobStore = toReadable(
```
fkling, line 55: This needs to be reactive
```suggestion
    $: blobStore = toReadable(
```

client/web-sveltekit/src/routes/search/PreviewPanel.svelte
```diff
@@ -0,0 +1,154 @@
...
+        client
+            .query(BlobPageQuery, {
+                repoName: result.repository,
+                revspec: result.commit ?? '', // TODO: default value?
+                path: result.path,
+            })
+            .then(mapOrThrow(result => result.data?.repository?.commit?.blob ?? null))
+    )
+
+    const highlightStore = toReadable(
```
fkling, line 65: This needs to be reactive

```suggestion
    $: highlightStore = toReadable(
```

client/web-sveltekit/src/routes/search/PreviewPanel.svelte
```diff
@@ -0,0 +1,154 @@
...
+    } from '$lib/shared'
+    import { settings } from '$lib/stores'
+    import { toReadable } from '$lib/utils/promises'
+    import { Alert } from '$lib/wildcard'
+
+    // TODO: should I be importing this from a page? Seems funky.
+    import {
+        BlobPageQuery,
+        BlobSyntaxHighlightQuery,
+    } from '../[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/page.gql'
```
fkling, line 39: Maybe leave a comment that explains why the query is shared (caching the query response) and that we'll be looking for a better solution.

client/web-sveltekit/src/routes/search/PreviewPanel.svelte
```diff
@@ -0,0 +1,154 @@
...
+            <CodeMirrorBlob
+                blobInfo={{
+                    repoName: result.repository,
+                    commitID: result.commit ?? '',
+                    revision: '',
+                    filePath: result.path,
+                    content: $blobStore.value.content ?? '',
+                    languages: $blobStore.value.languages ?? [],
+                }}
+                highlights={$highlightStore.value ?? ''}
```
fkling, line 106: I _think_ you want to also check `$highlightsStore.pending` here. Otherwise wouldn't it be possible that you are using the data from a previous `BlobSyntaxHighlightQuery` request?

client/web-sveltekit/src/routes/search/PreviewPanel.svelte
fkling: Yeah, we need to clean up / remove the global styles.

client/web-sveltekit/src/routes/search/PreviewPanel.svelte
```diff
@@ -0,0 +1,154 @@
...
+
+<style lang="scss">
+    .preview-container {
+        display: flex;
+        flex-direction: column;
+        height: 100%;
+    }
+
+    .header {
+        // Keep in sync with the action bar
```
fkling, line 121: Depending on _what_ exactly needs to be kept in sync, there are a couple of alternatives we can consider (nothing blocking):

- Create a reusable 'bar-like' component which applies the header style
- Create a CSS module with those styles and import it where needed.
- Create/use a SCSS mixin (never really tried that).

client/web-sveltekit/src/routes/search/SearchResults.svelte
```diff
@@ -16,23 +17,32 @@
     import { mdiCloseOctagonOutline } from '@mdi/js'
     import type { Observable } from 'rxjs'
     import { tick } from 'svelte'
+    import { writable } from 'svelte/store'
+
+    import { limitHit } from '@sourcegraph/branded'
```
fkling, line 22: I think you can import that from `$lib/branded` (and if not, it should be added there)
```suggestion
    import { limitHit } from '$lib/branded'
```

client/web-sveltekit/src/routes/search/SearchResults.svelte
```diff
@@ -72,6 +82,8 @@
     $: resultsToShow = results.slice(0, count)
     $: expandedSet = cacheEntry?.expanded || new Set<SearchMatch>()
 
+    $: previewResult = writable<ContentMatch | SymbolMatch | PathMatch | null>(cacheEntry?.preview ?? null)
```
fkling, line 85: Do you need the type parameter? Shouldn't it infer it from `cacheEntry.preview`?
```suggestion
    $: previewResult = writable(cacheEntry?.preview ?? null)
```

client/web-sveltekit/src/routes/search/SearchResults.svelte
```diff
@@ -118,10 +133,10 @@
 </div>
 
 <div class="search-results">
-    <div style:width={sidebarWidth}>
+    <div style:width={`clamp(14rem, ${$filtersSidebarPosition * 100}%, 35%)`}>
```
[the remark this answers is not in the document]
fkling: I think you can even change this to
```suggestion
    <div style:width="clamp(14rem, {$filtersSidebarPosition * 100}%, 35%)">
```

Svelte allows you to interpolate attribute values this way.

client/web-sveltekit/src/routes/search/SearchResults.svelte
```diff
@@ -152,6 +167,14 @@
             {/if}
         </div>
     </div>
+    {#if $previewResult}
+        <Separator currentPosition={previewSidebarPosition} />
+        <div style:width={`clamp(10rem, ${100 - $previewSidebarPosition * 100}%, 50%)`}>
```
fkling, line 172:
```suggestion
        <div style:width="clamp(10rem, {100 - $previewSidebarPosition * 100}%, 50%)">
```
(see above)

client/web-sveltekit/src/routes/search/SearchResults.svelte
```diff
@@ -152,6 +167,14 @@
...
+            {#key $previewResult}
```
fkling, line 173: Ah, this is why the preview panel worked, even without reactivity... in general not a problem, but if you look at `PreviewPanel` in isolation it appears "broken".

client/web-sveltekit/src/testing/integration.ts
```diff
@@ -177,19 +168,29 @@ class Sourcegraph {
...
-        return async () => {
-            // Wait for the search results page to be "ready"
-            await this.page.getByRole('heading', { name: 'Filter results' }).waitFor()
-            return this.page.evaluate(
-                ([events]) => {
-                    for (const event of events) {
-                        for (const source of window.$$sources) {
-                            source.dispatchEvent(new MessageEvent(event.event, { data: JSON.stringify(event.data) }))
+        return {
+            publish: async (events: SearchEvent | SearchEvent[]): Promise<void> => {
```
fkling, line 172: I'd make this a variadic function

```suggestion
            publish: async (...events: SearchEvent[]): Promise<void> => {
```

then it can be used in the following ways:

```
publish(event) // single event
publish(event1, event2) // multiple events
publish(...events) // array of events
```

client/web-sveltekit/src/testing/integration.ts
```diff
@@ -112,23 +119,7 @@ class Sourcegraph {
...
-                    skipped: [],
-                    durationMs: 100,
-                },
-            },
-            {
-                event: 'done',
-                data: {},
-            },
-        ]
+    public mockSearchStream(): MockSearchStream {
```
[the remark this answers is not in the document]
fkling: I love it! I didn't think about making an API that can be called multiple times to publish more events. Brilliant :)

fkling (approved):
Very nice!

client/web-sveltekit/src/routes/search/SearchResults.svelte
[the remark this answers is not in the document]
vovakulikov: I see that with current implementation we have more flexibility about how we control width of different blocks/panels on the page. But ideally I think I would like to have some like

```tsx
<Panels>
   <Panel name='Filters' minWidth='14rem' maxWidth='35%' side='right' >
     .....
   </Panel>  
   
   /* Any non panel content */

   <Panel name='File preview' minWidth='14rem' maxWidth='35%' side='left' >
     .....
   </Panel>  
</Panels> 
```  

Maybe a bit verbose with Panels component but no math on consumer and in Panel component we could use reference from Panels component instead of node.parent (this would solve problem if we have more complex layout with additional wrappers)

client/web-sveltekit/src/routes/search/PreviewPanel.svelte
vovakulikov: There was a problem in react version that I couldn't adjust heights in action bars (header in the search results block) and header in the file preview panel with just grid layout of the search result page (because resizable panel create separate block and sub-grids are still not supported in many browsers. So I ended up on just syncing them manually. 

But here I think we could setup search result page grid properly since resizable panels don't create any additional block but implemented with flat div elements, so grid still can be applied to different part of the file preview (including header)

fkling: I always forget about grid... definitely worth investigating.

client/web-sveltekit/src/testing/integration.ts
camdencheek: Oh, much more elegant!

client/web-sveltekit/src/routes/search/PreviewPanel.svelte
vovakulikov: But again as I mentioned above I think panels should be seperate components (in my opinion) and this would break the grid here. So 

- Create a reusable 'bar-like' component which applies the header style
This is probably the best option because CSS module seems too lose and mixin might be hard to read and unwrap when it's too specific.

client/web-sveltekit/src/routes/search/PreviewPanel.svelte
camdencheek: > Yeah, we need to clean up / remove the global styles.

This sounds like something that will only get harder as we build out more UI that might silently depend on the current global styles. Should we prioritize this?

client/web-sveltekit/src/testing/integration.ts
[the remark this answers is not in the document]
camdencheek: Thank you!

client/web-sveltekit/src/routes/search/SearchResults.svelte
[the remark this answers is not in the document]
camdencheek: @fkling that didn't seem to work for me. I think I need the backticks for string interpolation

client/web-sveltekit/src/routes/search/PreviewButton.svelte
```diff
@@ -0,0 +1,23 @@
...
+    import type { ContentMatch, PathMatch, SymbolMatch } from '$lib/shared'
+
+    import { getSearchResultsContext } from './searchResultsContext'
+
+    export let result: ContentMatch | SymbolMatch | PathMatch
+
+    const searchResultContext = getSearchResultsContext()
+</script>
+
+<button on:click={() => searchResultContext.setPreview(result)}>Preview</button>
```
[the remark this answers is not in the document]
camdencheek: Ah, I did, but ran into issues with the button padding. I just pushed a change to use the `variant="link"` styling as the baseline and just overrode the padding.

client/web-sveltekit/src/routes/search/PreviewPanel.svelte
camdencheek: Ha, yes. We run into this all the time on the backend. Also, the definition of "column" is weird and loose given the translation between bytes, UTF8 code points, and UTF16 code points.

client/web-sveltekit/src/routes/search/PreviewPanel.svelte
camdencheek: The update of `$highlightsStore` and `$blobStore` will happen in the same transaction (micro update?), correct? In that case, if we change the file being pointed to, both promises will be updated at the same time and this _should_ just be an unresolved promise as soon as we'd change files (I think)

client/web-sveltekit/src/routes/search/SearchResults.svelte
camdencheek: Whoops, got caught by auto-import. Thanks!

client/web-sveltekit/src/routes/search/SearchResults.svelte
camdencheek: Sure enough. Fixed!

client/web-sveltekit/src/routes/search/SearchResults.svelte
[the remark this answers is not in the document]
fkling: Too bad. Maybe it only works for real attributes, like `id="..."` and not those shorthands.

client/web-sveltekit/src/routes/search/PreviewPanel.svelte
fkling: Lets do it rather sooner than later but I don't expect it to be a big issue with the current scope.

client/web-sveltekit/src/routes/search/PreviewButton.svelte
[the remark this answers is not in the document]
fkling: Have you tried size small or whatever it's called?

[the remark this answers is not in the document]
camdencheek: Yes, I have. There is still padding, and it also makes the text size small

PR #61024, opened 2024-03-12
Check for scheduled or processing jobs when scheduling permission syncs

### LINKED ISSUE #61023: Permission sync jobs are being scheduled even when a sync has already been scheduled or is in progress.

pjlast:
Closes #61023 

Fixes an issue where multiple permission sync jobs are scheduled for users and repositories at the same time.

The existing SQL queries don't take into account that a permissions sync could be in progress or queued when checking for users/repos with the oldest permissions, which leads to a lot of overlapping permission syncs being scheduled. This could cause significant delays in the permissions sync system as a whole, since a single permissions sync now turns into 5+ consecutive syncs for the same user.

## Test plan

Verified on my local machine that multiple permission jobs are no longer scheduled for the same user/repo.


eseliger (approved):
Change LGTM, but can we get a test to prove this behavior?

PR #61033, opened 2024-03-12
svelte: new results summary indicator

jasonhawkharris:
This PR adds the new [search results indicator](https://www.figma.com/file/EPgl0qVbqGSumvf52brhda/Code-Search?type=design&node-id=1-52&mode=design&t=Bg1BGlk8kkdIi1JK-0) to the Svelte web app. This is an easier way for customers to see how their current search is performing, and to inform them of any errors, warnings or suggestions. 

For the most part, it fetches data from the backend. However, two of the use cases (timeouts and fetching) use elapsed time that is kept track of by the client. We then use this value to determine which actions we should show.

**States**
*Completed search state*
Screenshot 2024-03-22 at 2 21 08 PM

*Skipped items state*
Screenshot 2024-03-22 at 3 03 35 PM

*Fetching/Loading state*
Screenshot 2024-03-22 at 2 21 49 PM

*Taking too long state*
Screenshot 2024-03-22 at 2 46 45 PM

*Error state*
Screenshot 2024-03-22 at 2 39 36 PM

I am still learning about conventions and good practices in Svelte. I'd love for a more seasoned Svelte engineer to review this PR with this in mind. (Hello @fkling) For example, am I breaking things up too much? 

## Test plan
- Manually checked all use cases with various inputs
- Ships with a suite of unit tests


client/web-sveltekit/src/routes/search/StreamingProgress.svelte
```diff
@@ -17,6 +18,8 @@
         error: mdiAlertCircle,
     }
     let searchAgainDisabled = true
+    let elapsedSeconds: number
+    let elapsedMilliseconds: number
```
fkling, line 22: It doesn't look like these need to be defined here? you are not doing anything with them.

client/web-sveltekit/src/lib/search/ResultsIndicator.svelte
```diff
@@ -0,0 +1,220 @@
...
+    export let hasSuggestedItems: boolean
+    export let searchProgress: Progress
+    export let togglePopover: () => void
+    export let trigger: (node: HTMLElement) => void
+
+    // Create variables to be passed into the SecondCounter component
+    // These variables will be used to determine the amount of time
+    // the search has run without a response
+    export let elapsedSeconds: number = 0
+    export let elapsedMilliseconds: number = 0
```
fkling, line 20: It doesn't seem seem necessary to export those... am I missing something?

client/web-sveltekit/src/lib/search/ResultsIndicator.svelte
```diff
@@ -0,0 +1,220 @@
...
+            elapsedSeconds = Math.floor(delta / 1000)
+            elapsedMilliseconds = delta % 100
+            if (now - lastUpdate >= 5) {
+                displaySeconds = elapsedSeconds
+                displayMilliseconds = elapsedMilliseconds
+                lastUpdate = now
+            }
+        }, 1300)
+    })
+
```
fkling, line 39: I think it could be useful to factor this out into a store, so that you only have to do something like this:

```js
const elapsedTime = createElapsedTimeStore({start: Date.now(), interval: 1300})
$: console.log($elapsedTime)
```

Basically hide the implementation detail of how to compute the elapsed time.

Regardless though, the issue with the way you are using `setInterval` here is that you never remove the interval. At the very least you'd have to return a function from `onMount` to clean up the interval, e.g.

```js
const id = setInterval(...)
return () => window.clearInterval(id)
```

client/web-sveltekit/src/lib/search/ResultsIndicator.svelte
```diff
@@ -0,0 +1,220 @@
...
+        margin-right: 1rem;
+    }
+
+    div.duration-badge {
+        display: flex;
+        flex-flow: row-nowrap;
+        margin-top: 0.5rem;
+    }
+
+    div.loading-action-message {
```
fkling, line 214: No need to prefix all of these with `div`.

client/web-sveltekit/src/routes/search/StreamingProgress.svelte
```diff
@@ -36,15 +36,16 @@
...
-                <Icon svgPath={mdiChevronDown} inline />
-            </button>
-        </svelte:fragment>
-    </Button>
+    <ResultsIndicator
+        {hasSkippedItems}
+        {sortedItems}
+        {hasSuggestedItems}
+        trigger={registerTrigger}
+        togglePopover={toggle}
```
fkling, line 44: It's unfortunate that actions cannot be used with components the same way as the can be with elements.

However, seeing how the button in `<ResultsIndicator />` just calls `togglePopover`, why not keep the button element outside of `ResultsIndiciator` and have _that_ component really just be the _content_ of the button.
That avoid having to add these two properties to integrate with the popover.

I.e.

```svelte
<Popover ...>
  <Button variant="secondary" size="sm" outline>
    <svelte:fragment slot="custom" let:buttonClass>
      <button use:registerTrigger class="{buttonClass} progress-button" on:click={() => toggle()}>
        <ResultsIndicator .../>
      </button>
    </svelte:fragment>
</Popover>
```

client/web-sveltekit/src/lib/search/ResultsIndicator.svelte
```diff
@@ -0,0 +1,220 @@
...
+                    <LoadingSpinner inline />
+                    <div class="messages">
+                        <div class="progress-info-message">
+                            Fetching results... {elapsedSeconds}.{elapsedMilliseconds}s
+                        </div>
+                        {#if elapsedSeconds < MAX_SEARCH_DURATION}
+                            <div class="loading-action-message">Running search...</div>
+                        {:else}
+                            <div class="duration-badge">
+                                <div class={`info-badge duration`}>Taking too long?</div>
```
fkling, line 69:
```suggestion
                                <div class="info-badge duration">Taking too long?</div>
```

client/web-sveltekit/src/lib/search/ResultsIndicator.svelte
```diff
@@ -0,0 +1,220 @@
...
+
+<Button variant="secondary" size="sm" outline>
+    <svelte:fragment slot="custom" let:buttonClass>
+        <button use:trigger class="{buttonClass} progress-button" on:click={() => togglePopover()}>
+            {#if loading && !searchProgress}
+                <div class="loading">
+                    <LoadingSpinner inline />
+                    <div class="messages">
+                        <div class="progress-info-message">
+                            Fetching results... {elapsedSeconds}.{elapsedMilliseconds}s
```
fkling, line 63: Maybe date-fn already has a method for formatting time a specific way, then you don't have to do this manually. And if not I'd create  helper function for that.

client/web-sveltekit/src/lib/search/ResultsIndicator.svelte
```diff
@@ -0,0 +1,220 @@
...
+    export let togglePopover: () => void
+    export let trigger: (node: HTMLElement) => void
+
+    // Create variables to be passed into the SecondCounter component
+    // These variables will be used to determine the amount of time
+    // the search has run without a response
+    export let elapsedSeconds: number = 0
+    export let elapsedMilliseconds: number = 0
+    let displaySeconds: number = 0
+    let displayMilliseconds: number = 0
```
fkling, line 22: These don't seem to be used.

client/web-sveltekit/src/lib/search/ResultsIndicator.svelte
jasonhawkharris: What do you mean when you say a "store"

fkling: This: https://learn.svelte.dev/tutorial/readable-stores

client/web-sveltekit/src/lib/search/resultsIndicator/InfoBadge.test.ts
fkling, line 1: Did you find this easier to write than playwright tests?

client/web-sveltekit/src/lib/search/resultsIndicator/ResultsIndicator.svelte
```diff
@@ -0,0 +1,101 @@
...
+        let startTime = Date.now()
+        setInterval(() => {
+            const now = Date.now()
+            elapsedDuration = now - startTime
+            // once search has completed, reset the startTime
+            if (done) {
+                startTime = Date.now()
+            }
+        }, 1300)
+    })
```
fkling, line 35:
```suggestion
    onMount(() => {
        let startTime = Date.now()
        const interval = setInterval(() => {
            const now = Date.now()
            elapsedDuration = now - startTime
            // once search has completed, reset the startTime
            if (done) {
                startTime = Date.now()
            }
        }, 1300)
        return () => clearInterval(interval)
    })
```
 

You need to clear the interval when the component unmounts (by returning a cleanup callback). Also you probably don't want to keep the interval running when the search is done. However, since you start the interval on mount there wouldn't be a way to restart it. Which makes me think that you probably want to start the interval when `state === 'loading'`

You could create a store with a `start` and `stop` method for starting and stopping the interval. That could look like this:

```typescript
interface Timer extends Readable<number> {
   start(): void
   stop(): void
}

export function createTimer(): Timer {
    const {subscribe, set} = writable(0, () => {
      // Cleanup function is called when the last
      // subscriber unsubscribes.
      return () => clearInterval(timer)
    })
    let timer: number | undefined
    
    return {
      subscribe,
      start() {
        if (!timer) {
          const startTime = Date.now()
          timer = setInterval(() => {
              set(Date.now() - startTime)
          }, 1000)
        }
      },
      stop() {
        clearInterval(timer)
        timer = undefined
      }
    }
}
```

(this is a rough version that should work ok for this case)

which can then be used like so:


```typescript
const elapsedDuration = createTimer()

$: if (state === 'loading') {
    elapsedDuration.start()
} else {
    elapsedDuration.stop()
}
```

and you can get the current value by accessing `$elapsedDuration` (variable name prefixed with `$`). For example:


```typescript
$: console.log($elapsedDuration)
```

client/web-sveltekit/src/lib/search/resultsIndicator/InfoBadge.svelte
```diff
@@ -0,0 +1,24 @@
...
+</div>
+
+<style lang="scss">
+    .error-text {
+        color: var(--danger);
+    }
+
+    .progress-message {
+        font-size: 0.9rem;
+        margin-left: 0.2rem;
```
camdencheek, line 22: General composability/style nit: margins should be left to consumer of a component. Otherwise we can't drop this component in somewhere else. Not a huge deal since this is only used one place, but it's a good habit to get into.

Similar comment applies elsewhere in the PR

client/web-sveltekit/src/lib/search/resultsIndicator/SuggestedAction.svelte
```diff
@@ -0,0 +1,90 @@
...
+    export let hasSkippedItems: boolean
+    export let hasSuggestedItems: boolean
+    export let mostSevere: Skipped
+
+    const INTERPUNCT = '\u00B7'
+
+    $: done = progress.done
+</script>
+
+{#if progress}
```
fkling, line 16: What's the type of `Progress`? Surprised to see that it can be nullable.

client/web-sveltekit/src/lib/search/resultsIndicator/TimeoutMessage.svelte
```diff
@@ -0,0 +1,58 @@
...
+    const CENTER_DOT = '\u00B7' // interpunct
+</script>
+
+<div class={`info-badge ${isError && 'error-text'}`}>
+    <div class="duration-badge">
+        <div class="info-badge duration">Taking too long?</div>
+        <div class="separator">{CENTER_DOT}</div>
+        <div class="action-badge">
+            Use
+            <a href="https://sourcegraph.com/docs/code-search/types/search-jobs" target="_blank"> Search Job </a>
```
jasonhawkharris, line 13: @camdencheek Is this okay? Or should I find another way to link to the docs?

camdencheek: We should use the `/help` redirect for this, otherwise it will always point to the latest docs rather than the docs for the version that is actually running

client/web-sveltekit/src/lib/search/resultsIndicator/SuggestedAction.svelte
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    .info-badge.duration {
+        background: var(--warning);
+        color: black;
+    }
+
+    .info-badge.error-text {
+        background: var(--danger);
+    }
```
fkling: You can nest CSS selectors and use `&` to avoid repeating the common class name. E.g. 

```suggestion
    .info-badge {
        background-color: var(--primary);
        border-radius: 3px;
        color: var(--white);
        padding-left: 0.4rem;
        padding-right: 0.2rem;

        &.duration {
            background: var(--warning);
            color: black;
        }

        &.error-text {
            background: var(--danger);
        }
    }
```

client/web-sveltekit/src/routes/search/SearchResults.svelte
```diff
@@ -133,17 +133,12 @@
...
     <div style:width={`clamp(14rem, ${$filtersSidebarPosition * 100}%, 35%)`}>
-        <DynamicFiltersSidebar {selectedFilters} streamFilters={$stream.filters} searchQuery={queryFromURL} {loading} />
+        <DynamicFiltersSidebar {selectedFilters} streamFilters={$stream.filters} searchQuery={queryFromURL} {state} />
     </div>
     <Separator currentPosition={filtersSidebarPosition} />
     <div class="results">
         <aside class="actions">
-            {#if loading}
-                <div>
-                    <LoadingSpinner inline />
```
jasonhawkharris, line 143: Removed this because it caused to loading spinners to load. We don't need this here since we already have it on the actual component.

client/web-sveltekit/src/lib/search/resultsIndicator/InfoBadge.svelte
```diff
@@ -0,0 +1,24 @@
...
+    {getProgressText(searchProgress).visibleText}
+</div>
+
+<style lang="scss">
+    .error-text {
+        color: var(--danger);
+    }
+
+    .progress-message {
+        font-size: 0.9rem;
```
camdencheek, line 21: Can we use `<small>` for this instead? Tossing in a custom font size is a little code-smelly. Similar comment applies to the rest of the PR

client/web-sveltekit/src/lib/search/resultsIndicator/TimeoutMessage.svelte
```diff
@@ -0,0 +1,59 @@
...
```
fkling, line 13: This should probably link to the docs page of the instance. So maybe just
```suggestion
            <a href="/docs/code-search/types/search-jobs" target="_blank"> Search Job </a>
```

? I'm not sure.

client/web-sveltekit/src/lib/search/resultsIndicator/InfoBadge.svelte
```diff
@@ -0,0 +1,24 @@
...
+    import { getProgressText } from '$lib/branded'
+    import type { Progress } from '$lib/shared'
+
+    export let searchProgress: Progress
+    export let state: 'error' | 'complete' | 'loading'
+
+    $: isError = state === 'error'
+</script>
+
+<div class={`progress-message ${isError && 'error-text'}`}>
```
camdencheek, line 11: Take a look a [this](https://learn.svelte.dev/tutorial/classes)

```suggestion
<div class="progress-message" class:error-text={isError}>
```

client/web-sveltekit/src/lib/search/utils.ts
```diff
@@ -44,3 +44,7 @@ export function lineToMatchGroup(match: LineMatch): MatchGroup {
         endLine: match.lineNumber,
     }
 }
+
+export const capitalizeFirstLetter = (str: string): string => {
+    return str.charAt(0).toUpperCase() + str.slice(1)
+}
```
fkling, line 50: lodash has a `capitalize` function. Since we already have that as dependency you might as well just use that one instead of rolling your own.

client/web-sveltekit/src/routes/search/SearchResults.svelte
```diff
@@ -19,11 +19,11 @@
     import { tick } from 'svelte'
     import { writable } from 'svelte/store'
 
+    import { limitHit } from '@sourcegraph/branded'
```
fkling, line 22:
```suggestion
    import { limitHit } from '$lib/branded'
```

client/web-sveltekit/src/lib/search/resultsIndicator/ProgressMessage.svelte
```diff
@@ -0,0 +1,49 @@
+<script lang="ts">
+    import InfoBadge from '$lib/search/resultsIndicator/InfoBadge.svelte'
+    import type { Progress } from '$lib/shared'
+
+    export let state: 'error' | 'loading' | 'complete'
+    export let progress: Progress
+    export let loading: boolean
+    export let isError: boolean
```
camdencheek, line 8: There seems to be a lot of overlap in the information provided by these dependencies. Do we actually need all of these?

client/web-sveltekit/src/routes/search/StreamingProgress.svelte
```diff
@@ -25,23 +27,19 @@
...
     $: sortedItems = sortBySeverity(progress.skipped)
     $: openItems = sortedItems.map((_, index) => index === 0)
     $: suggestedItems = sortedItems.filter((skipped): skipped is Required<Skipped> => !!skipped.suggested)
     $: hasSuggestedItems = suggestedItems.length > 0
+    $: isError = state === 'error'
 </script>
 
 <Popover let:registerTrigger let:toggle placement="bottom-start">
-    <Button variant="secondary" size="sm" outline>
+    <Button variant={`${isError ? 'danger' : 'secondary'}`} size="sm" outline>
```
fkling, line 46: This should work too:

```suggestion
    <Button variant="{isError ? 'danger' : 'secondary'}" size="sm" outline>
```

client/web-sveltekit/src/lib/search/resultsIndicator/InfoBadge.svelte
camdencheek: Looks like there are a couple of other places where we can avoid string interpolation as well

client/web-sveltekit/src/lib/search/resultsIndicator/ProgressMessage.svelte
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    .suggested-action {
+        display: flex;
+        flex-flow: row nowrap;
+        margin-left: 0.2rem;
+    }
+
+    .running-search {
+        color: var(--gray-06);
```
camdencheek, line 47: Should this be `var(--text-muted)` instead? I think that's probably a safer bet that it'll look good in both light and dark mode

client/web-sveltekit/src/lib/search/resultsIndicator/ProgressMessage.svelte
```diff
@@ -0,0 +1,49 @@
...
+
+{#if loading}
+    <div class="progress-message">Fetching results... {(elapsedDuration / 1000).toFixed(1)}s</div>
+    <div class={`action-container ${isError && 'error-text'}`}>
+        <div class="suggested-action">
+            {#if elapsedDuration <= maxSearchDuration}
+                <div class="running-search">Running Search</div>
+            {/if}
+        </div>
+    </div>
```
camdencheek, line 21: There seems to be one (or or maybe two?) more layers of divs than necessary here. Should we collapse these?

client/web-sveltekit/src/lib/search/resultsIndicator/ResultsIndicator.svelte
```diff
@@ -0,0 +1,101 @@
...
+    import TimeoutMessage from '$lib/search/resultsIndicator/TimeoutMessage.svelte'
+    import type { Progress, Skipped } from '$lib/shared'
+
+    export let hasSkippedItems: boolean
+    export let sortedItems: Skipped[]
+    export let hasSuggestedItems: boolean
+    export let progress: Progress
+    export let state: 'error' | 'complete' | 'loading'
+
+    const MAX_SEARCH_DURATION = 8000
```
camdencheek, line 18: What does MAX_SEARCH_DURATION acutally describe? Is this the threshold above which we suggest search jobs? If so, could we maybe rename this "SEARCH_JOB_THRESHOLD" or something to avoid making it sound like this is a search limit?

client/web-sveltekit/src/lib/search/resultsIndicator/ResultsIndicator.test.ts
```diff
@@ -0,0 +1,44 @@
...
+                title: 'Display limit hit',
+                message: 'you hit the display limit',
+                severity: 'info',
+            }],
+            hasSuggestedItems: true,
+            progress: {
+                matchCount: 600,
+                durationMs: 5260,
+                skipped: [],
+            },
```
fkling: This is a good example of why I suggested earlier to derive the values for `hasSkippedItems` and others from the `progress` object directly instead of passing them via props. As it is your data is inconsistent. You are setting `hasSkippedItems: true` but `skipped` is an empty array.

Ideally the API of the component cannot be used in a way that makes it inconsistent.

Alternatively, if `ResultsIndicator` is not accessing `progress.skipped`, narrow down the type of the `progress` prop to only include the properties that the component actually needs.

client/web-sveltekit/src/lib/search/resultsIndicator/ResultsIndicator.svelte
```diff
@@ -0,0 +1,101 @@
...
+
+    onMount(() => {
+        let startTime = Date.now()
+        setInterval(() => {
+            const now = Date.now()
+            elapsedDuration = now - startTime
+            // once search has completed, reset the startTime
+            if (done) {
+                startTime = Date.now()
+            }
```
camdencheek, line 33: This doesn't look quite right. After the search is done, we don't want to keep updating this, right? So maybe we should keep track of the timer and unsubscribe once the search completes?

client/web-sveltekit/package.json
```diff
@@ -10,6 +10,7 @@
     "preview": "vite preview",
     "test": "playwright test",
     "test:dev": "PORT=5173 playwright test --ui",
+    "test:svelte": "vitest --run",
```
fkling, line 13: FWIW you can just execute `pnpm vitest` to run all unit tests, although I guess having it here gives you better autocompletion.
If we want to keep this we should (eventually) rename the commands to something more meaningful.

dev/linters/bodyclose/BUILD.bazel
fkling, line 1: I think these can all be reverted. No idea where these are coming from.

client/web-sveltekit/src/lib/search/resultsIndicator/ResultsIndicator.svelte
```diff
@@ -0,0 +1,101 @@
...
+     * TODO: @jasonhawkharris Explore combining 'complete' and 'done' values.
+     * The values do refer to different objects. 'complete' is the state returned
+     * from the stream. 'done' is the state returned from the progress object (which)
+     * also exists as a smaller object inside of the search stream. Do they need to?
+     */
+    $: done = progress.done || state === 'complete'
+</script>
+
+<div class="indicator">
+    <div class="icon">
```
fkling, line 55: Do we have a global `.icon` class that does something? If not I guess you can remove the class name here. If yes, we should (eventually) find a way to move away from global classes.

camdencheek:
Page reflow when loading completes. We should make sure this has a fixed height


https://github.com/sourcegraph/sourcegraph/assets/12631702/c178c569-9138-498d-a74b-c8233fd16b86

camdencheek:
Spacing on the left margin of the info icon doesn't look like enough. Looks like we add some more padding in the designs too. Most easily visible when hovering and you can see the background color of the element

CleanShot 2024-03-22 at 14 48 55@2x

client/web-sveltekit/src/lib/search/resultsIndicator/SuggestedAction.svelte
```diff
@@ -0,0 +1,91 @@
...
+    import { capitalizeFirstLetter } from '$lib/search/utils'
+    import type { Progress, Skipped } from '$lib/shared'
+
+    export let isError: boolean
+    export let progress: Progress
+    export let hasSkippedItems: boolean
+    export let hasSuggestedItems: boolean
+    export let mostSevere: Skipped
+
+    const INTERPUNCT = '\u00B7'
```
camdencheek, line 11: I don't feel strongly about this, but non-ASCII text _is_ allowed in code files. That way ya don't need to go look up what "INTERPUNCT" is when reading the source 😄

client/web-sveltekit/src/lib/search/resultsIndicator/SuggestedAction.svelte
```diff
@@ -0,0 +1,91 @@
...
+    .code-font {
+        background-color: var(--secondary);
+        border-radius: 3px;
+        color: var(--text-body);
+        font-family: var(--code-font-family);
+        font-size: 0.7rem;
+        padding-top: 0.2rem;
+        padding-bottom: 0.2rem;
+        padding-left: 0.2rem;
+        padding-right: 0.2rem;
```
camdencheek, line 56:
```suggestion
        padding: 0.2rem;
```

client/web-sveltekit/src/lib/search/resultsIndicator/SuggestedAction.svelte
```diff
@@ -0,0 +1,91 @@
...
+
+    .info-badge.duration {
+        background: var(--warning-2);
+        color: var(--text-body);
+    }
+
+    .info-badge.error-text {
+        background: var(--danger-2);
+        color: var(--text-body);
+    }
```
camdencheek, line 75: Personal preference: I like nesting these in the `.info-badge` block with `&.duration` so that it's more visually clear what the hierarchy is

client/web-sveltekit/src/lib/search/resultsIndicator/TimeoutMessage.svelte
```diff
@@ -0,0 +1,58 @@
...
+            for background search
+        </div>
+    </div>
+</div>
+
+<style lang="scss">
+    .duration-badge {
+        display: flex;
+        flex-flow: row-nowrap;
+        // margin-top: 0.2rem;
```
camdencheek, line 23: stray comment?

client/web-sveltekit/src/lib/search/resultsIndicator/TimeoutMessage.svelte
```diff
@@ -0,0 +1,58 @@
...
+
+    .error-text {
+        color: var(--danger);
+    }
+
+    .info-badge {
+        border-radius: 3px;
+        color: var(--text-body);
+        padding-left: 0.2rem;
+        padding-right: 0.2rem;
```
camdencheek, line 34:
```suggestion
        padding: 0 0.2rem;
```

client/web-sveltekit/src/routes/search/SearchResults.svelte
camdencheek, line 22: I think we prefer importing from `$lib/branded` rather than importing from `@sourcegraph` directly

client/web-sveltekit/src/routes/search/SearchResults.svelte
```diff
@@ -208,7 +203,7 @@
             align-items: center;
             // Explictly set height to avoid jumping when loading spinner is
             // shown/hidden.
-            height: 3rem;
+            // height: 4rem;
```
camdencheek, line 206: stray comment?

client/web-sveltekit/src/routes/search/StreamingProgress.svelte
```diff
@@ -115,31 +120,33 @@
...
     }
 
-    label {
-        display: block;
-    }
-
     .progress-button {
         border: none;
+        border-radius: 3px;
+        padding: 0.5rem 0.5rem 0.5rem 0.2rem;
```
camdencheek, line 140: It's suspicious to me that we adjust the padding in one direction. Is there a margin somewhere that we can remove so this element has consistent padding?

camdencheek:
The most common scenario where we are missing results is when we exclude forked or archived repos, so this will almost always show like this. I don't think this is actually valuable though, and it's kinda a confusing presentation of information (why do we only show excluded forks and not excluded archived repos? is it saying 104 forks are searched or excluded? why is include forked, which is phrased as an action, not clickable?)

IMO, we should only display these things in the component itself for the scenarios in the design, which is just when a search limit is hit.

CleanShot 2024-03-22 at 15 03 29@2x

client/web-sveltekit/src/lib/search/resultsIndicator/ResultsIndicator.svelte
camdencheek: Resolving this since Felix also addressed it and his comment is more comprehensive

client/web-sveltekit/src/lib/search/resultsIndicator/TimeoutMessage.svelte
camdencheek: It should use `/help`, which redirects to the version of the docs corresponding to the instance

jasonhawkharris: Are we sure about the `/help` redirect? It doesn't seem to work.

client/web-sveltekit/package.json
jasonhawkharris: Hmm... test:svelte seems pretty intuitive to me. Do you have a suggestion for another name for this test?

client/web-sveltekit/src/lib/search/resultsIndicator/InfoBadge.svelte
jasonhawkharris: What do you mean by code-smelly? I don't know this term.

client/web-sveltekit/src/lib/search/resultsIndicator/ResultsIndicator.svelte
jasonhawkharris: Ahhh... good call. That's a much better name. I'll change it.

client/web-sveltekit/src/lib/search/resultsIndicator/TimeoutMessage.svelte
camdencheek: Did you remove the `/docs` portion? This URL redirects for me: `https://sourcegraph.sourcegraph.com/help/code-search/types/search-jobs`

client/web-sveltekit/src/lib/search/resultsIndicator/InfoBadge.svelte
camdencheek: A "code smell" is just something that doesn't pass the sniff test -- something that suggests there are bigger problems. 

In this case, manually setting font size is a suggestion that we probably aren't using standardized styling for fonts, which can make the web app feel non-cohesive. We don't want too many font sizes, styles, colors, etc., so using semantic tags like `<small>` and letting the base style sheets decide what that means in terms of actual size usually makes for more maintainable code.

client/web-sveltekit/src/lib/search/resultsIndicator/ResultsIndicator.svelte
jasonhawkharris: Keeping the icon class because I'm using it to add margins.

client/web-sveltekit/src/lib/search/resultsIndicator/SuggestedAction.svelte
jasonhawkharris: I went back and forth on calling this `CENTER_DOT` (more descriptive) and `INTERPUNCT` (what it actually is). I'll change it back to `CENTER_DOT` since understanding is the priority here, term accuracy.

client/web-sveltekit/src/lib/search/resultsIndicator/SuggestedAction.svelte
jasonhawkharris: This looks to be leftover from an earlier on iteration and is no longer needed. Removing it.

client/web-sveltekit/src/lib/search/resultsIndicator/InfoBadge.test.ts
jasonhawkharris: I did. I also did this at the suggestion of Michael.

client/web-sveltekit/src/lib/search/resultsIndicator/SuggestedAction.svelte
camdencheek: I think I might have buried the lede here. I was actually talking about putting a _literal_ interpunct here. So anyone reading this can _see_ exactly what the character is. Again, nbd, but it's an option if you think that makes it more clear

camdencheek: CleanShot 2024-03-23 at 13 20 43@2x

bahrmichael:
Removed myself as a reviewer because there's quite a bit of feedback already. Don't want to add more input that could be conflicting with others.

client/web-sveltekit/src/lib/search/resultsIndicator/ResultsIndicator.svelte
```diff
@@ -0,0 +1,119 @@
...
+</div>
+
+<style lang="scss">
+    .icon {
+        padding-right: 0.4rem;
+    }
+
+    .dropdown-icon {
+        padding-left: 1.2rem;
+    }
```
camdencheek, line 94: Here's another example where I think `padding` is actually the wrong choice because this is space _between_ elements in our div. So we couldn't, for example, take the icon element and move it into its own component because the padding would no longer make sense there.

Since this is spacing _within_ the component (and doesn't escape _outside_ the component), this is a good candidate for a flexbox `gap` or a `margin`

client/web-sveltekit/src/lib/search/resultsIndicator/ResultsIndicator.svelte
```diff
@@ -0,0 +1,119 @@
...
+        align-content: center;
+        align-items: flex-start;
+        display: flex;
+        flex-flow: column nowrap;
+    }
+
+    .running-search {
+        color: var(--text-muted);
+        padding-top: 0.3rem;
+    }
```
camdencheek, line 113: This is another spot where we can simplify the DOM and remove padding that needs to stay in sync. Let's move the `action-container` div outside of the `if` branch and remove the spacing from the `SuggestedAction` component. 

We should address this before merging

client/web-sveltekit/src/lib/search/resultsIndicator/SuggestedAction.svelte
```diff
@@ -0,0 +1,89 @@
...
+    $: mostSevere = sortedItems[0]
+    $: done = progress.done
+</script>
+
+<div class="action-container" class:error-text={isError}>
+    <div class="suggested-action">
+        {#if done && !hasSkippedItems}
+            <small>
+                <div class="more-details">See more details</div>
+            </small>
```
camdencheek, line 27: nit: `<small>` is text markup, so it should live inside the div, which is a block element.

client/web-sveltekit/src/lib/search/resultsIndicator/SuggestedAction.svelte
```diff
@@ -0,0 +1,89 @@
...
+
+        {#if done && mostSevere && Object.hasOwn(mostSevere, 'suggested')}
+            <div class="separator">{CENTER_DOT}</div>
+            <div class="action-badge">
+                <small>
+                    {capitalize(mostSevere?.suggested ? mostSevere.suggested.title : '')}&nbsp;
+                    <span class="code-font">
+                        <small>{mostSevere.suggested?.queryExpression}</small>
+                    </span>
+                </small>
```
camdencheek, line 46: why do we have nested `<small>` here?

client/web-sveltekit/src/lib/search/resultsIndicator/ResultsIndicator.svelte
camdencheek: @jasonhawkharris I don't think this has been resolved. We do need to unsubscribe from the timer, otherwise it will just continue running forever, potentially making `elapsedDuration` do weird things.

We should address this before merge

client/web-sveltekit/src/lib/search/resultsIndicator/TimeoutMessage.svelte
```diff
@@ -0,0 +1,43 @@
...
+        background-color: var(--warning-2);
+        border-radius: 3px;
+        color: var(--text-body);
+        padding-left: 0.2rem;
+        padding-right: 0.2rem;
+    }
+
+    .separator {
+        padding-left: 0.4rem;
+        padding-right: 0.4rem;
```
camdencheek, line 36: nit: collapsing these into `padding: 0 0.4rem` generally communicates "this padding should be horizontally symmetrical" better

dev/linters/bodyclose/BUILD.bazel
jasonhawkharris: @camdencheek Any idea why these might have been added?

camdencheek: Did you run `bazel generate` instead of `sg bazel generate`?

jasonhawkharris: yes. Damnit.

camdencheek (approved):
I think this is good to merge! 🥳 (once CI passes)

I think we should follow up with some design tweaks. In particular, I do not think we should show "104 forked" since 1) it implies those are part of the search results, not excluded, and 2) there are other things excluded that aren't mentioned (archived). Instead, I think we should have some sort of "some results excluded" message that guides users to the popover for more info. 

@jasonhawkharris, could you create an issue for this followup and tag me and @taiyab?

CleanShot 2024-03-26 at 11 19 51@2x

vovakulikov (commented):
I'm bit late here, but just want to highlight that we skipped search jobs UI here. Not a blocker or regression but something we need to add before public rollout (whenever it happens)

camdencheek:
@vovakulikov Did we? What are you referring to here?

CleanShot 2024-03-28 at 12 13 10@2x

vovakulikov:
@camdencheek this screenshot from the react version, right? I don't see search job UI in the popover in svelte version

PR #61036, opened 2024-03-12
web/ExternalsAuth: fix dotcom sign-in widget

bobheadxi:
With the recent migration to SAMS on Sourcegraph.com, this widget has stopped working because it used to pull from dotcom config and look for some known auth providers. These auth providers no longer exist, because SAMS is the only one now, so we use some SAMS magic to direct users to the SAMS version of the per-external-service login pages. This is similar to what we do in the VSCode extension.

The most interesting diff is `client/web/src/auth/components/ExternalsAuth.tsx` - everything else is just changing the prop-drilled context.

There seems to be something wrong with the CSS, presumably because it's been missing for a while, but please don't make me fix that 😁 

## Test plan

`sg start dotcom`:

image

The links don't work directly for logging in to a local "dotcom" instance without some additional configuration, so we copy the target URLs and change the link to be `sourcegraph.com`, and pretend to use them to log in to Sourcegraph.com (after logging out):

```diff
- https://sourcegraph.test:3443/.auth/openidconnect/login?prompt_auth=gitlab&pc=sams
+ https://sourcegraph.com/.auth/openidconnect/login?prompt_auth=gitlab&pc=sams
```

Login works as expected.


unknwon (approved):

vdavid (approved):
Beautiful

rrhyne (approved):
Thanks for putting this together! It will certainly have better conversion with this method!

Issue #60859, opened 2024-03-05
Remove/adjust prototype specific messaging/UI elements


fkling:
We should make a decision about whether we keep messaging/UI elements that inform the user that this is a prototype/the experimental web app. And if yes what we want to show. Currently we have the following elements:

- "Info banner" at the top of the page (like global alerts)
- Icons next to links to indicate that they lead back to the React app

I think it makes sense to show _some_ indication that this is the experimental web app. I'm not sure about whether we should give people an easy way to opt-out. Ideally it's easier to complain (and for us to fix it) rather than to disable the prototype.

fkling:
Decision we made:

- Do not decorate links that leave the Svelte app
- Provide a simple way to _permanently_ opt out of the Svelte rollout
- Replace the beaker button in the React app with a button to opt back into the Svelte rollout.

Engineering work is tracked in #61298

PR #61038, opened 2024-03-12
svelte: Update messaging and link indicators

fkling:
Relates to #60859 

This commit removes the info banner and adds an experimental badge with popover instead. I'm sure the messaging in the popover can be improved. Maybe we'll give people a way to permanently opt-out.

This also removes the styling for external links, which, toghether with the changes to the repo URL matching, means that the `data-sveltekit-reload` attributes are not necessary anymore.

Because I made changes to the popover logic, namely making it possible to show and hide the popover on hover, I used that opportunity to migrate from the deprecated popperjs to floating-ui, which resolves some issues with the tooltip in hovercards.
Eventually we should probably standardize on using melt-ui for popovers too (or whatever library we will use eventually) but for now I still find floating-ui to be easier to use than the corresponding melt-ui components.

| Tooltip before | Tooltip after |
|--------|--------|
|  |  |

## Test plan

Manual testing


client/web-sveltekit/src/routes/Header.svelte
```diff
@@ -38,18 +41,37 @@
             {/each}
         </ul>
     </nav>
-    <Tooltip tooltip="Leave experimental web app">
-        <a href={reactURL} data-sveltekit-reload>
-            <Icon svgPath={mdiFlaskOutline} --color="var(--oc-green-6)" />
-        </a>
-    </Tooltip>
+    <Popover let:registerTrigger showOnHover>
```
vovakulikov, line 44: I understand that this isn't the place for this change, just thinking out loud (I believe I mentioned this before but just for the record once again)

So I think understand the current API (kind of render props approach from react, and angular composition templates) I still think though that we could share these props (register trigger) via svelte context (or whenever it's called in svelte) and use Popover compound components instead (like we do in wildcard react components)

fkling: You mean something like:

```svelte
<Popover>
  <Popover.Trigger> .... </Popover.Trigger>
  ...
</Popover>
```

?

vovakulikov: Correct

PR #61040, opened 2024-03-12
Clean-up Dotcom specific checks

evict:
This PR cleans up some Dotcom specific checks. It removes a check on:

- Listing organizations: allows site-admins on Dotcom to list organizations. People can make organizations still, so it makes sense we should be able to list them so we can remove them.
- User settings adminstration by a site-admin: on Dotcom we no longer have private code, so there shouldn't be any secrets related to codehosts in user settings.

## Test plan
CI tests.

### DIFF
--- a/cmd/frontend/graphqlbackend/orgs_test.go
+++ b/cmd/frontend/graphqlbackend/orgs_test.go
@@ -3,8 +3,6 @@ package graphqlbackend
 import (
 	"testing"
 
-	"github.com/graph-gophers/graphql-go/errors"
-
 	"github.com/sourcegraph/sourcegraph/internal/database/dbmocks"
 	"github.com/sourcegraph/sourcegraph/internal/dotcom"
 	"github.com/sourcegraph/sourcegraph/internal/types"
@@ -52,7 +50,7 @@ func TestOrgs(t *testing.T) {
 	})
 }
 
-func TestListOrgsForCloud(t *testing.T) {
+func TestListOrgsForDotcom(t *testing.T) {
 	orig := dotcom.SourcegraphDotComMode()
 	dotcom.MockSourcegraphDotComMode(true)
 	defer dotcom.MockSourcegraphDotComMode(orig)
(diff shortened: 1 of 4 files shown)


eseliger (approved):
Nice cleanup, code change LGTM, security-wise this is your call 😇

cmd/frontend/graphqlbackend/user.go
```diff
@@ -375,18 +375,10 @@ func (r *UserResolver) settingsSubject() api.SettingsSubject {
 }
 
 func (r *UserResolver) LatestSettings(ctx context.Context) (*settingsResolver, error) {
-	// 🚨 SECURITY: Only the authenticated user can view their settings on
-	// Sourcegraph.com.
-	if dotcom.SourcegraphDotComMode() {
```
evict, line 380: @unknwon do you remember why this check was introduced? I couldn't find much information about it!

andreeleuterio: You are correct sir, this dates back to two PLG efforts ago 😅

cmd/frontend/graphqlbackend/orgs_test.go
```diff
@@ -52,7 +50,7 @@ func TestOrgs(t *testing.T) {
 	})
 }
 
-func TestListOrgsForCloud(t *testing.T) {
+func TestListOrgsForDotcom(t *testing.T) {
```
unknwon, line 53: I think we can remove this test entirely? Two test cases are almost identical now.

cmd/frontend/graphqlbackend/user.go
unknwon: Back to 2022 when we wanted private code on dotcom? 😂

cmd/frontend/graphqlbackend/user_test.go
```diff
@@ -180,35 +180,40 @@ func TestUser_LatestSettings(t *testing.T) {
 	t.Run("only allowed by authenticated user on Sourcegraph.com", func(t *testing.T) {
```
unknwon, line 180: This entire test suite can be removed as dotcom-specific?

unknwon (approved):
All good except I think we can be a bit more aggressive in few cases 😁 Thanks for cleaning things up!

PR #61044, opened 2024-03-12
executors: Fix version detection for V2 payloads

eseliger:
This is.. bad. This broke docker auth via executor secrets because the V0 payload didn't support it. This could also affect non-utf-8 encoded virtual machine file contents.

## Test plan

Verified that the new syntax works correctly.


camdencheek (approved):
Is there more context about this somewhere? I don't really know what I'm looking at.

eseliger:
Yes: https://sourcegraph.slack.com/archives/C03JR7S7KRP/p1710184258933689 sorry, I was about to reach out in your team channel but was just getting a second thing ready for review :D

PR #61045, opened 2024-03-12
executor: Fix global secrets across namespaces colliding

eseliger:
Found this while debugging another issue:
When there are two executor secrets for both global scopes we currently support that have the same name, they will appear as both Global Secret, and Overwrites Global Secret in the UI.

So far so confusing, so let's fix that (because they're not actually overwriting anything, scopes are separate namespaces).

Turns out this also breaks detection for uniqueness of secret values in user namespaces, so when two global DOCKER_AUTH_CONFIG secrets exist (one for each scope), users cannot overwrite this secret value anymore, getting an error message like below.

image

## Test plan

Manually verified that things work correctly now and adjusted a test.

### DIFF
--- a/internal/database/executor_secrets.go
+++ b/internal/database/executor_secrets.go
@@ -401,7 +401,7 @@ var executorSecretsColumns = []*sqlf.Query{
 	sqlf.Sprintf("key"),
 	sqlf.Sprintf("value"),
 	sqlf.Sprintf("encryption_key_id"),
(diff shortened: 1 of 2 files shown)


bahrmichael (approved):

BolajiOlajide (approved):

BolajiOlajide:
Ooopsie.

PR #61047, opened 2024-03-12
svelte: Improve TS types for GraphQL mocks

fkling:
Before this commit the "mock types" had been generated by simply applying `DeepPartial` (which is like `Partial` but applies recursively) to the existing GraphQL types.

This commit generates the TS types "from scratch". This allows us to customize how TS types for mocks are defined. Concrete differences (so far):

- Mock types for interfaces have a `__typename` field that is a union of the type names of all implementations of that interface. This should make it easier to define mocks of concreate types for fields that accept an interface.
- Fields that take an enum are now defined as string union, which allows speciying the enum value as string literal (with completion support), instead of having to import the actual enum.

Example of generated code:

```ts
export interface AggregationModeAvailabilityMock {
  __typename?: 'AggregationModeAvailability',
  /**
   * Boolean indicating if the mode is available
   */
  available?: Scalars['Boolean']['output'],
  /**
   * The SearchAggregationMode
   */
  mode?: 'AUTHOR' | 'CAPTURE_GROUP' | 'PATH' | 'REPO' | 'REPO_METADATA',
  /**
   * If the mode is unavailable the reason why
   */
  reasonUnavailable?: Scalars['String']['output'] | null,
}
```

Note that this only affect type mocks, not operation mocks. Those require more effort to process. I might do that in a future PR.

## Test plan

`pnpm build` -> to inspect graphql codegen output
`pnpm check` -> to find TypeScript issues

Manual inspection of generated files.

### DIFF
--- a/client/web-sveltekit/dev/graphql-type-mocks.cjs
+++ b/client/web-sveltekit/dev/graphql-type-mocks.cjs
@@ -1,6 +1,15 @@
 // @ts-check
-const { isObjectType, visit, concatAST } = require('graphql')
-const { isScalarType } = require('graphql')
+const {
+    visit,
+    concatAST,
+    isObjectType,
+    isNonNullType,
+    isListType,
+    isInterfaceType,
+    isUnionType,
+    isEnumType,
+    isScalarType,
+} = require('graphql')
 const logger = require('signale')
 
 /**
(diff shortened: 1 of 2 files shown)


client/web-sveltekit/dev/graphql-type-mocks.cjs
```diff
@@ -17,26 +26,123 @@ function documentsToAST(documents) {
...
+            .map(type => generateTSTypeForGraphQLType(type, indent))
+            .join(' | ')
+    }
+    if (isObjectType(type) || isInterfaceType(type)) {
+        return getMockTypeName(type)
+    }
+    if (isEnumType(type)) {
+        return type
+            .getValues()
+            .map(value => `'${value.name}'`)
```
vovakulikov, line 117: Does this mean that mocks type will have only string literals of enum? and actual enum values won't be supported?

vovakulikov (approved):

client/web-sveltekit/dev/graphql-type-mocks.cjs
fkling: If by 'actual enum value' you mean something like `AlertType.WARNING` I think that should still work since `Alert.WARNING` will be of type `'WARNING'`.

fkling: It should be easy to add the actual enum as possible value too though. If we run into issues we can still do that.

PR #60911, opened 2024-03-07
rename "Subscriptions" to "Enterprise licenses" to avoid confusion with Cody Pro

sqs:
Some users have gotten Sourcegraph Enterprise license keys confused with Cody Pro subscriptions because the former was also called "Subscriptions" in the user settings area ([example](https://discord.com/channels/969688426372825169/1212987159791800351/1213121019397546024)). This renames "Subscriptions" to "Enterprise licenses" in the UI. The API and URLs remain unchanged for compatibility.

Docs PR: https://github.com/sourcegraph/docs/pull/125

## Test plan

Run `sg start dotcom` and confirm that the sidebar labels and page titles are updated.

PR #61049, opened 2024-03-13
rename 'Enterprise licenses' to 'Enterprise subscriptions' to accurately represent enterprise accounts

bobheadxi:
https://github.com/sourcegraph/sourcegraph/pull/60911 renamed "Subscriptions" to "Enterprise licenses", but the two are not equivalent, as "subscriptions" are an entity above individual licenses - for example `Create new Enterprise license` is not equivalent to the previous `Create new product subscription`. This entity is important because there are some elements of these subscriptions that are more subscription-oriented, than license-oriented - for example, Cody Gateway rate limits exist at an "Enterprise subscription" level and are not necessarily tied to individual licenses (though they can affect what is evaluated at the "subscription" level).

In [RFC 885 Sourcegraph Enterprise Portal](https://docs.google.com/document/d/1tiaW1IVKm_YSSYhH-z7Q8sv4HSO_YJ_Uu6eYDjX7uU4/edit#heading=h.tdaxc5h34u7q) outlining how Core Services plans to change this area (and also the existing use cases of a "subscription"-like entity), we've taken to referring to these as "Enterprise subscriptions" to differentiate from PLG ones. This naming more accurately reflects the model while avoiding confusion with PLG subscriptions.

See thread: https://sourcegraph.slack.com/archives/C04MEF2T33M/p1710292839852869?thread_ts=1709789262.898779&cid=C04MEF2T33M

This PR applies the new naming and also fixes a few spots missed in https://github.com/sourcegraph/sourcegraph/pull/60911

## Test plan

`sg start dotcom`

image
image
image

### DIFF
--- a/client/web/src/enterprise/site-admin/dotcom/productSubscriptions/SiteAdminLicenseKeyLookupPage.tsx
+++ b/client/web/src/enterprise/site-admin/dotcom/productSubscriptions/SiteAdminLicenseKeyLookupPage.tsx
@@ -47,11 +47,11 @@ export const SiteAdminLicenseKeyLookupPage: React.FunctionComponent<React.PropsW
 
     return (
         <div className="site-admin-product-subscriptions-page">
-            <PageTitle title="Product subscriptions" />
+            <PageTitle title="Enterprise subscriptions" />
             <PageHeader
                 path={[{ text: 'License key lookup' }]}
                 headingElement="h2"
-                description="Find matching licenses and their associated product subscriptions"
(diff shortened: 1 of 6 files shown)


sqs (approved):

eseliger (approved):
Can you also fixup the handbook?

bobheadxi:
> Can you also fixup the handbook?

PR is in https://github.com/sourcegraph/handbook/pull/8710, and https://github.com/sourcegraph/docs/pull/150 for docs as well 😄

Issue #57499, opened 2023-10-09
Bulk Update and Individual Management of Embeddings


gabtorre:
#### Feature request description

Requested by https://github.com/sourcegraph/accounts/issues/548

The current system doesn't allow for a bulk update or deletion of embeddings, which results in a significant overhead for users who are iterating configs over many repos.

Bulk Update Feature: A mechanism to forcibly update all embeddings collectively.

Individual Management: The capability to directly update or delete a specific embedding without influencing others.

Auto-indexing index management is attached as an example.

![Screenshot 2023-10-09 at 4 13 04 PM](https://github.com/sourcegraph/sourcegraph/assets/69164745/a61a0646-c613-4c75-9aa3-3af9bf3efc8b)

#### Describe alternatives you've considered.

Manual updates

jtibshirani:
Closing this out as we're not investing in embeddings at the moment.

PR #61052, opened 2024-03-13
worker: remove embeddings

keegancsmith:
I noticed a bunch of logspam since we effectively disabled embeddings by always having its configuration turned off.

```
WARN worker.repo-embedding-job.Handle repo/handler.go:62 embeddings are not configured or disabled
```

Test Plan: go test ./cmd/worker/...


keegancsmith:
@stefanhengl should we keep the janitor around to help clean up?

stefanhengl:
> @stefanhengl should we keep the janitor around to help clean up?

I don't think so. It just operates on the worker table which we don't need anymore. Talking of tables ... should we also remove the embeddings worker tables? What about the site admin pages?

jtibshirani:
@keegancsmith would you be okay if we closed this out? We had discussed as a team and decided we'd like to keep embeddings around since they are so useful for evals and prototyping, and the code is well-contained.

nihaals:
Is the plan to still not expose this to users and essentially have it for internal use? Embeddings currently feel like they're in limbo as existing ones still exist and are shown in the UI but you can't seem to delete them and it's unclear if just removing the embedding store is fine or if users are meant to manually delete entries from the database too since #57499 wasn't done in time.

jtibshirani:
@nihaals the recommendation is just to remove the embeddings service. To reclaim space, you can also remove the `embeddings` directory from blobstore. Embeddings take up little room in the main Postgres database, so it is not critical to delete entries there.

nihaals:
> Embeddings take up little room in the main Postgres database, so it is not critical to delete entries there.

Could this leave it in an inconsistent state where essentially references to embeddings that no longer exist are stored and potentially cause problems in the future if embeddings are used again? I also assume the admin page will still show repos having embeddings that don't actually exist anymore. Repos that no longer exist still have embeddings stored as shown by the total embedded repos count so it doesn't seem like any leftover data is purged.

jtibshirani:
We have ways of making sure we can "pick up where we left off" and not get into an inconsistent state, like forcing a reindex. We also don't have plans at the moment to bring back the current embeddings service. In future versions, we'll also hide references to "embeddings" in the admin page so there won't be confusion.

Issue #60745, opened 2024-02-24
Search: digression between indexed and unindexed search with `file:contains.content()`


camdencheek:
Noticed while debugging something else, but these two searches return different results:

```
repo:github.com/sourcegraph/sourcegraph$ file:contains.content("ResourceStateEvents": null\s*\}\s*)

repo:github.com/sourcegraph/sourcegraph$@HEAD~1 file:contains.content("ResourceStateEvents": null\s*\}\s*)
```

The unindexed search yields no results. I have not looked into this at all, but it's easily reproducible locally.

/cc @sourcegraph/search-platform

PR #61053, opened 2024-03-13
searcher: use same logic for IsRegExp as zoekt

keegancsmith:
We use the same logic as indexed search (and commit search) on deciding if something is a regex. In theory this should mean the same thing, but in practice we have bugs so lets use consistent behaviour.

https://github.com/sourcegraph/sourcegraph/blob/095a429bec15d40fd506a8d20ad76a152b2fcd02/internal/search/zoekt/query.go#L149-L151

This issue was found when investigating why a query like `file:contains.content(foo\s*bar)` would have results with index:yes but not with index:no. Turns out the Pattern node currently doesn't set the relevant annotations. That will be fixed in another commit. Currently our UX highlights the inner string with regex highlighting, so right now this is the expected behaviour. We can debate as part of keyword work whether the pattern inside content should require /re/ escaping.

Test Plan: CI. Note I tried to add a test case for this but it was quite somewhat difficult to:
- adjust protocol's tests to generate a pattern with no annotations.
- test consistency across zoekt and searcher.

If anyone has advice on how to test, I would appreciate it. More than likely the best tests will come in the next PR which fixes the missing annotations.

Fixes https://github.com/sourcegraph/sourcegraph/issues/60745


cmd/searcher/protocol/query.go
```diff
@@ -27,7 +27,7 @@ func FromJobNode(node query.Node) QueryNode {
 		return &PatternNode{
 			Value:     n.Value,
 			IsNegated: n.Negated,
-			IsRegExp:  n.Annotation.Labels.IsSet(query.Regexp),
+			IsRegExp:  !n.Annotation.Labels.IsSet(query.Literal),
```
stefanhengl, line 30: We should probably leave a comment. I can see myself coming across this in 3 months reverting this change thinking it must be a bug.

mmanela: What does the query get determined to be literal? Is that coming from the client?

keegancsmith: @stefanhengl please take a look at the commit I just pushed that makes this less fragile.

@mmanela this stuff is from our internal query parser from a search query. We have this logic copy pasted across deciding if the pattern value is literal or regex. I just pushed a commit to make this clearer.

internal/search/query/parser.go
```diff
@@ -73,6 +74,27 @@ func (node Pattern) String() string {
...
 }
 
+// IsRegExp returns true if the Pattern.Value should be interpreted as a Regex
+// otherwise returns false for a Literal.
+//
+// Note: This checks that the relevant annotation is set, which occasionally
+// we regress on setting. In such situations it will return that the Pattern
+// is Regex. Use this method so we have consistent behaviour across our
+// backends rather than directly checking annotations when building queries
+// for your backend.
```
jtibshirani, line 84: Nice, I like how this documents how we interpret the labels. Previously it was unclear what the default behavior is when no labels are set.

jtibshirani (approved):
> If anyone has advice on how to test, I would appreciate it.

I believe this is covered by our existing job util tests (now that we are including the pattern info everywhere) and end-to-end tests. As for enforcing consistency, using a shared method like this feels solid.

I looked for references to `Labels.IsSet` and saw a few more places where we manually check for `Literal` and `Regexp`. Maybe we could update those too in this PR? Overall we could plan to move away from directly checking labels in favor of simple methods like the ones you added.

keegancsmith:
> > If anyone has advice on how to test, I would appreciate it.
> 
> I believe this is covered by our existing job util tests (now that we are including the pattern info everywhere) and end-to-end tests. As for enforcing consistency, using a shared method like this feels solid.

Yeah now that I added the pattern we pass to searcher in the other PR, those tests cover this. Before it didn't really capture this.

> I looked for references to `Labels.IsSet` and saw a few more places where we manually check for `Literal` and `Regexp`. Maybe we could update those too in this PR? Overall we could plan to move away from directly checking labels in favor of simple methods like the ones you added.

Yes there are a few more places. However, they aren't as obvious a use case (just yet!) and one I believe has a bug I wanna double check before changing. I will follow-up with those places.

PR #61056, opened 2024-03-13
symbols: respect limits, merge properly

### LINKED ISSUE #60698: symbols: keyword search interpreted differently for indexed vs unindexed
The query `repo:^github\.com/sourcegraph/sourcegraph$ repo link picker type:symbol` has two results via indexed search and no results if via unindexed search. The indexed path is the correct one here, since there is a symbol called RepoLinkPicker. Technically what Zoekt is doing is searching for any document which has those symbols not just a single symbol with those terms. So I suspect there is a bigger issue here with how symbol search interprets queries.

- Indexed link https://sourcegraph.sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/sourcegraph%24+repo+link+picker+type:symbol+&patternType=keyword&sm=0 
- Unindexed link https://sourcegraph.sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/sourcegraph%24%40main+repo+link+picker+type:symbol+&patternType=keyword&sm=0

Follow-up from an extra I noticed while debugging https://github.com/sourcegraph/sourcegraph/issues/60379

stefanhengl:
fixes #60698

This fixes the following bugs for unindexed symbol search:

1. Now we respect the limit given by "frontend". Before, the upper limit of unindexed symbol search was hard coded to 500. This limit could not be overwritten with the `count:` filter. 
2. Merge symbol matches properly, applying the same logic as we do for indexed search: Before, symbol matches within the same file were simply appended which created duplicates for AND/OR queries if separate terms of the query matched the same symbol.

Test plan:
- New unit test
- Partially covered by existing tests

### DIFF
--- a/cmd/symbols/internal/api/handler.go
+++ b/cmd/symbols/internal/api/handler.go
@@ -11,12 +11,9 @@ import (
 	internalgrpc "github.com/sourcegraph/sourcegraph/internal/grpc"
 	"github.com/sourcegraph/sourcegraph/internal/grpc/defaults"
 	"github.com/sourcegraph/sourcegraph/internal/search"
-	"github.com/sourcegraph/sourcegraph/internal/search/result"
 	proto "github.com/sourcegraph/sourcegraph/internal/symbols/v1"
 	internaltypes "github.com/sourcegraph/sourcegraph/internal/types"
 )
-
-const maxNumSymbolResults = 500
 
 type grpcService struct {
 	searchFunc   types.SearchFunc
@@ -60,21 +57,12 @@ func NewHandler(
 	handleStatus func(http.ResponseWriter, *http.Request),
 	ctagsBinary string,
 ) http.Handler {
-	searchFuncWrapper := func(ctx context.Context, args search.SymbolsParameters) (result.Symbols, error) {
-		// Massage the arguments to ensure that First is set to a reasonable value.
-		if args.First < 0 || args.First > maxNumSymbolResults {
-			args.First = maxNumSymbolResults
-		}
-
-		return searchFunc(ctx, args)
-	}
-
 	rootLogger := logger.Scoped("symbolsServer")
 
 	// Initialize the gRPC server
 	grpcServer := defaults.NewServer(rootLogger)
 	proto.RegisterSymbolsServiceServer(grpcServer, &grpcService{
-		searchFunc:   searchFuncWrapper,
+		searchFunc:   searchFunc,
 		readFileFunc: readFileFunc,
 		ctagsBinary:  ctagsBinary,
 		logger:       rootLogger.Scoped("grpc"),
--- a/cmd/symbols/internal/database/store/search.go
+++ b/cmd/symbols/internal/database/store/search.go
@@ -44,7 +44,14 @@ func scanSymbols(rows *sql.Rows, queryErr error) (symbols []result.Symbol, err e
 	return symbols, nil
 }
 
+const defaultLimit = 100
+
 func (s *store) Search(ctx context.Context, args search.SymbolsParameters) ([]result.Symbol, error) {
+	limit := defaultLimit
+	if args.First > 0 {
+		limit = args.First
+	}
+
 	return scanSymbols(s.Query(ctx, sqlf.Sprintf(
 		`
 			SELECT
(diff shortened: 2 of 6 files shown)


cmd/symbols/internal/database/store/search.go
```diff
@@ -44,7 +44,14 @@ func scanSymbols(rows *sql.Rows, queryErr error) (symbols []result.Symbol, err e
 	return symbols, nil
 }
 
+const defaultLimit = 100
```
stefanhengl, line 47: This is the same limit Rockskip uses

internal/search/result/file.go
```diff
@@ -164,15 +166,19 @@ func (fm *FileMatch) Select(selectPath filter.SelectPath) Match {
 	return nil
 }
 
-// AppendMatches appends the line matches from src as well as updating match
-// counts and limit.
 func (fm *FileMatch) AppendMatches(src *FileMatch) {
 	// TODO merge hunk matches smartly
 	fm.ChunkMatches = append(fm.ChunkMatches, src.ChunkMatches...)
-	fm.Symbols = append(fm.Symbols, src.Symbols...)
+	fm.appendSymbols(src)
```
keegancsmith, line 172: how often is AppendMatches called in a FileMatch where you need this dedup logic? If its per "job leaf" then this should be fine since that is a relatively small number. However, if this happens as we stream in this effectively becomes an O(n^2logn) algorithm over all calls to AppendMatches. I think it is the former so should be fine. But just wanna call that out.

Additionally, we should probably rename this method to indicate it is doing more work than just append to avoid surprises.

keegancsmith (approved):

internal/search/result/file.go
stefanhengl: Renamed it to "mergeSymbols". 

Apart from the leaf logic of flat jobs touched in this PR, AppendMatches is called in `Deduper.Add`, which is odd, because no caller of `Deduper` ever reads the deduped results. I will follow up with a PR that removes the logic.

cmd/symbols/internal/api/handler.go
```diff
@@ -11,13 +11,10 @@ import (
 	internalgrpc "github.com/sourcegraph/sourcegraph/internal/grpc"
 	"github.com/sourcegraph/sourcegraph/internal/grpc/defaults"
 	"github.com/sourcegraph/sourcegraph/internal/search"
-	"github.com/sourcegraph/sourcegraph/internal/search/result"
 	proto "github.com/sourcegraph/sourcegraph/internal/symbols/v1"
 	internaltypes "github.com/sourcegraph/sourcegraph/internal/types"
 )
 
-const maxNumSymbolResults = 500
```
jtibshirani, line 19: Do we have any historical knowledge of why this was added? I do wonder if we need some limit here (but a much higher one), as we don't perform any sort of DB paging and streaming, and could be pulling back a huge number of results? 

We could do something like validate the limit is in the range [0, 5000] and throw an error otherwise?

internal/search/result/symbol.go
```diff
@@ -142,6 +143,45 @@ func (s *SymbolMatch) URL() *url.URL {
 	return base
 }
 
+func compareSymbolMatches(a, b *SymbolMatch) int {
+	if v := cmp.Compare(a.Symbol.Line, b.Symbol.Line); v != 0 {
```
jtibshirani, line 147: Curious, why did we switch to cmp.Compare? Naively, the old way of comparing looked more efficient :)

cmd/symbols/internal/api/handler.go
stefanhengl: > Do we have any historical knowledge of why this was added?

🕵️ No. I traced this back to the first commit: "Publish Sourcegraph as open source" from Oct 2018. Back then it was called `maxFirst` and the code comes without comments. 

> We could do something like validate the limit is in the range [0, 5000] and throw an error otherwise?

The behavior in the PR is now consistent with Rockskip which we encourage to enable for repos that are larger than 100MB. I don't recall that we have had issues with OOMs of the symbol service, I may be wrong though. 

If we want to set an upper limit (lower than count:all), maybe frontend would be the better spot? This way we get a consistent behavior for Rocksip and Sqlite.

stefanhengl: EDIT: I changed the PR to enforce the range [0, 50_000]. This is high enough so that we never hit it during normal search (default limit 10001) and it is also high enough to allow Search Jobs to set higher limits than interactive search.

PR #61100, opened 2024-03-13
search: use SexpVerbose for TestNewPlanJob

keegancsmith:
We include this information for zoekt. It helps us test another commit which ensures we use the correct pattern type for searcher.

Test Plan: CI

### DIFF
--- a/internal/search/searcher/search.go
+++ b/internal/search/searcher/search.go
@@ -155,7 +155,6 @@ func (s *TextSearchJob) Attributes(v job.Verbosity) (res []attribute.KeyValue) {
 	case job.VerbosityBasic:
 		res = append(res,
 			attribute.Bool("indexed", s.Indexed),
-			attribute.Stringer("patternInfo", s.PatternInfo),
 		)
 	}
 	return res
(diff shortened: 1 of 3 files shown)


internal/search/searcher/search.go
```diff
@@ -155,6 +155,7 @@ func (s *TextSearchJob) Attributes(v job.Verbosity) (res []attribute.KeyValue) {
 	case job.VerbosityBasic:
 		res = append(res,
 			attribute.Bool("indexed", s.Indexed),
+			attribute.Stringer("patternInfo", s.PatternInfo),
```
jtibshirani, line 158: Instead of adding this, could we instead switch from `SexpPretty` -> `SexpVerbose` inside the job tests? I believe that will use the `VerbosityMax` case above.

Also there's a `fallthrough` above, which means we could add `patternInfo` twice in the `VerbosityMax` case.

keegancsmith: This does make it consistent with how we print attributes for the zoekt job which makes it useful for catching bugs in these golden tests. But you point out a good bug and your change is the path of least resistance. So will update this

jtibshirani (approved):

PR #61101, opened 2024-03-13
Add Cody context filters to the site config and expose them via GraphQL

taras-yemets:
Closes https://github.com/sourcegraph/sourcegraph/issues/61605

Adds Cody context filters to the site config and corresponding GraphQL resolvers.
The `codyContextFilters` query has a required `version` parameter to simplify future API evolution. Returns "cody.contextFilters" site config param value as JSON.

Example query
```
query {
    site {
	codyContextFilters(version: V1) {
            raw
        }
    }
}
```

| "cody.contextFilters" in the site config | Query result |
| -- | -- |
|Screenshot 2024-04-11 at 11 05 57|Screenshot 2024-04-11 at 11 06 05|
|Screenshot 2024-04-11 at 11 09 13|Screenshot 2024-04-11 at 11 09 20|
|not defined|Screenshot 2024-04-11 at 11 11 00|

## Test plan
- CI
- Tested manually


valerybugakov (commented):
API console response before setting up the `cody.contextFilters` value locally. Should we return nulls without panic messages?

Screenshot 2024-04-08 at 12 09 24

arafatkatze (approved):
This PR does what it claims to do. 

There is a longer goal with this PR for context inclusion/exclusion stuff etc so I don't grok the whole picture just yet but for now this PR does what it claims to do so 👍 from me

cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -7246,6 +7246,34 @@ type DefaultSettings implements SettingsSubject {
...
+    """
+    Regular expression which matches a set of repository names.
+    """
+    repoNamePattern: String!
+}
+
+"""
+Rules defining the repositories that will never be shared by Cody with third-party LLM providers.
+"""
+type CodyContextFilters {
```
chwarwick, line 7263: A key thing you want to consider when making a schema is how the error propagation will work, and it can be difficult to change later due to introducing breaking changes.

With `include: [CodyContextFilterItem]` and `exclude: [CodyContextFilterItem]`, meaning include/exclude and each CodyContextFilterItem is optional, this will mean that any error will be "in place" with the rest of the include/exclude still existing.

For example if 1 item within the exclude errors, what should the behavior be? Just continue on and ignore it? 

An alternative could be to make all fields in the `CodyContextFilters` and their CodyContextFilterItems required that would push any error up to the first nullable field being the top level `codyContextFilters` field and provide a better indication that the contextFilters were not able to be determined.

cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -7246,6 +7246,34 @@ type DefaultSettings implements SettingsSubject {
...
+    List of rules specifying repositories that Cody may include as context in requests to third-party LLMs.
+    If defined, only repositories matching these rules will be considered for sharing.
+    If not defined, all repositories may be shared.
+    """
+    include: [CodyContextFilterItem]
+    """
+    List of rules specifying repositories that Cody should excluded from context in requests to third-party LLMs.
+    These rules are applied only to repositories matching the include rules.
+    """
+    exclude: [CodyContextFilterItem]
```
chwarwick, line 7274: How do include and exclude interact? It's not specified in the documentation here.  Can a user provide both an include and exclude and if so what takes precedence?  What happens if I include and exclude the same repo? 

Should this be a single array of items with a type that says if the instance is using an allow or deny list?

chwarwick (commented):
It is worth a little extra time up front to make sure you get the schema right to lock down the behavior you want to enforce.  Since there is some time until the next release you could still make breaking changes but it's nice to avoid those if possible.

Also probably worth noting that those resolvers/settings are experimental and eventually dependent on Client versions >= X to not give the wrong impression.

cmd/frontend/graphqlbackend/schema.graphql
taras-yemets: Include rules define the repositories that can be used by Cody as context in requests to third-party LLMs. If not specified, all repositories are considered as included.
Exclude rules if specified are applied for repositories meeting the include rules. 

Do you think it needs to be further clarified in the fields' descriptions?

taras-yemets: Having a single list may not be flexible enough ([thread](https://docs.google.com/document/d/1PMeJHvY0ia2nMewFvs4vrFcmDbRDw9kSyljil-K8Pt8/edit?disco=AAABKBGa2MQ)).

Also, we are going to support RE2 syntax and it's quite tricky to invert regexes with it. RE2 doesn't support backreferences and look-around assertions and it would be quite tricky to achieve "include everything except for ..." effect ([example](https://docs.google.com/document/d/1PMeJHvY0ia2nMewFvs4vrFcmDbRDw9kSyljil-K8Pt8/edit#bookmark=id.v5fqodttttx2)).

I would be happy to chat about it more!

chwarwick: My point is that as a consumer (the clients) of this, I don't know what I should be doing with this date.  It's going to need to be documented very specifically somewhere what the order of operations is between include/exclude.

Any very small misunderstanding or misinterpretation of the rules could cause large differences in behavior.

taras-yemets: Agree!
I think the [feature docs page](https://sourcegraph.com/docs/cody/capabilities/ignore-context#cody-ignore) would be a good place for these details (as well as [minimum supported client versions](https://github.com/sourcegraph/sourcegraph/pull/61101#pullrequestreview-1986387778)).

cmd/frontend/graphqlbackend/schema.graphql
chwarwick: An example here would be what happens if one of the exclude filters is an invalid regex?  With the current setup it would still return all the other include/exclude rules and an error for that 1 item.  What should clients do? I would tend to think it should be disable Cody for everything because we know that there is an attempt to exclude something but we don't know what.  The currently way of reporting this error makes it difficult for the client to know the expected behavior.  

Also my expectation of disabling may not be the correct expectation but that is also sort of the point that it's left up to consumers to decide.

taras-yemets: I share your concern.
When one field from a config fails to resolve it's an error case for the whole `codyContextFilters` field. Will fix it.
Thank you!

taras-yemets: Made all fields required (https://github.com/sourcegraph/sourcegraph/pull/61101/commits/798b49888f2776b0b4cee404181f947448e9c114).

If `"cody.contextFilters"` field is not defined in the site config, GraphQL query will resolve to empty config:
```
{
  "data": {
    "site": {
      "codyContextFilters": {
        "exclude": [],
        "include": []
      }
    }
  }
}
```

schema/site.schema.json
```diff
@@ -2570,6 +2570,27 @@
       },
       "group": "Cody"
     },
+    "cody.contextFilters": {
+      "description": "Rules defining the repositories that will never be shared by Cody with third-party LLM providers.",
```
keegancsmith, line 2574: Is this language correct? In particular third-party LLM providers. Say we hosted our own LLM, would we then ignore these rules for that? I can imagine that is the case since, but just want some confirmation :)

keegancsmith (approved):
LGTM. I have high level design feedback with respect to the GraphQL. I would argue we shouldn't be modeling the shape of the config in graphql. I think we `codyContextItems` should return a string which is the marshalled JSON of the ignore rules. Two reasons:
- A client should always want all fields. A subset of fields is an invalid. IE the whole config needs to live together.
- Simplifies API / reduces scope of what we need to support w.r.t. API evolution.

However, I'm not a graphQL expert so I don't know if what I am saying is heresy.

schema/site.schema.json
taras-yemets: @aramaraju, what do you think about the wording here?

taras-yemets:
> LGTM. I have high level design feedback with respect to the GraphQL. I would argue we shouldn't be modeling the shape of the config in graphql. I think we `codyContextItems` should return a string which is the marshalled JSON of the ignore rules. Two reasons:
> 
> * A client should always want all fields. A subset of fields is an invalid. IE the whole config needs to live together.
> * Simplifies API / reduces scope of what we need to support w.r.t. API evolution.
> 
> However, I'm not a graphQL expert so I don't know if what I am saying is heresy.

Thanks for your review, @keegancsmith!
I think both exposing the Cody context filters via GraphQL in the same shape as the site config and as a stringified JSON could work for our current use case.

We discussed it with @valerybugakov (thanks for your input, Valery!) and he pointed out that, on one hand, clients will always need an entire filters config, thus they won't benefit from the granularity of the current approach. On the other hand, having types defined on the API level is more reliable than parsing opaque JSON structures on the client to extract the rules. With the opaque JSON structure, it's still possible to parse it incorrectly and use a subset of fields, this approach does not really prevent bugs when the subset of rules is used.

I also browsed the current GraphQL schema looking for similar cases. We return stringified JSON for the entire configurations (e.g., site config, settings, code host configs, etc.), but not for their subsets. It would probably be more consistent with the rest of the codebase to return explicit schema 🤔 

I'm not a GraphQL expert either and I don't know the idiomatic way to handle such problems. I'm curious what other folks think of it.

taras-yemets:
https://github.com/sourcegraph/sourcegraph/pull/61101/commits/3a3a8621d1aa5d44d61ffd4701aa9a247c2badca refactors to return the marshaled JSON of the ignore rules (as string).



Context:
- https://github.com/sourcegraph/sourcegraph/pull/61101#pullrequestreview-1991003745
- https://sourcegraph.slack.com/archives/C02UC4WUX1Q/p1712751174229009

cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -7517,6 +7527,14 @@ type Site implements SettingsSubject {
     Site-admin only.
     """
     codyGatewayRateLimitStatus: [CodyGatewayRateLimitStatus!]
+
+    """
+    Rules defining the repositories that will never be shared by Cody with third-party LLM providers.
+    For details see https://sourcegraph.com/docs/cody/capabilities/ignore-context
+    """
+    codyContextFilters(
+        version: Int!
```
taras-yemets, line 7536: @chwarwick, do you think version should be required? 🤔

cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -7246,6 +7246,16 @@ type DefaultSettings implements SettingsSubject {
...
 }
 
+"""
+Rules defining the repositories that will never be shared by Cody with third-party LLM providers.
+"""
+type CodyContextFilters {
+    """
+    A JSON string representing rules specifying repositories that Cody may include as context in requests to third-party LLMs.
+    """
+    raw: String
```
eseliger, line 7256: I recommend using the custom type `JSONString` here, it'll marshal nicely in the response so you don't have to call json.Parse a second time. 

Example: https://sourcegraph.sourcegraph.com/api/console#%7B%22query%22%3A%22query%20%7B%5Cn%20%20currentUser%20%7B%5Cn%20%20%20%20externalAccounts%20%7B%5Cn%20%20%20%20%20%20nodes%20%7B%5Cn%20%20%20%20%20%20%20%20id%5Cn%20%20%20%20%20%20%20%20accountData%5Cn%20%20%20%20%20%20%7D%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%2C%22variables%22%3A%22%22%7D

taras-yemets: https://github.com/sourcegraph/sourcegraph/pull/61101/commits/08992e7c688c71dd297cb8cb75f777d5489d7e5a

cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -7517,6 +7528,12 @@ type Site implements SettingsSubject {
     Site-admin only.
     """
     codyGatewayRateLimitStatus: [CodyGatewayRateLimitStatus!]
+
+    """
+    Rules defining the repositories that will never be shared by Cody with third-party LLM providers.
+    For details see https://sourcegraph.com/docs/cody/capabilities/ignore-context
+    """
+    codyContextFilters(version: Int!): CodyContextFilters!
```
taras-yemets, line 7536: TODO: use an enum for version (similar to [SearchVersion](https://github.com/sourcegraph/sourcegraph/blob/08992e7c688c71dd297cb8cb75f777d5489d7e5a/cmd/frontend/graphqlbackend/schema.graphql#L2658-L2674))
```suggestion
    codyContextFilters(version: CodyContextFiltersVersion!): CodyContextFilters!
```

PR #61102, opened 2024-03-13
search: set Regexp annotation on patterns for FileContainsPatterns

keegancsmith:
This was the root cause of a bug where searcher and zoekt treated the pattern it received inconsistently. The way file:contains.content() works is it always takes a regex.

Additionally we came across a case for cody context not setting inputs as literal. This likely fixes a bug, but I did not test further. This came up when auditing direct creations of query.Pattern.

Test Plan: added a unit test which ensures searcher uses a regex for the pattern.

Stacked on https://github.com/sourcegraph/sourcegraph/pull/61100

Part of https://github.com/sourcegraph/sourcegraph/issues/60745

### DIFF
--- a/internal/search/codycontext/query_transformer.go
+++ b/internal/search/codycontext/query_transformer.go
@@ -120,7 +120,9 @@ func queryStringToKeywordQuery(queryString string) (*keywordQuery, error) {
 
 	patternNodes := make([]query.Node, 0, len(transformedPatterns))
 	for _, p := range transformedPatterns {
-		patternNodes = append(patternNodes, query.Pattern{Value: p})
+		node := query.Pattern{Value: p}
+		node.Annotation.Labels.Set(query.Literal)
+		patternNodes = append(patternNodes, node)
 	}
 
 	if len(patternNodes) > 0 {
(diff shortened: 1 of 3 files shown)


internal/search/codycontext/query_transformer.go
```diff
@@ -120,7 +120,9 @@ func queryStringToKeywordQuery(queryString string) (*keywordQuery, error) {
 
 	patternNodes := make([]query.Node, 0, len(transformedPatterns))
 	for _, p := range transformedPatterns {
-		patternNodes = append(patternNodes, query.Pattern{Value: p})
+		node := query.Pattern{Value: p}
```
jtibshirani, line 123: Thanks for catching this! I believe we haven't seen many errors around it, because we remove punctuation before creating these patterns (see `term_utils.go`).

However, this is a better approach as I don't think we should be removing punctuation! For example, a search term like `syntax.Regexp` is meaningful. I'll make a note to follow up.

internal/search/job/jobutil/job.go
```diff
@@ -84,7 +84,9 @@ func NewBasicJob(inputs *search.Inputs, b query.Basic) (job.Job, error) {
 	if len(fileContainsPatterns) > 0 {
 		newNodes := make([]query.Node, 0, len(fileContainsPatterns)+1)
 		for _, pat := range fileContainsPatterns {
-			newNodes = append(newNodes, query.Pattern{Value: pat})
+			node := query.Pattern{Value: pat}
+			node.Annotation.Labels.Set(query.Regexp)
```
jtibshirani, line 88: For my education: do we have a notion of "default" for these labels, if you don't specify `Literal` or `Regexp`?

jtibshirani (approved):

PR #61104, opened 2024-03-13
telemetry-gateway: discard oversized events after logging some content

bobheadxi:
I forgot that I never got around to discarding problematic, oversized events. This change makes us properly discard them so they stop retrying, and also logs some metadata - including a snippet of the problematic event for diagnostics.

A [bucket was created](https://sourcegraph.slack.com/archives/C06CCJR4K9R/p1709954993017009?thread_ts=1707984656.886109&cid=C06CCJR4K9R), but I currently don't have time to implement uploading this data to the bucket - for now, let's just do a simple log and discard.

This is currently causing high memory usage in Telemetry Gateway, and a lot of Sentry events.

## Test plan

CI


jac (approved):
🙇‍♂️

PR #61106, opened 2024-03-13
Update tooling for end-to-end Bazel images

willdollman:
There is a handful of tooling around base images builds - switching to Bazel means that this needs updated.

### `sg wolfi` tooling

- [x] `sg wolfi image` builds using Bazel
- Added `--legacy` flag to revert to old behaviour
- [x] `sg wolfi image` uses local custom package repo - this is actually not possible, so will revert to legacy build method
- [x] `sg wolfi update` refreshes all lockfiles

### Buildkite pipeline

- [x] Images are built using Bazel
- [x] If the YAML or Lockfile are modified, add a step that ensures they are in sync
- [x] Uses branch-specific package repo

### Buildkite auto-update pipeline

- [x] Refreshes lockfiles rather than updating hashes

### Documentation

- [x] New build and update process
- [x] How to add new images

## Test plan

- Local testing of sg


willdollman:
Added docs here

dev/ci/internal/ci/wolfi_operations.go
```diff
@@ -298,12 +304,14 @@ func addWolfiOps(c Config) (packageOps, baseImageOps *operations.Set) {
...
+
+		apkoOps = WolfiCheckApkoLocks()
 	}
 
-	return packageOps, baseImageOps
+	return packageOps, baseImageOps, apkoOps
 }
 
-// wolfiRebuildAllBaseImages adds operations to rebuild all Wolfi base images and push to registry
+// wolfiRebuildAllBaseImages [deprecated] adds operations to rebuild all Wolfi base images and push to registry
```
Strum355, line 314: You can probably mark it as deprecated [with this method instead](https://stackoverflow.com/questions/7849663/how-do-you-mark-code-as-deprecated-in-go), depending on the editor setup, this should then be displayed properly

image

dev/ci/scripts/wolfi/build-base-image.sh
```diff
@@ -99,58 +63,48 @@ if [[ "$IS_MAIN" != "true" && "$branch_repo_exists" == "true" ]]; then
...
-  docker push "sourcegraph/wolfi-${name}-base:$tag"
-  docker tag "$local_image_name" "sourcegraph/wolfi-${name}-base:latest"
-  docker push "sourcegraph/wolfi-${name}-base:latest"
-fi
+docker tag "${local_image_name}" "${remote_image_name}:${tag}"
+docker push "${remote_image_name}:${tag}"
+docker tag "${local_image_name}" "${remote_image_name}:latest"
+docker push "${remote_image_name}:latest"
 
+# TODO: Update message to reflect that getting this result in main requires further work
```
Strum355, line 101: :warning: TODO :eyes:

dev/ci/scripts/wolfi/apko-check-lock.sh
```diff
@@ -0,0 +1,50 @@
...
+set -euf -o pipefail
+
+cd "$(dirname "${BASH_SOURCE[0]}")/../../../.."
+REPO_DIR=$(pwd)
+
+BRANCH="${BUILDKITE_BRANCH:-'default-branch'}"
+MAIN_BRANCH="main"
+IS_MAIN=$([ "$BRANCH" = "$MAIN_BRANCH" ] && echo "true" || echo "false")
+
+echo "Is-main is $IS_MAIN"
```
Strum355, line 12: Probably need to do another pass over all the bash scripts to make sure temp stuff is removed

Strum355 (approved):
Stamping so this can be merged once the TODOs are all resolved :slightly_smiling_face:

dev/ci/internal/ci/wolfi_operations.go
willdollman: Nice, thanks

dev/ci/scripts/wolfi/apko-check-lock.sh
willdollman: Thanks, cleaned up

PR #61117, opened 2024-03-13
sg/msp: Cloud Deploy IAM permissions for a user supplied Service Account

jac:
To better support deployments from CI pipelines which already are configured with a GCP Service Account this change allows the email of that service account to be specified as part of the rollout spec.

Access is granted to that service account to then allow it to create releases as part of your CI pipeline.

- [x] Verify the `<pipeline_id>_clouddeploy` bucket is created when the pipeline is created making it available to apply IAM rules to. If the bucket is only create when the pipeline is used for the first time a more broad scope will need to be granted
  - edit: only created on first use so we now manually create it when provisioning the pipeline
```
// Needs access to `<pipeline_id>_clouddeploy` bucket
_ = storagebucketiammember.NewStorageBucketIamMember(stack, id.TerraformID("clouddeploy"), &storagebucketiammember.StorageBucketIamMemberConfig{
	Bucket: pointers.Stringf("%s_clouddeploy", deliveryPipeline.PipelineID),
	Role:   pointers.Ptr("roles/storage.admin"),
	Member: pointers.Stringf("serviceAccount:%s", sa),
})
```

## Test plan

Applied to `msp-testbed-test` environment. Created a release through buildkite by supplying `sourcegraph-aspect-workflow@aspect-dev.iam.gserviceaccount.com` as the service account in the spec


dev/managedservicesplatform/spec/rollout.go
```diff
@@ -11,6 +11,9 @@ type RolloutSpec struct {
 	// Deploy. Note that this does NOT prevent manual deploys from happening
 	// directly in Cloud Run.
 	Suspended *bool `yaml:"suspended,omitempty"`
+	// ServiceAccount is the email address of the service account to provision IAM access to create
+	// releases for. Can be used to give access to the Service Account used in your CI pipeline
+	ServiceAccount *string `yaml:"serviceAccount,omitempty"`
```
bobheadxi, line 16:
```suggestion
	// ServiceAccount is the email address of the service account to provision IAM access to create
	// releases for. Can be used to give access to the Service Account used in your CI pipeline,
	// instead of using the default operator SA.
	ServiceAccount *string `yaml:"serviceAccount,omitempty"`
```

dev/managedservicesplatform/stacks/cloudrun/cloudrun.go
```diff
@@ -434,6 +437,53 @@ func NewStack(stacks *stack.Set, vars Variables) (crossStackOutput *CrossStackOu
...
 			ContentType: pointers.Ptr("application/gzip"),
 		})
+
+		// Provision Service Account IAM to create releases
+		sa := pointers.DerefZero(vars.RolloutPipeline.OriginalSpec.ServiceAccount)
+		if sa != "" {
+			id := id.Group("serviceaccount")
+			// Permission to create releases
+			_ = projectiammember.NewProjectIamMember(stack, id.TerraformID("releaser"), &projectiammember.ProjectIamMemberConfig{
+				Project: pointers.Ptr(vars.ProjectID),
```
bobheadxi, line 447: Presumably the generated SA will use the same roles and permissions as the user-provided SA - could we share the setup in a helper?

Also, should we set this up in the `iam` module instead?

bobheadxi: > That depends, if done as is where access is scoped to individual buckets (<projectid>-cloudrun-skaffold, <pipeline_uid>_clouddeploy) then no as those are defined here

Ah sure, this makes sense - would probably make sense to assign permissions for the other SA option here as well

unknwon (approved):

dev/managedservicesplatform/stacks/cloudrun/cloudrun.go
```diff
@@ -518,3 +541,57 @@ func makeContainerResourceLimits(r spec.EnvironmentInstancesResourcesSpec) map[s
...
 	}
 }
+
+type cloudDeployIAMConfig struct {
+	serviceAccounts    []string
+	skaffoldBucketName *string
+	pipelineBucketName *string
+}
+
+func cloudDeployIAM(vars Variables, id resourceid.ID, stack cdktf.TerraformStack, config cloudDeployIAMConfig) {
```
bobheadxi, line 551:
```suggestion
func addCloudDeployIAM(vars Variables, id resourceid.ID, stack cdktf.TerraformStack, config cloudDeployIAMConfig) {
```

bobheadxi: Maybe a docstring here too, indicating that we need to do this here because the Cloud Deploy pipeline must be provisioned here

bobheadxi (approved):

PR #61119, opened 2024-03-13
auth: do not fail single identity per provider check if it's the same identity

unknwon:
When the user is:

1. Already authenticated
2. Completed an auth flow using an already linked external identity/account

We should not fail the single-identity-per-provider check because that is not a new identity.

This addresses the issue in the following scnerio:

1. Sign in VSCode and dotcom using GitHub
1. Sign out VSCode but stay sign-in on sourcegraph.com
2. Try to sign in VSCode again using the exact same GitHub account
3. White auth error screen on sourcegraph.com.

## Test plan

CI and tested e2e locally

### DIFF
--- a/cmd/frontend/auth/user_test.go
+++ b/cmd/frontend/auth/user_test.go
@@ -327,6 +327,20 @@ func TestGetAndSaveUser(t *testing.T) {
 				expNewUserCreated:                false,
 				expSavedExtAccts:                 map[int32][]extsvc.AccountSpec{},
 			},
+			{
+				description: "single identity per user mode accepts the same external identity from same provider",
+				op: GetAndSaveUserOp{
+					ExternalAccount:       ext("st1", "s1", "c1", "s1/u1"),
+					UserProps:             userProps("u1", "u1@example.com"), // This user exists in the DB already and has an external account for st1, s1, c1
+					SingleIdentityPerUser: true,
+				},
+				createIfNotExistIrrelevant: true,
+				expUserID:                  1,
+				expSavedExtAccts: map[int32][]extsvc.AccountSpec{
+					1: {ext("st1", "s1", "c1", "s1/u1")},
+				},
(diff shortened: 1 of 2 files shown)


cmd/frontend/auth/user_test.go
```diff
@@ -327,6 +327,20 @@ func TestGetAndSaveUser(t *testing.T) {
 				expNewUserCreated:                false,
 				expSavedExtAccts:                 map[int32][]extsvc.AccountSpec{},
 			},
+			{
+				description: "single identity per user mode accepts the same external identity from same provider",
```
eseliger, line 331: Just to triple check, without this patch the test fails, correct?

eseliger (approved):

cmd/frontend/auth/user_test.go
unknwon: Dang, you caught me, was missing `actorUID` field.. but now fails bunch of other things ...

eseliger: that's what I thought - I had to fight this suite pretty hard to make my test work 😆

unknwon:
Ah damn... auto-merged 😂

unknwon:
Let me fix the test in a follow up

cmd/frontend/auth/user_test.go
unknwon: https://github.com/sourcegraph/sourcegraph/pull/61121

chrsmith (commented):
Post-hoc LGTM. We talked about this on a Zoom huddle and this seems like it should do the trick.

Also, I'll state the obvious that it sure would be awesome if we wrote some more unit tests for this function and then carefully refactored the code to be a little more sensible. (Which we don't have time to do now, and once things are in a stable, steady state doesn't seem all that important because things are in a stable, steady state. :P )

PR #61121, opened 2024-03-13
auth: fix ineffective test

unknwon:
Follow up on https://github.com/sourcegraph/sourcegraph/pull/61119#discussion_r1523821235

## Test plan

CI

### DIFF
--- a/cmd/frontend/auth/user_test.go
+++ b/cmd/frontend/auth/user_test.go
@@ -329,6 +329,7 @@ func TestGetAndSaveUser(t *testing.T) {
 			},
 			{
 				description: "single identity per user mode accepts the same external identity from same provider",
+				actorUID:    1,
 				op: GetAndSaveUserOp{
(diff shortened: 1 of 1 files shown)


cmd/frontend/auth/user_test.go
```diff
@@ -339,7 +340,9 @@ func TestGetAndSaveUser(t *testing.T) {
 				expSavedExtAccts: map[int32][]extsvc.AccountSpec{
 					1: {ext("st1", "s1", "c1", "s1/u1")},
 				},
-				expNewUserCreated: false,
+				expCalledGrantPendingPermissions: true,
+				expCalledCreateUserSyncJob:       true,
```
eseliger, line 344: Hmm should this be true?

eseliger (approved):

cmd/frontend/auth/user_test.go
unknwon: Logically I thought it should be false if all done right, but the code logic flows to the end (in this test case), which calls them regardless I think.

PR #61122, opened 2024-03-13
web: Update rxjs to v7

fkling:
A side goal for the web rewrite is to leave the existing code base in a better state than before. I recently [added a hacky workaround](https://github.com/sourcegraph/sourcegraph/blob/da5ddc99b65398c3cb14f075db217f1fb374095a/client/web-sveltekit/vite.config.ts#L82-L101) to make the Svelte version work properly with different rxjs versions. But the whole point of the rewrite is to not have to do these things anymore. So this is my attempt of upgrading rsjx in the main repo to v7.

I worked through the list of breaking changes in the [rxjs documentation](https://rxjs.dev/deprecations/breaking-changes) and fixed TypeScript issues to the best of my abilities.

Most notable changes:

- The custom `combineLatestOrDefault` operator was rewritten to avoid using rxjs internals, and the `.lift` method (side note: the corresponding tests do not cover all expected behavior, but issues had been caught through other tests)
- Where necessary `.toPromise()` was replaced with `lastValueFrom` or `firstValueFrom`. My assumption was that since we don't get runtime errors for the existing code, it's save to assume that the corresponding observables emit at least one value, i.e. `.toPromise()` did not return `undefined`. Only in some places I added a default value where it was easy to deduce what it should be.
- The generic type in `of()` was removed
- The generic type in `concat()` was removed
- `Subject.next` seemed to have allowed `undefined` to be passed even if the subject's types didn't allow that. If the subject's type couldn't be extended to include `undefined` I changed the code to not pass `undefined`.
- The generic type signature of `defaultIfEmpty` changed.
- Where possible I replaced `Subscribable` with `ObservableInput`, but we also have a copy of the old `Subscribable` interface in the `sourcegraph` package, and that makes things more complicated.
- I simplified unnecessary Promise/Observable interop where necessary.

A lot of the complex rxjs logic and usage of changed interfaces, such as `Subscribable`, is in extensions related code, which is not used in the web app anymore, but is still at least imported in the browser extensions code. Most of it is probably not used anymore, which makes the migration somewhat simpler.

I suggest hiding whitespace changes. At the beginning I tried to chunk the work in meaningful commits, but in the end I just fixed issues that the CI reported. Still, it might be easier to look at commits individually.

## Test plan

CI

Did a quick manual test of running Sourcegraph via `sg start web-standalone` and browsed a couple of pages without issue.


client/codeintellify/src/hoverifier.test.ts
```diff
@@ -27,17 +27,18 @@ import {
...
+
+    beforeEach(() => {
+        dom = new DOM()
         testcases = dom.createCodeViews()
     })
 
     let subscriptions = new Subscription()
 
     afterEach(() => {
+        dom.cleanup()
```
fkling, line 41: I got strange `Executing a cancelled action` errors in this test suite, but only if all tests ran. When I ran them individually with `it.only` they passed.
For some mystical reason, changing the setup to instantiate the DOM per test makes the error go away.
I found [this](https://github.com/ReactiveX/rxjs/issues/5671) related to the issue, but that wasn't helpful.

client/shared/src/api/client/api/common.ts
```diff
@@ -59,37 +58,18 @@ export const wrapRemoteObservable = <T>(
...
+            return new Observable(subscriber => {
+                const proxyObserver: Parameters<typeof proxySubscribable['subscribe']>[0] = {
+                    [proxyMarker]: true,
+                    // @ts-expect-error - this was previously typed as any
+                    next: value => subscriber.next(value),
+                    error: error => subscriber.error(asError(error)),
+                    complete: () => subscriber.complete(),
+                }
+                return syncRemoteSubscription(proxySubscribable.subscribe(proxyObserver))
+            })
```
fkling, line 72: I _think_ that's equivalent. `mergeMap` expects to be returned an `Observable` and I didn't see an obvious reason why it couldn't be.

`next: value => subscriber.next(value)` returns a type error but I suspect that would have been an issue before too if `args` hadn't been defined as `any[]`.

client/shared/src/api/client/mainthread-api.test.ts
```diff
@@ -19,7 +19,7 @@ describe('MainThreadAPI', () => {
 
     describe('graphQL', () => {
         test('PlatformContext#requestGraphQL is called with the correct arguments', async () => {
-            const requestGraphQL = sinon.spy(_options => EMPTY)
+            const requestGraphQL = sinon.spy(_options => of({ data: null, errors: [] }))
```
fkling, line 22: This was one example where using `lastValueFrom` elsewhere threw an unexpected error because of the observable not emitting a value.
I think it's OK to assume that `requestGraphQL` would always emit or throw an error.

client/shared/src/api/contract.ts
```diff
@@ -13,7 +12,7 @@ import type { MaybeLoadingResult } from '@sourcegraph/codeintellify'
 import type * as clientType from '@sourcegraph/extension-api-types'
 import type { GraphQLResult } from '@sourcegraph/http-client'
 
-import type { ReferenceContext } from '../codeintel/legacy-extensions/api'
+import type { DocumentHighlight, ReferenceContext } from '../codeintel/legacy-extensions/api'
```
fkling, line 15: These are two different `DocumentHighlights` but it seems we are trying to treat them the same in our APIs. I had to do some juggling to get this to (partially) work.

client/web-sveltekit/vite.config.ts
```diff
@@ -73,33 +73,6 @@ export default defineConfig(({ mode }) => {
...
-                        //   is supposed to work.
-                        // - Using a custom plugin that implements a custom resolveId function is somehow not being
-                        //   run in the production build for `rxjs` imports. Maybe it has something to do with the
-                        //   interop between vite and sveltekit.
-                        if (importer?.includes('@sourcegraph/telemetry')) {
-                            source = source.replace('/index.js', '')
-                        }
-                        return this.resolve(source, importer, options)
-                    },
-                },
```
fkling, line 102: 🥳

vovakulikov (approved):
I own you 10 euros. I've checked browser extension with this PR and it worked as it worked on the main (note if you're going to check it yourself, the most recent version of gitlab doesn't support browser extension because of project name error, our own instance where we still use old version for testing this is where I run browser extension with this PR https://gitlab.sgdev.org/)

client/codeintellify/src/hoverifier.test.ts
fkling: I ran them in the console. The change I made here fixes the issue, but I don't know why.

PR #61131, opened 2024-03-14
cleanup: Remove 'Dump' terminology from backend

varungandhi-src:
Reviewer request:
- Christoph - please check if my understanding of the commitgraph bits is correct
- Anton - please check that my general understanding of state of `lsif_uploads`/`lsif_dumps` is correct
- Other parts - Anyone is free to review

Background information:

- `lsif_dumps` is a view which subsets out `lsif_uploads` ([source](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@e7484252eb449c27e1fb2ca22ab7fa8bd757a818/-/blob/migrations/frontend/squashed.sql?L3102-3126)). It checks that the state is either 'completed' or 'deleting'.
  - `lsif_dumps_with_repository_name` is a view which further subsets out `lsif_dumps` and checks that `deleted_at IS NULL` ([source](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@e7484252eb449c27e1fb2ca22ab7fa8bd757a818/-/blob/migrations/frontend/squashed.sql?L3137-3163)).
    - [ ] Is it guaranteed that the state is `deleting` then `deleted_at` is non null? It seems reasonable, but I need to double-check. Or can we return uploads which are currently mid-deletion (which seems non-ideal)? I briefly looked at [`SET state = 'deleting'`](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/sourcegraph%24%40ec02bad+SET+state+%3D+%27deleting%27&patternType=standard&sm=0) and [`SET state = 'deleted'`](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/sourcegraph%24%40ec02bad+SET+state+%3D+%27deleted%27&patternType=standard&sm=0), and in both places, we don't seem to be setting the `deleted_at` column. 🤔
    - There are 3 usages of this view. In all 3 cases, it seems like we really only want `state = 'completed'` uploads.
      - [`getDumpsByIDsQuery`](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@e7484252eb449c27e1fb2ca22ab7fa8bd757a818/-/blob/internal/codeintel/uploads/internal/store/uploads.go?L247-267)
      - [`definitionDumpsQuery`](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@e7484252eb449c27e1fb2ca22ab7fa8bd757a818/-/blob/internal/codeintel/uploads/internal/store/uploads.go?L579-625)
      - [`findClosestDumpsFromGraphFragmentQuery`](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@e7484252eb449c27e1fb2ca22ab7fa8bd757a818/-/blob/internal/codeintel/uploads/internal/store/commitgraph.go?L440-461)

For now, I'm just removing the 'Dumps' terminology because it is not self-explanatory, and is sometimes used interchangeably with an entirely different term 'Uploads', making reading the code confusing. If there is agreement on the expected behavior, I can file follow-up issues so that we rename the views, and change the logic to check `state = 'completed'` rather than `deleted_at is NULL`.

Side-notes:

- `lsif_dumps` is directly used in [referenceIDsQuery](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@ec02bad50e68a07354dddb189c8c7f56bd047225/-/blob/internal/codeintel/uploads/internal/store/uploads.go?L478-516) which makes it seems like we can potentially return references inside `state = 'deleting'` uploads.
  - This seems wrong given that [this usage site of `referenceIDsQuery`](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@ec02bad50e68a07354dddb189c8c7f56bd047225/-/blob/internal/codeintel/uploads/internal/store/uploads.go?L413-463) where we attempt to locate uploads which refer to package names from any of the arguments.

## Test plan

Covered by existing tests

### DIFF
--- a/internal/codeintel/codenav/internal/lsifstore/locations_by_position.go
+++ b/internal/codeintel/codenav/internal/lsifstore/locations_by_position.go
@@ -76,7 +76,7 @@ func (s *store) GetBulkMonikerLocations(ctx context.Context, tableName string, u
 		totalCount += len(monikerLocations.Locations)
 	}
 	trace.AddEvent("TODO Domain Owner",
-		attribute.Int("numDumps", len(locationData)),
+		attribute.Int("numUploads", len(locationData)),
 		attribute.Int("totalCount", totalCount))
 
 	max := totalCount
@@ -94,9 +94,9 @@ outer:
 			}
 
 			locations = append(locations, shared.Location{
-				DumpID: monikerLocations.DumpID,
-				Path:   row.URI,
-				Range:  newRange(row.StartLine, row.StartCharacter, row.EndLine, row.EndCharacter),
+				UploadID: monikerLocations.UploadID,
+				Path:     row.URI,
+				Range:    newRange(row.StartLine, row.StartCharacter, row.EndLine, row.EndCharacter),
 			})
 
 			if len(locations) >= limit {
@@ -176,9 +176,9 @@ func (s *store) getLocations(
 			for _, monikerLocation := range monikerLocations {
 				for _, row := range monikerLocation.Locations {
 					locations = append(locations, shared.Location{
-						DumpID: monikerLocation.DumpID,
-						Path:   row.URI,
-						Range:  newRange(row.StartLine, row.StartCharacter, row.EndLine, row.EndCharacter),
+						UploadID: monikerLocation.UploadID,
+						Path:     row.URI,
+						Range:    newRange(row.StartLine, row.StartCharacter, row.EndLine, row.EndCharacter),
 					})
 				}
 			}
@@ -492,7 +492,7 @@ func deduplicateLocations(locations []shared.Location) []shared.Location {
 
 func locationKey(l shared.Location) string {
 	return fmt.Sprintf("%d:%s:%d:%d:%d:%d",
-		l.DumpID,
+		l.UploadID,
 		l.Path,
 		l.Range.Start.Line,
 		l.Range.Start.Character,
@@ -555,7 +555,7 @@ func (s *store) GetMinimalBulkMonikerLocations(ctx context.Context, tableName st
 		totalCount += len(monikerLocations.Locations)
 	}
 	trace.AddEvent("TODO Domain Owner",
-		attribute.Int("numDumps", len(locationData)),
+		attribute.Int("numUploads", len(locationData)),
 		attribute.Int("totalCount", totalCount))
 
 	max := totalCount
@@ -573,9 +573,9 @@ outer:
 			}
 
 			locations = append(locations, shared.Location{
-				DumpID: monikerLocations.DumpID,
-				Path:   row.URI,
-				Range:  newRange(row.StartLine, row.StartCharacter, row.EndLine, row.EndCharacter),
+				UploadID: monikerLocations.UploadID,
+				Path:     row.URI,
+				Range:    newRange(row.StartLine, row.StartCharacter, row.EndLine, row.EndCharacter),
 			})
 
 			if len(locations) >= limit {
--- a/internal/codeintel/codenav/observability.go
+++ b/internal/codeintel/codenav/observability.go
@@ -12,17 +12,17 @@ import (
 )
 
 type operations struct {
-	getReferences          *observation.Operation
-	getImplementations     *observation.Operation
-	getPrototypes          *observation.Operation
-	getDiagnostics         *observation.Operation
-	getHover               *observation.Operation
-	getDefinitions         *observation.Operation
-	getRanges              *observation.Operation
-	getStencil             *observation.Operation
-	getClosestDumpsForBlob *observation.Operation
-	snapshotForDocument    *observation.Operation
-	visibleUploadsForPath  *observation.Operation
+	getReferences                     *observation.Operation
+	getImplementations                *observation.Operation
+	getPrototypes                     *observation.Operation
+	getDiagnostics                    *observation.Operation
+	getHover                          *observation.Operation
+	getDefinitions                    *observation.Operation
+	getRanges                         *observation.Operation
+	getStencil                        *observation.Operation
+	getClosestProcessedUploadsForBlob *observation.Operation
+	snapshotForDocument               *observation.Operation
+	visibleUploadsForPath             *observation.Operation
 }
 
 var m = new(metrics.SingletonREDMetrics)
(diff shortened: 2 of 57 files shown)


internal/codeintel/codenav/types.go
```diff
@@ -179,7 +179,7 @@ type ImplementationsCursor struct {
 
 // cursorAdjustedUpload
 type CursorToVisibleUpload struct {
-	DumpID                int             `json:"dumpID"`
+	UploadID              int             `json:"uploadID"`
```
varungandhi-src, line 182: I think this is OK because cursors are opaque, and are only associated with a single request. So when a Sourcegraph version upgrade is applied, live Find references actions with pagination will break if you do 'Show more references'. I think that's an OK tradeoff because:
- Pagination is generally not that common
- Even if you have multiple pages, you don't always click 'Load more'
- We don't have any documented requirements around zero-downtime upgrades.

internal/codeintel/uploads/shared/types.go
```diff
@@ -55,10 +55,12 @@ func (u Upload) SizeStats() UploadSizeStats {
...
 
-// TODO - unify with Upload
-// Dump is a subset of the lsif_uploads table (queried via the lsif_dumps_with_repository_name view)
+// ProcessedUpload is a subset of the lsif_uploads table
+// (queried via the lsif_dumps_with_repository_name view)
 // and stores only processed records.
-type Dump struct {
+//
+// The State must be 'completed', see TODO(id: completed-state-check).
+type ProcessedUpload struct {
```
varungandhi-src, line 63: This is the main change, and most of the other changes are for consistency.

varungandhi-src: On second thought, perhaps this should be called `CompletedUpload` to match up with `state = 'completed'`?

internal/codeintel/codenav/transport/graphql/root_resolver.go
```diff
@@ -195,7 +195,7 @@ func (r *gitBlobLSIFDataResolver) VisibleIndexes(ctx context.Context) (_ *[]reso
 	return &resolvers, nil
 }
 
-func dumpToUpload(expected uploadsshared.Dump) *uploadsshared.Upload {
```
kritzcreek, line 198: Can we change this name too?

internal/codeintel/codenav/transport/graphql/root_resolver_diagnostics.go
```diff
@@ -79,7 +79,7 @@ func (r *diagnosticResolver) Location(ctx context.Context) (resolverstubs.Locati
 		ctx,
 		r.locationResolver,
 		shared.UploadLocation{
-			Dump:         r.diagnostic.Dump,
+			Dump:         r.diagnostic.Upload,
```
kritzcreek, line 82: Can we change this field name or would that be breaking for the GraphQL API?

kritzcreek (approved):
Changes to commitgraph LGTM. Thanks for the cleanup, one less confusing term for the same thing!

internal/codeintel/uploads/shared/types.go
kritzcreek: I agree `Completed` is a slight bit clearer, but I'm not sure if it justifies you having to change all this again :D

varungandhi-src: Changed in https://github.com/sourcegraph/sourcegraph/pull/61131/commits/acfda10a7a3513a82b927157c6ccf48eb2c836b1

internal/codeintel/codenav/transport/graphql/root_resolver_diagnostics.go
varungandhi-src: This field isn't being serialized (there is no associated struct tag, and I don't see any 'dump' fields in our GraphQL API), so it should be OK to rename this, I just missed it.

internal/codeintel/codenav/transport/graphql/root_resolver.go
varungandhi-src: Removed this function (there were 2 copies of it), and made it a method on `CompletedUpload`.

internal/codeintel/codenav/internal/lsifstore/locations_by_position.go
```diff
@@ -76,7 +76,7 @@ func (s *store) GetBulkMonikerLocations(ctx context.Context, tableName string, u
 		totalCount += len(monikerLocations.Locations)
 	}
 	trace.AddEvent("TODO Domain Owner",
-		attribute.Int("numDumps", len(locationData)),
+		attribute.Int("numUploads", len(locationData)),
```
keynmol, line 79: I guess this will mean that historic data will be disjoint from the point when this PR makes it onto the instances?

internal/codeintel/codenav/observability.go
```diff
@@ -46,17 +46,17 @@ func newOperations(observationCtx *observation.Context) *operations {
...
-		visibleUploadsForPath:  op("VisibleUploadsForPath"),
+		getReferences:                     op("getReferences"),
+		getImplementations:                op("getImplementations"),
+		getPrototypes:                     op("getPrototypes"),
+		getDiagnostics:                    op("getDiagnostics"),
+		getHover:                          op("getHover"),
+		getDefinitions:                    op("getDefinitions"),
+		getRanges:                         op("getRanges"),
+		getStencil:                        op("getStencil"),
+		getClosestCompletedUploadsForBlob: op("GetClosestCompletedUploadsForBlob"),
```
keynmol, line 57: same comment about metrics as https://github.com/sourcegraph/sourcegraph/pull/61131/files#r1525023924

keynmol (approved):
Nothing jumps out at me apart from the impact on particular metrics, but I don't think it's a big deal IMO.

internal/codeintel/codenav/internal/lsifstore/locations_by_position.go
varungandhi-src: It's an attribute on traces here. So I don't think there is a time series as such, but yes, the field name will be different. I think it should be fine...? After all, we sometimes introduce new fields in traces too.

internal/codeintel/codenav/observability.go
varungandhi-src: With metrics, it's a bit iffy since those do cut across time, but I think it makes sense to just get this done with, I don't think anyone is particularly looking at these.

internal/codeintel/codenav/types.go
varungandhi-src: Because the `uploadID` field will be missing from the request, it will be automatically set to 0 because of how Go's marshalling works. So we might start showing older results, I'm not 100% sure as I'd have to trace through the code to see if there might be an inconsistency which would cause an error or if we would actually just continue returning results but old ones. I don't think it's worth doing that right now.

PR #61134, opened 2024-03-14
Adding Token counting support for bring your own key customers using Redis cache

arafatkatze:
# PR description(Claude Generated)

## Motivation
This PR aims to introduce token counting functionality for customers who use their own API keys (BYOK - Bring Your Own Key) with various language models. Token counting is crucial for accurately tracking usage and costs associated with different models.

## Approach

[See Video](https://www.loom.com/share/73f6b9a50b1b474a83bc9d97411dcb50?sid=f674ac85-83af-4b43-bb84-f57ce6600105)

To efficiently handle potentially high volumes of token counting requests, especially for features like autocomplete, we leverage Redis as a caching mechanism. The token usage data for different models is stored in the Redis cache, allowing for quick access and updates.

## Redis Caching
1. Token usage data is stored in the Redis cache with a TTL (Time to Live) of 30 minutes.
2. This approach prevents the need to persist every single request to the database, which could be inefficient and resource-intensive.
3. After 30 minutes, the token usage data is pulled from the Redis cache and sent to a ping mechanism.
The data team can then process and analyze the token usage data from that point onwards.

## Tokenizer Logic
- To improve code organization and reusability, the tokenizer logic has been moved to a separate folder.
- This allows the tokenizer to be used by multiple components and facilitates easier maintenance and updates.

## Model Support

The token counting functionality has been extended to support various models and providers, including:
- Bedrock
- Azure OpenAI
- OpenAI
- Anthropic

The PR introduces the ability to build and integrate new models with the tokenizer, enabling flexibility for future additions.

## Testing
The token counting functionality has been thoroughly tested with the supported models and providers.
Personal API keys and setups were used to ensure the correctness and reliability of the implementation.
Edge cases and different configurations were considered and verified.

## Collaboration
If you have any of the supported setups working with your own API key, you are encouraged to try out the token counting feature and provide feedback. For testing specific edge cases or configurations (e.g., Bedrock), feel free to reach out to me via DM, and I'll be happy to assist you with the necessary credentials or guidance.

## Future Enhancements
Explore the possibility of adding support for more models and providers based on user requirements and feedback(Chris/stephen can chime in here)

## Test plan

- [x] Implement Redis caching for token usage data
- [x]   Move tokenizer logic to a separate folder for reusability
- [x]  Add support for Bedrock, Azure OpenAI, OpenAI, and Entropiq models
- [x]  Test token counting functionality with personal API keys and setups
- [x]  Verify edge cases and different configurations
- [ ]  Update documentation to reflect the new token counting feature(Will do this after Data team approves this and also I add stuff on the end of [Cody analytics repo](https://github.com/sourcegraph/cody-analytics) )

Please review the PR and provide any feedback or suggestions for improvement. If you have any questions or need further clarification, don't hesitate to reach out. Thank you for your time and consideration!


internal/completions/client/azureopenai/openai.go
```diff
@@ -263,7 +266,9 @@ func streamChat(
 		entry, err := resp.ChatCompletionsStream.Read()
 		// stream is done
 		if errors.Is(err, io.EOF) {
-			return nil
+			tokenManager := tokenusage.NewTokenUsageManager()
```
arafatkatze, line 269: NOTE: I realize that some outputs and some place will show me the tokenusage in the responses too so maybe I can use those in some places. It still shouldn't deter from the essence of this PR. 

Also it seems like the. current implementation is generic enough to be used as it is and we can probably make use of the response token numbers as an efficiency measure later.

internal/completions/tokenusage/tokenusage.go
```diff
@@ -0,0 +1,72 @@
...
+func (m *TokenUsageManager) TokenizeAndCalculateUsage(inputText, outputText, model, feature string, stream bool) error {
+	tokenizer, err := tokenizer.NewTokenizer(model)
+	if err != nil {
+		return err
+	}
+
+	inputTokens, _ := tokenizer.Tokenize(inputText)
+	outputTokens, _ := tokenizer.Tokenize(outputText)
+
+	baseKey := fmt.Sprintf("%s:%s:%s:", model, feature, streamDescription(stream))
```
dadlerj, line 30: Just to confirm—this set of possible strings here will be consistent across all instances, right? I think so looking at how they're generated above. Our schema needs to be explicitly structure, and won't adapt easily to arbitrary strings.

As part of this PR, we would need to go add these possible keys to our BigQuery ping schema so that we get the data in a usable format. Is it possible to create a list of keys now?

Another option, if we expect the list of possible values to grow quickly, would be to change the format of the usageData field in the GetAllTokenUsageData func below to be an array of `{"description": ${baseKey}, "tokens": ${tokenCount}}`?

arafatkatze: > this set of possible strings here will be consistent across all instances, right?

Yes in the sense that we are not doing anything in this string that will make it different on the basis of an instance but different instances shall focus on different models so for some all the metrics are for the anthropic model and for some its openai model. 

Some will have slightly different names for the exact same model because of using bedrock etc the model naming convention could be changed. So all in all there is a huge overlap but there could still be a lot of different cases. 

> Another option, if we expect the list of possible values to grow quickly, would be to change the format of the usageData field in the GetAllTokenUsageData func below to be an array of {"description": ${baseKey}, "tokens": ${tokenCount}}?

That I think makes even more sense because I do think that this combination of providers and different models could potentially grow to 15-20 and then more with time so if it helps your team with data processing to do it the way you suggested then I would be down to make that change. Just give me a thumbs up here.

internal/completions/tokenusage/tokenusage.go
```diff
@@ -0,0 +1,72 @@
...
+	return "non-stream"
+}
+
+func (m *TokenUsageManager) updateTokenCounts(key string, tokenCount int) {
+	currentTokens, _ := m.Cache.GetInt(key)
+	newTokens := currentTokens + tokenCount
+	m.Cache.SetInt(key, newTokens)
+}
+
+func (m *TokenUsageManager) GetTokenCounts(key string) (int, bool) {
```
chwarwick, line 52: If these are ever increasing values are you sure that an int is sufficent to hold the number of tokens? Considering we look at pricing per million tokens I suspect an int is not large enough.

chwarwick (changes requested):
This is a good start, but right now it's not clear to me what the tokens counts in the ping are inteted to be all time use vs this period use. 

Additionaly other pings aggregate data from the local db's event logs which is nice because it leaves a record of this locally that can be queryable.  Considering the most common use here is customers in BYOK mode likely self hosted having this aggregated data in the local event log may be neccessary so customer can extract it.

internal/completions/tokenusage/tokenusage.go
dadlerj: That change in ping data structure would be great, thank you!

slimsag:
General question: if we lose all redis cache data, what is the impact on whatever functionality this provides?

All data in Redis depoyments is generally considered ephemeral to us. i.e., customers are (an have historically been) advised that if they do not back up Redis data and all of it is lost in e.g. an accident, then the only 'real' impact on users will be that they need to sign in again - and in some cases things may be a little bit slower until the caches re-populate.

Given that, any concerns with using Redis for this functionality?

internal/rcache/rcache.go
```diff
@@ -88,6 +89,28 @@ func (r *Cache) SetWithTTL(key string, b []byte, ttl int) {
...
+func (r *Cache) GetInt(key string) (int, bool) {
+	b, found := r.Get(key)
+	if !found {
+		return 0, false
+	}
+	// Convert byte slice to int
+	value, err := strconv.Atoi(string(b))
+	if err != nil {
+		log15.Warn("failed to convert value to int", "value", string(b), "error", err)
+		return 0, false
```
slimsag, line 109: This seems bad for other uses, if for some reason the value here is wrong (e.g. key conflict with other usage in our codebase or something) then we would start getting zero values back here and not really know what is going on. The log warning also doesn't mention the `key` so we wouldn't know what it is related to.

I would prefer to see this function return an error instead, for the caller to log appropriately, and replace the `bool` return parameter with a 'does not exist' error instead.

internal/completions/tokenusage/tokenusage.go
slimsag: seconded

internal/completions/tokenusage/tokenusage.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/sourcegraph/sourcegraph/internal/completions/tokenizer"
+	"github.com/sourcegraph/sourcegraph/internal/rcache"
+)
+
+type TokenUsageManager struct {
+	Cache *rcache.Cache
+}
+
+func NewTokenUsageManager() *TokenUsageManager {
```
slimsag, line 15: In a lot of places in this PR (like, per-request and stuff) you are calling `NewTokenUsageManager` - which is calling `rcache.NewWithTTL`. I think this is not good for performance or GC.

Instead, see if you can pass an instance of `*TokenUsageManager` around places and not call `NewTokenUsageManager` so much.

internal/completions/tokenusage/tokenusage.go
arafatkatze: Done

internal/completions/tokenusage/tokenusage.go
arafatkatze: Great Catch 🙇

internal/completions/tokenusage/tokenusage.go
```diff
@@ -0,0 +1,72 @@
...
+	baseKey := fmt.Sprintf("%s:%s:%s:", model, feature, streamDescription(stream))
+
+	// Calculate and update token counts in Redis
+	m.updateTokenCounts(baseKey+"input", len(inputTokens))
+	m.updateTokenCounts(baseKey+"output", len(outputTokens))
+	return nil
+}
+
+// Helper function to get the description of the request type
+func streamDescription(stream bool) string {
```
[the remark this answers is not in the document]
arafatkatze: Cool I just wanted to be comprehensive but you are right I don't see how this adds value to the tokenization. 

I do think that whether its chat or autocomplete does make a difference though

internal/completions/tokenusage/tokenusage.go
```diff
@@ -0,0 +1,72 @@
...
+	currentTokens, _ := m.Cache.GetInt(key)
+	newTokens := currentTokens + tokenCount
+	m.Cache.SetInt(key, newTokens)
+}
+
+func (m *TokenUsageManager) GetTokenCounts(key string) (int, bool) {
+	return m.Cache.GetInt(key)
+}
+
+func (m *TokenUsageManager) GetAllTokenUsageData() map[string]int {
```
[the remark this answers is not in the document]
arafatkatze: @dadlerj This looks pretty cool, what do you think?(The only suggestion being that input/output tokens will be in different keys)


The current format  per your suggestion is 
```

{
descriptions: ["model1:chat_feature:input", "model2:chat_feature:output", "model3:completion_feature:input" ],
tokens:           [4000,                                          500,                        600]

}

```

Is the format good enough for you dan or would you suggest something different like this recommendation?

internal/completions/tokenusage/tokenusage.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/sourcegraph/sourcegraph/internal/rcache"
+)
+
+type TokenUsageManager struct {
+	Cache *rcache.Cache
+}
+
+func NewTokenUsageManager() *TokenUsageManager {
+	return &TokenUsageManager{
+		Cache: rcache.NewWithTTL("LLMUsage", 1800),
```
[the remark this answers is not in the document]
arafatkatze: Great catch.

Theres two ways that I think we can deal with this
1. using Sorted sets of reddis where each entry is separate and then we delete keys older than 30 mins and then do cumulative operations 
Cons: 
- Hard to implement
- Probably risky if done imperfectly

3. When doing the pings every 30 minutes we can just :delete the key on it being successfully put into the struct [here](https://github.com/sourcegraph/sourcegraph/pull/61134#discussion_r1528301428)

Cons: There might be a very slight offset of a few operations because of how redis handles concurrent calls.  If a key is updated in Redis right after it's read but before it's deleted, those updates could be lost. 
The way to handle this con is to have a shared mutex lock
```

func (m *TokenUsageManager) TokenizeAndCalculateUsage(inputText, outputText, model, feature string) error {
	m.mu.Lock() // Lock the mutex at the start
	defer m.mu.Unlock() // Ensure the mutex is unlocked at the end

	tokenizer, err := tokenizer.NewTokenizer(model)
	if err != nil {
		return err
	}

	inputTokens, _ := tokenizer.Tokenize(inputText)
	outputTokens, _ := tokenizer.Tokenize(outputText)

	baseKey := fmt.Sprintf("%s:%s:", model, feature)

	m.updateTokenCounts(baseKey+"input", int64(len(inputTokens)))
	m.updateTokenCounts(baseKey+"output", int64(len(outputTokens)))
	return nil
}

func (m *TokenUsageManager) GetAllTokenUsageData() TokenUsageSummary {
	m.mu.Lock() // Lock the mutex at the start
	defer m.mu.Unlock() // Ensure the mutex is unlocked at the end

	allKeys := m.Cache.ListAllKeys()
	var summary TokenUsageSummary

	for _, key := range allKeys {
		cleanedKey := strings.SplitN(key, "LLMUsage:", 2)[1]
		value, found := m.Cache.GetInt(cleanedKey)
		if !found {
			continue
		}
		summary.Descriptions = append(summary.Descriptions, cleanedKey)
		summary.Tokens = append(summary.Tokens, value)

               // Delete the key after processing
            err := m.Cache.Delete(cleanedKey)
            if err != nil {
               log15.Warn("Failed to delete key", "key", cleanedKey, "error", err)
            }
	}


	return summary
}
```

I think the second choice works and if you give me a thumbs up then I will add it

internal/completions/tokenusage/tokenusage.go
```diff
@@ -0,0 +1,72 @@
...
+// Helper function to get the description of the request type
+func streamDescription(stream bool) string {
+	if stream {
+		return "stream"
+	}
+	return "non-stream"
+}
+
+func (m *TokenUsageManager) updateTokenCounts(key string, tokenCount int) {
+	currentTokens, _ := m.Cache.GetInt(key)
```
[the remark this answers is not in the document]
arafatkatze: Great catch. 

I definitely should have thought about this. 

Thanks for the catch.

arafatkatze:
@slimsag you asked a great question

> All data in Redis depoyments is generally considered ephemeral to us. i.e., customers are (an have historically been) advised that if they do not back up Redis data and all of it is lost in e.g. an accident, then the only 'real' impact on users will be that they need to sign in again - and in some cases things may be a little bit slower until the caches re-populate.

I wanna go with this as the answer. 

image

internal/rcache/rcache.go
arafatkatze: Good recommendation: This reminds me that this PR could have many different points of failure so I fixed the error handling all over the place.

internal/completions/tokenusage/tokenusage.go
arafatkatze: Makes sense I did that

internal/completions/tokenusage/tokenusage.go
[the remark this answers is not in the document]
arafatkatze: Thanks I will make a change for this

cmd/cody-gateway/internal/httpapi/completions/anthropic.go
```diff
@@ -130,7 +130,7 @@ type anthropicResponse struct {
 }
 
 type AnthropicHandlerMethods struct {
-	anthropicTokenizer *tokenizer.Tokenizer
+	anthropicTokenizer *tokenizer.TiktokenTokenizer
```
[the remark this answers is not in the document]
arafatkatze: Yeah makes sense

internal/completions/tokenizer/tokenizer.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	_ "embed"
+	"strings"
+
+	"github.com/pkoukk/tiktoken-go"
+)
+
+// Tokenizer is an interface that all tokenizer implementations must satisfy.
+type Tokenizer interface {
```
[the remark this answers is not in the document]
arafatkatze: Yeah I actually used that. The idea is to use this generically as we grow to add more models and then other stuff will be implementation of this.

internal/completions/client/anthropic/anthropic.go
```diff
@@ -84,12 +92,26 @@ func (a *anthropicClient) Stream(
...
 			return err
 		}
 	}
+	if dec.Err() != nil {
+		return dec.Err()
+	}
+
+	tokenManager := tokenusage.NewTokenUsageManager()
+	err = tokenManager.TokenizeAndCalculateUsage(inputText(requestParams.Messages), completedString, "anthropic/"+requestParams.Model, string(feature), true)
+	return err
```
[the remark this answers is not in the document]
arafatkatze: Makes sense and I made the changes.

Just for further clarification: 
- Can we say in general that telemetry failures in general should be NON blocking on the main operation and therefore we shouldn't error out because of them?  Would that be the reason to not error out here?

internal/completions/tokenizer/tokenizer.go
```diff
@@ -0,0 +1,46 @@
...
+	Tokenize(text string) ([]int, error)
+}
+
+type TiktokenTokenizer struct {
```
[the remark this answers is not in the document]
arafatkatze: Makes sense, I fixed this.

internal/completions/tokenusage/tokenusage.go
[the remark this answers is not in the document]
arafatkatze: @dadlerj @chwarwick I have changed the format. If you are cool with this this please resolve this issue.

cmd/cody-gateway/internal/httpapi/completions/anthropic.go
```diff
@@ -45,7 +45,7 @@ func NewAnthropicHandler(
 	autoFlushStreamingResponses bool,
 ) (http.Handler, error) {
 	// Tokenizer only needs to be initialized once, and can be shared globally.
-	anthropicTokenizer, err := tokenizer.NewAnthropicClaudeTokenizer()
+	anthropicTokenizer, err := tokenizer.NewAnthropicClaudeTokenizer("anthropic/claude-2")
```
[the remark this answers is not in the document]
arafatkatze: > Does that model name ever get used because it will likely be wrong in many cases.

So I finally refactored it to look [like this](https://sourcegraph.com/-/editor?remote_url=https%3A%2F%2Fgithub.com%2Fsourcegraph%2Fsourcegraph.git&branch=count-token&file=cmd%2Fcody-gateway%2Finternal%2Fhttpapi%2Fcompletions%2Fanthropic.go&editor=VSCode&version=2.2.16&start_row=47&start_col=57&end_row=47&end_col=57) essentially the idea here is that we initialize a tokenizer which is the only thing exported from that package and give it a model name and then it "magically" figures out the right [tokenizer to use](https://sourcegraph.com/-/editor?remote_url=https%3A%2F%2Fgithub.com%2Fsourcegraph%2Fsourcegraph.git&branch=count-token&file=internal%2Fcompletions%2Ftokenizer%2Ftokenizer.go&editor=VSCode&version=2.2.16&start_row=26&start_col=8&end_row=36&end_col=1) based on whether its openai/azure/claude or something else. I didn't yet add the granularity to be able to specifically double down on claude-2 so the model name is actually not used.

@philipp-spiess says that for claude 3 onwards [the request itself](https://sourcegraph.slack.com/archives/C05497E9MDW/p1710514378197969?thread_ts=1710514215.874919&cid=C05497E9MDW) would give use the token count so perhaps I can learn how to add that somehow in the requests we make to anthropic and force this to only work with claude 2.x /claude instant models.

image

  
Anyway chris, I am open to ideas here: I am also happy to make followup changes to use the value of claude3+ tokens in a separate PR too. What do you think?

internal/completions/tokenusage/tokenusage.go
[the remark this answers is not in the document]
arafatkatze: Done

cmd/cody-gateway/internal/httpapi/completions/anthropic.go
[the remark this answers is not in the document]
philipp-spiess: @arafatkatze There's too much context in this PR and I haven't caught up on all of this but regarding your ping: The `/messages` API that we're moving to give you exact token counts (this is not related to Claude 3, so even Claude 2 will get us usage data on that API).

Ideally we would always use usage data as reported by the inference APIs if these exist because models change and new models come with new tokenizers. I don't have context as to why these changes are important for us right now but if we can avoid having to add tokenizers for every new model being added, I imagine it would help a ton going forward.

internal/completions/client/anthropic/anthropic.go
```diff
@@ -84,12 +97,28 @@ func (a *anthropicClient) Stream(
...
+	return nil
+}
+
+func inputText(messages []types.Message) string {
+	allText := ""
+	for _, message := range messages {
+		allText += message.Text
+	}
+	return allText
 }
```
philipp-spiess, line 122: This seems wrong, wouldn't there also be tokens necessary to explain to the LLM what the boundaries are between user/system/and assistant messages?

arafatkatze: Yeah I though I had fixed it but I hadn't. I will fix that soon.

cmd/cody-gateway/internal/httpapi/completions/anthropic.go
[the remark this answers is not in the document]
arafatkatze: Great idea Philip. If this is the case then Ideally I should get rid of tokenizer for anthropic because it serves no purpose if the `/messages` api can take care of the usage counting for all the models of anthropic now(as show on docs and tested by us). Basically for any model related to anthropic we should first fully migrate to the `/messages` api  so first we do 
1. https://github.com/sourcegraph/sourcegraph/pull/60953 
2. I add support for `/messages` api for amazon bedrock because they too support [token output now](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-messages.html#model-parameters-anthropic-claude-messages-request-response) and that does have the output tokens that we are looking for
3.  Once both those PRs are merged I rebase this PR on top of them and then get rid of the anthropic tokenizer altogether because getting rid of it while adding. support for claude 3 is easier than doing shenanigans to support claude 3 token counting. 

@chwarwick can you give me a thumbs up for this and then I will implement it?

cmd/cody-gateway/internal/httpapi/completions/anthropicmessages.go
```diff
@@ -280,7 +280,7 @@ func (a *AnthropicMessagesHandlerMethods) parseResponseAndUsage(logger log.Logge
 	return promptUsage, completionUsage
 }
 
-func isFlaggedAnthropicMessagesRequest(tk *tokenizer.Tokenizer, r anthropicMessagesRequest, cfg config.AnthropicConfig) (*flaggingResult, error) {
+func isFlaggedAnthropicMessagesRequest(tk *tokenizer.TiktokenTokenizer, r anthropicMessagesRequest, cfg config.AnthropicConfig) (*flaggingResult, error) {
 	return isFlaggedRequest(tk,
```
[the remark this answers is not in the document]
arafatkatze: I will do the refactoring after getting rid of the tokenizer see [here](https://github.com/sourcegraph/sourcegraph/pull/61134#discussion_r1530930597)

internal/completions/tokenusage/tokenusage.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/sourcegraph/sourcegraph/internal/completions/tokenizer"
+	"github.com/sourcegraph/sourcegraph/internal/rcache"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+)
+
+type TokenUsageManager struct {
```
rafax, line 12: nit: call this Manager, so it's used as `tokenusage.Manager`, not `tokenusage.TokenUsageManager`

rafax: https://go.dev/wiki/CodeReviewComments#package-names

internal/completions/tokenusage/tokenusage.go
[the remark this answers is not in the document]
rafax: > The way to handle this con is to have a shared mutex lock

This won't work if you have more than one process exporting usage (and it seems like it gets exported from `frontend` which generally runs multiple replicas)?

internal/completions/tokenusage/tokenusage.go
arafatkatze: Makes sense

👍

internal/completions/tokenusage/tokenusage.go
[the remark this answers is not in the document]
arafatkatze: Yeah you are right, that won't work. So I did something different. Basically I am now gonna retrieve a value from the redis and then use DecrBy and then because that operation in itself is atomic it will only decrease it by the value that it sent to our dashboards.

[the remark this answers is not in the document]
arafatkatze: CC @rafax @slimsag

internal/updatecheck/client.go
```diff
@@ -868,13 +879,13 @@ func Start(logger log.Logger, db database.DB) {
...
 
-	const delay = 30 * time.Minute
+	const delay = 5 * time.Minute
 	scopedLog := logger.Scoped("updatecheck")
 	for {
 		check(scopedLog, db)
 
 		// Randomize sleep to prevent thundering herds.
-		randomDelay := time.Duration(rand.Intn(600)) * time.Second
+		randomDelay := time.Duration(rand.Intn(10)) * time.Second
```
arafatkatze, line 888: Currently set to this for testing but I will change it before the merge

internal/completions/tokenusage/tokenusage.go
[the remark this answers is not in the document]
arafatkatze: Got rid of ttl alltogether and now I just decrement the keys atomically when sending to redis.

cmd/frontend/internal/bg/store_token_usage_in_postgres.go
```diff
@@ -0,0 +1,47 @@
+package bg
+
+import (
+	//nolint:logging // TODO move all logging to sourcegraph/log
```
slimsag, line 4: what's this needed for? I suspect it can be removed

internal/updatecheck/client.go
slimsag: definitely need to undo these changes, same for `const delay` above

cmd/cody-gateway/internal/httpapi/completions/anthropic_test.go
```diff
@@ -37,7 +37,7 @@ func TestIsFlaggedAnthropicRequest(t *testing.T) {
 		RequestBlockingEnabled:         true,
 		AllowedPromptPatterns:          []string{strings.ToLower(validPreamble)},
 	}
-	tk, err := tokenizer.NewAnthropicClaudeTokenizer()
+	tk, err := tokenizer.NewAnthropicClaudeTokenizer("anthropic/claude-2")
```
slimsag, line 40: same things w.r.t. constants in this file

cmd/cody-gateway/internal/httpapi/completions/anthropicmessages_test.go
```diff
@@ -161,7 +161,7 @@ func TestIsFlaggedAnthropicMessagesRequest(t *testing.T) {
 }
 
 func TestAnthropicMessagesRequestJSON(t *testing.T) {
-	_, err := tokenizer.NewAnthropicClaudeTokenizer()
+	_, err := tokenizer.NewAnthropicClaudeTokenizer("anthropic/claude-2")
```
slimsag, line 164: Also with this function call `tokenizer.NewAnthropicClaudeTokenizer`

internal/rcache/rcache.go
```diff
@@ -108,6 +110,67 @@ func (r *Cache) SetWithTTL(key string, b []byte, ttl int) {
...
+	value, err := strconv.ParseInt(string(b), 10, 64) // 10 is the base, 64 is the bit size
+	if err != nil {
+		return 0, false, errors.Newf("failed to convert value to int", "value", string(b), "error", err)
+	}
+	return value, true, nil
+}
+
+// IncrbyInt64 increments the integer value of a key by the given amount.
+// It returns the new value after the increment.
+func (r *Cache) IncrbyInt64(key string, value int64) (int64, error) {
```
slimsag, line 136: I think you have a typo here `IncrbyInt64` -> `IncrByInt64`

slimsag: same for `DecrbyInt64`

cmd/frontend/internal/bg/store_token_usage_in_postgres.go
```diff
@@ -0,0 +1,47 @@
...
+func StoreTokenUsage(ctx context.Context, db database.DB) error {
+	recorder := telemetryrecorder.New(db)
+	tokenManager := tokenusage.NewManager()
+	tokenUsageData, err := tokenManager.FetchTokenUsageDataForAnalysis()
+	if err != nil {
+		return err
+	}
+	convertedTokenUsageData := make(telemetry.EventMetadata)
+	for key, value := range tokenUsageData {
+		convertedTokenUsageData[telemetry.ConstString(key)] = value
```
slimsag, line 38:
```suggestion
		convertedTokenUsageData[key] = value
```

Can we just write this instead? It's weird that `telemetry.ConstString` is a thing TBH, would be ideal to not expose it from that package if that works.

internal/completions/client/anthropic/anthropic.go
```diff
@@ -99,29 +112,41 @@ func (a *anthropicClient) Stream(
...
 			return err
 		}
 
 	}
+	if dec.Err() != nil {
+		return dec.Err()
+	}
 
-	return dec.Err()
+	return nil
```
slimsag, line 149: This if statement + `return nil` is exactly the same as `return dec.Err()` - no need to make this change.

Actually the `if` statement added is potentially more error-prone, because it relies on `dec.Err()` returning the same error twice

internal/completions/tokenusage/tokenusage.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func (m *Manager) RetrieveAndResetTokenUsageData() (map[string]interface{}, error) {
+	tokenUsageData, err := m.fetchTokenUsageData(true) // true to decrement counts
+	if err != nil {
+		return nil, err
+	}
+
+	// Grouping token usage data under a 'models' key
+	modelsData := make([]map[string]interface{}, 0)
```
slimsag, line 75:
```suggestion
	modelsData := make([]map[string]interface{}, 0, len(tokenUsageData))
```

more optimal

cmd/frontend/internal/bg/store_token_usage_in_postgres.go
```diff
@@ -0,0 +1,47 @@
...
+
+		// Wait for 1 minute before the next execution
+		time.Sleep(time.Minute)
+	}
+}
+
+func StoreTokenUsage(ctx context.Context, db database.DB) error {
+	recorder := telemetryrecorder.New(db)
+	tokenManager := tokenusage.NewManager()
+	tokenUsageData, err := tokenManager.FetchTokenUsageDataForAnalysis()
```
slimsag, line 32: It occurs to me that as the code is written today the following could happen:

* Usage is recorded in redis
* The update client runs and sends recorded redis data to our telemetry servers (or e.g. in a fully offline deployment, fails to send data to our telemetry servers.) Either way, `RetrieveAndResetTokenUsageData` gets called and resets all counters to zero,
* Then 1s later a loop of `ScheduleStoreTokenUsage` runs, but since `RetrieveAndResetTokenUsageData` ran before it now the usage is set to zero. No usage is recorded.

This could lead to the last minute of data being lost somewhat infrequently, which leads to the overall numbers being incorrect and suspect.

The only obvious way I see to address this is to make sure that `StoreTokenUsage` directly prior to any `RetrieveAndResetTokenUsageData` calls.

I also suggest changing the `// Wait for 1 minute before the next execution` to 5 minutes, as that would surface this type of bug better (and 1 minute is fairly aggressive anyway)

slimsag (changes requested):
Left some feedback inline, you're definitely on the right track though! Give me a ping on Slack when you need another look.

cmd/frontend/internal/bg/store_token_usage_in_postgres.go
arafatkatze: This comment stayed because I copy paste a lot of stuff. I fixed it.

cmd/cody-gateway/internal/httpapi/completions/anthropic.go
```diff
@@ -33,7 +33,7 @@ func NewAnthropicHandler(
 	autoFlushStreamingResponses bool,
 ) (http.Handler, error) {
 	// Tokenizer only needs to be initialized once, and can be shared globally.
-	anthropicTokenizer, err := tokenizer.NewAnthropicClaudeTokenizer()
+	anthropicTokenizer, err := tokenizer.NewTokenizer("anthropic")
```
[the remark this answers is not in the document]
arafatkatze: Makes so much sense 🙇

internal/rcache/rcache.go
arafatkatze: Good catch. I made the fix

internal/completions/tokenizer/tokenizer_test.go
```diff
@@ -91,7 +91,7 @@ Assistant: My answer is (`,
 }
 
 func TestTokenize(t *testing.T) {
-	tk, err := tokenizer.NewAnthropicClaudeTokenizer()
+	tk, err := tokenizer.NewAnthropicClaudeTokenizer("testModel")
```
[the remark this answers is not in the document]
arafatkatze: I know what you mean. Basically I got rid of exposing `NewAnthropicClaudeTokenizer` and then now I only use `tokenizer.NewTokenizer `everywhere and I use model names as constants so this makes everything a whole lot cleaner and the interface is pretty clean too.

cmd/frontend/internal/bg/store_token_usage_in_postgres.go
arafatkatze: sadly no 😢

internal/completions/client/anthropic/anthropic.go
arafatkatze: 👍

cmd/frontend/internal/bg/store_token_usage_in_postgres.go
dadlerj: >I also suggest changing the // Wait for 1 minute before the next execution to 5 minutes, as that would surface this type of bug better (and 1 minute is fairly aggressive anyway)

I agree with this!

cmd/frontend/internal/bg/store_token_usage_in_postgres.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		return err
+	}
+	convertedTokenUsageData := make(telemetry.EventMetadata)
+	for key, value := range tokenUsageData {
+		castedKey := telemetry.ConstString(key) // Cast the key to telemetry.ConstString
+		convertedTokenUsageData[castedKey] = value
+	}
+
+	if err := recorder.Record(ctx, "llmTokenCounter", "modelUsage", &telemetry.EventParameters{
```
dadlerj, line 39:
```suggestion
	if err := recorder.Record(ctx, "cody.llmTokenCounter", "record", &telemetry.EventParameters{
```

To confirm—this event occurs every [1] minute(s), right? So the _action_ here is more of a "record" or a "store" action than a "model was used", right? Let me know if the logic behind these changes and questions makes sense

internal/completions/tokenusage/tokenusage.go
```diff
@@ -0,0 +1,133 @@
...
+func (m *Manager) RetrieveAndResetTokenUsageData() (map[string]interface{}, error) {
+	tokenUsageData, err := m.fetchTokenUsageData(true) // true to decrement counts
+	if err != nil {
+		return nil, err
+	}
+
+	// Grouping token usage data under a 'models' key
+	modelsData := make([]map[string]interface{}, 0, len(tokenUsageData))
+	for model, tokens := range tokenUsageData {
+		modelData := map[string]interface{}{
```
dadlerj, line 77: Not a dealbreaker, but any reason not to make this a struct type? It's probably fine to leave it simple like this, but it might be worth adding a comment here that this structure shouldn't be changed without a corresponding change to our BigQuery telemetry database.

cmd/frontend/internal/bg/store_token_usage_in_postgres.go
arafatkatze: Makes sense I added a fix

internal/completions/tokenusage/tokenusage.go
arafatkatze: Makes sense so Made a struct type

cmd/frontend/internal/bg/store_token_usage_in_postgres.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+func ScheduleStoreTokenUsage(ctx context.Context, db database.DB) {
+	for {
+		err := storeTokenUsageinDb(ctx, db)
+		if err != nil {
+			fmt.Printf("Error storing token usage: %v\n", err)
+		}
+
+		// Wait for 1 minute before the next execution
```
slimsag, line 21:
```suggestion
		// Wait for 5 minutes before the next execution
```

internal/rcache/rcache.go
```diff
@@ -108,6 +110,67 @@ func (r *Cache) SetWithTTL(key string, b []byte, ttl int) {
...
+	if err != nil {
+		return newValue, errors.Newf("failed to execute redis command", "cmd", "INCRBY", "error", err)
+	}
+
+	if r.ttlSeconds > 0 {
+		// Optionally, set a TTL on the key if ttlSeconds is specified for the cache.
+		err = kv().Expire(r.rkeyPrefix()+key, r.ttlSeconds)
+		if err != nil {
+			return newValue, errors.Newf("failed to execute redis command", "cmd", "INCRBY", "error", err)
+			// Decide whether to return this error or not based on your error handling policy.
```
slimsag, line 147:
```suggestion
```

internal/rcache/rcache.go
```diff
@@ -108,6 +110,67 @@ func (r *Cache) SetWithTTL(key string, b []byte, ttl int) {
...
+	if err != nil {
+		return newValue, errors.Newf("failed to execute redis command", "cmd", "DECRBY", "error", err)
+	}
+
+	if r.ttlSeconds > 0 {
+		// Optionally, set a TTL on the key if ttlSeconds is specified for the cache.
+		err = kv().Expire(r.rkeyPrefix()+key, r.ttlSeconds)
+		if err != nil {
+			return newValue, errors.Newf("failed to execute redis command", "cmd", "DECRBY", "error", err)
+			// Decide whether to return this error or not based on your error handling policy.
```
slimsag, line 167:
```suggestion
```

internal/updatecheck/client.go
```diff
@@ -868,13 +910,13 @@ func Start(logger log.Logger, db database.DB) {
 	}
 	started = true
 
-	const delay = 30 * time.Minute
+	const delay = 5 * time.Minute
```
slimsag, line 913: @arafatkatze please make sure you undo this change and the one below.

slimsag (approved):

PR #47517, opened 2023-02-11
remove enableLegacyExtensions and all now-unreachable code

sqs:
This is part of the extension API deprecation (https://docs.google.com/document/d/10vtoe-kpNvVZ8Etrx34bSCoTaCCHxX8o3ncCmuErPZo/edit).

There are a lot more refactors to do after this, but this removes the admin-facing `enableLegacyExtensions` setting and some of the root-level instantiation and usage of the extensions controller API.

## Test plan

Check that hovers still work on dotcom.

PR #61138, opened 2024-03-14
web: Remove extensions controller remnants from client/web

fkling:
Support for extensions, even legacy support, [had been removed](https://github.com/sourcegraph/sourcegraph/pull/47517) for quite a while. Now it seems that none of the code actually needs the extensions controller shim. At most they are registering the current user or document, but there seems to be no code that actually reads that information.

There is a lot of code that interfaces with extension controller in `client/shared/` which is used by `client/browser/`. This needs a separate pass for cleaning up.

Did quite a bit for ripgrep-ing to find occurrences of `extensionsController`, `extHostAPI`, etc.

## Test plan

Verified that code intel hovercards, go to definition and find references work.


vovakulikov (approved):
Good bye, old "friend", thanks for clean up

PR #61139, opened 2024-03-14
search: simplify Deduper

stefanhengl:
The behavior of `Deduper` was very surprising to me. 
Apart from the expected deduping based on `Match.Key()`, `Deduper` kept track of the deduped results but none of the call sites ever read the results. On top of that, `Deduper` modified the matches passed to it via `Deduper.Add` which, looking at the call sites, I don't think was intentional, because all call sites keep track of results themselves. The test I fix in the second commit documents that behavior.

Notes:
- I also removed NewDedupingStream which was unused.

## Test plan
- updated unit test
- CI


stefanhengl:
@camdencheek tagging you here because you might have more context around `Deduper`.

camdencheek (approved):
IIRC, the history here is that `Deduper` used to handle the responsibilities of `Merger` in addition to its current capacity as a pure deduper. So, previously, it was also used for `And` and `Or` jobs, which _do_ need to merge the matches together.

I think your change is safe from a "preserve behavior" perspective. I do think this uncovers an open question about the semantics of `select:` though: should `select:` merge results with the same key? Right now it won't and it will just ignore any additional results from the same file (for example). Given that this is a pretty difficult edge case to hit, even intentionally, I don't think answering this question is urgent.

stefanhengl:
> IIRC, the history here is that `Deduper` used to handle the responsibilities of `Merger` in addition to its current capacity as a pure deduper. So, previously, it was also used for `And` and `Or` jobs, which _do_ need to merge the matches together.

That makes a lot sense. Thanks for the context!

PR #61140, opened 2024-03-14
SG Docker Commands

jamesmcnamara:
Adds first class support for docker commands to `sg start`. This includes watching a bazel target and reloading.
Additionally, 
* Refactors out common command parameters to be shared
* Adds a logfile parameter to all commands to redirect logs
For instance, Grafana looked like this:
```yaml
  grafana:
    cmd: |
      if [[ $(uname) == "Linux" ]]; then
        # Linux needs an extra arg to support host.internal.docker, which is how grafana connects
        # to the prometheus backend.
        ADD_HOST_FLAG="--add-host=host.docker.internal:host-gateway"

        # Docker users on Linux will generally be using direct user mapping, which
        # means that they'll want the data in the volume mount to be owned by the
        # same user as is running this script. Fortunately, the Grafana container
        # doesn't really care what user it runs as, so long as it can write to
        # /var/lib/grafana.
        DOCKER_USER="--user=$UID"
      fi

      echo "Grafana: serving on http://localhost:${PORT}"
      echo "Grafana: note that logs are piped to ${GRAFANA_LOG_FILE}"
      docker run --rm ${DOCKER_USER} \
        --name=${CONTAINER} \
        --cpus=1 \
        --memory=1g \
        -p 0.0.0.0:3370:3370 ${ADD_HOST_FLAG} \
        -v "${GRAFANA_DISK}":/var/lib/grafana \
        -v "$(pwd)"/dev/grafana/all:/sg_config_grafana/provisioning/datasources \
        grafana:candidate >"${GRAFANA_LOG_FILE}" 2>&1
    install: |
      mkdir -p "${GRAFANA_DISK}"
      mkdir -p "$(dirname ${GRAFANA_LOG_FILE})"
      docker inspect $CONTAINER >/dev/null 2>&1 && docker rm -f $CONTAINER
      bazel build //docker-images/grafana:image_tarball
      docker load --input $(bazel cquery //docker-images/grafana:image_tarball --output=files)
    env:
      GRAFANA_DISK: $HOME/.sourcegraph-dev/data/grafana
      # Log file location: since we log outside of the Docker container, we should
      # log somewhere that's _not_ ~/.sourcegraph-dev/data/grafana, since that gets
      # volume mounted into the container and therefore has its own ownership
      # semantics.
      # Now for the actual logging. Grafana's output gets sent to stdout and stderr.
      # We want to capture that output, but because it's fairly noisy, don't want to
      # display it in the normal case.
      GRAFANA_LOG_FILE: $HOME/.sourcegraph-dev/logs/grafana/grafana.log
      IMAGE: grafana:candidate
      CONTAINER: grafana
      PORT: 3370
      # docker containers must access things via docker host on non-linux platforms
      DOCKER_USER: ''
      ADD_HOST_FLAG: ''
      CACHE: false
```
But can now be written:
```yaml
 grafana:
    target: //docker-images/grafana:image_tarball
    docker:
      image: grafana:candidate
      ports:
        - 3370
      flags:
        cpus: 1
        memory: 1g
      volumes:
        - from: $HOME/.sourcegraph-dev/data/grafana
          to: /var/lib/grafana
        - from: $(pwd)/dev/grafana/all
          to: /sg_config_grafana/provisioning/datasources

      linux:
        flags:
          # Linux needs an extra arg to support host.internal.docker, which is how grafana connects
          # to the prometheus backend.
          add-host: host.docker.internal:host-gateway

          # Docker users on Linux will generally be using direct user mapping, which
          # means that they'll want the data in the volume mount to be owned by the
          # same user as is running this script. Fortunately, the Grafana container
          # doesn't really care what user it runs as, so long as it can write to
          # /var/lib/grafana.
          user: $UID
    # Log file location: since we log outside of the Docker container, we should
    # log somewhere that's _not_ ~/.sourcegraph-dev/data/grafana, since that gets
    # volume mounted into the container and therefore has its own ownership
    # semantics.
    # Now for the actual logging. Grafana's output gets sent to stdout and stderr.
    # We want to capture that output, but because it's fairly noisy, don't want to
    # display it in the normal case.
    logfile: $HOME/.sourcegraph-dev/logs/grafana/grafana.log

    env:
      # docker containers must access things via docker host on non-linux platforms
      CACHE: false
```
## Test plan
Unit tests for docker command generation and I tested that the postgres_exporter reloads on file changes


dev/sg/internal/run/command.go
```diff
@@ -291,34 +237,59 @@ type outputOptions struct {
...
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to get user home directory")
+		}
+		logfile = filepath.Join(home, strings.Replace(strings.Replace(logfile, "~/", "", 1), "$HOME", "", 1))
+	}
+	parent := filepath.Dir(logfile)
+	if err := os.MkdirAll(parent, os.ModePerm); err != nil {
+		return nil, err
+	}
+	return os.Create(logfile)
```
burmudar, line 290:
```suggestion
       // we don't have to worry about the file existing already and growing large, since this will truncate the file if it exists
	return os.Create(logfile)
```

dev/sg/internal/run/docker_command_test.go
```diff
@@ -0,0 +1,148 @@
...
+func TestCompileGrafanaCommand(t *testing.T) {
+	want := `docker inspect grafana > /dev/null 2>&1 && docker rm -f grafana
+docker load -i ./fake_img.tar
+
+mkdir -p $HOME/.sourcegraph-dev/data/grafana
+mkdir -p $(pwd)/dev/grafana/all
+
+echo hello
+docker run --rm --name grafana ` +
+		`-v $HOME/.sourcegraph-dev/data/grafana:/var/lib/grafana ` +
```
burmudar, line 77: can we not remove the backticks here and do
```
docker run --rm --name grafana  \
 -v $HOME/.sourcegraph-dev/data/grafana:/var/lib/grafana
```

dev/sg/internal/run/docker_commmand.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	for _, env := range toSortedPairs(dc.GetDockerEnv(isLinux)) {
+		fmt.Fprintf(&cmd, ` -e %s="%s"`, env.Key, env.Value)
+	}
+	fmt.Fprintf(&cmd, " %s %s", dc.Docker.Image, dc.Config.Args)
+	return cmd.String()
+
+}
+
+func (dc DockerCommand) GetCmd(isLinux bool, bin string) string {
```
burmudar, line 188: nit: I think the params order should be `bin string, isLinux bool`

dev/sg/internal/run/docker_commmand.go
```diff
@@ -0,0 +1,230 @@
...
+	cleanup := fmt.Sprintf("docker inspect %s > /dev/null 2>&1 && docker rm -f %s", dc.Config.Name, dc.Config.Name)
+	load := dc.GetDockerImage(bin)
+	docker := dc.GetDockerCommand(isLinux)
```
burmudar, line 191: why aren't we using the same thing we do on 202 aka `runtime.GOOS == "linux"` ?

jamesmcnamara: Testing. I want to be able to test that the docker command output changes based on that param.

dev/sg/internal/run/docker_command_test.go
jamesmcnamara: I don't think so because the generated string doesn't have those newlines. I tried it and the test failed.

burmudar: Ah right. Unfortunate :(

dev/sg/internal/run/docker_commmand.go
burmudar: Right gotcha 👍🏼

burmudar (approved):
nice

burmudar:
Build failed because of a gofmt issue which is probably due to the comment I suggested

Issue #61055, opened 2024-03-13
Expose an API to surface usage and rate limits info from Cody Gateway Redis Cache


thenamankumar:
We want to expose usage and rate limits data for easy debugging purposes. Cody Gateway redis cache is the source of truth for how rate limits ARE getting applied, not necessarily how they should get applied.

The source of truth for how they SHOULD get applied is/will be SSC only. But to debug the issues we need an easy way to compare how-they-should vs how-they-are .

The only goal of exposing this API is to easily surface the value of rate limits config from the cody gateway redis storage for easy debugging.

PR #61144, opened 2024-03-14
Expose Rate Limit Status from Cody Gateway

thenamankumar:
closes: https://github.com/sourcegraph/sourcegraph/issues/61055

This PR exposes a dotcom-only API for debugging purposes to easily surface the value of per feature rate limits config & usage for a user from the cody gateway redis storage. This API will only be used for debugging purposes. 

CleanShot 2024-03-19 at 12 49 27@2x

## Test plan
- Visit `/.api/console`
- Perform the following query:
``` graphql
query {
  currentUser {
    id
    codyGatewayRateLimitStatus {
      feature 
      interval
      limit
      nextLimitReset
      usage
    }
  }
}
```
- It should respond with the details as shown in screenshot above.
- If it returns in 401 error, make sure `CODY_GATEWAY_DOTCOM_ACCESS_TOKEN` is configured correctly.

### DIFF
--- a/cmd/frontend/graphqlbackend/schema.graphql
+++ b/cmd/frontend/graphqlbackend/schema.graphql
@@ -6638,6 +6638,11 @@ type User implements Node & SettingsSubject & Namespace {
     """
     codySubscription: CodySubscription
     """
+    The rate limit status for the user stored in Cody Gateway.
+    On dotcom only the user and site admins can access this field.
+    """
+    codyGatewayRateLimitStatus: [CodyGatewayRateLimitStatus!]
+    """
     DEPRECATED
     Whether the user has enrolled for Cody Pro.
     """
--- a/cmd/frontend/graphqlbackend/user.go
+++ b/cmd/frontend/graphqlbackend/user.go
@@ -233,6 +233,32 @@ func (r *UserResolver) CodySubscription(ctx context.Context) (*CodySubscriptionR
 	return &CodySubscriptionResolver{subscription: subscription}, nil
 }
 
+func (r *UserResolver) CodyGatewayRateLimitStatus(ctx context.Context) (*[]RateLimitStatus, error) {
+	if !dotcom.SourcegraphDotComMode() {
+		return nil, errors.New("this feature is only available on sourcegraph.com")
+	}
+
+	// 🚨 SECURITY: Only the user and admins are allowed to access the user's
+	// settings, because they may contain secrets or other sensitive data.
+	if err := auth.CheckSiteAdminOrSameUserFromActor(r.actor, r.db, r.user.ID); err != nil {
+		return nil, err
+	}
+
+	limits, err := cody.GetGatewayRateLimits(ctx, r.user.ID, r.db)
+	if err != nil {
+		return nil, err
+	}
+
+	rateLimits := make([]RateLimitStatus, 0, len(limits))
+	for _, limit := range limits {
+		rateLimits = append(rateLimits, &codyRateLimit{
+			rl: limit,
+		})
+	}
+
+	return &rateLimits, nil
+}
+
 func (r *UserResolver) CreatedAt() gqlutil.DateTime {
 	return gqlutil.DateTime{Time: r.user.CreatedAt}
 }
--- a/internal/cody/rate_limits.go
+++ b/internal/cody/rate_limits.go
@@ -3,10 +3,16 @@ package cody
 import (
 	"context"
 	"encoding/hex"
+	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
+	"sort"
+
+	"github.com/sourcegraph/log"
 
 	"github.com/sourcegraph/sourcegraph/internal/accesstoken"
+	"github.com/sourcegraph/sourcegraph/internal/codygateway"
 	"github.com/sourcegraph/sourcegraph/internal/conf"
 	"github.com/sourcegraph/sourcegraph/internal/conf/conftypes"
 	"github.com/sourcegraph/sourcegraph/internal/database"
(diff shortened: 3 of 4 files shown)


cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -6638,6 +6638,10 @@ type User implements Node & SettingsSubject & Namespace {
     """
     codySubscription: CodySubscription
     """
+    The rate limit status for the user stored in Cody Gateway.
```
chrsmith, line 6641: This doesn't quite sound right. Perhaps we need to instead think of this as something more abstract?

Cody Gateway is only used for Cody _except_ Cody Enterprise under BYOK mode. And in that case, maybe we still want to enforce some sort of rate limiting... so maybe we should name this `codyRateLimits` instead?

And describe it as:

"""Rate limits for Cody features for the given user.

Typically this is managed by Cody Gateway. But for "bring-your-own-key" configurations, per-user rate limiting is managed within the Sourcegraph instance."""

Does that make sense?

It seems wrong to call this `codyGatewayRateLimitStatus` when Cody Gateway will not be used in every supported case. And in the cases where Cody Gateway _isn't_ used, we may still want to expose this information.

chrsmith, line 6641: ⚠️ Since the GraphQL handler requires site-admin permissions, we should call that out in this `.graphql` file as well. (Just so a user doesn't think they can call this, only to find out at runtime that they can't.)

internal/cody/rate_limits.go
```diff
@@ -17,33 +21,83 @@ import (
 
 // RefreshGatewayRateLimits refreshes the rate limits for the user on Cody Gateway.
 func RefreshGatewayRateLimits(ctx context.Context, userID int32, db database.DB) (error, int) {
+	resp, err := newGatewayRequestForUser(ctx, userID, db, http.MethodPost, "/v1/limits/refresh", nil)
+	if err != nil {
+		return err, http.StatusInternalServerError
```
chrsmith, line 26: I'd expect to see a call to `logger.Error(...)` here, but perhaps that is redundant.
Is it the case we already do that within `newGatewayRequestForUser`?

internal/cody/rate_limits.go
```diff
@@ -17,33 +21,83 @@ import (
...
+	}
+	if resp == nil {
+		return nil, http.StatusOK
```
chrsmith, line 29: When is this the case? Does this mean that the user ID isn't found on the Cody Gateway side? It seems like this should be treated as some sort of error... since Cody Gateway should be returning _something_, right?

chrsmith (commented):
Just some notes, this seems fine. But I'd like to make sure we think about this in the context of how "per-user rate limiting" should be part of the Cody Enterprise product.

We perform per-user rate limits for Cody PLG, both as that's how we differentiate the Cody Free and Cody Pro tiers. But for Enterprise scenarios, this would be a good capability to have too. (e.g. ensuring that users of a Cody Enterprise subscription cannot consume more than 10% of all available quota each day.)

We don't need to have a perfect idea for how this would be wired up. But just thinking through what the API should look would avoid us needing to redo it later.

What do you think?

vdavid:
@thenamankumar, as per your request [here](https://sourcegraph.slack.com/archives/C05PC7AKFQV/p1710417760350909), I've checked out the branch, ran `sh start dotcom`, went to `https://sourcegraph.test:3443/api/console`.
I got
```
{
  "errors": [
    {
      "message": "non-200 response fetching limits from Gateway: 401",
      "path": [
        "currentUser",
        "codyGatewayRateLimitStatus"
      ]
    }
  ],
  "data": {
    "currentUser": {
      "id": "VXNlcjox",
      "codyGatewayRateLimitStatus": null
    }
  }
}
```
which, I think, is not the intended outcome? (The test plan doesn't clarify)

internal/cody/rate_limits.go
thenamankumar: This is when the gateway is not the configured provider.

thenamankumar:
> @thenamankumar, as per your request [here](https://sourcegraph.slack.com/archives/C05PC7AKFQV/p1710417760350909), I've checked out the branch, ran `sh start dotcom`, went to `https://sourcegraph.test:3443/api/console`. I got
> 
> ```
> {
>   "errors": [
>     {
>       "message": "non-200 response fetching limits from Gateway: 401",
>       "path": [
>         "currentUser",
>         "codyGatewayRateLimitStatus"
>       ]
>     }
>   ],
>   "data": {
>     "currentUser": {
>       "id": "VXNlcjox",
>       "codyGatewayRateLimitStatus": null
>     }
>   }
> }
> ```
> 
> which, I think, is not the intended outcome? (The test plan doesn't clarify)

You have the save local config issue which I was struggling with, more context [here](https://sourcegraph.slack.com/archives/C04MYFW01NV/p1710832780776889).

internal/cody/rate_limits.go
thenamankumar: Added

internal/cody/rate_limits.go
```diff
@@ -17,33 +23,90 @@ import (
...
+	// Make sure the limits are always returned in the same order, since the map
+	// above doesn't have deterministic ordering.
+	sort.Slice(rateLimits, func(i, j int) bool {
+		return rateLimits[i].Feature < rateLimits[j].Feature
+	})
+
+	return rateLimits, nil
+}
+
+func newGatewayRequestForUser(ctx context.Context, userID int32, db database.DB, method string, pathname string, body io.Reader) (*http.Response, error) {
```
vdavid, line 92: The naming doesn't seem right here. For one thing, I strongly prefer function names to start with a verb. In our Go codebase, when we do `func new****`, we use those pretty much as constructors. In this case, `new***Request***`, I'd expect this to return a new request. But it actually runs the constructed request and returns the response, so the naming is misleading. Something like `callCodyGateway`?

Another tip: maybe only pass the `AccessTokenStore` from `db` to narrow down that interface? `db` is pretty wide and we only use that single store.

vdavid (approved):
Overall good, left some comments proposing a few small changes. Please review them before merging.

cmd/frontend/graphqlbackend/user.go
```diff
@@ -233,6 +233,32 @@ func (r *UserResolver) CodySubscription(ctx context.Context) (*CodySubscriptionR
...
 
+func (r *UserResolver) CodyGatewayRateLimitStatus(ctx context.Context) (*[]RateLimitStatus, error) {
+	if !dotcom.SourcegraphDotComMode() {
+		return nil, errors.New("this feature is only available on sourcegraph.com")
+	}
+
+	// 🚨 SECURITY: Only the user and admins are allowed to access the user's
+	// settings, because they may contain secrets or other sensitive data.
+	if err := auth.CheckSiteAdminOrSameUserFromActor(r.actor, r.db, r.user.ID); err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
thenamankumar: No, it won't be better. This is a debugging API, the goal is to surface the error not hide it being a plain text message. 

1. user.go:243: `CheckSiteAdminOrSameUserFromActor` already returns a well-formated error string `"must be authenticated as the authorized user or site admin"` as error message. For every single use of this function in our codebase we do return back the error from the function call directly. [Examples](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/sourcegraph$+CheckSiteAdminOrSameUserFromActor&patternType=keyword&sm=0).
2. user.go:249 This is a debug only API and no one is using it, so internal logging doesn't help with anything, not even debugging. Keep it plain and simple and returning the error as response will only make debugging easier, rather than only getting a plain text response and then having to check actual error message in GCP. But anyways I have already added internal logs for most of the error cases in rate_limits.go already, which also covers the `GetGatewayRateLimits` func.

internal/cody/rate_limits.go
thenamankumar: I have renamed it to `requestGatewayWithUserCreds`.

I prefer passing db store in the args vs particular model stores because:
- that is being followed everywhere throughout the codebase.
- the caller does not need to care about which particular models will be used.
- later if the function needs something more from DB, we won't need to update every single function call.

PR #61161, opened 2024-03-14
[sg] share runner implementation between run and start

jamesmcnamara:
Instead of putting any work into `sg run` I just created a shared runner between `sg run` and `sg start`. I also added a flag to `sg start` to run commands directly because `sg start` has more flags and features built in, and also it seems silly to have two commands that do the same thing. I'd love to just make `sg run` and alias for `sg start --commands` or something.

Took a bunch of code to do this because the prior implementations were pretty fragile.

## Test plan
Ran locally.


dev/sg/sg_start.go
```diff
@@ -94,7 +98,10 @@ sg start -describe single-program
 				Name:  "profile",
 				Usage: "Starts up pprof on port 6060",
 			},
-
+			&cli.BoolFlag{
+				Name:  "commands",
+				Usage: "Signifies that you will be passing in a list of commands to run",
```
bobheadxi, line 103:
```suggestion
				Usage: "Signifies that you will be passing in individual commands to run, instead of a set of commands",
```

dev/sg/sg_start.go
```diff
@@ -94,7 +98,10 @@ sg start -describe single-program
 				Name:  "profile",
 				Usage: "Starts up pprof on port 6060",
 			},
-
+			&cli.BoolFlag{
+				Name:  "commands",
```
bobheadxi, line 102:
```suggestion
				Name:  "commands",
				Aliases: []string{"c"}
				
```

bobheadxi: I think you should also update `BashComplete` to work with `-c`, otherwise the completion will suggest commandsets still when you actually want commands

bobheadxi (commented):
I must say I'm not very up-to-date on the `run`/`start` internals these days, so review is very cursory 😛 Overall, just a single mechanism (at least internally) makes sense to me

> I also added a flag to sg start to run commands directly so because sg start has more flags and features built in, and also it seems silly to have two commands that do the same thing. I'd love to just make sg run and alias for sg start --commands or something.

I think deprecating `run` entirely makes sense too, `sg start -c foobar` is nice

bobheadxi (approved):

dev/sg/sg_start.go
jamesmcnamara: Unfortunately "-c" is already being used by "--critical" debug level. Either we could change that or do you have a better name for `--commands`?  `--services` and '-s'?

bobheadxi: Hm maybe `-cmd` as an alias?

jamesmcnamara: @bobheadxi Okay this ended being a HUGE rabbit hole of working on the bash completions. The bad news is that I don't have it working here, but the semi-good news is that I do have [this](https://github.com/sourcegraph/sourcegraph/pull/61239) WIP PR that upgrades to v3 of CLI which seems to not have the crashing issues but does have some other known issues I'm tracking. I can keep chipping away at that and when it gets released we should get a suite of up updated bash command improvements.

PR #61162, opened 2024-03-14
Svelte: search results styling updates

camdencheek:
This makes a handful of small styling updates to the Svelte implementation of the search results page to bring it more in line with the designs and the react app. It is not comprehensive, but it takes care of the big things.

## Test plan

Manual testing. See [Loom recording](https://www.loom.com/share/a5dd1ac1700d43be98ca483ec3c85f25)

### DIFF
--- a/client/web-sveltekit/src/lib/search/dynamicFilters/CountBadge.svelte
+++ b/client/web-sveltekit/src/lib/search/dynamicFilters/CountBadge.svelte
@@ -30,3 +30,10 @@
         {/if}
     </span>
 {/if}
+
+<style lang="scss">
+    span.count :global(span) {
+        background-color: var(--secondary-2);
+        color: var(--text-muted);
+    }
+</style>
--- a/client/web-sveltekit/src/routes/search/FileContentSearchResult.svelte
+++ b/client/web-sveltekit/src/routes/search/FileContentSearchResult.svelte
@@ -165,10 +165,7 @@
         }
 
         &:hover {
-            // Set the background color of the whole div to also change the color of the padding,
-            // but also make --background-color transparent to unset the background color of CodeExcerpt.
             background-color: var(--subtle-bg-2);
-            --background-color: transparent;
         }
 
         a {
--- a/client/web-sveltekit/src/routes/search/FileContentSearchResult.svelte
+++ b/client/web-sveltekit/src/routes/search/FileContentSearchResult.svelte
@@ -74,20 +74,18 @@
     }
 
     let visible = false
-    let highlightedHTMLRows: string[][] | undefined
+    let highlightedHTMLRows: Promise<string[][]> | undefined
     $: if (visible) {
         // If the file contains some large lines, avoid stressing syntax-highlighter and the browser.
         if (!result.chunkMatches?.some(chunk => chunk.contentTruncated)) {
             // We rely on fetchFileRangeMatches to cache the result for us so that repeated
             // calls will not result in repeated network requests.
-            fetchFileRangeMatches({
+            highlightedHTMLRows = fetchFileRangeMatches({
                 result,
                 ranges: expandedMatchGroups.map(group => ({
                     startLine: group.startLine,
                     endLine: group.endLine,
                 })),
-            }).then(result => {
-                highlightedHTMLRows = result
             })
         }
     }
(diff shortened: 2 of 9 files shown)


client/web-sveltekit/src/lib/search/dynamicFilters/CountBadge.svelte
```diff
@@ -30,3 +30,10 @@
         {/if}
     </span>
 {/if}
+
+<style lang="scss">
+    span.count :global(span) {
+        background-color: var(--secondary-2);
+        color: var(--text-muted);
+    }
+</style>
```
camdencheek, line 39: I don't love breaking out of the design framework here, but we're doing that in the React app too.

client/web-sveltekit/src/routes/search/FileContentSearchResult.svelte
```diff
@@ -74,18 +74,20 @@
...
             // calls will not result in repeated network requests.
-            highlightedHTMLRows = fetchFileRangeMatches({
+            fetchFileRangeMatches({
                 result,
                 ranges: expandedMatchGroups.map(group => ({
                     startLine: group.startLine,
                     endLine: group.endLine,
                 })),
+            }).then(result => {
+                highlightedHTMLRows = result
```
camdencheek, line 90: Unrelated simplification that means we don't need to have repetitive `{#await}` branches in the markup

client/web-sveltekit/src/routes/search/RepoSearchResult.svelte
```diff
@@ -71,12 +71,15 @@
...
+                <p class="p-2 m-0" use:highlightRanges={{ ranges: descriptionMatches }}>
+                    {limitDescription(description)}
+                </p>
+            </div>
         {/key}
-    {/if}
-    {#if badges.length > 0}
+    {/if}<!--
+        Intentional weird comment to avoid adding an empty line to the body
+    -->{#if badges.length > 0}
```
camdencheek, line 80: This was surprising. Because of the way Svelte inserts default slots, this newline between the `{#if}` blocks was treated literally, making the body non-empty (containing a newline), which meant we were showing an extra separator line

client/web-sveltekit/src/routes/search/SearchResult.svelte
```diff
@@ -57,7 +57,7 @@
         align-items: center;
     }
 
-    .body {
+    .body:not(:empty) {
```
camdencheek, line 60: I couldn't figure out how to get Svelte to just skip adding the `body` div, so this is a small hack to avoid adding the `border-bottom` for an empty div

fkling (approved):
While you at it, I think the background color of the preview panel should be the same as on the file page (white in light mode).

client/web-sveltekit/src/routes/search/FileContentSearchResult.svelte
camdencheek: Oh! Right. Promises aren't cancellable 🤦 

Okay, will update this to use the new promise helpers instead or remove it from this PR

client/web-sveltekit/src/lib/search/dynamicFilters/CountBadge.svelte
camdencheek: Well...my hope with this is that we will update the design system so this is a supported style. Until that happens, `:global` is (IMO) a really great way to flag where we are breaking out of the design system.

fkling: Makes sense. I agree that with `:global` it's at least explicit that something's going on here that might needs to be revisited.

camdencheek: also cc @taiyab in case you have opinions RE: fitting these things into the design system

camdencheek:
> the background color of the preview panel should be the same as on the file page

Done!

client/web-sveltekit/src/routes/search/FileContentSearchResult.svelte
camdencheek: I went ahead and just reverted this to keep the PR styling focused

PR #61172, opened 2024-03-14
dotcom: MockSourcegraphDotComMode requires a T for cleanup

keegancsmith:
I had a suspicion another test was failing due to racing with reading dotcom.SourcegraphDotComMode and another test didn't unset it. It turned out this wasn't the case, but I ended improving the API to avoid this issue. Most call sites should be easier to read.

Test Plan: go test


camdencheek (approved):
❤️

Issue #60682, opened 2024-02-21
Cloning very large repos with GitHub App installation tokens can fail


eseliger:
On a customer instance we recently saw a clone take so long that the installation token would expire in between the start of the clone where the token is refreshed and the point at which we run `git remote show ${remoteURL}`.

eseliger:
Workaround that we used:

This is an issue, because `clone` clones to a tmp dir first, and if the post fetch actions fail, it removes it.
For updates, however, we run git fetch right inside the dir at the target location. So even if post actions fail, an update will persist the changes.

We make use of that and create an empty repo on disk in the expected location. After a server restart, it'll be marked as cloned in the DB, after ~45s, repo-updater should trigger a git fetch, and eventually populate the repo, even if the post fetch action fails again. In followup fetches, the post actions should pass again, because we won't run into timeouts.

To do that: Assume `github.com/sourcegraph/src-cli` is the repo failing.

Run: `mkdir -p /data/repos/github.com/sourcegraph/src-cli`, then initialize a repo using `git -c init.defaultBranch=main init --bare /data/repos/github.com/sourcegraph/src-cli/.git`.
Then restart gitserver.

PR #61179, opened 2024-03-14
gitserver: make vcs syncer implementation request a fresh clone URL in between each remote git operation

ggilmore:
Fixes: https://github.com/sourcegraph/sourcegraph/issues/60682 

The key insight into fixing https://github.com/sourcegraph/sourcegraph/issues/60682 is to make our `vcssyncer` implementation (that performs the fetch and clone operations) stop using a "static" remote URL for all of it's operations. 

Before, when we used a Github App installation, Sourcegraph will receive an [authentication token](https://docs.github.com/en/rest/apps/installations?apiVersion=2022-11-28#revoke-an-installation-access-token) that's valid for one hour. We embed this token in the remote URL that we use to configure git for the fetch and git remote show head (to figure out what the default branch is pointing to) operations. When a huge repository like https://github.com/sgtest/megarepo gets cloned - it's likely that it will take > 1 hour to clone. This means that the auth information embedded the remote URL will expire in between the fetch and remote head operations - which would fail the overall "clone" operation. 

The solution to this is to introduce a new  abstraction: 

```go
// RemoteURLSource is a source of a remote URL for a repository.
//
// The remote URL may change over time, so it's important to use this interface
// to get the remote URL every time instead of caching it at the call site.
type RemoteURLSource interface {
	// RemoteURL returns the latest remote URL for a repository.
	RemoteURL(ctx context.Context) (*vcs.URL, error)
}
```

The `RemoteURLSource` encapsulates all the logic to figure out what remote URL to use for a given repository (including authentication information, token refreshing, etc.). Instead of passing around a static remote URL, the `RemoteURLSource` is stored as a struct member on each of the syncers. Whenever the syncer performs a git operation (like fetch) - it consults the source for the latest remoteURL to use - and only uses that url for the current operation. This means that individual git operations now have the chance to "refresh" the token in the remote URL (if necessary) which solves the original problem.

---

Test plan:

- Create a github app installation using https://sourcegraph.com/docs/admin/external_service/github#using-a-github-app
- Add a humongous repository such as https://github.com/sgtest/megarepo to the instance that will take over an hour to clone
- See that the clone actually completes successfully


internal/github_apps/auth/auth.go
```diff
@@ -182,7 +186,30 @@ func (t *InstallationAuthenticator) storeInCache(ctx context.Context) error {
 // It makes a request to the GitHub API to generate a new installation access token for the
 // installation associated with the Authenticator.
 // Returns an error if the request fails.
-func (t *InstallationAuthenticator) Refresh(ctx context.Context, cli httpcli.Doer) error {
+func (t *InstallationAuthenticator) Refresh(ctx context.Context, cli httpcli.Doer) (err error) {
+	//defer func() {
```
ggilmore, line 190: This will be removed soon, this is just a local hack I created to try to revoke tokens without waiting an hour.

cmd/gitserver/internal/executil/executil.go
```diff
@@ -300,6 +300,11 @@ func RunCommandWriteOutput(ctx context.Context, cmd wrexec.Cmder, writer io.Writ
 	}
 
 	err = cmd.Wait()
+	var exitErr *exec.ExitError
+	if errors.As(err, &exitErr) {
+		// Redact the stderr output if we have it.
+		exitErr.Stderr = []byte(redactor(string(exitErr.Stderr)))
```
pjlast, line 306: Whenever I see this []byte -> string -> []byte conversion pattern I wonder if we wouldn't be better off just having []byte implementations of these functions

pjlast (approved):

mmanela (approved):

cmd/gitserver/internal/integration_tests/clone_test.go
```diff
@@ -93,10 +112,10 @@ func TestClone(t *testing.T) {
 
 	// Should have acquired a lock.
 	mockassert.CalledOnce(t, locker.TryAcquireFunc)
-	// Should have reported status. 21 lines is the output git currently produces.
+	// Should have reported status. 24 lines is the output git currently produces.
 	// This number might need to be adjusted over time, but before doing so please
 	// check that the calls actually use the args you would expect them to use.
-	mockassert.CalledN(t, lock.SetStatusFunc, 21)
+	mockassert.CalledN(t, lock.SetStatusFunc, 24)
```
ggilmore, line 118: For future readers, this number increased by 3 due to the 3 extra `tryWrite` calls I added in `gitVCSSyncer.Clone` (the additional logs before / after SetHEAD + an additional log once the fetch finished): https://github.com/sourcegraph/sourcegraph/pull/61179/files#diff-22d28b810252fd7a9c6e94ce2c5bbbb1a61f079b6a6d8407661f1ae28f282250R111

cmd/gitserver/internal/executil/executil.go
eseliger, line 306: since we use `err` not `exitErr` further down, is `exitErr` a reference or a copy of `err`?

PR #61182, opened 2024-03-15
Chore: clean up .gitignore

camdencheek:
Just a quick cleanup while I was taking a look at the .gitignore file to figure out why some files were unintentionally being ignored.

## Test plan

Removing these did not uncover any ignored files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,11 +78,7 @@ eb-bundle.zip
 
 .gtm/
 
-cmd/src/debug
 cmd/gitserver/debug
-cmd/indexer/debug
-
-/.gtm/
 
 # Client
 node_modules
@@ -100,9 +96,6 @@ graphql-operations.ts
 *.module.scss.d.ts
 *.module.css.d.ts
 dll-bundle
-
-# Extensions
-/client/extension-api/dist
 
 # Direnv
 .envrc
@@ -131,10 +124,6 @@ storybook-static/
 sg.config.overwrite.yaml
 # sg Google Cloud API OAuth token
 .sg.token.json
-
-# Generated sitemaps are not committed, they're hosted in a GCS bucket.
-sitemap/
-sitemap_query.db
 
 # Lighthouse CI reports
 .lighthouseci/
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -177,8 +164,8 @@ index.scip
 .aspect/bazelrc/ci.generated.bazelrc
 
 # nix
-result
-result-*
+/result
+/result-*
```
camdencheek, line 170: @Strum355, this was causing the `internal/search/result` directory to be ignored. Assuming that wasn't intentional, I guessed that any `nix` `result` directories will be at the root of the repo. Is that right?

.gitignore
```diff
@@ -132,10 +123,6 @@ sg.config.overwrite.yaml
 # sg Google Cloud API OAuth token
 .sg.token.json
 
-# Generated sitemaps are not committed, they're hosted in a GCS bucket.
-sitemap/
-sitemap_query.db
-
```
camdencheek, line 138: Sitemap tool no longer exists

.gitignore
```diff
@@ -101,9 +95,6 @@ graphql-operations.ts
 *.module.css.d.ts
 dll-bundle
 
-# Extensions
-/client/extension-api/dist
-
```
camdencheek, line 106: Extension API build no longer goes to the `dist` dir

.gitignore
```diff
@@ -78,11 +78,7 @@ eb-bundle.zip
 
 .gtm/
 
-cmd/src/debug
 cmd/gitserver/debug
-cmd/indexer/debug
-
-/.gtm/
```
camdencheek, line 85: Duplicated by a more general rule above.

.gitignore
```diff
@@ -78,11 +78,7 @@ eb-bundle.zip
 
 .gtm/
 
-cmd/src/debug
 cmd/gitserver/debug
-cmd/indexer/debug
```
camdencheek, line 83: No longer exist

.gitignore
camdencheek: cc @jasonhawkharris -- this is why you couldn't find `internal/search/result/match.go` in your editor that one time

eseliger (approved):
Thanks for cleanup :)

.gitignore
Strum355: Thats correct yea, thx for the fix!

PR #61184, opened 2024-03-15
ci: Fetch tag before checkout in backcompat test

varungandhi-src:
Unblock PRs on main

## Test plan

Run CI

### DIFF
--- a/dev/backcompat/bazel-backcompat.sh
+++ b/dev/backcompat/bazel-backcompat.sh
@@ -34,7 +34,7 @@ fi
 echo "--- :git: fetch origin v${tag}"
 # For back-compat tests, the shared fetching code doesn't guarantee that
 # the tag has been fetched (because not all tests require tags), so fetch it here.
-git fetch origin "v${tag}"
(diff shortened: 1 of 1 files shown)


dev/backcompat/bazel-backcompat.sh
```diff
@@ -31,6 +31,11 @@ else
   fi
 fi
 
+echo "--- :git: fetch v${tag}"
+# For back-compat tests, the shared fetching code doesn't guarantee that
+# the tag has been fetched (because not all tests require tags), so fetch it here.
+git fetch "v${tag}"
```
camdencheek, line 37: Do we want to add a `--depth=1` here? I don't think we want to pull the full repo history on a miss.

varungandhi-src: Done, I incorrectly assumed that if you specified a tag name, it would only fetch said tag.

camdencheek (approved):
Approved assuming CI passes

burmudar (approved):

PR #61207, opened 2024-03-18
Cody Gateway: add StarCoder2 models

valerybugakov:
## Context

- Adds support for `starcoder2-7b` and `starcoder2-15b` models hosted by Fireworks dedicated deployments.
- Client-side feature flag https://github.com/sourcegraph/cody/pull/3417
- [Slack discussion](https://sourcegraph.slack.com/archives/C0514L18QHG/p1709046679684259)

## Test plan

<code>curl -vS -X POST http://localhost:9992/v1/completions/fireworks -H 'Authorization: bearer <SGD_TOKEN>' -d '{"stream":false,"max_tokens":50, "model": "starcoder2-7b", "stop_sequences": ["\n\n"], "prompt": "const value = ", "stream":false}' -H 'X-sourcegraph-feature: code_completions'</code>

<code>curl -vS -X POST http://localhost:9992/v1/completions/fireworks -H 'Authorization: bearer <SGD_TOKEN>' -d '{"stream":false,"max_tokens":50, "model": "starcoder2-15b", "stop_sequences": ["\n\n"], "prompt": "const value = ", "stream":false}' -H 'X-sourcegraph-feature: code_completions'</code>


client/web/src/enterprise/site-admin/dotcom/productSubscriptions/ModelBadges.tsx
```diff
@@ -52,6 +52,8 @@ function modelBadgeVariant(model: string, mode: 'completions' | 'embeddings'): '
             case 'fireworks/starcoder':
             // Bespoke alternative models hosted for us by Fireworks. These are also allowed on the
             // Cody Gateway side
+            case 'accounts/sourcegraph/models/starcoder2-15b':
+            case 'accounts/sourcegraph/models/starcoder2-7b':
```
philipp-spiess, line 56:
```suggestion
            case 'sourcegraph/accounts/sourcegraph/models/starcoder2-15b':
            case 'sourcegraph/accounts/sourcegraph/models/starcoder2-7b':
```

but maybe we don't want to add this here yet since this is only for the UI for when we manually edit enterprise licenses which we don't need for now anyhow. And once we start needing it we probably want to use a shorthand syntax like `fireworks/starcoder2` anyways. WDYT?

philipp-spiess: Wait actually further down you only look at `fireworks/starcoder` to get the model support so this is not needed at all

cmd/cody-gateway/internal/httpapi/completions/upstream.go
```diff
@@ -491,7 +491,9 @@ func isAllowedModel(allowedModels []string, model string) bool {
 		}
 
 		// Expand virtual model names
-		if m == "fireworks/starcoder" && (model == "fireworks/"+fireworks.Starcoder7b ||
+		if m == "fireworks/starcoder" && (model == "fireworks/"+fireworks.StarcoderTwo15b ||
+			model == "fireworks/"+fireworks.StarcoderTwo7b ||
+			model == "fireworks/"+fireworks.Starcoder7b ||
```
philipp-spiess, line 496: This will make it instantly usable for all enterprise instances. I think we should not depend on the `firework/starcoder` allowlist for this and instead manually add the two models for PLG only for now. We can later roll this into the same virtual model list if we decide we want to give everyone access to it.

philipp-spiess: I saw that below you already added the individual models to the allowlist for PLG users so this change can be reverted

internal/completions/httpapi/codecompletion.go
```diff
@@ -48,8 +48,12 @@ func allowedCustomModel(model string) string {
 	case "fireworks/starcoder",
 		"fireworks/starcoder-16b",
 		"fireworks/starcoder-7b",
+		"fireworks/starcoder2-15b",
+		"fireworks/starcoder2-7b",
 		"fireworks/" + fireworks.Starcoder16b,
 		"fireworks/" + fireworks.Starcoder7b,
+		"fireworks/" + fireworks.StarcoderTwo15b,
+		"fireworks/" + fireworks.StarcoderTwo7b,
```
philipp-spiess, line 56:
```suggestion
		"fireworks/starcoder2-15b",
		"fireworks/starcoder2-7b",
		"fireworks/" + fireworks.Starcoder16b,
		"fireworks/" + fireworks.Starcoder7b,
```

If you add the virtual model names, let's make sure that clients use those to connect exclusively.

philipp-spiess (approved):

PR #61209, opened 2024-03-18
Expose an embeddings handler that doesn't implement rate-limiting

rafax:
Expose an additional Embeddings handler that does not implement rate-limiting (reduced overhead + we don't have token counts for our models).

This is proof-of-concept only - will be used for performance testing in staging (off by default). 
## Test plan

- tested locally
- old functionality still E2E tested

### DIFF
--- a/cmd/cody-gateway/internal/httpapi/embeddings/handler.go
+++ b/cmd/cody-gateway/internal/httpapi/embeddings/handler.go
@@ -27,6 +27,7 @@ import (
 )
 
 const usageHeaderName = "X-Token-Usage"
+const baseLoggerScope = "embeddingshandler"
 
 func NewHandler(
 	baseLogger log.Logger,
@@ -36,7 +37,7 @@ func NewHandler(
 	mf ModelFactory,
 	allowedModels []string,
 ) http.Handler {
-	baseLogger = baseLogger.Scoped("embeddingshandler")
+	baseLogger = baseLogger.Scoped(baseLoggerScope)
 
 	return featurelimiter.HandleFeature(
 		baseLogger,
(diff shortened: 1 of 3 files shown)


cmd/cody-gateway/internal/httpapi/embeddings/handler.go
```diff
@@ -36,160 +37,176 @@ func NewHandler(
...
-			act := actor.FromContext(r.Context())
-			logger := act.Logger(sgtrace.Logger(r.Context(), baseLogger))
-			// This will never be nil as the rate limiter middleware checks this before.
-			// TODO: Should we read the rate limit from context, and store it in the rate
-			// limiter to make this less dependent on these two logics to remain the same?
-			rateLimit, ok := act.RateLimits[codygateway.FeatureEmbeddings]
-			if !ok {
-				response.JSONError(logger, w, http.StatusInternalServerError, errors.Newf("rate limit for %q not found", string(codygateway.FeatureEmbeddings)))
-				return
+		embeddingsHandler(baseLogger, eventLogger, allowedModels, mf),
```
rafax, line 48: Diff isn't great here - the only real change is extracting existing logic to `embeddingsHandler` so it can be used from both public methods.

cmd/cody-gateway/shared/config/config.go
```diff
@@ -242,6 +243,8 @@ func (c *Config) Load() {
 	c.Attribution.Enabled = c.GetBool("CODY_GATEWAY_ENABLE_ATTRIBUTION_SEARCH", "false", "Whether attribution search endpoint is available.")
 
 	c.Sourcegraph.TritonURL = c.Get("CODY_GATEWAY_SOURCEGRAPH_TRITON_URL", "https://embeddings-triton-direct.sgdev.org/v2/models/ensemble_model/infer", "URL of the Triton server.")
+	c.Sourcegraph.UnlimitedEmbeddings = c.GetBool("CODY_GATEWAY_SOURCEGRAPH_UNLIMITED_EMBEDDINGS", "false", "Enable unlimited embeddings.")
```
chrsmith, line 246: We should had a big, long doc comment explaining what `CODY_GATEWAY_SOURCEGRAPH_UNLIMITED_EMBEDDINGS` controls and when it should be used.

"Enable unlimited embeddings" is clear and all... but why would we want his? What does "unlimited" mean in this context? What things are being disabled, etc.

You call that out in this PR's description (🙏) but I'd suggest you put something similar in the code, too. So it's easier to understand without digging through the git history.

cmd/cody-gateway/internal/httpapi/handler.go
```diff
@@ -193,6 +194,16 @@ func NewHandler(
 			"/embeddings",
 			attributesOpenAIEmbeddings,
 			embeddingsHandler)
+
+		if config.Sourcegraph.UnlimitedEmbeddings {
```
chrsmith, line 198: Having some API endpoints available or not available based on config is kind of confusing. e.g. we cannot really say if it is definitively part of the API or not.

What would you think about _always_ enabling the API endpoint. But having its auth check _also_ check the config setting? e.g. "400: unlimited embeddings are not enabled"

Would that arguably make our API "clearer" in some way?

cmd/cody-gateway/internal/httpapi/embeddings/handler.go
```diff
@@ -36,160 +37,176 @@ func NewHandler(
...
+			response.JSONError(logger, w, resolvedStatusCode, errors.Wrap(err, "failed to marshal response"))
+			return
+		}
+
+		w.Header().Add("Content-Type", "application/json; charset=utf-8")
+		// Write implicitly returns a 200 status code if one isn't set yet
+		if resolvedStatusCode <= 0 {
+			resolvedStatusCode = 200
+		}
+		_, _ = w.Write(data)
```
chrsmith, line 208: If we don't explicitly call `w.WriteHeader` first, I believe when we call `w.Write` it set that header automatically (defaulting to 200).

Is that what we want? Wouldn't it be slightly better to explicitly set the status code? (I'm just confused here, since maybe it's possible `resolvedStatusCode` would be something other than 200, and yet the HTTP response we write would implicitly be 200.
https://pkg.go.dev/net/http#ResponseWriter

cmd/cody-gateway/internal/httpapi/embeddings/handler.go
```diff
@@ -36,160 +37,176 @@ func NewHandler(
...
-					// Record original code if the status error is a custom one
-					if originalCode, ok := statusCodeErr.IsCustom(); ok {
-						upstreamStatusCode = originalCode
-					}
-					return
-				}
+func embeddingsHandler(baseLogger log.Logger, eventLogger events.Logger, allowedModels []string, mf ModelFactory) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+
+		act := actor.FromContext(r.Context())
```
chrsmith, line 77: Nit: We call `r.Context()` several times throughout the lifetime of this function. (Twice in these two lines of code, even.)

Wouldn't it be cleaner and clearer to just introduce it as a local variable at the top of the function? e.g.

```go
ctx := r.Context()
act := actor.FromContext(ctx)
logger := act.Logger(sgtrace.Logger(ctx, baseLogger))
```

etc.?

It's not a major thing, but IMHO it helps a big. Although given the number of moving parts, I'd suggest we break this function up even more. Perhaps having a dedicated function for calculating all of the HTTP headers, or giving the `defer func()` a name.

chrsmith (approved):
Modulo some suggestions for trying to clean up the code or make it easier to read, this all seems fine. Though if this really is just for performance testing, can we update the URL route to make that clearer? e.g.

`/totally-internal-one-off-for-perf-testing/v1/unlimited-embeddings/`?

rafax:
> Modulo some suggestions for trying to clean up the code or make it easier to read, this all seems fine. Though if this really is just for performance testing, can we update the URL route to make that clearer? e.g.
> 
> `/totally-internal-one-off-for-perf-testing/v1/unlimited-embeddings/`?

I can control the suffix easily, but modifying the prefix (up to and including the `v1`) requires creating a new router, which requires changing the existing `register*` methods to accept a router, which is something we don't really want to do?

rafax:
@chrsmith your other suggestions make sense, my plan is to revert this whole commit once I'm done with performance testing, so I'll apply those in a separate PR (post-revert)

Issue #59026, opened 2023-12-15
Offer choice for Go to definition when target definition is ambiguous


varungandhi-src:
In some cases, such as when we have search-based code intel, or when even with precise code intel, there are multiple sources for the same symbol/definition (or the indexer isn't able to distinguish between the two due to some lazy rules such as that of C++ templates), it is valuable to be able to select a definition site during Go to definition.

Here is an example of how GoLand/IntelliJ does it.

![](https://github.com/sourcegraph/sourcegraph/assets/93103176/9ba73866-0085-47de-9181-647acc992be4)

It would be valuable to have a similar UI compared to the current behavior where it picks one result randomly.

peterguy:
@varungandhi-src this is a UX-only (or mostly) change, right? The multiple sources are already available to the UI?

varungandhi-src:
@peterguy in part yes, in part no.

Today, it's possible that the server returns multiple definition results for the same range. https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/cmd/frontend/graphqlbackend/codeintel.codenav.graphql?L94-112

So the client can change the UI to show all of these, and that is fixable without server-side changes.

However, the results provided today are not rich enough to provide a good UX on the client-side. This would be fixed with https://github.com/sourcegraph/sourcegraph/issues/59470

PR #61217, opened 2024-03-18
[DO NOT MERGE] SCIP-based GraphQL API

varungandhi-src:
Part of #62148

TODO:
- [x] Sketch out ref panel like functionality
- [x] Describe how fuzzy navigation might be accommodated.
- [ ] Design new API for hover docs
- [x] Describe how file navigation might be accommodated.
- [x] Describe how build identifiers/filter might be accommodated.
- [x] ~Customizable grouping?~ Let client handle this for now.
- [x] Add implementation notes on how to accommodate search-based code intel in this framework.

## Test plan

n/a

### DIFF
--- a/cmd/frontend/graphqlbackend/codeintel.codenav.graphql
+++ b/cmd/frontend/graphqlbackend/codeintel.codenav.graphql
@@ -275,12 +275,20 @@ type GitBlobLSIFData implements TreeEntryLSIFData {
     snapshot(indexID: ID!): [SnapshotData!]
 }
 
+input LookupSCIPSymbol {
+    name: SymbolNameComparator!
+    provenance: ProvenanceComparator!
+}
+
 """
 In the future, we may want to extend this to allow passing in a suffix or a "symbol pattern".
 """
-input LookupSCIPSymbol {
-    name: String!
-    provenance: Provenance!
+input SymbolNameComparator {
+    eq: String
+}
+
+input ProvenanceComparator {
+    eq: Provenance
 }
 
 type SCIPSymbol {
--- a/cmd/frontend/graphqlbackend/codeintel.codenav.graphql
+++ b/cmd/frontend/graphqlbackend/codeintel.codenav.graphql
@@ -132,6 +132,8 @@ type CodeGraphData {
 
     # EXTENSION: We can add a build identifier here to return different
     # Documents for (say) Linux vs Windows.
+    # EXTENSION: We can add a fileSymbolInfo field here to allow 'Find usages' for the document.
+}
 
 type CodeGraphToolInfo {
     name: String
@@ -150,13 +152,13 @@ type Occurrence {
     symbol: String
     range: Range!
     roles: [SymbolRole!]
-    # We can add diagnostics etc. here in the future if needed.
+    # EXTENSION: We can add diagnostics etc. here in the future if needed.
 }
 
 input OccurrencesFilter {
     rangeFilter: RangeFilter
     rolesFilter: RolesFilter
-    # We can add a DiagnosticsFilter etc. here in the future if needed.
+    # EXTENSION: We can add a DiagnosticsFilter etc. here in the future if needed.
 }
 
 input RangeFilter {
--- a/cmd/frontend/graphqlbackend/codeintel.codenav.graphql
+++ b/cmd/frontend/graphqlbackend/codeintel.codenav.graphql
@@ -247,19 +247,6 @@ type GitBlobLSIFData implements TreeEntryLSIFData {
     ): Hover
 
     """
-    Either one of 'symbol' or 'range' must be provided, but this isn't enforced at the GraphQL
-    layer due to the lack of support for input unions.
-    https://github.com/graphql/graphql-wg/blob/main/rfcs/InputUnion.md
-    """
-    usages(
-        symbol: LookupSCIPSymbol,
-        range: LookupRange,
-        filter: UsagesFilter,
-        first: Int,
-        after: String,
-    ): UsageConnection!
-
-    """
     Code diagnostics provided through LSIF.
     """
     diagnostics(first: Int): DiagnosticConnection!
@@ -273,6 +260,19 @@ type GitBlobLSIFData implements TreeEntryLSIFData {
     SCIP snapshot data (similar to the additional information from the `scip snapshot` command) for each SCIP Occurrence.
     """
     snapshot(indexID: ID!): [SnapshotData!]
+}
+
+extend type Query {
+    usagesForSymbol(
+        symbol: LookupSCIPSymbol,
+        """
+        Non-optional due to implementation limitations
+        """
+        range: LookupRange!,
+        filter: UsagesFilter,
+        first: Int,
+        after: String,
+    ): UsageConnection!
 }
 
 input LookupSCIPSymbol {
@@ -297,8 +297,24 @@ type SCIPSymbol {
 }
 
 input LookupRange {
-    start: LookupPosition!
-    end: LookupPosition!
+    """
+    Defaults to instance-wide search if the repo is not specified.
+
+    This field is non-optional due to implementation limitations.
+    """
+    repo: String!
+    """
+    Defaults to HEAD of the default branch if not specified.
+    """
+    revision: String
+    """
+    Defaults to repo-wide search if the path is not specified.
+
+    This field is non-optional due to implementation limitations.
+    """
+    path: String!
+    start: LookupPosition
+    end: LookupPosition
 }
 
 input LookupPosition {
@@ -310,7 +326,6 @@ input UsagesFilter {
     and: [UsagesFilter!]
     or: [UsagesFilter!]
     not: UsagesFilter
-    # TODO: Can we be flexible here and allow patterns while still maintaining fast queries.
     repository: RepositoryFilter
     path: PathFilter
     kind: SymbolUsageKind
--- a/cmd/frontend/graphqlbackend/codeintel.codenav.graphql
+++ b/cmd/frontend/graphqlbackend/codeintel.codenav.graphql
@@ -563,7 +563,31 @@ An empty filter allows all kinds of usages for all paths in all repositories.
 However, if the symbol used for lookup is a file-local symbol or a
 repository-local symbol, then usages will automatically be limited to the
 same file or same repository respectively.
+
+Filters are combined monoidally, i.e. all filters are applied. For example,
+
+```
+{and: [u1, u2], or: [u3, u4], not: u5}
+```
+
+would be equivalent to:
+
+```
+{and: [{and: [u1, u2]}, {or: [u3, u4]}, {not: u5}]}
+```
 """
+# Design Rationale:
+#
+# 1. The ref panel did separate queries for 'this repo' vs 'all other repos'
+#    https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/client/web/src/enterprise/codeintel/searchBased.ts?L105:13-105:23#tab=references
+#    That requires `NOT` (at least for a `repo`).
+# 2. The ref panel has a way to specify paths for filtering. This requires AND.
+#    AND will not be part of the first implementation pass. However,
+#    it would allow fetching less data instead of post-filtering client-side.
+# 3. Currently, the ref panel has a setting for only showing precise results.
+#    Once syntactic is added, if we want to allow the user to specify either
+#    syntactic or precise (but not search-based), we'd need OR.
+#    OR will also not be needed for the first implementation pass.
 input UsagesFilter {
     and: [UsagesFilter!]
     or: [UsagesFilter!]
--- a/cmd/frontend/graphqlbackend/codeintel.codenav.graphql
+++ b/cmd/frontend/graphqlbackend/codeintel.codenav.graphql
@@ -49,11 +49,18 @@ extend type GitBlob {
     symbolInfo(line: Int!, character: Int!): SymbolInfo
 
     """
+    Return the code graph data associated with this blob.
+
+    If there are multiple tools (i.e. name and version pairs) which
+    have uploaded precise indexes for this blob, then this API will
+    return multiple results even if
+    filter == { provenance: { equals: Precise } }.
+
     If filter.provenance? is not provided, then the API will go through
     each provenance one by one in the order Precise -> Syntactic -> SearchBased
     and stop when some data is available.
     """
-    codeGraphData(filter: CodeGraphDataFilter): CodeGraphData
+    codeGraphData(filter: CodeGraphDataFilter): [CodeGraphData!]
 }
 
 input CodeGraphDataFilter {
--- a/cmd/frontend/graphqlbackend/codeintel.codenav.graphql
+++ b/cmd/frontend/graphqlbackend/codeintel.codenav.graphql
@@ -76,16 +76,17 @@ Invariants:
 # 1. Request code graph data, specifying 'filter' if needed.
 #    Either the scipDocument or occurrences is fine.
 # 2. When a hover is triggered, attempt to look up occurrence in document or list
-#    or some other lookup structure constructed from those.
-# 3a. If one matching occurrence is found, use symbolName + range to call
-#     usagesForSymbol
+#    of occurrences (or some other lookup structure constructed from those)
+#    based on the source range.
+# 3a. If one matching occurrence is found, use Occurrence.symbol and range to call
+#     usagesForSymbol.
 # 3b. If zero matching occurrences are found, use only range to call usagesForSymbol.
 # 3c. If more than one matching occurrence is found, use some more complex logic
 #     to offer a symbol picker.
-#     - Problem: For a symbol picker, ideally you'd use SymbolInformation.display_name,
-#       which is stored on SymbolInformation. However, display_name is not supported
-#       today by most SCIP indexers. And we don't have any API for getting the detailed
-#       SCIP SymbolInformation for a bunch of occurrences.
+#     - Problem: For a symbol picker, ideally you'd use SymbolInformation.display_name
+#       in SCIP. However, display_name is not supported today by most SCIP indexers.
+#       And we don't have any API for getting the detailed SCIP SymbolInformation
+#       for a bunch of occurrences.
 type CodeGraphData {
     scipDocument: AssociatedSCIPDocument
 
@@ -95,7 +96,7 @@ type CodeGraphData {
     Invariant:
     forall sym in symbols.
         exists occ in occurrences such that
-            sym.name == occ.symbolName && occ.roles.contains(Definition)
+            sym.name == occ.symbol && occ.roles.contains(Definition)
     """
     # TODO: Should this be paginated to gracefully handle large generated files?
     symbols: [SymbolInformation!]
@@ -501,6 +502,11 @@ input LookupPosition {
     """
     character: Int!
 }
+
+# NOTE(id: filter-vs-comparator-terminology)
+#
+# 'Filter' is used for complex objects whereas 'Comparator' is used
+# for primitive objects.
 
 """
 An empty filter allows all kinds of usages for all paths in all repositories.
--- a/cmd/frontend/graphqlbackend/codeintel.codenav.graphql
+++ b/cmd/frontend/graphqlbackend/codeintel.codenav.graphql
@@ -108,7 +108,14 @@ type CodeGraphData {
 }
 
 type Occurrence {
-    symbolName: String
+    """
+    Symbol name using syntax specified by the SCIP schema.
+    https://github.com/sourcegraph/scip/blob/main/scip.proto#L147-L188
+
+    This value will generally be used in conjunction with
+    the `usagesBySymbol` API.
+    """
+    symbol: String
     range: Range!
     roles: [SymbolRole!]
     # We can add diagnostics etc. here in the future if needed.
--- a/cmd/frontend/graphqlbackend/codeintel.codenav.graphql
+++ b/cmd/frontend/graphqlbackend/codeintel.codenav.graphql
@@ -197,8 +197,6 @@ type GitBlobLSIFData implements TreeEntryLSIFData {
 
     """
     A list of definitions of the symbol under the given document position.
-
-    DEPRECATED: Use usagesForSymbol with a Definition filter instead.
     """
     definitions(
         """
@@ -215,12 +213,10 @@ type GitBlobLSIFData implements TreeEntryLSIFData {
         When specified, it filters references by filename.
         """
         filter: String
-    ): LocationConnection!
+    ): LocationConnection! @deprecated(reason: "Superseded by usagesForSymbol; use the Definition filter")
 
     """
     A list of references of the symbol under the given document position.
-
-    DEPRECATED: Use usagesForSymbol with a Reference filter instead.
     """
     references(
         """
@@ -252,12 +248,10 @@ type GitBlobLSIFData implements TreeEntryLSIFData {
         When specified, it filters references by filename.
         """
         filter: String
-    ): LocationConnection!
+    ): LocationConnection! @deprecated(reason: "Superseded by usagesForSymbol; use the Reference filter")
 
     """
     A list of implementations of the symbol under the given document position.
-
-    DEPRECATED: Use usagesForSymbol with Implementations filter instead.
     """
     implementations(
         """
@@ -289,12 +283,10 @@ type GitBlobLSIFData implements TreeEntryLSIFData {
         When specified, it filters implementation by filename.
         """
         filter: String
-    ): LocationConnection!
+    ): LocationConnection! @deprecated(reason: "Superseded by usagesForSymbol; use the Implemention filter")
 
     """
     A list of prototypes of the symbol under the given document position.
-
-    DEPRECATED: Use usagesForSymbol with a Supers filter instead.
     """
     prototypes(
         """
@@ -326,7 +318,7 @@ type GitBlobLSIFData implements TreeEntryLSIFData {
         When specified, it filters prototypes by filename.
         """
         filter: String
-    ): LocationConnection!
+    ): LocationConnection! @deprecated(reason: "Superseded by usagesForSymbol; use the Super filter")
 
     """
     The hover result of the symbol under the given document position.
--- a/cmd/frontend/graphqlbackend/codeintel.codenav.graphql
+++ b/cmd/frontend/graphqlbackend/codeintel.codenav.graphql
@@ -120,6 +120,8 @@ type CodeGraphData {
     forall sym in symbols.
         exists occ in occurrences such that
             sym.name == occ.symbol && occ.roles.contains(Definition)
+
+    These map 1-1 with the symbols in scipDocument.data
     """
     symbols: SymbolInformationConnection
 
--- a/cmd/frontend/graphqlbackend/codeintel.codenav.graphql
+++ b/cmd/frontend/graphqlbackend/codeintel.codenav.graphql
@@ -368,13 +368,14 @@ extend type Query {
         """
         Symbol to perform the lookup for.
 
-        If provided, then the start and end position in 'range' are optional.
+        If provided, then the start and end position in `range` are optional.
 
-        If not provided, then the start and end position in 'range' must be set.
-        In that case, the results will be merged across all semantic symbols
-        in the source range specified by [start, end]
+        If not provided, and if multiple symbols are detected at the same range,
+        the combined results for all symbols will be returned.
+        In that case, `Usage.symbol.name` can be used to group results.
         """
         symbol: SymbolComparator,
+
         """
         Information about an existing source range for a usage of the symbol.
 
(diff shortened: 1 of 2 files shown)


cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -247,6 +269,57 @@ type GitBlobLSIFData implements TreeEntryLSIFData {
...
 
+union SCIPSymbolOrRange = SCIPSymbol | Range
+
+"""
+In the future, we may want to extend this to allow passing in a suffix or a "symbol pattern".
+"""
+type SCIPSymbol {
+    name: String!
+    provenance: Provenance!
+}
```
varungandhi-src, line 280: TODO: Should this implement `WithProvenance` with a fine-grained data source?

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -105,6 +105,21 @@ type GitBlobLSIFData implements TreeEntryLSIFData {
...
+        """
+        after: String
+
+        """
+        When specified, indicates that this request should be paginated and
+        the first N results (relative to the cursor) should be returned. i.e.
+        how many results to return per page.
+        """
+        first: Int
+
```
varungandhi-src, line 122: TODO: Technically, this is fixing a hole in the old API, but maybe we don't care about it, since this will be superseded by usages.

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -247,6 +269,57 @@ type GitBlobLSIFData implements TreeEntryLSIFData {
...
+type SCIPSymbol {
+    name: String!
+    provenance: Provenance!
+}
+
+type SymbolUsagesFilter {
+    # TODO: There needs to be a way to say 'current repo only' or any repos.
+    # Allow file filter
+    # Allow kind filter
+}
```
varungandhi-src, line 286: TODO: Document connections with Chromium code search, IntelliJ, and Linear's filtering API.

cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -6172,17 +6172,55 @@ type HighlightedFile {
...
+    """
+    provenance: Provenance!
+    """
+    Opaque fine-grained information describing the data source.
+
+    Provided only for debugging.
+
+    This field should be ignored when checking structural equality.
+    """
+    dataSource: String!
```
varungandhi-src, line 6207: Anton's note: We should make this optional if this is only for debugging.

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -247,6 +275,77 @@ type GitBlobLSIFData implements TreeEntryLSIFData {
...
+    character: Int!
+}
+
+input UsagesFilter {
+    and: [UsagesFilter!]
+    or: [UsagesFilter!]
+    not: UsagesFilter
+    # TODO: Can we be flexible here and allow patterns while still maintaining fast queries.
+    repository: String
+    path: String
```
varungandhi-src, line 307: Anton's note: We should make these objects for more flexibility, even if it is just a string for now.

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -231,6 +246,19 @@ type GitBlobLSIFData implements TreeEntryLSIFData {
...
+    layer due to the lack of support for input unions.
+    https://github.com/graphql/graphql-wg/blob/main/rfcs/InputUnion.md
+    """
+    usages(
+        symbol: LookupSCIPSymbol,
+        range: LookupRange,
+        filter: UsagesFilter,
+        first: Int,
+        after: String,
+    ): UsageConnection!
```
varungandhi-src, line 260: Anton's Q: Why not split this into RangeUsageConnection and SymbolUsageConnection? This will cause more complexity in the outputs because of input dependence.

Varun's A: The output needs to have some complexity anyways because of needing to support search-based/syntactic/precise and some information will not be present depending on the data source. (Anton's idea: In GraphQL, you'd typically use union types for results so that caller can check the type and cast.)

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -47,6 +47,64 @@ extend type GitBlob {
     Experimental: This API is likely to change in the future.
     """
     symbolInfo(line: Int!, character: Int!): SymbolInfo
+
+    codeGraphData: CodeGraphData
```
varungandhi-src, line 51: Should we pass in a `ProvenanceComparator` as a parameter here? Should we pass anything else?

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -47,6 +47,64 @@ extend type GitBlob {
...
+}
+
+"""
+Invariants:
+1. forall sym in symbols.
+     (sym.provenance, sym.dataSource) == (scipDocument.provenance, scipDocument.dataSource)
+"""
+type CodeGraphData {
```
varungandhi-src, line 59: This invariant is a bit annoying; should we remove the duplication the `(provenance, dataSource)` pair of fields? Right now, we have Provenance on `SymbolInformation` so that you can check it after invoking `usagesBySymbol`

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -47,6 +47,64 @@ extend type GitBlob {
...
+type Occurrence {
+    symbolName: String
+    range: Range!
+    roles: [SymbolRole!]
+    # We can add diagnostics etc. here in the future if needed.
+}
+
+input OccurrencesFilter {
+    rangeFilter: RangeFilter
+    rolesFilter: RolesFilter
```
varungandhi-src, line 84: TODO: Add documentation here describing how this can be used for the symbols side bar so that the symbols sidebar is consistent with the navigation in the file. E.g. you can have a precise symbols sidebar if you have precise code intel instead of having to use `type:symbol` search in the file.

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -113,17 +173,19 @@ type GitBlobLSIFData implements TreeEntryLSIFData {
...
         The line on which the symbol occurs (zero-based, inclusive).
         """
-        line: Int!
+        line: Int
 
         """
         The character (not byte) of the start line on which the symbol occurs (zero-based, inclusive).
         """
-        character: Int!
+        character: Int
```
varungandhi-src, line 188: TODO: Revert these changes, we don't care about this API anymore

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -247,6 +313,216 @@ type GitBlobLSIFData implements TreeEntryLSIFData {
...
+    kind: SymbolUsageKind
+}
+
+input RepositoryFilter {
+    name: StringComparator!
+}
+
+input PathFilter {
+    name: StringComparator!
+}
```
varungandhi-src, line 476: TODO: Should we remove the `!` here?

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -247,6 +313,216 @@ type GitBlobLSIFData implements TreeEntryLSIFData {
...
+
+type UsageRange {
+    repo: String!
+    revision: String!
+    range: Range!
+}
+
+type Usage {
+    symbol: SymbolInformation!
```
varungandhi-src, line 501: TODO: Think through the non-null cases more carefully, especially for `UsageConnection`

cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -6172,17 +6172,83 @@ type HighlightedFile {
     """
     html: String!
     """
-    Base64 encoded JSON payload of LSIF Typed with syntax highlighting data.
+    DEPRECATED: Base64 encoded JSON payload of SCIP with syntax highlighting data.
+    Use document instead.
```
mmanela, line 6176: Should `lsif: String!` become `lsif: String` so that we can omit it when `document` is present?

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -47,6 +47,103 @@ extend type GitBlob {
...
+
+    """
+    Occurrences are guaranteed to be sorted by range.
+    """
+    # TODO: Should this be paginated to gracefully handle large generated files?
+    occurrences(filter: OccurrencesFilter): [Occurrence!]
+}
+
+type Occurrence {
+    symbolName: String
```
olafurpg, line 111: "symbol name" is not defined in SCIP and could be confused with `SymbolInformation.displayName` (which is intentionally not named `.name`)
```suggestion
    symbol: String
```

cmd/frontend/graphqlbackend/schema.graphql
varungandhi-src: No, that would be a breaking change, and we're not making breaking changes here for now.

cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -6172,17 +6172,83 @@ type HighlightedFile {
     """
     html: String!
     """
-    Base64 encoded JSON payload of LSIF Typed with syntax highlighting data.
+    DEPRECATED: Base64 encoded JSON payload of SCIP with syntax highlighting data.
```
varungandhi-src, line 6175: Use `@deprecated` attribute instead of `DEPRECATED`.

cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -6172,17 +6172,83 @@ type HighlightedFile {
...
+type AssociatedSCIPDocument implements WithProvenance {
+    """
+    Base64 encoded SCIP payload.
+
+    After decoding, the resulting byte buffer can be parsed with
+    pre-generated bindings in the sourcegraph/scip repository or
+    manually generating bindings using
+    https://github.com/sourcegraph/scip/blob/main/scip.proto
+    """
+    data: String!
```
camdencheek, line 6237: I wonder: would it be more natural to use proto's canonical JSON encoding for a GraphQL API?

camdencheek: I guess the downside is then there is temptation to not use the SCIP bindings and just use the deserialized JSON object directly. Also it doesn't look like the bindings support JSON serialization. So nevermind, ignore me.

varungandhi-src: 1. We don't want to update our GraphQL API for every change to SCIP, because most of our indexers may not adopt the new features right away. Some features also add a significant amount of complexity (e.g. this one being proposed which exposes full type structures https://github.com/sourcegraph/scip/pull/231), and I don't think we should expose that in our GraphQL API. Exposing the raw SCIP allows a customer to upload updated indexes using the new SCIP feature and implement custom functionality.
2. Protobuf allows a client to parse the data in a streaming manner, which can allow for optimizations like interning-while-parsing (e.g. for symbol names), as well as just skipping needless input, without allocating a lot of memory.

Some of the important Document data is already exposed by the `symbols` and `occurrences` fields in the API today. So 'light' clients can make use of it directly. 'Heavy' clients can take on a bit more complexity by parsing the Protobuf for full access to all the details.

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -247,6 +352,216 @@ type GitBlobLSIFData implements TreeEntryLSIFData {
...
+same file or same repository respectively.
+"""
+input UsagesFilter {
+    and: [UsagesFilter!]
+    or: [UsagesFilter!]
+    not: UsagesFilter
+    repository: RepositoryFilter
+    path: PathFilter
+    kind: SymbolUsageKind
+}
```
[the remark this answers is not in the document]
varungandhi-src: Yes, good point, added that in.

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
varungandhi-src: Good point, I've replaced `symbolName` with `symbol`.

cmd/frontend/graphqlbackend/schema.graphql
camdencheek: Ah, makes sense. Thanks!

cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -6172,17 +6172,83 @@ type HighlightedFile {
...
+    document: AssociatedSCIPDocument!
+    """
     A list of the desired line ranges. Each list is a list of lines, where each element is an HTML
     table row '<tr>...</tr>' string. This is useful if you only need to display specific subsets of
     the file.
     """
     lineRanges(ranges: [HighlightLineRange!]!): [[String!]!]!
 }
 
+enum Provenance {
```
eseliger, line 6191: Tip: Provenance is very generic, and GraphQL doesn't have namespaces so I recommend to go with a slightly more specific term here, like `SCIPLookupProvenance` or something like that.

cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -6172,17 +6172,83 @@ type HighlightedFile {
...
+    Opaque fine-grained information describing the data source.
+
+    Provided only for debugging.
+
+    This field should be ignored when checking structural equality.
+    """
+    dataSource: String
+}
+
+type AssociatedSCIPDocument implements WithProvenance {
```
eseliger, line 6228: The `AssociatedSCIPDocument` is returned from the above API directly, not the interface so I'm curious what the value of having an interface here is?

cmd/frontend/graphqlbackend/schema.graphql
eseliger: By shoving this into a string field, you lose the ability to do a subselect, which kind of defeats GraphQL. That said, if we really need a proto representation here, I don't think that is blocking.

However, from an APIs perspective, it feels a little cleaner to return JSON.

If having to support streaming decoding here, I think I have another concern: 
Is this gonna be a huge string sent over? Base64 encoding adds another bit of overhead to the size of the proto, and you need to unmarshal the whole JSON GQL response to then grab this string to then streaming decode it. 
Our backend has to build up the whole response in memory, and the clients load it into memory entirely. Is that a concern?

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -47,6 +47,111 @@ extend type GitBlob {
     Experimental: This API is likely to change in the future.
     """
     symbolInfo(line: Int!, character: Int!): SymbolInfo
+
+    """
+    If filter.provenance? is not provided, then the API will go through
+    each provenance one by one in the order Precise -> Syntactic -> SearchBased
+    and stop when some data is available.
```
eseliger, line 54: I think this could be moved to filter.provenance, and here we could instead document what this API does more generally.

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -47,6 +47,111 @@ extend type GitBlob {
...
+input CodeGraphDataFilter {
+    provenance: ProvenanceComparator
+}
+
+"""
+Invariants:
+1. forall sym in symbols.
+     (sym.provenance, sym.dataSource) == (scipDocument.provenance, scipDocument.dataSource)
+"""
+# Implementation notes:
```
eseliger, line 68: Nice usage of `#` vs `"""`

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -47,6 +47,111 @@ extend type GitBlob {
...
+    symbols: [SymbolInformation!]
+
+    """
+    Occurrences are guaranteed to be sorted by range.
+    """
+    # TODO: Should this be paginated to gracefully handle large generated files?
+    occurrences(filter: OccurrencesFilter): [Occurrence!]
+}
+
+type Occurrence {
```
eseliger, line 111: Not as further down: No namespaces means this is a global type, Occurrence is quite generic. Maybe consider prefixing it with something.

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -47,6 +47,111 @@ extend type GitBlob {
...
+
+    """
+    List of symbols with definitions in this document.
+
+    Invariant:
+    forall sym in symbols.
+        exists occ in occurrences such that
+            sym.name == occ.symbol && occ.roles.contains(Definition)
+    """
+    # TODO: Should this be paginated to gracefully handle large generated files?
```
eseliger, line 101: Probably, adding pagination after the fact is a breaking change. You could use a connection type today, but always pass a first: 100000 for a simple v0.

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -247,6 +352,256 @@ type GitBlobLSIFData implements TreeEntryLSIFData {
...
+    """
+    blob: GitBlob!
+
+    """
+    Instead of blob { content }, allows accessing a sub-span of the content
+    using relative coordinates from the range of this usage. If linesBefore
+    or linesAfter is negative or exceeds the number of available lines,
+    the value is interpreted as until the start/end of the file.
+    """
+    surroundingBlobContent(surroundingLines: SurroundingLines = {linesBefore: 0, linesAfter: 0}): String!
```
eseliger, line 597: `0` meaning "no value given" is a very go-ey thing, I think you don't need a default value here, since `surroundingLines: SurroundingLines` already indicates nullable and the backend can then decide how to represent that in the given programming language

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -247,6 +352,256 @@ type GitBlobLSIFData implements TreeEntryLSIFData {
...
+"""
+input UsagesFilter {
+    and: [UsagesFilter!]
+    or: [UsagesFilter!]
+    not: UsagesFilter
+    repository: RepositoryFilter
+    path: PathFilter
+    kind: SymbolUsageKindComparator
+    provenance: ProvenanceComparator
+}
```
eseliger, line 526: wow! sophisticated API 😬 

Curious, how are `and`, `or`, and `not` combined if all are set?

cmd/frontend/graphqlbackend/schema.graphql
mmanela: `WithProvenance` is also used on `SymbolInformation`

cmd/frontend/graphqlbackend/schema.graphql
mmanela: The thing here is that we want clients to use the SCIP proto-bindings for interacting with this field

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -47,6 +47,111 @@ extend type GitBlob {
...
+#    Either the scipDocument or occurrences is fine.
+# 2. When a hover is triggered, attempt to look up occurrence in document or list
+#    of occurrences (or some other lookup structure constructed from those)
+#    based on the source range.
+# 3a. If one matching occurrence is found, use Occurrence.symbol and range to call
+#     usagesForSymbol.
+# 3b. If zero matching occurrences are found, use only range to call usagesForSymbol.
+# 3c. If more than one matching occurrence is found, use some more complex logic
+#     to offer a symbol picker.
+#     - Problem: For a symbol picker, ideally you'd use SymbolInformation.display_name
```
mmanela, line 86: How does the current reference panel handle this situation?

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -247,6 +352,256 @@ type GitBlobLSIFData implements TreeEntryLSIFData {
...
+
+# Implementation note: In the future, we may want to extend this
+# to allow passing in a suffix or a "symbol pattern".
+# See https://github.com/sourcegraph/sourcegraph/issues/59957
+input SymbolNameComparator {
+    equals: String
+}
+
+input ProvenanceComparator {
+    equals: Provenance
```
mmanela, line 419: Would the idea we add other operators here like substr later on? 

also @eseliger  is this pattern idiomatic for graphql? Or can we achieve this differently

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -47,6 +47,111 @@ extend type GitBlob {
...
+    """
+    List of symbols with definitions in this document.
+
+    Invariant:
+    forall sym in symbols.
+        exists occ in occurrences such that
+            sym.name == occ.symbol && occ.roles.contains(Definition)
+    """
+    # TODO: Should this be paginated to gracefully handle large generated files?
+    symbols: [SymbolInformation!]
```
mmanela, line 102: Would all these symbols also  be in the scipDocument, or will this also contain info from other sources?

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -47,6 +47,111 @@ extend type GitBlob {
...
+
+type Occurrence {
+    """
+    Symbol name using syntax specified by the SCIP schema.
+    https://github.com/sourcegraph/scip/blob/main/scip.proto#L147-L188
+
+    This value will generally be used in conjunction with
+    the `usagesBySymbol` API.
+    """
+    symbol: String
```
mmanela, line 119: Will we use `Occurrence` type to also return items from search based references? How would the descriptor look like if that is the case?

cmd/frontend/graphqlbackend/schema.graphql
varungandhi-src: I'm not sure if interfaces are conventionally used this way, but the way I was using it here was to mark different types as having some common fields. E.g. in the future, if we want to add more provenance information, it might make sense to add that in the interface, so that you get a GraphQL error for all types which don't support that field.

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
varungandhi-src: There is an implicit ordering in fidelity: `SearchBased < Syntactic < Precise`. In the future, we might want to have an extra field `fidelityEqualToOrGreaterThan`. For example, right now the `CodeGraphDataFilter` allows only passing a single `Provenance`, it doesn't have a richer structure with `OR` unlike the `UsagesFilter`.

This pattern is modeled after comparators in Linear's GraphQL API. 

https://studio.apollographql.com/public/Linear-API/variant/current/schema/reference/inputs/StringComparator

`substr` doesn't make much sense on enums IMO.

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
varungandhi-src: I'll try to describe my rationale for this machinery:

1. The ref panel currently has queries to search all repos other than the current repo ([code](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/client/web/src/enterprise/codeintel/searchBased.ts?L105:13-105:23#tab=references)). That requires `NOT` (at least for a `repo`).
2. The ref panel has a way to specify paths for filtering. So while `AND` is technically not needed for the first pass (and I won't implement it initially), it would allow fetching less data instead of post-filtering client-side.
3. Currently, the ref panel has a setting for only showing precise results. Once syntactic is added, if we want to allow the user to specify either syntactic or precise (but not search-based), we'd need OR. OR will also not be needed for the first implementation pass.

So I've pre-emptively sketched it out here; all of this won't be in the first or perhaps even second version of the API, except for NOT.

OK, so interaction. GraphQL doesn't really have a way to model sym types AFAICT. What I really want to write here is something like:

```
input enum UsagesFilter {
    case atom(repository: RepositoryFilter, ...) @default
    case and([UsagesFilter!])
    case or([UsagesFilter!])
    case not(UsagesFilter)
}
```

Given the inability to do that, I've just flattened that all here, similar to how the Linear API does it. https://studio.apollographql.com/public/Linear-API/variant/current/schema/reference/inputs/IssueFilter

Given that this is meant as a filter, I think the semantics should be monoidal combination, i.e. apply all the filters.

```
UsageFilter{and: [u1, u2], or: [u3, u4], not: u5}

-> (desugar)

UsageFilter{
  and: [
    UsageFilter{and: [u1, u2]},
    UsageFilter{or: [u3, u4]},
    UsageFilter{not: u5}}
  ]
} 
```

eseliger: Sounds good and makes sense to me! Would be great to have some form of this comment in the doc strings maybe :)

cmd/frontend/graphqlbackend/schema.graphql
varungandhi-src: Renamed to `CodeGraphDataProvenance`

cmd/frontend/graphqlbackend/schema.graphql
varungandhi-src: I've removed this interface for now; we can add it retroactively if needed, thanks to structural typing.

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
varungandhi-src: Renamed to `SCIPOccurrence`

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
varungandhi-src: Added pagination for `CodeGraphData.symbols`, `CodeGraphData.occurrences` as well as `AssociatedSCIPDocument.data`

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
varungandhi-src: It does not. https://github.com/sourcegraph/sourcegraph/issues/59026

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
varungandhi-src: They should map 1-1 with symbols in the `scipDocument` above; updated the comment here.

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
varungandhi-src: Based on the initial plan, no. The search-based code 'Find usages' will go through this:

```
# 3b. If zero matching occurrences are found, use only range to call usagesForSymbol.
```

Even if we do change that, it shouldn't matter to the client so long as the client is implemented using the described algorithm without making assumptions of which provenances may or may not have occurrences.

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
varungandhi-src: I've added the equivalence part in the docstring and the design rationale as a `#` comment.

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
varungandhi-src: Here, `0` doesn't mean "no value given", it literally means `0` because that makes sense as a default -- only the single line from the blob will be returned in the default case. I wrote the 0 explicitly here as it serves as documentation for an API client for what the behavior will be if they omit this parameter. Another potential default would be `INT32_MAX` (i.e. get the full blob contents), but we're deliberately picking `0` here, not `INT32_MAX`.

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
varungandhi-src: Moved this comment.

varungandhi-src:
Thanks everyone for the feedback and questions @eseliger, @mmanela, @camdencheek. I think I've addressed most of other people's comments. Overall, API shape-wise, I think we're close to where we want to be.

The main TODO left is the API for the hover doc. If we think about it broadly, I think we really want to think about "get me information about this symbol" (or if a symbol is not available, such as in the case of search-based, then a range), which may include:
1. Markdown formatted documentation
2. A signature, maybe with highlighting information
3. Relationships to other symbols etc.

The first field is already part of `SymbolInformation`. The second and third would also make sense on `SymbolInformation`. The API for getting a `SymbolInformation` based on a symbol/range would be an additive change, so I don't think that's too pressing to discuss right away.

I'm going to close this PR for now. I will tag people individually for subsequent PRs on an as-needed basis, especially if there is a big divergence from what has been discussed here.

Issue #61110, opened 2024-03-13
Make cloud-ephemeral runtype push images to private registry


burmudar:
During a cloud-ephemeral run push images to the private-registry for cloud ephemeral deploments 

Depends on: 
- https://github.com/sourcegraph/sourcegraph/issues/61109
- https://github.com/sourcegraph/sourcegraph/issues/61208

burmudar:
Fixed in https://github.com/sourcegraph/sourcegraph/pull/61237

PR #61237, opened 2024-03-18
ci: add cloud ephemeral runtype to Cloud Ephemeral deployments

### LINKED ISSUE #61208: ci: Create cloud-ephemeral runtype
Create a runtype `CloudEphemeral` which has a branch matcher that matches on `cloud-ephemeral/<branch>`.

ci runs of `CloudEphemeral` is practically the same `MainDryRun` with the only change being that it pushes to the private registry `us-central1-docker.pkg.dev/sourcegraph-ci/cloud-ephemeral`

burmudar:
Closes #61208,#61110

This adds the Cloud Ephemeral runtype which is a runtype which:
- Matches on branches which starts with `cloud-ephemeral/`
- Does everything a main-dry-run does except pushes images to the registry `sourcegraph-ci/cloud-ephemeral`

## Test plan
* CI
* [Test run](https://buildkite.com/sourcegraph/sourcegraph/builds/265572)

### DIFF
--- a/dev/ci/push_all.sh
+++ b/dev/ci/push_all.sh
@@ -114,6 +114,7 @@ elif [[ "$BUILDKITE_BRANCH" =~ ^main-dry-run/.*  ]]; then
   prod_tags+=("insiders")
   push_prod=false
 elif [[ "$BUILDKITE_BRANCH" =~ ^cloud-ephemeral/.* ]]; then
+  # Cloud Ephemeral images need a proper semver version
   dev_tags+=("insiders" "${PUSH_VERSION}")
   prod_tags+=("insiders")
   push_prod=false
--- a/dev/ci/push_all.sh
+++ b/dev/ci/push_all.sh
@@ -113,6 +113,11 @@ elif [[ "$BUILDKITE_BRANCH" =~ ^main-dry-run/.*  ]]; then
   dev_tags+=("insiders")
   prod_tags+=("insiders")
   push_prod=false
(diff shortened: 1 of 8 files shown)


jhchabran (approved):

dev/ci/push_all.sh
```diff
@@ -113,6 +113,12 @@ elif [[ "$BUILDKITE_BRANCH" =~ ^main-dry-run/.*  ]]; then
   dev_tags+=("insiders")
   prod_tags+=("insiders")
   push_prod=false
+elif [[ "$BUILDKITE_BRANCH" =~ ^cloud-ephemeral/.* ]]; then
+  # Cloud Ephemeral images need a proper semver version
+  dev_tags+=("insiders" "${PUSH_VERSION}")
```
burmudar, line 118: Thinking of just pushing `$PUSH_VERSION` here instead of all the other tags 🤔

filiphaftek (approved):
Changes are nice!

dev/ci/push_all.sh
filiphaftek: @burmudar Do we have to enforce this naming convention `^cloud-ephemeral/`? Initially I though there is no requirement for branch name.

burmudar: I can change it so that we use an environment variable instead 🤔 that _might_ actually be much nicer since then we don't have to push a special branch. I'll try it out and see how it is

PR #61247, opened 2024-03-18
Search: expect trailing newlines in chunk matches

camdencheek:
This:
1) Bumps Zoekt to include https://github.com/sourcegraph/zoekt/pull/747
2) updates all the consumers of our APIs to trim the trailing newline before splitting
3) updates searcher to also include trailing newlines in chunk matches

## Test plan

Manual testing in the UI to ensure everything looks right (in particular, no extra blank lines). Verified every changed test output.


cmd/searcher/internal/search/hybrid_test.go
```diff
@@ -196,118 +189,89 @@ Hello world example in go
...
 		Pattern: protocol.PatternInfo{
 			Query:        &protocol.PatternNode{Value: ""},
 			IncludePaths: []string{"."},
 		},
-		Want: `
-added.md
-changed.go
-unchanged.md
-`,
+		Want: autogold.Expect("added.md\nchanged.go\nunchanged.md"),
```
keegancsmith, line 235: the introduction of autogold made this test far less readable. I'm not really sure why you introduced it?

cmd/searcher/internal/search/hybrid_test.go
```diff
@@ -196,118 +189,89 @@ Hello world example in go
 				},
 			},
 		},
-		Want: `
-added.md:1:1:
+		Want: autogold.Expect(`added.md:1:1:
 hello world I am added
-changed.go:1:1:
+changed.go:1:2:
```
keegancsmith, line 194: these numbers shouldn't be changing. You needed to update the `toString` method to correctly count the number of lines we are matching on.

cmd/searcher/internal/search/search_test.go
```diff
@@ -243,41 +243,41 @@ fmt.Println("Hello world")
...
-		want: autogold.Expect(`main.go:1:3:
+		want: autogold.Expect(`main.go:1:4:
 package main
 
 import "fmt"
 `),
 	}, {
 		arg: protocol.PatternInfo{Query: &protocol.PatternNode{Value: "package main\n\\s*import \"fmt\"", IsRegExp: true}, IsCaseSensitive: false, PatternMatchesPath: true, PatternMatchesContent: true},
-		want: autogold.Expect(`main.go:1:3:
+		want: autogold.Expect(`main.go:1:4:
```
keegancsmith, line 253: like in hybrid search I don't see why this should change. It is asserting on how many lines we matched. Yes you have changed what content looks like, but then you should update how we marshal the result so that it still says 3 lines.

cmd/searcher/internal/search/search_test.go
```diff
@@ -745,7 +744,7 @@ func toString(m []protocol.FileMatch) string {
 			buf.WriteString(strconv.Itoa(int(cm.ContentStart.Line) + strings.Count(cm.Content, "\n") + 1))
 			buf.WriteByte(':')
 			buf.WriteByte('\n')
-			buf.WriteString(cm.Content)
+			buf.WriteString(strings.TrimSuffix(cm.Content, "\n"))
```
keegancsmith, line 747: this will change the test output fixtures, but I wonder if we should avoid this trimming just to correctly assert we include the new line?

cmd/searcher/internal/search/hybrid_test.go
```diff
@@ -148,39 +148,32 @@ Hello world example in go`, typeFile},
...
 		Pattern: protocol.PatternInfo{Query: &protocol.PatternNode{Value: "world"}},
-		Want: `
-added.md:1:1:
+		Want: autogold.Expect(`added.md:1:1:
 hello world I am added
-changed.go:6:6:
-	fmt.Println("Hello world")
-unchanged.md:1:1:
+changed.go:6:7:
+fmt.Println("Hello world")
```
keegancsmith, line 158: I don't understand why this changed (the indentation). I think this is likely an autogold thing which is now not asserting correctly on what we want to find.

internal/search/result/file.go
```diff
@@ -279,7 +279,7 @@ func (cm ChunkMatch) MatchedContent() []string {
 // between lines in a multiline match, but it allows us to keep providing the
 // LineMatch representation for clients without breaking backwards compatibility.
 func (h ChunkMatch) AsLineMatches() []*LineMatch {
-	lines := strings.Split(h.Content, "\n")
+	lines := strings.Split(strings.TrimSuffix(h.Content, "\n"), "\n")
```
keegancsmith, line 282: your implementation doesn't include the `\n` for each preview. I guess that is fine since then we atleast don't break clients that still use linematches?

keegancsmith (approved):
How do we communicate this extra new line to consumers of our APIs that we don't control?

Do we need to update src-cli?

cmd/searcher/internal/search/hybrid_test.go
camdencheek: Ah, sorry. The tests stripped whitespace for comparison, which is not very helpful to test the change here. I'll revert the autogold change and split the test changes into reviewable commits (one commit to stop stripping whitespace, one commit that changes behavior).

cmd/searcher/internal/search/search_test.go
```diff
@@ -290,15 +302,15 @@ func main() {
...
+// No newline at end of chunk
+main.go:1:7:
 package main
 
 import "fmt"
 
 func main() {
 fmt.Println("Hello world")
 }
-
```
camdencheek, line 301: This is a notable test. Before, it matched the trailing newline, which caused us to include it in the chunk. Now, we always include it in the chunk. So before, we were overcounting the number of lines because we weren't expecting the trailing newline to be included.

cmd/searcher/internal/search/search_test.go
```diff
@@ -740,13 +755,17 @@ func toString(m []protocol.FileMatch) string {
...
 			buf.WriteByte(':')
-			buf.WriteString(strconv.Itoa(int(cm.ContentStart.Line) + strings.Count(cm.Content, "\n") + 1))
+			buf.WriteString(strconv.Itoa(lastLine))
 			buf.WriteByte(':')
 			buf.WriteByte('\n')
 			buf.WriteString(cm.Content)
-			buf.WriteByte('\n')
+			if !strings.HasSuffix(cm.Content, "\n") {
+				buf.WriteString("\n// No newline at end of chunk\n")
+			}
```
camdencheek, line 768: Added this to specifically call out in the test when there is no trailing newline in the chunk.

cmd/searcher/internal/search/hybrid_test.go
camdencheek: Alright, fixed. The test diffs are much more readable now.

cmd/searcher/internal/search/hybrid_test.go
camdencheek: Ah, yep, thanks. For some reason I was thinking this was the column, so adding a newline character would bump it by one. Fixed 👍

cmd/searcher/internal/search/search_test.go
camdencheek: Fixed 👍

cmd/searcher/internal/search/search_test.go
camdencheek: Yeah, agreed. I added a specific "no newline at end of chunk" message to make it more obvious what's going on here

cmd/searcher/internal/search/hybrid_test.go
camdencheek: I think this had to do with trimming spaces. In any case, reverting the autogold changes undid this.

internal/search/result/file.go
camdencheek: That was my thought. We specifically only keep `LineMatches` around for backcompat, so I figured if I can easily avoid breaking that I should.

camdencheek:
Going to go ahead and merge this, but happy to follow up with any additional comments

keegancsmith (commented):
Updates LGTM. Also nice idea on the inclusion of the no newline in the tests.

PR #61252, opened 2024-03-19
fix regression and add test

burmudar:
Generated command was missing a `;`
## Test plan
CI and [bazel-do invoke](https://buildkite.com/sourcegraph/sourcegraph/builds/265638#018e55e9-270a-4234-b1d9-9da8b7c40f63)


jhchabran (approved):
Thanks for adding a test :)

PR #61261, opened 2024-03-19
logging: migate rockskip search away from log15

burmudar:
migrate from log15 to `sourcegraph/log`
## Test plan
CI

### DIFF
--- a/internal/rockskip/search.go
+++ b/internal/rockskip/search.go
@@ -11,10 +11,11 @@ import (
 	"github.com/amit7itz/goset"
 	"github.com/grafana/regexp"
 	"github.com/grafana/regexp/syntax"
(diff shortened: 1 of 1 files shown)


internal/rockskip/search.go
```diff
@@ -564,7 +570,6 @@ func regexMatch(conditions Conditions, regex string, isCaseSensitive bool) *sqlf
 		return conditions.regexI(regex)
 	}
 
-	log15.Error("None of the conditions matched", "regex", regex)
```
burmudar, line 567: Wondering if it is ok to remove. I don't think it provides much value but that is just my assessment after a few minutes

Strum355 (approved):

PR #61265, opened 2024-03-19
feat: prototype hotkey components for webapp rewrite

### LINKED ISSUE #61145: Support for global keyboard shortcuts
We should have a simple API to allow a component to register a global keyboard shortcut. Similar to what we have in React (but better).

An example use case would be to register a keyboard shortcut to focus to search input or to collapse/expand the repo sidebar.

A couple of points to consider:

- A simple, straightforward way to specify keybindings. Personally I like [CodeMirror's approach](https://codemirror.net/docs/ref/#view.KeyBinding) (e.g. `Ctrl-Space`).
- "What you see is what you mean". When specifying e.g. `Alt-t` as keybinding works as expected, even though `Alt-t` produces `†` ([current workaround](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@9bbce03c7a474ad16b2486fbe9ba0e63c2ab3c49/-/blob/client/shared/src/keyboardShortcuts/keyboardShortcuts.ts?L46-47)). Not sure if that's even possible given the information available on the [`KeyboardEvent`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent).
- Global keyboard shortcuts shouldn't be triggered (by default) on input elements, so there needs to be a way to detect whether the event originates from within an input (like the current React version).
- Maybe this doesn't need to be a component? It could be just a function that is called curing component initialization. But maybe that makes teardown or conditional binding more difficult.

Further things to keep in mind but not necessary for an initial implementation:

- Eventually it should be possible to automatically show list of all currently available keybindings, e.g. in a help dialog.
- Certain keyboard bindings might become user configurable.

---

The current `<Shortcut />` React component is a good starting point as well as the CodeMirror source.

bahrmichael:
Closes #61145 

This PR introduces a new function to register hotkeys in the svelte app:

```
export function registerHotkey({ keys, handler, allowDefault, ignoreInputFields }: HotkeySetupOptions): {
    bind: (options: HotkeyOptions) => void
}
```

https://github.com/sourcegraph/sourcegraph/pull/61265/files#diff-19ffc44bcfd649cf8bffe0c7ba2425ecac33668d2a0aeaa334dbaac50db1e47fR109-R111

You can use the function in Svelte components' `<script>` section:

```
    const hk = registerHotkey({
        keys: {
            key: 'ctrl+o',
            mac: 'command+o',
        },
        handler: () => alert('ctrl+o'),
    })

    $: hk.bind({
        keys: {
            key: 'ctrl+o',
        },
        handler: () => alert('ctrl+o (rebind)'),
    })
```

---

**The info below is somewhat outdated, but kept for context**

Here's how the current version stacks against the requirements:

`A simple, straightforward way to specify keybindings. Personally I like [CodeMirror's approach](https://codemirror.net/docs/ref/#view.KeyBinding) (e.g. Ctrl-Space).`

Here's a simple example how a new component that wraps `hotkeys-js` can look like:

```html
<Hotkey key="Alt+T" run={() => alert('test')} />
```

`"What you see is what you mean". When specifying e.g. Alt-t as keybinding works as expected, even though Alt-t produces †. Not sure if that's even possible given the information available on the KeyboardEvent`

My tests showed that `hotkeys-js` properly maps the given example to alt+t. It looks like they solved this and other cases for us. You can play around with other hotkeys here: https://wangchujiang.com/hotkeys-js/

`Global keyboard shortcuts shouldn't be triggered (by default) on input elements, so there needs to be a way to detect whether the event originates from within an input (like the current React version).`

`hotkeys-js` [by default ignores input elements](https://github.com/jaywcjlove/hotkeys-js?tab=readme-ov-file#filter). They solve this for us.

`Maybe this doesn't need to be a component? It could be just a function that is called curing component initialization. But maybe that makes teardown or conditional binding more difficult.`

It seems to work well as a component in my tests. With components we have the upside of de-registering hotkeys when their component's lifecycle ends.

`Eventually it should be possible to automatically show list of all currently available keybindings, e.g. in a help dialog.`

`hotkeys-js` comes with `hotkeys.getAllKeyCodes()` which prints all registered hotkeys. We only need to process the resulting list and make it pretty.

`Certain keyboard bindings might become user configurable.`

I included a function in the prototype component that could map a key before registering it. Let me know if this approach is flexible enough.

## Test plan

- Tested this manually in various iterations
- Most of the new code is covered by unit tests
- For now the new code is not used anywhere yet, and once we register hotkeys we should add e2e tests that make sure the desired action happens when a hotkey is used.


client/web-sveltekit/src/lib/Hotkey.svelte
```diff
@@ -0,0 +1,24 @@
...
+        // Here we can look up custom mappings
+        return key;
+    }
+
+    hotkeys(getEvaluatedKey(), function(event, _handler){
+        if (preventDefault) {
+            event.preventDefault();
+        }
+        run();
+    });
```
fkling, line 19: While I don't suspect that it's common to change `key` during the lifetime oft he component, somebody will likely do it at some point. So this should be setup in a way to handle this. I.e. if `key` changes we would have to unbind the key rebind it (I assume). This could look something like this:


```js
let boundKey: string|undefined
$: if (boundKey !== key) {
  if (boundKey) {
    hotkeys.unbind(boundKey)
  }
  boundKey = key
  hotkeys(key, handler)
}
```

We still need to unbind on destroy of course.

client/web-sveltekit/src/lib/Hotkey.svelte
```diff
@@ -0,0 +1,24 @@
...
+
+    onDestroy(() => {
+        hotkeys.unbind(key);
```
fkling, line 22: Is it possible to define which function to unbind? I hope we'll never do that (maybe we should add some runtime logic that guards against this) but I'm wondering what happens when the component is used multiple times with the same key.

fkling (commented):
That looks very promising, thank you! Some initial thoughts:

- If I understand correctly, "filter" for contenteditable elements has to be added, it's not done by default. CodeMirror uses contenteditable elements, so we would have to add that if the library doesn't do this by default.
- It doesn't look like there is support for a `Mod` modifier that is `Ctrl` on Linux and `Command` on macOS. In general I'd be OK if we don't support this and instead require keybindings for both platforms, but that should still be handled transparently (i.e. `ctrl+t, command+t` shouldn't register two keybindings on macOS).

client/web-sveltekit/src/lib/Hotkey.svelte
bahrmichael: I tried to push this a bit further, and introduced a new `ReactiveHotkey` that recreates the binding component when any of the inputs change. In theory this should also update the binding when the function that runs it changes. Let me know what you think about this approach. I feel like it's a bit cleaner to use svelte-concepts like `#key` instead of doing the re-binding myself.

`
{#key [keyDefault, keyOverride, run, preventDefaultEffect]}
    <Hotkey key={keyOverride || keyDefault} run={run} preventDefaultEffect={preventDefaultEffect}/>
{/key}
`

client/web-sveltekit/src/lib/Hotkey.svelte
bahrmichael: I experimented a bit with this and found this a bit more tricky. It looks like the used library doesn't care about conflicts, and just registers multiple bindings.

What I'm observing is that Svelte will mount multiple components, and then call the code in each of them. We don't seem to have a lot control about what happens, unless we move the binding to a later stage. I've got a prototype for this in https://github.com/sourcegraph/sourcegraph/pull/61265/commits/7ccb05c2b03f29c22402288ce136292272fcc0b9, where I move the binding into `onMount` and then do a check if that hotkey is already registered. It seems like the hotkey is not yet registered if we do the check outside of `onMount`.

Currently my approach display an alert, but we can change that to unbinding the previous hotkey. My gut feeling is that we should prevent multiple bindings for the same hotkey, what do you think?

bahrmichael:
> If I understand correctly, "filter" for content-editable elements has to be added, it's not done by default. CodeMirror uses content-editable elements, so we would have to add that if the library doesn't do this by default.

I think the library supports this by default. You can see in the video of the PR description that it behaves differently when I focus an input field (pressed the same key both times, one time it shows the alert and one time it just writes the character).

> It doesn't look like there is support for a `Mod` modifier that is `Ctrl` on Linux and `Command` on macOS. In general I'd be OK if we don't support this and instead require keybindings for both platforms, but that should still be handled transparently (i.e. `ctrl+t, command+t` shouldn't register two keybindings on macOS).

I need a bit more input on what the expected user experience should be. Here's what I currently understand:
* When I register a hotkey `ctrl+t` that should trigger equally on linux, mac, and windows (all with ctrl+t).
* When I register two hotkeys `ctrl+t` and `command+t` it should run the action on all systems that support `ctrl`, but also for the other one on systems that support `command`. I'm not sure if multiple hotkeys for the same action is something we could and should prevent.
* I may want to register a hotkey that's `ctrl+t` on linux and windows, but `command+t` on mac. Here I would need the modifier key that you mentioned. With `hotkeys-js` I assume that we need to use the `isMacPlatform` from `@sourcegraph/common` and then change what we register with hotkeys.

client/web-sveltekit/src/lib/Hotkey.svelte
fkling: > Currently my approach display an alert, but we can change that to unbinding the previous hotkey. My gut feeling is that we should prevent multiple bindings for the same hotkey, what do you think?

I agree. I would simply call `console.error` though. Unbinding implies that we know which of the two is preferred, but we can't know that.
Ideally we could include the location of the other hotkey in the error message but I don't think that's possible.

client/web-sveltekit/src/lib/ReactiveHotkey.svelte
```diff
@@ -0,0 +1,14 @@
+<script lang="ts">
+    import Hotkey from '$lib/Hotkey.svelte';
+
+    export let run: () => void;
+    export let keyDefault: string;
+    // is there a better way for optional props? currently I'm using a default value, and then make sure it gets ignored with the || oerator
+    export let keyOverride: string = "";
```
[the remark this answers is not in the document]
bahrmichael: Removed that with the newest iteration.

bahrmichael:
> I agree. I would simply call console.error though. Unbinding implies that we know which of the two is preferred, but we can't know that.

I found a way to unbind the function that was registered in a given component, instead of all functions for that hotkey. You can pass a second parameter `method` to the unbind command. The version in https://github.com/sourcegraph/sourcegraph/pull/61265/commits/ce2f3d19cb2f57ad5d626e5952af0e49f2807a07 demonstrates that it only deregisters the expected hotkey. You can try it yourself by pushing the hotkey, going through the alerts, and then pushing the hotkey again. On the second push you should only get one alert.

> Ideally we could include the location of the other hotkey in the error message but I don't think that's possible.

I agree, but haven't found a way to give more context rather than "some other Hotkey component has also registered that".

client/web-sveltekit/src/lib/ReactiveHotkey.svelte
```diff
@@ -0,0 +1,14 @@
...
+    export let preventDefaultEffect: boolean = true;
+
+</script>
+
+{#key [keyDefault, keyOverride, run, preventDefaultEffect]}
+    <Hotkey key={keyOverride || keyDefault} run={run} preventDefaultEffect={preventDefaultEffect}/>
+{/key}
```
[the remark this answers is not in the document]
bahrmichael: Gotcha, thank you :)

client/web-sveltekit/src/lib/Hotkey.svelte
```diff
@@ -0,0 +1,51 @@
+<script lang="ts">
+    import hotkeys from 'hotkeys-js';
+    import type {KeyHandler, HotkeysEvent} from 'hotkeys-js';
+    import {onDestroy, onMount} from 'svelte';
+    import {isLinuxPlatform, isMacPlatform, isWindowsPlatform} from '$root/client/common';
```
fkling, line 5: Please add "external" dependencies to `$lib/common` instead and import from there.

client/web-sveltekit/src/lib/Hotkey.svelte
```diff
@@ -0,0 +1,51 @@
...
+
+    export let run: () => void;
+    export let key: string = '';
```
fkling, line 8: I'd make `key` required. It shouldn't be possible to pass no keys at all.

client/web-sveltekit/src/lib/Hotkey.svelte
```diff
@@ -0,0 +1,51 @@
...
+    export let linux: string = '';
+    export let mac: string = '';
+    export let windows: string = '';
+
+    export let unbindAfterInvoke: boolean = false;
```
fkling, line 13: That's an interesting idea and it's great to see that that works. I would not include it in the first iteration and instead keep things as simple as possible. We can always add this later when we need it.

client/web-sveltekit/src/lib/Hotkey.svelte
```diff
@@ -0,0 +1,51 @@
...
+        } else if (isLinuxPlatform() && keys.linux) {
+            return keys.linux;
+        } else if (isWindowsPlatform() && keys.windows) {
+            return keys.windows;
+        } else {
+            return keys.key;
+        }
+    }
+
+    let evaluatedKey: string = evaluateKey({mac, linux, windows, key});
```
fkling, line 27: I don't think you need this line, but I'm not sure how reactive statements interact with `onMount` (but we can't use onMount anyway if we want to react to key changes).

client/web-sveltekit/src/lib/Hotkey.svelte
```diff
@@ -0,0 +1,51 @@
...
+    $: evaluatedKey = evaluateKey({mac, linux, windows, key});
+    export let preventDefault: boolean = true;
```
fkling, line 29: I'd move this to the top to the other props.

client/web-sveltekit/src/lib/Hotkey.svelte
```diff
@@ -0,0 +1,51 @@
...
+        }
+    }
+
+    onMount(() => {
+        if (hotkeys.getAllKeyCodes().map(k => k.shortcut).includes(evaluatedKey)) {
+            console.error(`The hotkey "${evaluatedKey}" has already been registered by another Hotkey component.`);
+        }
+
+        hotkeys(evaluatedKey, handler);
+    });
```
fkling, line 49: We should just make this reactive, then it works with changed keys. Something like this:

```suggestion
    let boundKey: string | undefined
    $: if (evaluatedKey !== boundKey) {
        if (boundKey) {
           hotkeys.unbind(boundKey, handler)
        }
     
        if (hotkeys.getAllKeyCodes().map(k => k.shortcut).includes(evaluatedKey)) {
            console.error(`The hotkey "${evaluatedKey}" has already been registered by another Hotkey component.`);
        }

        boundKey = evaluatedKey
        hotkeys(evaluatedKey, handler);
    }
```

fkling (commented):
Nice! Regarding the function approach: not sure what's better. Personally I'm not a big fan of the practice in the React world to make things components that don't have to be components. But maybe it's not too bad either.

bahrmichael:
I wrote a function based approach based on your suggestion @fkling in https://github.com/sourcegraph/sourcegraph/pull/61265/commits/dd242a80eab6ce63a5e90730080db7740922d833. It works as expected.

My personal feeling is that the component based approach is a bit easier to read, and has less overwrite things going on (in particular overwriting the current* variables). This is a bit scrappy though, and maybe can be improved further. What do you think?

client/web-sveltekit/src/lib/Hotkey.svelte
bahrmichael: I would also not include this functionality when we release it. This was intended more as a test to see if we can unbind one function but not another from the same hotkey.

client/web-sveltekit/src/lib/Hotkey.svelte
bahrmichael: Resolving this as I moved to the function-based approach where this doesn't apply anymore.

client/web-sveltekit/src/lib/Hotkey.svelte
bahrmichael: Resolving this as I moved to the function-based approach where this doesn't apply anymore.

client/web-sveltekit/src/lib/Hotkey.svelte
bahrmichael: Resolving this as I moved to the function-based approach. Moved the error message over as well.

client/web-sveltekit/src/lib/ReactiveHotkey.svelte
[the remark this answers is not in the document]
bahrmichael: Applied this approach, see https://github.com/sourcegraph/sourcegraph/pull/61265/#issuecomment-2017763713.

bahrmichael:
Idea: Move this to common so other clients can use this as well.

client/web-sveltekit/src/lib/Hotkey.test.ts
bahrmichael, line 1: This test doesn't test the integration with hotkeys-js, because integrating with it got very tricky. We would have to do a lot of complicated mocking and workarounds to test that we correctly call the hotkeys functions. The main issue here was that hotkeys exports a default function as well as some sub-functions.

If we want to really test this, I feel like a proxy around hotkeys-js would be best.

bahrmichael: I also tried to test the isContentField function, which however quickly turned into a lot of config changes. That's because currently we seem to run in a node-only environment, and jsdom is not set up yet.

fkling: Have a look at the other unit tests, I think you can enable jsdom for a test file via a comment at the top of the file.

client/web-sveltekit/src/lib/Hotkey.ts
```diff
@@ -0,0 +1,151 @@
...
+    if (keys.windows && isWindowsPlatform()) {
+        return keys.windows
+    }
+
+    return keys.key
+}
+
+export function isElement(t: any): t is Element {
+    return t instanceof Element
+}
```
camdencheek, line 24: Does this need to be exported?

client/web-sveltekit/src/lib/Hotkey.ts
```diff
@@ -0,0 +1,151 @@
...
+            .includes(currentKey)
+    ) {
+        // Instead of printing an error, we can also use hotkey's "single" option, which will automatically unregister any
+        // existing hotkey with the same key and scope.
+        console.error(`The hotkey "${currentKey}" has already been registered by another Hotkey component.`)
+    }
+    let wrappedHandler = wrapHandler(handler, allowDefault, ignoreInputFields)
+
+    // By default, hotkeys-js ignores input fields. Unfortunately this filter can only be set globally, and will apply to all hotkeys.
+    // We work around this by always checking input fields, and then applying a custom filter in the wrappedHandler function.
```
camdencheek, line 126: Wow, that's annoying!

client/web-sveltekit/src/lib/Hotkey.ts
```diff
@@ -0,0 +1,151 @@
...
+    /**
+     * Whether the handler should be executed when the user focuses an input field.
+     */
+    ignoreInputFields?: boolean
+}
+
+/**
+ * Creates a global keyboard shortcut. Needs to be called during
+ * component initialization.
+ */
```
camdencheek, line 108: What counts as "component initialization"? Is this specifically inside `onMount`? And if so, should this return an "unregister" function?

camdencheek (approved):

client/web-sveltekit/src/lib/Hotkey.ts
fkling: @camdencheek component initialization basically is the constructor of the component. Any code that's not reactive and not in a callback.

client/web-sveltekit/src/lib/Hotkey.ts
```diff
@@ -0,0 +1,151 @@
...
+export function registerHotkey({ keys, handler, allowDefault, ignoreInputFields }: HotkeySetupOptions): {
```
vovakulikov, line 109: I'm a little bit confused by API, so if I got it right we can either 
- pass options like keys, handler, allowDefault, etc, as arguments for `registerHotkey` and bind hotkey 
- pass `HotkeyOptions` through  bind return function and then bind 

And if use the second option here we will reuse some arguments from first call, my question here is
what is there reason for having multiple options for binding and setting hotkey? 

cc @fkling

client/web-sveltekit/src/lib/Hotkey.ts
vovakulikov: I think the reason why this should be called during "component initialization" is because we use `onDestroy` hook here from svelte. Which I don't fully understand (correct me if I'm wrong) what if we could provide an unbind API so consumer should unsubscribe themself?

client/web-sveltekit/src/lib/Hotkey.ts
vovakulikov: What is the reason for this usage with reactive operator `$`, is it just imperative API to set hotkey dynamically in some callback or by some conditions? 

```
const hk = registerHotkey({
        keys: {
            key: 'ctrl+o',
            mac: 'command+o',
        },
        handler: () => alert('ctrl+o'),
    })

    $: hk.bind({
        keys: {
            key: 'ctrl+o',
        },
        handler: () => alert('ctrl+o (rebind)'),
    })
    ```

client/web-sveltekit/src/lib/Hotkey.ts
bahrmichael: > And if so, should this return an "unregister" function?

Added a `unregister` function that gets exported alongside the `bind` function. I don't see where we would need that yet, but it wasn't difficult to add.

> Which I don't fully understand (correct me if I'm wrong) what if we could provide an unbind API so consumer should unsubscribe themself?

I think it's better to default to automatically doing the right thing. I would probably forget to unsubscribe the hotkey, and then we get duplicates all over the place. You can do so manually with the upcoming `unregister` function, but I would keep the `onDestroy` hook.

client/web-sveltekit/src/lib/Hotkey.ts
fkling: @vovakulikov It would allow us to change the key binding on the fly. If you think that's premature we can remove it for now.

bahrmichael: As @fkling said the reason for the different interfaces is that we may want to rebind the hotkey during a component's lifecycle. That would happen when we allow users to change hotkeys.

As for the example with the `$` you're right, the idea is that we may want to change the hotkey or the action at a later time during the component's lifecycle.

vovakulikov: Yeah I thought about this, I think it makes sense, just not sure about API around registerHotkey and bind function, do we need to have them both. I mean can it be just one function registerHotkey, but then I think onDestroy within this function if we use it with $ operator won't work, so maybe this is how it needs to be, but just double check

client/web-sveltekit/src/lib/Hotkey.ts
fkling: I'm kind of split between these options. On one hand I like flexibility, and returning an unsubscribe function would allow callers to handle it however they want.
On the other hand I like it when things "just work" and you can't forget to unsubscribe, as Michael said.

Maybe more importantly we have to establish a common pattern. Should we make our APIs as foolproof as possible and handle things like unsubscribing internally, or not.

client/web-sveltekit/src/lib/Hotkey.test.ts
bahrmichael: Thank you, that fixed it :)

bahrmichael:
I'm not seeing any blocking comments/discussion, and will therefore merge this. Please poke me if you still have some input that needs to be addressed.

PR #61268, opened 2024-03-19
Adds documentation for the codeintel commitgraph

kritzcreek:
As this is documentation purely meant for someone working on the `commitgraph` code I decided to add it "in-tree" right next to relevant source files.
I still need to fill in how we build the data algorithmically, but I've gotten to a point where I describe what constitutes an annotated commitgraph.
I've added you as reviewers because that first part is ready to be reviewed.

## Test plan

documentation changes


internal/codeintel/uploads/internal/commitgraph/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Loading data
+
+For the marked as dirty repository:
+- Get full commitgraph (non-annotated) from git server. All commits and their relationships
+- Get metadata for all uploaded indices (indexer name, commit, root directory)
+
+### What's an annotated commitgraph
+
+A fully annotated commitgraph contains a set of visible uploads for every commit. We can never see multiple visible uploads from the same indexer _and_ root at the same commit.
```
mmanela, line 31: What do you mean by "we can never see"?

internal/codeintel/uploads/internal/commitgraph/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Updating the commitgraph
+1. Load data
+2. Compute new commitgraph
+3. Diff old vs new and do incremental updates
+
+### Loading data
+
+For the marked as dirty repository:
+- Get full commitgraph (non-annotated) from git server. All commits and their relationships
```
mmanela, line 26: We pull the full git commit history each time?

internal/codeintel/uploads/internal/commitgraph/README.md
```diff
@@ -0,0 +1,105 @@
...
+}
+```
+
+We'll look at a couple of examples with increasing complexity to understand what constitutes visibility.
+
+#### Visibility rules
+
+Our examples will be drawings of commitgraphs.
+
+We will use `I1`..`In` to denote indexer+root combinations or "index keys". For example `I1` could correspond to indices created by `"scip-go"` in the `"backend/"` root directory.
```
mmanela, line 55: How does the root directory of the index affect the shadowing? Should that be part of the shadowing algo below?

internal/codeintel/uploads/internal/commitgraph/README.md
kritzcreek: For the example given here it means this is not valid, because it has multiple entries for the same indexer + root combination.
```
commitgraph = {
  "deadbeef" => {
    ("scip-go", "backend/") => (UploadId(23), 0),
    ("scip-go", "backend/") => (UploadId(25), 0),
    ("scip-go", "worker/") => (UploadId(24), 1),
    ("scip-typescript", "frontend/") => (UploadId(33), 1),
  }
}
```

internal/codeintel/uploads/internal/commitgraph/README.md
kritzcreek: Yes, well the output of `git log --pretty="%H %P" --topo-order` (so no commit messages or changes, just hashes)

internal/codeintel/uploads/internal/commitgraph/README.md
mmanela: Gotcha

internal/codeintel/uploads/internal/commitgraph/README.md
kritzcreek: `("scip-go", "backend/")` only shadows other uploads with the same indexer and root. If you also upload `("scip-go", "backend/database/")` you'll get duplicate results for files inside `backend/database/`. My mathy description earlier included an ordering on paths, but in the current implementation only equality is considered.

varungandhi-src:
Let's avoid checking in binary files here. The [docs](https://sourcegraph.com/docs/dev/how-to/documentation_implementation#adding-images-to-the-documentation) say we should use the sourcegraph-assets bucket for images over 100kB, and technically, the ones in this PR are currently smaller than that, but I think there's a potential risk if you end up making some changes and those end up going past 100kB.

I've created a top-level [dev-docs folder](https://console.cloud.google.com/storage/browser/sourcegraph-assets/dev-docs?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&folder=true&organizationId=true&project=sourcegraph-de&prefix=&forceOnObjectsSortingFiltering=false) in the bucket, you should be able to add images there.

Once that's done, you can force-push to your branch (hopefully that should avoid people fetching the old commit?).

internal/codeintel/uploads/internal/commitgraph/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+We will use single uppercase letters in boxes with rounded corners to denote commits.
+Colored boxes mark commits with uploads, and black boxes those with no uploads.
+A commit with multiple uploads will have additional colored outlines.
+
+Boxes with sharp corners record the set of visible uploads for a commit. An entry like `I1 : A(3)` means this commit sees an upload for `I1` made at commit `A` at a depth of 3.
```
varungandhi-src, line 61: Minor notation suggestion: Perhaps it should be `A+3`/`A↦3`/`A↣3`, or some other rightwards arrow, (to convey "3 commits ahead of A") rather than `A(3)`? 

Plus is also evocative, as you add `+1` to the numeric value as you go along.

Given that there is redundancy between the colors and the zero distances, I'm mildly in favor of removing the colors altogether, but it's not a big deal either way.

internal/codeintel/uploads/internal/commitgraph/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+**Rule 3: Uploads for the same index key shadow uploads at greater depth**
+
+![commitgraph_2.png](img/commitgraph_2.png)
+
+In this example `C` now has an upload for `I1`, which means it sees the upload at depth 0, rather than the one from `B` at depth 1.
+
+`E` sees the upload for `I2` at `D` at a distance of 1, rather than `B` because the upload at `B` has a distance of 2.
+The notion of shadowing can be a bit unintuitive here, as `D` does *not* need to be on the path from `B` to `E` in order to shadow `B`.
+
```
varungandhi-src, line 85:
```suggestion
`E` sees the upload for `I2` at `D` at a distance of 1, rather than `B` because the upload at `B` has a distance of 2.
The notion of shadowing can be a bit unintuitive here, as `D` does *not* need to be on the path from `B` to `E` in order to shadow `B`.

In practice, this means that if a codebase uses merge commits and also indexes commits not on the `main` branch, it is possible to fall back to uploads on commits that aren't on the `main` branch.
```

internal/codeintel/uploads/internal/commitgraph/README.md
```diff
@@ -0,0 +1,105 @@
+# commitgraph
+
+Problem: Given a commit hash, find me the set of index uploads most relevant to that commit.
```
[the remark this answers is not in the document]
kritzcreek: I'd say that happens in a layer above the commit graph. Filtering on paths and indexer name is only applied _after_ the `lsif_nearest_uploads` table has been queried with repository_id and commit hash.

I'll add that the repository id is also an input.

internal/codeintel/uploads/internal/commitgraph/README.md
```diff
@@ -0,0 +1,105 @@
...
+- the index is still being built, or the commit is in the queue to be indexed
+- indices have been deleted, either by an explicit user action or retention policy
+
+To solve this problem we build an annotated copy of the git commit graph in Postgres, so that we can return so called "visibleUploads" quickly.
+
+## High-level architecture
+
+- repository listing (table `lsif_dirty_repositories`), isDirty flag is set whenever an upload is completed
+- Worker process updates commitgraph for dirty repositories and clears flag
+- TODO: When information is requested for an unknown commit in a repository we ask git server for a fragment of the commitgraph that includes the commit and "graft" it onto our existing graph
```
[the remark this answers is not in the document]
kritzcreek: This section isn't fleshed out yet. As this is more of an operational concern maybe it shouldn't even go in here? I thought it might provide some helpful context to get an idea of how often/when this code is being run.

internal/codeintel/uploads/internal/commitgraph/README.md
```diff
@@ -0,0 +1,105 @@
...
+Conceptually:
+
+`commitgraph : Map<Commit, Map<(Indexer, Root), (UploadId, Distance)>>`
+
+Here's what an example entry for a commit `deadbeef` might look like for our conceptual model:
+
+```go
+commitgraph = {
+  "deadbeef" => {
+    ("scip-go", "backend/") => (UploadId(23), 0),
```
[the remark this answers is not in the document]
kritzcreek: Yeah this is correct. Here's the list of distinct indexers present on S2:

```
pgsql=> select distinct indexer from lsif_uploads;
     indexer
-----------------
 lsif-go
 scip-java
 scip-ruby
 scip-python
 scip-typescript
 scip-clang
 scip-go
 rust-analyzer
(8 rows)
```

[It's hashed and stored from here](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@8f9ed348688f0086d76ab1e7f48f5851e278af2b/-/blob/internal/codeintel/uploads/internal/store/commitgraph.go?L170)

internal/codeintel/uploads/internal/commitgraph/README.md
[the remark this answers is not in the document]
varungandhi-src: > As this is more of an operational concern maybe it shouldn't even go in here?

There are a few sub-parts to this as I see it:
1. The core conceptual model, which is mostly abstract.
2. The overall architecture/implementation in code
3. Operational docs: e.g. with examples of RAM usage, trouble-shooting steps with links to example failures/bad behavior. (You don't have to do this right now)

Out of these, 1. belongs here. For 2., I'd say that still belongs here because of none of the details of this mechanism are exposed to customers (via the GraphQL API). For 3., I think that should be in the docs for site-admins.

After the initial motivation/summary, I think it would make sense to have 2 first, and then link the conceptual model from the bit where you're describing the logic, as that would be a more detailed explanation of what the code is trying to do. But you could also do it differently.

internal/codeintel/uploads/internal/commitgraph/README.md
kritzcreek: I got positive feedback on the colored outlines from Matt so I'm inclined to leave them as is. A `+` might've been a bit clearer, than what we have but I don't want to redo all the graphs and I think with the explanation given here it's still pretty clear.

kritzcreek:
I've reorganized things a bit, and think this is ready to go in now. I decided to not document the implemented algorithm too much as it's easy enough to get by reading the code, if you have the specification for the structure it's producing (this document).

internal/codeintel/uploads/internal/commitgraph/README.md
varungandhi-src: Yeah, I don't think it's a big deal. You used Excalidraw, right? I believe that it has save/load functionality, so next time you could save the images for editing if there are some changes requested.

(Tldraw is also similar to Excalidraw, is free, and I know that has save/load functionality.)

kritzcreek: Yeah it's excalidraw. If I do this again I might try to lean GH's builtin support for Mermaid diagrams and see how custom I can get things. That way I could avoid the Google Cloud bucket...

internal/codeintel/uploads/internal/commitgraph/README.md
```diff
@@ -0,0 +1,116 @@
...
+### Linked commits
+
+Because storing the full map of visible uploads per commit can be expensive, we store _links_ to ancestors were possible.
+This is purely an optimization to reduce the amount of storage we use.
+
+A commit is stored as a link iff:
+
+- It does not have associated uploads
+- It is not a merge commit
+- It does not have a child that is a merge commit
```
varungandhi-src, line 84: Is this fully correct or missing an edge case related to leafs/roots?

internal/codeintel/uploads/internal/commitgraph/README.md
```diff
@@ -0,0 +1,116 @@
...
+For these commits we store the first ancestor commit that isn't a link, as well as the depth to that ancestor.
+The visible uploads for those commits can then be computed by retrieving the visible uploads for the linked commit and increasing all their depths by the depth between the linked commits.
+
+![commitgraph_3.png](https://storage.googleapis.com/sourcegraph-assets/dev-docs/commitgraph/commitgraph_3.png)
+
+In this example both `B` and `C` are stored as links. `E` is not stored as a link, because its child `F` is a merge commit.
+
+## Architecture & operations
+
+At a high level the annotated commit graph is kept up-to-date by a worker that runs periodically. We maintain a table `lsif_dirty_repositories` that tracks what repositories have had new indexes uploaded or old ones deleted.
```
varungandhi-src, line 95: Suggestion: Mention the worker by name for easier searching

internal/codeintel/uploads/internal/commitgraph/README.md
```diff
@@ -0,0 +1,116 @@
...
+When a repository is marked as dirty we update its annotated commit graph and remove its dirty flag.
+
+### Loading data
+
+For a repository marked as dirty we fetch the following data before computing the annotated commit graph:
+
+- Get full commit graph (non-annotated) from git server. All commits and their relationships
+- Get metadata for all uploaded indexes (indexer name, commit, root directory)
```
varungandhi-src, line 103:
```suggestion
For a repository marked as dirty we fetch the following data before computing the annotated commit graph:

- Full non-annotated commit graph (non-annotated) from git server, i.e. every parent for every commit
- Metadata for all uploaded indexes (indexer name, commit, root directory)
```

Since the previous sentence has 'we fetch', 'Get' at the start of the bullet looks a bit odd.

varungandhi-src (approved):
Left some minor non-blocking suggestions

internal/codeintel/uploads/internal/commitgraph/README.md
kritzcreek: When you say "by name" do you mean the file it's defined in (`internal/codeintel/uploads/internal/background/commitgraph/job_commitgraph.go`) or its go struct name (`commitGraphUpdater`)?

kritzcreek: NVM I just went with the Go name, that should be easy enough to search.

internal/codeintel/uploads/internal/commitgraph/README.md
kritzcreek: I don't think there are edge cases for leafs. For roots if the root does not have an upload it's not stored at all, and so is neither of the commits following it until we get the first upload or merge commit. I'm not sure if that behaviour is deliberate.

internal/codeintel/uploads/internal/commitgraph/README.md
```diff
@@ -0,0 +1,126 @@
...
+- Commits and their visible uploads are stored in `lsif_nearest_uploads`
+  - The visibility maps are stored as `jsonb` objects
+- Linked commits are stored in `lsif_nearest_uploads_links`
+
+These tables are updated via a diffing mechanism. We create temporary tables and insert the full graph into them.
+We then run a query to compare the actual table and the temporary table and insert/delete/update individual rows.
+
+### TODO: Queries for commits not yet in the annotated commit graph
+
+When information is requested for an unknown commit in a repository we ask git server for a fragment of the commit graph that includes the commit and "graft" it onto our existing graph
```
mmanela, line 126: @varungandhi-src @kritzcreek I heard about this before, I am curious how important this feature is? Is this solving the problem of the gap between new commits and indexing completing?

kritzcreek: Great question @mmanela. It's not about indexing not having completed, and more about git server and commitgraph being out-of-sync. We only recompute the (annotated) commit graph when the set of uploads changes, not whenever the set of commits change.

mmanela: But that should be a fixed period of time right? Since we assume the CI should upload a new index after each commit? 

And if we serve index for an older commit, won't references potentially not match line numbers? Do we handle that?

kritzcreek: > But that should be a fixed period of time right? Since we assume the CI should upload a new index after each commit?

That is entirely dependent on the customer's particular setup. I think the most common would be to only index commits on the `main` branch (using auto-indexing), but all of that can be customized and customers that upload from CI have even more freedom.

> And if we serve index for an older commit, won't references potentially not match line numbers? Do we handle that?

My guess right now would be that yes references from older indexes can be off, but I'm still learning new things about the codebase all the time.

PR #61272, opened 2024-03-19
Upgrade GPT 4 Turbo on PLG

### LINKED ISSUE #2919: AWS CodeCommit integration roadmap
This issue captures the work (which spans multiple releases) to improve Sourcegraph's integration with AWS CodeCommit.

- Repository permissions https://github.com/sourcegraph/sourcegraph/issues/2872

Customers using AWS CodeCommit (not exhaustive):

- https://app.hubspot.com/contacts/2762526/company/555146901

philipp-spiess:
Closes #2919

Switching from the hardcoded date version to `gpt-4-turbo-preview` as recommended in https://openai.com/blog/new-embedding-models-and-api-updates.

## Test plan

```
curl 'https://sourcegraph.test:3443/.api/completions/stream' \
-i \
-X POST \
-H 'authorization: TOKEN' \
--data-raw '{"messages":[{"speaker":"human","text":"What is your name?"}],"maxTokensToSample":30,"temperature":0.2,"topP":0.95,"topK":0,"model":"openai/gpt-4-turbo-preview","stopSequences":[],"timeoutMs":5000,"stream":true}'
HTTP/2 200

event: completion
data: {"completion":"","stopReason":""}

event: completion
data: {"completion":"I","stopReason":""}

event: completion
data: {"completion":"I'm","stopReason":""}

event: completion
data: {"completion":"I'm an","stopReason":""}
```

### DIFF
--- a/internal/conf/computed.go
+++ b/internal/conf/computed.go
@@ -1228,11 +1228,14 @@ func openaiDefaultMaxPromptTokens(model string) int {
 		return 7_000
 	case "gpt-4-32k":
 		return 32_000
-	case "gpt-3.5-turbo", "gpt-3.5-turbo-instruct", "gpt-4-1106-preview",
-		"gpt-4-turbo-preview":
+	case "gpt-3.5-turbo-instruct":
 		return 4_000
-	case "gpt-3.5-turbo-16k":
+	case "gpt-3.5-turbo-16k",
+		"gpt-3.5-turbo":
 		return 16_000
+	case "gpt-4-1106-preview",
+		"gpt-4-turbo-preview":
+		return 128_000
 	default:
 		return 4_000
 	}
--- a/internal/conf/computed.go
+++ b/internal/conf/computed.go
@@ -1228,7 +1228,8 @@ func openaiDefaultMaxPromptTokens(model string) int {
 		return 7_000
 	case "gpt-4-32k":
 		return 32_000
-	case "gpt-3.5-turbo", "gpt-3.5-turbo-instruct", "gpt-4-1106-preview":
+	case "gpt-3.5-turbo", "gpt-3.5-turbo-instruct", "gpt-4-1106-preview",
+		"gpt-4-turbo-preview":
 		return 4_000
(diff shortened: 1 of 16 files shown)


arafatkatze (approved):

internal/conf/computed.go
```diff
@@ -1228,7 +1228,8 @@ func openaiDefaultMaxPromptTokens(model string) int {
 		return 7_000
 	case "gpt-4-32k":
 		return 32_000
-	case "gpt-3.5-turbo", "gpt-3.5-turbo-instruct", "gpt-4-1106-preview":
+	case "gpt-3.5-turbo", "gpt-3.5-turbo-instruct", "gpt-4-1106-preview",
```
chwarwick, line 1231: This looks oddly low, why are the max prompt tokens so low for gpt-4-preview it has a large context window.

chwarwick (approved):

internal/conf/computed.go
philipp-spiess: I did not think this through but yes 100% this is a bug

philipp-spiess: Should we set this to the actual maximum here as per docs (so 128k?) I suppose the clients won't use all of that anyways or how does that work

chwarwick: It should be lower, I think for Claude we're around 12k.  The clients _i think_ use 60% of this limit for the initial chat content.

chwarwick: I think the way to think of this limit is the lower of the model context window or the max size of a fully constructed chat/prompt to send.

philipp-spiess: Yeah the client limits it for PLG at least

philipp-spiess: Ah, gotcha, I'll set it to something more reasonable then.

philipp-spiess: Limiting it to 16k so our enterprise customers don't get surprised. I think they can overwrite this anyways? and for plg the client applies other limits anyways

PR #61273, opened 2024-03-19
Cleanup Cody Gateway upstreams.go

chrsmith:
This PR applies some general code cleanups and refactoring to `upstream.go`, which IMO is a bit too gnarly for being such a crucial part of the codebase.

Changes:

**Added a `getAPIURLByFeature(codygateway.Feature) string` function to the methods interface**

Added a new `getAPIURLByFeature` function to the `upstreamHandlerMethods[ReqT UpstreamRequest]` interface.

This removes one of the parameters to `makeUpstreamHandler`, and also makes more sense IMHO. Because the `upstreamHandlerMethods` interface is a facade in front of an LLM API, and so it seemed rather awkward to rely on a one-off function for determining the API URL to use.

As part of this, I also moved the package-level constants for the API URLs to just be constant literals in the `getAPIURLByFeature` functions. This may not be as clear-cut of an improvement, but given that the `const` value was only ever read in one place, it seems better to not "pollute" the package namespace with it. And just keep it hard-coded in that one function.

**Rename `d` to `profDetector`**

Given the vast distance between declaration and usage, renaming `d` to `profDetector` seems like a good call. 

There's a lot of great advice in finding balance between short and long identifiers in Go here:
https://dave.cheney.net/practical-go/presentations/gophercon-israel.html#_the_larger_the_identifiers_scope_the_larger_its_name
The general rule I try to remember is just:

> The greater the distance between a name’s declaration and its uses, the longer the name should be.

**Replace `r.Context()` with `ctx`**

It seems preferable to have a local variable named `ctx` than keep on calling `r.Context()`. This (IMHO) keeps things easier to read. But a more practical argument for doing so is that it doesn't lead to bugs if we need to _alias_ the context later.

e.g. a common pattern is that within an HTTP handler you'd replace `ctx` with something like `ctx = context.WithTimeout(ctx, ...)` or something. And that isn't possible if you keep calling `r.Context()`.

(But I'd be lying if I wasn't just removing this because I personally just want to avoid all the unnecessary function calls and keep the code more tidy and concise.)

There are a couple of other changes I'll explain in PR comments.

## Test plan

Relying on existing unit tests.

(I'd love to write more, but first we need to untangle all of the dependencies in `makeUpstreamHandler`, as mocking out ~~12~~ now 11 parameters is going to be a PITA.

### DIFF
--- a/cmd/cody-gateway/internal/httpapi/completions/upstream.go
+++ b/cmd/cody-gateway/internal/httpapi/completions/upstream.go
@@ -63,10 +63,12 @@ const phrasePrefixLength = 5
 //
 // Methods do not need to be concurrency-safe, as they are only called sequentially.
 type upstreamHandlerMethods[ReqT UpstreamRequest] interface {
+	// getAPIURLByFeature returns the upstream API endpoint to call for the given feature.
+	getAPIURLByFeature(codygateway.Feature) string
+
 	// validateRequest can be used to validate the HTTP request before it is sent upstream.
 	// Returning a non-nil error will stop further processing and return the given error
-	// code, or a 400.
-	// Second return value is a boolean indicating whether the request was flagged during validation.
+	// code. Defaulting to 400.
 	//
 	// The provided logger already contains actor context.
 	validateRequest(context.Context, log.Logger, codygateway.Feature, ReqT) (int, *flaggingResult, error)
@@ -100,6 +102,9 @@ type UpstreamRequest interface {
 	BuildPrompt() string
 }
 
+// makeUpstreamHandler a big deal. This method will produce an http.Handler that will handle converting
+// the Cody Gateway user's request to the backing LLM ("upstream provider"). This is how we provide a
+// consistent way for providing logging, telemetry, rate limiting, etc. across multiple upstream providers.
 func makeUpstreamHandler[ReqT UpstreamRequest](
 	baseLogger log.Logger,
 	eventLogger events.Logger,
@@ -111,7 +116,6 @@ func makeUpstreamHandler[ReqT UpstreamRequest](
 	// provider names defined clientside, i.e. "anthropic" or "openai".
 	upstreamName string,
 
-	upstreamAPIURL func(feature codygateway.Feature) string,
 	allowedModels []string,
 
 	methods upstreamHandlerMethods[ReqT],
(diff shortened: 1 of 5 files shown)


cmd/cody-gateway/internal/httpapi/completions/upstream.go
```diff
@@ -123,9 +127,7 @@ func makeUpstreamHandler[ReqT UpstreamRequest](
 	autoFlushStreamingResponses bool,
 	patternsToDetect []string,
 ) http.Handler {
-	baseLogger = baseLogger.Scoped(upstreamName).
-		// This URL is used only for logging reason so we default to the chat endpoint
-		With(log.String("upstream.url", upstreamAPIURL(codygateway.FeatureChatCompletions)))
```
chrsmith, line 128: This is a functional change I'm introducing with this PR... but I think this current behavior is problematic enough to consider it a bug. (Right?)

What is the point of adding the `upstream.url` field to the `baseLogger` (and adding it as context in every log line produced by that logger) if it will be _wrong_. (Since the upstream URL is dependent on the feature that is used.)

In other words, I'd argue that having no `upstream.url` is preferable to having the wrong `upstream.url`.

If this is important to preserve, wouldn't we just want to wrap `baseLogger` within the body of the `upstreamHandler := func(w http.ResponseWriter, r *http.Request) {` later in the function? (So we can ensure it is always right?)

Or is there a large overhead to creating a new scoped logger, enough so that we would prefer to just do it once when registering the upstream handler and not on every HTTP request. (Even if the value for that log line's attribute would be wrong in some cases.)

cmd/cody-gateway/internal/httpapi/completions/upstream.go
```diff
@@ -136,220 +138,94 @@ func makeUpstreamHandler[ReqT UpstreamRequest](
...
-			var body ReqT
-			if err := json.NewDecoder(r.Body).Decode(&body); err != nil {
-				response.JSONError(logger, w, http.StatusBadRequest, errors.Wrap(err, "failed to parse request body"))
-				return
-			}
-			status, flaggingResult, err := methods.validateRequest(r.Context(), logger, feature, body)
-			if err != nil {
-				if status == 0 {
-					response.JSONError(logger, w, http.StatusBadRequest, errors.Wrap(err, "invalid request"))
-				}
```
chrsmith, line 206: This is a bug in the current implementation. (And was removed in the updated code.)

We write the HTTP response here if the implementation of `validateRequest` did not return a status value. However, we write to the same `http.ResponseWriter` later. That will fail (panic even?) because you cannot write the HTTP header with the status code twice.

So the new code removes this `if status == 0` block, and has a `if status < 400` block right before we call `response.JSONError` and then `return`.

chrsmith: The new code is hard to see in the diff, but the new code to handle this looks like:

```go
			// Ensure that we return a meaningful error to the client for validation failures.
			if status < 400 {
				status = http.StatusBadRequest
				err = errors.Wrap(err, "invalid request")
			}
			response.JSONError(logger, w, status, err)
			return
```

rafax (approved):
LGTM, adding @philipp-spiess who introduced the URL-switching logic to reviewers (but pushing that to handlerMethods 👍 )

philipp-spiess (approved):

PR #61276, opened 2024-03-19
Svelte: add copy file button

camdencheek:
This adds the "copy file path" button to the svelte web app. According to telemetry, this is one of the more used features on the page and it was easy enough to add. 

## Test plan
Manual testing (see video) and added a playwright test

https://github.com/sourcegraph/sourcegraph/assets/12631702/5e7e9fc4-1c52-4760-a8d7-a7a063783c74


client/web-sveltekit/src/routes/search/FileSearchResultHeader.svelte
```diff
@@ -21,13 +31,55 @@
...
+            --color: var(--body-color);
+        }
+    }
+
+    :global(.copy-button-focus-container:hover),
+    :global(.copy-button-focus-container:focus-within) {
+        .copy-button {
+            visibility: visible;
+        }
+    }
```
camdencheek, line 84: This is the hackiest part. Basically, we need to show the button whenever the search result _container_ is hovered or focused. However, we can't refer to CSS classes directly, nor can we pass a class name into the CSS.

I think this is okay, but an alternative here would be to use javascript or CSS variables to set the visible property. I think I generally prefer the pure CSS solution, but lemme know if you think differently.

fkling: Mmh. What do you think about inverting this. Attach a `data-visible-on-hover` data attribute somewhere and target this in the `SearchResult.svelte` component? That seems a bit cleaner to me, but maybe not?

camdencheek: I'm not sure I'm following. If I moved the visibility behavior to a consuming component, wouldn't that mean I still have the same problem where I need to use `:global` to target this component? Or are you saying data attributes don't deal with these scoping things?

fkling: The parent component has to use `:global` but not this component. Basically instead of doing

```css
:global(.class-on-parent:hover) {
   ...
}
```

here


you do something like

```css
.article:hover :global([data-visible-on-hover]) {
  ...
}
```

in `SearchResults.svelte`.

In order words, instead of the header "reaching out" the container "reaches inside". 

And using `data-*` simply as an indicator that this is not a class used for styling but an "API" to target.

camdencheek: Aah, gotcha. I like that -- will push an update. Thanks!

fkling: At first glance "reaching in" seems a bit cleaner to me that "reaching out" but I'm not sure. That's why I'm asking you 😉

camdencheek: Okay, I just pushed a commit that did just that. I also pulled the button out into its own component, which I think made things a bit cleaner and also let me make the component itself ignore any visibility concerns

client/web-sveltekit/src/routes/search/FileSearchResultHeader.svelte
```diff
@@ -1,3 +1,7 @@
+<script context="module">
+    export const focusContainerClass = 'copy-button-focus-container'
+</script>
+
```
fkling, line 4: This can be removed now.

fkling (approved):
Awesome!

fkling:
FWIW You also need to update the `BUILD.bazel` file to include the dependency.

camdencheek:
Ah, thank you! Was trying to figure out what I mucked up

Issue #61278, opened 2024-03-19
Add missing request validation for all LLM providers in Cody Gateway


chrsmith:
Today, only the Anthropic implementation of `upstreamMethods` has any logic in the `validateRequest` function. For example:

```go
func (a *AnthropicHandlerMethods) validateRequest(ctx context.Context, logger log.Logger, _ codygateway.Feature, ar anthropicRequest) error {
	if ar.MaxTokensToSample > int32(a.config.MaxTokensToSample) {
		return errors.Errorf("max_tokens_to_sample exceeds maximum allowed value of %d: %d", a.config.MaxTokensToSample, ar.MaxTokensToSample)
	}
	return nil
}
```

However, we should perform the same type of validation as applicable for Fireworks and OpenAI-based LLMs as well. Any sort of hard-caps for ensuring system stability should be enforced consistently.

PR #61280, opened 2024-03-19
Generalize Cody Gateway request flagging

chrsmith:
This PR continues my crusade to tidy up the Cody Gateway upstream code and try to make our anti-abuse measures a bit more general.

The crux of this PR is the following:

```diff
-	validateRequest(context.Context, log.Logger, codygateway.Feature, ReqT) (int, *flaggingResult, error)
+	validateRequest(context.Context, log.Logger, codygateway.Feature, ReqT) error
+	shouldFlagRequest(context.Context, log.Logger, ReqT) (*flaggingResult, error)
```

Previously we were performing "abuse flagging" as part of the request `validateRequest` function. And also had an awkward function signature, where the `int` return value signified an HTTP status code (or defaulted to zero, which would then be defaulted again into `http.BadRequest`.)

So this PR changes three things:

### Simplify `validateRequest`

Now, the function just returns an error. A non-nil response will serve a `400 Bad Request`, along with the error.

:warning: This does change the behavior for our OpenAI endpoint. We were previously returning a more accurate `501 Not Implemented` if trying to use that provider for "chat". But I'd argue that serving the less accurate HTTP error response (400) is probably less important than a simpler function signature for the `upstreamHandlerMethods` interface.

And if we _really_ wanted to do that, we could check for the appropriate status code some other way. e.g. `errors.As(err, &httpStatusCodeError)

### Add new `shouldFlagRequest` function

We call `shouldFlagRequest` as part of processing an upstream request immediately after calling `validateRequest`. This provides an arguably cleaner way to isolate code for flagging LLM requests on a per-upstream provider basis. Though in practice, I expect all of our implementations here to just wrap the `isFlaggedRequest` function. (Like is the case for Anthropic now.)

⚠️ Part of my refactoring here is that I removed the LLM check that was inside of `isFlaggedAnthropicRequest`. (So with this change, we would _start_ flagging requests sent to "claude-2.0", "claude-v1", etc.)

My understanding is that this is what we want to do in general. So this is the right fix. The concern however is that it could lead to a sudden change in requests being flagged as abuse. (e.g. a spike in false-positives when using those older LLMs, etc.)

Since discussions about <REDACTED> shouldn't be made in this issue tracker, ping me on Slack. Because of the way <REDACTED> works today, I don't think this will be a problem.

### Opportunistic code cleanups

This PR also made a few refactoring for clarity and maintainability. If you feel I went too far in this regard, or would rather I tease those changes out into a separate PR let me know.

But in general I'm looking for patterns like:

```
if predicate {
    long block of code
} else {
    short block of code
}
```

And inverting the predicate so that the predicate, "short block of code", and the start of the "long block of code" can all be seen together. (Otherwise it arguably is more difficult to read the code as a human.)

Similarly, updating patterns like:

```
if predicate {
   long block of code
   return x
}
return y
```

to just be `if !predicate { return y }`. Then, the "long block of code" does not need to be indented. (e.g. having multiple return statements to avoid overall cyclomatic complexity, if I am using that term right.)

### Next Steps

Provide this PR looks good, then my next step is to follow up and fix https://github.com/sourcegraph/sourcegraph/issues/61278. That is add new configuration knobs like `fireworksconfig.RequestBlockingEnabled` and have fireworks and OpenAI requests be routed through `isFlaggedRequest`.

That shouldn't cause any disruption, and will mean we start flagging abusive requests in paths that we aren't today. But I'd need some help to understand how to roll that out in such a way we can understand if it is causing an unexpected uptick in false-positives, etc.

And, after that, my master plan is to update `isFlaggedRequest` to introduce a new type of flag: "flagged-for-moderation". These would be requests that we want to set aside for an async review (e.g. running it through a proper moderation classifier as part of <REDACTED>.)

## Test plan

Relying on current tests, and aside from trivial things there shouldn't be any functional changes here. (But when I go back to actually start flagging new types of requests, will certainly be adding the relevant tests.)

### DIFF
--- a/cmd/cody-gateway/internal/httpapi/completions/anthropic_test.go
+++ b/cmd/cody-gateway/internal/httpapi/completions/anthropic_test.go
@@ -1,6 +1,7 @@
 package completions
 
 import (
+	"context"
 	"encoding/json"
 	"strings"
 	"testing"
@@ -9,10 +10,18 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"github.com/sourcegraph/log/logtest"
+
 	"github.com/sourcegraph/sourcegraph/cmd/cody-gateway/shared/config"
 
 	"github.com/sourcegraph/sourcegraph/cmd/cody-gateway/internal/tokenizer"
 )
+
+type mockPromptRecorder struct{}
+
+func (*mockPromptRecorder) Record(ctx context.Context, prompt string) error {
+	return nil
+}
 
 func TestIsFlaggedAnthropicRequest(t *testing.T) {
 	validPreamble := "You are cody-gateway."
@@ -22,20 +31,36 @@ func TestIsFlaggedAnthropicRequest(t *testing.T) {
 		PromptTokenBlockingLimit:       20000,
 		MaxTokensToSampleFlaggingLimit: 1000,
 		ResponseTokenBlockingLimit:     1000,
+		RequestBlockingEnabled:         true,
 	}
 	cfgWithPreamble := config.AnthropicConfig{
 		PromptTokenFlaggingLimit:       18000,
 		PromptTokenBlockingLimit:       20000,
 		MaxTokensToSampleFlaggingLimit: 1000,
 		ResponseTokenBlockingLimit:     1000,
+		RequestBlockingEnabled:         true,
 		AllowedPromptPatterns:          []string{strings.ToLower(validPreamble)},
 	}
 	tk, err := tokenizer.NewAnthropicClaudeTokenizer()
 	require.NoError(t, err)
 
+	// Helper function for calling the AnthropicHandlerMethod's shouldFlagRequest, using the supplied
+	// request and configuration.
+	callShouldFlagRequest := func(t *testing.T, ar anthropicRequest, cfg config.AnthropicConfig) (*flaggingResult, error) {
+		t.Helper()
+		anthropicUpstream := &AnthropicHandlerMethods{
+			anthropicTokenizer: tk,
+			promptRecorder:     &mockPromptRecorder{},
+			config:             cfg,
+		}
+		ctx := context.Background()
+		logger := logtest.NoOp(t)
+		return anthropicUpstream.shouldFlagRequest(ctx, logger, ar)
+	}
+
 	t.Run("missing known preamble", func(t *testing.T) {
 		ar := anthropicRequest{Model: "claude-2", Prompt: "some prompt without known preamble"}
-		result, err := isFlaggedAnthropicRequest(tk, ar, cfgWithPreamble)
+		result, err := callShouldFlagRequest(t, ar, cfgWithPreamble)
 		require.NoError(t, err)
 		require.True(t, result.IsFlagged())
 		require.False(t, result.shouldBlock)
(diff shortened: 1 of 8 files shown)


cmd/cody-gateway/internal/httpapi/completions/anthropic_test.go
```diff
@@ -22,20 +31,36 @@ func TestIsFlaggedAnthropicRequest(t *testing.T) {
 		PromptTokenBlockingLimit:       20000,
 		MaxTokensToSampleFlaggingLimit: 1000,
 		ResponseTokenBlockingLimit:     1000,
+		RequestBlockingEnabled:         true,
```
chrsmith, line 34: > Why do we need to set this config value now, when previously the unit tests were fine without it?

Previously the tests were calling `isFlaggedAnthropicRequest` directly, and in this PR that function has been removed and its logic merged into `shouldFlagRequest`.

But within `shouldFlagRequest` there is the line `flaggingResult.shouldBlock = flaggingResult.shouldBlock && ar.config.RequestBlockingEnabled`. (i.e. `shouldFlagRequest` takes that configuration knob into account.) Where `isFlaggedAnthropicRequest` did not.

I would think that maybe this is a sign of a design wart of sorts? What do you think of doing the following:

1. Update `flaggingConfig` to contain a `RequestBlockingEnabled bool` field.
2. Run the `shouldBlock = shouldBlock && cfg.RequestBlockingEnabled` inside of `isFlaggedRequest`.

That way, each implementation of `shouldFlagRequest` doesn't need to do its one-off configuration check. And it's the responsibility of `isFlaggedRequest` instead.

Make sense? What do y'all think?

chrsmith:
Rebased on top of `main` and force-pushed, since this PR was stacked on top of https://github.com/sourcegraph/sourcegraph/pull/61273 and that has since been merged.

chwarwick:
> ⚠️ This does change the behavior for our OpenAI endpoint. We were previously returning a more accurate 501 Not Implemented if trying to use that provider for "chat". But I'd argue that serving the less accurate HTTP error response (400) is probably less important than a simpler function signature for the upstreamHandlerMethods interface.

Quick first pass, this could have some undesirable effects because of the [http client](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@7583026a003eaa04f99e3e6cf7ec31a4a9abb67d/-/blob/internal/httpcli/client.go?L684-689) used in sourcegraph that auto reties.

chrsmith:
> this could have some undesirable effects because of the [http client](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@7583026a003eaa04f99e3e6cf7ec31a4a9abb67d/-/blob/internal/httpcli/client.go?L684-689) used in sourcegraph that auto reties.

There's only one way to find out! 😁 

OK, there are several ways to find out. But I'm going to assume this won't be a problem, since a 400 status code means that by definition the problem is client-side and therefore shouldn't be retried blindly. (At least not with the same payload, headers, etc.)
https://httpwg.org/specs/rfc9110.html#status.400

chwarwick:
> > this could have some undesirable effects because of the [http client](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@7583026a003eaa04f99e3e6cf7ec31a4a9abb67d/-/blob/internal/httpcli/client.go?L684-689) used in sourcegraph that auto reties.
> 
> There's only one way to find out! 😁
> 
> OK, there are several ways to find out. But I'm going to assume this won't be a problem, since a 400 status code means that by definition the problem is client-side and therefore shouldn't be retried blindly. (At least not with the same payload, headers, etc.) https://httpwg.org/specs/rfc9110.html#status.400

Looks like we do retry 400's and default is 20 retries

chrsmith:
> Looks like we do retry 400's and default is 20 retries.

Thanks for keeping me honest @chwarwick.

I may be missing something, but that seems totally broken on multiple dimensions. I'll try to send out a PR and see if we can fix that. But I'll update this PR to rewire the "no seriously, return a 501" status code so that we don't have clients resending the same request 20x times before giving up.

However, the current behavior I think does return a 400 for other validation errors. But I'll double check that. (Since that might have a bad interaction with the HTTP client wrapper we use today, but at least this PR should preserve the current behavior. But obviously I wouldn't want to make things worse.)

chwarwick (approved):
Aside from possible unintended consequences from changing return status code the rest looks good. Deferring to your judgment regrading if dealing with the retries needs to be avoided.

PR #61286, opened 2024-03-20
web: Fix codeintel hovercards not working on first load

### LINKED ISSUE #61282: Codeintel hovercards don't seem to work on initial page load
See [Slack report](https://sourcegraph.slack.com/archives/C05MHAP318B/p1710872479765969)

### LINKED ISSUE #61828: Panics in guardrails still observed

fkling:
Fixes #61282 

Codeintel hovercards are not appearing when the file page initially loads. The codeintel CodeMirror extension is never created because `getOrCreateCodeIntelAPI` never resolves.
That's because the `platform.settings` observable never completes, which is necessary for `lastValueFrom` to work (why it works after navigating to a different page remains a mystery).

The simple fix would be to use `firstValueFrom`, but upon closer inspection, `getOrCreateCodeIntelAPI` is only used by the codemirror blob. I refactored it to avoid having to rely on platform context.

## Test plan

Manual testing. Loading a file page and hovering over a symbol shows a hovercard.


bahrmichael (approved):

camdencheek (approved):
Thanks Felix!

PR #61290, opened 2024-03-20
Refactor URL helpers

fkling:
We currently have quite a few URL helpers for which is not obvious how they are supposed to be used. Additionally we often "prettify" certain parts of URL when generating one but that's easy to forget. 

For example, `formatLineOrPositionOrRange` is defined in `client/web/src/util/url.ts` but is used together with `addLineRangeQueryParameter` or `appendLineRangeQueryParameter` defined in `client/common/util/url.ts`.
There is also `toPositionOrRangeQueryParameter` which does the same as `formatLineOrPositionOrRange` but takes a different input.

This commit attempts to improve this situation in various ways:

- Reduce the API surface. Instead provide a `SourcegraphURL` class that should be used for parsing, manipulating and converting a URL to a string.
- Rename helper functions that operate on `git:` URLs (`makeRepoURI` and `parseRepoURI`) to make their purpose clearer.
- Reduce the scope/visibility of certain helpers.

The general design principals behind `SourcegraphURL` are the following:
- callers shouldn't (have to) make assumptions about what effect certain mutations have (e.g. that `setViewState` changes the hash)
- the API should be obvious to use and difficult to misuse
- converting an instance back to a string should be predictable. E.g. that's why if you pass in only URL search parameters, you get, at most, a URL search fragment back.

The API of the class is not final and I expect it to change as more helpers will be incorporated into this class (e.g. function for parsing repo URLs).

Example:

Before:

```js
const url = appendLineRangeQueryParameter(
  fileURL,
  toPositionOrRangeQueryParameter({
    position: { line: lineNumber + 1, character: offset + 1 },
  })
)
```

After:

```js
const url = SourcegraphURL.from(fileUrl)
  .setLineRange({
    line: lineNumber + 1,
    character: offset + 1,
  })
  .toString()
```  

I feel this API is an improvement but let me know what you think!

I also updated jsdom because a newer version added proper support for `URLSearchParams#size` (which in the end I think I didn't need anymore but it's better when our tools are as spec compliant as possible).

## Test plan

Unit tests

Did manual testing of the redirect logic in `client/web/src/repo/RepositoryFileTreePage.tsx` (which revealed that `SourcegraphURL` cannot be passed to `Navigate`'s `to` directly, for whatever reason).


client/browser/src/shared/code-hosts/shared/util/selections.test.ts
fkling, line 1: These tests have been moved from `common/src/util/url.test.ts`.

client/browser/src/shared/code-hosts/shared/util/selections.ts
```diff
@@ -2,12 +2,45 @@ import { isEqual } from 'lodash'
...
+            line: lpr.endLine || lpr.line,
+            character: lpr.endCharacter || lpr.character || 0,
+        },
+    }
+}
+
+// `lprToRange` sets character to 0 if it's undefined. Only - 1 the character if it's not 0.
+const characterZeroIndexed = (character: number): number => (character === 0 ? character : character - 1)
+
+export function lprToSelectionsZeroIndexed(lpr: LineOrPositionOrRange): Selection[] {
```
fkling, line 24: I moved this here because it was only used inside `browser`.

client/common/src/util/url.test.ts
```diff
@@ -21,136 +134,56 @@ describe('encodeURIPathComponent', () => {
...
-                    },
-                    active: {
-                        line: 4,
-                        character: 0,
-                    },
-                    isReversed: false,
-                },
-            ]
-        )
+describe('parse legacy hash', () => {
```
fkling, line 137: I moved these tests here.

client/common/src/util/url.ts
fkling, line 1: This file contains the most significant changes. Most functions have been made "private" and are used by `SourcegraphURL`.

client/shared/src/util/url.ts
```diff
@@ -178,7 +169,7 @@ const parsePosition = (string: string): Position => {
  * information to fetch the contents of.
  * @deprecated Migrate to using URLs to the Sourcegraph raw API (or other concrete URLs) instead.
  */
-export function parseRepoURI(uri: RepoURI): ParsedRepoURI {
+export function parseRepoGitURI(uri: RepoURI): ParsedRepoURI {
```
fkling, line 172: I renamed this to make the purpose a bit clearer (it's not used to parse "app repo URLs").

client/shared/src/util/url.ts
```diff
@@ -223,6 +214,21 @@ export function parseRepoURI(uri: RepoURI): ParsedRepoURI {
     return { repoName, revision, commitID, filePath: filePath || undefined, position, range }
 }
 
+/**
+ * The inverse of parseRepoGitURI, this generates a string from parsed values.
+ * Example output: `git://github.com/gorilla/mux?SHA#mux.go:3,5-4,9`
+ */
+export function makeRepoGitURI(parsed: ParsedRepoURI): RepoURI {
```
fkling, line 221: Similarly I moved this next to its parse function and renamed it.

client/shared/src/util/url.ts
```diff
@@ -260,163 +266,6 @@ export interface AbsoluteRepoFilePosition
...
-    line = typeof line === 'number' && isNaN(line) ? undefined : line
-    character = typeof character === 'number' && isNaN(character) ? undefined : character
-    if (line === undefined) {
-        return { line: undefined, character: undefined }
-    }
-    return { line, character }
-}
-
 /** Encodes a repository at a revspec for use in a URL. */
 export function encodeRepoRevision({ repoName, revision }: RepoSpec & Partial<RevisionSpec>): string {
```
vovakulikov, line 270: If we're ok with having more complex method in `SourcegraphURL` class, can this URL builder also can be part of these methods?

The problem I have with URLs generation that even if you find `SourcegraphURL` you never know that consumer won't patch the URL in some weird non-standard for us way. We store in URL a lot of important things and it makes it very fragile if you're working with this manually directly in consumer, I remember that we re-wrote some parameters implicitly on the search result page and consumer logic was broken because of this. There is also logic about `buildSearchURLQuery` that seems like very odd piece of URL generation to me every time I have to work with it. Maybe with more common and better API from Felix in `SourcegraphURL` we could simplify it a lot for consumers 

So, proposition for follow up PRs, have a main entry class for working with URL and forbid working with URL manually through eslint. (There should be some exclusions I think for literal link URLs but this should be ok)

cc @fkling @camdencheek

client/web/src/repo/RepositoryFileTreePage.tsx
```diff
@@ -61,24 +61,34 @@ export const RepositoryFileTreePage: FC<RepositoryFileTreePageProps> = props =>
     if (objectType === 'blob' && hashLineNumberMatch) {
         const startLineNumber = parseInt(hashLineNumberMatch[1], 10)
         const endLineNumber = hashLineNumberMatch[2] ? parseInt(hashLineNumberMatch[2].slice(1), 10) : undefined
-        const url = appendLineRangeQueryParameter(
-            location.pathname + location.search,
-            `L${startLineNumber}` + (endLineNumber ? `-${endLineNumber}` : '')
+        // Navigate's to doesn't seem to work with the SourcegraphURL object, so we need to convert it to a string
+        return (
+            <Navigate
+                to={SourcegraphURL.from({ pathname: location.pathname, search: location.search })
```
vovakulikov, line 67: Can we just pass the whole location object here?

client/web/src/repo/blob/panel/BlobPanel.tsx
```diff
@@ -64,11 +64,9 @@ function useBlobPanelViews({
     const location = useLocation()
 
     const position = useMemo(() => {
-        const parsedHash = parseQueryAndHash(location.search, location.hash)
-        return parsedHash.line !== undefined
-            ? { line: parsedHash.line, character: parsedHash.character || 0 }
-            : undefined
-    }, [location.hash, location.search])
+        const lineRange = SourcegraphURL.from(location).lineRange
+        return lineRange.line !== undefined ? { line: lineRange.line, character: lineRange.character || 0 } : undefined
```
vovakulikov, line 68: Can we omit this condition, interface of the lineRage seems the same as this function expect we return?

vovakulikov (commented):
Left some minor comment and suggestions, didn't run it locally yet so just comments now)

Will try it on Monday. So far feels like a right improvement, I like how consumers became more verbose (in a good way) and less confusing, thanks

client/shared/src/util/url.ts
fkling: Yep, that's my goal, I just didn't want to spend too much time to consolidate all functions before getting feedback that we want to do it this way.

I'd love to have only a single `SourcegraphURL` class that handles all URL manipulation/creation.

client/web/src/repo/RepositoryFileTreePage.tsx
fkling: No, because we don't want to use the existing `hash` value.

client/web/src/repo/blob/panel/BlobPanel.tsx
fkling: I don't know, I haven't looked at how `position` is used. In theory `lineRange` can return a range with an end line, here it seems that we only want to get the start position. I kept it close to the original code to be on the safe side.

But yes, maybe `lineRange` could be returned directly.

client/web/src/repo/RepositoryFileTreePage.tsx
vovakulikov: Understood, I thought that helper logic takes into account having hash value and takes search params first if they are both are presented

client/browser/src/shared/backend/extension-api-conversion.tsx
```diff
@@ -1,11 +1,16 @@
 import type { TextDocumentPositionParameters } from '@sourcegraph/client-api'
-import { type AbsoluteRepoFilePosition, toURIWithPath } from '@sourcegraph/shared/src/util/url'
+import { type AbsoluteRepoFilePosition, makeRepoGitURI } from '@sourcegraph/shared/src/util/url'
 
 export const toTextDocumentPositionParameters = (
     position: AbsoluteRepoFilePosition
 ): TextDocumentPositionParameters => ({
     textDocument: {
-        uri: toURIWithPath(position),
+        uri: makeRepoGitURI({
```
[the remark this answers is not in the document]
fkling: Not only but the whole git URI business seems to be legacy. Here too I wanted to focus first on the handful of APIs I'm familiar with before refactoring even more APIs.

client/common/src/util/url.ts
```diff
@@ -95,6 +243,245 @@ export function findLineKeyInSearchParameters(searchParameters: URLSearchParams)
...
+ * This class encapsulates the logic for creating and manipulating Soucegraph URLs.
+ * Not all methods are applicable to all types of URLs. If a method is not applicable
+ * to the URL, the operation will be a no-op.
+ *
+ * See the individual method documentation for details.
+ *
+ * Using this class to manipulate URLs is preferred over manual string manipulation
+ * because it ensures that the URL is prettified when converted to a string.
+ */
+export class SourcegraphURL {
```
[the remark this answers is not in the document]
fkling: I'm happy too move the class to the top, I don't want to move them to funding to separate modules though. IMO it's nice to have them truly private. If they exported from other modules it's too easy for someone to use them.

client/common/src/util/url.ts
```diff
@@ -95,6 +243,245 @@ export function findLineKeyInSearchParameters(searchParameters: URLSearchParams)
...
+     * ```
+     * const url = new SourcegraphURL('/foo?bar')
+     * url.setViewState('references')
+     * url.toString() // => '/foo?bar#L1:2-3:4&tab=references'
+     * ```
+     *
+     * @template V The type that describes the view state (typically a union of string constants).
+     * @param viewState the view state to set
+     */
+    public setViewState<V extends string = string>(viewState: V | undefined): this {
```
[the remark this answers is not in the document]
fkling: Yeah, it's the name I picked up from other functions. We can already change it to something that makes more sense.

client/common/src/util/url.ts
```diff
@@ -95,6 +243,245 @@ export function findLineKeyInSearchParameters(searchParameters: URLSearchParams)
...
+ *
+ * Using this class to manipulate URLs is preferred over manual string manipulation
+ * because it ensures that the URL is prettified when converted to a string.
+ */
+export class SourcegraphURL {
+    private url: URL
+    private hasPathname: boolean
+    private hasOrigin: boolean
+
+    private constructor(url: string | URL) {
```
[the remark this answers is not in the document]
fkling: Absolutely. I'd like to move all repo URL specific helpers into this class eventually.

I actually started out experimenting with some generic route matching generator, that works generate parsing and creation functions from a given path pattern and additional config.
But that felt over engineered, especially since I didn't know how exactly all these helpers are used. So in the end I started with something that I was most familiar with.
But I really want to fold all URL helpers into this class one way or another.

client/web/src/repo/RepositoryFileTreePage.tsx
fkling: I was considering a more "sophisticated" approach, but everything felt like too magical.
Here we know that the hash contains legacy information so it should not be used in the new URL. And the caller shouldn't have to know what exactly happens when they call `setLineRange` or `setViewState` either. The generated URL should be opaque to them.

client/common/src/util/url.ts
```diff
@@ -1,91 +1,239 @@
...
+    if (existingLineRangeKey) {
+        params.delete(existingLineRangeKey)
+    }
+
+    return range !== '' ? new URLSearchParams([[range, ''], ...params]) : params
+}
+
+/**
+ * Tells if the given fragment component is a legacy blob hash component or not.
+ * Legacy fragments have the structure `#L<line>:<character>-<line>:<character>$<viewState>`.
```
camdencheek, line 173: I've always wondered: any chance you know why we moved line number from fragment to query? It always made more sense in the fragment to me

camdencheek (approved):
The motivation makes sense to me. URL construction always makes me a bit nervous, so nice to see it getting a bit better type support. I scanned through it a bit but honestly don't think I can give a review that adds much confidence to this change, so :LGTM:

client/common/src/util/url.ts
fkling: I think that was before my time. I don't know why we did that.

PR #61294, opened 2024-03-20
Feature flags: allow users to control their own overrides

camdencheek:
Currently, only site admins can add overrides. We'd like to use feature flag overrides to opt out of the svelte prototype. This changes it so users can add overrides to their own account.


camdencheek:
Related to https://github.com/sourcegraph/sourcegraph/issues/61298

fkling (commented):
Well, that seemed easy :)

fkling (approved):

PR #61296, opened 2024-03-20
Fix result container collision (fix search result page)

vovakulikov:
Hotfix for regression after this PR https://github.com/sourcegraph/sourcegraph/commit/97d0fae787428074b1759313bf5592fd22c5c6ab

Apparently css modules files with the same filename generates the same class hash in esbuild 

## Test plan
- Check that Search result page (react version) looks ok


camdencheek (approved):
Oh haha, beat me to it. Thanks Vova!

PR #61297, opened 2024-03-20
Remove profanity and phrase detection 

rafax:
Remove profanity and phrase detection - it makes the code more complex (one more thing we do "just-for-vintage-anthropic"), makes API calls slightly slower, and (as far as I know) is not used by anyone. 

I did take a quick look at collected data, and many of the detected phrases are substrings of valid words (like parse or class), so it's not very useful overall.

## Test plan

- unit and e2e tests pass


chrsmith (approved):
LGTM. +1 for removing an external dependency, and +10 since this is largely redundant with the blocklist we added later. (And profanity isn't necessary a red flag, so much as more specific terms as well.)

chwarwick (approved):
I noticed the same things being captured.

PR #61301, opened 2024-03-20
Search: make lang filters always case insensitive

### LINKED ISSUE #60575: Add .C and .H files to C++ file types

Currently, `.c` and `.h` files are registered as C++ file types. In order for `.C` and `.H` files to be returned when filtering on `lang:C++` case sensitivity needs to be turned off, but C++ is a case-sensitive language so the rest of the search will contain false positives (e.g. enumField and EnumField are different) 

#### Is your feature request related to a problem? If so, please describe.

Some customers who use `.C` and `.H` files get false positives when searching with case-sensitivity disabled in order for those files to be included as C++ files. For example, `lang:C++ enumField` with case sensitivity turned **on**, will not include `.C` and `.H` files. When case sensitivity is turned **off**, then results for both `EnumField` and `enumField` will be returned, which includes false positives, as `EnumField` differs from `enumField`. 

#### Describe alternatives you've considered.

An alternative is to use regex search with `(file:.*\.C OR file:.*\.H)`, with case sensitivity enabled.

#### Additional context

go-enry file extensions listed here: https://github.com/go-enry/go-enry/blob/v2.8.4/data/extension.go

mmanela:
This change updates file extension lang matching to be case-insensitive. This is more flexible and consistent with GitHub search also.  

Fixes #60575

## Test plan
- Ensure unit tests of all impacted compontents are updated and passing
- Manually test locally  for both content based and file based lang matching


internal/search/zoekt/query_test.go
```diff
@@ -143,7 +143,7 @@ func TestQueryToZoektQuery(t *testing.T) {
 			}
 
 			if !queryEqual(got, zoektQuery) {
-				t.Fatalf("mismatched queries\ngot  %s\nwant %s", got.String(), zoektQuery.String())
+				t.Fatalf("mismatched queries during [%s] \ngot  %s\nwant %s", tt.Name, got.String(), zoektQuery.String())
```
mmanela, line 146: @jtibshirani  I am having trouble getting this test to pass due to this case

```json
		{
			Name:           "Languages get passed as file filter",
			Type:           search.TextRequest,
			Query:          `file:\.go$ lang:go`,
			WantZoektQuery: `file:"\\.go(?m:$)" file:"\\.go(?m:$)"`,
		},
```

I am unable to get the parsed `WantZoektQuery` to match. As is the output is this

```
            got  (and file_regex:"(?m:\\.go$)" file_regex:"(?im:\\.GO$)")
            want (and file_regex:"(?m:\\.go$)" file_regex:"(?m:\\.go$)")
 ```
 
 But if I update the `WantZoektQuery` to the following 
 ```
 WantZoektQuery: `file:"\\.go(?m:$)" file:"(?i)\\.go(?m:$)"`
 ```
 
 
 The result is this
 ```
             got  (and file_regex:"(?m:\\.go$)" file_regex:"(?im:\\.GO$)")
            want (and file_regex:"(?m:\\.go$)" case_file_regex:"(?im:\\.GO$)")
 ```
 
 Notice zoekt now changes `file_regex `to `case_file_regex`

Comment (jtibshirani) [reply internal/search/zoekt/query_test.go:146]: @mmanela and I caught up offline. We will update `WantZoektQuery` to directly check the Zoekt query's string output instead of using the custom Zoekt query language as it does now. That will make all these tests less fragile and easier to work with.

File: internal/search/zoekt/query_test.go [hunk 2]
```diff
@@ -16,77 +17,83 @@ import (
...
+			Name:            "Languages get passed as file filter",
+			Type:            search.TextRequest,
+			Query:           `file:\.go$ lang:go`,
+			WantZoektOutput: `(and file_regex:"(?m:\\.go$)" file_regex:"(?im:\\.GO$)")`,
+		},
+		{
+			Name:            "Languages still use case_insensitive in case sensitivity mode",
+			Type:            search.TextRequest,
+			Query:           `file:\.go$ lang:go case:true`,
+			WantZoektOutput: `(and case_file_regex:"(?m:\\.go$)" case_file_regex:"(?im:\\.GO$)")`,
```
mmanela, line 96: @jtibshirani  Per your feedback, this is the one wart of this approach. I validated manually in UX that this does work correctly. But it showd the contradiction. Zoekt thinks the file match regex is case sensitive (since it says `case_file_regex` ) BUT inside of that regex is a `(?i)` which undoes that and allows us to match all cases. 

If we didn't want that there are two options
1. In zoekt query we can sniff the `(?i)` prefix and change the sensitivity of that match. 
2. Change the data type of Include/ExcludePatterns to not just be string but a struct with more info


I am not sure either are "worth" it right now but open for feedback :)

mmanela: But how do we know it's a language filter? Since at this point it's just a file path regex , Unless we do my #2 above

jtibshirani: Indeed, I was suggesting something similar to your (2) above. I'd say -- let's give it a try since it's the most solid thing to do, and see how simple we can get it! If it proves messy, we can just make sure it's well tested and move on.

mmanela: I looked into this and discussed with @jtibshirani  and we decided not to try to change Include/ExcludePatterns since in reality this case is meant to not be a file pattern and the right solution is to fully enable content based support support. So, we will go with this solution for now and document this as another reason to fully finish content based lang support and remove the older support.

internal/search/query/helpers.go
```diff
@@ -28,6 +28,15 @@ func UnionRegExps(values []string) string {
 	return "(?:" + strings.Join(values, ")|(?:") + ")"
 }
 
+func CaseInsensitiveRegExp(value string) string {
+	// Don't add (?i) if it is already present.
```
jtibshirani, line 32: Checking I understand -- this is just to keep the regexes simple, but isn't needed for correctness?

internal/search/zoekt/query_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"slices"
 	"testing"
 
+	gpCmp "github.com/google/go-cmp/cmp"
```
jtibshirani, line 8: I believe we can delete `queryEqual` and `zoektQStringCompare`, so then we don't need the `cmp` import above. And then this doesn't need to be a named import!

internal/search/job/jobutil/exhaustive_job_test.go
```diff
@@ -35,6 +35,30 @@ func TestNewExhaustive(t *testing.T) {
 		WantPager autogold.Value
 		WantJob   autogold.Value
 	}{
+		{
+			Name:  "case insensitive lang match",
```
jtibshirani, line 39: Maybe we can make this test turn on case sensitivity, since this was the source of the bug -- language filters with a case sensitive query.

jtibshirani (commented):
This looks good to me! Double-checking -- did you manually test both indexed and unindexed search, which you can do by searching an older commit or using the query parameter `index:no`? (I wish we had integration tests that cover both frontend + Zoekt, but I don't think we do!)

internal/search/query/helpers.go
mmanela: We were appending this string in lots of places and I figured to add a helper and make sure its never going to add twice (but it would still work fine in that case)

mmanela:
@jtibshirani  Yes I did test with index:no

internal/search/zoekt/query_test.go
mmanela: Good catch

internal/search/job/jobutil/exhaustive_job_test.go
mmanela: Good call

jtibshirani (approved):

jtibshirani:
Let's merge! I just changed the title to match our usual format.

PR #61306, opened 2024-03-20
Log changes to a user's Cody Gateway limits

chrsmith:
Currently there is no guaranteed log line written when we update a user's Cody Gateway quota. (Perhaps we usually write something to the "Security Event Logs", but it's unclear what `auditlog-expansion` governs and if it is set for sourcegraph.com.)

This PR just adds a log line so that you can quickly grep through GCP logs to see a record of changes made to user quotas.

## Test plan

NA

### DIFF
--- a/cmd/frontend/graphqlbackend/user.go
+++ b/cmd/frontend/graphqlbackend/user.go
@@ -873,6 +873,7 @@ func (r *schemaResolver) SetUserCompletionsQuota(ctx context.Context, args SetUs
 	if err := auth.CheckCurrentUserIsSiteAdmin(ctx, r.db); err != nil {
 		return nil, err
 	}
+	requestingActor := actor.FromContext(ctx)
 
 	if args.Quota != nil && *args.Quota <= 0 {
 		return nil, errors.New("quota must be 1 or greater")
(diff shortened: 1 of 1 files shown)


thenamankumar (approved):

cmd/frontend/graphqlbackend/user.go
```diff
@@ -897,9 +897,12 @@ func (r *schemaResolver) SetUserCompletionsQuota(ctx context.Context, args SetUs
 	if err := r.db.Users().SetChatCompletionsQuota(ctx, user.ID, quota); err != nil {
 		return nil, err
 	}
-	if featureflag.FromContext(ctx).GetBoolOr("auditlog-expansion", false) {
 
-		// Log an event when a user's Completions quota is updated
+	// Log that the user's completions quota was updated.
+	r.logger.Info("setting user completions quota",
+		log.Int("userID", int(user.ID)),
+		log.Int32p("quota", args.Quota))
```
rafax, line 904: It'd be useful to log the old value (for people that already have an override) and the actor performing the change (banbot? support? other site admin?)

chrsmith: Ah yeah, good suggestions! I'll add those and merge.

PR #61309, opened 2024-03-20
svelte: Inject window.context when serving the Svelte index.html page

### LINKED ISSUE #61307: Embed JSContext in index.html
Even if it's not used in the app yet, this would fix an issue with running `sg start web-standalone` against S2 while it is configured to serve the SvelteKit app by default.

At the moment the page errors because the devserver is setup to extract `window.context` from the origin page and embed it into its own local index.html page. Since the server sends the SvelteKit app, `window.context` is not present.
To other devs this error won't be obvious and the only way to fix this is to opt-out of SvelteKit on S2.

Since we want to us `window.context` eventually we might as well spent the time now to embed it into the SvelteKit app.

fkling:
Closes #61307 

This commit extends the server to parse SvelteKit's index.html file as a template and inject `window.context`.
The SvelteKit app doesn't use window.context yet but this change fixes an issue when running `sg start web-standalone` against S2 and S2 is configured to serve the SvelteKit app.

Couple of notes:

- I originally wanted to use a single `app.html` file for both "integrated" and standalone development. Having two separate files is not ideal, but was the easier way.
- I mostly copied the logic for generating and caching the template from the code that serves the esbuild bundle. There didn't seem to be an obvious way to make that reusable.

## Test plan

Manual testing by running

```
sg start enterprise
pnpm dev:enterprise
```

and inspecting the HTML returned by `https://sourcegraph.test:3443`

I couldn't verify that it indeed works with `sg run web-standalone` because I can't run an instance and the standalone web app, but I looked at our esbuild devserver setup and verified that we indeed extract `window.context` from the origin server response.


client/web-sveltekit/package.json
```diff
@@ -4,6 +4,7 @@
   "scripts": {
     "dev": "vite dev",
     "dev:dotcom": "vite dev --mode=dotcom",
+    "dev:enterprise": "DEPLOY_TYPE=dev vite build --watch",
```
fkling, line 7: I added this make it easier to create a SvelteKit build that works with `sg start enterprise` but we should really spend some time to make this part of the `sg start enterprise` command itself.

cmd/frontend/internal/app/ui/sveltekit.go
```diff
@@ -57,13 +63,51 @@ func useSvelteKit(r *http.Request) bool {
 	return false
 }
 
+var (
+	loadSvelteKitTemplateMu sync.RWMutex
+	loadSvelteKitTemplateCache = map[string]*template.Template{}
+)
+
+func loadSvelteKitTemplate(name string) (*template.Template, error) {
```
camdencheek, line 71: is `name` ever anything other than `index.html`? If not, we can simplify this quite a bit by just wrapping this in a [sync.OnceValues](https://pkg.go.dev/sync#OnceValues).

cmd/frontend/internal/app/ui/sveltekit.go
```diff
@@ -57,13 +63,51 @@ func useSvelteKit(r *http.Request) bool {
...
+	loadSvelteKitTemplateMu.Lock()
+	loadSvelteKitTemplateCache[name] = tmpl
+	loadSvelteKitTemplateMu.Unlock()
+	return tmpl, nil
+}
+
 // renderSvelteKit writes SvelteKit's fallback page to the provided writer
-func renderSvelteKit(dst io.Writer) error {
-	file, err := assets.Provider.Assets().Open("_sk/index.html")
+func renderSvelteKit(dst io.Writer, data any) error {
```
camdencheek, line 98: why is `data` typed as `any` here? Shouldn't we have a known type for this?

camdencheek (approved):

cmd/frontend/internal/app/ui/sveltekit.go
fkling: I copied this from `renderTemplate` (in `tmpl.go`). 🤷🏻

camdencheek: 🙃

cmd/frontend/internal/app/ui/sveltekit.go
fkling: Ah, but what about !env.InsecureDev though? It seems we don't want to cache when in dev mode? But maybe that's not what that env var is doing.

camdencheek: Huh. I guess that's so you can change the assets live without restarting the process?

fkling: Yep... how would that work with the "OnceValues" thing? I guess I could define the template loading function separately and either call it directly or the retuned function from OnceValues.

camdencheek: Yeah, that's how I would probably do it at least. But also, feel free to ignore. Dev env stuff makes things messy

Issue #61221, opened 2024-03-18
Svelte: Add general wildcard resizable panel components


vovakulikov:
We already have some first implementation of panels, and this is how filter panel and file preview panel work on the search result page. But the problem with current implementation that these panels are disconnected and don't change the size of other panels on the page as you resize them. This leads to the weird states when you can collapse one of the panel completely. 

Also in the near future (see [Figma design](https://www.figma.com/file/EPgl0qVbqGSumvf52brhda/Code-Search?type=design&node-id=1397%3A21372&mode=design&t=5WGzUUfQ3itkH70C-1)) the whole blob UI page will be built on top of panels layouts with panel groups and nested panel. 

We can use https://react-resizable-panels.vercel.app as an implementation reference

PR #61319, opened 2024-03-21
Svelte [wildcard]: Add svelte resizable panels

vovakulikov:
Closes https://github.com/sourcegraph/sourcegraph/issues/61221

This PR adds more general approach to resizable panel UI compared to what we have in react and in svelte project so far. As reference implementation picks one of react library for panels ([link here](https://react-resizable-panels.vercel.app/)). In most helper functions this PR basically just ports implementation (some files in utils directory). 

Code base isn't small, so I tried to split this into commits in a way it makes sense, so I would recommend to go commit by commit 

[Loom video](https://www.loom.com/share/23cd70e8cc7747e59166423fa4c1ae69?sid=7e4a93e2-8ac0-420f-9b90-fe03a4cc175a) 

I highly recommend to check out the demo of this component and share your feedback about this. A few details about how it works
- All panels are built on top of flex box layout 
- All panels use relative (percentage) values
- Each panel is configurable meaning you can set default/min/max sizes in relative values
- Panels are shrinkable, it means that if we have 3 panels (A, B, C) and we increase panel C we will decrease panel B until it reaches its minSize, than we start decrease panel A.
- It does support nested panels (so from my point of view we're ready to implement new blob UI page layout with panels)
- Implementation is ready to support keyboard resizing but I omit this part in this PR (it's too big already)
- Size persisting will be added in further PRs (code base is ready for it just didn't want to increase big PR even more)

## Test plan
- Check CI 
- Check storybook demo (check different browsers, big three Safari, Firefox, Chrome)
- Since this just adds yet another component to svelte wildcard this all checks at this time


client/web-sveltekit/src/lib/wildcard/resizable-panel/utils/determinePivotIndices.ts
```diff
@@ -0,0 +1,11 @@
+import { getResizeHandleElementIndex } from '../dom/getResizeHandleElementIndex'
+
+export function determinePivotIndices(
```
vovakulikov, line 3: Since our Layout is just list of numbers, calculating index is important when it comes to resizing. 
This helper finds index of resize panels that arounds resize handler which is moving

client/web-sveltekit/src/lib/wildcard/resizable-panel/utils/adjustLayoutByDelta.test.ts
fkling, line 1: What's up with the (duplicate ?) test names?

fkling (commented):
Left a couple of comments in-line. I mostly looked at the Svelte files.

client/web-sveltekit/src/lib/wildcard/resizable-panel/Panel.svelte
```diff
@@ -0,0 +1,52 @@
+<script lang="ts">
+    import { getContext, onDestroy } from 'svelte'
+    import classNames from 'classnames'
+    import type { PanelGroupContext } from './types'
+    import { getId } from './utils/common'
+
+    export let id: string | null = null
+    export let className: string = ''
```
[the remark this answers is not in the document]
camdencheek: Yeah, particularly with a Panel component, I'd expect it _only_ handles layout. Any custom styling should be added to a wrapper element IMO.

client/web-sveltekit/src/lib/wildcard/resizable-panel/PanelResizeHandle.svelte
```diff
@@ -0,0 +1,91 @@
...
+    onMount(() => {
+        const resizeHandler = getResizeHandler(resizeHandleId)
+
+        assert(element, 'Element ref not attached')
+
+        const setResizeHandlerState = (action: ResizeHandlerAction, isActive: boolean, event: ResizeEvent) => {
+            if (isActive) {
+                switch (action) {
+                    case 'down': {
+                        stateStore.set('drag')
```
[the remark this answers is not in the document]
vovakulikov: I tried just common variable  assignment statement like `stateStore = 'drag'` I thought that these variables should work reactively as long as they stay within one component

client/web-sveltekit/src/lib/wildcard/resizable-panel/Panel.svelte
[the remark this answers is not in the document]
vovakulikov: Ok, it was my last change to do class overriding, go with data attribute (they already exist there (data-panel-group, data-panel ..etc)

client/web-sveltekit/src/lib/wildcard/resizable-panel/utils/adjustLayoutByDelta.test.ts
vovakulikov: This is just how test look in the original library, because they all about increasing size of the first panel, all other things are just different inputs to the function to simulate different cases, we could describe cases but it would be too long. 

I can adjust this if it's confusing

vovakulikov:
@fkling resolved all comments but unit tests, I can name all test with more human readable way but I think it's not a blocker, let me know if there is anything else, thanks for review

client/web-sveltekit/src/lib/wildcard/resizable-panel/Panel.svelte
```diff
@@ -0,0 +1,50 @@
...
+    )
+
+    $: styles = getPanelStyles(panelId)
+</script>
+
+<div
+    class="panel"
+    style={$styles}
+    bind:this={panelElement}
+    {id}
```
fkling, line 33: Should this be `id={panelId}` instead?

client/web-sveltekit/src/lib/wildcard/resizable-panel/PanelGroup.svelte
```diff
@@ -0,0 +1,213 @@
...
+            // TODO [VK]: Add auto saved layout support here.
+
+            // Validate even saved layouts in case something has changed since last render
+            // e.g. for pixel groups, this could be the size of the window
+            const nextLayout = validatePanelGroupLayout(
+                unsafeLayout,
+                panels.map(panelData => panelData.constraints)
+            )
+
+            layoutStore.set(nextLayout)
```
fkling, line 58: Up to you which API to use, it doesn't make a difference. Just want to point out that you *could* write `$layoutStore = nextLayout` as well (same for other uses of `store.set`).

client/web-sveltekit/src/lib/wildcard/resizable-panel/PanelResizeHandle.svelte
```diff
@@ -0,0 +1,87 @@
...
+                        resizeHandler(event)
+                        break
+                    }
+                    case 'up': {
+                        $state = 'hover'
+                        stopDragging()
+                    }
+                }
+            } else {
+                state.set('inactive')
```
fkling, line 50:
```suggestion
                $state = 'inactive'
```

for consistency

client/web-sveltekit/src/lib/wildcard/resizable-panel/ResizablePanels.stories.svelte
```diff
@@ -0,0 +1,130 @@
...
+        padding: 0.5rem;
+        font-size: 1rem;
+        display: flex;
+        overflow: hidden;
+        height: 100%;
+    }
+
+    :global([data-resize-handle]) {
+        flex: 0 0 3px;
+    }
```
fkling, line 115: Could we make that element be `flex: 0 0 content` by default? Or is there a reason not do that?

client/web-sveltekit/src/routes/search/FileSearchResultHeader.svelte
fkling, line 1: I think you need to rebase on main, the `f` was removed.

fkling (approved):
Great! Regarding the tests, I was just confused that there are multiple tests with the same name (or maybe I didn't look properly). That doesn't seem to make it easy to identify when a test fails.

client/web-sveltekit/src/lib/wildcard/resizable-panel/PanelResizeHandle.svelte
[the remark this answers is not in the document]
vovakulikov: Sorry I meant that I used just a common variable assignment 
`state = "drag"`, like in svelte hello world counter example. No store or any other reactive primitives

client/web-sveltekit/src/lib/wildcard/resizable-panel/ResizablePanels.stories.svelte
vovakulikov: it would require to always provide child element to the resize handler component, otherwise handler will collapse since there is no element. I will add it be default but just note that this won't remove these rule `flex: 0 0 3px;` here

vovakulikov: Just realised that this won't work because there is no easy way to override class styles with data attribute rule from outside. So this means that having will always require child element.

fkling: You could also use CSS variables to allow overriding specific styles. E.g. `--resize-handler-width`, `--resize-handle-background-color`, `--resize-handle-hover-background-color`.

vovakulikov: Yeah, CSS API is better here I think than override. Will add this when I will be migrating search result page panels 😉, thanks for suggestions

client/web-sveltekit/src/lib/wildcard/resizable-panel/types.ts
```diff
@@ -0,0 +1,50 @@
...
+    id: PanelId
+    order: number | undefined
+    constraints: PanelConstraints
+    getPanelElement: () => HTMLElement
+}
+
+export interface PanelConstraints {
+    defaultSize?: number | undefined
+    maxSize?: number | undefined
+    minSize?: number | undefined
```
camdencheek, line 27: Do these have to be percentages? I think it'd be pretty common that I have a panel with a logical min width that is fixed (e.g. it looks bad if it's less than 20rem).

client/web-sveltekit/src/lib/wildcard/resizable-panel/types.ts
```diff
@@ -0,0 +1,50 @@
...
+    collapsedSize?: number | undefined
+    collapsible?: boolean | undefined
```
camdencheek, line 29: Why does the panels element need to know about collapse behavior? Couldn't this just be reactive behavior? e.g. update `minSize` or swap out the component when the user clicks the collapse button

vovakulikov:
Next step is to migrate search result page to panels components, I'm out today for search web issues (busy with Cody web) if anyone wants to do it go for it 😉 (otherwise I will do it myself begging of the next week so don't feel obligated)

client/web-sveltekit/src/lib/wildcard/resizable-panel/utils/common.ts
```diff
@@ -0,0 +1,32 @@
+import type { PanelId, PanelInfo } from '../types'
+
+let counter = 0
+
+export function getId(): string {
+    counter++
+    return counter.toString()
+}
```
camdencheek, line 8: If our IDs are generated on the fly, doesn't that mean they are nondeterministic? Will that make it difficult to persist the layout (which we don't do yet, right?)

client/web-sveltekit/src/lib/wildcard/resizable-panel/ResizablePanels.stories.svelte
```diff
@@ -0,0 +1,130 @@
...
+
+<Story name="Horizontal panels">
+    <section class="root">
+        <PanelGroup direction="horizontal">
+            <Panel defaultSize={30} minSize={20}>
+                <div class="item">left</div>
+            </Panel>
+            <PanelResizeHandle>
+                <div class="handle" />
+            </PanelResizeHandle>
```
camdencheek, line 21: Could we make this more ergonomic with slots? 

e.g. instead of having to wrap our own class with `<PanelResizeHandle>`, could we just do `<div class="handle" slot="handle"/>`? Then we wouldn't need to make `PanelResizeHandle` a public part of the API and it would let us provide a default style which can be overridden with a slotted component.

camdencheek (commented):
Oops, just a little too late 😄 Nothing blocking anyways.

General question: how do you feel about the complexity of this? It's a pretty big positive diff. I did try to at least scan through all the javascript to make sure I understand what it's doing. Most of it is fairly approachable, but there is definitely some hackery mixed in.

fkling:
I was surprised to see such a huge diff. Luckily most of it is tests.
I also wonder if it needs to be this complex, but I've not spent any time on this so I'm not familiar with the layout problem. Maybe it is just a complex problem.

client/web-sveltekit/src/lib/wildcard/resizable-panel/types.ts
vovakulikov: Correct, all panels are controlled by relative values.

> e.g. it looks bad if it's less than 20rem

It works both ways, I can resize my browser's window in a way that 20rem also looks bad because it pushes other panels on the page as browser window becomes smaller.

In theory we could support a proper fixed sizes logic but it would be even more tricky than relative values. 

The idea behind relative panel minWidth is basically what we have we flexbox layout. I think this what we want here but let's see how it would work with real UI. Would be happy to change it.

client/web-sveltekit/src/lib/wildcard/resizable-panel/types.ts
vovakulikov: Collapse here means a bit different thing, as panels becomes smaller and reaches minSize there a way to collapse this panel automatically to collapse size. It's supported out of the box by panel component.

client/web-sveltekit/src/lib/wildcard/resizable-panel/utils/common.ts
vovakulikov: Correct, if you want persist logic active you have to pass id and order props.

client/web-sveltekit/src/lib/wildcard/resizable-panel/types.ts
camdencheek: Yeah, for sure. Let's give it a try 🙂 At worst, I think we could probably set a `min-width` on the inner element if we needed

client/web-sveltekit/src/lib/wildcard/resizable-panel/ResizablePanels.stories.svelte
vovakulikov: I actually don't know. @fkling ideas? I thought that svelte slots should be static. So just having slot on element won't give you its ref for example, also note that ResizeHandle component contains resize logic calls, I would like to have it separately from panel group component

PR #61322, opened 2024-03-21
svelte: Persist theme choice

fkling:
Until now selecting a different theme didn't persist it in temporary settings. This commit fixes it. I first tried an approach via context (hence the changes to setting the context) but that didn't work as expected (but the changes are still an improvement).

I also added a small piece of JavaScript to the HTML pages to set the selected theme as soon as possible. This does not have any visible effect yet because the style sheets are loaded lazily but in the future we could add a set of minimal styles to set the background color.

## Test plan

Manual testing.

### DIFF
--- a/client/web-sveltekit/src/lib/theme.ts
+++ b/client/web-sveltekit/src/lib/theme.ts
@@ -1,17 +1,73 @@
-import { derived, writable, type Readable } from 'svelte/store'
+import { type Writable, writable, type Updater, derived, type Readable } from 'svelte/store'
+import { temporarySetting, type TemporarySettingStore } from './temporarySettings'
 
-import { createMappingStore } from './utils'
+const LOCAL_STORAGE_THEME_KEY = 'sourcegraph-theme'
 
 export enum Theme {
-    Light,
-    Dark,
-    System,
+    Light = 'Light',
+    Dark = 'Dark',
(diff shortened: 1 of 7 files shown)


bahrmichael:
Lgtm, but i can't figure out how to approve on mobile.

bahrmichael (approved):
Lgtm

client/web-sveltekit/src/lib/theme.ts
```diff
@@ -1,17 +1,73 @@
...
+ * to temporary settings.
  */
-export const theme = writable(Theme.System)
+export const theme: Writable<Theme> = (function () {
+    let theme: Theme = (localStorage.getItem(LOCAL_STORAGE_THEME_KEY) ?? Theme.System) as Theme
+    let themeSettingStore: TemporarySettingStore<'user.themePreference'> | undefined
+
+    const { subscribe } = writable(theme, set => {
+        if (!themeSettingStore) {
+            themeSettingStore = temporarySetting('user.themePreference')
```
vovakulikov, line 22: Not a problem of this PR, but just a reminder for us for future time, we have to revisit temporal settings storage, since it has internally async API but it tries to mimic sync api in consumers it leads to inconsistent flickering when we're getting overrides value in response to save client value.

PR #61332, opened 2024-03-21
Enable flagging for OpenAI-routed LLM requests

chrsmith:
This PR enables our flagging/abuse-detection logic for requests sent to OpenAI.

To do this, it was only a matter of updating the `OpenAIHandlerMethods`'s `shouldFlagRequest(...)` function. And having it call `isFlaggedRequest(...)`. However, the bulk of this PR is to do the wiring in such a way that making the necessary configuration data available makes sense.

## Changes

1. Introduce `config.FlaggingConfig`

All of our anti-abuse parameters are part of `AnthropicConfig`. Rather than add the same fields to `OpenAIConfig` (and later, other LLM configuration objects), I moved them to a stand-alone `FlaggingConfig` struct.

These flagging configuration values can then be loaded in-bulk by calling this function:

```go
// loadFlaggingConfig loads the common set of flagging-related environment variables for
// an LLM provider. The expectation is that the env vars all share the provider-specific
// prefix, and a flagging config-specific suffix.
//
// IMPORTANT: Some of the env vars loaded are _required_. So be sure that they are all
// set before calling loadFlaggingConfig for a new LLM provider.
func (c *Config) loadFlaggingConfig(cfg *FlaggingConfig, envVarPrefix string) {
```

So all of the configuration loading for Anthropic is just a matter of calling:

```go
// Load configuration settings specific to how we flag Anthropic requests.
c.loadFlaggingConfig(&c.Anthropic.FlaggingConfig, "CODY_GATEWAY_ANTHROPIC")
```

2. Add `FlaggingConfig` to `OpenAIConfig`

Next, we just add the same `FlaggingConfig` bag of settings to `OpenAIConfig`. So the same configuration settings are available there, too.

However, the question becomes how do we _load_ these values? Do we want to add `CODY_GATEWAY_OPENAI_BLOCKED_PROMPT_PATTERNS`, and friends? That would be super-annoying to maintain. So to start with, I just load the same Anthropic settings for OpenAI.

Chiefly:

```go
// Load configuration settings specific to how we flag OpenAI requests.
//
// HACK: Rather than duplicate all of the env vars to independently control how we flag
// Anthropic and OpenAI models, we are just reusing the same settings for Anthropic.
// If we follow this better, we should just rename the env vars to something general, e.g.
// "CODY_GATEWAY_FLAGGING_" and just load a single, shared flagging config.
c.loadFlaggingConfig(&c.OpenAI.FlaggingConfig, "CODY_GATEWAY_ANTHROPIC")
```

I believe the long-term solution is to tease out all of these abuse-related env vars, into their own specific configuration bag. (e.g. `CODY_GATEWAY_FLAGGING`.) And only wire through LLM-specific env vars on an as-needed basis.

⚠️ Some ramifications of this:

- It assumes that the "tuning" for the two LLMs is more or less the same. e.g. if we set a custom `RESPONSE_TOKEN_BLOCKING_LIMIT` for Anthropic, that the same limit would apply to OpenAI too. However, that cannot be true because they would use different tokenizers.

In fact, isn't it possible that different LLM _models_ from the same provider could use different tokenizers?) I'm not sure if there is a great, one-size-fits-all type solution here.

- More importantly, this also means that we will -- unless we add some more logic, or update our `.tf` configuration -- have `REQUEST_BLOCKING_ENABLED=true` for OpenAI. (Since it is the case for Anthropic at the moment.) That may be risky, and a more conservative approach would be to set `REQUEST_BLOCKING_ENABLED=false` for OpenAI, manually inspect flagged requests from that model, and then enable it.

Since there's an ongoing REDACTED involving REDACTED, I'm inclined to just enable this. I am guessing that our existing abuse flagging isn't aggressive enough to disrupt legitimate OpenAI traffic. Even if things like token counts have different meanings. (e.g. the wiggle room between legitimate traffic and abusive traffic is X%, and X% of tokens is larger than any disparity between Anthropic and OpenAI tokenizers. But I could be totally wrong here.)

3. Update callers of `isFlaggedRequest` and the function itself

Rather than having `isFlaggedRequest` take a `config.FlaggingConfig` object directly (which wouldn't be too unreasonable), I instead kept the package's `flaggingConfig` struct.

That is nearly identical as `config.FlaggingConfig`, but is missing `MaxTokensToSample`. I added some comments here and there to clarify that surprising fact. (e.g. pointing out the difference between and `MaxTokensToSample` and `MaxTokensToSampleFlaggingLimit`, since auto-completion might trick you into using the wrong one.)

4. Add some tests specifically for `isFlaggedRequest`

Finally, just add some specific tests for `isFlaggedRequest`. Since the tests we have today (which are still valuable!) add a lot of ceremony to go through the `AnthropicHandlers` interface, and then call the `upstreamMethods.shouldFlagRequest`, etc.

And now that things are a bit more decoupled, it's far easier to just test the logic within `isFlaggedRequest` directly.

## Test plan

Added `flagging_test.go`, and generally cleaned things up. But we still do not have more robust end-to-end or integration type tests, for how the `http.Handler` that processes these requests work.

But we're getting closer. And hopefully soon we can do that without it being a major slog.

### DIFF
--- a/cmd/cody-gateway/internal/httpapi/completions/anthropic.go
+++ b/cmd/cody-gateway/internal/httpapi/completions/anthropic.go
@@ -132,27 +132,20 @@ func (a *AnthropicHandlerMethods) getAPIURLByFeature(feature codygateway.Feature
 }
 
 func (a *AnthropicHandlerMethods) validateRequest(ctx context.Context, logger log.Logger, _ codygateway.Feature, ar anthropicRequest) error {
-	if ar.MaxTokensToSample > int32(a.config.MaxTokensToSample) {
-		return errors.Errorf("max_tokens_to_sample exceeds maximum allowed value of %d: %d", a.config.MaxTokensToSample, ar.MaxTokensToSample)
+	maxTokensToSample := a.config.FlaggingConfig.MaxTokensToSample
+	if ar.MaxTokensToSample > int32(maxTokensToSample) {
+		return errors.Errorf("max_tokens_to_sample exceeds maximum allowed value of %d: %d", maxTokensToSample, ar.MaxTokensToSample)
 	}
 	return nil
 }
 
 func (a *AnthropicHandlerMethods) shouldFlagRequest(ctx context.Context, logger log.Logger, ar anthropicRequest) (*flaggingResult, error) {
-	cfg := a.config
 	result, err := isFlaggedRequest(a.anthropicTokenizer,
 		flaggingRequest{
 			FlattenedPrompt: ar.Prompt,
 			MaxTokens:       int(ar.MaxTokensToSample),
 		},
-		flaggingConfig{
-			AllowedPromptPatterns:          cfg.AllowedPromptPatterns,
-			BlockedPromptPatterns:          cfg.BlockedPromptPatterns,
-			PromptTokenFlaggingLimit:       cfg.PromptTokenFlaggingLimit,
-			PromptTokenBlockingLimit:       cfg.PromptTokenBlockingLimit,
-			MaxTokensToSampleFlaggingLimit: cfg.MaxTokensToSampleFlaggingLimit,
-			ResponseTokenBlockingLimit:     cfg.ResponseTokenBlockingLimit,
-		},
+		makeFlaggingConfig(a.config.FlaggingConfig),
 	)
 	if err != nil {
 		return nil, err
@@ -166,7 +159,6 @@ func (a *AnthropicHandlerMethods) shouldFlagRequest(ctx context.Context, logger 
 		if err := a.promptRecorder.Record(ctx, ar.BuildPrompt()); err != nil {
 			logger.Warn("failed to record flagged prompt", log.Error(err))
 		}
-		result.shouldBlock = result.shouldBlock && a.config.RequestBlockingEnabled
 	}
 	return result, nil
 }
--- a/cmd/cody-gateway/internal/httpapi/completions/flagging.go
+++ b/cmd/cody-gateway/internal/httpapi/completions/flagging.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	"github.com/sourcegraph/sourcegraph/cmd/cody-gateway/internal/tokenizer"
+	"github.com/sourcegraph/sourcegraph/cmd/cody-gateway/shared/config"
 	"github.com/sourcegraph/sourcegraph/internal/trace"
 	"github.com/sourcegraph/sourcegraph/lib/errors"
 )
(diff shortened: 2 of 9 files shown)


cmd/cody-gateway/internal/httpapi/completions/anthropic.go
```diff
@@ -166,7 +159,6 @@ func (a *AnthropicHandlerMethods) shouldFlagRequest(ctx context.Context, logger
 		if err := a.promptRecorder.Record(ctx, ar.BuildPrompt()); err != nil {
 			logger.Warn("failed to record flagged prompt", log.Error(err))
 		}
-		result.shouldBlock = result.shouldBlock && a.config.RequestBlockingEnabled
```
chrsmith, line 169: This was pushed down into `isFlaggedRequest`. Since the `flaggingConfig` object has an `RequestBlockingEnabled` field, we no longer need to have this check repeated in multiple places.

cmd/cody-gateway/internal/httpapi/completions/flagging.go
```diff
@@ -22,7 +23,25 @@ type flaggingConfig struct {
 	PromptTokenBlockingLimit       int
 	MaxTokensToSampleFlaggingLimit int
 	ResponseTokenBlockingLimit     int
+
+	// If false, the flaggingResult returned from isFlaggedRequest will never have shouldBlock set to true.
```
vdavid, line 27: Nit: I find this hard to read. How about
```suggestion
	// If false, flaggingResult.shouldBlock will always be false when returned by isFlaggedRequest.
```

cmd/cody-gateway/internal/httpapi/completions/openai.go
```diff
@@ -126,9 +126,15 @@ func (*OpenAIHandlerMethods) validateRequest(_ context.Context, _ log.Logger, fe
 	return nil
 }
 
-func (*OpenAIHandlerMethods) shouldFlagRequest(_ context.Context, _ log.Logger, _ openaiRequest) (*flaggingResult, error) {
-	// TODO[#61278]: Add missing request validation for all LLM providers in Cody Gateway.
-	return nil, nil
+func (oaih *OpenAIHandlerMethods) shouldFlagRequest(ctx context.Context, logger log.Logger, req openaiRequest) (*flaggingResult, error) {
```
vdavid, line 129: My editor keeps on complaining if the receiver names are different within a struct. It seems like a best practice to always use the same receiver name throughout the functions of a struct.

```suggestion
func (o *OpenAIHandlerMethods) shouldFlagRequest(ctx context.Context, logger log.Logger, req openaiRequest) (*flaggingResult, error) {
```

vdavid (approved):
Thanks for the six individual commits, it was a breeze to review it commit by commit!
Thanks for the thorough comments! <3
Left two nits.

cmd/cody-gateway/internal/httpapi/completions/flagging.go
chrsmith: Yup, definitely a clear improvement.

cmd/cody-gateway/internal/httpapi/completions/openai.go
chrsmith: Good catch. I didn't notice that we were using `o` elsewhere.

What IDE are you using? Apparently VS Code or the way I've configured it is failing me :'(

vdavid: I use IntelliJ IDEA. This is what my warnings look like:

CleanShot 2024-03-21 at 21 30 33@2x

PR #61336, opened 2024-03-22
minor changes

chwarwick:
We are always using the message api so I think we need to always remove the whitespace only stop sequence (made a minor edit there to remove all whitespace).


philipp-spiess (approved):
I think it's a good call. In the future, if there are issues with this approach, we can implement the stopSequence on our end too. It's confusing as to why this is not possible to set anymore in the new API

PR #61343, opened 2024-03-22
Feature flags: relax some constraints

camdencheek:
This fixes two small pain points with automated treatment of feature flags:
1) A user cannot add an override unless the feature flag has already been created and a default has been set. However, in all our code, we require the caller to specify a default if the feature flag doesn't exist, so this is not a necessary limitation. It's also particularly annoying because only site admins can create a feature flag.
2) In order to update an override, a user has to first fetch the override to see if it exists, then update it if it exists or create it if it doesn't. This modifies create to just overwrite if the override already exists.

I vaguely remember these things being complaints in the past as well, so hopefully this will make feature flags a bit more usable.

## Test plan

Existing tests covered this already. I updated them to reflect the new relaxed constraints. Also did a manual e2e test to ensure this will work for the things I want to use it for.


migrations/frontend/1711003437_remove_feature_flag_foreign_key/up.sql
```diff
@@ -0,0 +1,14 @@
...
+--  * Make migrations backwards-compatible (old readers/writers must continue to work)
+--  * If you are using CREATE INDEX CONCURRENTLY, then make sure that only one statement
+--    is defined per file, and that each such statement is NOT wrapped in a transaction.
+--    Each such migration must also declare "createIndexConcurrently: true" in their
+--    associated metadata.yaml file.
+--  * If you are modifying Postgres extensions, you must also declare "privileged: true"
+--    in the associated metadata.yaml file.
+
+ALTER TABLE feature_flag_overrides
+DROP CONSTRAINT feature_flag_overrides_flag_name_fkey;
```
camdencheek, line 14: First change: remove the foreign key constraint on the flag name. This allows users to add overrides for feature flags that do not have defaults set by the admin.

internal/featureflag/featureflag.go
```diff
@@ -78,6 +78,10 @@ type FeatureFlagRollout struct {
 	Rollout int32
 }
 
+func (f *FeatureFlagRollout) Evaluate(flagName string, userID int32) bool {
```
camdencheek, line 81: Just extracted this so I could use it separately from a full `FeatureFlag` object

internal/database/feature_flags.go
```diff
@@ -331,7 +287,18 @@ func (f *featureFlagStore) CreateOverride(ctx context.Context, override *ff.Over
...
+		-- NOTE: this only upserts for user overrides, not
+		-- org overrides. Postgres does not allow an ON CONFLICT
+		-- clause targeting two different unique constraints. Since
+		-- this just exists for convenience and an override can also
+		-- be explicitly updated, it should be okay.
+		ON CONFLICT (namespace_user_id, flag_name)
+		DO UPDATE SET
+			flag_value = EXCLUDED.flag_value,
+			updated_at = now(),
+			deleted_at = NULL
```
camdencheek, line 300: Second change: modify create to upsert instead. This allows users to just call the `createFeatureFlagOverride` GraphQL method if they want to update their override. Before, this would lead to a unique key violation, but now we just update the override if it already exists. A caller used to have to look up the feature flag override to get its id, then call update if it existed and create if it did not. Now it's safe to just always call create.

Due to postgres limitations, we cannot make this work for orgs as well. But since orgs are much more rarely used, I think that's okay. Once we upgrade to Postgres 15, we should be able to use the `MERGE` operator instead which appears to get around this limitation.

internal/database/feature_flags.go
```diff
@@ -183,50 +183,6 @@ func (f *featureFlagStore) CreateBool(ctx context.Context, name string, value bo
 
 var ErrInvalidColumnState = errors.New("encountered column that is unexpectedly null based on column type")
 
-func scanFeatureFlagAndOverride(scanner dbutil.Scanner) (*ff.FeatureFlag, *bool, error) {
```
camdencheek, line 186: I just inlined this because it's pretty specific to the function that calls it.

PR #61344, opened 2024-03-22
svelte: Update sg start enterprise-sveltekit command set

fkling:
This commit updates the enterprise-sveltekit command set to work with the current esbuild setup. It adds a new command to start the vite build process. The build process is quite slow so this is not a good way to develop the SvelteKit application but it's as close to the real setup as possible.

## Test plan

```
sg start enterprise-sveltekit
```

### DIFF
--- a/sg.config.yaml
+++ b/sg.config.yaml
@@ -417,6 +417,12 @@ commands:
       # Needed so that node can ping the caddy server
       NODE_TLS_REJECT_UNAUTHORIZED: 0
 
+  web-sveltekit:
+    description: Enterprise version of the web sveltekit app
+    cmd: pnpm --filter @sourcegraph/web-sveltekit dev:enterprise
+    install: |
+      pnpm install
+
   web-standalone-http:
     description: Standalone web frontend (dev) with API proxy to a configurable URL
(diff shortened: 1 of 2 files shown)


sg.config.yaml
```diff
@@ -417,6 +417,12 @@ commands:
       # Needed so that node can ping the caddy server
       NODE_TLS_REJECT_UNAUTHORIZED: 0
 
+  web-sveltekit:
+    description: Enterprise version of the web sveltekit app
+    cmd: pnpm --filter @sourcegraph/web-sveltekit dev:enterprise
```
camdencheek, line 422: TIL `--filter` can work that way

camdencheek (approved):
Nice! LGTM

sg.config.yaml
fkling: I stole that from other commands in this file 😉

vovakulikov: > TIL --filter can work that way

Just for my education, what is other way it can work? (I remember when we moved to pnpm from yarn it confused me a little bit that all sub packages should be selected through filter flag)

PR #61348, opened 2024-03-22
httpcli: add prometheus metric for monitoring the rate that Sourcegraph issues requests to external services

ggilmore:
This PR adds a Prometheus metric that tracks the rate that Sourcegraph issues HTTP request that target external hosts (e.g. github.com, gitlab.com, etc.)

For every HTTP request, the metric tracks the hostname, HTTP method, and response code. 

I have also added a new "External HTTP Request Rate" section the frontend Grafana dasboard that plots this metric. See the example screenshot below.

screencapture-sourcegraph-test-3443-debug-grafana-d-frontend-frontend-2024-03-22-13_18_13

## Test Plan  

- Some rote unit tests
- Generated the above screenshot on my local sourcegraph instance with `sg start monitoring`

### DIFF
--- a/monitoring/definitions/frontend.go
+++ b/monitoring/definitions/frontend.go
@@ -2,6 +2,7 @@ package definitions
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/sourcegraph/sourcegraph/monitoring/definitions/shared"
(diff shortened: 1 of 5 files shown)


monitoring/definitions/frontend.go
```diff
@@ -62,6 +64,16 @@ func Frontend() *monitoring.Dashboard {
 				},
 				Multi: true,
 			},
+			{
+				Label: "HTTP request hostname",
+				Name:  httpRequestHostVariableName,
+				OptionsLabelValues: monitoring.ContainerVariableOptionsLabelValues{
+					Query:         "src_http_client_external_request_count",
+					LabelName:     "host",
+					ExampleOption: "ping.sourcegraph.com",
```
eseliger, line 73: I think we're phasing out ping in favor of pings, so to avoid confusion maybe?

```suggestion
					ExampleOption: "pings.sourcegraph.com",
```

monitoring/definitions/frontend.go
```diff
@@ -62,6 +64,16 @@ func Frontend() *monitoring.Dashboard {
 				},
 				Multi: true,
 			},
+			{
```
eseliger, line 67: this lives on frontend, but it's not specific to frontend, right? I.e., we show all requests from all services here?

eseliger (approved):

PR #61354, opened 2024-03-24
svelte: Remove transitive dependency on all of @sourcegraph/wildcard (via eventLogger)

fkling:
One reason why the development server is relatively slow is that we are pulling in a lot of (transitive) dependencies that are not used. Specifically the whole wildcard package is included.

I used [dependency-cruiser](https://github.com/sverweij/dependency-cruiser) to find out why all of wildcard is included and eventually discovered that it was because of importing `eventLogger`.

`eventLogger` imports `observerQuerySelector` from `dom.ts` which in turn imports `useObservable` from `@sourcegraph/wildcard`, which of course imports all of wildcard:

wildcard-index-deps

Since `eventLogger` is deprecated and we'll have to implement proper support for event logging anyway I've removed the dependency for now. This avoids pulling in all of wildcard and reduces the number of request made for the homepage by almost half:

Before: 

After: 

**Additional notes**

In future PR we could add dependency-cruiser to help us keep our dependencies "clean". E.g. it can help us detect transitive dependencies to React. We can then refactor our code so that non-React modules don't depend on React modules.

I don't see why `useObservable` is part of `client/wildcard`. Maybe it should be moved to `client/shared`.

## Test plan

`pnpm dev` and `pnpm build`

### DIFF
--- a/client/web-sveltekit/src/lib/logger.ts
+++ b/client/web-sveltekit/src/lib/logger.ts
@@ -4,11 +4,9 @@
 import { onMount } from 'svelte'
 
 // Dev build breaks if this import is moved to `$lib/web` ¯\_(ツ)_/¯
-import { eventLogger, type EventLogger } from '@sourcegraph/web/src/tracking/eventLogger'
+import type { EventLogger } from '@sourcegraph/web/src/tracking/eventLogger'
 
 import { PUBLIC_ENABLE_EVENT_LOGGER } from '$env/static/public'
-
-export { eventLogger }
 
 /**
  * Can only be called during component initialization. It logs a view event when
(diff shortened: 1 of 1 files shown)


vovakulikov (approved):
Previously it wasn't problem since tree shaking managed index.ts exports from packages properly, it looks like esbuild doesn't have it (in our configuration at least). 

The `shared` package is a pile of everything (legacy logic, code host, ..etc), intention of having useObservable in the wildcard was that this is a "clean" (no deep deps file) piece of logic/helper that should live alongside with shared UI.

fkling:
Tree shaking works fine in production build, but the development server serves all source files as is. I mean, I guess I could configure some of the dependencies as external, but it's probably still good to restructure the dependencies a bit and don't just rely on build tools.

PR #61367, opened 2024-03-25
code monitors: don't fail job if repo empty

stefanhengl:
Previously we failed a code monitor search if any of the revisions could not be resolved. For example, this happens if an empty repo is added to Sourcegraph. Now we log the error and continue with the job. The log message is shown in the Code Monitors log view.

image

Notes:
- The cm_trigger_jobs table had a field `log_contents: TEXT` which was unused. However, I anticipate we might want to add more logs in the future so `TEXT` is not the ideal type. Instead, I add a new column `logs: JSON[]`. This way we can add richer metadata to the logs if we want to.
- I have to drop `log_contents` in a separate PR, otherwise the backcompat tests fail.

### Test plan:
- manual testing
  - I added an empty repo to reproduce the error locally
  - I created a search job and verified that it passes with 0 results and that the logs are properly displayed in the UI (see screenshot)
- new unit test


internal/database/code_monitor_trigger_jobs.go
```diff
@@ -90,6 +110,16 @@ func (s *codeMonitorStore) UpdateTriggerJobWithResults(ctx context.Context, trig
 	return s.Store.Exec(ctx, sqlf.Sprintf(logSearchFmtStr, queryString, resultsJSON, triggerJobID))
 }
 
+const updateTriggerJobLogsFmtStr = `
+UPDATE cm_trigger_jobs
+SET execution_logs = execution_logs || %s::json
```
eseliger, line 115: the execution_logs column is currently expected to contain `ExecutionLogEntry` structs, https://sourcegraph.com/-/editor?remote_url=github.com%2Fsourcegraph%2Fsourcegraph&branch=main&file=internal%2Fexecutor%2Fstore.go&editor=VSCode&version=2.2.16&start_row=10&start_col=0&end_row=19&end_col=0

This is currently executors specific, but it shouldn't be. To avoid confusion, it might be good to either
- Universally use `ExecutionLogEntry`
- Or to add a new column to the table (this is how we generally store additional execution information in other workers) **Edit:** I see we already had a column for that, could we use `log_contents` here?

internal/codemonitors/search.go
```diff
@@ -181,14 +183,27 @@ func hookWithID(
...
+			// and continue. This can happen for empty repos.
+			var revErr *gitdomain.RevisionNotFoundError
+			if errors.As(err, &revErr) {
+				if err := db.CodeMonitors().UpdateTriggerJobWithLogs(
+					ctx,
+					triggerID,
+					database.TriggerJobExecutionLogEntry{Message: err.Error()},
+				); err != nil {
+					return err
+				}
```
eseliger, line 203: would we need a `continue` statement here? Otherwise `commitHashes` gets an empty string appended, not sure if this is a problem?

internal/codemonitors/search.go
```diff
@@ -181,14 +183,27 @@ func hookWithID(
...
+		// Fail early for context cancellation.
 		if err := ctx.Err(); err != nil {
 			return ctx.Err()
 		}
 
 		res, err := gs.ResolveRevision(ctx, args.Repo, rev, gitserver.ResolveRevisionOptions{NoEnsureRevision: true})
 		if err != nil {
-			return err
+			// If the revision is not found, update the trigger job with the error message
+			// and continue. This can happen for empty repos.
```
eseliger, line 194: Should this only apply to HEAD? Or do we not care about explicitly mentioned revisions that are not found either?

internal/codemonitors/search.go
peterguy: or an `else`?:
```
if err != nil {
  ...
} else {
  commitHashes - append(...)
}
```

internal/codemonitors/search.go
```diff
@@ -181,14 +183,27 @@ func hookWithID(
...
+			var revErr *gitdomain.RevisionNotFoundError
+			if errors.As(err, &revErr) {
+				if err := db.CodeMonitors().UpdateTriggerJobWithLogs(
+					ctx,
+					triggerID,
+					database.TriggerJobExecutionLogEntry{Message: err.Error()},
+				); err != nil {
+					return err
```
peterguy, line 202: Returning the database error doesn't seem incredibly helpful here. Can we log the database error and return the original error? Or maybe there's another/better option?

internal/database/code_monitor_trigger_jobs.go
stefanhengl: > The cm_trigger_jobs table had a field log_contents: TEXT which was unused. However, I anticipate we might want to add more logs in the future so TEXT is not the ideal type. Instead, I decided to reuse execution_logs: JSON[], which was unused, too. This way we can add richer metadata to the logs if we want to.

I think []JSON is a better fit. I can add a new column.

internal/codemonitors/search.go
stefanhengl: IMO we shouldn't fail if a revision is not found, but it is probably more a product question.  

I am thinking of the use case where someones wants to monitor all "develop" branches in an org. It's a nice property that the monitor keeps working even repos exist that don't have the branch. WDYT?

internal/codemonitors/search.go
```diff
@@ -181,14 +183,27 @@ func hookWithID(
...
+				}
+			} else {
+				return err
+			}
 		}
```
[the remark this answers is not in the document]
stefanhengl: > Just ignore it on creation/update and let it be logged when the monitor runs?

That seems acceptable to me. @eseliger WDYT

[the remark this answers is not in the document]
eseliger: Yep, I think that makes sense. Even if a repo is empty now, I want to be informed about future changes, right?

internal/codemonitors/search.go
eseliger: Agreed, we can also silent skip over?

stefanhengl: > Agreed, we can also silent skip over?

@eseliger You mean silent skipping during creation, editing and execution? With the current state of the PR we silent skip during creation and editing, but log during execution.

eseliger: I meant during creation and editing, logging but not failing during execution seems fine to me

peterguy (approved):
All functionality works: creating, modifying, and running monitors that include empty repositories now works, with "revision not found" messages in the run logs.

A mild suggestion I have is to consider prepending "WARNING: " or something to the "revision not found" messages to avoid the appearance of "successfully failed" statuses.

Issue #61368, opened 2024-03-25
Cody Web: bring back the old deleted codebase of `code-shared` and `cody-ui`


vovakulikov:
At the beginning cody web (and it still depends on it) on `cody-ui` and `cody-shared` packages, later these packages were moved to the cody repo. But cody has evolved since then and now it doesn't have dependencies on these packages. 

The only reason why Cody-web is still working is because these packages live only npm as published code, in this issue we should bring these packages back to the main repo that we can change logic there freely 

Also part of this we should write a small RFC as a conclusion for the further work on cody-web

peterguy:
[Cody Web RFC](https://docs.google.com/document/d/1we2H0W7hszfnqY17N6l9biFZTGTa6dtSMvEpng0lHJA/edit#heading=h.trqab8y0kufp)

PR #61376, opened 2024-03-25
Cody web: Bring back old packages from git history

vovakulikov:
Closes: https://github.com/sourcegraph/sourcegraph/issues/61368

This PR brings back old `cody-ui` and `cody-shared` packages back into the main monorepo. Since these packages have evolved and eventually were removed from Cody repo I found as I think the [most older commit](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody@0b60dceffac6bfc06532c8213568b68a5cb41367) in Cody repo where versions of packages have the same value as we update them last time in the main repo. But during development I found that there were some changes in Cody repo which didn't update the versions and I had to patch some files again manually. 

Also it probably makes sense to wait on this PR and merge it after April release. I doubt we will have some urgent work about cody web in the next week so waiting worth it. 

## Test plan
-We basically should manually test all cody web functional just to make sure we don't break anything here.


package.json
```diff
@@ -16,7 +16,7 @@
     "format:check": "prettier '**/{*.{js?(on),ts?(x),graphql,md,scss},.*.js?(on)}' --config prettier.config.js --check --write=false",
     "format:ci": "prettier $( { git diff --diff-filter=d --name-only \"$(git merge-base $COMMIT_SHA $BUILDKITE_PULL_REQUEST_BASE_BRANCH)\" $COMMIT_SHA && git ls-files --other --modified --exclude-standard ; } | grep -E '\\.(js|json|ts|tsx|graphql|md|scss)$' | xargs) --config prettier.config.js --check --write=false",
     "_lint:js": "DOCSITE_LIST=\"$(./dev/docsite.sh -config doc/docsite.json ls)\" NODE_OPTIONS=\"--max_old_space_size=16192\" eslint",
-    "lint:js:changed": "pnpm _lint:js $(git diff --diff-filter=d --name-only origin/main... | grep -E '\\.[tj]sx?$' | xargs)",
+    "lint:js:changed": "pnpm _lint:js $(git diff --diff-filter=d --name-only origin/main... | grep -E '\\.[tj]sx?$' | xargs) --fix",
```
thenamankumar, line 19: This might have been left here by mistake, I guess you tried running it with `--fix` to fix the issues locally and then the change got committed. I suggest not add the `--fix` flag.

client/cody-shared/.gitignore
```diff
@@ -0,0 +1,2 @@
+node_modules/
```
thenamankumar, line 1: do we not want to ignore `dist/` and `out/` as well? Not sure if there are generated still but saw them in `.eslintignore` above.

client/cody-shared/src/chat/preamble.ts
```diff
@@ -0,0 +1,111 @@
+import type { Message } from '../sourcegraph-api'
+
+export interface Preamble {
+    actions: string
+    rules: string
+    answer: string
+}
+
+const actions = `You are Cody, an AI-powered coding assistant created by Sourcegraph. You work inside a text editor. You have access to my currently open files. You perform the following actions:
```
thenamankumar, line 9: I expect all of this to have been considerably changed by now. I do not expect this PR to address that. But it will be good to collect these high-impact changes required to get Cody Web on parity with IDE clients, somewhere. Probably in the RFC you wrote. Which I haven't reviewed yet.

client/cody-shared/src/chat/prompts/default-prompts.json
```diff
@@ -0,0 +1,32 @@
+{
+  "commands": {
```
thenamankumar, line 2: This looks like it was added after the v0.0.9 release. But good to keep the latest things here, making sure it works.

client/cody-shared/src/chat/recipes/agent-recipes.ts
```diff
@@ -0,0 +1,52 @@
...
+import type { Recipe, RecipeID } from './recipe'
+import { TranslateToLanguage } from './translate'
+
+function nullLog(filterLabel: string, text: string, ...args: unknown[]): void {
+    // Do nothing
+}
+
+export const registeredRecipes: { [id in RecipeID]?: Recipe } = {}
+
+export function getRecipe(id: RecipeID): Recipe | undefined {
```
thenamankumar, line 18: I did a grep and found this file not being used anywhere. We should remove this.

client/cody-shared/src/chat/recipes/code-question.ts
thenamankumar, line 1: This recipe is not used anywhere. We should not import this.

client/cody-shared/src/chat/recipes/agent-recipes.ts
thenamankumar, line 1: I did a grep and found this file not being used anywhere, only browser-recipes is used. We should remove this.

client/cody-shared/src/chat/recipes/context-search.ts
thenamankumar, line 1: This recipe is not used anywhere, we should not import this.

client/cody-shared/src/chat/recipes/fixup.ts
thenamankumar, line 1: This recipe is not used and should not be imported.

client/cody-shared/src/chat/recipes/generate-pr-description.ts
thenamankumar, line 1: This recipe is not used anywhere and we can skip importing it.

client/cody-shared/src/chat/recipes/generate-release-notes.ts
thenamankumar, line 1: This recipe is not used anywhere and we can skip importing it.

client/cody-shared/src/chat/recipes/git-log.ts
thenamankumar, line 1: This recipe is not used anywhere and we can skip importing it.

client/cody-shared/src/chat/recipes/next-questions.ts
thenamankumar, line 1: Next question is not implemented on the web. We can skip importing it.

client/cody-shared/src/chat/recipes/recipe.ts
```diff
@@ -0,0 +1,43 @@
...
+export interface RecipeContext {
+    editor: Editor
+    intentDetector: IntentDetector
+    codebaseContext: CodebaseContext
+    responseMultiplexer: BotResponseMultiplexer
+    firstInteraction: boolean
+}
+
+export type RecipeID =
+    | 'chat-question'
```
thenamankumar, line 17: A bunch of these recipies are not implemented on the web and technically can not be. We should not import those recipes. I have commented on the files for those recipes as well.

client/cody-shared/src/codebase-context/index.ts
```diff
@@ -0,0 +1,326 @@
...
+    numCodeResults: number
+    numTextResults: number
+}
+
+export class CodebaseContext {
+    private embeddingResultsError = ''
+
+    constructor(
+        private config: Pick<Configuration, 'useContext' | 'serverEndpoint' | 'experimentalLocalSymbols'>,
+        private codebase: string | undefined,
```
thenamankumar, line 38: none of these contexts other than `unifiedContextFetcher` works for web. So we can refactor this and remove all the other args and code related to them.

But to keep the changes in the PR minimal and only focus on directly importing the code to sourcegraph repo, we can work on this improvement separately later on. Provided we continue building on top of this.

client/cody-shared/src/prompt/constants.ts
```diff
@@ -0,0 +1,11 @@
+export const DEFAULT_MAX_TOKENS = 7000
```
thenamankumar, line 1: Can we make sure these are upto date with the current constants in the Cody repo?

thenamankumar:
Reviewed the PR and tested it locally. LGTM. There is a bunch of code which is IDE-specific but is imported as part of this PR, which is obvious given the fact that the imported code belonged to the shared lib package. It makes sense to do that refactoring separately and not as part of this PR.

But I have still highlighted the recipes which are not used on the web and won't be used in future as well, due to their IDE-specific use case. We should not import them as part of this PR. Removing the recipes should be fairly straightforward.

package.json
vovakulikov: Yes, just left it by mistake, thanks for catching

client/cody-shared/src/chat/preamble.ts
vovakulikov: Yeah, this is still under discussions, we don't know yet what this feature parity means (probably just a subset of features that we have in IDE clients but might be something different because of search/cody integration project everyone talked about it recently)

client/cody-shared/src/codebase-context/index.ts
vovakulikov: Yeah, if something requires more than just delete imports or file I would prefer to work on this in follow up since this PR is just about bringing logic back without any changes if possible. But thanks for the context, will mark this for further things in cody web.

client/cody-shared/src/prompt/constants.ts
vovakulikov: Checked this, still have 7k there

vovakulikov:
@thenamankumar thanks for the review and for your suggestions, good catch on unused/unsupported recipes. Removed all things by your comments 😉

peterguy (approved):
Looks well-reviewed; approving to unblock

Issue #61342, opened 2024-03-22
Add deprecated notices to the GraphQL completions resolver(s)


chwarwick:
Currently there are no official consumers of Cody via GraphQL.  It would be best to keep it that way.  Add a deprecated with a reason that consumers should use the non-graphql endpoint that

PR #61381, opened 2024-03-25
Adding warnings to the consumers of GQL endpoint for completions api

arafatkatze:
Solves https://github.com/sourcegraph/sourcegraph/issues/61342 

## Test plan
When you run a local instance and use the completions endpoint you will see a yellowed out sign at the bottom and a warning
image


RXminuS (approved):

slimsag:
It'd be nice to include a link to some docs about what/where the non-graphql endpoint is

otherwise lgtm

PR #61393, opened 2024-03-26
search: add ENV to disable hybrid search

stefanhengl:
This gives us flexibility during troubleshooting issues if we suspect hybrid search might be the issue.

Test plan:
CI

### DIFF
--- a/cmd/searcher/internal/search/search.go
+++ b/cmd/searcher/internal/search/search.go
@@ -53,6 +53,8 @@ type Service struct {
 	// single call to git archive. This mainly needs to be less than ARG_MAX
 	// for the exec.Command on gitserver.
 	MaxTotalPathsLength int
+
+	UseHybridSearch bool
 }
 
 func (s *Service) search(ctx context.Context, p *protocol.Request, sender matchSender) (err error) {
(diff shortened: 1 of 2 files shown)


cmd/searcher/internal/search/search.go
```diff
@@ -53,6 +53,8 @@ type Service struct {
 	// single call to git archive. This mainly needs to be less than ARG_MAX
 	// for the exec.Command on gitserver.
 	MaxTotalPathsLength int
+
+	DisableHybridSearch bool
```
mmanela, line 57: Put a comment on why you may want to do this

cmd/searcher/shared/shared.go
```diff
@@ -43,6 +43,8 @@ var (
 	backgroundTimeout = env.MustGetDuration("PROCESSING_TIMEOUT", 2*time.Hour, "maximum time to spend processing a repository")
 
 	maxTotalPathsLengthRaw = env.Get("MAX_TOTAL_PATHS_LENGTH", "100000", "maximum sum of lengths of all paths in a single call to git archive")
+
+	disableHybridSearch = env.MustGetBool("DISABLE_HYBRID_SEARCH", false, "set to true to disable hybrid search")
```
mmanela, line 47: Where do they set this variable? Is it on the site admin page somewherE?

keegancsmith: maybe include in documentation here that this will avoid unindexed search consulting indexed search for speeding up searches.

@mmanela this is set as an environment variable on the searcher service. In the case of a cloud instance, we will update the relevant resources for a customer to do this.

cmd/searcher/internal/search/search.go
keegancsmith: agreed, some information about it being a manual breaker to stop communication with zoekt.

keegancsmith (approved):

PR #61395, opened 2024-03-26
msp_delivery: infer pipeline id from service_id and region

jac:
Update `msp_delivery` bazel macro arguments to make it simpler to use
## Test plan

CI


bobheadxi (approved):
Nice 😁

PR #61396, opened 2024-03-26
Add Sonnet for free users

philipp-spiess:
## Test plan

- Trivial change (just adding a model to the allowlist) CI will test

### DIFF
--- a/internal/completions/httpapi/chat.go
+++ b/internal/completions/httpapi/chat.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/sourcegraph/log"
 
+	"github.com/sourcegraph/sourcegraph/internal/completions/client/anthropic"
 	"github.com/sourcegraph/sourcegraph/internal/completions/client/fireworks"
 	"github.com/sourcegraph/sourcegraph/internal/completions/types"
 	"github.com/sourcegraph/sourcegraph/internal/conf/conftypes"
@@ -75,9 +76,9 @@ func isAllowedCustomChatModel(model string, isProUser bool) bool {
 	if isProUser {
 		switch model {
 		case
-			"anthropic/claude-3-haiku-20240307",
-			"anthropic/claude-3-sonnet-20240229",
-			"anthropic/claude-3-opus-20240229",
+			"anthropic/" + anthropic.Claude3Haiku,
+			"anthropic/" + anthropic.Claude3Sonnet,
+			"anthropic/" + anthropic.Claude3Opus,
 			"fireworks/" + fireworks.Mixtral8x7bInstruct,
 			"openai/gpt-3.5-turbo",
 			"openai/gpt-4-1106-preview",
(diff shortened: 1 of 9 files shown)


internal/completions/httpapi/chat.go
```diff
@@ -74,7 +74,9 @@ func isAllowedCustomChatModel(model string, isProUser bool) bool {
 	// When updating these two lists, make sure you also update `allowedModels` in codygateway_dotcom_user.go.
 	if isProUser {
 		switch model {
-		case "anthropic/claude-3-sonnet-20240229",
+		case
+			"anthropic/claude-3-haiku-20240307",
```
philipp-spiess, line 78: This was an oversight. It currently works because we do not go to the Sourcegraph instance for Haiku in Chat. We need to roll this out before we land https://github.com/sourcegraph/cody/pull/3551 though

internal/completions/httpapi/chat.go
```diff
@@ -93,7 +95,10 @@ func isAllowedCustomChatModel(model string, isProUser bool) bool {
 		}
 	} else {
 		switch model {
-		case // Remove after the Claude 3 rollout is complete
+		case
+			"anthropic/claude-3-sonnet-20240229",
+			"anthropic/claude-3-haiku-20240307",
```
philipp-spiess, line 100: Adding haiku here as well since we want to use it for the `fast` model. This does not mean it will be part of the model dropdown.

chrsmith (approved):
This seems reasonable enough.

Don't block on this, but mostly questions for my own understanding and figuring out future-directions:

- Where do these strings come from? These are defined in the Anthropic APIs docs or something?
- Does it make sense to introduce constant values for these? Having the string is probably clearer, but since we are already using it in 2x places... it is on the verge of tripping over some linter rules, and/or being a code smell.
- (Long term, now today) Would we want these to be controlled server-side, right? e.g. the backing Sourcegraph instance may want to allow/block certain LLM models (e.g. for cost reasons, or due to licensing agreements, etc.?)

philipp-spiess:
> Where do these strings come from? These are defined in the Anthropic APIs docs or something?

Yep exactly, these are from the Anthropic docs: https://docs.anthropic.com/claude/docs/models-overview. We only prefix them with the respective provider name. 

> Does it make sense to introduce constant values for these? Having the string is probably clearer, but since we are already using it in 2x places... it is on the verge of tripping over some linter rules, and/or being a code smell.

Yeah you're right, we kinda do this already to some extend for Fireworks but for Claude it's a bit of a mess. Let me add it here since for the Claude 3 models there's the date suffix that makes it easy to introduce typos actually.

> (Long term, now today) Would we want these to be controlled server-side, right? e.g. the backing Sourcegraph instance may want to allow/block certain LLM models (e.g. for cost reasons, or due to licensing agreements, etc.?)

Agree but for now custom models are only supported on the dotcom instance (where we maintain this allowlist). For Enterprise, only the model that is configured in the site config can be used.

internal/completions/httpapi/chat.go
```diff
@@ -74,8 +75,10 @@ func isAllowedCustomChatModel(model string, isProUser bool) bool {
 	// When updating these two lists, make sure you also update `allowedModels` in codygateway_dotcom_user.go.
 	if isProUser {
 		switch model {
-		case "anthropic/claude-3-sonnet-20240229",
-			"anthropic/claude-3-opus-20240229",
+		case
```
thenamankumar, line 78: In which case are these old non-gateway endpoints used? Do enterprises use these for BYOK?

philipp-spiess: @thenamankumar for chat we use this in both PLG and Enterprise, regardless of Cody Gateway or BYOK. There was one "exception" that I added to ship Claude 3 faster which I am reverting in [this PR](https://github.com/sourcegraph/cody/pull/3551)

PR #61398, opened 2024-03-26
Fix links on /docs/admin/executors/deploy_executors_kubernetes

mike-r-mclaughlin:
Change links pointing directly to https://docs.sourcegraph.com to relative links

## Test plan

View in browser and make sure the links go to the new docs site.


loujar (changes requested):
@mike-r-mclaughlin just realized this is in the sourcegraph/sourcegraph repo. we should be committing doc changes in https://github.com/sourcegraph/docs going forward

mike-r-mclaughlin:
oops.  wrong place for docs ....

PR #61400, opened 2024-03-26
gomod: update zoekt for List performance improvements (#61397)

michaellzc:
https://github.com/sourcegraph/zoekt/compare/f2577c25dc61...931974dd953e

- 06de7adb72 shards: use selectRepoSet in List
- 931974dd95 shards: respect scheduler and use smarter synchronization for List

Test Plan: tested in zoekt repo and this CI

(cherry picked from commit b98f333ba5f16f7ad410ba6335758f3aa3f4a4b0)

backport https://github.com/sourcegraph/sourcegraph/pull/61397

## Test plan

CI

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -169,7 +169,7 @@ require (
 	github.com/shurcooL/httpgzip v0.0.0-20190720172056-320755c1c1b0
 	github.com/slack-go/slack v0.10.1
 	github.com/smacker/go-tree-sitter v0.0.0-20231219031718-233c2f923ac7
-	github.com/sourcegraph/go-ctags v0.0.0-20231024141911-299d0263dc95
+	github.com/sourcegraph/go-ctags v0.0.0-20240325032535-71f1ec6e46e4
 	github.com/sourcegraph/go-diff v0.6.2-0.20221123165719-f8cd299c40f3
 	github.com/sourcegraph/go-jsonschema v0.0.0-20221230021921-34aaf28fc4ac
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/sourcegraph/sourcegraph
 
-go 1.21
```
eseliger, line 3: this feels risky, was that required?

michaellzc: I hope not. let me revert this

mmanela: Ugh this is the rub I guess, we have changes backporting that may already take advantage of 1.22. (like the loop capture change), so not sure the impact of backporting and leaving old go version @eseliger

michaellzc: reverted, 🤞 CI passes

eseliger: sourcegraph/zoekt asks for go 1.22, maybe it actually is required :/

michaellzc: > sourcegraph/zoekt asks for go 1.22, maybe it actually is required :/

we can maintain a 1.21 branch of zoekt if needed. let's see

mmanela: but this PR is updating to a new build of zoekt

eseliger: > Ugh this is the rub I guess, we have changes backporting that may already take advantage of 1.22. (like the loop capture change)

Good point, that might actually be risky. But just randomly bumping the go version in an emergency patch also feels risky :/

keegancsmith:
I will create a new backport for this that is far less risky. ie doesn't change go versions and only includes our targetted changes instead of all changes to zoekt main since 5.3

PR #61401, opened 2024-03-26
Update JB sign-in instructions

taylorsperry:
Update the sign-in instructions for both Enterprise and Free/Pro users. @MaedahBatool , if you think it would be helpful to include a screenshot (I don't know if it's necessary?) I've added an example below: 
Screenshot 2024-03-26 at 10 35 22 AM

## Test plan

Passing CI


taylorsperry:
closing in favor of https://github.com/sourcegraph/docs/pull/198

PR #61404, opened 2024-03-26
Docs redirect

jamesmcnamara:
It always drives me crazy that the legacy docs redirect takes you to the main docs page and not the page you were looking at. So I added a small snippet to set the href to point to the relevant new docs page
## Test plan
Tested locally.


dev/sg/internal/run/installer.go
```diff
@@ -51,6 +51,9 @@ type InstallManager struct {
 }
 
 func Install(ctx context.Context, env map[string]string, verbose bool, cmds []Installer) error {
+	if len(cmds) == 0 {
+		return nil
+	}
```
jamesmcnamara, line 56: This just fixes a bug in `sg run` if you have a single task which doesn't have an installer.

unknwon (approved):

Issue #60741, opened 2024-02-23
Telemetry: Track origin of signup


eseliger:
Copied from [slack conversation](https://sourcegraph.slack.com/archives/C05EMJM2SLR/p1708629693587119):

Hey Source folks - could I put in a telemetry request? I'm looking at the ExternalAuthSignupSucceeded event and I was wondering if we could tag it with metadata for whether the signup was:
Triggered from the web (aka user signs up first, and then installs)
Triggered from the IDE (aka user installs first, and then signs up through the IDE flow)
I believe we can since we have logic that re-routes users back to the IDE after the Hubspot work/personal form during the onboarding flow if they trigger signup from the IDE. Can I get your help adding logging for this?
Why:
In our effort to move retention, we're breaking down the funnel in each step: signup, install, first DAU action in the product, and then DAU actions in subsequent days.
What makes this difficult to currently measure, however, is that without that metadata, we can't tell whether or not a user started with signup before or after install. So our metrics are always kind of muddy
(ie we get twisted up with things like: "the dashboard shows signup --> install rate is 10%, but that's because a lot of the signup events shouldn't even be counted because they happened after install, so we don't really know what it is")

eseliger:
cc @chenkc805

dadlerj:
Just curious—do users go _direct_ from the editor to the /.auth URL and into the auth flow? Or do they land on a dotcom page and then see an option to sign in from GitHub/GitLab/etc.?

eseliger:
Yes, directly to `/.auth` is my understanding, to trigger a flow where it's almost invisible that you connect to dotcom vs SAMS.

chenkc805:
Hey @eseliger , bumping this, as this is pretty important for understanding how to improve retention. How can I help get this prioritized with the Source team? Context here: https://sourcegraph.slack.com/archives/C06K09TB8UR/p1711395129411779

eseliger:
Checking

chenkc805:
Thank you for re-opening this @eseliger . We've been intending to investigate this on the Cody Prime team but have 0 bandwidth to do so

eseliger:
Pardon, this was some hiccup of Linear getting out of date and me touching the tickets so it triggered a reconciliation.

I will move this in linear to the cody prime team

PR #61406, opened 2024-03-26
Add hint to telemetry on where signup came from

eseliger:
This has been requested for PLG here: https://github.com/sourcegraph/sourcegraph/issues/60741

To better track the order of events (signup first or install IDE extension first), we're adding an additional field to the event logged when a new user is created that indicates that a signup was initiated by the IDE.

Closes https://github.com/sourcegraph/sourcegraph/issues/60741

## Test plan

Code review, asked PLG to help review.


cmd/frontend/auth/user.go
```diff
@@ -185,13 +189,18 @@ func GetAndSaveUser(ctx context.Context, db database.DB, op GetAndSaveUserOp) (n
...
-		})
+		argMap := map[string]any{}
+		for k, v := range op.UserCreateEventProperties {
+			argMap[k] = v
+		}
+		// NOTE: The conventional name should be "service_type", but keeping as-is for
+		// backwards capability.
+		argMap["serviceType"] = acct.AccountSpec.ServiceType
+
+		args, err := json.Marshal(argMap)
```
bobheadxi, line 203: FYI see docstring above:

```
		// Legacy event - new event is recorded outside this closure to cover
		// more scenarios. We retain the legacy event because it is still
		// exported by the legacy Cloud exporter, remove in future release.
```

the v2 event recording is somewhere in here, we should update that callsite too as the data team is working towards v2 telemetry

eseliger: I didn't touch what method is being called here - merely added some additional field. I don't know much about telemetry V2 or the different formats, so I'm a little scared to break the data for the PLG team 🙈 if you could handhold me here, ideally in a followup or stacked PR, that'd be appreciated 😬

cmd/frontend/internal/auth/openidconnect/middleware.go
```diff
@@ -359,7 +361,20 @@ func AuthCallback(db database.DB, r *http.Request, usernamePrefix string, getPro
 		return c.Value
 	}
 	anonymousId, _ := cookie.AnonymousUID(r)
-	newUserCreated, actor, safeErrMsg, err := getOrCreateUser(r.Context(), db, p, oauth2Token, idToken, userInfo, &claims, usernamePrefix, &hubspot.ContactProperties{
+
+	// PLG: If the user has been created just now, we want to log an additional event
```
dadlerj, line 365:
```suggestion
	// PLG: If the user has been created just now, we want to log an additional property on the event
```

cmd/frontend/auth/user.go
bobheadxi: The V2 event is recorded separately, so you just need to add the same attribute over there and it won't be breaking, but can also do that follow up tomorrow :)

dadlerj: +1 to this if possible. 

Typically v2 events can only handle numeric event properties, so to avoid complex property changes, I would recommend changing this to be a `map[string]int`. And instead of `"IDE"`, we'll use the number 1 to mean IDE, for example, and the number 2 to mean web :shrug:. This current version is fine for now though.

See here: https://github.com/sourcegraph/sourcegraph/blob/4f0b4a272ed61127b72e0972743afd829c6fed08/cmd/frontend/auth/user.go#L309-L319

dadlerj (commented):
LGTM. This will work for dotcom, since all of our signups use OIDC auth today with SAMS. 

That wasn't the case before SAMS, though, and we would have had to replicate this code in a bunch of places... I'd want to be sure that we plan to only use SAMS on dotcom and that it will only ever be OIDC?

cmd/frontend/internal/auth/openidconnect/middleware.go
```diff
@@ -359,7 +361,20 @@ func AuthCallback(db database.DB, r *http.Request, usernamePrefix string, getPro
...
 	}
 	anonymousId, _ := cookie.AnonymousUID(r)
-	newUserCreated, actor, safeErrMsg, err := getOrCreateUser(r.Context(), db, p, oauth2Token, idToken, userInfo, &claims, usernamePrefix, &hubspot.ContactProperties{
+
+	// PLG: If the user has been created just now, we want to log an additional property on the event
+	// that indicates that the user has initiated the signup from the IDE extension.
+	userCreateEventProperties := map[string]any{}
+	if dotcom.SourcegraphDotComMode() {
+		u, err := url.Parse(state.Redirect)
+		if err == nil {
```
chrsmith, line 370: It always seems odd to see us not handle an error case.

Would it make sense to write this as:

```
if err != nil {
    logger.Warn("dotcom signup, but unable to parse redirect URL. Not recording telemetry.", log.Error(err))
} else {

File: cmd/frontend/internal/auth/openidconnect/middleware.go [hunk 4]
```diff
@@ -359,7 +361,20 @@ func AuthCallback(db database.DB, r *http.Request, usernamePrefix string, getPro
 		return c.Value
 	}
 	anonymousId, _ := cookie.AnonymousUID(r)
-	newUserCreated, actor, safeErrMsg, err := getOrCreateUser(r.Context(), db, p, oauth2Token, idToken, userInfo, &claims, usernamePrefix, &hubspot.ContactProperties{
```
chrsmith, line 362: I know this is a little beyond the scope of this PR, but as a code cleanup would you mind adding a local variable to the top of the function `ctx := r.Context()` and then replacing all calls to `r.Context()` with a reference to `ctx`? It's debatable, but perhaps a clearer way to do the same thing. (As well as avoid bugs related to later if we needed to alias the context by adding a timeout or deadline.)

cmd/frontend/internal/auth/openidconnect/user.go
```diff
@@ -27,7 +27,7 @@ type ExternalAccountData struct {
 // getOrCreateUser gets or creates a user account based on the OpenID Connect token. It returns the
 // authenticated actor if successful; otherwise it returns a friendly error message (safeErrMsg)
 // that is safe to display to users, and a non-nil err with lower-level error details.
-func getOrCreateUser(ctx context.Context, db database.DB, p *Provider, token *oauth2.Token, idToken *oidc.IDToken, userInfo *oidc.UserInfo, claims *userClaims, usernamePrefix string, hubSpotProps *hubspot.ContactProperties) (newUserCreated bool, _ *actor.Actor, safeErrMsg string, err error) {
+func getOrCreateUser(ctx context.Context, db database.DB, p *Provider, token *oauth2.Token, idToken *oidc.IDToken, userInfo *oidc.UserInfo, claims *userClaims, usernamePrefix string, userCreateEventProperties map[string]any, hubSpotProps *hubspot.ContactProperties) (newUserCreated bool, _ *actor.Actor, safeErrMsg string, err error) {
```
chrsmith, line 30: ... It might be time to bite the bullet and require a custom type to bundle up all of these parameters.

- `ctx context.Context`
- `db database.DB`
- `p *Provider`
- `token *oauth2.Token`
- `idToken *oidc.IDToken`
- `userInfo *oidc.UserInfo`
- `claims *userClaims`
- `usernamePrefix string`
- `userCreateEventProperties map[string]any`
- `hubSpotProps *hubspot.ContactProperties`

e.g.

```go
// getOrCreateUserParams bundles up all of the options for completing the
// login-or-new-user process.
type getOrCreateUserParams struct {
     ...
     List the fields, whether or not they are optional or required. etc.
     ...
}

// Validate checks that all required parameters are present, and that
// the supplied data is well-formed.
func (o getOrCreateUserOptions) Validate() error {
    ...
}
```

cmd/frontend/internal/auth/openidconnect/user_test.go
```diff
@@ -77,6 +77,7 @@ func TestAllowSignup(t *testing.T) {
 				},
 				&userClaims{},
 				test.usernamePrefix,
+				nil,
```
chrsmith, line 80: When dealing with 10+ params, it's super-useful to call out what this stray `nil` is referring to.

```suggestion
				nil,  // UserCreateEventProperties
```

cmd/frontend/internal/auth/openidconnect/middleware.go
```diff
@@ -359,7 +361,20 @@ func AuthCallback(db database.DB, r *http.Request, usernamePrefix string, getPro
...
+
+	// PLG: If the user has been created just now, we want to log an additional property on the event
+	// that indicates that the user has initiated the signup from the IDE extension.
+	userCreateEventProperties := map[string]any{}
+	if dotcom.SourcegraphDotComMode() {
+		u, err := url.Parse(state.Redirect)
+		if err == nil {
+			if strings.EqualFold(u.Query().Get("requestFrom"), "CODY") {
```
chrsmith, line 371: Can we leave a comment to explain where this `requestFrom` URL query parameter is coming from? e.g. if there is a random change within the bowels of the Cody extension, this would suddenly break.

So having some reference to the other repo might at least provide some context for how to keep these things in-sync.

Alternatively, can we just always pass through the `requestFrom` if present? e.g.

```
userCreateEventProperties["signup_source"] = fmt.Sprintf("plg-user:requestFrom:%v", u.Query().Get("requestFrom"))
```

e.g. must the value _exactly_ be IDE? Why not just leave it as "CODY" (or whatever values we change it to in the future)?

chrsmith (approved):
As far as I understand, the function change here is to just wire through the `map[string]any` to record the telemetry. That seems reasonable enough.

But there are a few ways I'd gently suggest we try to leave this code in a better state than we found it.

eseliger:
> That wasn't the case before SAMS, though, and we would have had to replicate this code in a bunch of places... I'd want to be sure that we plan to only use SAMS on dotcom and that it will only ever be OIDC? 

There's some other PLG specific logic in OIDC already to pass along information about which auth provider to pick on the SAMS side, so I think this is fine and a much simpler change than touching all auth providers -> less risk.

cmd/frontend/internal/auth/openidconnect/user.go
eseliger: Moving it into a bag struct makes it a runtime error for missing required parameters, while forgetting a non-optional parameter is a compile time error.

I totally agree that this is fairly unreadable, but I also feel relatively comfortable knowing that at least that part I didn't mess up so that it'll bubble up as a runtime error

cmd/frontend/internal/auth/openidconnect/middleware.go
eseliger: Added a comment and switched to a boolean instead, wdyt?

cmd/frontend/internal/auth/openidconnect/middleware.go
eseliger: done

cmd/frontend/internal/auth/openidconnect/middleware.go
eseliger: done, thx for reminder :)

cmd/frontend/internal/auth/openidconnect/middleware.go
chrsmith: 🤷 :shipit: .

Personally, IMHO, I'd just keep it as a string and pass along the `"requestFrom"` URL query parameter unmodified. That way we aren't looking for a hard-coded string, which seems a little too fragile. e.g. if the Cody clients start sending some new value, we'll still be passing the `signup_source` key, it will just have a new value. Rather than not having the `signup_source` key at all.

However, I don't have much insight into the backend where we read/use this data. So it could be that `signup_source_is_cody` is preferable, since perhaps in the future we'd have other codepaths set that flag as well.

eseliger: From another thread on here, it sounds like future telemetry will only allow numerical values for privacy reasons, that's why I opted for `bool` now. But I don't have a strong opinion here, or really any strong knowledge on telemetry at all so happy to have a conversation :)

cmd/frontend/auth/user.go
bobheadxi: @eseliger see https://github.com/sourcegraph/sourcegraph/pull/61460 :) Feel free to just merge into this PR

> And instead of "IDE", we'll use the number 1 to mean IDE, for example, and the number 2 to mean web 🤷. This current version is fine for now though.

@dadlerj I think the source i.e. `"IDE"`, etc isn't tracked in either the V1 or the V2 telemetry of this event at all today, though we have a numeric identifier for `serviceVariant`, which is this enum: https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/internal/extsvc/types.go?L144

chenkc805 (approved):

PR #61419, opened 2024-03-27
chore: remove unused darwin-arm64 cross binaries for codeintel-qa

Strum355:
No longer needed since https://github.com/sourcegraph/sourcegraph/pull/60569

## Test plan

ctrl+f + bazel query to find any references


jhchabran (approved):
Thanks for thinking about this!

PR #61420, opened 2024-03-27
Use simple projects endpoint for GItLab permissions syncing

### LINKED ISSUE #60742: Gitlab /projects usage improvements
#### Question description

From an EM at Gitlab:

> I'm wondering if you could connect me with an engineer or EM who is familiar with requests to the GitLab GET /projects API endpoint made from the user agent Sourcegraph-Bot. I'm assuming these are calls from Sourcegraph to index a GitLab user's repos. We noticed a lot of the requests are quite slow and may be returning more data than needed, and I was hoping I could talk to someone at Sourcegraph about potentially making more efficient use of the endpoint.

#### Additional context

Internal Slack thread:
https://sourcegraph.slack.com/archives/C05EMJM2SLR/p1707513264604149?thread_ts=1707512398.176149&cid=C05EMJM2SLR

cc @ggilmore

pjlast:
Closes #60742

An EM at GitLab (in the issue above) suggested this as an optimisation to how we use the `/projects` endpoint for GitLab, since we hit it a lot. It simply adds the `simple=true` flag to the query parameters when doing permissions syncing requests. When set, GitLab returns a reduced subset of fields from the `/projects` endpoint. The reduced set of fields are sufficient for permissions syncing purposes, and very crude tests on my end showed that it does have an impact on the total request time, so it feels like a change with only upsides.

## Test plan

Tested locally.


internal/authz/providers/gitlab/sudo.go
```diff
@@ -230,6 +230,7 @@ func listProjects(ctx context.Context, client *gitlab.Client, listInternalRepos
 
 	q := make(url.Values)
 	q.Add("per_page", "100") // 100 is the maximum page size
+	q.Add("simple", "true")
```
eseliger, line 233: doc string please :)

mmanela (approved):

PR #61424, opened 2024-03-27
GraphQL: add ability to query by user ID for debugging

camdencheek:
When debugging, I fairly regularly need to look up a user by an ID from a log message or a trace or something. In order to do that, I currently resort to a hack like writing a test like the following:

```go
func TestMarshalUserID(t *testing.T) {
	t.Fatal(gql.MarshalUserID(38428))
}
```

then using the generated ID to query with GraphQL:

```graphql
query {
    node(id:"VXHalsdgh=") {
        ...on User { ... }
    }
}
```

This is not an obvious solution for what should be a simple operation, especially because when we're trying to look up a user by ID, we're frequently in a hurry, which then makes the most obvious solution just hitting the database directly.

This PR just adds `intID` as a queryable field on the `user()` GraphQL query. It's documented as for debugging to hopefully discourage people from preferring this to GraphQL IDs in clients.

## Test plan

Added a test. Manually tested. 

CleanShot 2024-03-27 at 10 11 17@2x

### DIFF
--- a/cmd/frontend/graphqlbackend/schema.graphql
+++ b/cmd/frontend/graphqlbackend/schema.graphql
@@ -1717,6 +1717,10 @@ type Query {
         Query the user by verified email address.
         """
         email: String
+        """
+        Query the user by ID. Clients should prefer using GraphQL IDs. This is primarily intended for debugging.
+        """
+        intID: Int
     ): User
     """
     List all users.
--- a/cmd/frontend/graphqlbackend/user.go
+++ b/cmd/frontend/graphqlbackend/user.go
@@ -36,6 +36,7 @@ func (r *schemaResolver) User(
 	args struct {
 		Username *string
 		Email    *string
+		IntID    *int32
 	},
 ) (*UserResolver, error) {
 	var err error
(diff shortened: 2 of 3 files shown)


evict (approved):

cmd/frontend/graphqlbackend/schema.graphql
```diff
@@ -1717,6 +1717,10 @@ type Query {
         Query the user by verified email address.
         """
         email: String
+        """
+        Query the user by ID. Clients should prefer using GraphQL IDs. This is primarily intended for debugging.
+        """
+        intID: Int
```
bobheadxi, line 1723: minor nit would be to call this databaseID, as I think the user resolvers return that, but not a big deal :)

camdencheek: Oh, good catch! I didn't realize we actually expose that. Totally agree. Will make a followup

camdencheek: https://github.com/sourcegraph/sourcegraph/pull/61427

PR #61425, opened 2024-03-27
embeddings: don't schedule jobs if disabled

stefanhengl:
We return early, either if embeddings are disabled or if there are no repos matching any policy.

Before, we scheduled embeddings for all repos returned by `ListMinimalRepos(ctx, database.ReposListOptions{IDs: nil})` in `ScheduleRepositoriesForPolicy`. Reading the documentation of `RepoListOptions.IDs`  "When zero-valued, this is omitted from the predicate set", this might cause repos to get embedded even if embeddings are disabled and no policies are defined. This causes 1 call to gitserver per repo.

Test plan:
CI


eseliger (approved):
Oh boi, you don’t wanna know how often I ran into this issue before with empty lists. 

Great find!

PR #61426, opened 2024-03-27
Blob View [svelte]: Add last commit data to the blobview

jasonhawkharris:
This PR adds the data from the last commit to the blob view in the bottom panel. Figma design can be found [here](https://www.figma.com/file/EPgl0qVbqGSumvf52brhda/Code-Search?type=design&node-id=1947-37160&mode=design&t=VCPhpphrJ6BmbjPV-0). As you can see, I have not added ownership information yet. I plan to do so in a follow-up PR. I think it's also worth evaluating whether we even need that information present, considering the state of Own.

Screenshot 2024-03-27 at 2 19 59 PM
Screenshot 2024-03-27 at 2 20 13 PM

Sidenote: does anyone know how to remove the svelte button from the bottom right of the screen? I couldn't figure it out, but it would be helpful to move it in this case.

## Test plan
1. manual testing of UI
3. helper functions ship with unit tests


camdencheek:
> does anyone know how to remove the svelte button from the bottom right of the screen? I couldn't figure it out, but it would be helpful to move it in this case.

When I'm doing final testing/screenshots and stuff, I usually do `pnpm build && pnpm preview` which builds a production version of the app. It doesn't have the live reloading, but it's more optimized and doesn't have the svelte picker. I'm not sure if there's an easy way to disable it in dev mode 🤔

client/web-sveltekit/src/lib/TabPanel.svelte
```diff
@@ -5,6 +5,7 @@
     import { type TabsContext, KEY } from './Tabs.svelte'
 
     export let title: string
+    export let iconPath: string | null = null
```
camdencheek, line 8: Is this used?

client/web-sveltekit/src/lib/repo/LastCommit.svelte
```diff
@@ -0,0 +1,102 @@
+<script lang="ts">
+    import { formatDistanceToNow } from 'date-fns'
+
+    import Avatar from '$lib/Avatar.svelte'
+    import type { Avatar_User } from '$testing/graphql-type-mocks'
```
camdencheek, line 5: I think we should import this from `Avatar.gql` rather than the type mocks package (which is generated from that file)

client/web-sveltekit/src/lib/repo/LastCommit.svelte
```diff
@@ -0,0 +1,102 @@
...
+    let avatar: Avatar_User = {
+        __typename: 'User',
+        avatarURL: avatarURL,
+        displayName: displayName,
+        username: displayName,
+    }
+
+    function getFirstNameAndLastInitial(name: string): string {
+        const names = name.split(' ')
+        return `${names[0]} ${names[names.length - 1].charAt(0).toUpperCase()}.`
```
camdencheek, line 22: What if a name only has one word? e.g. [seebs](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@4a87b94e63fe9be68457b37cdc7c56b941406f55/-/blob/.mailmap?L55-55)

client/web-sveltekit/src/lib/repo/LastCommit.svelte
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    function extractCommitMessage(commitMessage: string): string {
+        let split = commitMessage.split(' ')
+        let msg = split.slice(0, split.length - 1)
+        return msg.join(' ')
```
camdencheek, line 28: This appears to just be removing the last word of the commit message? Is that right?

client/web-sveltekit/src/lib/repo/LastCommit.svelte
```diff
@@ -0,0 +1,102 @@
...
+        cm = extractCommitMessage(commitMessage)
+        return cm.length > 30 ? cm.substring(0, 30) + '...' : cm
+    }
+
+    $: commitMessageNoSHA = truncateIfNeeded(commitMessage)
+    $: commitSHA = extractSHA(commitMessage)
+</script>
+
+<div class="latest-commit">
+    <div class="commit-info">
```
camdencheek, line 52: Why do we have two divs here? I expect we could flatten these

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte
```diff
@@ -147,6 +149,17 @@
...
                 </TabPanel>
             </Tabs>
+            {#if latestCommit}
+                <LastCommit
+                    commitURL={latestCommit.canonicalURL}
+                    avatarURL={latestCommit.author.person.avatarURL}
+                    displayName={latestCommit.author.person.displayName}
+                    commitMessage={latestCommit.subject}
+                    commitDate={latestCommit.author.date}
+                />
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte:159]: Given that we're just passing in a commit, should we just make `LatestCommit` take whatever that type is rather than having to unwrap it here?

client/web-sveltekit/src/lib/repo/LastCommit.svelte
```diff
@@ -0,0 +1,102 @@
...
+            <Avatar {avatar} />
+            <div class="display-name">
+                <small>{getFirstNameAndLastInitial(displayName)}</small>
+            </div>
+        </div>
+
+        <div class="commit-message">
+            <small>
+                {commitMessageNoSHA}
+                (<a href={commitURL}>{commitSHA}</a>)
```
camdencheek, line 63: Wondering whether the whole commit message should be clickable 🤔

client/web-sveltekit/src/lib/repo/LastCommit.svelte
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    function convertToElapsedTime(commitDateString: string): string {
+        const commitDate = new Date(commitDateString)
+        return formatDistanceToNow(commitDate, { addSuffix: true })
+    }
+
+    function truncateIfNeeded(cm: string): string {
+        cm = extractCommitMessage(commitMessage)
+        return cm.length > 30 ? cm.substring(0, 30) + '...' : cm
```
camdencheek, line 44: Rather than explicitly truncating, should we just set a max-width on the div and set [text-overflow: ellipsis](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow)?

client/web-sveltekit/src/lib/repo/LastCommit.svelte
jasonhawkharris: I like how Taiyab has designed it as it is, but I could see the benefit.

camdencheek: Ah, I think that's my confusion here. I'm pretty sure what you're calling the "commit SHA" here is actually the PR number, which is part of the commit message that we detect and turn into a link to the github PR

jasonhawkharris: ahhh... okay. I'll change the name to reflect that.

camdencheek: Well, it's not just changing the name. I think it's also pointing to the wrong URL. I'd take a look at how we recognize PR numbers in the current webapp and see if we can copy that. I think it's also acceptable to split the "link to external PR" part into a separate PR if you want to keep this one more tightly scoped

client/web-sveltekit/src/lib/repo/LastCommit.svelte
jasonhawkharris: Yup!

camdencheek: Why are we removing the last word of the commit message?

client/web-sveltekit/src/lib/repo/LastCommit.svelte
```diff
@@ -0,0 +1,95 @@
...
+    export let displayName: string
+    export let commitMessage: string
+    export let commitDate: string
+
+    let avatar: Avatar_User = {
+        __typename: 'User',
+        avatarURL: avatarURL,
+        displayName: displayName,
+        username: displayName,
+    }
```
fkling, line 18: You should add a `LastCommit.gql` file that defines a fragment for the data this component needs. Have a look at [`HistoryPanel.gql`](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@d882ad23ba0abe115db05bd7d788311df7828af0/-/blob/client/web-sveltekit/src/lib/repo/HistoryPanel.gql) as an example. That file also shows how to compose the data dependency for `Avatar`.

Then the appropriate GraphQL query needs to be updated to include that fragment. As you already discovered we are already fetching the commit history, so it could just be merged into `GitHistory_HistoryConnection`. I think there is also a field on `Repository` that returns the last commit but I'm not sure.

client/web-sveltekit/src/lib/repo/LastCommit.svelte
jasonhawkharris: Since the last word of the commit message is the PR number, we need to extract that and place it manually so that we can link just that string, and not the entire string.

client/web-sveltekit/src/lib/repo/LastCommit.svelte
jasonhawkharris: ahhhh.... okay. That makes sense. Thank you!

client/web-sveltekit/src/lib/repo/LastCommit.svelte
camdencheek: Yes, I think that's the right idea. But the last word of the commit message is _not always_ a PR number. That's just a convention that GitHub uses by default. But you can specify your own commit message. Or not use GitHub. We have a bunch of commits in sourcegraph/sourcegraph that do not have a PR number in them. 

I'd recommend looking at [how we do it in the React app](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@4a87b94e63fe9be68457b37cdc7c56b941406f55/-/blob/client/web/src/repo/commit/CommitMessageWithLinks.tsx?L13-28)

client/web-sveltekit/src/lib/repo/LastCommit.svelte
```diff
@@ -0,0 +1,98 @@
...
+        let split = commitMessage.split(' ')
+        let msg = split.slice(0, split.length - 1)
+        return msg.join(' ')
+    }
+
+    function extractPRNumber(cm: string): string {
+        let cmWords = cm.split(' ')
+        let sha = cmWords[cmWords.length - 1]
+        return sha.slice(1, sha.length - 1)
+    }
```
fkling, line 38: This seems too GitHub specific. We cannot assume that the last "word" is a PR number nor that it's wrapped in `(...)`.
If we want to do this for GitHub we should to some proper parsing (e.g. find a substring that matches `\(#\d+\)`) and we should only do this for commits/repos that are connected to GitHub.

As an inspiration, here is a function we use to link PR numbers to their GitHub URL in text: https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@d882ad23ba0abe115db05bd7d788311df7828af0/-/blob/client/web/src/repo/linkifiy/get-links.ts?L94

client/web-sveltekit/src/lib/repo/LastCommit.svelte
jasonhawkharris: It looks like `GitHistory_HistoryConnection` already has all the data I need for this, so I don't think it's necessary to add a `LastCommit.gql`. Do you agree?

fkling: The idea is that every component defines its own data dependencies so they can change independently.
It's not a problem if two different components define overlapping data dependencies. If they are used in the same GraphQL query the field will be deduplicated.

Imagine `LastCommit.svelte` needs additional data at some point. It's much easier to add the field in its corresponding graphql file than to track down the query that needs to be changed.
Similarly, maybe the component need less data at some point. Then the field can be safely removed without affecting any other components which might get data from the same (composed) query. And if no component requires a specific field anymore and it's removed from all fragments it won't be requested from the server anymore.

Overall this should Maske GraphQL queries much more manageable and scalable.

client/web-sveltekit/src/lib/repo/LastCommit.svelte
jasonhawkharris: Decided against this for the time being. The `text-overflow` idea actually seems less intuitive to me than using a separate function.

client/web-sveltekit/src/lib/repo/utils.test.ts
```diff
@@ -0,0 +1,127 @@
...
+        {
+            title: 'capitalizes correctly',
+            input: 'John smith',
+            expected: 'John S.',
+        },
+        {
+            title: 'capitalizes correctly',
+            input: 'jOhN sMiTH',
+            expected: 'John S.',
+        },
```
camdencheek, line 40: Question (maybe for @taiyab): should we be messing with the capitalization of someone's name (or for that matter, abbreviations of that name)? People set what they want their display name to be in git. Names like [Nyck de Vries](https://en.wikipedia.org/wiki/Nyck_de_Vries) are pretty common in some places, and the non-capitalized first character is part of the name.

A sample of edge cases just from the `sourcegraph/sourcegraph` repo:

```
abschu
Andre Marcelo-Tanner
Bunny
chwarwick
dan-mckean
Dax McDonald
dependabot[bot]
ErikaRS
garo (they/them)
IDOKO EMMANUEL
Iskander (Alex)
Jackson R
Jean du
Jean-Hadrien Chabran
JeanPaulSG
L. R.
Leo Succar-Ferré
LockedOutofHeaven
Loïc Guychard
MrStacks
n8maynard
oacnhpkqxjhufwumkkqnshvlmheokqv
Petri-Johan Last
xin.xing
ᴊ. ᴄʜᴇɴ
无闻
하광일 /
```

client/web-sveltekit/src/lib/repo/LastCommit.svelte
```diff
@@ -0,0 +1,87 @@
...
+    import {
+        truncateIfNeeded,
+        extractPRNumber,
+        getFirstNameAndLastInitial,
+        convertToElapsedTime,
+    } from '$lib/repo/utils'
+
+    export let latestCommit: LastCommitProps
+
+    interface LastCommitProps {
```
camdencheek, line 13: I'd recommend exporting this interface in a `<script context="module">` so consumers of this component can name the type they're creating.

client/web-sveltekit/src/lib/repo/LastCommit.svelte
```diff
@@ -0,0 +1,87 @@
...
+        </div>
+    </div>
+
+    <div class="commit-message">
+        {#if PRNumber}
+            <small>{commitMessageNoPR}</small>
+            (<a href={latestCommit.canonicalURL}><small>{PRNumber}</small></a>)
+        {:else}
+            <a href={latestCommit.canonicalURL}><small>{commitMessageNoPR}</small></a>
+        {/if}
```
camdencheek, line 53: I don't think we should deconstruct and reconstruct the message like this. There's no guarantee that a PR number is at the end of a commit. That's just the way GitHub generates the default message. As an example, revert PRs [quote the PR title that was reverted](https://github.com/sourcegraph/sourcegraph/pull/60498). 

Instead, I would recommend using the regex you wrote for PR numbers and replacing all instances with links. Then we would render the "linkified" HTML here. 

Second, I think the link is incorrect. A PR number should not link to the commit, it should link to the PR on GitHub.

client/web-sveltekit/src/lib/repo/LastCommit.svelte
fkling: I'd strongly advice for using text-ellipsis. It avoids two problems:

- unnecessarily truncated commit message on wide screens
- incomplete commit message for screenreaders

client/web-sveltekit/src/lib/repo/LastCommit.svelte
```diff
@@ -0,0 +1,89 @@
...
+        canonicalURL: string
+        author: {
+            date: string
+            person: {
+                avatarURL: string | null
+                displayName: string
+                name: string
+            }
+        }
+    }
```
fkling, line 15: If you create a `LastCommit.gql` file with the right fragment then you don't have to declare this type, it will be generated for you from the `.gql` file.

client/web-sveltekit/src/lib/repo/utils.ts
```diff
@@ -64,3 +65,50 @@ export async function resolveRevision(
...
+}
+
+function hasPRNumber(cm: string): boolean {
+    let words = cm.split(' ')
+    for (let word of words) {
+        if (/\(#(\d+)\)/.test(word)) {
+            return true
+        }
+    }
+    return false
```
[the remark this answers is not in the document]
jasonhawkharris: Nice! TIL!

client/web-sveltekit/src/lib/repo/utils.test.ts
jasonhawkharris: Okay, we are now preserving capitalization on the user's name.

client/web-sveltekit/src/lib/repo/utils.ts
```diff
@@ -64,3 +65,50 @@ export async function resolveRevision(
...
+
+export function convertToElapsedTime(commitDateString: string): string {
+    const commitDate = new Date(commitDateString)
+    return formatDistanceToNow(commitDate, { addSuffix: true })
+}
+
+export function truncateIfNeeded(cm: string): string {
+    cm = extractCommitMessage(cm)
+    return cm.length > 23 ? cm.substring(0, 23) + '...' : cm
+}
```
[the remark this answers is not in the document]
jasonhawkharris: Moved everything out of the utils file ended up deleting most of these helpers anyways. I'm using text-overflow to generate the truncation now per @camdencheek's earlier suggestion.

client/web-sveltekit/src/lib/repo/utils.ts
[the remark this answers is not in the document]
jasonhawkharris: Compare this to the screen shots in the initial post on this PR:

Screenshot 2024-03-29 at 11 49 28 AM
Screenshot 2024-03-29 at 11 50 37 AM

IMO, this is clearer, and presents fewer edge cases we'd need to consider.  Thoughts @fkling @camdencheek?

client/web-sveltekit/src/lib/repo/utils.ts
```diff
@@ -64,3 +65,50 @@ export async function resolveRevision(
...
+    return `${names[0]} ${names[names.length - 1].charAt(0).toUpperCase()}.`
+}
+
+export function extractPRNumber(cm: string): string | null {
+    if (!hasPRNumber(cm)) {
+        return null
+    }
+    let cmWords = cm.split(' ')
+    let sha = cmWords[cmWords.length - 1]
+    return sha.slice(1, sha.length - 1)
```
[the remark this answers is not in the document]
jasonhawkharris: See notes starting from [here](https://github.com/sourcegraph/sourcegraph/pull/61426#discussion_r1544686542)

client/web-sveltekit/src/lib/repo/LastCommit.svelte
jasonhawkharris: Do you think it's okay to link to whatever the value of `cannonicalURL` is here instead of the corresponding PR? I don't see that information present in the data.

jasonhawkharris: The issue here is that for commit messages that contain the PR number, linking the commit link (which is what the data provides) only goes to the commit, not the PR. Even when the whole line is highlighted, it's misleading that it would link to the commit instead of the PR. 

Seems like there are two options to me: 
1. Generate a PR link on the backend, and link that instead of the commit. 
2. Keep the commit link instead, and link the entire line.

client/web-sveltekit/src/lib/repo/LastCommit.svelte
jasonhawkharris: Could you help me understand why that would be a better approach than just declaring the type?

fkling: The type will be generated from the GraphQL schema and thus will always be in sync and correct.

client/web-sveltekit/src/lib/repo/LastCommit.gql
```diff
@@ -0,0 +1,14 @@
...
+    id
+    abbreviatedOID
+    subject
+    canonicalURL
+    author {
+        date
+        person {
+            displayName
+            name
+            avatarURL
```
camdencheek, line 11: Should this be `...Avatar_User` or something?

client/web-sveltekit/src/lib/repo/LastCommit.svelte
```diff
@@ -0,0 +1,75 @@
...
+    import Avatar from '$lib/Avatar.svelte'
+
+    import type { LastCommitFragment } from './LastCommit.gql'
+
+    export let latestCommit: LastCommitFragment
+
+    $: user = latestCommit.author.person
+    $: canonicalURL = latestCommit.canonicalURL
+    $: commitMessage = latestCommit.subject
+    $: commitDate = formatDistanceToNow(latestCommit.author.date, { addSuffix: false })
```
camdencheek, line 13: I think this should drop the `about` prefix and add the `ago` suffix. Right now, it reads "about 1 year" but I think it makes more sense to say "1 year ago"

client/web-sveltekit/src/lib/repo/LastCommit.svelte
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    .display-name {
+        margin-left: 0.4rem;
+    }
+
+    .latest-commit {
+        display: flex;
+        flex-flow: row nowrap;
+        padding-right: 0.5rem;
```
camdencheek, line 61: This `padding-right` should be handled in the parent container, not in the component itself.

client/web-sveltekit/src/lib/repo/LastCommit.svelte
```diff
@@ -0,0 +1,75 @@
...
+        white-space: nowrap;
+        justify-content: space-between;
+        margin-right: 0.5rem;
+    }
+
+    .user-info {
+        display: flex;
+        flex-flow: row nowrap;
+        align-items: center;
+        margin-right: 0.6rem;
```
camdencheek, line 73: I see 3 different margin sizes in this component: 0.4rem, 0.5rem, and 0.6rem. We don't enforce specific spacing sizes, but it's generally limited to power-of-two fractions (e.g. 0.5, 0.25, 0.125, 0.375). Let's stick to 0.5 for all of these so it's all consistent.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+page.svelte
```diff
@@ -2,13 +2,13 @@
...
     import Icon from '$lib/Icon.svelte'
+    import Readme from '$lib/repo/Readme.svelte'
     import SidebarToggleButton from '$lib/repo/SidebarToggleButton.svelte'
     import { sidebarOpen } from '$lib/repo/stores'
     import { createPromiseStore } from '$lib/utils'
-    import Readme from '$lib/repo/Readme.svelte'
-    import type { RepoPage_Readme } from './page.gql'
 
     import type { PageData } from './$types'
+    import type { RepoPage_Readme } from './page.gql'
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+page.svelte:11]: Is this used?

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte
```diff
@@ -70,6 +71,7 @@
     $: ({ revision = '', parentPath, repoName, resolvedRevision } = data)
     $: fileTreeStore.set({ repoName, revision: resolvedRevision.commitID, path: parentPath })
     $: commitHistoryQuery = data.commitHistory
+    $: latestCommit = commitHistory?.nodes[0]
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte:74]: This doesn't look like it's actually _using_ your new `LastCommitQuery`, is it?

client/web-sveltekit/src/lib/repo/LastCommit.svelte
camdencheek, line 1: The resize behavior on this component doesn't look right. Seems it doesn't actually shrink, and it overlaps with the tabs once the screen is too small. 

I would expect the `max-width` to apply to the commit message (not the component as a whole) since that's really the only resizeable element here.

https://github.com/sourcegraph/sourcegraph/assets/12631702/46571fa2-b7b4-449a-9433-05214c7b7cfb

camdencheek, line 1: Looks to me like the text isn't quite vertically centered?

CleanShot 2024-04-01 at 11 38 18@2x

client/web-sveltekit/src/lib/repo/LastCommit.svelte
jasonhawkharris: I'm using a library for this, and it doesn't give you an option to remove the prefix, only the suffix. Would adding back the suffix be enough?

camdencheek: Oh! In the React [Timestamp component](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@698bc0590435b619f2f6aa68823d59e2e3e4f450/-/blob/client/branded/src/components/Timestamp/Timestamp.tsx?L47-47), we have a `noAbout` option. I didn't realize that wasn't built into the library.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+page.svelte
Comment (jasonhawkharris) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+page.svelte:11]: Pretty sure it's used in line 15.

Comment (camdencheek) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+page.svelte:11]: You're totally right. I just completely misread the diff 🤦

client/web-sveltekit/src/lib/repo/LastCommit.svelte
camdencheek: Any reason you decided against using the `Timestamp.svelte` component?

camdencheek (approved):
I think the max width should be larger. On large screens, we've got a lot of real estate to work with that we're not using, and the truncated commit message is often pretty useless. 

CleanShot 2024-04-02 at 14 08 42@2x

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte
```diff
@@ -66,17 +68,21 @@
...
 
     $: ({ revision = '', parentPath, repoName, resolvedRevision } = data)
     $: fileTreeStore.set({ repoName, revision: resolvedRevision.commitID, path: parentPath })
     $: commitHistoryQuery = data.commitHistory
+    $: lastCommitQuery = data.lastCommit
     $: if (!!commitHistoryQuery) {
         // Reset commit history when the query observable changes. Without
         // this we are showing the commit history of the previously selected
         // file/folder until the new commit history is loaded.
         commitHistory = null
```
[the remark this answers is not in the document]
Comment (camdencheek) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte:81]: confirmed that, when switching to a new file, we show the wrong last commit until the new one loads

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/layout.gql
```diff
@@ -10,6 +10,19 @@ query GitHistoryQuery($repoName: String!, $revspec: String!, $first: Int, $after
     }
 }
 
+query LastCommitQuery($repoName: String!, $revspec: String!, $filePath: String) {
```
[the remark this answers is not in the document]
Comment (jasonhawkharris) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/layout.gql]: yup!

fkling:
I think `width` is another thing that shouldn't be controlled by the component itself most of the time.
I see no reason why the component shouldn't use as much space as available here (with some reasonable visual separation from the tab).

PR #61428, opened 2024-03-27
msp: add yaml langserver magic comment for schema

Strum355:
If opening a generated MSP service file in your editor, but as part of a different workspace/repo, then vscode doesnt pick up what schema file to use for it, resulting in no intellisense. The yaml langserver supports a magic comment to point to a relative file, that works in this case :slightly_smiling_face: 

## Test plan

Validated locally with local sg build and `sg msp init ...`

### DIFF
--- a/dev/sg/msp/example/job.template.yaml
+++ b/dev/sg/msp/example/job.template.yaml
@@ -1,3 +1,4 @@
+# yaml-language-server: $schema=../../schema/service.schema.json
 service:
   kind: job
   id: {{ .ID }}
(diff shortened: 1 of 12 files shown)


jac (approved):
This is great!

As a bonus this also works makes it work in Zed!

bobheadxi:
Hmm the managed-services repo already has a workspace config file to use the schema, but if this helps why not :)

Strum355:
> Hmm the managed-services repo already has a workspace config file to use the schema, but if this helps why not :)

That only works if youve got the repo open, in my case I was working in sg/sg and opened up a newly created msp service file in my sg/sg editor workspace

PR #61429, opened 2024-03-27
gitserver: Mark perforce repos correctly as recently fetched

eseliger:
Our git logic determines the last fetch date by looking at the mtime of the FETCH_HEAD or HEAD file. Since for perforce we don't have a FETCH_HEAD, and HEAD is not touched because we don't do an actual git fetch, just run p4-fusion, this is not actually set correctly.

Thus, the search will show a red icon saying this repo is _severely_ out of date, incorrectly.

This fixes the issue by adjusting the mtime of HEAD after a successful p4-fusion fetch.

## Test plan

Replicated the setup locally and verified that after this change the repo is now correctly showing as fresh.

### DIFF
--- a/cmd/gitserver/internal/vcssyncer/perforce.go
+++ b/cmd/gitserver/internal/vcssyncer/perforce.go
@@ -7,6 +7,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strconv"
+	"time"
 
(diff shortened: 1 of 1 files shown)


peterguy (approved):
Hah, good catch!

mmanela (approved):

pjlast (approved):

PR #61433, opened 2024-03-27
Fix INC 290

thenamankumar:
context: https://sourcegraph.slack.com/archives/C06R89KF3AB/p1711569914628259
We have noticed a huge uptick in Claude 3 usage which was tracked down to one abusing user. The user has the correct quota of 1 but has over 2k API keys. The current theory is that they are reusing API keys and abusing the 1 quota.

Root cause:
If the user uses a new access token to make a request to cody gateway and the actor is not saved in the cache against that new token, we would reset the usage. This allowed users to use Cody without limits by keep generating new tokens. 

The fix is to check the cache for actor against the user id while `fetchAndCache` and not reset the usage if actor is found in the cache aginst the user id. 

## Test plan

https://www.loom.com/share/724bb4383d864805a7b61a4ccb46303c

### DIFF
--- a/cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
+++ b/cmd/cody-gateway/internal/actor/dotcomuser/dotcomuser.go
@@ -90,25 +90,29 @@ func (s *Source) fetchAndCache(ctx context.Context, token string, oldAct *actor.
 	} else {
 		act = newActor(s, token,
 			resp.Dotcom.CodyGatewayDotcomUserByToken.DotcomUserState, s.concurrencyConfig)
-
-		// It might be the case that this is a new token for which the actor is not stored in the cache yet.
-		// But the actor against the actor.ID (user_id) might be already stored in the cache. This is possible
-		// in the case user have used a different token before this.
-		// So we need to check cache for the actor against the act.ID (user_id), otherwise the cache will be reset,
-		// every time a new token is used by the same user.
-		data, hit := s.cache.Get(act.ID)
-		if hit {
-			var actorFromCache *actor.Actor
-			if err := json.Unmarshal(data, &actorFromCache); actorFromCache != nil && err == nil {
-				return actorFromCache, nil
-			}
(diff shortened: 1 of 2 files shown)


chrsmith (commented):
I don't think this is what we want, because it fails to persist the actor within the cache by its access token. (i.e. we don't call `s.cache.Set(token, data)` like we used to on line 118.

While that's not something we want to continue doing in the future, we need to maintain that invariant until we exclusively rely on caching the data by the user ID.

Instead, I think we'd want to move your "lookup actor by ID" check lower, to line 112, before we call `s.maybeResetUsageData`.

Maybe the right thing to do here is:

```go
// The caller didn't know what the previous actor state was.
// But now we do. So it could be the case that we can use the existing
// cached data.
if oldAct == nil {
    oldAct = json.Unmarshall(s.cache.Get(act.ID)
}
```

... then we call `maybeResetUsageData` like before?

thenamankumar:
> I don't think this is what we want, because it fails to persist the actor within the cache by its access token. (i.e. we don't call `s.cache.Set(token, data)` like we used to on line 118.
> 
> While that's not something we want to continue doing in the future, we need to maintain that invariant until we exclusively rely on caching the data by the user ID.
> 
> Instead, I think we'd want to move your "lookup actor by ID" check lower, to line 112, before we call `s.maybeResetUsageData`.
> 
> Maybe the right thing to do here is:
> 
> ```go
> // The caller didn't know what the previous actor state was.
> // But now we do. So it could be the case that we can use the existing
> // cached data.
> if oldAct == nil {
>     oldAct = json.Unmarshall(s.cache.Get(act.ID)
> }
> ```
> 
> ... then we call `maybeResetUsageData` like before?

Makes total sense, Updated the PR

chrsmith (approved):
@thenamankumar said he ran the changes locally and it all looked legit. I pushed some unit tests, which confirm the new behavior.

So this should be good to deploy! 🚀

PR #61434, opened 2024-03-27
Fix and optimize repo exclusions

eseliger:
There was a bug in this logic: For some code host types, we would overwrite the value of exact if both Name and ID were specified in an exclusion rule, for example. Instead, we should OR match on _any_ given exact match. 

This is just a side-effect though, the actual improvement here is that we no longer need to iterate over thousands of rules for simple exact matches. We have a customer who wants to use potentially up to 20000 statements, and during code host syncing, this would need to run 20000 times, for a total of 400,000,000 exact comparisons. 

I added a benchmark for this. Before:

```
goos: darwin
goarch: arm64
pkg: github.com/sourcegraph/sourcegraph/internal/repos
BenchmarkExcludeExact-10            8205            144040 ns/op               0 B/op          0 allocs/op
PASS
```

After:

```
goos: darwin
goarch: arm64
pkg: github.com/sourcegraph/sourcegraph/internal/repos
BenchmarkExcludeExact-10        57618390                20.75 ns/op            0 B/op          0 allocs/op
PASS
```

## Test plan

Adjusted existing tests, other existing tests are still passing, and added a benchmark to verify the improvement.

### DIFF
--- a/internal/repos/awscodecommit.go
+++ b/internal/repos/awscodecommit.go
@@ -87,9 +87,10 @@ func newAWSCodeCommitSource(svc *types.ExternalService, c *schema.AWSCodeCommitC
 
 	var ex repoExcluder
 	for _, r := range c.Exclude {
-		ex.AddRule().
+		// Either Name OR ID must match.
+		ex.AddRule(NewRule().
 			Exact(r.Name).
-			Exact(r.Id)
+			Exact(r.Id))
 	}
 	if err := ex.RuleErrors(); err != nil {
 		return nil, err
--- a/internal/repos/exclude.go
+++ b/internal/repos/exclude.go
@@ -24,10 +24,17 @@ type excludeFunc func(input any) bool
 // repo should be excluded. In that call, all the rules are OR'd together. If
 // one rule excludes, the repo is excluded.
 type repoExcluder struct {
-	rules []*rule
+	exactRules map[string]struct{}
+	rules      []*rule
 }
 
 func (e *repoExcluder) ShouldExclude(input any) bool {
+	if inputString, ok := input.(string); ok {
+		if _, exists := e.exactRules[strings.ToLower(inputString)]; exists {
+			return true
+		}
+	}
+
 	for _, r := range e.rules {
 		if r.Excludes(input) {
 			return true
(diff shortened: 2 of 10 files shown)


internal/repos/awscodecommit.go
```diff
@@ -87,9 +87,10 @@ func newAWSCodeCommitSource(svc *types.ExternalService, c *schema.AWSCodeCommitC
 
 	var ex repoExcluder
 	for _, r := range c.Exclude {
-		ex.AddRule().
+		// Either Name OR ID must match.
+		ex.AddRule(NewRule().
```
mmanela, line 91: Why does this make it an OR?

eseliger: Just documenting what this is supposed to do, that's what it should always have done but because Exact was overwriting and not appending before this change, a rule of `{name:a,id:123}` would previously have overwritten the exact match for `a` with `123`, so this would match anything with 123, and nothing with `a`.

internal/repos/exclude_test.go
```diff
@@ -61,9 +70,11 @@ func TestRuleExcludes(t *testing.T) {
 	t.Run("multiple conditions", func(t *testing.T) {
 		r := &rule{}
 		r.Exact("foobar")
+		r.Exact("foobar2")
```
mmanela, line 73: In real life what does exact path look like? repo url? or fully qualified repo (mmanela/foo)?

internal/repos/exclude.go
```diff
@@ -24,10 +24,17 @@ type excludeFunc func(input any) bool
 // repo should be excluded. In that call, all the rules are OR'd together. If
 // one rule excludes, the repo is excluded.
 type repoExcluder struct {
-	rules []*rule
+	exactRules map[string]struct{}
+	rules      []*rule
 }
 
 func (e *repoExcluder) ShouldExclude(input any) bool {
```
mmanela, line 31: What can input be besides a string?

mmanela (approved):

internal/repos/exclude_test.go
eseliger: It is the name of the repo on the code host. How that looks like precisely depends on the code host - for type Other (aka you give us a git URL) it can be pretty much anything (plus normalization). `erik`, `golang`, `sourcegraph/sourcegraph`. Most code hosts have a `<org>/<name>` structure, but not very strict.

internal/repos/exclude.go
eseliger: a github.Repo for example, when passed to a `Generic` rule evaluator, it can cast it back and check for some metadata properties.

Some examples of other types are at the bottom of this file for size constraint exclusion rules

Issue #61298, opened 2024-03-20
Add reliable way to opt out of the Svelte experiment


fkling:
We want to rollout specific pages of the Svelte prototype on S2. To do this we can enabled the `web-next-rollout` feature flag. This feature flag tells the server to serve the Svelte application for specific routes.

We should also provide a simple way for the user to opt-out of this. It's technically possible to do this by adding feature flag overrides to the URL, but that ultimately results in  a broken experience:

- The server respects the override and serve the React version. The override is persisted in local storage and and whenever we check for the feature flag value it will now be false. Most importantly that causes links to stay inside the React app.
- Problem: When the user goes to S2 again (e.g. in a new tab) they will be served the Svelte version (the server doesn’t now about any override since it’s not in the URL). This might already be surprising to the user. Now if they navigate to the link in the React version and e.g. click Search Home in the menu they will stay in the React version because that link is affected by the local override. This is an inconsistent experience.

Things to do:

- Allow user's to override feature flags for themselves. If we can permanently set the feature flag on the server we avoid the server state/local state mismatch.
- Update the opt-out link to set the feature flag on the server instead.
- Replace the current beaker button in React app with a more descriptive button to opt back in and possible also opt into the full svelte experience (i.e. the repo pages) by overriding the `web-next` feature flag (which is what the beaker button currently does but only via the URL, not permanently)

camdencheek:
Okay, so, looking closer, I do not think feature flags are going to be the best way to move forward. They'll require some significant updates to get working for this work.
1) Currently, we only have "create", "delete", and "update" operations. No "create or update", which means we need to make one request to fetch the override ID and a second to update or create it. Not ideal.
2) Modifying "create" to act like "create or update" is also a pain since we have two unique constraints and Postgres does not allow "ON CONFLICT" clauses for two different constraints. (Looking forward to the MERGE command in PG15).
3) The DB constraints on feature flags also require that the feature flag exists in order to create an override for it. Also not ideal for this scenario.
4) The discoverability of feature flags is still not great. 

Our settings API already allows patching settings, so I think maybe we just make the svelte webapp and svelte rollout experimental features.

fkling:
Can we easily read the settings on the server during the request?

camdencheek:
Yes, we can

PR #61440, opened 2024-03-27
Svelte: persist opt in/out

camdencheek:
Basically, when you opt into or out of the svelte experiment, it persists that choice. The rollout flag is now only effectively used if you have not done either of these actions. Once you opt in or out, that choice is remembered by saving a feature flag override.

Previously, enabling the feature was done with URL-level feature flag overrides, which is problematic because they don't stay enabled/disabled.

This should be the last thing needed before enabling by default on S2. Closes https://github.com/sourcegraph/sourcegraph/issues/61298

## Test plan

https://github.com/sourcegraph/sourcegraph/assets/12631702/038cfb5b-2d30-46a6-8299-7fde46562e77

### DIFF
--- a/client/web-sveltekit/src/routes/+layout.ts
+++ b/client/web-sveltekit/src/routes/+layout.ts
@@ -62,10 +62,10 @@ export const load: LayoutLoad = async ({ fetch }) => {
             }
             return result.data.evaluatedFeatureFlags
         },
-        disableSvelteFeatureFlags: async () => {
+        disableSvelteFeatureFlags: async (userID: string) => {
             const mutationResult = await client.mutation(
                 DisableSveltePrototype,
-                { userID: result?.data?.currentUser?.id },
+                { userID },
                 { requestPolicy: 'network-only', fetch }
             )
             if (!mutationResult.data || mutationResult.error) {
--- a/client/web-sveltekit/src/routes/+layout.ts
+++ b/client/web-sveltekit/src/routes/+layout.ts
@@ -6,7 +6,7 @@ import { getGraphQLClient } from '$lib/graphql'
 import type { Settings } from '$lib/shared'
 
 import type { LayoutLoad } from './$types'
-import { Init, EvaluatedFeatureFlagsQuery, GlobalAlertsSiteFlags } from './layout.gql'
+import { Init, EvaluatedFeatureFlagsQuery, GlobalAlertsSiteFlags, DisableSveltePrototype } from './layout.gql'
 
 // Disable server side rendering for the whole app
 export const ssr = false
@@ -61,6 +61,16 @@ export const load: LayoutLoad = async ({ fetch }) => {
                 throw new Error(`Failed to fetch evaluated feature flags: ${result.error}`)
             }
             return result.data.evaluatedFeatureFlags
+        },
+        disableSvelteFeatureFlags: async () => {
+            const mutationResult = await client.mutation(
+                DisableSveltePrototype,
+                { userID: result?.data?.currentUser?.id },
+                { requestPolicy: 'network-only', fetch }
+            )
+            if (!mutationResult.data || mutationResult.error) {
+                throw new Error(`Failed to disable svelte feature flags: ${result.error}`)
+            }
         },
     }
 }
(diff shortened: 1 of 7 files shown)


client/web-sveltekit/src/routes/+layout.ts
```diff
@@ -62,5 +62,15 @@ export const load: LayoutLoad = async ({ fetch }) => {
...
+        disableSvelteFeatureFlags: async () => {
+            const mutationResult = await client.mutation(
+                DisableSveltePrototype,
+                { userID: result?.data?.currentUser?.id },
+                { requestPolicy: 'network-only', fetch }
+            )
+            if (!mutationResult.data || mutationResult.error) {
+                throw new Error(`Failed to disable svelte feature flags: ${result.error}`)
+            }
+        },
```
fkling, line 74: That's exactly how I expected it to be done 🙂

client/web-sveltekit/src/routes/+layout.ts
```diff
@@ -62,5 +62,15 @@ export const load: LayoutLoad = async ({ fetch }) => {
             }
             return result.data.evaluatedFeatureFlags
         },
+        disableSvelteFeatureFlags: async () => {
+            const mutationResult = await client.mutation(
+                DisableSveltePrototype,
+                { userID: result?.data?.currentUser?.id },
```
fkling, line 68: What will happen if `userID` ends up being `undefined`?

client/web-sveltekit/src/routes/Header.svelte
```diff
@@ -1,33 +1,25 @@
...
-            if (feature !== 'web-next' && feature !== 'web-next-rollout') {
-                urlCopy.searchParams.append('feat', feature)
-            }
-        }
-        urlCopy.searchParams.append('feat', '-web-next')
-        urlCopy.searchParams.append('feat', '-web-next-rollout')
-        return urlCopy.toString()
-    })($page.url)
+    async function handleOptOut() {
+        await $page.data.disableSvelteFeatureFlags()
```
fkling, line 20: Should we pass this in as a prop instead? I'm not sure yet how to best use `$page.data`. Also we should either catch any error and show it.

client/web-sveltekit/src/routes/Header.svelte
```diff
@@ -91,6 +78,16 @@
         background-color: var(--color-bg-1);
     }
 
+    .opt-out {
+        background: none;
+        border: none;
+        padding: 0;
+        font: inherit;
```
fkling, line 85: You can also use [`all: unset;`](https://developer.mozilla.org/en-US/docs/Web/CSS/all) instead.

bahrmichael (approved):

fkling (approved):
I think we should also make the icon in the React app a bit ... "better". I don't know what to do exactly though. Maybe @taiyab has an idea. But if not I guess it's fine.

client/web-sveltekit/src/routes/+layout.ts
camdencheek: Oh, actually, this is a type error that I missed. Updating this so it takes a (non-optional) user ID as a param and disables the opt out

client/web-sveltekit/src/routes/Header.svelte
camdencheek: Ah, yeah. I guess we should probably prefer passing things in rather than circumventing the component signature.

client/web-sveltekit/src/routes/Header.svelte
camdencheek: TIL!

vovakulikov: But you need to be careful with `all: unset` since it resets all styles (not only default ones that we haven't specified), in this case `all: unset` should be before you set rules like `color` and `text-decoration`

camdencheek:
Will follow up with some design updates, but don't want to block rollout on that

taiyab:
> I think we should also make the icon in the React app a bit ... "better". I don't know what to do exactly though. Maybe @taiyab has an idea. But if not I guess it's fine.

I'm fine with it. Let's just keep going and we can come back to this later.

PR #61463, opened 2024-03-28
searcher: remove unused dependency on dbconn pkg

keegancsmith:
In a recent incident it was bought up that hot fixing searcher had a risk since it speaks to the DB. I investigated and noticed it doesn't actually connect, but does have dbconn as an unused transitive dep.

This commit refactors out the bit of logic used by searcher into a pure package. To do so I did a few renames, but functionally there is no change in logic other than the searcher binary shedding some heavy dependencies.

Additionally I updated the go-dbconn-import check to ensure searcher isn't accidently using the database again.

Test Plan: CI

### DIFF
--- a/cmd/searcher/internal/search/BUILD.bazel
+++ b/cmd/searcher/internal/search/BUILD.bazel
@@ -41,7 +41,7 @@ go_library(
         "//internal/search/backend",
         "//internal/search/casetransform",
         "//internal/search/query",
-        "//internal/search/zoekt",
+        "//internal/search/zoektquery",
         "//internal/searcher/v1:searcher",
         "//internal/trace",
         "//lib/codeintel/languages",
--- a/cmd/searcher/internal/search/zoekt_search.go
+++ b/cmd/searcher/internal/search/zoekt_search.go
@@ -11,53 +11,53 @@ import (
 	"github.com/sourcegraph/conc/pool"
 	"github.com/sourcegraph/log"
 	"github.com/sourcegraph/zoekt"
-	zoektquery "github.com/sourcegraph/zoekt/query"
+	"github.com/sourcegraph/zoekt/query"
 
 	"github.com/sourcegraph/sourcegraph/cmd/searcher/protocol"
 	"github.com/sourcegraph/sourcegraph/internal/api"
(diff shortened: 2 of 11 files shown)


keegancsmith:
@michaellzc FYI searcher doesn't speak to the DB. Are there deployment assets that we need to update to avoid db sidecars/etc?

@burmudar out of interest is `dev/check/go-dbconn-import.sh` still run in CI or do I need to update some bazel magic?

burmudar:
> @michaellzc FYI searcher doesn't speak to the DB. Are there deployment assets that we need to update to avoid db sidecars/etc?
> 
> @burmudar out of interest is `dev/check/go-dbconn-import.sh` still run in CI or do I need to update some bazel magic?

@keegancsmith I believe we have a `nogo` linter that does that but let me double check and confirm

burmudar:
> > @michaellzc FYI searcher doesn't speak to the DB. Are there deployment assets that we need to update to avoid db sidecars/etc?
> > @burmudar out of interest is `dev/check/go-dbconn-import.sh` still run in CI or do I need to update some bazel magic?
> 
> @keegancsmith I believe we have a `nogo` linter that does that but let me double check and confirm

@keegancsmith confirmed! see [dbconn nogo linter](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@748e8b226b4cf02f0610f7638ef268b3f72000eb/-/blob/dev/linters/dbconn/dbconn.go) and it is registered with `nogo` [here](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/BUILD.bazel?L304)

stefanhengl (approved):

michaellzc:
> FYI searcher doesn't speak to the DB. Are there deployment assets that we need to update to avoid db sidecars/etc?

No sidecar, but we do have network policy in place to permit who can talk to db on Cloud.

The source of truth is here:

https://github.com/sourcegraph/sourcegraph/blob/1c5f545a4597a03a2b9df066017ef8351d0ad7b8/lib/servicecatalog/service-catalog.yaml#L46-L58

PR #61465, opened 2024-03-28
Add syntactic indexing experimental feature (disabled)

keynmol:
We seem to have

- Feature flags: https://sourcegraph.com/docs/dev/how-to/use_feature_flags#when-to-use-feature-flags - **I don't think they fit here**
- Site config: https://sourcegraph.com/docs/admin/config/site_config - designed to be enabled/disabled by site admins. **At this point, I don't think this should be an admin-facing setting**
- Settings? https://sourcegraph.com/docs/admin/config/settings. Which seem to contain `experimentalFeatures` object that contains features similar to what we're building: https://sourcegraph.com/docs/admin/how-to/enable-experimental-feature. **Let's put it there**

The `schema.go` file is generated automatically.

## Test plan

N/A


varungandhi-src (approved):
This looks OK, I'm guessing you wanted to add checks for this flag (similar to `autoIndexingEnabled()`) in a different PR?

keynmol:
Yeah, I want to add accessors only alongside the code that uses them.

PR #61466, opened 2024-03-28
Follow UP Pr to adding a GQL warning for not using Graphql while using Cody API 

arafatkatze:
Coming from https://github.com/sourcegraph/sourcegraph/pull/61381#issuecomment-2021634637

image

## Test plan
When you run a local instance and use the completions endpoint you will see a yellowed out sign at the bottom and a warning

### DIFF
--- a/cmd/frontend/graphqlbackend/completions.graphql
+++ b/cmd/frontend/graphqlbackend/completions.graphql
@@ -4,8 +4,8 @@ extend type Query {
     """
     completions(input: CompletionsInput!, fast: Boolean = false): String!
         @deprecated(
-            reason: "Consumers should use non graphql endpoints when communicating the completions endpoint. Those endpoints can be found in this file -> https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/services/AuthProvider.ts"
-        )
+            reason: "GraphQL API for completions is deprecated and will be removed in the future"
+    )
 }
 
 """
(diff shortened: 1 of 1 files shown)


cmd/frontend/graphqlbackend/completions.graphql
```diff
@@ -3,6 +3,9 @@ extend type Query {
     Returns a string of completion responses
     """
     completions(input: CompletionsInput!, fast: Boolean = false): String!
+        @deprecated(
+            reason: "Consumers should use non graphql endpoints when communicating the completions endpoint. Those endpoints can be found in this file -> https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/services/AuthProvider.ts"
```
slimsag, line 7: Can you link to (or create if needed) a page on https://sourcegraph.com/docs that explains with e.g. curl examples how to use the Cody API? That would be the best-case scenario here

Afraid no site admin would know what to do with that typescript file.

slimsag: (just an idea, otherwise not sure the motivation for linking the typescript file)

arafatkatze: image

slimsag: okay, sounds fine then

taras-yemets (approved):

arafatkatze:
Merged already in another PR

PR #61467, opened 2024-03-28
PLG: Add missing events

vdavid:
- Fixes https://github.com/sourcegraph/cody/issues/2361

Added the three new event types and triggering them when needed.

Caveats/questions:
- Wasn't sure what to use for the exact name of the editors, I used "JetBrains", "VS Code", and "NeoVim".
- About the naming for the newer telemetryRecorder thing, I just came up with the feature names (e.g. `cody.onboarding.openMarketplace` and the metadata format, not sure if that's needed/right)

## Test plan

Ran the site locally and checked if each of the events get triggered.

I couldn't actually find a way to trigger the "Cody Features" view 🤔 but the logging is there the same way as for the setup instructions page, so it should work the same way.

### DIFF
--- a/client/web/src/cody/onboarding/instructions/JetBrains.tsx
+++ b/client/web/src/cody/onboarding/instructions/JetBrains.tsx
@@ -1,8 +1,12 @@
-import { useState } from 'react'
+import { useState, useEffect } from 'react'
 
 import classNames from 'classnames'
 
+import type { TelemetryRecorder } from '@sourcegraph/shared/src/telemetry'
 import { Button, H2, Link, Text } from '@sourcegraph/wildcard'
+
+import { eventLogger } from '../../../tracking/eventLogger'
+import { EventName } from '../../../util/constants'
 
 import styles from '../CodyOnboarding.module.scss'
 
@@ -10,12 +14,25 @@ export function JetBrainsInstructions({
     onBack,
     onClose,
     showStep,
+    telemetryRecorder,
 }: {
     onBack?: () => void
     onClose: () => void
     showStep?: number
+    telemetryRecorder: TelemetryRecorder
 }): JSX.Element {
     const [step, setStep] = useState<number>(showStep || 0)
+    const marketplaceUrl = 'https://plugins.jetbrains.com/plugin/9682-sourcegraph-cody--code-search'
+
+    useEffect(() => {
+        if (step === 0) {
+            eventLogger.log(EventName.CODY_EDITOR_SETUP_VIEWED, { editor: 'JetBrains' })
+            telemetryRecorder.recordEvent('cody.editorSetupViewed', 'view', { metadata: { jetBrains: 1 } })
+        } else if (step === 1) {
+            eventLogger.log(EventName.CODY_EDITOR_FEATURES_VIEWED, { editor: 'JetBrains' })
+            telemetryRecorder.recordEvent('cody.editorFeaturesViewed', 'view', { metadata: { jetBrains: 1 } })
+        }
+    }, [step, telemetryRecorder])
 
     return (
         <>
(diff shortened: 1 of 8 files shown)


client/web/src/cody/onboarding/instructions/JetBrains.tsx
```diff
@@ -1,21 +1,38 @@
...
     onBack?: () => void
     onClose: () => void
     showStep?: number
+    telemetryRecorder: TelemetryRecorder
 }): JSX.Element {
     const [step, setStep] = useState<number>(showStep || 0)
+    const marketplaceUrl = 'https://plugins.jetbrains.com/plugin/9682-sourcegraph-cody--code-search'
+
+    useEffect(() => {
+        if (step === 0) {
```
chrsmith, line 28: `step === 0` seems suuuuper fragile. And not having any sort of comment on top of `const [step, setStep] = ...` seems problematic.

Can you update the code to add some comments here? Or even better, define some constants for readability? e.g. `if (step === steps.choseEditor)` or similar.

... that might be a little annoying because we are using the same functionality in all of these.

But would it make sense then to generalize this? e.g.

```ts
export interface CreateReactWidgetThingeyOptions {
    marketplaceUrl: string;
    editorName: string;  // e.g. "JetBrains"
    eventMetadata: Object;  // Sent with telemetry, e.g. { jetBrains: 1 }
}

// CreateReactWidgetThingey will create the same "pick an editor" widget,
// just customized with the specific image, onClick URL, etc.
export function CreateReactWidgetThingey(opts): ReactWidget {
    ...
```

chrsmith: It seems like we can do a little bit of work and leave this part of the codebase a lot cleaner than we found it.

chrsmith (approved):
As far as I know, this code all looks good. Though it seems like there is a good opportunity for us to generalize these widgets, and just create a single reusable one. Rather than duplicating the specific logic/behavior/styling again and again. What do you think?

client/web/src/cody/onboarding/instructions/JetBrains.tsx
vdavid: I've introduced an enum to replace the magic values in https://github.com/sourcegraph/sourcegraph/pull/61467/commits/33b27453016b7a088fb5e6c1f806f54bef1a9c71, but didn't do the generalization just yet.

chrsmith: Thanks. No need to block merging this PR without the bigger refactoring... although, you know, that would be score you some pretty sweet "code karma". :D

vdavid: Okay, I did another round because there was a 94-line chunk that was _triplicated_, so I've extracted that into a component in this commit: https://github.com/sourcegraph/sourcegraph/pull/61467/commits/8795f5090b214ae5d235bb266645bfb527fa5f3e
But the rest of it is custom enough so extracting the common parts would've actually been more confusing than helpful.

client/web/src/cody/onboarding/instructions/JetBrains.tsx
```diff
@@ -1,25 +1,45 @@
...
+    const [step, setStep] = useState<EditorStep>(showStep || 0)
+    const marketplaceUrl = 'https://plugins.jetbrains.com/plugin/9682-sourcegraph-cody--code-search'
+
+    useEffect(() => {
+        if (step === EditorStep.SetupInstructions) {
+            eventLogger.log(EventName.CODY_EDITOR_SETUP_VIEWED, { editor: 'JetBrains' })
+            telemetryRecorder.recordEvent('cody.editorSetupViewed', 'view', { metadata: { jetBrains: 1 } })
+        } else if (step === EditorStep.CodyFeatures) {
+            eventLogger.log(EventName.CODY_EDITOR_FEATURES_VIEWED, { editor: 'JetBrains' })
+            telemetryRecorder.recordEvent('cody.editorFeaturesViewed', 'view', { metadata: { jetBrains: 1 } })
```
bobheadxi, line 36: Nit - the feature here is the `editorSetup` or `editorFeatures` pages, while the _action_ is `view`, so there is no need to repeat the `view` verb here. Instead, this should be:

```
telemetryRecorder.recordEvent('cody.editorSetupPage', 'view', { metadata: { jetBrains: 1 } })
 telemetryRecorder.recordEvent('cody.editorFeaturesPage', 'view', { metadata: { jetBrains: 1 } })
```

bobheadxi: The goal should be to align all related events under shared `feature`, and then have each `feature` record multiple types of `actions`. This helps the data team cluster their datasets better, turning multiple-TB queries into just some GBs :)

I will try and update the inline SDK docs with more details

PR #61469, opened 2024-03-28
PLG: Add "no subscription" alert

vdavid:
- Part of https://github.com/sourcegraph/self-serve-cody/issues/506

I had to add an unexpected CSS rule to get the alert to display correctly because of its parent having `display: flex`.

Unrelated: fixed a few warnings around aria labels to improve accessibility.

## Test plan

Ran locally. This is how it looks: 
CleanShot 2024-03-28 at 15 03 36@2x

### DIFF
--- a/client/web/src/cody/subscription/CodySubscriptionPage.tsx
+++ b/client/web/src/cody/subscription/CodySubscriptionPage.tsx
@@ -59,7 +59,9 @@ export const CodySubscriptionPage: React.FunctionComponent<CodySubscriptionPageP
     const parameters = useSearchParameters()
 
     const utm_source = parameters.get('utm_source')
-    const redirectedForNoSubscription = useMemo(() => parameters.get('noSubscription') === '1', [parameters])
+
+    // Used when redirected from the SSC site for not having a subscription.
+    const showNoSubscriptionMessage = useMemo(() => parameters.get('noSubscription') === '1', [parameters])
 
     const codyPaymentsUrl = useCodyPaymentsUrl()
     const manageSubscriptionRedirectURL = `${codyPaymentsUrl}/cody/subscription`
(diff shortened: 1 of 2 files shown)


client/web/src/cody/subscription/CodySubscriptionPage.tsx
```diff
@@ -59,6 +59,7 @@ export const CodySubscriptionPage: React.FunctionComponent<CodySubscriptionPageP
     const parameters = useSearchParameters()
 
     const utm_source = parameters.get('utm_source')
+    const redirectedForNoSubscription = useMemo(() => parameters.get('noSubscription') === '1', [parameters])
```
chrsmith, line 62: There isn't actually any redirection going on though, right? It's just "if we see this URL query parameter, we display a notification?

So perhaps `showNoSubscriptionMessage` would be a more appropriate or less confusing name?

client/web/src/cody/subscription/CodySubscriptionPage.tsx
```diff
@@ -336,7 +340,7 @@ export const CodySubscriptionPage: React.FunctionComponent<CodySubscriptionPageP
                             </div>
                         </div>
                     </div>
-                    <div className="border d-flex flex-column flex-1 bg-1 border p-3 rounded">
```
chrsmith, line 339: Just to confirm, removing this `border` class is intentional, right?

chrsmith (approved):
Some minor nits, but seems reasonable.

client/web/src/cody/subscription/CodySubscriptionPage.tsx
vdavid: Yes, it was a duplicate.

client/web/src/cody/subscription/CodySubscriptionPage.tsx
vdavid: Renamed and added a comment to describe the use case in https://github.com/sourcegraph/sourcegraph/pull/61469/commits/41779afd967ca6d9a4553b6f5c7ed2fb2266a79d.

Issue #60679, opened 2024-02-21
Redirect users from `/get-cody` to `/cody`


vdavid:

iskyOS:
Just adding a screenshot for detail. `/get-cody` is currently missing the login/signup buttons, so for users that land on it, they are basically stranded.

`/get-cody` also still references the desktop app lower on the page, which is deprecated.

![CleanShot 2024-02-21 at 16 45 55](https://github.com/sourcegraph/sourcegraph/assets/13739780/7c9b6547-e89b-46ba-932d-4e868cd4523c)

chenkc805:
@iskyOS 2 questions:
- how do users even reach this page today?
- how many people hit this page over the past 2 weeks?

iskyOS:
@chenkc805 A couple notes on this issue:

- This issue is only problematic for unauthed users who land on this page.
- Those auth buttons have been fixed now, so this page isn't a dead end for unauthed users anymore. BUT the page still talks about the Cody desktop app, which is gone.

> @iskyOS 2 questions:
> 
> * how do users even reach this page today?
The authentication buttons on sourcegraph.com actually route people through this URL when they're signing up (at least that's my understanding). For example, this is the link one of those auth buttons uses: 'https://sourcegraph.com/.auth/openidconnect/login?prompt_auth=github&pc=sams&&redirect=/get-cody'

The issue is for random and unauthed users, who shouldn't commonly end up on this page. However there are a few old blog posts of ours that point to it, and and I believe it's mentioned in some of our older youtube videos + 3rd party videos.

> * how many people hit this page over the past 2 weeks?
Slightly over 10,000 users in the past 2 weeks, but we don't know how many of those were authenticated vs unauthed.

PR #61470, opened 2024-03-28
PLG: Redirect user from /get-cody if unauthed

vdavid:
- Closes https://github.com/sourcegraph/sourcegraph/issues/60679

Just a very simple two-line change.

## Test plan

Tested authed and un-authed with and without the change: I could repro the old behavior and saw that this change fixes it. See it work in this 1-min Loom: https://www.loom.com/share/056ebfe326244c18966e1500e735fa49


chrsmith (approved):
🚀

PR #58484, opened 2023-11-22
Remove centralized observability deps

daxmc99:
These are now unused

## Test plan

Dep update, CI

PR #61471, opened 2024-03-28
bazel: add prometheus-gcp image

burmudar:
EDIT

In light of https://github.com/sourcegraph/sourcegraph/pull/58484 seems this was purposely removed

## Test plan
https://buildkite.com/sourcegraph/sourcegraph/builds/266763

### DIFF
--- a/docker-images/prometheus/BUILD.bazel
+++ b/docker-images/prometheus/BUILD.bazel
@@ -60,3 +60,35 @@ oci_push(
     image = ":image",
     repository = image_repository("prometheus"),
 )
+
+oci_image(
+    name = "image_gcp",
+    base = "@wolfi_prometheus_gcp_base",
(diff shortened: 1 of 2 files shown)


docker-images/prometheus/BUILD.bazel
```diff
@@ -60,3 +60,35 @@ oci_push(
...
+    driver = "docker",
+    image = ":image_gcp",
+    tags = [
+        "exclusive",
+        "requires-network",
+    ],
+)
+
+oci_push(
+    name = "gcp_candidate_push",
```
Strum355, line 91: why "gcp" instead of "prometheus"? 🤔

burmudar: we already have one for `prometheus` this is just for gcp using a different wolfi base

PR #61474, opened 2024-03-28
k8s/resource: Add pkg for creating k8s configmap resources

jdpleiness:
Add `configmap` pkg to create k8s configmaps with patterns and defaults common to Sourcegraph.

## Test plan

### DIFF
--- a/internal/k8s/resource/configmap/BUILD.bazel
+++ b/internal/k8s/resource/configmap/BUILD.bazel
@@ -1,0 +1,31 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library")
+load("//dev:go_defs.bzl", "go_test")
+
+go_library(
+    name = "configmap",
+    srcs = ["configmap.go"],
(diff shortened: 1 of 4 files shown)


internal/k8s/resource/configmap/BUILD.bazel
```diff
@@ -0,0 +1,31 @@
...
+        "@io_k8s_api//core/v1:core",
+        "@io_k8s_apimachinery//pkg/apis/meta/v1:meta",
+    ],
+)
+
+go_test(
+    name = "configmap_test",
+    srcs = [
+        "configmap_test.go",
+        "example_test.go",
```
DaedalusG, line 21: What is example_test.go actually getting called for? Just to see if the config map default can be marshaled to json and yaml?

DaedalusG (approved):
This looks good pretty straightforward following the same pattern as operator repo -- see the comment in slack before merging I think though?
https://sourcegraph.slack.com/archives/C05EH3JP15Z/p1711755583228859

PR #61476, opened 2024-03-28
Adding changes to redis caching to be able to support redis store for durability

arafatkatze:
Originally suggested by Stephen [here](https://github.com/sourcegraph/sourcegraph/pull/61134#issuecomment-2000758669)
image
Adding the ability to support a new type of redis in the redis cache so that more durable connections can be made. 

This [redis uses AOF ](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@redis-store-add/-/blob/cmd/server/shared/assets/redis-store.conf.tmpl?L12)which makes it very resilient to the redis going down
image

This is a prelude to this pr https://github.com/sourcegraph/sourcegraph/pull/61134 

## Test plan
Make sure the tests pass


internal/rcache/rcache.go
```diff
@@ -31,12 +31,22 @@ func DeleteOldCacheData(c redis.Conn) error {
...
 // New creates a redis backed Cache
 func New(keyPrefix string) *Cache {
 	return &Cache{
 		keyPrefix: keyPrefix,
+		useStore:  false,
+	}
+}
+
+// New creates a redis backed Cache
+func NewWithRedisStore(keyPrefix string) *Cache {
```
RXminuS, line 46: It feels like this will very quickly lead to a lot of constructors? Maybe better to introduce a config struct here so store & ttl can be configured optionally?

Also just in terms of readability it's not clear to me that if useStore is `false` it now would use the `cache` instance. As it's not a "on/off" feature but more a "type selector" maybe an enum like`storeType` would be more readable?

arafatkatze: > It feels like this will very quickly lead to a lot of constructors?

The chances of that are slim to none. In actuality the only reason we are adding this constructor is because We want to use a more durable redis for caching.

Although generally across sourcegraph the idea is that we will generally use the caching redis which is less durable so for just about anything else we can use the old constructor. Its only for our usecase of having to use a more durable redis that tolerates failure(while preserving keys) that we have to use this.

For most other cases that redis is not needed. 

> Also just in terms of readability it's not clear to me that if useStore is false it now would use the cache instance. That could just be my context, but as it's not a "on/off" feature but more a "type selector" maybe an enum likestoreType would be more readable? I am a die-hard rust-acean though so we use Enum's for everything 😁

Makes total sense and I made the change.

RXminuS (approved):

PR #61478, opened 2024-03-28
Add syntax highlighting for Pkl files

mmanela:
Fixes PRIV-2879

Adds syntax highlighting for the [Pkl configuration](https://pkl-lang.org/) language.  

This does NOT add pkl files to the `lang:` filter. 

**Key Changes/Notes/Questions**
1. Added reference to the Pkl tree-sitter package
2. Added references to the new language in the syntax highlighter rust project
3. Added the Pkl tree sitter queries (.scm files) (got them from the [pkl tree sitter repo](https://github.com/apple/tree-sitter-pkl/blob/main/queries/highlights.scm))
5. Add reference to `pkl` files in `gosyntect/langauges.go` and `highlight/languages.go` to ensure it is picked up properly since pkl does not exist in go-enry.
6. Generated unit test snapshot for Pkl

 
image

## What about updating the Docs?
If we agree these are the minimal steps to add a highlighter, I can update the docs with these steps. The one wrinkle is handling if the given language exists in 
- Syntect
- Enry

If they do or don't exist in those places some of the steps may change a tiny bit I think. 

## Test plan
- [x] Update unit tests
- [x] Validate local instance colors pkl correctly
- [ ] validate in S2 after deploy


docker-images/syntax-highlighter/Cargo.Bazel.lock
```diff
@@ -45,7 +45,7 @@
       "version": "1.0.2",
       "repository": {
         "Http": {
-          "url": "https://crates.io/api/v1/crates/adler/1.0.2/download",
+          "url": "https://static.crates.io/crates/adler/1.0.2/download",
```
eseliger, line 48: I have no clue about rust, but I wonder how this change comes to be. Did they just change their domains at some point?

eseliger (commented):
I don't feel confident to approve this, but I don't see anything that screams wrong at me, and it works so.. 
I'll defer for final approval to the Graph team :)

Thanks for tagging me for the learning experience, and congrats on shipping a customer request in <48h after we heard about it on a call!

docker-images/syntax-highlighter/crates/syntax-analysis/queries/pkl/locals.scm
```diff
@@ -0,0 +1,38 @@
...
+(objectProperty
+  (identifier) @local.definition)
+
+(typedIdentifier
+  (identifier) @local.definition)
+
+; References
+
+(variableExpr
+  (identifier) @local.reference)
```
varungandhi-src, line 38: I'm pretty sure that Pkl files can reference entities in other files, so these are not correct. Since this PR is for highlighting, please remove the locals navigation queries.

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__pkl.pkl.snap
```diff
@@ -0,0 +1,97 @@
+---
+source: crates/syntax-analysis/src/highlighting/tree_sitter.rs
+expression: "snapshot_treesitter_syntax_kinds(&document, &contents)"
+---
+  import "pigeon1.pkl"
+//^^^^^^ Keyword
+//       ^^^^^^^^^^^^^ StringLiteral
```
varungandhi-src, line 7: `pigeon1.pkl` should get marked as `IdentifierNamespace`, not `StringLiteral`.

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__pkl.pkl.snap
```diff
@@ -0,0 +1,97 @@
...
+  amends "pigeon2.pkl"
+//^^^^^^ Identifier
```
varungandhi-src, line 9: `amends` should be marked as a `Keyword`, not as `Identifier`.

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__pkl.pkl.snap
```diff
@@ -0,0 +1,97 @@
...
+//        ^^^^^^^ Identifier
+//               ^ PunctuationDelimiter
+//                ^^^ Identifier
```
varungandhi-src, line 12: These are incorrect, `extends` and `amends` have syntax similar to `import`. Please add a test for `extends` too. https://github.com/apple/tree-sitter-pkl/blob/main/grammar.js#L128-L131

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__pkl.pkl.snap
```diff
@@ -0,0 +1,97 @@
...
+//       ^ IdentifierOperator
+//          ^^^^^^ Identifier
+    name = "Great green macaw"
+//  ^^^^ Identifier
+//       ^ IdentifierOperator
+//         ^^^^^^^^^^^^^^^^^^^ StringLiteral
+    diet = "Berries"
+//  ^^^^ Identifier
+//       ^ IdentifierOperator
+//         ^^^^^^^^^ StringLiteral
```
varungandhi-src, line 30: This line doesn't add anything over the previous one, please remove it.

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__pkl.pkl.snap
```diff
@@ -0,0 +1,97 @@
...
+  import "pigeon1.pkl"
+//^^^^^^ Keyword
+//       ^^^^^^^^^^^^^ StringLiteral
+  amends "pigeon2.pkl"
+//^^^^^^ Identifier
+//        ^^^^^^^ Identifier
+//               ^ PunctuationDelimiter
+//                ^^^ Identifier
+  
+  name = "Great green macaw"
```
varungandhi-src, line 14: Please add tests for the following inside strings: https://pkl-lang.org/main/current/language-reference/index.html#strings

1. Escape sequences, especially `\"`, and `\\` and `\u{<codepoint>}`
2. String interpolation
3. Multiline strings, with and without interpolation, with and without leading whitespace.
4. Strings with custom delimiters
5. Multiline strings with custom delimiters (not super obvious that this is supported, but seems plausible)

docker-images/syntax-highlighter/crates/syntax-analysis/queries/pkl/highlights.scm
```diff
@@ -0,0 +1,138 @@
...
+
+(stringConstant) @string
+(slStringLiteral) @string
+(mlStringLiteral) @string
+
+(escapeSequence) @escape
+
+(intLiteral) @number
+(floatLiteral) @number
+
```
varungandhi-src, line 42: Please add tests for [duration literals](https://pkl-lang.org/main/current/language-reference/index.html#durations) and [data sizes](https://pkl-lang.org/main/current/language-reference/index.html#data-sizes). It's weird that these are not recorded in the grammar, maybe that needs to be fixed upstream first.

https://pkl-lang.org/main/current/language-reference/index.html#durations

I think maybe we should mark the suffixes (`min` in the context `5.min`) as keywords.

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__pkl.pkl.snap
```diff
@@ -0,0 +1,97 @@
...
+    taxonomy {
+//  ^^^^^^^^ Identifier
+      species = "Ara ambiguus"
+//    ^^^^^^^ Identifier
+//            ^ IdentifierOperator
+//              ^^^^^^^^^^^^^^ StringLiteral
+    }
+  }
+  
+  woodPigeon {
```
varungandhi-src, line 40: Please remove the code here and below this, since it's not testing anything new.

varungandhi-src (changes requested):
Thank you for the PR. I appreciate your enthusiasm to get your hands dirty and help out.

I noticed that there are lots of test cases missing. During the first pass, we should be aiming for coverage of as many of the various syntax features as possible, so that we don't have to keep coming back to fix small things (it's also quite likely that if some things work and other don't, users won't like the UX but they won't dislike it enough to complain about it).

I've outlined some in the inline comments, some more that are missing (non-exhaustive):
- Type annotations (simple identifiers as well as generic types)
- Escaped identifiers such as `` `class` ``.
- Chained amends expression
- Field accesses
- Class definitions
- Method definitions
- Method calls
- `new` expressions
- Various property keywords such as `fixed`, `hidden`, `local`, `const` etc.
- Annotations such as `@Deprecated`

You can go through the language reference (https://pkl-lang.org/main/current/language-reference/index.html) and add test cases for these.

If it's not possible to pick out some syntax features based on the grammar, we should still add a test case for that and note in a comment that it is because of an upstream limitation.

I do not feel comfortable merging this in its current state given the low coverage. Please re-request code review once you've made the suggested changes.

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting.rs
```diff
@@ -167,11 +168,12 @@ struct SublimeLanguageName {
...
-            match file_info.extension() {
-                Some("ncl") => return TreeSitterLanguageName::new("nickel"),
-                _ => {}
-            };
+            let extension = file_info.extension().unwrap_or("");
+            if !extension.is_empty() {
+                if let Some(parser_id) = ParserId::from_file_extension(extension) {
+                    return TreeSitterLanguageName::new(parser_id.name());
+                }
+            }
```
varungandhi-src, line 176: The layering here is wrong - the caller (from Go) should be passing the language `"Pkl"`, we should try to avoid doing more language detection here. We already run language detection in Go as pass a language as input to the caller here. This old code is basically a workaround and should be removed eventually or be made consistent with the rest instead of this special hack.

The correct fix would be to modify `func getLanguagesByExtension` in Go and add an extra `map[string]string unsupportedByEnryExtensionsMap` until Linguist and consequently go-enry get support for Pkl.

mmanela: @varungandhi-src I first made this change while struggling to understand why the unit test (test all files) was failing . That test path relies on either the file under test matching in syntect or the hard code work around that was here for nickel already based on extension.

How do you suggest I re-work that test?

mmanela:
@varungandhi-src Thanks for the thorough review ( I was eager for you to show me the ropes here ).


Couple questions /comments
1. By add unit tests in this case you mean make my pkl test file cover all the key language features you layout right?

2. The scm file I copied from the tree sitter pkl repo, I just assumed it was correct so thanks for poking at that. Good opportunity to understand more there and make sure that file is correct.

3. I will make a second pass today and make as many changes as I can.

docker-images/syntax-highlighter/crates/syntax-analysis/queries/pkl/locals.scm
mmanela: 1. So remove this whole file right?

2. For my own education can you explain why this jumped out to you as wrong ( it's my first time using or reading tree sitter queries )?

varungandhi-src: > So remove this whole file right?

I'm not 100% sure if we leave these blank for other languages where we don't support locals navigation or not have these files at all, but I think removing seems reasonable.

> For my own education can you explain why this jumped out to you as wrong ( it's my first time using or reading tree sitter queries )?

It doesn't have anything to do with Tree-sitter queries specifically but the purpose of this file (it's called `locals.scm`), which is to mark local variables for faster code navigation without extra queries from the client to the server. We're not implementing code navigation functionality here, so it's out-of-place for this PR.

If your question was about language semantics, I saw 'classes' and 'methods'; it would be unusual for a language to have such abstraction facilities if they were confined to a single file.

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting.rs
varungandhi-src: The test code can be changed to pass in the language for the Pkl case. It might sound a bit weird (why would you want to spread this logic in more places instead of keeping it confined here?), but we want to ideally remove the language detection logic in the highlighter, and mimic the situation where the language is passed in from the outside (which is what should be happening in production), so it makes sense for the test to mimic what the Go code would do in production.

I don't think it's a HUGE deal, it's more work than what you've done here, but if there is code that starts accidentally relying on this codepath for something other than Pkl (because now you're suddenly doing language detection for many cases!), then it will be harder to undo.

varungandhi-src:
> By add unit tests in this case you mean make my pkl test file cover all the key language features you layout right?

I skimmed through the language reference (https://pkl-lang.org/main/current/language-reference/index.html) and noted things which looked missing. I did not look at everything in the language reference. When you're adding test cases, you should add test cases for the different syntactic features as written in the language reference (my list is a subset).

Generally, it is also good to add test cases for "interesting" combinations of features, e.g. how I wrote "Multiline strings, with and without interpolation, with and without leading whitespace." What exactly is an "interesting" combination is hard to define, but it's useful to think through variations. You may also want to look at the grammar.js file directly for inspiration, or the test cases in the tree-sitter-pkl repo. (https://github.com/apple/tree-sitter-pkl/tree/main/corpus)

> The scm file I copied from the tree sitter pkl repo, I just assumed it was correct so thanks for poking at that. Good opportunity to understand more there and make sure that file is correct.

1. Please add a link and attribution at the top of the file as the original is Apache 2.0 licensed.
2. That's why we have the snapshot tests; you see there is this line https://github.com/apple/tree-sitter-pkl/blob/main/queries/highlights.scm#L101 but in your snapshot output, `amends` was getting marked as an identifier. I think this might be because the order matters -- generally the identifier queries should come at the end, and keywords and operators should come at the beginning. So the upstream queries are buggy (and if you look at the language reference, that also has some incorrect highlighting for code snippets that doesn't match the intent of the queries).

    If you're feeling so inclined, you could submit an issue or PR upstream too and point them to your improved queries.

docker-images/syntax-highlighter/crates/syntax-analysis/queries/pkl/highlights.scm
mmanela: I added them to the test file but I am not sure how to mark the duration and unit suffices as keywords. I looked into the grammer.js but I see what you meant that they don't explicitly seem to label those

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__pkl.pkl.snap
mmanela: I tried a while to figure out how to do this but nothing I tried worked. I probably am missing knowledge here. If this is critical, we can chat later and just handle unresolved feedback in one session. 

I also notice other languages like Typescript also list things like this as StringLiteral
image

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__pkl.pkl.snap
mmanela: Done

mmanela:
@varungandhi-src  Thanks for your detailed information. I made a second attempt to address as much of your feedback as I could figure out.

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__pkl.pkl.snap
varungandhi-src: Thanks for bringing up TypeScript, I didn't realize we were doing that, this means we have a code nav bug with TypeScript. https://github.com/sourcegraph/sourcegraph/issues/61508

FWIW, the "right" fix for this would be to add a query like:

```
(importClause "import" (stringConstant) @identifier.module)
```

However, I don't think we should do that, because from a highlighting POV, you probably want this to be highlighted like a string literal anyways. I suggested this earlier as a workaround for the incorrect assumption in the client-side code for interactive occurrence filtering (https://sourcegraph.slack.com/archives/C05EA9KQUTA/p1711010206641949?thread_ts=1710936490.391979&cid=C05EA9KQUTA, see `const INTERACTIVE_OCCURRENCE_KINDS` in the client code), but in hindsight, I don't think we should make the highlighting wrong to work around bugs.

Let's leave this as StringLiteral.

docker-images/syntax-highlighter/crates/syntax-analysis/queries/pkl/highlights.scm
varungandhi-src: > I added them to the test file but I am not sure how to mark the duration and unit suffices as keywords. I looked into the grammer.js but I see what you meant that they don't explicitly seem to label those

One good way to check the tree form is to check out the grammar repo and modify one of the corpus files, and run `tree-sitter test --update` using the Tree-sitter CLI. Here is a diff I got:

```
diff --git a/corpus/number/numberUnderscores.txt b/corpus/number/numberUnderscores.txt
index 16c98e9..c60f793 100644
--- a/corpus/number/numberUnderscores.txt
+++ b/corpus/number/numberUnderscores.txt
@@ -26,2 +26,4 @@ res21 = 0o755
 res22 = 0o7_5_5__
+res23 = 5.min
+res24 = true.xor(false)
 
@@ -98,2 +100,14 @@ res22 = 0o7_5_5__
     (identifier)
-    (intLiteral)))
+    (intLiteral))
+  (classProperty
+    (identifier)
+    (propertyCallExpr
+      (intLiteral)
+      (identifier)))
+  (classProperty
+    (identifier)
+    (methodCallExpr
+      (trueLiteral)
+      (identifier)
+      (argumentList
+        (falseLiteral)))))
```

Based on that, you could make this patch to handle the pure literal cases:

```
diff --git a/docker-images/syntax-highlighter/crates/syntax-analysis/queries/pkl/highlights.scm b/docker-images/syntax-highlighter/crates/syntax-analysis/queries/pkl/highlights.scm
index 4ac335590ae..f3db64ed547 100644
--- a/docker-images/syntax-highlighter/crates/syntax-analysis/queries/pkl/highlights.scm
+++ b/docker-images/syntax-highlighter/crates/syntax-analysis/queries/pkl/highlights.scm
@@ -126,2 +126,6 @@
 
+; Units of measure
+(propertyCallExpr (intLiteral) (identifier) @identifier.type)
+(propertyCallExpr (floatLiteral) (identifier) @identifier.type)
+
 (parameterList (typedIdentifier (identifier) @variable.parameter))
diff --git a/docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__pkl.pkl.snap b/docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__pkl.pkl.snap
index 06533df8eba..8aff890a0d8 100644
--- a/docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__pkl.pkl.snap
+++ b/docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__pkl.pkl.snap
@@ -221,3 +221,3 @@ expression: "snapshot_treesitter_syntax_kinds(&document, &contents)"
 //             ^ PunctuationDelimiter
-//              ^^^ Identifier
+//              ^^^ IdentifierType
   xySeconds = (x + y).s
@@ -239,3 +239,3 @@ expression: "snapshot_treesitter_syntax_kinds(&document, &contents)"
 //             ^ PunctuationDelimiter
-//              ^ Identifier
+//              ^ IdentifierType
 //                 ^^^^^^^^^^^^^^^^^^^^^^^^ Comment
@@ -607 +607,3 @@ expression: "snapshot_treesitter_syntax_kinds(&document, &contents)"
 //            ^^^^^^ StringLiteral
+  
+
```

However, because this is also supported for more complex expressions, when the grammar sees `foo.s`, it is not clear whether it is OK to unambiguously interpret that as marking it as a duration in seconds, or whether if the `foo` object actually has an `s` field, then this does a field projection instead of a wrapping operation. Since we can only mark durations in the literal cases, it would lead to inconsistent highlighting in other cases.

I've filed a bug against the tree-sitter-pkl repo for this: https://github.com/apple/tree-sitter-pkl/issues/6

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/tree_sitter.rs
```diff
@@ -431,6 +432,32 @@ mod test {
...
+            || language_from_syntect.raw.to_lowercase() == "plain text"
+        {
+            let extension = filepath.extension().unwrap_or_default();
+            if !extension.is_empty() {
+                if let Some(parser_id) = ParserId::from_file_extension(extension.to_str().unwrap())
+                {
+                    return TreeSitterLanguageName::new(parser_id.name());
+                }
+            }
+        }
```
varungandhi-src, line 456:
```suggestion
        if_chain! {
            if language_from_syntect.raw.is_empty()
                || language_from_syntect.raw.to_lowercase() == "plain text";
            if let Some(extension) = filepath.extension();
            if let Some(parser_id) = ParserId::from_file_extension(extension.to_str().unwrap());
            then {
                return TreeSitterLanguageName::new(parser_id.name());
            }
        }
```

After adding `use if_chain::if_chain` (you can see other examples in the code), since this feature isn't stabilized yet. https://github.com/rust-lang/rust/issues/53667

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/tree_sitter.rs
```diff
@@ -484,12 +511,7 @@ SELECT * FROM my_table
             let mut contents = String::new();
             file.read_to_string(&mut contents)?;
 
-            let language = &crate::highlighting::test::SYNTAX_SET
-                .with(|syntax_set| {
-                    FileInfo::new(filepath.to_string_lossy().as_ref(), &contents, None)
-                        .determine_language(syntax_set)
-                })
-                .unwrap();
+            let language = get_language_for_test(&filepath, &contents);
```
varungandhi-src, line 514: There is another code path doing a similar dance in `fn test_files_with_locals`, please update that too to avoid divergence.

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting.rs
varungandhi-src: Please follow up on this comment before marking this comment thread as resolved.

> The correct fix would be to modify `func getLanguagesByExtension` in Go and add an extra `map[string]string` `unsupportedByEnryExtensionsMap` until Linguist and consequently go-enry get support for Pkl.

At the moment, you've modified the `baseHighlightConfig`, but that's the wrong place to modify because we want to make sure `.pkl` is detected as the Pkl language everywhere else too, not just for highlighting.

I strongly suggest not touching the Nickel-related code in this PR instead of removing it as it's not related to Pkl.

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__pkl.pkl.snap
```diff
@@ -0,0 +1,607 @@
...
+  x = 42
+//^ Identifier
+//  ^ IdentifierOperator
+//    ^^ NumericLiteral
+  message1 = "\(x + 2) plus \(x * 2) is \(0x80)"
+//^^^^^^^^ Identifier
+//         ^ IdentifierOperator
+//           ^ StringLiteral
+//              ^ Identifier
+//               ^ StringLiteral
```
varungandhi-src, line 62: This is not correct, `+` shouldn't be getting marked as a `StringLiteral`.

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__pkl.pkl.snap
```diff
@@ -0,0 +1,607 @@
...
+//^^^^^ Identifier
+//      ^ IdentifierOperator
+//        ^^^^^^ StringLiteral
+//              ^ PunctuationDelimiter
+//               ^^^^^^^^ IdentifierFunction
+//                        ^^^^^^^ StringLiteral
+  
+  
+  # Functions
+//  ^^^^^^^^^ IdentifierType
```
varungandhi-src, line 122: `#` is not [valid comment syntax](https://pkl-lang.org/main/current/language-reference/index.html#comments), please remove that here and below.

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__pkl.pkl.snap
```diff
@@ -0,0 +1,607 @@
...
+//   ^ IdentifierOperator
+//     ^^^ NumericLiteral
+  notANumber = NaN
+//^^^^^^^^^^ Identifier
+//           ^ IdentifierOperator
+//             ^^^ IdentifierType
+  positiveInfinity = Infinity
+//^^^^^^^^^^^^^^^^ Identifier
+//                 ^ IdentifierOperator
+//                   ^^^^^^^^ IdentifierType
```
varungandhi-src, line 196: NaN and Infinity should be marked as builtins rather than as types.

docker-images/syntax-highlighter/crates/syntax-analysis/queries/pkl/highlights.scm
```diff
@@ -0,0 +1,159 @@
...
+"abstract" @keyword
+"amends" @include
+"as" @keyword
+"class" @keyword
+"const" @keyword
+"else" @keyword
+"extends" @keyword
+"external" @keyword
+(falseLiteral) @constant.builtin
```
varungandhi-src, line 67:
```suggestion
(falseLiteral) @boolean
```

docker-images/syntax-highlighter/crates/syntax-analysis/queries/pkl/highlights.scm
```diff
@@ -0,0 +1,159 @@
...
+"out" @keyword
+(outerExpr) @variable.builtin
+"read" @function.method.builtin
+"read?" @function.method.builtin
+"read*" @function.method.builtin
+"super" @variable.builtin
+(thisExpr) @variable.builtin
+"throw" @function.method.builtin
+"trace" @function.method.builtin
+(trueLiteral) @constant.builtin
```
varungandhi-src, line 96:
```suggestion
(trueLiteral) @boolean
```

docker-images/syntax-highlighter/crates/syntax-analysis/queries/pkl/highlights.scm
```diff
@@ -0,0 +1,159 @@
...
+"import*" @include
+"in" @keyword
+"is" @keyword
+"let" @keyword
+"local" @keyword
+(moduleExpr "module" @include)
+"module" @keyword
+"new" @keyword
+"nothing" @type.builtin
+(nullLiteral) @constant.builtin
```
varungandhi-src, line 85:
```suggestion
(nullLiteral) @constant.null
```

docker-images/syntax-highlighter/crates/syntax-analysis/queries/pkl/highlights.scm
```diff
@@ -0,0 +1,159 @@
...
+"as" @keyword
+"class" @keyword
+"const" @keyword
+"else" @keyword
+"extends" @keyword
+"external" @keyword
+(falseLiteral) @constant.builtin
+"fixed" @keyword
+"for" @keyword
+"function" @keyword
```
varungandhi-src, line 70:
```suggestion
"function" @keyword.function
```

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__pkl.pkl.snap
```diff
@@ -0,0 +1,607 @@
...
+//^^^^^^^^ Identifier
+    ["Pigeon"] = "Piggy the Pigeon"
+//   ^^^^^^^^ StringLiteral
+//             ^ IdentifierOperator
+//               ^^^^^^^^^^^^^^^^^^ StringLiteral
+  }
+  
+  entries2 {
+//^^^^^^^^ Identifier
+    ...entries1
```
varungandhi-src, line 542: Missing occurrence for the `...`?

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__pkl.pkl.snap
```diff
@@ -0,0 +1,607 @@
...
+//               ^^^ Keyword
+    name = "Pigeon"
+//  ^^^^ Identifier
+//       ^ IdentifierOperator
+//         ^^^^^^^^ StringLiteral
+  }
+  
+  @Deprecated { message = "Replace with \(pigeonName)" }
+//^ IdentifierOperator
+// ^^^^^^^^^^ IdentifierType
```
varungandhi-src, line 389: Should ideally be IdentifierAttribute for the full range, instead of `IdentifierOperator` and `IdentifierType`

varungandhi-src (commented):
Looking much better, will do a further pass later.

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting.rs
mmanela: Thanks, I missed that, will revise .

docker-images/syntax-highlighter/crates/syntax-analysis/queries/pkl/highlights.scm
mmanela: And as you mentioned, I left this as is for now

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__pkl.pkl.snap
mmanela: Ok will leave as is , thanks for the info on how to go about it. I understand tree-sitter queries a lot more now

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting.rs
mmanela: Updated, thanks for catching.

docker-images/syntax-highlighter/Cargo.Bazel.lock
varungandhi-src: It's a recent change. https://blog.rust-lang.org/2024/03/11/crates-io-download-changes.html

docker-images/syntax-highlighter/crates/syntax-analysis/queries/pkl/highlights.scm
```diff
@@ -0,0 +1,175 @@
...
+
+; Identifiers
+(objectSpread ("..." @identifier.operator) (variableExpr))
+
+(classProperty (identifier) @property)
+(classProperty (identifier) @property)
+(objectProperty (identifier) @property)
+
+(parameterList (typedIdentifier (identifier) @variable.parameter))
+(objectBodyParameters (typedIdentifier (identifier) @variable.parameter))
```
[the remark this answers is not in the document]
mmanela: I added the following example which does parse as ObjectBodyParamters
```
// dynamic object with object body parameters
birds = new Dynamic {
  default { _ ->
    species = "Bird"
  }
  ["Falcon"] {}
}

```


```
module [0, 0] - [6, 0]
  (classProperty [0, 0] - [5, 1]
    (identifier [0, 0] - [0, 5])
    (newExpr [0, 8] - [5, 1]
      (type [0, 12] - [0, 19]
        (qualifiedIdentifier [0, 12] - [0, 19]
          (identifier [0, 12] - [0, 19])))
      (objectBody [0, 20] - [5, 1]
        (objectProperty [1, 2] - [3, 3]
          (identifier [1, 2] - [1, 9])
          (objectBody [1, 10] - [3, 3]
            (objectBodyParameters [1, 12] - [1, 16]
              (typedIdentifier [1, 12] - [1, 13]
                (identifier [1, 12] - [1, 13])))
            (objectProperty [2, 4] - [2, 20]
              (identifier [2, 4] - [2, 11])
              (slStringLiteral [2, 14] - [2, 20]))))
        (objectEntry [4, 2] - [4, 15]
          (slStringLiteral [4, 3] - [4, 11])
          (objectBody [4, 13] - [4, 15]))))))
          ```

PR #61479, opened 2024-03-28
k8s/resource: Add pkg for creating k8s serviceaccount resources

jdpleiness:
Add `serviceaccount` pkg to create k8s service accounts with patterns and defaults common to Sourcegraph.

## Test plan


Chickensoupwithrice (approved):
on 🔥 let's go

PR #61481, opened 2024-03-28
Add note for supplying context when changing per-user overrides

chrsmith:
Sourcegraph instances have the ability to set a per-user "quota override" which is used for Cody PLG, as a knob for governing access to the system.

However, this has been kinda painful to work with in-practice, because there isn't any context about when or why a change was made. (e.g. Why was this user's quota set higher or lower? Who made the change? How long should it be in effect?)

This PR does (I think) all of the wiring to just add a new `users.completions_quota_note` column to the `users` table. And the associated read/write actions to expose it from our GraphQL schema. And finally adding the new field to the User Quota page... all so that 🤞  we can just see a note displayed like `"User quota set by <tool> for <reason>. (<date>)"`.

⚠️ I have no idea what I'm doing, and wasn't able to successfully run Sg locally. So I kinda sorta need some help 👉👈😢.

> I'll bug people on Slack, but things like `sg generate` fail before updating `schema.md`. And when running locally, the default configuration for GitHub OAuth doesn't seem to work. (So maybe it's a client-side configuration issue? Or I just need to fetch the secrets?)
>
> Anyways, in the interest of time, if you could just patch this locally and sanity check the User Quota page can read/write the quota note I would greatly appreciate it.

## Long-term Plan

This is essentially a stop-gap solution to make it easier to understand and diagnose quota-related issues. But this definitely isn't where we want to land long-term.

My thinking is that this quarter we'd make the following changes to Sourcegraph.com / Cody PLG:

- Move the source of truth for user subscription information to the SSC backend. Today, Cody Gateway fetches the allowed models and/or user limits from dotcom. And dotcom in-turn fetches the user's subscription information from the SSC backend. But we'll just cut dotcom out of that equation. And have Cody Gateway read subscription information (and the models/quotas that are allowed as a result) directly from the SSC backend.
- As part of that, we'll also migrate the "quota overrides" component of dotcom into the SSC backend. So at that point we'll be able to drop the `users.code_completions_quota` and GraphQL methods like `SetCodeCompletionsQuota(...)` from the codebase entirely. (And instead, this information will be managed by the SSC backend. Which is the right long-term solution, since this functionality is specific to Cody PLG. And doesn't really apply to a prototypical Sourcegraph instance.

Both of those things are planned for the current quarter by the PLG team. So hopefully just having a simple text field will buy us enough time to have a better story from the SSC backend.

e.g. a more robust support for recording quota changes. Or being able to more easily set quota overrides per-Cody Gateway feature. Rather than having mostly duplicate GraphQL endpoints for both "completions" (chats?) and "code-completions".

## Test plan

Added some tests. But I was not able to try it out locally. (See note above.)

### DIFF
--- a/client/web/src/user/settings/quota/UserQuotaProfilePage.tsx
+++ b/client/web/src/user/settings/quota/UserQuotaProfilePage.tsx
@@ -9,7 +9,12 @@ import { LoaderButton } from '../../../components/LoaderButton'
 import { PageTitle } from '../../../components/PageTitle'
 import type { Scalars } from '../../../graphql-operations'
 
-import { SET_USER_CODE_COMPLETIONS_QUOTA, SET_USER_COMPLETIONS_QUOTA, USER_REQUEST_QUOTAS } from './backend'
+import {
+    SET_USER_CODE_COMPLETIONS_QUOTA,
+    SET_USER_COMPLETIONS_QUOTA,
+    SET_USER_COMPLETIONS_QUOTA_NOTE,
+    USER_REQUEST_QUOTAS,
+} from './backend'
 
 interface Props {
     user: {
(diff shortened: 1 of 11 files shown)


chrsmith:
@bobheadxi has shown me the error of my ways, and that we should solve this problem another way. Closing this PR out.

PR #61486, opened 2024-03-28
Generalize prompt recording for all upstream LLM providers

chrsmith:
We had a `PromptRecorder` interface used within our Anthropic handlers to temporarily record flagged requests for troubleshooting spam and abuse. However, now that we offer more LLMs for Cody PLG, we should generalize that capability so that it is part of all of our "upstream handlers". (And not something specific to just Anthropic.)

The diff for this PR is kinda noisy, but really only three changes were made:

- Moving the `type PromptRecorder interface {` from `anthropic.go` to `flagging.go`.
- Adding `flaggedPromptRecorder PromptRecorder` as a parameter to `makeUpstreamHandler`.
- Opportunistic code cleanups. e.g. removing the TODO around making the `dotcomPromptRecorder` configurable.

## Test plan

Just shuffling some code around. The type checker will ensure things are good to go.


vdavid (approved):
Went through the code in detail and I think I got it all. Thanks for mentioning "The diff for this PR is kinda noisy" in the PR desc., it set the expectations right so I wasn't annoyed by the PR indeed being kinda noisy. 🙂 
(Maybe next time use separate commits for the different changes? It should save more time on the code review than it takes to split the commits.)

rafax (approved):

chrsmith:
Applied a couple of changes to address feedback:

- We do not record `chat_completions`, since that is just noise until we get the false-positive rate there under control. (I think with the updates to REDACTED this should be better.)
- I also recorded a little more data in the Redis key, adding feature and user ID, in addition to the trace ID.

PR #61488, opened 2024-03-29
msp/runtime: export contract and helpers for direct usage

bobheadxi:
Exports package `github.com/sourcegraph/sourcegraph/lib/managedservicesplatform/runtime/contract` to expose previously-internal env-configuration (the "MSP contract") loading, helpers, and configuration. This allows programs that aren't 100% in the MSP runtime ecosystem yet (hopefully reducing future barrier to entry), or are in particular scenarios where they might not want to use runtime ([e.g. MSP jobs which doesn't have proper runtime support yet](https://sourcegraph.slack.com/archives/C06062P5TS5/p1711663052720289)), to integrate with MSP-provisioned infrastructure like BigQuery, PostgreSQL, and even Sentry etc.

Simple example:

```go
e, _ := contract.ParseEnv([]string{"MSP=true"})

c := contract.New(logtest.Scoped(t), mockServiceMetadata{}, e)
assert.NotZero(t, c)
assert.True(t, c.MSP)

writer, _ := c.BigQuery.GetTableWriter(ctx, "my-table")
writer.Write(...)
```

## Test plan

There are no functionality changes, only code moving. But, ran some sanity checks:

```
sg run msp-example
sg run telemetry-gateway 
sg run pings
```


jac (approved):

unknwon (approved):

lib/managedservicesplatform/runtime/contract/contract.go
```diff
@@ -0,0 +1,131 @@
...
+
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+	"github.com/sourcegraph/sourcegraph/lib/pointers"
+)
+
+// Contract loads standardized MSP-provisioned (Managed Services Platform)
+// configuration.
+type Contract struct {
+	// MSP indicates if we are running in a live Managed Services Platform
+	// environment. In local development, this should generally be false.
```
chrsmith, line 46: Can you elaborate on "In local development, this should generally be false"? (Asking because I honestly want to know.)

Suppose I want to build an MSP-backed service, but also want it to be easy to run locally. And I have the world's greatest `docker-compose.yml` file for spinning up all the necessary infrastructure... what should I do in that case?

Is the idea that I would need to avoid using the `Contract` struct entirely? Or use some other shim/library to connect to my local Postgres, Redis instances in a way that satisfies things like `postgreSQLContract`?

e.g. what exactly is the line between where MSP ends and local infrastructure begins? And how can I cleanly separate those two?

lib/managedservicesplatform/runtime/contract/contract.go
```diff
@@ -0,0 +1,131 @@
...
+	Port int
+	// ExternalDNSName is the DNS name the service uses, if one is configured.
+	ExternalDNSName *string
+
+	// RedisEndpoint is the full connection string of a MSP Redis instance if
+	// provisioned, including any prerequisite authentication.
+	RedisEndpoint *string
+
+	// PostgreSQL has helpers and configuration for MSP PostgreSQL instances.
+	PostgreSQL postgreSQLContract
```
chrsmith, line 62: We are exporting the `Contract` and `Contract.PostgreSQL`. But `postgreSQLContract` isn't usable outside of this package. So isn't this essentially useless to callers? 

IIRC, this is caught by some linters for that reason. But maybe this is intentional?

lib/managedservicesplatform/runtime/contract/contract.go
```diff
@@ -0,0 +1,131 @@
...
+	Diagnostics diagnosticsContract
+
+	// internal configuration for MSP internals that are not exposed to service
+	// developers.
+	internal internalContract
+}
+
+type ServiceMetadataProvider interface {
+	Name() string
+	Version() string
```
chrsmith, line 79: Super-minor nit: Because I love me obnoxious doc comments, it would be useful to add a note explaining what the recommended format for `Name` and `Version` should be.

e.g. will name return the MSP service slug like `Sourcegraph-accounts`, or a human-friendly "Sourcegraph Accounts"? Similarly, where does `Version` come from? Is it a semver? Git SHA when built? Or "implementation dependent"?

lib/managedservicesplatform/runtime/contract/contract.go
```diff
@@ -0,0 +1,131 @@
...
+	internal := internalContract{
+		logger:  logger,
+		service: service,
+	}
+	isMSP := env.GetBool("MSP", "false", "indicates if we are running in a MSP environment")
+
+	return Contract{
+		MSP:             isMSP,
+		EnvironmentID:   env.Get("ENVIRONMENT_ID", "unknown", "MSP Service Environment ID"),
+		Port:            env.GetInt("PORT", "", "service port"),
```
chrsmith, line 102: JFYI: this is one of the thing that's kinda annoying when using MSP.

Here's a basic config struct I wrote for a new app:

```go
package config

import (
	"github.com/sourcegraph/sourcegraph/lib/managedservicesplatform/runtime"
)

type Config struct {
	SAMSClientCredentials struct {
		ClientID     string
		ClientSecret string
	}
}

// Load implements the MSP service.ConfigLoader contract.
func (c *Config) Load(env *runtime.Env) {
	c.SAMSClientCredentials.ClientID = env.Get("SAMS_CLIENT_ID", "", "SAMS Client ID")
	c.SAMSClientCredentials.ClientSecret = env.Get("SAMS_CLIENT_SECRET", "", "SAMS Client Secret")
}
```

You'd think then, that there are only two _required_ env vars. However, the MSP runtime errors out unless `PORT` is set. (Probably because it is getting the default value of "" and choking later, but maybe we should just randomly assign the port as needed in that case?)

Anyways, this isn't a big deal, but it's just a "hidden requirement" when using the MSP infrastructure. But it's easy enough to identify/fix, so don't worry about changing anything.

lib/managedservicesplatform/runtime/contract/contract.go
```diff
@@ -0,0 +1,131 @@
...
+func extractBearer(h http.Header) (string, error) {
+	var token string
+
+	if authHeader := h.Get("Authorization"); authHeader != "" {
+		typ := strings.SplitN(authHeader, " ", 2)
+		if len(typ) != 2 {
+			return "", errors.New("token type missing in Authorization header")
+		}
+		if strings.ToLower(typ[0]) != "bearer" {
+			return "", errors.Newf("invalid token type %s", typ[0])
```
chrsmith, line 124: Totally unimportant, overly pedantic nit-pick because I am a fundamentally broken human being:

Reading the MDN docs, I think "authorization scheme" is the right term to use here.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization#syntax

```suggestion
			return "", errors.Newf("invalid token scheme %s", typ[0])
```

lib/managedservicesplatform/runtime/contract/contract.go
```diff
@@ -0,0 +1,131 @@
...
+		Diagnostics: loadDiagnosticsContract(logger, env, defaultGCPProjectID, internal, isMSP),
+
+		internal: internal,
+	}
+}
+
+func extractBearer(h http.Header) (string, error) {
+	var token string
+
+	if authHeader := h.Get("Authorization"); authHeader != "" {
```
chrsmith, line 118: TIL: there is built-in support for header name canonicalization 🤯 .
https://pkg.go.dev/net/textproto#CanonicalMIMEHeaderKey
https://cs.opensource.google/go/go/+/refs/tags/go1.22.1:src/net/textproto/reader.go;l=650

lib/managedservicesplatform/runtime/contract/contract_test.go
```diff
@@ -0,0 +1,34 @@
...
+		c := contract.New(logtest.Scoped(t), mockServiceMetadata{}, e)
+		assert.NotZero(t, c)
+		assert.True(t, c.MSP)
+
+		// Expected to error, as there are missing required env vars.
+		err = e.Validate()
+		assert.Error(t, err)
+	})
+
+	// TODO: Add more validation tests
```
chrsmith, line 33: - Passing `nil` or `[]string` to `contract.ParseEnv` should work?
- Is this case sensitive? What about `MsP=true`
- How are awe parsing that value, e.g. `MSP=1` or `MSP=false`? 

Also, seems like we'd want to also confirm that various services were at least created. e.g.

```go
// Services that will always be provided.
assert.NotNilt(t, c.Diagnostics)
// Services that will not based on configuration.
assert.Nil(t, c.BigQuery)
```
-

lib/managedservicesplatform/runtime/contract/diagnostics.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/sourcegraph/log"
+
+	"github.com/sourcegraph/sourcegraph/lib/managedservicesplatform/runtime/internal/opentelemetry"
+	"github.com/sourcegraph/sourcegraph/lib/pointers"
+)
+
+type diagnosticsContract struct {
+	// DiagnosticsSecret can be used to authenticate diagnostics requests.
+	//
+	// 🚨 SECURITY: Do NOT use to authenticate sensitive data access.
```
chrsmith, line 19: What does this mean then? The comment and security warning seem to contradict one another. 

Is this saying:

> Do not use this string to authenticate access to any other sensitive data. This should only be used for governing access to diagnostic information. (And should still be treated with great care like any other application secrets.)

lib/managedservicesplatform/runtime/contract/diagnostics.go
```diff
@@ -0,0 +1,158 @@
...
+				defaultGCPProjectID),
+			OtelSDKDisabled: env.GetBool("OTEL_SDK_DISABLED", "false", "disable OpenTelemetry SDK"),
+		},
+		SentryDSN: env.GetOptional("SENTRY_DSN", "Sentry error reporting DSN"),
+
+		internal: internal,
+		msp:      msp,
+	}
+	if c.DiagnosticsSecret == nil {
+		logger.Warn("DIAGNOSTICS_SECRET not set, diagnostics handlers will not check for secrets")
```
chrsmith, line 51: Is "not check for secrets" accurate? From the other comments, shouldn't this say: "diagnostic handlers will not have any authorization checks"?

lib/managedservicesplatform/runtime/contract/diagnostics.go
```diff
@@ -0,0 +1,158 @@
...
+func (c diagnosticsContract) RegisterDiagnosticsHandlers(r HandlerRegisterer, state ServiceState) {
+	diagnosticsLogger := c.internal.logger.Scoped("diagnostics")
+
+	// Only enable Prometheus metrics endpoint if we are not in a MSP environment,
+	// i.e. in local dev.
+	if !c.msp {
+		// Prometheus standard endpoint is '/metrics', we use the same for
+		// convenience.
+		r.Handle("/metrics", promhttp.Handler())
+		// Warn because this should only be enabled in dev
```
chrsmith, line 95: Similar to the question above, we are saying `!c.msp` means "local-only". But if we are running locally, then why are we assuming that there is a Prometheus instance looking for data? Shouldn't we govern registering these handlers IFF the service is configured that way?

Or is this more of a ":shrug: if Prometheus wants to look for the data, there it is"?

Super-pedantic, minor point: Since this is a complete sentence, it should end in a period.

lib/managedservicesplatform/runtime/contract/diagnostics.go
```diff
@@ -0,0 +1,158 @@
...
+		diagnosticsLogger.Warn("enabled Prometheus metrics endpoint at '/metrics'")
+	}
+
+	// Simple auth-less version reporter
+	r.Handle("/-/version", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
chrsmith, line 100: Does `HandlerRegisterer` expose the underlying router? Could we clean this up with:

```go
// Register common MSP endpoints. All of which allow unauthenticated access.
r.HandlePrefix("/-/", func(r HttpRegisterer) {
    r.Handle("/version", ...)
    r.Handle("/healthz", ...)
})

File: lib/managedservicesplatform/runtime/contract/diagnostics.go [hunk 12]
```diff
@@ -0,0 +1,158 @@
...
+					log.String("query", r.URL.Query().Encode()),
+					log.Error(err))
+
+				w.WriteHeader(http.StatusInternalServerError)
+				_, _ = w.Write([]byte("healthz: " + err.Error()))
+				return
+			}
+
+			w.WriteHeader(http.StatusOK)
+			_, _ = w.Write([]byte("healthz: ok"))
```
chrsmith, line 122: Nit: You can clean this this pattern up by just using the (extremely poorly named) `http.Error` function.
https://pkg.go.dev/net/http#Error

```go
http.Error(w, "healthz: ok", http.StatusOK)
```

IMHO it's a lot cleaner and less clunky than `_, _ =` which also makes linters unhappy because you re ignoring an `error` return value.

lib/managedservicesplatform/runtime/contract/diagnostics.go
```diff
@@ -0,0 +1,158 @@
...
+		})))
+}
+
+// DiagnosticsAuthMiddleware uses DIAGNOSTICS_SECRET to authenticate requests to
+// next. It is used for debug endpoints that require some degree of simple
+// authentication as a safeguard.
+//
+// 🚨 SECURITY: Do NOT use this to authenticate sensitive data access.
+func (c diagnosticsContract) DiagnosticsAuthMiddleware(next http.Handler) http.Handler {
+	hasDiagnosticsSecret := func(w http.ResponseWriter, r *http.Request) (yes bool) {
```
chrsmith, line 132: Do we need this to be a local variable `hasDiagnosticsSecret` and have it return `yes bool`?

Since we already have `next http.Handler`, we can just simplify this by replacing:

```go
	if c.DiagnosticsSecret != nil {
		token, err := extractBearer(r.Header)
		if err != nil {
			// chrsmith: By spec, you should return 401 if the credentials
			// are missing or malformed. Returning 400 in this
			// situation makes things harder to diagnose.
			http.Error(w, err.Error(), http.StatusUnauthorized)
			return
		}

		result := subtle.ConstantTimeCompare([]byte(token), []byte(*c.DiagnosticsSecret))
		if result == 0 {
			// chrsmith: By returning 403 we may be leaking the
			// existence of the endpoint. But maybe it's OK (otherwise use 404).
			http.Error(w, err.Error(), http.StatusForbidden)
			return
		}
	}

	// No diagnostic secret was set, or the check passed.
	next.ServeHTTP(w, r)
```

chrsmith (approved):
High-level, this all looks good and as far as I understand what you are trying to do, this looks great. :shipit: 

I left a slew of code review comments, but pretty much all of them are minor and/or optional (I to a fault try to provide as much feedback as possible to improve things, so please don't take it as me trying to criticize or suggest your code isn't great to begin with!)

The only thing that I'd love to better understand (unrelated to this PR) is what the story is for "local development when using MSP services". e.g. what is the best way to "use as much of the MSP code/logic", but plug in local endpoints for connecting to the database or redis.

chrsmith:
@bobheadxi friendly ping. You plan on merging this, right? 😅 

I'd like to hook up the changes in this PR to the `abuse-ban-bot` to make some sorely needed improvements. e.g. persisting a record of when/why we banned a user in a database/BigQuery, or (🤞) being able to post messages directly to #alerts-gatekeeper-prod with a summary of the job's results. (e.g. number of users banned, etc.)

bobheadxi:
@chrsmith sorry for the delay! I was OOO Monday-yesterday - I will address your review comments tonight (my time) and get this merged ASAP

lib/managedservicesplatform/runtime/contract/contract.go
```diff
@@ -0,0 +1,131 @@
...
+//
+//	// Check for errors on Env retrieval (missing/invalid values, etc.)
+//	if err := e.Validate(); err != nil { ... }
+//
+//	// Use Contract helpers and configuration values
+//	writer, _ := c.BigQuery.GetTableWriter(ctx, "my-table")
+//	writer.Write(...)
+//
+// For more help, please reach out to #discuss-core-services or refer to
+// go/msp: https://handbook.sourcegraph.com/departments/engineering/teams/core-services/managed-services/platform/
```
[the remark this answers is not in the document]
bobheadxi: Yeah, but it sounds like the account owners cancelled too late, and IMO they do come in handy, so... I'll keep suggesting them for now 😅 In the even they ever do get discontinued, it should be fairly easy to find them all again and change them up

lib/managedservicesplatform/runtime/contract/contract.go
bobheadxi: The idea is to generally allow the use of `Contract` in local development:

1. Pretty much all configurations are env-vars, so if you find the right one to replace in dev, you can make things work pretty seamlessly (e.g. Redis is just 1 env var).
2. For special configurations like Postgres, there is `PGDSN` which allows for a template string to point to an arbitrary PG instance (since production uses a magic Cloud SQL connection string instead) - this is documented on `OpenDatabase` docstring already
3. For all other scenarios, you can use `Contract.MSP` to wire up your own custom behaviours as you wish

Admittedly, this is pretty case-by-case currently - I opened an issue to track writing up a docs page about this:

lib/managedservicesplatform/runtime/contract/contract.go
bobheadxi: The "sub-contract" types are intentionally useless to callers, which is intentional: the internal configuration values, naming, etc have changed a lot, so we don't want callers to be passing around `postgreSQLContract` or constructing them on their own (they should instead rely on the corresponding env vars).

When you export a private type like this, `Contract.PostgreSQL.OpenDatabase` etc are still available, you just can't pass the type around (except for the top-level `Contract` type)

bobheadxi: I pushed a5eca895de4186cdbbf6b5d53179c8dbb9cb546c to explain this in the `Contract` docstring, let me know if that still falls short and I'll follow up with more details!

lib/managedservicesplatform/runtime/contract/contract.go
bobheadxi: You're right on both `e.g.`, and docstrings definitely be added - I've included some in 048f29e86bbd0d325f92437e33f9627f6b72364e to explain what we want from each and what we use them for, thank you!

lib/managedservicesplatform/runtime/contract/contract.go
bobheadxi: All runtime services can also be run using `-help` to list all referenced environment variables and their required-ness - I added 70ca7b1ecaec5016cd06fb448a6bf147a51b2ea5 to document callsites that lead to information in `-help`, as well as on `Contract` that you can use this to get a list of all required env vars.

lib/managedservicesplatform/runtime/contract/contract_test.go
bobheadxi: Let me follow up on the tests front, as this is a sizeable diff - a lot of core env stuff is copy-pasta from the normal sourcegraph equivalent, and there are minimal logic diffs here so nothing should be broken

lib/managedservicesplatform/runtime/contract/diagnostics.go
bobheadxi: Your suggestion is about right - it can be used to authenticate basic diagnostics checks (we primarily use it for `/-/healthz` right now), and it is not regularly rotated. Added 275a368ae459edea3ce34ec9fe7f7b8d9a429df3 with your suggestion, thanks!

lib/managedservicesplatform/runtime/contract/diagnostics.go
bobheadxi: Yes, you're right - improved the message here, and added a docstring, for 131331b88d5ebb1f733d2c28703cc4d99b964d77

lib/managedservicesplatform/runtime/contract/diagnostics.go
bobheadxi: > Or is this more of a ":shrug: if Prometheus wants to look for the data, there it is"?

This is it - also, Prometheus `/metrics` are pull-based, so having it available is harmless. Tried to make it more clear in d1c5c5c2ec8bfe0b715c85a10d3ccd10da9f3afc, but I think we should mention this stuff in https://github.com/sourcegraph/managed-services/issues/1073

lib/managedservicesplatform/runtime/contract/diagnostics.go
bobheadxi: I'm preferring to keep the API surface here simple so that authors using bespoke libraries have less things to mock/stub out/worry about potential differences in pathing handling and what not. For this case, we just very explicitly set up the exact paths we are interested in. Being specific on the few endpoints we add here also makes the "magic" endpoints easier to search for as well - WDYT?

lib/managedservicesplatform/runtime/contract/diagnostics.go
bobheadxi: While nice, this sadly also adds a `fmt.Fprintln(w, error)` according to the source code, which is kind of annoying 😬

lib/managedservicesplatform/runtime/contract/diagnostics.go
bobheadxi: I originally did to make the next part (encoding what the returned handler even does) easier to read:

```
		if !hasDiagnosticsSecret(w, r) {
			return
		}
		next.ServeHTTP(w, r)
	})
```

bobheadxi: I think it's a bit clearer this way to explain "this checks if the request is 'authorized' by our standards", then a separate block (`hasDiagnosticsSecret` definition) explaining when we decide a request is authorized

bobheadxi:
@chrsmith thanks for the extensive review comments and feedback! Opened a few issues to address follow-ups, and addressed most of your open comments - happy to discuss unaddressed points further post-merge :) Going to land this now so that it's at least available first

PR #61494, opened 2024-03-29
ci: set tag refspec for semgrep when working with tags

burmudar:
The semgrep check fails on tags since it uses the branch refspec but on a tag those do not exist see https://buildkite.com/sourcegraph/sourcegraph/builds/266947#018e8a09-c8ad-4a3e-bf74-92cdb048f8da
## Test plan
CI and https://buildkite.com/sourcegraph/sourcegraph/builds/266955#018e8a2f-a110-409e-83c0-5d196b193857


Strum355 (approved):

shivasurya:
Thank you for handling this🙏🏻

PR #61496, opened 2024-03-29
k8s/resource: Add pkg for creating k8s ingress resources

jdpleiness:
Add `ingress` pkg to create k8s ingress with patterns and defaults common to Sourcegraph.

## Test plan


DaedalusG (approved):
Looks good, same general feedback as here: https://github.com/sourcegraph/sourcegraph/pull/61474#pullrequestreview-1969685740

PR #61501, opened 2024-03-29
k8s/resource: Add pkg for creating k8s deployment resources

jdpleiness:
Add `deployment` pgk to create k8s deployment with patterns and defaults common to Sourcegraph.

## Test plan


DaedalusG (approved):
🛠️

PR #61502, opened 2024-03-29
Update GQL docs for user activity metrics

dadlerj:
Update GQL docs to explain the difference between two types of user activity statistics. This is a common source of confusion, and this appears to be the best place to document the difference. Open to suggestions for other spots.

Recent example: https://sourcegraph.slack.com/archives/C05EMJM2SLR/p1711735853127489

## Test plan

CI


dadlerj:
Sorry for tagging you Robert, I don't know who else to go to 🤦 . Just a quick documentation update.

bobheadxi (approved):
Not terribly familiar with this area, but this is just docstring improvements and those are always great, so approve to unblock 😁

PR #61505, opened 2024-03-29
k8s/resource: Add pkg for creating k8s stateful set resources

jdpleiness:
Add `statefulset` pkg to create k8s statefulsets with patterns and defaults common to Sourcegraph.

## Test plan

### DIFF
--- a/internal/k8s/resource/statefulset/statefulset.go
+++ b/internal/k8s/resource/statefulset/statefulset.go
@@ -1,0 +1,152 @@
+package statefulset
+
+import (
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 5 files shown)


internal/k8s/resource/statefulset/statefulset.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/sourcegraph/sourcegraph/lib/pointers"
+)
+
+// NewStatefulSet creates a new k8s StatefulSet with default values.
+//
+// Default values include:
+//
+//   - MinReadySeconds: 10
+//   - RevisionHistoryLimit: 10
+//   - UpdateStrategy: RollingUpdateStrategy
```
DaedalusG, line 18: Nice

DaedalusG (approved):
Same as: https://github.com/sourcegraph/sourcegraph/pull/61474#pullrequestreview-1969685740

PR #61512, opened 2024-04-01
build-tracker: trigger build-metrics pipeline on build.finished

Strum355:
Triggers a buildkite pipeline when a build.finished event is received from buildkite in order to collect metrics & information about the build (coming in a later PR).

The details we attach as part of the build can be iterated upon as needed. For the most part, we can get all the details of the build that ultimately is triggering this by using the buildkite cli to query the build by its ID from the `DEVX_TRIGGERED_FROM_BUILD_ID` env var

A drawback appears to be that it will always show the owner of the token as the author/creator of the build (although the committer is preserved in `BUILDKITE_BUILD_AUTHOR` env var of the jobs)

Depends on https://github.com/sourcegraph/managed-services/pull/1096

## Test plan

Tested with payloads fetched from the webhook log, and with a personal token (see builds here: https://buildkite.com/sourcegraph/devx-build-metrics/)

### DIFF
--- a/dev/build-tracker/config/config.go
+++ b/dev/build-tracker/config/config.go
@@ -5,15 +5,17 @@ import "github.com/sourcegraph/sourcegraph/lib/managedservicesplatform/runtime"
 const DefaultChannel = "#william-buildchecker-webhook-test"
 
 type Config struct {
-	BuildkiteToken string
-	SlackToken     string
-	SlackChannel   string
-	Production     bool
-	DebugPassword  string
+	BuildkiteWebhookToken string
+	BuildkiteToken        string
+	SlackToken            string
+	SlackChannel          string
+	Production            bool
+	DebugPassword         string
 }
 
 func (c *Config) Load(env *runtime.Env) {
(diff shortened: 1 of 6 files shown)


dev/build-tracker/config/config.go
```diff
@@ -5,15 +5,17 @@ import "github.com/sourcegraph/sourcegraph/lib/managedservicesplatform/runtime"
...
+	SlackToken            string
+	SlackChannel          string
+	Production            bool
+	DebugPassword         string
 }
 
 func (c *Config) Load(env *runtime.Env) {
-	c.BuildkiteToken = env.Get("BUILDKITE_WEBHOOK_TOKEN", "", "")
+	c.BuildkiteWebhookToken = env.Get("BUILDKITE_WEBHOOK_TOKEN", "", "")
+	c.BuildkiteToken = env.Get("BUILDKITE_TOKEN", "", "")
```
Strum355, line 18: We need to create a token in GSM for this cc @burmudar maybe theres a bot account we have?

burmudar (approved):

Issue #46026, opened 2023-01-02
before: search keyword in diff search is ineffective


sqs:
In the query [`file:featureflag type:diff before:7months`](https://sourcegraph.sourcegraph.com/search?q=context:sourcegraph+f:featureflag+type:diff+before:7months&patternType=standard&sm=0&groupBy=author) (which I was doing to see who to thank for building our feature flag system!), the `before:` operator is not taking effect.

The results include matches from 10 and 12 days ago. I expect no results newer than 7 months.

![image](https://user-images.githubusercontent.com/1976/210198821-d4fc9cf2-53a4-4c12-bf6a-d170e24397e2.png)

sqs:
Hmm, it might be that `after:` and `before:` got switched, since `after:` shows no results (which is unexpected, since there are certainly results in the last 7 months!).

https://sourcegraph.sourcegraph.com/search?q=context%3Asourcegraph+f%3Afeatureflag+type%3Adiff+repo%3Asourcegraph%24+after%3A6months&patternType=standard&sm=0&groupBy=author

camdencheek:
The [natural date parser we use](https://github.com/tj/go-naturaldate) expects relative dates of the form "7 months ago". Based on a couple of tests, it appears to be tokenizing the `7`, failing to find the keyword `ago`, and interpreting `7months` as `7:00 today`, which would explain why there doesn't seem to be any filtering going on. 

We used that package so we could move quickly, but we might eventually want to just write our own.

limitedmage:
Can we instead use search query code intel so that the user hovers over the time string and sees the timestamp? Similar to what we do for eg. regex 
![image](https://user-images.githubusercontent.com/206864/210264828-f3ef61eb-b021-44c2-a0d9-a7bf4586b2fe.png)

limitedmage:
This would also fix the ambiguity of timezone parsing (we always parse exact times as UTC if they don't have a timezone.

camdencheek:
Ooh, that's a cool idea. I could definitely expose the date parsing via the API, then it would just be a matter of hooking up the hovers to call that

jtibshirani:
@camdencheek can we close this out now that you improved the semantics of 'before' in https://github.com/sourcegraph/sourcegraph/pull/61513?

camdencheek:
[...]

PR #61513, opened 2024-04-01
RFC: point-in-time searching

camdencheek:
# Summary

This is a proposal to add the ability to do point-in-time searches. I'd like to add the ability to search a repo's state at a timestamp rather than at a specific commit, branch, or tag. This PR adds a `rev:at.time(5 days ago, my-branch)` predicate that allows a user to do just that. 

[Related discussion in Slack](https://sourcegraph.slack.com/archives/C05R619V4F8/p1711652215691879)

# Motivation
- For Code Insights, this is useful for chart drilldown so we can generate searches that more closely represent the results that produced that data point (currently we use a diff search, which is not a very good proxy for a code insights data point).
- For general code searching, it's often useful for exploratory searching to look at the state of a repo (or repos) at a given point in time. I often find myself doing something like `HEAD~1000` hoping to find a commit that's approximately in the time period I'm interested in. On the command line, I can use `git log --before`, but that's not available outside of commit/diff search. 
- Customers often use search contexts to collect a set of tags together into "release" contexts, but that requires building tooling that generates these search contexts, and not all repos will necessarily have the same tagging conventions. This provides a zero-config way to get less precise, but functionally similar behavior. 

# Discussion

## Syntax

None of the syntax is set in stone. Would really love feedback on this part. ~To stop myself from bikeshedding on the exact syntax, this PR is implemented with the simplest thing I could come up with: `rev:ancestor.at(branch, 5 days ago)`, or `rev:ancestor.at(5 days ago)`, which is equivalent to `rev:ancestor.at(HEAD, 5 days ago)`.~ This PR implements the syntax as `rev:at.time(5 days ago)`. This defaults to targeting HEAD, but can take an optional second argument to choose a different revision as the tip like `rev:at.time(2021-01-01, cc/my-branch)`.

Currently, there is no user-facing way to add the revision to the repo filter because I couldn't find any syntax I liked for that. However, due to a [quirk in how our query language is parsed](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@997aecb3624c339fd0bdf2d4decef756388b84df/-/blob/internal/search/query/transformer.go?L571-573), all `rev:` values need to be representable in the `repo:reponame@rev` position, so this super hackily json+base64 encodes the predicate struct. Note that a user will never actually see that, it's just an internal detail. I think this is acceptable for now, and if we decide on a good syntax it will be a non-breaking change.

## Behavior

Because of force pushes, merge commits, and overriding commit dates, we cannot guarantee to yield the state of a branch at a given date. However, we _can_ make a feature that 1) almost always matches user intent, and 2) is well-defined.

So, to define the behavior: `RevAtTime` returns the nearest ~(by generation number) ancestor commit~ commit in the linear history defined by `--first-parent` that has a committer timestamp (not author timestamp) before the requested timestamp. ~In the case of a tie in generation number (which can happen with non-linear commit histories introduced by merge commits), we return the commit with the greater timestamp.~

~Merges also make this a little complicated, so we might want to linearize the commit history with `--first-parent`.~

~Consider the following commit graph where e is a merge commit:~
```
  a
 / \
b   |
|   |
c   d
 \ /
  e
```

~If e is after the target date, but c and d are before, which commit should we yield as the target? As implemented, we’ll always select the newest one, but if the newest one is d, then we’ll skip b because it’s not an ancestor of d (even though it is an ancestor of another commit that matches the before filter).~

Git commiter dates are not guaranteed to be monotonically increasing (they can be overridden with `GIT_COMMIT_DATE`), but in practice they _usually_ are. Author dates are significantly more variable because commits are often authored before they are merged (think cherry picks). Since the goal of this feature is to find the state of a branch at a given timestamp, I think committer date is more appropriate than author date and this should not be configurable.

## Performance

As implemented, this requires a gitserver query during the repo resolution step to resolve a `date=` rev into a commit hash. The `git log` command it runs is generally fast (<50ms for sourcegraph/sourcegraph), but its cost scales with the depth of the commit graph and the distance it has to traverse to find the first matching commit. I expect the cost of these queries will be dominated by the unindexed searches they generate (non-HEAD searches will generally be unindexed).

After some perf testing, to limit the cost of repeated queries of this type, I've added a small in-memory cache.

# Test Plan

- Added predicate parsing tests
- Added git CLI tests
- Added API integration tests
- Manual testing of completions and hover docs


cmd/gitserver/internal/git/gitcli/revattime.go
```diff
@@ -0,0 +1,31 @@
...
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"time"
+
+	"github.com/sourcegraph/sourcegraph/internal/api"
+)
+
+func (g *gitCLIBackend) AncestorAtTime(ctx context.Context, spec string, t time.Time) (api.CommitID, error) {
```
eseliger, line 13: thx for implementing this as a proper gRPC call aligned with our current work to migrate all the things :)

camdencheek: I'm as excited to get rid of Exec as you are 😄

eseliger: :dab:

sqs:
This is cool! Should we support `git rev-parse` syntax like `git show 'HEAD@{3 months ago}'` here? `rev:"{3 months ago}"`? It looks weird but maybe that could be more standard for people familiar with that syntax.

eseliger:
One thing that is weird about the above ref syntax is that git uses reflog for determining these dates, not the commit/author date, which camden is using in this PR (and intuitively makes more sense IMO).

So the above syntax that you provided means "HEAD as it was on Sourcegraph 3 months ago". 
Here's that for a freshly cloned repo:

```
git rev-parse 'HEAD@{3 days ago}'
warning: log for 'HEAD' only goes back to Wed, 3 Apr 2024 00:16:15 +0200
9761f2597af73fa2db530f0afc3c7258538c9dd2
```

camdencheek:
Yep what Erik said. I agree the syntax is more intuitive, but `rev@{3 days ago}` has meaning to git that is different than what we're doing here (and doesn't actually make much sense in the context of a headless repo), so I don't know that we want to reuse the syntax exactly. Maybe we should use that syntax as inspiration though and pick something closer to that, e.g. `repo:myrepo@HEAD%{5 days}` or something. Will noodle on that.

keegancsmith (commented):
-   I like this a lot and would use it fairly often.
    -   I often want to start a search from like 2 years ago to check a hypothesis.
-   I like how it is introduced into our syntax.

-   how do we make it clear which timestamp we search?
    -   I guess we only allow and show commit timestamp?
    -   for repos that we convert to git (eg perforce) is the commit timestamp useful or is it actually the author timestamp?
-   timezones for non-relative time queries
    -   relative is fine because 5 days ago is relative to time.Now.
    -   but non-relative (absolute?) without TZ in it is likely parsed as UTC by the server (or your macbook's timezone in your devenv)
    -   I suspect a user would want it to be the browser TZ.
    -   This also then brings up questions about sharing links with those sort of timestamps in them.
-   how do we handle non-linear history?
    -   I like the idea you propose of &#x2013;first-parent. That feels like the more useful value for user intent.
    -   I'm unsure about `--date-order`? It feels like we should stick to a cut vertex in the commit graph. I suppose there are people who really do create funky graphs in how they do dev, and in that case my intuition of cut vertex is wrong.
-   how do we handle non-monotonic commit or author times? EDIT: you address this and I forgot to update my notes about it.
-   `rev:ancestor.at(my-branch, 5 days ago)` how could this conflict with actual rev names?
-   concern around performance implications if people start using it in search contexts.

bahrmichael:
This looks really cool! I would probably abuse it for bisect searches (unless there's something better for this).

My only concern is around performance. You mentioned a fast 50ms query. Do you know if there's an upper bound to how long this could take, or could it become increasingly slower as repos get larger and commit histories longer? Mainly concerned about noisy neighbor problems, esp. related to automated queries (e.g. from code insights).

camdencheek:
> timezones for non-relative time queries

Yeah, that's tricky. We run into this with `before` and `after` too. The server uses UTC, but users expect relative days to mean local time. There's some more discussion [here](https://github.com/sourcegraph/sourcegraph/issues/46026), along with an idea to provide a hover with the resolved date (which I think is a good idea). I personally don't think we should interpret queries differently for different users just because of the sharability aspect.

> It feels like we should stick to a cut vertex in the commit graph.

I would generally agree, but there's not really an easy way to do this without abandoning the git CLI entirely AFAIK. The exception to that being if we use `--first-parent`, the graph is linear, so every vertex is a cut vertex.

The downside of `--first-parent` is it skips the history of all merged branches. So for places that heavily use merge commits, the granularity of the search will be at the "PR" level rather than at the commit level, which can be quite meaningful for long-lived branches. I'm starting to be convinced that this is a worthwhile tradeoff though, and that the simplicity of `--first-parent` outweights the loss of granularity.

> how could this conflict with actual rev names?

The predicate parser only looks for things of the shape `rev:ancestor.at([^)]*)`. I expect it to be vanishingly unlikely that it would conflict, and in the case that it does, we support quoting values like `rev:"ancestor.at(abc)"` (at least on paper...I haven't actually tested this in a while).

> concern around performance

My thought here is that perf cost will be dominated by actually doing the unindexed searches. That's part is no different than a search context that contains a bunch of unindexed revs. Comparatively, the only added cost is the commit resolution, which is non-zero, but not massive either.

As a data point, the `chromium` repo (largest I have locally with 1.2M commits) takes ~12 seconds to traverse its entire history. It takes 100ms with `--before="1 year ago"`. That seems like an acceptable perf tradeoff to me for a somewhat niche feature.

And in the case perf does become a significant concern, the work the Source team has done to virtualize git calls means we could pretty easily add a cache or index to this specific lookup.

cmd/gitserver/internal/git/gitcli/revattime.go
```diff
@@ -0,0 +1,31 @@
...
+	r, err := g.NewCommand(ctx, WithArguments(
+		"log",
```
mmanela, line 15: Is this where you are planning to add `--first-parent`?

cmd/gitserver/internal/git/observability.go
```diff
@@ -289,6 +304,7 @@ type operations struct {
 	getCommit       *observation.Operation
 	archiveReader   *observation.Operation
 	resolveRevision *observation.Operation
+	revAtTime       *observation.Operation
```
mmanela, line 307: Is there a reason we name the operation different than the backend method?

jtibshirani:
The proposed behavior makes sense to me. +1 to using `--first-parent`, this feels easiest to understand for users.

> My thought here is that perf cost will be dominated by actually doing the unindexed searches.

Now that we use hybrid search by default (yay!) many unindexed searches are very fast. But as you say, there's a lot we can do here -- for example a simple cache of time expression -> hash could really help. Also I assume the command scales logarithmically with history length (not linear) so there's not a huge concern with large histories?

Personally I find the naming a bit surprising because I don't think about "git ancestors" in my day to day dev work :) Throwing out some other ideas:
* `rev:commit.at(...)`
* `rev:at.time(...)` -> this feels consistent with the format we use for others like `repo:has.topic`

camdencheek:
>  Also I assume the command scales logarithmically with history length (not linear) so there's not a huge concern with large histories?

It does not. Git cannot assume a binary search is safe because commit dates are not guaranteed to be monotonic. So it has to iterate commit-by-commit. (I'm also not even sure Git has the indexes that would be needed to do a binary search, but that's another matter entirely)

> Personally I find the naming a bit surprising

Thank you! I also dislike the naming, so I really appreciate the suggestions. 

For the examples you gave, do they still make sense to you for non-HEAD commits? I read `rev:at.time(yesterday)` to mean "HEAD at yesterday", which I think is a good default. But what about if I wanted to search the history of other branches? Does `rev:at.time(my-branch, yesterday)` make sense? Or maybe `rev:at.time(yesterday, my-branch)` to put the optional param last? Obviously, we can document this all, but an intuitive syntax is preferable 🙂

camdencheek:
Or maybe we could even get funky and allow multiple predicates like `rev:branch(my-branch):at.time(yesterday)`

cmd/gitserver/internal/git/observability.go
camdencheek: Nope, just an oversight. Will fix, thanks!

jtibshirani:
> It does not. Git cannot assume a binary search is safe because commit dates are not guaranteed to be monotonic.

Oof, I see! Then a cache feels really valuable, even in the initial implementation, as a way of reducing load and making sure follow-up queries are fast.

camdencheek:
> cache feels really valuable, even in the initial implementation,

Done!

camdencheek:
Updates from feedback:
- I added `--first-parent` to linearize the history and make it easier to explain
- I updated the syntax to `rev:at.time(yesterday)` or `rev:at.time(yesterday, my-branch)`. Along with that, I've renamed all the backend pieces to match the "rev at time" naming. 
- I added a simple LRU cache to the commit resolution to mitigate the cost of repeated, expensive git commands

camdencheek:
Since there seems to be general agreement on direction, I'll work towards productionizing this PR (add tests mostly). Feel free to add more feedback in the mean time, but I'll explicitly request a review once it's ready for full code review

cmd/gitserver/internal/git/gitcli/revattime.go
```diff
@@ -0,0 +1,53 @@
...
+type revAtTimeCacheKey struct {
+	spec string
+	t    time.Time
+}
+
+func (g *gitCLIBackend) RevAtTime(ctx context.Context, spec string, t time.Time) (api.CommitID, error) {
+	key := revAtTimeCacheKey{spec, t}
+	if entry, ok := g.revAtTimeCache.Get(key); ok {
+		return entry, nil
+	}
```
camdencheek, line 31: Not sure whether this should live in the `gitCLIBackend` or whether I should wrap the `gitCLIBackend` in a struct that provides the cache for this method.

cmd/gitserver/internal/git/gitcli/revattime.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/sourcegraph/sourcegraph/internal/api"
+)
+
+// A simple in-process cache to mitigate the cost of any regular
+// rev:at.time() queries.
+//
+// Each cache entry takes ~128 bytes, so 8192 entries should keep this cache
+// around 1MB.
+var globalRevAtTimeCache, _ = lru.New[revAtTimeCacheKey, api.CommitID](8192)
```
mmanela, line 20: Doesn't this need to have the repo id/name as part of the key?

camdencheek: Ha, yup, sure does. Thanks for the quick catch. Saved me some debugging time 😄

mmanela:
Just learned about the `before:` and `after:` query params we have. Would there be any confusion in this proposal with those existing filters?

camdencheek:
> Would there be any confusion in this proposal with those existing filters?

Honestly, probably. But I do think the syntax difference is justified and we should not try to unify them. `before` and `after` are for commit and diff search _only_. Time range queries make sense for commit and diff search because we are returning sets of commits. This RFC is different in that we are still searching a repo at a specific _point_ in time. `rev:at.time()` will always resolve to a single commit per repo, and our search results are file contents (not commits).

I think any confusion should be resolvable with good documentation (which I will add along with this addition).

camdencheek:
Okay! I think this is ready for a full code review now. I've added tests and query completions/hovers since the last update.

cmd/gitserver/internal/server_grpc.go
```diff
@@ -1200,6 +1200,59 @@ func (gs *grpcServer) ResolveRevision(ctx context.Context, req *proto.ResolveRev
...
+				Spec: e.Spec,
+			})
+			if err != nil {
+				return nil, err
+			}
+			return nil, s.Err()
+		}
+
+		gs.svc.LogIfCorrupt(ctx, repoName, err)
+		// TODO: Better error checking.
```
bahrmichael, line 1241: I assume you'll address this before merging.

internal/search/query/repo_revs.go
```diff
@@ -127,12 +144,44 @@ func ParseRepositoryRevisions(repoAndOptionalRev string) (ParsedRepoFilter, erro
...
-func ParseRevisionSpecifier(spec string) RevisionSpecifier {
+func ParseRevisionSpecifier(spec string) (RevisionSpecifier, error) {
 	if strings.HasPrefix(spec, "*!") {
-		return RevisionSpecifier{ExcludeRefGlob: spec[2:]}
+		return RevisionSpecifier{ExcludeRefGlob: spec[2:]}, nil
 	} else if strings.HasPrefix(spec, "*") {
-		return RevisionSpecifier{RefGlob: spec[1:]}
+		return RevisionSpecifier{RefGlob: spec[1:]}, nil
+	} else if strings.HasPrefix(spec, revAtTimePrefix) {
+		aat, err := ParseRevAtTime(spec[4:])
```
bahrmichael, line 156: What does `aat` sand for?

bahrmichael (approved):
No blockers :)

client/shared/src/search/query/filters.ts
```diff
@@ -235,7 +234,6 @@ export const FILTERS: Record<NegatableFilter, NegatableFilterDefinition> &
         negatable: true,
         description: negated =>
             `${negated ? 'Exclude' : 'Include only'} results from file paths matching the given search pattern.`,
-        discreteValues: () => [...predicateCompletion('file')],
```
jtibshirani, line 238: Could you explain this change?

internal/search/query/date_format.go
```diff
@@ -75,7 +75,7 @@ func ParseGitDate(s string, now func() time.Time) (time.Time, error) {
 
 	// Human date
 	n := now()
-	if t, err := naturaldate.Parse(s, n); err == nil && t != n {
+	if t, err := naturaldate.Parse(s, n, naturaldate.WithDirection(naturaldate.Past)); err == nil && t != n {
```
jtibshirani, line 78: Could you explain this change?

internal/search/repos/repos.go
```diff
@@ -528,6 +528,18 @@ func (r *Resolver) normalizeRepoRefs(
 			globs = append(globs, gitdomain.RefGlob{Include: rev.RefGlob})
 		case rev.ExcludeRefGlob != "":
 			globs = append(globs, gitdomain.RefGlob{Exclude: rev.ExcludeRefGlob})
+		case rev.RevAtTime != nil:
+			commitOID, ok, err := r.gitserver.RevAtTime(ctx, repo.Name, rev.RevAtTime.RevSpec, rev.RevAtTime.Timestamp)
+			if err != nil {
+				if errors.Is(err, context.DeadlineExceeded) || errors.HasType(err, &gitdomain.BadCommitError{}) {
+					return nil, err
+				}
+				reportMissing(RepoRevSpecs{Repo: repo, Revs: []query.RevisionSpecifier{rev}})
```
jtibshirani, line 537: Checking I understand -- if a time is too far in the past for some repo, we'll expose this as "missing" like we do for revisions that don't exist.

cmd/gitserver/internal/git/gitcli/revattime_test.go
```diff
@@ -0,0 +1,100 @@
...
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/sourcegraph/sourcegraph/internal/api"
+)
+
+func TestGitCLIBackend_RevAtTime(t *testing.T) {
```
jtibshirani, line 13: Nice tests 😎

jtibshirani (commented):
This looks good to me! Just left some questions, and I'm going to stare at the git command another time before approving :)

cmd/gitserver/internal/git/gitcli/revattime_test.go
camdencheek: Thanks!

internal/search/query/date_format.go
camdencheek: Ah, yeah. The other place we use relative dates is `before:` and `after:`. The `naturaldate` library also supports relative dates in the future, but for the purpose of `before`, `after`, and `rev:at.time()`, a date in the future is very unlikely to be what the user actually intends. So it tells the `naturaldate` parser that the thing it's parsing should be something in the past. e.g. `5 days` will be interpreted as "5 days ago" rather than "5 days from now"

client/shared/src/search/query/filters.ts
camdencheek: Ha, yeah. Predicates are handled separately [here](https://github.com/sourcegraph/sourcegraph/pull/61513/files#diff-f14b9aee7bdf8db39a3a793d0a68d25bd112f70e06f9e9d88abd27e21c21cb2eR518). The predicates added to `discreteValues` were never used, which made debugging very confusing. Additionally, `discreteValues` are expected to be literals, not snippets, so forcing all predicates to go through `discreteValues` would break snippet expansion / templating.

I think this is likely just a leftover from an old refactor, and it can probably be cleaned up more, but for now "get rid of the lines that do nothing" seemed like the best change for this PR.

internal/search/repos/repos.go
camdencheek: Not quite. An error will only be returned if the _starting revision_ can't be found. e.g. if you do `rev:at.time(yesterday, my-branch)`, and `my-branch` doesn't exist for a repo. Or the repo itself doesn't exist (which should be rare since we just listed it from the database).

In the case that there is no commit that exists at the requested time, that is a "no result" scenario and there is no error or "missing" notification.

So:
- `repo:sourcegraph/sourcegraph rev:at.time(yesterday, my-branch-that-does-not-exist)`  will yield an alert that `my-branch-does-not-exist` does not exist
- `repo:sourcegraph/sourcegraph rev:at.time(20 years ago)` will just have no results.

I think this is reasonable behavior, but I'm certainly willing to be persuaded otherwise.

cmd/gitserver/internal/server_grpc.go
camdencheek: Oh, this was actually just copied from another chunk of code that I did not write. Given I don't actually understand the intent behind the TODO, I'm gonna let the Source team own fixing that one 🙂

internal/search/query/repo_revs.go
camdencheek: Whoops. That was a leftover from when this was called "AncestorAtTime". Will rename it to a more sane var name

internal/search/query/date_format.go
jtibshirani: Ah I see. So this will immediately change the behavior for `before` and `after` filters like "5 days"? That seems okay to me, but it's good to be aware of this.

internal/search/repos/repos.go
jtibshirani: Ah, now I see the commit ID is empty in this case, and gitserver returns "ok" only if `res.GetCommitSha() != ""`. It'd be nice to add documentation to the client interface about what "ok" means exactly. 

I don't have a strong preference, but it feels cleaner to treat this as an error everywhere so we don't have the "no error, but not ok" state, which I always find a bit confusing.

camdencheek: > treat this as an error 

Is this "treat it as an error" from a user perspective or a Go perspective? I can certainly modify the signature to return a special error type, but want to make sure I'm understanding correctly

jtibshirani: I was mainly thinking about the gitserver client, which has the signature
```
RevAtTime(ctx context.Context, repo api.RepoName, spec string, t time.Time) (api.CommitID, bool, error)
```
I don't have a strong feeling about the client behavior, and could see an argument either way (just returning no results vs. matching what we do for missing rev specs).

jtibshirani (approved):
I've now pondered relative dates and `--first-parent` and this still looks good to me :)

internal/search/repos/repos.go
camdencheek: > "no error, but not ok" state

Real optionals in Go sure would be nice 😄

camdencheek: I've renamed the return value to "found" for clarity (instead of "ok" which, though idiomatic, implies a "non-ok" state, which is weird when there's also a possible error).

I tried two other things:
- Return a sentinel error value. I didn't love this because the nonexistence of a commit before the target time does not feel like an error to me. That's an expected outcome for some searches. Also, we'd need to assert on the error type, which causes us to lose some type safety. 
- Return an empty string as a sentinel that no such commit exists. I also didn't like this because it's easy to miss that the return value can be empty, and that should be handled specially. The nice thing about the third return value is that it forces the caller to do something with it.

RE: client behavior, I'm also not tied to the current behavior. Going to go ahead and merge as-is, but certainly won't be offended if someone follows up changing it.

internal/gitserver/v1/gitserver.proto
eseliger, line 1: This diff seems a bit noisy, what tool are you using to format? 
Just trying to keep merge conflicts small 😬

internal/gitserver/v1/gitserver.proto
```diff
@@ -166,6 +174,17 @@ message ResolveRevisionResponse {
   string commit_sha = 1;
 }
 
+message RevAtTimeRequest {
+  string repo_name = 1;
+  bytes rev_spec = 2;
+  google.protobuf.Timestamp time = 3;
+}
```
eseliger, line 181: bit of a weird ask but this will help me with a future migration 😬 

Also, can you add a quick docstring here what time and revspec are used for so callers don't have to dig all the way through to the gitcli package?

```suggestion
message RevAtTimeRequest {
  // repo_name is the name of the repo to run the blame operation in.
  // Note: We use field ID 2 here to reserve 1 for a future repo int32 field.
  string repo_name = 2;
  bytes rev_spec = 3;
  google.protobuf.Timestamp time = 4;
}
```

internal/gitserver/v1/gitserver.proto
```diff
@@ -75,83 +75,91 @@ service GitserverService {
...
-  // If the given repo is not cloned, it will be enqueued for cloning and a NotFound
-  // error will be returned, with a RepoNotFoundPayload in the details.
+  // If the given repo is not cloned, it will be enqueued for cloning and a
+  // NotFound error will be returned, with a RepoNotFoundPayload in the details.
   rpc ResolveRevision(ResolveRevisionRequest) returns (ResolveRevisionResponse) {
     option idempotency_level = IDEMPOTENT;
   }
+  rpc RevAtTime(RevAtTimeRequest) returns (RevAtTimeResponse) {
+    option idempotency_level = NO_SIDE_EFFECTS;
+  }
```
eseliger, line 154: What does it do? 😬 
Also, maybe a quick docstring about well-known errors this can return (RevisionNotFoundPayload for example)

```suggestion
  // RevAtTime ...
  //
  // If the given repo is not cloned, it will be enqueued for cloning and a
  // NotFound error will be returned, with a RepoNotFoundPayload in the details.
  rpc RevAtTime(RevAtTimeRequest) returns (RevAtTimeResponse) {
    option idempotency_level = NO_SIDE_EFFECTS;
  }
```

internal/gitserver/commands.go
```diff
@@ -936,6 +937,35 @@ func (c *clientImplementor) ResolveRevision(ctx context.Context, repo api.RepoNa
...
+		RevSpec:  []byte(spec),
+		Time:     timestamppb.New(date),
+	}
+	res, err := client.RevAtTime(ctx, req)
+	if err != nil {
+		return "", false, err
+	}
+
+	return api.CommitID(res.GetCommitSha()), res.GetCommitSha() != "", nil
+
```
eseliger, line 966:
```suggestion
```

internal/gitserver/commands.go
```diff
@@ -936,6 +937,35 @@ func (c *clientImplementor) ResolveRevision(ctx context.Context, repo api.RepoNa
 	return api.CommitID(res.GetCommitSha()), nil
 }
 
+func (c *clientImplementor) RevAtTime(ctx context.Context, repo api.RepoName, spec string, date time.Time) (_ api.CommitID, ok bool, err error) {
```
eseliger, line 940: Can we get a quick test of this method? 

Here's a good example of one: https://sourcegraph.com/github.com/sourcegraph/sourcegraph@main/-/blob/internal/gitserver/commands_test.go?L2507%3A1-2554%3A1

cmd/gitserver/internal/server_grpc.go
eseliger: haha I should give up here at this point 😆 feel free to leave it in, I'll do a pass of this file once we're done with migration.

cmd/gitserver/internal/git/iface.go
```diff
@@ -60,6 +61,12 @@ type GitBackend interface {
 	// If the revspec can not be resolved to a commit, a RevisionNotFoundError is returned.
 	ResolveRevision(ctx context.Context, revspec string) (api.CommitID, error)
 
+	// RevAtTime returns the OID of the nearest ancestor of `spec` that has a
+	// commit time before the given time. To simplify the logic, it only
+	// follows the first parent of merge commits to linearize the commit
+	// history. The intent is to return the state of a branch at a given time.
```
eseliger, line 67:
```suggestion
	// history. The intent is to return the state of a branch at a given time.
	// If revspec is not found, a RevisionNotFoundError is returned.
```

cmd/gitserver/internal/server_grpc.go
```diff
@@ -1185,6 +1185,59 @@ func (gs *grpcServer) ResolveRevision(ctx context.Context, req *proto.ResolveRev
...
+
+	if err := gs.checkRepoExists(ctx, repoName); err != nil {
+		return nil, err
+	}
+
+	revspec := string(req.GetRevSpec())
+
+	backend := gs.getBackendFunc(repoDir, repoName)
+
+	// First, try to resolve the revspec so we can return a useful RevisionNotFound error
```
eseliger, line 1210: I think this should be done in the gitcli layer, it should own returning nice errors. So the RevAtTime method should be documented to return this well-known error type and do this check. 
That will also make sure that any git corruption errors are logged properly (currently missing for the other gitcli invocation)

cmd/gitserver/internal/server_grpc.go
```diff
@@ -1185,6 +1185,59 @@ func (gs *grpcServer) ResolveRevision(ctx context.Context, req *proto.ResolveRev
 	}, nil
 }
 
+func (gs *grpcServer) RevAtTime(ctx context.Context, req *proto.RevAtTimeRequest) (*proto.RevAtTimeResponse, error) {
```
eseliger, line 1188: A test alá https://sourcegraph.com/github.com/sourcegraph/sourcegraph@main/-/blob/cmd/gitserver/internal/server_grpc_test.go?L255%3A1-330%3A1 would be appreciated :)

cmd/gitserver/internal/server_grpc.go
```diff
@@ -1185,6 +1185,59 @@ func (gs *grpcServer) ResolveRevision(ctx context.Context, req *proto.ResolveRev
...
+	}
+
+	repoName := api.RepoName(req.GetRepoName())
+	repoDir := gs.fs.RepoDir(repoName)
+
+	if err := gs.checkRepoExists(ctx, repoName); err != nil {
+		return nil, err
+	}
+
+	revspec := string(req.GetRevSpec())
```
eseliger, line 1206: Is an empty revspec an allowed condition?

internal/gitserver/client.go
```diff
@@ -359,6 +359,16 @@ type Client interface {
...
 	ResolveRevision(ctx context.Context, repo api.RepoName, spec string, opt ResolveRevisionOptions) (api.CommitID, error)
 
+	// RevAtTime returns the OID of the nearest ancestor of `spec` that has a
+	// commit time before the given time. To simplify the logic, it only
+	// follows the first parent of merge commits to linearize the commit
+	// history. The intent is to return the state of a branch at a given time.
+	//
+	// If `spec` does not exist, an error will be returned. In the case
+	// no commit in the ancestry of `spec` before the time `t`, no error
+	// is returned, but the second return value `found` will be false.
```
eseliger, line 369: Usually we return a RevisionNotFoundError for the former case, can we do that and also document it here?

cmd/gitserver/internal/git/gitcli/clibackend.go
```diff
@@ -11,16 +13,18 @@ import (
...
 	return &gitCLIBackend{
-		logger:   logger,
-		rcf:      rcf,
-		dir:      dir,
-		repoName: repoName,
+		logger:         logger,
+		rcf:            rcf,
+		dir:            dir,
+		repoName:       repoName,
+		revAtTimeCache: globalRevAtTimeCache,
```
eseliger, line 20: Idea: 

This cache seems very application specific - maybe it is better suited at the caller level which'll let you even save on the request entirely. Ofc that is less federated, but most instances don't restart that often and don't have that many replicas.

cmd/gitserver/internal/git/gitcli/revattime.go
```diff
@@ -0,0 +1,54 @@
...
+		return entry, nil
+	}
+
+	r, err := g.NewCommand(ctx, WithArguments(
+		"log",
+		"--format=format:%H", // only hash
+		"--first-parent",     // linearize history
+		fmt.Sprintf("--before=%d", t.Unix()),
+		"--max-count=1", // only one commit
+		spec,
```
eseliger, line 40: I think you'll need a 

```
if err := checkSpecArgSafety(string(spec)); err != nil {
		return nil, err
	}
```

here, otherwise this might be prone to argument injection

cmd/gitserver/internal/git/gitcli/revattime.go
```diff
@@ -0,0 +1,54 @@
...
+	))
+	if err != nil {
+		return "", err
+	}
+
```
eseliger, line 45: make sure to close r after here, otherwise we'll keep a process handle dangling

cmd/gitserver/internal/git/gitcli/revattime_test.go
eseliger, line 13: A test for 

- Invalid rev spec -> RevisionNotFoundError
- Not found rev spec -> RevisionNotFoundError

would be great!

cmd/gitserver/internal/git/iface.go
eseliger: Please document that it returns empty string for "no earlier commit" :)

internal/gitserver/v1/gitserver.proto
eseliger: and that it returns empty hash for "no earlier commit"

eseliger (approved):
Looks good, thanks for treading through the new gitserver API stuff and adding it in the new format!

I left a bunch of comments on the gitserver stuff, feel free to tell me to shut up and just merge it, just some suggestions on making sure we maintain a high quality API surface :)

internal/gitserver/v1/gitserver.proto
camdencheek: Hmm. I'm using `buf format -w`, which I thought was the default?

cmd/gitserver/internal/git/gitcli/revattime.go
camdencheek: Oh, I was thinking that since I was doing `io.ReadAll`, which ensures reading to EOF, a `Close()` was unnecessary. I guess it doesn't hurt though, and decouples creating the reader from consuming the reader.

internal/gitserver/v1/gitserver.proto
eseliger: I'm not sure how we format hoenstly, seems like CI doesn't enforce any standard here 🤔

cmd/gitserver/internal/git/gitcli/revattime.go
eseliger: The close also cancels a context cancel func, so unless `io.ReadAll` does a `if r, ok := r.(Closer); ok { r.Close() }` I _think_ it's still necessary 🤔 would need to re-read the code to be sure

cmd/gitserver/internal/git/gitcli/revattime.go
camdencheek: Ooh, good catch, thank you!

cmd/gitserver/internal/server_grpc.go
camdencheek: Moving resolution into the CLI backend made it more clear this accepts all the same things that `ResolveRevision` accepts (including an empty revspec). Will also document this

cmd/gitserver/internal/server_grpc.go
camdencheek: Done!

internal/gitserver/commands.go
camdencheek: Done!

internal/gitserver/client.go
camdencheek: Done!

cmd/gitserver/internal/git/iface.go
camdencheek: Done!

cmd/gitserver/internal/server_grpc.go
camdencheek: Done!

cmd/gitserver/internal/git/gitcli/clibackend.go
camdencheek: I don't think I'm going to move this because, for consistency, we want to cache based on SHA and not revspec, otherwise a query like `rev:at.time(2021-01-01, HEAD)` will always be a cache hit even though `HEAD` might change. Since we do the resolution in gitserver, I don't think it makes sense to pull it out into the client because then it would require an extra roundtrip to resolve the revision before making the `RevAtTime` request

eseliger: gotcha, makes sense 👍

cmd/gitserver/internal/git/gitcli/revattime_test.go
camdencheek: What's the difference between an invalid rev spec and an rev not found error?

eseliger: uh .. that was worded poorly 

not found _ref_ and invalid OID is what I meant: 

```
➜  sourcegraph git: git log fd71cf587cce30bc82f083cd7dfda8b0562c4665
fatal: bad object fd71cf587cce30bc82f083cd7dfda8b0562c4665
➜  sourcegraph git: git log THISISNOTABRANCH
fatal: ambiguous argument 'THISISNOTABRANCH': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

Interestingly, git returns different errors when a 40 char OID is given, vs any other arbitrarily sized revspec :|

camdencheek: Oh, interesting! TIL. Will add

PR #61514, opened 2024-04-02
Add maintenance APIs to Cody Gateway

chrsmith:
This PR adds a couple of new HTTP endpoints to Cody Gateway, one for listing any flagged prompts that are temporarily stored in Redis, and another for getting the contents of one of those prompts. This is to make it easier to land https://github.com/sourcegraph/abuse-ban-bot/pull/12, as well as inch us towards a SAMS-aware Cody Gateway.

# Next Steps

Things to do before merging this PR:

- [x] Write some unit tests to confirm everything is on the up and up.

Things to do after merging this PR:

- [ ] Register the "official" Cody Gateway (prod) SAMS client, and set the new env vars in Terraform. (I wasn't planning on adding any SAMS client credentials to the Cody Gateway dev environment, because any SAMS integrations with Cody Gateway only apply to Cody PLG for the near term. And registering another SAMS client in the SAMS-dev environment that isn't used would just be a hassle.)

After that:

- As part of the transition to making the SSC backend the source of truth for Cody Pro subscription data and overrides, I was thinking I'd add a few more of these "maintenance" handlers, such as:

```
GET   /maintenance/user/{samsExternalID}/rate-limits
PATCH /maintenance/user/{samsExternalID}/rate-limits
POST  /maintenance/user/{samsExternalID}/rate-limits/refresh
```

And then take on the refactoring to update the way we cache per-user rate limits in Redis to be keyed off of the PLG user's SAMS account external ID rather than Sourcegraph.com User ID... but that'll take some careful thought about how we gracefully cut over to that.

## Test plan

Unit tests for `NewMaintenanceHandler` are coming, promise! I'll add them before merging. But I wanted to get this out for review while colleges in the Eastern Hemisphere are at work.


vdavid:
@chrsmith 

> And then take on the refactoring to update the way we cache per-user rate limits in Redis to be keyed off of the PLG user's SAMS account external ID rather than Sourcegraph.com User ID... but that'll take some careful thought about how we gracefully cut over to that.

Btw, @thenamankumar and I went over that, and IIRC, the takeaway was that it'd _just work_: we could just start sending the SAMS ID instead of the dotcom user ID, Cody Gateway would start using those IDs as keys in Redis, and we're good to go. No migration is required because it'll only be a bunch of cache misses and stores, and all dotcom ID-based keys would expire after 30 days.

vdavid (approved):
Nice! This should be pretty helpful. Looks like it has all the parts we need, I didn't find anything missing.

internal/sams/auth.go
```diff
@@ -55,8 +55,7 @@ func (a *Authenticator) RequireScopes(requiredScopes []Scope, next http.Handler)
 					"attempt to authenticate using SAMS token without required scope",
 					log.Strings("gotScopes", gotScopes),
 					log.String("requiredScope", string(requiredScope)))
-				w.WriteHeader(http.StatusForbidden)
-				w.Write([]byte("missing required scope"))
+				http.Error(w, "Forbidden: Missing required scope", http.StatusForbidden)
```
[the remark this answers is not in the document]
chrsmith: Here we aren't relying on the built-in `http.StatusText` function to get the message to display in the response body. (Since we are providing a more detailed, custom message.) So there is less to gain from the "const value is a shorter identifier" approach. (Which again, is of dubious merit at best.)

cmd/cody-gateway/internal/httpapi/maintenance.go
```diff
@@ -0,0 +1,193 @@
...
+	"github.com/sourcegraph/sourcegraph/internal/redispool"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+
+	"github.com/sourcegraph/sourcegraph/internal/sams"
+)
+
+// NewMaintenanceHandler registers maintenance-related endpoints. These are APIs for
+// other services to call in order to inspect or update Cody Gateway internals.
+//
+// The registered endpoints require a SAMS authorization token, and Cody Gateway scopes.
```
rafax, line 25:
```suggestion
// The registered endpoints require a SAMS authentication token, and Cody Gateway scopes.
```

cmd/cody-gateway/internal/httpapi/maintenance.go
```diff
@@ -0,0 +1,193 @@
...
+			Handle(url, wrappedHandler).
+			Methods(method)
+	}
+
+	// Register the specific API endpoints.
+	const (
+		GET                    = http.MethodGet
+		scopeFlaggedPromptRead = "cody_gateway::flaggedprompts::read"
+	)
+	registerMaintenanceHandler(GET, "/flagged-requests", mHandlers.ListFlaggedPrompts, scopeFlaggedPromptRead)
```
rafax, line 77: nit: from the POV of a reader of this code, http.MethodGet is less exciting than a local const, so I'd actually be in favour on using the longer form here

cmd/cody-gateway/internal/httpapi/maintenance.go
```diff
@@ -0,0 +1,193 @@
...
+	registerMaintenanceHandler(GET, "/flagged-requests/{flaggedPromptKey}", mHandlers.GetFlaggedPrompt, scopeFlaggedPromptRead)
+
+	http.NewServeMux()
+
+	// Return the http.Handler. Yes, this is a horrible design wart, since we are using an
+	// "HTTP middleware" pattern, but not actually registering middleware. To address this,
+	// we'd need to rework Cody Gateway's httpapi.go and friends to pass around the mux.Router,
```
rafax, line 84: 😢

cmd/cody-gateway/internal/httpapi/maintenance.go
rafax, line 77: Do you expect that clients will call this, but not the prompt-scoped variant below? 

Seeing that this is a maintenance API we expose for our clients, I'd be in favour of returning both keys and values from this call (with  1 `KEYS` + N pipelined `GET` commands), instead of returning a list and having the client make N HTTP calls that in turn make N (not pipelined) `GET` calls)

chrsmith: Done. I'm mostly still recovering from PTSD induced by maintaining the massive REST API from something I worked on in the past.

After we had all of the authentication requirements, config settings, etc. we maintained a giant list of "registerHandler" type function calls here. And saving every possible character was important since it was already so crazy-wide.

But yes, `http.MethodGet` is cleaner on several dimensions.

cmd/cody-gateway/internal/httpapi/maintenance.go
chrsmith: > 😢

That's not criticism or anything, mostly just me trying to telegraph what I think a reasonable refactoring would be if we wanted to those changes.

Suppose down the road we did want to revisit how we register various HTTP endpoints from Cody Gateway, do you agree that would be a reasonable approach? Or is there some other pattern you would prefer? (Or would you just keep it as-is, using the `http.Handler` chaining we have today?)

```go
// Example
root := mux.NewRouter()

maintenanceRouter := mux.PathPrefix("/maintenance").Subrounter()
maintenanceHandlers.AttachToRouter(maintenanceRouter)

v1APIRouter := mux.PathPrefix("/v1").Subrounter()
apiHandlers.AttachToRouter(v1APIRouter)
```

cmd/cody-gateway/internal/httpapi/maintenance.go
chrsmith: I'm inclined to leave it as is, exposing a simpler API. But if this leads to problems (e.g. too slow, and/or too chatty) then I wouldn't be opposed to changing it.

PR #61522, opened 2024-04-02
[gitserver] ArchiveReader correctly handles paths with spaces

### LINKED ISSUE #61518: Gitserver returns an error for files with a space in the name
https://sourcegraph.slack.com/archives/C05EMJM2SLR/p1712046480031329

pjlast:
Closes #61518 

We used `bytes.Fields` to parse the output from the `git` command, but `bytes.Fields` splits on all whitespace. This leads to erroneous file paths when the file paths have spaces in them, since we'll only be using the final path segment.

This PR passes the `--name-only` flag to the `git ls-tree` command so that we don't have to do any parsing to get the path.

## Test plan

Unit test added.

### DIFF
--- a/cmd/gitserver/internal/git/gitcli/archivereader.go
+++ b/cmd/gitserver/internal/git/gitcli/archivereader.go
@@ -78,7 +78,13 @@ func (g *gitCLIBackend) verifyPaths(ctx context.Context, treeish string, paths [
 		if len(p) == 0 {
 			continue
 		}
-		pathSegments := bytes.Fields(p)
+
+		// Split into path segments by skipping only tab characters, since
+		// files/dirs can start with/contain spaces.
+		pathSegments := bytes.FieldsFunc(p, func(r rune) bool {
+			return r == rune('\t')
+		})
+
 		fileSet.Add(string(pathSegments[len(pathSegments)-1]))
 	}
 
--- a/cmd/gitserver/internal/git/gitcli/archivereader.go
+++ b/cmd/gitserver/internal/git/gitcli/archivereader.go
@@ -44,7 +44,7 @@ func buildArchiveArgs(format git.ArchiveFormat, treeish string, paths []string) 
 func pathspecLiteral(s string) string { return ":(literal)" + s }
 
 func (g *gitCLIBackend) verifyPaths(ctx context.Context, treeish string, paths []string) error {
-	args := []string{"ls-tree", treeish, "--"}
+	args := []string{"ls-tree", "--name-only", treeish, "--"}
 	args = append(args, paths...)
 	r, err := g.NewCommand(ctx, WithArguments(args...))
 	if err != nil {
@@ -75,17 +75,7 @@ func (g *gitCLIBackend) verifyPaths(ctx context.Context, treeish string, paths [
 	gotPaths := bytes.Split(bytes.TrimSpace(stdout), []byte("\n"))
 	fileSet := collections.NewSet[string]()
 	for _, p := range gotPaths {
-		if len(p) == 0 {
-			continue
-		}
-
-		// Split into path segments by skipping only tab characters, since
-		// files/dirs can start with/contain spaces.
-		pathSegments := bytes.FieldsFunc(p, func(r rune) bool {
-			return r == rune('\t')
-		})
-
-		fileSet.Add(string(pathSegments[len(pathSegments)-1]))
+		fileSet.Add(string(p))
 	}
 
 	pathsSet := collections.NewSet[string]()
--- a/cmd/gitserver/internal/git/gitcli/archivereader.go
+++ b/cmd/gitserver/internal/git/gitcli/archivereader.go
@@ -1,6 +1,7 @@
 package gitcli
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"io"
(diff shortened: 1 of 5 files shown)


cmd/gitserver/internal/git/gitcli/archivereader.go
```diff
@@ -78,7 +78,13 @@ func (g *gitCLIBackend) verifyPaths(ctx context.Context, treeish string, paths [
 		if len(p) == 0 {
 			continue
 		}
-		pathSegments := bytes.Fields(p)
+
+		// Split into path segments by skipping only tab characters, since
+		// files/dirs can start with/contain spaces.
+		pathSegments := bytes.FieldsFunc(p, func(r rune) bool {
+			return r == rune('\t')
```
keegancsmith: i believe tabs are allowed in file names. Your best bet is the null character, which we should be using in other places. This is the same thing used in robust shell scripts when piping between commands.

pjlast: @keegancsmith I went for a different approach by passing `--name-only` to `git ls-tree` so that we don't have to do any parsing at all.

cmd/gitserver/internal/git/gitcli/archivereader.go
```diff
@@ -44,7 +44,7 @@ func buildArchiveArgs(format git.ArchiveFormat, treeish string, paths []string)
 func pathspecLiteral(s string) string { return ":(literal)" + s }
 
 func (g *gitCLIBackend) verifyPaths(ctx context.Context, treeish string, paths []string) error {
-	args := []string{"ls-tree", treeish, "--"}
+	args := []string{"ls-tree", "--name-only", treeish, "--"}
```
keegancsmith, line 47: git normally supports a `-z` option to null seperate items. It often is quite important to use because you may get surprising behaviour like git quoting values in wierd file names.

eseliger: @pjlast can we incorporate that?

cmd/gitserver/internal/git/gitcli/archivereader_test.go
```diff
@@ -123,6 +125,22 @@ func TestGitCLIBackend_ArchiveReader(t *testing.T) {
...
 
+	t.Run("read file with space in name", func(t *testing.T) {
+		r, err := backend.ArchiveReader(ctx, "tar", string(commitID), []string{"dir1/ file3", "dir1/file with spaces"})
+		require.NoError(t, err)
+		t.Cleanup(func() { r.Close() })
+		tr := tar.NewReader(r)
+		contents := readFileContentsFromTar(t, tr, "dir1/ file3")
+		require.Equal(t, "ijkl\n", contents)
+
+		r, err = backend.ArchiveReader(ctx, "tar", string(commitID), []string{"dir1/ file3", "dir1/file with spaces"})
```
eseliger, line 136: Maybe a test for non ascii filenames as well?

cmd/gitserver/internal/git/gitcli/archivereader.go
```diff
@@ -43,16 +44,43 @@ func buildArchiveArgs(format git.ArchiveFormat, treeish string, paths []string)
 // See: https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-literal
 func pathspecLiteral(s string) string { return ":(literal)" + s }
 
+// scanNullLines is a split function for a [Scanner] that returns each null-terminated
+// line of text, stripped of any trailing end-of-line marker. The returned line may
+// be empty. The end-of-line marker is a null character.
+// The last non-empty line of input will be returned even if it has no
+// null character.
+func scanNullLines(data []byte, atEOF bool) (advance int, token []byte, err error) {
```
eseliger, line 52: Wonder if we could move this to `internal/byteutils` or so, seems fairly generic and most likely we already have 3-19 implementations of this somewhere 😆

pjlast: Will check!

cmd/gitserver/internal/git/gitcli/archivereader.go
```diff
@@ -44,15 +46,21 @@ func buildArchiveArgs(format git.ArchiveFormat, treeish string, paths []string)
...
 	args = append(args, paths...)
 	r, err := g.NewCommand(ctx, WithArguments(args...))
 	if err != nil {
 		return err
 	}
 	defer r.Close()
 
-	stdout, err := io.ReadAll(r)
+	scanner := bufio.NewScanner(r)
+	scanner.Split(byteutils.ScanNullLines)
```
keegancsmith, line 58: very minor perf point here. But using io.ReadAll, converting that to a string and then splitting on `\z` is much more performant since you end up doing 1 big allocation for the string data, then an allocation per "pointer" part of the string.

To be even more of a funny dude, does git not output the paths in the same order you insert them? You could probably totally avoid the collection.

eseliger: (1) shouldn't it reuse the buffer in the scanner?

keegancsmith: (1) is using a buffer than gets written over as you progress. The point I am making isn't totally worth it since I'm unsure how much of the GC thrash on gitserver comes from this function. But if git doesn't return paths in order you could probably write this function so that it allocates far less objects (but not necessarily less total space). The fun insight is you don't return any of the output from git so its all short lived. In the case you return some output in an error you can copy from the buffer into a string at that point.

Not gonna go into too much detail in this comment, but maybe this is a fun random hack hour to do. Rice this function w.r.t. allocations.

PR #61523, opened 2024-04-02
Reapply "gitserver: Introduce FS layer to encapsulate repo name conversions (#60627)" (#61487)

eseliger:
This reverts commit 0e8cbca5690e3677d47f44b8b6e0f8374a716e37.

After the initial version didn't work for case-sensitive filesystems, this is the next attempt at shipping this.

## Test plan

Tests and verified that the previous issue is resolved.


pjlast (approved):
🤞

Issue #61353, opened 2024-03-23
Code Insights export to CSV failed to export with data error: extended protocol limited to 65535 parameters


Galindo-J:
- **Sourcegraph version:** Observed on v5.3.2 & v5.2.4
- **Platform information:** 

#### Steps to reproduce:

1. Create an Insight with a user account using a query that runs over a large number of repos that the user does not have access to. 
2. Attempt to export the results to CSV.

#### Expected behavior:
Export is successful 

#### Actual behavior:
Export errors with the following
`failed to export data: failed to fetch all data for insight: fetching archived code insights data: extended protocol limited to 65535 parameters`

If you would like immediate help on this, please email support@sourcegraph.com (you can still create the issue, but there are no [SLAs](https://sourcegraph.com/support/) on issues like there are for support requests).

camdencheek:
[Slack context](https://sourcegraph.slack.com/archives/C05EA9KQUTA/p1710346770335499?thread_ts=1710201929.953689&cid=C05EA9KQUTA)

bahrmichael:
Recap from the Slack conversation:

There's this part in the code that lists all the repositories that a user does NOT have access to: https://sourcegraph.com/github.com/sourcegraph/sourcegraph@153e84aed28bd258fa1073e427a3757897e86365/-/blob/internal/insights/store/store.go?L900-906

We then send that list into a follow-up query. If there are too many repositories that you don't have access to, then 💥 .

A possible fix could be to remove this roundtrip through the application layer, and keep the evaluation in the database. To do that we need to bring the conditions of https://sourcegraph.com/github.com/sourcegraph/sourcegraph@153e84aed28bd258fa1073e427a3757897e86365/-/blob/internal/database/repos_perm.go?L84:6-84:21 into https://sourcegraph.com/github.com/sourcegraph/sourcegraph@153e84aed28bd258fa1073e427a3757897e86365/-/blob/internal/insights/store/store.go?L982.

bahrmichael:
Questions:

- Do we have tests that populate the database (preferably a real postgresql engine that can run and verify our query) with 10k+ records, or tests that I could base such one on? I would prefer to add a test that can prevent this bug from reappearing, but I'm afraid that this may take very long to run.

peterguy:
[...]

PR #61525, opened 2024-04-02
fix: allow more than 65535 repos in deny list

bahrmichael:
Closes https://github.com/sourcegraph/sourcegraph/issues/61353

This PR fixes an issue where too many user_repo_permissions would lead to an SQL error. We fix this by not loading all denied repo ids through the application layer back into the database, and instead loading the denied repo ids within the `WHERE` statement of the insights query.

TODO: 
- handle other call to GetUnauthorizedRepoIDsQuery
- create new ticket for early abort when we have more than 65535 predicates
- measure performance impact of avoiding roundtrip

## Test plan

- We didn't touch how permissions work, so this should be Covered by existing tests
- Added additional tests for the code that uses the resulting query


bahrmichael:
There are currently a lot of compilation issues, mostly from other packages. I'll focus on ruling out problems from the authz variations before tackling the inter-package refactoring.

bahrmichael:
Making sure that we don't negatively impact security may take a bit because there are nearly 1k lines in `repos_perm_test.go` which tests for correct auth.

internal/database/dbtest/dbtest.go
```diff
@@ -93,7 +93,7 @@ func getSchemaAtRev(t testing.TB, name, rev string) *schemas.Schema {
 // NewInsightsDB returns a connection to a clean, new temporary testing database with
 // the same schema as Sourcegraph's CodeInsights production Postgres database.
 func NewInsightsDB(logger log.Logger, t testing.TB) *sql.DB {
-	return newDB(logger, t, "insights", schemas.CodeInsights)
+	return newDB(logger, t, "insights", schemas.CodeInsights, schemas.Frontend)
```
bahrmichael, line 96: Todo for myself: Test if we can revert this change again.

bahrmichael: Update: We need this line so that the authz query can run (it access the repo and user_repo_permissions tables).

camdencheek: OH. I suddenly understand why we're doing it the way we're doing it now.

The Code Insights DB and the Frontend DB are two separate databases. A query cannot hit both of them. I'm pretty sure now this is why we queried the Frontend DB first, then used the results against the Code Insights DB.

So this means my comments about just using the query conditons directly were wrong. Sorry 😬

...so back to your ideas. Pagination might be tricky because this is a negative query. Thoughts?

bahrmichael: Couple ideas come to mind:
- Documenting a service limitation that customers with more than 65k restricted repos cannot run code insights queries
- Introducing a different auth and role mechanism to enable this with less predicates (sounds like a lot of work to me, including migration of stuff)
- Introducing a replication of permissions from the frontend to the code insights db
- Instead of sending predicates to the code insights db, filter out restricted repos in the application layer (sounds like the most pragmatic approach, will work on that today)

bahrmichael: I've spun up a new PR with the last idea: https://github.com/sourcegraph/sourcegraph/pull/61580

PR #61526, opened 2024-04-02
[perf] Do less string allocations when performing git blame

pjlast:
Improves our git blame performance by using `[]byte` operations instead of allocating strings.

Some numbers:

Using the `string` functions:
```
BenchmarkBlameBytes-10                                                    291399              4192 ns/op            7487 B/op         68 allocs/op
```

Using the `[]byte` functions:
```
BenchmarkBlameBytes-10                                                    344814              3397 ns/op            6783 B/op         45 allocs/op
```

And then anecdotal spot testing on `schema/schema.go`:

Using the `string` functions:
image

Using the `[]byte` functions:
image

## Test plan

Manual tests and benchmark

### DIFF
--- a/cmd/gitserver/internal/git/gitcli/blame.go
+++ b/cmd/gitserver/internal/git/gitcli/blame.go
@@ -2,6 +2,7 @@ package gitcli
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"fmt"
 	"io"
@@ -96,7 +97,7 @@ func (br *blameHunkReader) Read() (_ *gitdomain.Hunk, err error) {
 		}
 
 		// Read line from git blame, in porcelain format
-		line := br.sc.Text()
+		line := br.sc.Bytes()
 		annotation, fields := splitLine(line)
 
 		// On the first read, we have no hunk and the first thing we read is an entry.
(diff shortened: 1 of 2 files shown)


cmd/gitserver/internal/git/gitcli/blame.go
```diff
@@ -142,17 +142,17 @@ func (br *blameHunkReader) Close() error {
...
 		return nil, errors.Errorf("Expected at least 4 parts to hunkHeader, but got: '%s %s'", rev, content)
 	}
 
-	resultLine, err := strconv.Atoi(fields[1])
+	resultLine, err := strconv.Atoi(string(fields[1]))
 	if err != nil {
 		return nil, err
 	}
-	numLines, _ := strconv.Atoi(fields[2])
+	numLines, err := strconv.Atoi(string(fields[2]))
```
eseliger, line 155: heh oops 🙃

eseliger (approved):

cmd/gitserver/internal/git/gitcli/blame_test.go
```diff
@@ -430,6 +430,24 @@ func TestBlameHunkReader(t *testing.T) {
...
+		defer reader.Close()
+
+		for {
+			_, err := reader.Read()
+			if errors.Is(err, io.EOF) {
+				break
+			} else if err != nil {
+				b.Fatalf("blameHunkReader.Read failed: %s", err)
+			}
+		}
```
camdencheek, line 446: `defer` in a for loop can actually be pretty expensive, and seems to be skewing the results a bit. Would recommend closing manually. I suspect you'll get an even better improvement 🙂  

```suggestion

		for {
			_, err := reader.Read()
			if errors.Is(err, io.EOF) {
				break
			} else if err != nil {
				b.Fatalf("blameHunkReader.Read failed: %s", err)
			}
		}
		reader.Close()
```

cmd/gitserver/internal/git/gitcli/blame.go
```diff
@@ -166,32 +166,32 @@ func parseEntry(rev string, content string) (*gitdomain.Hunk, error) {
...
 	case "filename":
-		hunk.Filename = content
+		hunk.Filename = string(content)
 	case "previous":
 	case "boundary":
 	default:
 		// If it doesn't look like an entry, it's probably an unhandled git blame
 		// annotation.
-		if len(annotation) != 40 && len(strings.Split(content, " ")) != 3 {
+		if len(annotation) != 40 && len(bytes.Split(content, []byte(" "))) != 3 {
```
camdencheek, line 194: we could probably drop another allocation by using `bytes.Count` here instead of `len(bytes.Split(...))`

camdencheek (approved):
Nice!

cmd/gitserver/internal/git/gitcli/blame.go
camdencheek: Oh, I guess we expect this to be a rare branch anyways

PR #61527, opened 2024-04-02
svelte/ci: Run `svelte-check` in CI

fkling:
This commit updates the bazel build file to execute `svelte-check` in CI.

To make this work correctly we need to run additional commands to generate all the files that are needed by the code. Specifically that's `svelte-kit sync` and the graphql code/type generation. The code was refactored to make that work.

This commit also fixes issues detected by `svelte-check`.

IMPORTANT
Running `svelte-check` in CI differs in two ways from running it locally:

1. Locally `svelte-check` will also check all local `@sourcegraph/*`, because they symlink the orginal source code. In CI/Bazel those dependencies are already built and therefore are not checked by Bazel. However, this is what we want because the dependencies do not always conform to the rules set by `svelte-check`.
2. Source files that are not required for the production build have been excluded and are therefore not checked. That includes Storybook story files, playwright and vitest files and `src/testing/*`.

FWIW I'm happy to add story and tests files as well if others would prefer that.

## Test plan

```
bazel test //client/web-sveltekit:svelte-check
```

```
pnpm build
```

CI

### DIFF
--- a/client/web-sveltekit/BUILD.bazel
+++ b/client/web-sveltekit/BUILD.bazel
@@ -96,6 +96,7 @@ BUILD_DEPS = [
     ":node_modules/hotkeys-js",
     ":node_modules/prismjs",
     ":node_modules/sass",
+    ":node_modules/signale",
     ":node_modules/svelte",
     ":node_modules/ts-key-enum",
     ":node_modules/vite",
@@ -147,11 +148,19 @@ TESTS = glob([
 ])
 
 TEST_BUILD_DEPS = [
+    ":generate-graphql-types",
     ":node_modules/vitest",
-    ":node_modules/signale",
     ":node_modules/@testing-library/svelte",
     ":node_modules/@testing-library/user-event",
-]
+] + glob(
+    [
+        "src/testing/*",
+    ],
+    [
+        # Already inluded in TESTS
+        "src/testing/graphql-mocking.test.ts",
+    ],
+)
 
 vitest_test(
     name = "unit_tests",
--- a/client/web-sveltekit/BUILD.bazel
+++ b/client/web-sveltekit/BUILD.bazel
@@ -170,22 +170,6 @@ vitest_test(
     with_vitest_config = False,
 )
 
-# Input for graphql-codegen
-filegroup(
-    name = "graphql_sources",
-    srcs = glob(
-        [
-            "src/**/*.gql",
-            "src/**/*.ts",
-        ],
-        [
-            "src/lib/graphql-*.ts",
-            "src/testing/graphql-type-mocks.ts",
-            "src/**/*.gql.ts",
-        ],
-    ),
-)
-
 # Create binary that executes graphql-codegen
 ts_binary(
     name = "graphql_codegen_bin",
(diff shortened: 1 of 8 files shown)


client/web-sveltekit/dev/vite-graphql-codegen.ts
```diff
@@ -1,105 +1,104 @@
...
-            // The near-operation-file preset generates this import for every file, even if it's not used. This
-            // generally is not a problem, but the issue is reported by `pnpm check`.
-            // See https://github.com/dotansimha/graphql-code-generator/issues/4900
-            beforeOneFileWrite(_path, content) {
-                if (/^import type \* as Types from/m.test(content) && !/Types(\[|\.)/.test(content)) {
-                    return content.replace(/^import type \* as Types from .+$/m, '').trimStart()
-                }
-                return content
-            },
+const codgegenConfig: CodegenConfig = {
```
fkling, line 4: This was moved outside the plugin so that it can be reused by the standlone function. The config doesn't depend on any runtime parameters.

client/web-sveltekit/dev/vite-graphql-codegen.ts
```diff
@@ -1,105 +1,104 @@
-import { dirname } from 'node:path'
-import { fileURLToPath } from 'node:url'
-
 import { generate, type CodegenConfig } from '@graphql-codegen/cli'
 import type { Plugin } from 'vite'
 
-const __dirname = dirname(fileURLToPath(import.meta.url))
```
fkling, line 7: For some reason `__dirname` was already defined when running in Bazel, causing an error. We can just use `./dev/...` below. That means the codegen always has to run from the package root but I assume that's OK.

client/web-sveltekit/src/lib/wildcard/Button.svelte
```diff
@@ -6,12 +6,16 @@
 
     import { type BUTTON_DISPLAY, type BUTTON_SIZES, type BUTTON_VARIANTS, getButtonClassName } from './Button'
 
-    interface $$Props extends HTMLButtonAttributes {
+    type $$Props = {
```
fkling, line 9: Using `type` instead of `interface` didn't solve the issue with `data-*` not being recognized but it removed another error I got in neovim.

client/web-sveltekit/src/lib/wildcard/menu/Menu.svelte
```diff
@@ -34,7 +32,7 @@
     } = createDropdownMenu({
         open,
     })
-    setContext({ item, trigger, separator, builders, $open })
```
fkling, line 37: `$open` was wrong but also not used.

client/web-sveltekit/BUILD.bazel
```diff
@@ -138,11 +148,19 @@ TESTS = glob([
 ])
 
 TEST_BUILD_DEPS = [
+    ":generate-graphql-types",
```
fkling, line 151: No idea why this needs to be a dependency now. Without it one tests files because it (or a dependency) imports `*.gql` file.

camdencheek (approved):
WOOT! Thank you! I'm very excited for this 😄

camdencheek:
> svelte-check issues should be blocking

Well, right now, I think `svelte-check` is the only thing doing type checking right? I do think our tests should at least go through a type-check pass, but I don't really care too much if they don't pass all the lints.

In any case, it's fine to skip them for now. This PR is already a huge improvement.

bahrmichael (approved):

PR #61530, opened 2024-04-02
wolfi: add debugging and use switch for new branch

burmudar:
Wofli is currently timing out see [here](https://buildkite.com/sourcegraph/sourcegraph/builds/267161#018e9f87-33ff-457f-aac8-38fb8b41bc51) so we add some things to see WHY it is doing that
* Print out the commands
* Use switch instead of checkout since checkout can possibly hang when the repo is in a detached state
* swap out `:` for `true` since it is slightly more readable
## Test plan
CI


willdollman (approved):
Thanks for investigating!

PR #61532, opened 2024-04-02
auth/oidc: attempt to revokve SAMS session when sign out

unknwon:
Closes https://github.com/sourcegraph/sourcegraph-accounts/issues/163

Needing to sign out again on SAMS (accounts.sourcegraph.com) in addition to Sourcegraph.com has been a common UX confusion for PLG users.

This PR makes the Sourcegraph.com to attempting revoke the SAMS user session when a user perform a sign-out action on Sourcegraph.com. 

### As for the global var of the singleton...

Since this is _very dotcom-specific_ thing, I also do not feel great about prop drilling a `nil` SAMS client from the very top of the call chain, as well as populating the call chain. Open to ideas ... :D 

## Test plan

Booting up local SG in dotcom mode and SAMS instance.

https://github.com/sourcegraph/sourcegraph/assets/2946214/02b021cb-f4ff-43b4-85b0-7ed75c528406

## Before merge

- [x] Grant scopes to the dotcom SAMS client, `"sams::session::read", "sams::session::write"`


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/sourcegraph/sourcegraph
 
-go 1.22
+go 1.22.0
```
eseliger, line 3: unrelated?

go.mod
```diff
@@ -222,11 +222,11 @@ require (
...
 	go.uber.org/zap v1.26.0
-	golang.org/x/crypto v0.20.0
-	golang.org/x/net v0.21.0
-	golang.org/x/oauth2 v0.16.0
+	golang.org/x/crypto v0.21.0
+	golang.org/x/net v0.22.0
+	golang.org/x/oauth2 v0.18.0
 	golang.org/x/sync v0.6.0
-	golang.org/x/sys v0.17.0
+	golang.org/x/sys v0.18.0
```
eseliger, line 229: wonder if any of these are dangerous, i.e., have unintended side-effects?

cmd/frontend/internal/auth/openidconnect/session.go
```diff
@@ -39,20 +51,61 @@ func SignOut(w http.ResponseWriter, r *http.Request, sessionKey string, getProvi
...
 		return "", errors.WithMessage(err, "clearing OpenID Connect session data")
 	}
-	if data != nil {
-		p := getProvider(data.ID.ID)
-		if p == nil {
-			return "", errors.Errorf("unable to revoke token or end session for OpenID Connect because no provider %q exists", data.ID)
-		}
+	if data == nil {
+		return "", nil
+	}
```
eseliger, line 56: thx for early return

cmd/frontend/internal/auth/openidconnect/session.go
```diff
@@ -39,20 +51,61 @@ func SignOut(w http.ResponseWriter, r *http.Request, sessionKey string, getProvi
...
+			p.config.ClientSecret,
+			[]scopes.Scope{
+				"sams::session::read",
+				"sams::session::write",
+			},
+		)
+	})
+	if samsClientErr != nil {
+		return "", errors.Wrap(samsClientErr, "creating SAMS client")
+	}
```
eseliger, line 95: Is this expensive? If not, I'd err on the side of not caching this, for simplicity. Looking at the code, I don't see any heavy operations or caching happening here 🤔

cmd/frontend/internal/auth/openidconnect/session.go
```diff
@@ -39,20 +51,61 @@ func SignOut(w http.ResponseWriter, r *http.Request, sessionKey string, getProvi
...
+	if p == nil {
+		return "", errors.Errorf("unable to revoke token or end session for OpenID Connect because no provider %q exists", data.ID)
+	}
 
-		if p.oidc.RevocationEndpoint != "" {
-			if err := revokeToken(r.Context(), p, data.AccessToken, data.TokenType); err != nil {
-				return endSessionEndpoint, errors.WithMessage(err, "revoking OpenID Connect token")
-			}
+	endSessionEndpoint = p.oidc.EndSessionEndpoint
+	if p.oidc.RevocationEndpoint != "" {
```
eseliger, line 64: Just for my own entertainment, why could we not use the `revocationEndpoint` here vs relying on a magical cookie that's set with a wildcard plus a client with service account?

eseliger (approved):

cmd/frontend/internal/auth/openidconnect/session.go
unknwon: I asked myself this question too, so the conclusion so far is that the `revocationEndpoint` from OIDC-spec gives almost no security measures in terms of preventing abuse (i.e. anyone can fake/trick users to perform a sign-out on IdP, basically) 

Also see my assessment in https://github.com/sourcegraph/sourcegraph-accounts-sdk-go/pull/4#issuecomment-2013248142.

cmd/frontend/internal/auth/openidconnect/session.go
unknwon: It is primarily caching the access token (for an hour), but not expensive in the sense that we can get rate limited, or bring down SAMS, just logically wasteful and "bad behavior" 😂 I started with no singleton actually, maybe init new client on every sign-out is just fine. 🤔 Let's give others a spin see what they think.

go.mod
unknwon: `go mod tidy` insists on this change... 🤔

cmd/frontend/internal/auth/openidconnect/session.go
chrsmith: Having a trio of package-level global variables starts at "-30 points".

```go
var (
	samsClient     *sams.ClientV1
	samsClientOnce sync.Once
	samsClientErr  error
)
```

Creating a new SAMS client on every logout request is maybe "-20" points, but that's still -- using my HIGHLY ACCURATE AND PRECISE MATH -- like "10 points better".

The overhead here is that we need to (1) make a new API call to create a SAMS access token using the client credentials and (2) make another API request to actually revoke the session.

So in reality, it's just one more API call per signet than before... which does add latency and another thing that can unexpectedly fail. But having some package-level variables that may-or-may-not be initialized correctly, and should only be used within one unexpected function _looks wrong_.

So I'd just go with "new client every time", and if that causes a problem, do something a little more heavyweight, and/or stash a longer-lived SAMS client somewhere else.

chrsmith (approved):
As far as I understand how these things all work, this LGTM. 🚀

go.mod
evict: We are on 1.22.1, so I think we should change that? 🤔

evict (approved):
👍

go.mod
unknwon: Updated to 1.22.1!

go.mod
unknwon: Generally unless something is a show-stopper, the whole point of unit and integration tests is to ensure the CI should fail when we shouldn't upgrade right? 😁

unknwon:
Thanks for all the reviews! Merging as-is, happy to address post-merge feedback!

PR #61539, opened 2024-04-03
Add OpenAI models to supported chat models for enterprise customers

taras-yemets:
Adds GPT3.5, GPT4 Turbo, GPT 4 to the list of supported chat models for enterprise ([thread](https://sourcegraph.slack.com/archives/C012NU3PXA9/p1712103525033819?thread_ts=1712094412.441679&cid=C012NU3PXA9)).

## Test plan
- CI


vdavid (approved):
Looks good to me

arafatkatze (approved):
Makes sense as per the slack thread

chwarwick:
Just a note I don't think this will apply to Cloud customers, when a cloud customer is created the allowed models are set to only the models that are specified for chat & autocomplete.  Check the Cloud controller codebase 

cc @daxmc99

daxmc99:
Yes, by default cloud will set these models
https://sourcegraph.sourcegraph.com/github.com/sourcegraph/controller@main/-/blob/cmd/manager/api/v1/instanceinvariants_types.go?L313-324

And our automation will continually set them back unless configured in that instance's config.yaml to use a different model.

taras-yemets:
@daxmc99, the main goal of this PR was to make OpenAI models allowed by default. I made a similar change to the controller repo: https://github.com/sourcegraph/controller/pull/1463 ([Slack thread](https://sourcegraph.slack.com/archives/C03JR7S7KRP/p1712164602104869)).

PR #61545, opened 2024-04-03
ci: use build creator when the build is a release build

### LINKED ISSUE #61323: During release, BuildTracker should acknowledge correct author during failure
While testing an internal release, one of my builds resulted in a failure because the `--version` flag wasn't specified in the `sg` command bazel was executing. This resulted in BuildTracker notifying the #buildkite-main however, it specified the author of the last commit in the branch instead of the person who triggered the release.

CleanShot 2024-03-21 at 15 03 39@2x

This leads to noise. The build tracker should instead notify the person who triggered the release (be it internal or not)

burmudar:
Get the build creator when it is a release build and then try to fetch by name of the author

Closes  #61323

## Test plan
Unit Tests


dev/build-tracker/BUILD.bazel
```diff
@@ -3,7 +3,6 @@ load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
 load("//dev:oci_defs.bzl", "image_repository", "oci_image", "oci_push", "oci_tarball")
 load("@rules_pkg//:pkg.bzl", "pkg_tar")
 load("@container_structure_test//:defs.bzl", "container_structure_test")
-load("//dev:msp_delivery.bzl", "msp_delivery")
```
burmudar, line 6: buildifier complained

dev/build-tracker/build/build.go
```diff
@@ -104,20 +109,37 @@ func (b *Build) IsFinished() bool {
...
-func (b *Build) GetAuthorName() string {
-	if b.Author == nil {
-		return ""
+func (b *Build) IsReleaseBuild() bool {
+	// Release builds have two environment variables which distinguishes between internal / public releases
+	for _, key := range []string{"RELEASE_PUBLIC", "RELEASE_INTERNAL"} {
+		if _, ok := b.Env[key]; ok {
+			return true
+		}
 	}
```
BolajiOlajide, line 118: Do we instead need to check if either of these keys are `true` instead of checking for it's existence?

I ask because I wonder if there would ever be a scenario where we're running a normal CI job but with `RELEASE_PUBLIC` or `RELEASE_INTERNAL` set to false.

BolajiOlajide (approved):

dev/build-tracker/build/build.go
burmudar: Does hurt to check. I've added it

PR #61553, opened 2024-04-03
build-tracker: uncomment that which was commented

burmudar:
My merge commented this which it should not have
## Test plan
CI

### DIFF
--- a/dev/build-tracker/BUILD.bazel
+++ b/dev/build-tracker/BUILD.bazel
@@ -96,9 +96,9 @@ oci_push(
 )
 
 # automatic deployment not enabled for initial testing
-# msp_delivery(
-#     name = "msp_deploy",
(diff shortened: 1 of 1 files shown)


Strum355 (approved):

dev/build-tracker/BUILD.bazel
```diff
@@ -96,9 +96,9 @@ oci_push(
 )
 
 # automatic deployment not enabled for initial testing
```
Strum355, line 98: remove this line :P

Strum355 (changes requested):

dev/build-tracker/BUILD.bazel
burmudar: also have to add the loading of msp 😮‍💨

Strum355 (approved):

PR #61567, opened 2024-04-03
ci: make pushed images annotation collapsible

Strum355:
Shrinks down the output from the "Push xyz images" buildkite annotation!

image

image

## Test plan

See screenshots above for tested results

### DIFF
--- a/dev/ci/push_all.sh
+++ b/dev/ci/push_all.sh
@@ -35,22 +35,22 @@ function echo_append_annotation() {
     if [ "$arg" != "--tag" ]; then
       if [ "$formatted_tags" == "" ]; then
         # Do not insert a comma for the first element
-        formatted_tags="\`$arg\`"
+        formatted_tags="<code>$arg</code>"
       else
-        formatted_tags="${formatted_tags}, \`$arg\`"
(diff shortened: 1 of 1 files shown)


dev/ci/push_all.sh
```diff
@@ -146,15 +146,14 @@ fi
...
-echo -e "\n| Name | Registries | Tags |\n|---|---|---|" >> ./annotations/pushed_images.md
+echo -e "<details><summary>Click to expand table</summary><table>\n" >>./annotations/pushed_images.md
+echo -e "<tr><th>Name</th><th>Registries</th><th>Tags</th></tr>\n" >> ./annotations/pushed_images.md
 
 preview_tags "${dev_registries[*]}" "${dev_tags[*]}"
 if $push_prod; then
   preview_tags "${prod_registries[*]}" "${prod_tags[*]}"
 fi
 
-echo "--- done"
```
Strum355, line 156: These "done" kept misleading me into thinking the step was done when its not, and not all of them have a meaningful replacement (unlike e.g. the jobfile one below), so Im removing them

burmudar: Good call. It's also messed me up

burmudar (approved):

PR #61568, opened 2024-04-03
Disable embeddings generation for repos indexed on dotcom

chwarwick:
During the decommissioning of enterprise embeddings it was necessary to keep community embeddings being generated on sourcegraph.com in support of JetBrains users.  Since all PLG users now rely on symf for context it is possible to disable embeddings from being generated on sourcegraph.com.

This is the first of what will be a series of PRs removing the embeddings code from the codebase.

This disables embeddings creation on sourcegraph.com, but retains the Community/Pro rate limit information that is stored in the site config, by reading it directly from the config.
## Test plan

Verify that running dotcom mode querying for a users embeddings rate limit still works.
`sg start dotcom`

```graphql
{
  dotcom{
    codyGatewayDotcomUserByToken(token:"REDACTED"){
      username
      id
      codyGatewayAccess{
        embeddingsRateLimit{
          source
          limit
          intervalSeconds
          allowedModels
        }
      }
    }
  }
}
```

```json
{
  "data": {
    "dotcom": {
      "codyGatewayDotcomUserByToken": {
        "username": "chris",
        "id": "VXNlcjox",
        "codyGatewayAccess": {
          "embeddingsRateLimit": {
            "source": "PLAN",
            "limit": "1234",
            "intervalSeconds": 2592000,
            "allowedModels": [
              "openai/text-embedding-ada-002",
              "sourcegraph/triton"
            ]
          }
        }
      }
    }
  }
}
```


jtibshirani:
> This is the first of what will be a series of PRs removing the embeddings code from the codebase.

@chwarwick this PR makes sense to me, but could we please hold off on removing all embeddings code? I am actively using it for context evaluation at the moment (comparing remote keyword context vs. embeddings) and it's very helpful. I think the code's utility for evaluation and prototyping outweighs the cost of removing it right now, as it's quite well-isolated!

jtibshirani:
> Are you asking that I stop after this PR or skip this one too? I'm ok with either.

Sorry for the confusion, I only read your PR description before responding. Now I see it touches the embeddings config, in a way that makes it harder to enable and test embeddings locally. So I'd prefer we **not** merge this. I can revive it when the right time comes!

It would be great to disable embeddings jobs on dot com though, so we aren't incurring an unnecessary cost. (Perhaps we've already done this, I haven't checked!)

PR #61570, opened 2024-04-03
Cody: increase size of context snippet

jtibshirani:
While comparing context retrieval strategies, I noticed that keyword context
was returning much smaller result snippets than we did from embeddings. In
general, providing more surrounding context to the LLM can only help the
response quality.

This PR increases the number of context lines we return to 20. The reasoning:
* We didn't have problems with embeddings exceeding the context window size
(7000 tokens by default!). So we can use at least as much context as we did for
embeddings.
* Each embedding chunk is roughly 1024 chars (see [Cody context napkin
numbers](https://docs.google.com/spreadsheets/d/1xt-vEbvuQ8OoHXjaP28_YRn83ao377QKxrLsZ4J_62Q/edit#gid=0)).
I tested a variety of file types (Go, Typescript, Markdown), and this roughly
corresponded to 20 lines. It's also a conservative estimate.

## Test plan

Covered by existing unit tests. Also tested a bunch of our "golden queries"
manually in Cody Web and the LLM responses looked good.

### DIFF
--- a/internal/codycontext/context.go
+++ b/internal/codycontext/context.go
@@ -350,10 +350,10 @@ func fileMatchToContextMatches(fm *result.FileMatch) []FileChunkContext {
 	// To provide some context variety, we just use the top-ranked
 	// chunk (the first chunk) from each file
 
-	// 4 lines of leading context, clamped to zero
-	startLine := max(0, fm.ChunkMatches[0].ContentStart.Line-4)
+	// 5 lines of leading context, clamped to zero
+	startLine := max(0, fm.ChunkMatches[0].ContentStart.Line-5)
 	// depend on content fetching to trim to the end of the file
-	endLine := startLine + 8
+	endLine := startLine + 20
 
 	return []FileChunkContext{{
 		RepoName:  fm.Repo.Name,


internal/codycontext/context.go
```diff
@@ -350,10 +350,10 @@ func fileMatchToContextMatches(fm *result.FileMatch) []FileChunkContext {
 	// To provide some context variety, we just use the top-ranked
 	// chunk (the first chunk) from each file
 
-	// 4 lines of leading context, clamped to zero
-	startLine := max(0, fm.ChunkMatches[0].ContentStart.Line-4)
+	// 5 lines of leading context, clamped to zero
```
mmanela, line 353: What is rational for how much leading vs trailing context?

jtibshirani: There is not a strong rationale. I don't think we'll be able to make super principled, well-tested choices here until we have a way of evaluating snippet quality (like we're brainstorming in the eval doc).

keegancsmith (approved):
LGTM. Do we have any reasonable function to post process to avoid sending too many tokens in a snippet? Or maybe passing this through the same function our embeddings chunker uses to try and center around the match? Not for this PR, but just thinking of matching fun things like js files compiled down to one line.

PR #61573, opened 2024-04-03
Context: recognize more text extensions

jtibshirani:
While digging into context quality, I noticed that for some types of question,
our approach of diversifying results across code and docs really helps.
However, we include a fairly narrow set of files as docs/ text files.

This PR extends the list of extensions we recognize. To come up with the new
list, I looked through the [list of languages that GitHub linguist recognizes](https://github.com/github-linguist/linguist/blob/4ac734c15a96f9e16fd12330d0cb8de82274f700/lib/linguist/languages.yml),
searched for `type: prose`, and included everything that was reasonably common
and unambiguous.

## Test plan

Manual testing with Cody Web. Added a new unit test (mostly auto-generated :))


mmanela (approved):

internal/embeddings/embed/files.go
```diff
@@ -22,10 +22,24 @@ var autogeneratedFileHeaders = [][]byte{
...
-	"rst":      {},
-	"txt":      {},
+	"adoc":      {},
+	"asciidoc":  {},
+	"creole":    {},
+	"md":        {},
+	"markdown":  {},
+	"mediawiki": {},
+	"muse":      {},
+	"org":       {},
```
keegancsmith, line 32: thank you :D

keegancsmith (approved):
nice idea on relying on data from linguist

PR #61576, opened 2024-04-04
Svelte: simplify blob scrolling and fixup component composability

camdencheek:
This makes various changes to scrolling behavior in the Svelte prototype.

The biggest change is removing the "scroll file contents to full page size" behavior so that we can bring the page closer to the designs. 

There are a few other changes to make components/pages/layouts more independent and not assume things about the parent they are rendered into. In particular, I made a few changes to our layouts assuming they are being rendered into a flex box.

Comments inline. I recommend reviewing by hiding whitespace-only changes.

## Test plan

See [Loom video](https://www.loom.com/looms/videos)

### DIFF
--- a/client/web-sveltekit/src/lib/dom.ts
+++ b/client/web-sveltekit/src/lib/dom.ts
@@ -210,33 +210,6 @@ export const computeFit: Action<HTMLElement, void, ComputeFitAttributes> = node 
 }
 
 /**
- * Helper action to manage CSS classes on an element. This is used on svelte:body because
- * it doesn't support the class directive.
- * See https://github.com/sveltejs/svelte/issues/3105
- */
-export const classNames: Action<HTMLElement, string | string[]> = (node, classes) => {
-    // Converts the input to an array of non-empty strings.
-    // Empty strings are not valid inputs for classList and would throw an error.
-    function clean(classes: string | string[]): string[] {
-        return (Array.isArray(classes) ? classes : [classes]).filter(cls => cls.trim().length > 0)
-    }
-
-    classes = clean(classes)
-    node.classList.add(...classes)
-
-    return {
-        update(newClasses) {
-            node.classList.remove(...classes)
-            classes = clean(newClasses)
-            node.classList.add(...classes)
-        },
-        destroy() {
-            node.classList.remove(...classes)
-        },
-    }
-}
-
-/**
  * An action to move the attached element to the end of the document body.
  */
 export const portal: Action<HTMLElement> = target => {
--- a/client/web-sveltekit/src/lib/global-notifications/GlobalNotifications.svelte
+++ b/client/web-sveltekit/src/lib/global-notifications/GlobalNotifications.svelte
@@ -1,6 +1,7 @@
 <script lang="ts" context="module">
-    import { AlertType } from '$lib/graphql-types'
     import { formatDistanceStrict, isAfter } from 'date-fns'
+
+    import { AlertType } from '$lib/graphql-types'
 
     type PossibleAlertVariation = 'info' | 'warning' | 'danger'
 
@@ -33,13 +34,13 @@
 </script>
 
 <script lang="ts">
-    import { settings } from '$lib/stores'
-
-    import { Markdown } from '$lib/wildcard'
-    import DismissibleAlert from './DismissibleAlert.svelte'
-
-    import type { GlobalNotifications } from './GlobalNotifications.gql'
     import { differenceInDays, parseISO } from 'date-fns'
+
+    import { settings } from '$lib/stores'
+    import { Markdown } from '$lib/wildcard'
+
+    import DismissibleAlert from './DismissibleAlert.svelte'
+    import type { GlobalNotifications } from './GlobalNotifications.gql'
 
     export let globalAlerts: GlobalNotifications
 
@@ -116,11 +117,6 @@
 <style lang="scss">
     .root {
         width: 100%;
-        border-top: 1px solid var(--border-color-2);
-
-        &:empty {
-            display: none;
-        }
     }
 
     .proxy-link {
--- a/client/web-sveltekit/src/routes/+layout.svelte
+++ b/client/web-sveltekit/src/routes/+layout.svelte
@@ -109,6 +109,12 @@
 </div>
 
 <style lang="scss">
+    .inner-body {
+        display: flex;
+        flex-direction: column;
+        height: 100%;
+    }
+
     main {
         isolation: isolate;
         flex: 1;
@@ -116,11 +122,5 @@
         flex-direction: column;
         box-sizing: border-box;
         min-height: 0;
-    }
-
-    .inner-body {
-        display: flex;
-        flex-direction: column;
-        height: 100%;
     }
 </style>
--- a/client/web-sveltekit/src/routes/+layout.svelte
+++ b/client/web-sveltekit/src/routes/+layout.svelte
@@ -3,9 +3,8 @@
 
     import { browser, dev } from '$app/environment'
     import { isErrorLike } from '$lib/common'
-    import { classNames } from '$lib/dom'
     import { TemporarySettingsStorage } from '$lib/shared'
-    import { isLightTheme, setAppContext, scrollAll } from '$lib/stores'
+    import { isLightTheme, setAppContext } from '$lib/stores'
     import { createTemporarySettingsStorage } from '$lib/temporarySettings'
 
     import Header from './Header.svelte'
--- a/client/web-sveltekit/src/routes/+layout.svelte
+++ b/client/web-sveltekit/src/routes/+layout.svelte
@@ -94,13 +94,13 @@
     <meta name="description" content="Code search" />
 </svelte:head>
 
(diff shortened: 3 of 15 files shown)


client/web-sveltekit/src/lib/global-notifications/GlobalNotifications.svelte
```diff
@@ -116,11 +117,6 @@
 <style lang="scss">
     .root {
         width: 100%;
-        border-top: 1px solid var(--border-color-2);
-
-        &:empty {
-            display: none;
-        }
```
camdencheek, line 123: Because the component contains whitespace, this `:empty` was never triggering.

That meant we always had a `border-top`. Since we didn't add `box-sizing: border-box`, this meant the content page was one pixel larger than the viewport, which made a scrollbar show for the full page.

I don't think this `border-top` is actually necessary anyways so I just removed it. The alerts look good without it. I also removed the `:empty` because it wasn't actually doing what it claimed to be doing.

client/web-sveltekit/src/routes/+layout.svelte
```diff
@@ -94,37 +94,30 @@
     <meta name="description" content="Code search" />
 </svelte:head>
 
-<svelte:body use:classNames={$scrollAll ? '' : 'overflowHidden'} />
```
camdencheek, line 97: This removes the special "scrollAll" behavior so the blob page behaves more like the designs we're moving towards.

I think this is the only use of `classNames` -- should we get rid of it, or is it a useful utility to keep around?

client/web-sveltekit/src/routes/+layout.svelte
```diff
@@ -94,37 +94,30 @@
...
-        }
+        height: 100%;
     }
 
     main {
         isolation: isolate;
         flex: 1;
-        display: flex;
-        flex-direction: column;
-        box-sizing: border-box;
```
camdencheek, line 128: This is part of my newfound crusade of "components shouldn't need to know things about their parent". In this case, the things we rendered into the `main` slot depended on `main` being a column-oriented flex box. Instead, we just make `main` very normal and leave any flex-boxing to the child layouts (if they even need it).

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte
```diff
@@ -172,10 +172,8 @@
 <style lang="scss">
     section {
         display: flex;
-        flex: 1;
-        flex-shrink: 0;
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte:176]: Since flex layout is now handled by the parent, this doesn't need to have any flex properties.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte
```diff
@@ -172,10 +172,8 @@
...
         background-color: var(--code-bg);
-        min-height: 100vh;
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte:178]: This is just getting rid of the "scroll the blob content to the full page size"

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte
```diff
@@ -93,8 +93,10 @@
         {/await}
     {:else if blob}
         {#if blob.richHTML && !showRaw}
-            <div class={`rich ${markdownStyles.markdown}`}>
-                {@html blob.richHTML}
+            <div class="rich-scroller">
+                <div class={`rich-content ${markdownStyles.markdown}`}>
+                    {@html blob.richHTML}
+                </div>
             </div>
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte:100]: Added another layer here because I wanted the scroll container to be the full width even though the content is constrained to a smaller column.

client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte
```diff
@@ -60,50 +61,61 @@
     <title>{data.displayRepoName} - Sourcegraph</title>
 </svelte:head>
 
-<nav aria-label="repository">
-    <h1><a href="/{repoName}"><Icon svgPath={mdiSourceRepository} inline /> {displayRepoName}</a></h1>
-    <!--
+<div class="root">
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte:64]: Added a wrapper because I don't want to assume this is rendered in a flex box.

client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte
```diff
@@ -60,50 +61,61 @@
...
+                    </span>
+                </MenuLink>
+            {/each}
+        </DropdownMenu>
+        <RepoSearchInput repoName={data.repoName} />
+    </nav>
+
+    <div class="content">
+        <slot />
+    </div>
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte:109]: Added a wrapper because I don't want the child to have to assume it is rendered in a flex box.

client/web-sveltekit/src/routes/search/SearchResults.svelte
```diff
@@ -126,49 +126,62 @@
...
 </svelte:head>
 
-<div class="search">
-    <SearchInput {queryState} />
-</div>
-
-<div class="search-results">
-    <div style:width={`clamp(14rem, ${$filtersSidebarPosition * 100}%, 35%)`}>
-        <DynamicFiltersSidebar {selectedFilters} streamFilters={$stream.filters} searchQuery={queryFromURL} {state} />
+<div class="root">
```
camdencheek, line 129: Added a wrapper because I don't want to assume this is rendered in a flex box.

client/web-sveltekit/src/lib/dom.ts
```diff
@@ -209,33 +209,6 @@ export const computeFit: Action<HTMLElement, void, ComputeFitAttributes> = node
     }
 }
 
-/**
- * Helper action to manage CSS classes on an element. This is used on svelte:body because
- * it doesn't support the class directive.
- * See https://github.com/sveltejs/svelte/issues/3105
- */
```
camdencheek, line 216: Unused after removing the `scrollAll` behavior on the body

fkling:
I'll take a look at this later today.

> The biggest change is removing the "scroll file contents to full page size" behavior so that we can bring the page closer to the designs.

Do we explicitly not want full page scrolling or is the design underspecified. cc @vovakulikov, since that's important to you.

client/web-sveltekit/src/routes/+layout.svelte
```diff
@@ -94,37 +93,30 @@
...
+        min-height: 0;
```
vovakulikov, line 120: I understand (I think I do) this workaround with manually min sized on flex-child but let's leave a comment 
here for future questions, something like that we want to keep height fixed by container and children overflow:auto

client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte
```diff
@@ -166,4 +178,9 @@
     nav {
         color: var(--body-color);
     }
+
+    .content {
+        flex: 1;
+        min-height: 0;
```
Comment (vovakulikov) [inline client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte:184]: Same suggestion here, let's leave a comment about manually set of min-height

client/web-sveltekit/src/routes/search/SearchResults.svelte
```diff
@@ -126,49 +126,62 @@
     <title>{queryFromURL} - Sourcegraph</title>
```
vovakulikov, line 126: This file change is probably not necessary because changed the whole layout with resizable panels in other PRs

vovakulikov (approved):
Agree with changes. I would prefer to merge panels first to avoid git conflicts

camdencheek:
> Do we explicitly not want full page scrolling 

I personally do not want full-page scrolling 😄 I find it surprising and I think it makes the page feel broken. I am not against a full-screen mode, but I think that should be a mode (e.g. a button) rather than just scrolling away the header (which makes the page feel unstable). That said, with the new proposed designs, there is significantly less vertical space being used by headers, so I think there's a good chance a full-screen mode won't feel as necessary.

client/web-sveltekit/src/routes/+layout.svelte
camdencheek: Yeah, see [Setting min height to zero](https://ishadeed.com/article/min-max-css/) here. 

Basically, for flex items specifically, `min-height: auto` means `min-hight: content`. That means that the flex item will grow to fit its content rather than fitting inside the box it's given and scrolling the content. So setting it to an explicit `min-height: 0` says "this flex item can resize down to zero"

camdencheek: Basically, I've come to the conclusion that anywhere I want a flex item that is inner-scrollable, it should have `min-height: 0` added

camdencheek:
Going to abandon this because it's conflicting with more important work and I never did figure out how to fix scroll restoration. Not a complete waste -- I learned a lot, and I can implement the composability things in a more granular manner.

PR #61577, opened 2024-04-04
Improve git blame performance and fix previous commit location when file was renamed

### LINKED ISSUE #60137: Git Blame runs GetCommit a ton of times

### LINKED ISSUE #60159: Blame: Need to return OldPath for files that were renamed
When showing the "Show blame prior to this change" and the commit in question has renamed the file, we cannot reliably link to the file at the parent, because we don't know the previous name of the file. 

Example: (The commit from Dave)
https://sourcegraph.com/.api/blame/github.com/sourcegraph/sourcegraph@dd34d1b/stream/docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/syntect_html.rs

pjlast:
Closes #60137 
Closes #60159 

This PR does several things to improve Sourcegraph's Git Blame functionality:
- The previous commit of a change is now parsed from the output of `git blame`, which removes the need to fetch every commit in a blame just for the parent commit.
- The frontend now uses the `filename` field that's part of the git blame output. This filename is the original filename of the change, which allows for accurate navigation to the original file at the specified commit. Previously, if a file moved during a commit, this functionality would simply not work.

Overall this is a massive performance improvement, reducing the time it takes for a git blame on a large file to load from several seconds down to less than 500 milliseconds.

## Test plan

Manual tests, unit tests adjusted where necessary.


cmd/gitserver/internal/git/gitcli/blame.go
```diff
@@ -53,159 +52,111 @@ type blameHunkReader struct {
...
-			}
-			continue
-		}
-
-		// After that, we're either reading extras, or a new entry.
-		ok, err := parseExtra(br.cur, annotation, fields)
+// Read returns the next blame hunk. The returned hunk is only valid
+// until the next call to Read, and needs to be copied if it needs
+// to be stored.
+func (br *blameHunkReader) Read() (*gitdomain.Hunk, error) {
```
pjlast, line 68: This reader has been drastically simplified. `git blame --incremental` displays commits sequentially, so it's not necessary to keep an entire hashmap of previous commit data. We only need the current commit being processed.

The blame hunks are also deterministically structured: They start with the commit hash and end with the `filename` field, so I simplified the parsing and logic here. Once we read a `filename` field, we return the hunk. No funky logic required.

Finally, I'm using a single in-memory gitdomain.Hunk. Since hunks are immediately converted to a proto and sent, I don't think there is value in creating a new hunk object every time, especially since a vast majority of the hunks will share data and require minimal updates.

client/web/src/util/url.ts
```diff
@@ -29,6 +29,18 @@ export function replaceRevisionInURL(href: string, newRevision: string): string
...
+ * Returns a URL to a file at a specific commit.
+ */
+export function getURLToFileCommit(href: string, filename: string, revision: string): string {
+    const parsed = parseBrowserRepoURL(href)
+    parsed.revision = revision
+    parsed.filePath = '/-/blob/' + filename
+
+    const url = new URL(href, window.location.href)
+    return `/${parsed.repoName}@${parsed.revision}${parsed.filePath}${url.search}${url.hash}`
+}
```
eseliger, line 42: could we let cody generate one or two unit tests for this function to verify that it constructs valid URLs?

cmd/gitserver/internal/git/iface.go
```diff
@@ -94,6 +94,7 @@ type BlameRange struct {
 // BlameHunkReader is a reader for git blame hunks.
 type BlameHunkReader interface {
 	// Consume the next hunk. io.EOF is returned at the end of the stream.
+	// The returned hunk is only valid until the next call to Read.
```
eseliger, line 97: This feels a bit like an easy footgun to me 🤔 For example, when we start to use chunking or so to reduce the message count (tried that in another PR for another endpoint just today). Maybe changing from *Hunk to Hunk instead would be another option? That case it wouldn't have to allocate heap (but copy a bit more data).

client/web/src/repo/blame/useBlameHunks.ts
```diff
@@ -141,6 +142,7 @@ const fetchBlameViaStreaming = memoizeObservable(
                                     endLine: rawHunk.endLine,
                                     message: rawHunk.message,
                                     rev: rawHunk.commitID,
+                                    previousFilename: rawHunk.filename,
```
eseliger, line 145: Since we're renaming this field here, should it be called "previousFilename" in the API response as well? 

And oh -- we already exposed this in the API? I thought we'll have to read this from the previous section in the incremental output 🤔

cmd/gitserver/internal/git/gitcli/blame.go
```diff
@@ -53,159 +52,111 @@ type blameHunkReader struct {
...
 
-			return res, nil
+			// If we've finished reading the body we can return the hunk.
+			if done {
+				return br.cur, nil
+			}
 		}
 	}
+
+	// Return the scanner error if ther was one
```
eseliger, line 99:
```suggestion
	// Return the scanner error if there was one
```

cmd/gitserver/internal/git/gitcli/blame.go
```diff
@@ -53,159 +52,111 @@ type blameHunkReader struct {
...
+	// hunk header: <commit hash> <original file start line> <current file start line> <number of lines>
+	// followed by the hunk body.
+	// The hunk body is a list of lines in the format: <field name> <field value>
+	// A hunk body terminates with the filename field.
+	// If a hunk is part of the same commit as a previous hunk, only
+	// the difference in fields is returned, so the previous commit
+	// should be cached until a new commit is encountered.
+	for br.sc.Scan() {
+		// First parse the hunk header.
+		err := parseHeader(br.cur, br.sc.Bytes())
```
eseliger, line 79: If we stick to this reuse pattern, I think I would feel a bit better if we at least reset the br.cur here before parsing into it, otherwise if there's optional fields we might not be resetting them alright. But also check my other comment on this :)

cmd/gitserver/internal/git/gitcli/blame_test.go
eseliger, line 1: Could we get a test that covers the "file was renamed" case as well where we can clearly see that the filename differs?

internal/gitserver/gitdomain/common.go
```diff
@@ -230,14 +229,15 @@ func (m Message) Body() string {
...
-	Filename  string
+	StartLine        uint32 // 1-indexed start line number
+	EndLine          uint32 // 1-indexed end line number
+	StartByte        uint32 // 0-indexed start byte position (inclusive)
+	EndByte          uint32 // 0-indexed end byte position (exclusive)
+	CommitID         api.CommitID
+	PreviousCommitID api.CommitID
+	Author           Signature
+	Message          string
+	Filename         string
```
eseliger, line 240: Could we add a comment here what this is?

```suggestion
	// Filename is the name of the file at the commit prior to this change.
	Filename         string
```

cmd/gitserver/internal/git/gitcli/blame.go
```diff
@@ -53,159 +52,111 @@ type blameHunkReader struct {
...
-		if len(content) >= 2 && content[0] == '<' && content[len(content)-1] == '>' {
-			hunk.Author.Email = string(content[1 : len(content)-1])
-		}
+		hunk.Author.Email = string(bytes.Trim(content, "<>"))
 	case "author-time":
 		var t int64
 		t, err = strconv.ParseInt(string(content), 10, 64)
+		if err != nil {
+			return false, err
+		}
```
eseliger, line 150: oh oh

cmd/gitserver/internal/git/gitcli/blame.go
```diff
@@ -53,159 +52,111 @@ type blameHunkReader struct {
...
 		hunk.Author.Date = time.Unix(t, 0).UTC()
-	case "author-tz":
-		// do nothing
```
eseliger, line 183: I wonder if we don't actually have to consider this -- or is author-time always in UTC?

cmd/gitserver/internal/git/gitcli/blame.go
```diff
@@ -53,159 +52,111 @@ type blameHunkReader struct {
...
-	case "committer", "committer-mail", "committer-tz", "committer-time":
 	case "summary":
 		hunk.Message = string(content)
 	case "filename":
 		hunk.Filename = string(content)
+		return true, nil
```
eseliger, line 156: why the return here? can we add a comment?

cmd/gitserver/internal/git/gitcli/blame.go
```diff
@@ -53,159 +52,111 @@ type blameHunkReader struct {
...
-	case "boundary":
-	default:
-		// If it doesn't look like an entry, it's probably an unhandled git blame
-		// annotation.
-		if len(annotation) != 40 && bytes.Count(content, []byte(" ")) != 3 {
-			err = errors.Newf("unhandled git blame annotation: %s")
-		}
-		ok = false
+		commitID, _, _ := bytes.Cut(content, []byte(" "))
+		hunk.PreviousCommitID = api.CommitID(commitID)
```
eseliger, line 159: I would've thought that we also need to parse the filename in this section? 🤔

client/web/src/repo/blame/useBlameHunks.ts
```diff
@@ -49,9 +50,10 @@ export interface BlameHunk {
     }
     commit: {
         url: string
-        parents: {
-            oid: string
-        }[]
+        previous: {
+            rev: string
```
eseliger, line 54: for consistency, can we call this `commitID` as well?

client/web/src/repo/blame/useBlameHunks.ts
```diff
@@ -69,7 +71,10 @@ interface RawStreamHunk {
         Date: string
     }
     commit: {
-        parents: string[]
+        previous: {
+            commitID: string
+            filename: string
+        }
```
eseliger, line 77: this should be `| null`, no?

client/web/src/repo/blame/useBlameHunks.ts
eseliger: maybe previous also needs to be nullable?

client/web/src/repo/blob/BlameDecoration.tsx
```diff
@@ -219,12 +219,16 @@ export const BlameDecoration: React.FunctionComponent<BlameDecorationProps> = ({
                                 />
                             </div>
                         </div>
-                        {blameHunk.commit.parents.length > 0 && (
+                        {blameHunk.commit.previous.rev.length > 0 && (
```
eseliger, line 222: would be nice to instead check if previous is not null, and expect that rev is properly set then

client/web/src/util/url.ts
eseliger: friendly ping on test :)

cmd/frontend/internal/httpapi/stream_blame.go
```diff
@@ -182,8 +167,8 @@ type BlameHunkResponse struct {
 }
 
 type BlameHunkCommitResponse struct {
-	Parents []api.CommitID `json:"parents"`
-	URL     string         `json:"url"`
+	Previous gitdomain.PreviousCommit `json:"previous"`
```
eseliger, line 170: maybe maybe pointer so it's marshalling to `null` properly for creation commits where no prev exists?

cmd/gitserver/internal/git/gitcli/blame.go
pjlast: You'll see in the `parseHeader` function it resets all of the fields if it's a new commit

eseliger (commented):
not yet done, but waiting to resolve more commits

client/web/src/repo/blame/useBlameHunks.ts
pjlast: I was going for consistency with the main object, where the `commitID` is renamed to `rev` 🤷‍♂️

cmd/gitserver/internal/git/gitcli/blame.go
pjlast: Yeah author-time is a unix timestamp, so it should be UTC

client/web/src/repo/blame/useBlameHunks.ts
eseliger: hmm generally renaming fields feels like a red flag to me that it will cause confusion down the line, but I see why 🤔 maybe time to change both? 😬

eseliger (approved):
LGTM, pending a test in gitcli that covers the rename case. Nice!

I see the comment about the non-reusability of Hunk has been removed but aren't we still reusing the pointer? 🤔

pjlast:
@eseliger we make a copy of the object before sending it to the caller

PR #61580, opened 2024-04-04
fix: allow more than 65535 repos in deny list 

bahrmichael:
Closes issue https://github.com/sourcegraph/sourcegraph/issues/61353

Closes PR https://github.com/sourcegraph/sourcegraph/pull/61525

This PR fixes the error `extended protocol limited to 65535 parameters` by not sending a large list of predicates via an SQL query, but filtering out repos that a user should not have access to in the application layer.

The new filter runs with O(n) time complexity which should be equal or better than the previous SQL query. There is however more network traffic, as we now transfer all the repos that a user does not have access to from the database to the application.

## Test plan

- Added new unit and benchmarking tests
- Previous use cases are covered by existing test cases


internal/insights/store/store.go
```diff
@@ -976,11 +972,34 @@ func (s *Store) GetAllDataForInsightViewID(ctx context.Context, opts ExportOpts)
...
+func FilterSeriesPoints(denyList []api.RepoID, array []SeriesPointForExport) []SeriesPointForExport {
+	// If there is nothing to filter, then return early. This skips the assignment in the later for-loop.
+	if len(denyList) == 0 {
+		return array
+	}
+	// We turn the denyList into a map to ensure O(n) time complexity
+	denyMap := make(map[int32]bool)
+	for _, repoID := range denyList {
+		denyMap[int32(repoID)] = true
+	}
```
camdencheek, line 989: Go trick: we can use a zero-sized empty struct as the value to save some memory.

```suggestion
	denyMap := make(map[int32]struct{})
	for _, repoID := range denyList {
		denyMap[int32(repoID)] = struct{}{}
	}
```

camdencheek: Also, why do we convert to `int32`? Can't se use `map[api.RepoID]struct{}` directly?

internal/insights/store/store_test.go
```diff
@@ -983,14 +983,134 @@ func TestGetOffsetNRecordingTime(t *testing.T) {
...
+
+	// Second-best case: No assignments because everything is filtered out
+	b.Run("filter out all repos", func(b *testing.B) {
+		var denyList []api.RepoID
+		var points []SeriesPointForExport
+
+		for i := 0; i < b.N; i++ {
+			denyList = append(denyList, api.RepoID(i))
+			points = append(points, SeriesPointForExport{RepoId: optionalRepoID(api.RepoID(i))})
+		}
```
camdencheek, line 1018: This is not a correctly formed benchmark.

In a Go benchmark, the section inside the `b.N` loop is the piece being measured. `b.N` actually changes dynamically over the course of the benchmark to try to target 1 second worth of iterations (by default).

If you want to benchmark different sizes of slice, you can use sub benchmarks. 

```go
for _, size := range []int{0, 1, 10, 1000} {
    b.Run(strconv.Itoa(size), func(b *testing.B) {
        ...
    })
}
```

internal/insights/store/store.go
```diff
@@ -976,11 +972,34 @@ func (s *Store) GetAllDataForInsightViewID(ctx context.Context, opts ExportOpts)
...
 	return results, nil
 }
 
+func FilterSeriesPoints(denyList []api.RepoID, array []SeriesPointForExport) []SeriesPointForExport {
+	// If there is nothing to filter, then return early. This skips the assignment in the later for-loop.
+	if len(denyList) == 0 {
+		return array
+	}
+	// We turn the denyList into a map to ensure O(n) time complexity
+	denyMap := make(map[int32]bool)
```
camdencheek, line 986: This can be pre-allocated to avoid making extra allocations when it grows below

```suggestion
	denyMap := make(map[int32]bool, len(denyList))
```

internal/insights/store/store.go
```diff
@@ -976,11 +972,34 @@ func (s *Store) GetAllDataForInsightViewID(ctx context.Context, opts ExportOpts)
 		return nil, errors.Wrap(err, "fetching code insights data")
 	}
 
+	results = FilterSeriesPoints(denylist, results)
+
 	return results, nil
 }
 
+func FilterSeriesPoints(denyList []api.RepoID, array []SeriesPointForExport) []SeriesPointForExport {
```
camdencheek, line 980: nit: I think we should prefer a descriptive name rather `array`. Maybe `points`?

internal/insights/store/store.go
```diff
@@ -976,11 +972,34 @@ func (s *Store) GetAllDataForInsightViewID(ctx context.Context, opts ExportOpts)
 		return nil, errors.Wrap(err, "fetching code insights data")
 	}
 
+	results = FilterSeriesPoints(denylist, results)
```
camdencheek, line 975: Can you add a security note here? This is a very understated line 🙂

camdencheek (approved):
Nice! This looks good to me. Left a few comments, but nothing critical. Happy to provide a followup review if you'd like, but not necessary.

internal/insights/store/store.go
bahrmichael: Yes, it would be unfortunate if we loose that!

internal/insights/store/store.go
bahrmichael: Done!

internal/insights/store/store.go
bahrmichael: Done!

internal/insights/store/store_test.go
bahrmichael: Gotcha, I misunderstood how they work. I've changed it to include some higher values too. Thanks for the pointer and suggestion! The code snippets you provide are super helpful to understand your intention :)

internal/insights/store/store.go
bahrmichael: Not sure why I used int32, probably misunderstood an error. It's now using api.RepoID across the board.

internal/insights/store/store.go
```diff
@@ -976,11 +972,43 @@ func (s *Store) GetAllDataForInsightViewID(ctx context.Context, opts ExportOpts)
...
+	}
+	// We turn the denyList into a map to ensure O(n) time complexity
+	denyMap := make(map[api.RepoID]struct{})
+	for _, repoID := range denyList {
+		denyMap[repoID] = struct{}{}
+	}
+	// Based on https://stackoverflow.com/a/59051095
+	n := 0
+	for _, x := range points {
+		// If there is no RepoId on a point, then we can't filter by repo permissions and default to allowing access.
```
camdencheek, line 996: This is a change in benavior, right? Previously we were checking `repo_id NOT IN (...)`, but if `repo_id` is NULL, Postgres evaluates `NULL not in (1,2,3)` as NULL, which would mean this was previously excluded, right?

camdencheek: Example of what I'm talking about:

```
localhost sourcegraph@sourcegraph=# create table test (id int);
CREATE TABLE
Time: 22.762 ms
localhost sourcegraph@sourcegraph=# insert into test values (1), (2), (3);
INSERT 0 3
Time: 1.803 ms
localhost sourcegraph@sourcegraph=# select * from test;
┌────┐
│ id │
├────┤
│  1 │
│  2 │
│  3 │
└────┘
(3 rows)

Time: 0.826 ms
localhost sourcegraph@sourcegraph=# select * from test where NULL in (5,6,7);
┌────┐
│ id │
├────┤
└────┘
(0 rows)

Time: 0.935 ms
```

bahrmichael: I need to break this down myself (see below) to make sure I understand it right, and got to the same conclusion: The original code prevents users from seeing records where the ID is null.

This however means that no one (even admins) would be allowed to see those records, because their list of denied repos will just be empty and the check for `NULL not in ()` still yields false.

---

We have this original query

```
select iv.title, ivs.label, i.query, isrt.recording_time, rn.name, coalesce(sp.value, 0) as value, sp.capture
from %s isrt
    join insight_series i on i.id = isrt.insight_series_id
    join insight_view_series ivs ON i.id = ivs.insight_series_id
    join insight_view iv ON ivs.insight_view_id = iv.id
    left outer join %s sp on sp.series_id = i.series_id and sp.time = isrt.recording_time
    left outer join repo_names rn on sp.repo_name_id = rn.id
	where iv.unique_id = %s and %s
    order by iv.title, isrt.recording_time, ivs.label, sp.capture;
```

that we can condense down to

```
select *
from %s isrt
    left outer join %s sp
	where %s
```

The first `%s` is some table, the second is the series points table, and the third are all the predicates (including denied repos and regexes).

The denied repos part is essentially `sp.repo_id not in (%s)` where `%s` is a list of repo ids.

Filled in the query above looks like this:

```
select *
from %s isrt
    left outer join %s sp
	where sp.repo_id not in (%s)
```

We can further simplify this to

```
select *
from TABLE t
	where t.id not in (%s)
```

If id is in the list it should be blocked. The case we're discussing here is if the record should be blocked when t.id is NULL.

Here's a fiddle I used, where I get the same results as you:

https://www.db-fiddle.com/f/4jyoMCicNSZpjMt4jFYoz5/12981

camdencheek: TIL about db-fiddle! This all sounds right to me. Thanks for the detailed thought process

camdencheek (approved):

camdencheek:
FYI, "Closes issue #NUMBER" does not auto-close that issue, but `Closes #NUMBER` does. GitHub auto-close syntax is kinda picky 🙂

PR #61592, opened 2024-04-04
sg: detect and use dockerhub registry

burmudar:
Do the right thing if someone specifies the public docker hub registry.

## Test plan
```
👉 using Docker Hub registry index.docker.io/sourcegraph
👉 updating images to latest
⚠️ Registry credentials are not provided and could not be retrieved from docker config.
⚠️ You will be using anonymous requests and may be subject to rate limiting by Docker Hub.
❌ cannot load docker creds from store: failed to find store provider or credential helper for https://registry.hub.docker.com/v2
Error: bazel exited with exit code: 1
```

### DIFF
--- a/dev/sg/sg_ops.go
+++ b/dev/sg/sg_ops.go
@@ -202,6 +202,13 @@ func opsUpdateImages(
 				return errors.Errorf("invalid custom registry %q", registryType)
 			}
 
+			// might have specified the public registry url without knowing, so we check and create dockrhub!
+			if strings.Contains(registryType, "index.docker.io") {
+				registry = images.NewDockerHub("sourcegraph", dockerUsername, dockerPassword)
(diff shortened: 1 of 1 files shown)


jhchabran (approved):

dev/sg/sg_ops.go
```diff
@@ -202,6 +202,13 @@ func opsUpdateImages(
 				return errors.Errorf("invalid custom registry %q", registryType)
 			}
 
+			// might have specified the public registry url without knowing, so we check and create dockrhub!
+			if strings.Contains(registryType, "index.docker.io") {
+				registry = images.NewDockerHub("sourcegraph", dockerUsername, dockerPassword)
```
burmudar, line 207: We could extra the org from the provided registry but there is no actual usecase (at this point) for being able to pin images from other public docker registries

PR #61600, opened 2024-04-04
Consolidate DiffPath and Diff methods

eseliger:
Diff already has an option for scoping to a specific path, so merging these two together, to simplify the API surface we need to create for the new gRPC APIs around diffs.

## Test plan

Adjusted a few tests, hoping that integration and E2E tests cover this. Also ran some git commands locally to check if they produce roughly equivalent output. Code reviews will help as well.

### DIFF
--- a/internal/codeintel/codenav/gittree_translator.go
+++ b/internal/codeintel/codenav/gittree_translator.go
@@ -2,6 +2,7 @@ package codenav
 
 import (
 	"context"
+	"io"
 	"strconv"
 	"strings"
 
@@ -11,6 +12,7 @@ import (
 	"github.com/sourcegraph/sourcegraph/internal/codeintel/codenav/shared"
 	"github.com/sourcegraph/sourcegraph/internal/gitserver"
 	sgtypes "github.com/sourcegraph/sourcegraph/internal/types"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
 )
 
 // GitTreeTranslator translates a position within a git tree at a source commit into the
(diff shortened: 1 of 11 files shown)


internal/codeintel/codenav/gittree_translator.go
```diff
@@ -155,7 +157,27 @@ func (g *gitTreeTranslator) readCachedHunks(ctx context.Context, repo *sgtypes.R
...
+	fd, err := r.Next()
+	if err != nil {
+		if errors.Is(err, io.EOF) {
+			return nil, nil
+		}
+		return nil, err
+	}
+	if err := r.Close(); err != nil {
+		return nil, err
+	}
```
varungandhi-src, line 179: Given that the return type of `Diff` is `DiffFileIterator`, shouldn't we be calling `Next()` in a loop until we reach `EOF`? I'm confused as to why it is OK to only call `Next()` once here, and then call `Close()` even if we didn't reach EOF.

It seems like there isn't any existing test case exercising the case when one `Next()` call doesn't read in all the data?

varungandhi-src (commented):
Perhaps I'm misunderstanding something, but it seems like the iteration logic is not quite right.

internal/codeintel/codenav/gittree_translator.go
eseliger: Good catch, I added a test for that and put the r.Close() in a defer here as well, to make sure it's called in the empty case.

varungandhi-src: I see the test case with the close and the error checking -- my question is about the higher-level behavior. Previously, `DiffPath` was returning a slice, not an iterator, so it seems like `readHunks` was expecting to get all the hunks. Now that `Diff` returns an iterator, it seems like we're not iterating to the end by calling `Next` in a loop until we hit EOF. This would mean that the outward facing behavior of `readHunks` could potentially change in the iterator refilling case -- previously we would've gotten all hunks, but now we'd only get a subset of hunks.

Am I misunderstanding what is possible/is the new behavior identical to the old one?

eseliger: Ah I see what you mean now!

The iterator iterates over one _file_ at a time, and returns all hunks for said file at once.

So we can safely just look at the first file here, since we pass a single path, is my idea here.

Does that make sense?

varungandhi-src: Understood, thanks for the explanation.

PR #61614, opened 2024-04-04
auth: bitbucket: fix slice indexing error when constructing oauth error messages

ggilmore:
https://sourcegraph.slack.com/archives/C03JR7S7KRP/p1712180759513679?thread_ts=1712168903.358759&cid=C03JR7S7KRP

When a user signs in to Sourcegraph using a bitbucket oauth provider, there must be a corresponding Sourcegraph user with a verified email as their bitbucket account. (See https://sourcegraph.com/docs/admin/auth#linking-accounts-from-multiple-auth-providers). 

In the case where there isn't a matching Sourcegraph user with a verified email, we attempt to return an error message that looks like:

`No Sourcegraph user exists matching any of the verified emails: $VERIFIED_EMAILS.\n\nFirst error was: $ERROR` .

This was the original logic to construct the list of verified emails for the error message

```go
verifiedEmails := make([]string, 0, len(attempts))
	for i, attempt := range attempts {
		verifiedEmails[i] = attempt.email
	}
```

However, this logic has a fatal bug. The `make([]string, 0, len(attempts)`) constructs a slice with `0` len. Indexing into a slice with 0 length immediately throws an out-of-bounds panic. The correct thing to do here is to _append_ to the slice since it's initially empty.

This PR fixes this issue. 

## Test plan

New unit test

### DIFF
--- a/cmd/frontend/internal/auth/bitbucketcloudoauth/session_test.go
+++ b/cmd/frontend/internal/auth/bitbucketcloudoauth/session_test.go
@@ -3,6 +3,7 @@ package bitbucketcloudoauth
 import (
 	"context"
 	"encoding/json"
+	"github.com/stretchr/testify/assert"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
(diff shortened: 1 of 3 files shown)


cmd/frontend/internal/auth/bitbucketcloudoauth/session_test.go
```diff
@@ -306,3 +307,74 @@ func acct(serviceType, serviceID, clientID, accountID string) extsvc.AccountSpec
...
+	}
+
+	// Create issuer helper
+	issuer := &sessionIssuerHelper{
+		baseURL: u,
+		client:  bbClient,
+	}
+
+	// Assert no panic and that the returned error is the expected one
+	assert.NotPanics(t, func() {
```
eseliger, line 365: TIL about assert.notpanics!

eseliger (approved):
nice find!

cmd/frontend/internal/auth/bitbucketcloudoauth/session_test.go
ggilmore: Yeah. It's a bit redundant in this case (whether or not the test as a whole continues executing is sufficient, but I thought it was nice to be as explicit as possible here).

PR #61615, opened 2024-04-04
Remove clone progress DB experiment

eseliger:
This was an experiment from a while ago about storing the clone progress in the DB instead of making a request to gitserver to retrieve it. This has never been activated and is thus dead code, so removing it. With the switch to a dbworker we get better logging for free anyways.

## Test plan

Removed dead code, tests should cover any regressions.

### DIFF
--- a/client/web/src/site-admin/components/RepoMirrorInfo.tsx
+++ b/client/web/src/site-admin/components/RepoMirrorInfo.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react'
 
 import { Timestamp } from '@sourcegraph/branded/src/components/Timestamp'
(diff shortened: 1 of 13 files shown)


client/web/src/site-admin/components/RepoMirrorInfo.tsx
```diff
@@ -35,14 +35,6 @@ export const RepoMirrorInfo: React.FunctionComponent<
                                 </Tooltip>
                             </>
                         )}
-                        {mirrorInfo.cloneInProgress && (mirrorInfo.cloneProgress ?? '').trim() !== '' ? (
```
eseliger, line 38: was only set if the FF was on, so removing here.

pjlast (approved):

PR #61617, opened 2024-04-05
Re-enable bazel cody-gateway

jamesmcnamara:
As title. The issue here SHOULD be resolved by https://github.com/sourcegraph/sourcegraph/pull/61260. When I've tested this locally it seems to work fine. Could someone who had this problem try updating sg and testing it again?

## Test plan
Ran locally.

### DIFF
--- a/sg.config.yaml
+++ b/sg.config.yaml
@@ -1384,6 +1384,7 @@ commandsets:
       - blobstore
       - embeddings
       - telemetry-gateway
+    bazelCommands:
       - cody-gateway
     env:
       SOURCEGRAPHDOTCOM_MODE: true


rafax:
Do we have a solution for debugging Go with `bazel` commands (last time I checked / [asked](https://sourcegraph.slack.com/archives/C04MYFW01NV/p1710948381791479), we didn't)? Without debugging support, I think we shouldn't be moving Cody Gateway to bazel (as the only command that doesn't support debugging).

rafax:
I think the plan is not to merge this until we find a way to enable debugging with Bazel? Unassigning myself.

PR #61618, opened 2024-04-05
Svelte: add revision to result header and update styling

camdencheek:
This fixes an issue I noticed where we don't show the revision on a search result for non-HEAD repos. We definitely want this because a search like `repo:sourcegraph/sourcegraph rev:*refs/heads/cc/*` (searches for my branches) are pretty useless without this information. We were correctly linking to the revision already.

Additionally, when talking through design options with Taiyab, we just went ahead and paired on a few style updates to bring the search result headers closer to the figma designs.

## Test plan

Screenshots. Take a look.

Repo result without rev:
CleanShot 2024-04-04 at 19 45 46@2x

Repo result with rev:
CleanShot 2024-04-04 at 19 41 54@2x

Code result without rev:

CleanShot 2024-04-04 at 19 46 18@2x

Code result with rev:
CleanShot 2024-04-04 at 19 46 50@2x

Commit result with rev (all commit results have a rev):

CleanShot 2024-04-04 at 19 48 24@2x


client/web-sveltekit/src/lib/search/CodeHostIcon.svelte
```diff
@@ -16,7 +16,7 @@
     export let disableTooltip: boolean = false
 
     $: hostName = repository.split('/')[0]
-    $: svgPath = iconMap[hostName]
+    $: svgPath = iconMap[hostName] ?? mdiSourceRepository
```
camdencheek, line 19: Unrelated: I noticed for nonstandard code hosts, we were not rendering an icon. This adds a fallback icon for codehosts like `ghe.sgdev.org`. Ideally, we would get the code host type from the API and render the icon based on the type rather than the name, but this is better than nothing.

client/web-sveltekit/src/routes/search/RepoRev.svelte
```diff
@@ -0,0 +1,38 @@
...
+
+    $: href = `/${repoName}${rev ? `@${rev}` : ''}`
+    $: displayName = displayRepoName(repoName)
+    $: if (displayName !== repoName) {
+        // We only display part of the repository name, therefore we have to
+        // adjust the match ranges for highlighting
+        const delta = repoName.length - displayName.length
+        highlights = highlights.map(([start, end]) => [start - delta, end - delta])
+    }
+</script>
```
camdencheek, line 18: I extracted the icon/repo name/rev piece into its own component that is reused across all search results except Person.

taiyab (approved):
Looks good. We should make sure we come back round to do those final revision name styling updates.

fkling (approved):

client/web-sveltekit/src/routes/search/SearchResult.svelte
```diff
@@ -43,18 +40,12 @@
...
-        flex-shrink: 0;
-        --color: var(--text-muted);
-    }
-
     .title {
         flex: 1 1 auto;
         overflow: hidden;
         display: flex;
         flex-wrap: wrap;
+        align-items: center;
```
vovakulikov, line 48: @camdencheek I would change it to the `align-items: baseline` instead, just to align the repo name and commit with the file name which has different font and family values. (Note: code host icon probably would require align-self: center)

camdencheek: Oh! Thanks. I was struggling to get alignment right and I couldn't quite figure out why

camdencheek: Wow. that looks so much better. Thanks!

PR #61619, opened 2024-04-05
msp/opentelemetry: downgrade OTEL error logs, add metrics instead

bobheadxi:
We're seeing lots of fairly unactionable OTEL export errors in MSP services - this change downgrades the error logs to warn, to reduce Sentry noise, and adds various metrics (similar to #60775) on the success of exports.

## Test plan

```
sg run msp-example
sg run prometheus # collect metrics
```

Verify the `otel_trace_*` metrics are working. We can't test the metrics equivalent locally - see docstring - but there should be nothing breaking here.


cmd/msp-example/internal/example/example.go
```diff
@@ -141,8 +142,20 @@ func (s Service) Initialize(
...
+						// very interesting. If outgoing, include the host as well.
+						target := r.URL.Path
+						if r.RemoteAddr == "" { // no RemoteAddr indicates this is an outgoing request
+							target = r.Host + target
+						}
+						if operation != "" {
+							return fmt.Sprintf("%s.%s %s", operation, r.Method, target)
+						}
+						return fmt.Sprintf("%s %s", r.Method, target)
+					})),
```
bobheadxi, line 158: copy-pasta from the main sg/sg formatter

PR #61621, opened 2024-04-05
svelte: Refactor React<->SvelteKit routing integration

fkling:
## Context

In order for the SvelteKit app to work in production and to ensure that navigating between the React and SvelteKit app "just works" a couple of things need to happen:

- the server needs to know when to server the SvelteKit app for a given route (if at all).
- when navigating to a new page, the React needs to know whether that page should be served by the SvelteKit app or not. If it should it performs a page refresh, which will cause the server to send the SvelteKit app.
- the same applies to the SvelteKit app, but vice versa, except that it performs a page refresh automatically for every route that it doesn't support (the SvelteKit app is a subset of the React app).

Until now this logic was controlled by the `web-next` and `web-next-rollout` feature flags.
When `web-next` is set the server will serve the SvelteKit app for all routes supported by the app.
When `web-next-rollout` is set the server will server the SvelteKit app only for a certain subset of pages.

These two sets of route information are currently maintained separately on the server, the React app and the SvelteKit app.

## Changes

The current fragmentation doesn't make it easy to enable SvelteKit for additional routes or to add SvelteKit-only pages.

This commit refactors the code to leverage `window.context`, which is now also available in the SvelteKit app, to simplify the mechnism of route enablement.

High level explanation: The server sends a list of route names for which the SvelteKit app is enabled. The clients are responsible for mapping their routes to the server-sent route names and behave accordingly. This allows us to completely get rid of checking the `web-next` and `web-next-rollout` feature flags on the client for making navigation decisions.
We also only need to maintain the different lists of enabled routes on the server (which is done declaratively during route creation).

#### Route enablement on the server

Routes can now be marked as "SvelteKit" ready by passing them to `sk.Register` upon creation. This function also allows to specify the condition for when the route should be rendered by SvelteKit. Currently these conditions map back to the `web-next` and `web-next-rollout` feature flags. Example:

```go
sk.Register(
    repoRev.Path("/tree{Path:.*}").Methods("GET").Name(routeTree).Handler(...),
    sveltekit.EnableOptIn
)
```

#### Route mapping in the SvelteKit app

For each route we can now add a `_meta` export to the corresponding `+page.ts` or `+layout.ts` file that specifies the corresponding server route name. When the user navigates to a route that is not in the list of enabled SvelteKit routes the SvelteKit client will perform a page refresh.

Example:

```ts
export const _meta: RouteMeta = {
    serverRouteName: 'search'
}
```

#### Route mapping in the React app

Due to how routing works in the React app we have to use regular expression to map the requested pathname to the corresponding server route name. This is complicated by the fact that the repo route is essentially a "catch all" route.

```ts
const routeMap = [
    {
        route: PageRoutes.RepoContainer,
        serverRoute: 'tree',
        pathMatch: /^\/.+?\/-\/tree\//,
    }
]
```

#### More detailed details

In order to load the route meta data in the SvelteKit app I use vite's `import.meta.glob` feature, which allows us to import a export from multiple files (those get properly tree shaken in production). To make this work with pages and layouts that don't have such an export I added a custom plugin that emulates an empty export.
As mentioned in the docs, tree shaking works with `import.meta.glob` so that only the the imported values is included.

<details>
<summary>Code generated for production</summary>
<pre>
const up = {},
lp = {},
cp = {},
fp = {
  serverRouteName: 'repo'
},
pp = {
  serverRouteName: 'blob'
},
hp = {
  serverRouteName: 'tree'
},
dp = {},
vp = {
  serverRouteName: 'repo-branches'
},
mp = {},
yp = {},
gp = {
  serverRouteName: 'repo-commit'
},
bp = {
  serverRouteName: 'repo-commits'
},
kp = {
  serverRouteName: 'repo-stats'
},
_p = {},
wp = {
  serverRouteName: 'repo-tags'
},
Sp = {};
// ....
</pre>
</details>

### Caveats

Currently this whole system is based on server route names. However, in theory we have a single page application and there are quite a few routes that don't have dedicated route definitions on the server (e.g. site admin sub pages).
Therefore it's possible that we don't have the granuliarity that we would need. In the future we might have even less granularity. At the moment however I think this approach is fine. Eventually the server has to decide whether to serve the React app or the SvelteKit app and it can only do that if it performs the (initial) routing itself.

### Alternatives considered

I tried a couple of different approaches which I ultimately disregarded. I'll keep this brief and in no particular order:

- Instead of sending route names the server sents regular expressions to mach paths. The idea was that both apps could use the same data to make routing decisions and no client side changes would be necessary to add a new route. However it felt strange to let the server define patterns that ultimately depend/are controlled (to some degree) on the client. This also didn't work well with marking routes as "SvelteKit enabled" right where they are defined. I was also worried that we might run into a sitution that would require more custom route mapping behavior in the React app, that would then break this abstraction.
- Instead of using vite's `import.meta.glob` feature I tried to use a virtual module and do all the data extraction and code generation in a plugin. The benefit of this is that all data processing is done at built time. The disadvantage is that there wasn't a straightforward way to rerun the code generation  when any of the "+page.ts/+layout.ts" files were changed, added or removed. Doing out own parsing and code generation also introduces a lot more complexity and is not worth the tradeoff IMO.

## Test plan

Manual testing running a local instance.

1. Enabled `web-next-rollout` feature flag.
2. A fresh load of a SvelteKit-enabled site serves the SvelteKit application.
3. In the SvelteKit app:
    1. Navigating to supported but disabled route loads the React application
5. In the React app:
    1. Navigating to SK-supported route loads the SvelteKite app
    2. Navigating to SK-supported but disabled route stays within the React (with no page refresh)
    3. The beaker icon is shown only for pages that are _supported_ by the SK app
    4. Clicking the beaker icon loads the SvelteKit app.

----

1. Enabled `web-next` feature flag.
2. Same as above

----

1. Enabled `web-next` and disable `web-next-rollout`.
2. Going to `/search`, which is marked with `EnableRollout`, loads the SK app.

----

1. Disable `web-next` and `web-next-rollout`.
2. Going to `/search` loads the React app.


cmd/frontend/internal/app/ui/router.go
```diff
@@ -255,54 +259,54 @@ func InitRouter(db database.DB) {
...
 			// e.g. "mux.go - gorilla/mux - Sourcegraph"
 			fileName := path.Base(mux.Vars(r)["Path"])
 			return brandNameSubtitle(fileName, repoShortName(c.Repo.Name))
-		})))
+		}))), sveltekit.EnableOptIn)
 
 	// raw
 	repoRev.Path("/raw{Path:.*}").Methods("GET", "HEAD").Name(routeRaw).Handler(handler(db, serveRaw(logger, db, gitserver.NewClient("http.raw"))))
 
-	repo := r.PathPrefix(repoRevPath + "/" + routevar.RepoPathDelim).Subrouter()
```
fkling, line 284: This is exactly the same as `repoRev` in line 261. I don't think there is a reason to created a second subrouter. I assume this was an oversight.

cmd/frontend/internal/app/ui/router.go
```diff
@@ -255,54 +259,54 @@ func InitRouter(db database.DB) {
 			return
 		}
 		serveRepoHandler.ServeHTTP(w, r)
-	}))
+	})), sveltekit.EnableOptIn)
 
-	// tree
 	repoRev := r.PathPrefix(repoRevPath + "/" + routevar.RepoPathDelim).Subrouter()
-	repoRev.PathPrefix("/commits").Methods("GET").Name("repo-commits").Handler(brandedNoIndex("Commits"))
```
fkling, line 262: I moved this into the "table" below. I don't know if the order matters but I don't see an obvious reason why `/commits` should be different than any of the other routes.

cmd/frontend/internal/app/ui/router.go
```diff
@@ -255,54 +259,54 @@ func InitRouter(db database.DB) {
...
 			return brandNameSubtitle(fileName, repoShortName(c.Repo.Name))
-		})))
+		}))), sveltekit.EnableOptIn)
 
 	// raw
 	repoRev.Path("/raw{Path:.*}").Methods("GET", "HEAD").Name(routeRaw).Handler(handler(db, serveRaw(logger, db, gitserver.NewClient("http.raw"))))
 
-	repo := r.PathPrefix(repoRevPath + "/" + routevar.RepoPathDelim).Subrouter()
-
-	repo.PathPrefix("/batch-changes").Methods("GET").Name("repo-batch-changes").Handler(brandedIndex("Batch Changes"))
```
fkling, line 286: I moved that into the table below.

fkling:
Closing this for now because I'll investigate yet another approach.

Issue #60561, opened 2024-02-15
May - Cody backend work


cbart:

aramaraju:
@kalanchan Given that it is a configuration and it is optional, do we need a feature flag at all? Unless I'm mistaken, the fields won't even be shown on the server unless they enter it themselves.  @taras-yemets thoughts?

PR #61641, opened 2024-04-05
Cody: add repo-level context filters support

taras-yemets:
Closes https://github.com/sourcegraph/sourcegraph/issues/61606

1. Use different context filters for PLG and enterprise users.
2. Implement repo-level Cody context filter based on the site config rules.

Follow-up work:
- Use the correct matcher for PLG: https://github.com/sourcegraph/sourcegraph/issues/61816

Notes to reviewers:
1. Context filter logic depending on `.cody/.ignore` file remains unchanged, it was moved as-is to a separate file (`dotcom.go`).

## Test plan
- CI
- Added unit tests
- Tested manually: both enterprise and PLG scenarios work


cmd/frontend/internal/context/init.go
```diff
@@ -35,7 +35,7 @@ func Init(
 		embeddingsClient,
 		searchClient,
 		getQdrantSearcher,
-		codycontext.NewCodyIgnoreFilter(services.GitserverClient.Scoped("codycontext.ignore")),
+		codycontext.NewRepoContentFilter(services.GitserverClient.Scoped("codycontext.ignore")),
```
chwarwick, line 38: Is the context filter supposed to be dependent on dotcom/enterprise?

taras-yemets: Yes
https://github.com/sourcegraph/sourcegraph/blob/880beb650f84443b607057637cdca5e99b6eea96/internal/codycontext/filter.go#L15-L20

taras-yemets: Enterprise instances use site config to define include/exclude rules and ignore `.cody/.ignore` files (will be communicated via changelog and docs), PLG users can use only `.cody/.ignore` files.

keegancsmith, line 38: design: if repocontentfilter is only used by codycontextclient and they live in the same package, then I'd consider unexporting repocontentfilter and initializing it inside of codycontextclient.

internal/codycontext/dotcom.go
```diff
@@ -0,0 +1,167 @@
...
+			return fcc
+		}
+	}
+	return f.getFilter(repos, logger)
+}
+
+// getFilter returns the list of repos that can be filtered
+// their .cody/ignore files (or don't have one). If an error
+// occurs that repo will be excluded.
+func (f *dotcomRepoFilter) getFilter(repos []types.RepoIDName, logger log.Logger) ([]types.RepoIDName, FileChunkFilterFunc) {
```
keegancsmith, line 65: how big can this list of repos be? The code below won't scale to large sizes, so just checking this is relatively small.

internal/codycontext/dotcom.go
```diff
@@ -0,0 +1,167 @@
...
+	filters := make(map[api.RepoName]filterFunc, len(repos))
```
keegancsmith, line 66: minor: if possible, I'd key by repo ID. It's more stable and avoids issues around case sensitivity/etc.

internal/codycontext/dotcom.go
```diff
@@ -0,0 +1,167 @@
...
+			logger.Warn("repoContextFilter: couldn't get default branch, removing repo", log.Int32("repo", int32(repo.ID)), log.Error(err))
+			continue
+		}
+		// No commit signals an empty repo, should be nothing to filter
+		// Also we can't lookup the ignore file without a commit
+		if commit == "" {
+			logger.Info("repoContextFilter: empty repo removing", log.Int32("repo", int32(repo.ID)))
+			continue
+		}
+		matcher, err := getIgnoreMatcher(ctx, f.cache, f.client, repo, commit)
```
keegancsmith, line 83: can you just ask for `HEAD` as the commit here and avoid GetDefaultBranch call?

internal/codycontext/dotcom.go
```diff
@@ -0,0 +1,167 @@
...
+		}
+		return nil, err
+	}
+	defer fr.Close()
+
+	ignoreFileBytes, err := io.ReadAll(fr)
+	if err != nil {
+		return nil, err
+	}
+	ig, err := ignore.ParseIgnoreFile(bytes.NewReader(ignoreFileBytes))
```
keegancsmith, line 161: this is the ignore file used by search. Are you sure this is the same format you need to support?

Also you can pass fr directly to ParseIgnoreFile, you don't need to ReadAll first.

internal/codycontext/dotcom.go
```diff
@@ -0,0 +1,167 @@
...
+
+func isEnabled(c *conf.Unified) bool {
+	if c != nil && c.ExperimentalFeatures != nil && c.ExperimentalFeatures.CodyContextIgnore != nil {
+		return *c.ExperimentalFeatures.CodyContextIgnore
+	}
+	return false
+}
+
+func getIgnoreMatcher(ctx context.Context, cache safeCache[repoRevision, ignore.Matcher], client gitserver.Client, repo types.RepoIDName, commit api.CommitID) (*ignore.Matcher, error) {
+	cached, ok := cache.Get(repoRevision{Repo: repo, Commit: commit})
```
keegancsmith, line 136: it looks like you read and write from cache concurrently. I'm not sure if cache is concurrency safe?

internal/codycontext/dotcom.go
```diff
@@ -0,0 +1,167 @@
...
+func newDotcomFilter(client gitserver.Client) RepoContentFilter {
+	enabled := isEnabled(conf.Get())
+	//c, _ := lru.New[repoRevision, ignore.Matcher](128)
+	ignoreFilter := &dotcomRepoFilter{
+		cache:   newSafeCache[repoRevision, ignore.Matcher](128),
+		client:  client,
+		enabled: enabled,
+	}
+
+	go conf.Watch(func() {
```
keegancsmith, line 121: this shouldn't be a goroutine.

internal/codycontext/dotcom.go
```diff
@@ -0,0 +1,167 @@
...
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	f.enabled = enabled
+}
+
+func (f *dotcomRepoFilter) GetEnabled() bool {
+	f.mu.RLock()
+	defer f.mu.RUnlock()
+	return f.enabled
+}
```
keegancsmith, line 51: these functions are only used  internally, they shouldn't be exported.

internal/codycontext/dotcom.go
```diff
@@ -0,0 +1,167 @@
...
+		}
+
+		filters[repo.Name] = matcher.Match
+		filterableRepos = append(filterableRepos, repo)
+	}
+
+	return filterableRepos, func(fcc []FileChunkContext) []FileChunkContext {
+		filtered := make([]FileChunkContext, 0, len(fcc))
+		for _, fc := range fcc {
+			remove, ok := filters[fc.RepoName]
```
keegancsmith, line 96: very minor: a better name for remove I think is `ignore`. We are parsing an ignore file and `if !ignore(fc.Path)` reads better to me.

internal/codycontext/dotcom.go
```diff
@@ -0,0 +1,167 @@
...
+
+	ignoreFileBytes, err := io.ReadAll(fr)
+	if err != nil {
+		return nil, err
+	}
+	ig, err := ignore.ParseIgnoreFile(bytes.NewReader(ignoreFileBytes))
+	if err != nil {
+		return nil, err
+	}
+	cache.Add(repoRevision{Repo: repo, Commit: commit}, *ig)
```
keegancsmith, line 165: why not just store as a pointer in the cache? At write time you dereference and at read time you convert back into a pointer.

internal/codycontext/enterprise.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// newEnterpriseFilter creates a new RepoContentFilter that filters out
+// content based on the Cody context filters value in the site config.
+func newEnterpriseFilter() RepoContentFilter {
+	filter := &enterpriseRepoFilter{
+		cache: newSafeCache[api.RepoName, bool](128),
+		ccf:   conf.Get().SiteConfiguration.CodyContextFilters,
+	}
+	go conf.Watch(func() {
```
keegancsmith, line 29: not a goroutine.

internal/codycontext/enterprise.go
```diff
@@ -0,0 +1,95 @@
...
+
+// isRepoAllowed checks if repo name matches Cody context include and exclude rules from the site config and stores result in cache.
+func (f *enterpriseRepoFilter) isRepoAllowed(repoName api.RepoName) bool {
+	cached, ok := f.cache.Get(repoName)
+	if ok {
+		return cached
+	}
+
+	f.mu.RLock()
+	defer f.mu.RUnlock()
```
keegancsmith, line 66: you shouldn't be holding this mutex while doing matching. You should only be using this mutex to read in the value of f.ccf.

internal/codycontext/filter.go
```diff
@@ -1,191 +1,20 @@
...
-	ig, err := ignore.ParseIgnoreFile(bytes.NewReader(ignoreFileBytes))
-	if err != nil {
-		return nil, err
+func NewRepoContentFilter(client gitserver.Client) RepoContentFilter {
+	if dotcom.SourcegraphDotComMode() {
+		return newDotcomFilter(client)
 	}
-	cache.Add(repoRevision{Repo: repo, Commit: commit}, *ig)
-	return ig, nil
+	return newEnterpriseFilter()
```
keegancsmith, line 19: shouldn't the enterprise filter also be respecting ignore files inside the repo?

keegancsmith (commented):
Nice to see this take shape. Got quite a bit of feedback. At a high level this is missing observability. In particular prometheus metrics around performance and cache hits/misses would be very useful here. Additionally given you are refactoring some existing functionality around "dotcomfilter" I would hope for some sort of manual testing. The current tests look very unit test like so I am not confident at the integration time that every works well.

internal/codycontext/filter.go
taras-yemets: No, it shouldn't ([RFC](https://docs.google.com/document/d/1PMeJHvY0ia2nMewFvs4vrFcmDbRDw9kSyljil-K8Pt8/edit#bookmark=id.iewr8p3kry01)).

internal/codycontext/enterprise.go
```diff
@@ -0,0 +1,95 @@
...
+	if f.ccf == nil {
+		return allowByDefault
+	}
+
+	allowed := allowByDefault
+
+	if len(f.ccf.Include) > 0 {
+		for _, p := range f.ccf.Include {
+			include := regexp.MustCompile(p.RepoNamePattern).MatchString(string(repoName))
```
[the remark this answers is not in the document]
taras-yemets: https://github.com/sourcegraph/sourcegraph/blob/d4058e9ac6b5e9a67b7d6f21ece5fa0161b658c3/internal/conf/validate.go#L523-L529

Site config is validated on save (regexes are checked to [match RE2 syntax](https://pkg.go.dev/github.com/xeipuuv/gojsonschema#section-readme:~:text=regex.%20Go%20uses%20the%20RE2%20engine%20and%20is%20not%20ECMA262%20compatible.)), so I don't expect invalid patterns here. Though I agree that we'd better validate them.

cmd/frontend/internal/context/resolvers/context_test.go
```diff
@@ -3,6 +3,10 @@ package resolvers
 import (
 	"bytes"
 	"context"
+	"github.com/sourcegraph/sourcegraph/cmd/frontend/graphqlbackend"
+	"github.com/sourcegraph/sourcegraph/internal/codycontext"
+	"github.com/sourcegraph/sourcegraph/internal/dotcom"
+	"github.com/sourcegraph/sourcegraph/internal/observation"
```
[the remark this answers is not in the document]
taras-yemets: https://github.com/sourcegraph/sourcegraph/pull/61641/commits/0a5a77887c1cabceef5b429fc3152882168c822e

internal/codycontext/enterprise.go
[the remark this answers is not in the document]
taras-yemets: Refactored to validate patterns in `configure` method.

internal/codycontext/dotcom.go
taras-yemets: Fixed

cmd/frontend/internal/context/resolvers/context_test.go
```diff
@@ -208,38 +214,65 @@ func TestContextResolver(t *testing.T) {
...
+			want: []string{"testcode1.go", "ignore_me1.go", "ignore_me1.md"},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			dotcom.MockSourcegraphDotComMode(t, tt.dotComMode)
+
+			contextClient := codycontext.NewCodyContextClient(
+				observation.NewContext(logger),
```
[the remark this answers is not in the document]
taras-yemets: Fixed https://github.com/sourcegraph/sourcegraph/pull/61641/commits/f9c7d2efd6899d06ca49e4a6b26a9fa1c5131936

internal/codycontext/dotcom.go
taras-yemets: Renamed

internal/codycontext/dotcom.go
taras-yemets: Fixed

cmd/frontend/internal/context/init.go
taras-yemets: Addressed in https://github.com/sourcegraph/sourcegraph/pull/61641/commits/7f313cec6a8c7292f89edcc1ed8e41d0875b996b

internal/codycontext/cache.go
```diff
@@ -0,0 +1,36 @@
+package codycontext
+
+import (
+	lru "github.com/hashicorp/golang-lru/v2"
+)
+
+type safeCache[K comparable, V any] struct {
```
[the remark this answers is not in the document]
taras-yemets: Removed in https://github.com/sourcegraph/sourcegraph/pull/61641/commits/7a9674565f4baa0040c7f44d239bf52d5faeeabf

internal/codycontext/dotcom.go
taras-yemets: `getIgnoreMatcher` reads `.cody/.ignore` file content at a specific commit and creates an ignore matcher based on this file. Given that the ignore file content at a specific commit is stable, it should be safe to read and write to the cache concurrently. WDYT?

internal/codycontext/enterprise.go
```diff
@@ -0,0 +1,95 @@
...
+	allowedRepos := make([]types.RepoIDName, 0, len(repos))
+	for _, repo := range repos {
+		if f.isRepoAllowed(repo.Name) {
+			allowedRepos = append(allowedRepos, repo)
+		}
+	}
+	return allowedRepos, func(fcc []FileChunkContext) []FileChunkContext {
+		filtered := make([]FileChunkContext, 0, len(fcc))
+		for _, fc := range fcc {
+			isFromAllowedRepo := slices.ContainsFunc(allowedRepos, func(r types.RepoIDName) bool { return r.Name == fc.RepoName })
```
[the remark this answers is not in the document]
taras-yemets: Addressed in https://github.com/sourcegraph/sourcegraph/pull/61641/commits/da180d657b28f1185dbbb1f72402a9cc440f8232

internal/codycontext/dotcom.go
```diff
@@ -0,0 +1,168 @@
...
+	}, true
+}
+
+// newDotcomFilter creates a new RepoContentFilter that filters out
+// content based on the .cody/ignore file at the head of the default branch
+// for the given repositories.
+func newDotcomFilter(client gitserver.Client) RepoContentFilter {
+	enabled := isEnabled(conf.Get())
+	// ignore error since it only happens if cache size is not positive
+	cache, _ := lru.New[repoRevision, ignore.Matcher](128)
```
taras-yemets, line 115: This filter is used only for dotcom. The cache with size 128 doesn't seem helpful. Should we consider removing it, @keegancsmith?

taras-yemets: cc: @chwarwick

internal/codycontext/dotcom.go
taras-yemets: Nice catch!
Docs [claim](https://sourcegraph.com/docs/cody/capabilities/ignore-context#using-the-ignore-context-feature:~:text=The%20.codyignore%20file%20works%20similarly%20to%20a%20.gitignore%20file.) that `.cody/.ignore` file behaves similarly to `.gitignore` files, but the [ignore matcher we use for search](https://github.com/sourcegraph/zoekt/blob/c39011a14191514737c52ab3f0e06348915c087e/ignore/ignore.go) doesn't follow the [gitignore spec](https://git-scm.com/docs/gitignore). 

Created an issue to address it separately: https://github.com/sourcegraph/sourcegraph/issues/61816

internal/codycontext/dotcom.go
taras-yemets: As I understand the current implementation we call `GetDefaultBranch` to get commit ID. 

We use commit ID to differentiate 2 cases:
- when [repo/rev doesn't exist or other errors](https://github.com/sourcegraph/sourcegraph/blob/0b5e7fd490e920c2a3f8eb921f41b94006ef3233/internal/gitserver/commands.go#L1166-L1175) - we want to [ignore such repos](https://github.com/sourcegraph/sourcegraph/blob/5f7ca4e1ccfdb7bdc408b46d607d0130f89adfa5/internal/codycontext/dotcom.go#L70-L88)
- when repo/rev/commit exist, but `.cody/.ignore` file doesn't - we want to [allow such repos](https://github.com/sourcegraph/sourcegraph/blob/5f7ca4e1ccfdb7bdc408b46d607d0130f89adfa5/internal/codycontext/dotcom.go#L146-L153).

internal/codycontext/dotcom.go
taras-yemets: I don't expect this list to be big. 
E.g., for VSCode we query for a list of automatically (repos from workspace) and manually added (repo selector) repos. Not sure about the exact number of repos user can add manually (I'd be very surprised if more than 10) and if PLG users even have those repo selector (this `getFilter` will be used only for dotcom, and I currently don't have it in my VSCode). 
We also limit the number of code and text results in the context search query (VSCode extension currently [queries for 15 code and 5 text](https://github.com/sourcegraph/cody/blob/a109df44a5fb4501fea923dbaa71fb794d188665/lib/shared/src/sourcegraph-api/graphql/client.ts#L514-L515) results).  
I think the JetBrains extension has similar logic.  

Though we don't limit the number of repos in the [resolver](https://github.com/sourcegraph/sourcegraph/blob/341998d8749e96caedc1744dc36faec9967f50a4/cmd/frontend/internal/context/resolvers/context.go#L31) so if a consumer queries the endpoint with big number of repos and code/text results count, the efficiency/scalability problem will become more obvious.

internal/codycontext/dotcom.go
taras-yemets: removed from dotcom https://github.com/sourcegraph/sourcegraph/pull/61641/commits/c66902b1c8f944911581d3c07492a842073cca10

internal/codycontext/dotcom.go
taras-yemets: Removed cache from dotcom filter - https://github.com/sourcegraph/sourcegraph/pull/61641#discussion_r1561047662

internal/codycontext/dotcom.go
keegancsmith: I think as a follow-up PR you should enforce a hard limit somewhere and document why. Indeed it does sound like the number should be relatively small so this is acceptable. But the hard limit will save your ass further down the line and pre-empt conversations around if we want larger contexts.

internal/codycontext/dotcom.go
taras-yemets: Removed cache from dotcom filter - https://github.com/sourcegraph/sourcegraph/pull/61641#discussion_r1561047662

internal/codycontext/dotcom.go
keegancsmith: You should be able to tell this at the time of asking for the file based on the type of error returned (we have different errors for repo doesn't exist and file doesn't exist). However, our support of error codes can be a bit inconsistent in our git client. I'd encourage cleaning up our git cli to do this, but this is a minor point so I'd follow-up rather and get this merged sooner :)

internal/codycontext/dotcom.go
keegancsmith: cc @stefanhengl supporting the same format as .gitignore for search may be desirable if its backwards compatible with our current format. So if you find that to be the case would be cool to ping us to also implement.

internal/codycontext/dotcom.go
taras-yemets: Do you mean that the [resolver](https://github.com/sourcegraph/sourcegraph/blob/341998d8749e96caedc1744dc36faec9967f50a4/cmd/frontend/internal/context/resolvers/context.go#L31) should return a descriptive error if the repos number passed in args is greater than N?

internal/codycontext/context.go
```diff
@@ -122,7 +123,10 @@ func (c *CodyContextClient) GetCodyContext(ctx context.Context, args GetContextA
 
 	// Generating the content filter removes any repos where the filter can not
 	// be determined
-	filterableRepos, contextFilter := c.contentFilter.GetFilter(args.Repos, c.obsCtx.Logger)
+	filterableRepos, contextFilter, ok := c.contentFilter.GetFilter(args.Repos, c.obsCtx.Logger)
```
keegancsmith, line 126: minor: you have this obsCtx at the time of contentFilter creation. You should probably store the logger on the structs that need it rather than pass it around now at request time.

internal/codycontext/context.go
```diff
@@ -122,7 +123,10 @@ func (c *CodyContextClient) GetCodyContext(ctx context.Context, args GetContextA
...
+	if !ok {
+		return nil, errors.New("failed to create content filter")
```
keegancsmith, line 128: this error isn't very useful. I think this will end up in clients, so I would make it more informative. Something along the lines of telling users to contact there admin and tell them they have invalid cody ignore settings.

internal/codycontext/dotcom.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+		// No commit signals an empty repo, should be nothing to filter
+		// Also we can't lookup the ignore file without a commit
+		if commit == "" {
+			logger.Info("repoContextFilter: empty repo removing", log.Int32("repo", int32(repo.ID)))
+			continue
+		}
+		matcher, err := getIgnoreMatcher(ctx, f.client, repo, commit)
+		if err != nil {
+			logger.Warn("repoContextFilter: unable to process ignore file", log.Int32("repo", int32(repo.ID)), log.Error(err))
```
keegancsmith, line 76: Have you ever checked dotcom logs to see these log lines and others? I would be interested to know if you actually use them.

internal/codycontext/dotcom.go
```diff
@@ -0,0 +1,144 @@
...
+func (f *dotcomRepoFilter) getFilter(repos []types.RepoIDName, logger log.Logger) (_ []types.RepoIDName, _ FileChunkFilterFunc, ok bool) {
+	filters := make(map[api.RepoID]filterFunc, len(repos))
+	filterableRepos := make([]types.RepoIDName, 0, len(repos))
+	// use the internal actor to ensure access to repo and ignore files
+	ctx := actor.WithInternalActor(context.Background())
+	for _, repo := range repos {
+
+		_, commit, err := f.client.GetDefaultBranch(ctx, repo.Name, true)
+		if err != nil {
+			logger.Warn("repoContextFilter: couldn't get default branch, removing repo", log.Int32("repo", int32(repo.ID)), log.Error(err))
```
keegancsmith, line 65: minor: you shouldn't need to prefix your log messages with `repoContextFilter`. Instead you have be using a logger scoped to your component.

internal/codycontext/dotcom.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// getFilter returns the list of repos that can be filtered
+// their .cody/ignore files (or don't have one). If an error
+// occurs that repo will be excluded.
+func (f *dotcomRepoFilter) getFilter(repos []types.RepoIDName, logger log.Logger) (_ []types.RepoIDName, _ FileChunkFilterFunc, ok bool) {
+	filters := make(map[api.RepoID]filterFunc, len(repos))
+	filterableRepos := make([]types.RepoIDName, 0, len(repos))
+	// use the internal actor to ensure access to repo and ignore files
+	ctx := actor.WithInternalActor(context.Background())
```
keegancsmith, line 60: Now that you have removed the cache, you should pass in the request context and use that instead of bypassing auth.

internal/codycontext/enterprise.go
```diff
@@ -0,0 +1,174 @@
...
+		Help: "Incremented each time we have a cache miss on cody context filters.",
+	})
+)
+
+type filterItem struct {
+	RepoNamePattern *regexp.Regexp
+}
+
+type filtersConfig struct {
+	cache   *lru.Cache[api.RepoID, bool]
```
keegancsmith, line 36: There is a decent chance this cache is also not that useful given you say the size of a list of repos from the client is relatively small. However, lets leave it in and consult the prometheus metrics in the future to determine if its atleast doing something. I think the main thing this protects against is an admin generating a very large config. But I'd be tempted to just not have the cache and rather monitor response times.

keegancsmith (approved):
I agree the dotcom cache was likely not useful. I suspect the most expensive part before was reading the ignore file from git which I imagine was a pretty minor part of the work done in a codycontext request.

internal/codycontext/context.go
taras-yemets: Nice catch!
I already [store it on an enterprise filter struct](https://github.com/sourcegraph/sourcegraph/blob/da180d657b28f1185dbbb1f72402a9cc440f8232/internal/codycontext/enterprise.go#L51). Will refactor.
Thank you!

internal/codycontext/dotcom.go
taras-yemets: Addressed in https://github.com/sourcegraph/sourcegraph/pull/61641/commits/66019888d08f307edbbc96b76d74846c6dcef726

internal/codycontext/context.go
taras-yemets: That makes sense, thank you!
The error message text depends on the filter used (enterprise or dotcom), so I'd probably need to pass an actual error instead of the success indicator.
Will fix!

taras-yemets: Addressed in https://github.com/sourcegraph/sourcegraph/pull/61641/commits/cbf6a5fa5b5cb0b5f7bc6713b90793b668d0ae8e

internal/codycontext/dotcom.go
keegancsmith: yes

internal/codycontext/dotcom.go
taras-yemets: I didn't find matching logs on dotcom for the last 30 days. Though it's expected given that [search](https://sourcegraph.com/search?q=context:global+path:%22.cody/ignore%22&patternType=keyword&sm=0) for `.cody/ignore` returns only one result.

internal/codycontext/dotcom.go
taras-yemets: https://github.com/sourcegraph/sourcegraph/issues/61878

internal/codycontext/dotcom.go
taras-yemets: Added an item to the tracking issue:
- https://github.com/orgs/sourcegraph/projects/364/views/1?filterQuery=assignee:@me&pane=issue&itemId=53381339

internal/codycontext/dotcom.go
taras-yemets: Addressed in https://github.com/sourcegraph/sourcegraph/pull/61641/commits/b3cd88f6b1b4462fc527eee1bb34be8af3cea6fc

internal/codycontext/enterprise.go
taras-yemets: Added to the tracking issue:
- https://github.com/sourcegraph/sourcegraph/issues/60561

Issue #61613, opened 2024-04-04
Svelte: timer not reset when a new search is started


camdencheek:
https://github.com/sourcegraph/sourcegraph/assets/12631702/ff2e31cc-e632-4f1c-b3f0-357fbd9791ca

PR #61651, opened 2024-04-05
Svelte [results indicator]: Remove timer until search completes

jasonhawkharris:
Removes the live updating timer as it introduced tech debt that would be difficult to maintain. Page overall now looks less busy as a result. Solves #61613.
 
## Test plan
manual testing


camdencheek (approved):
cc @taiyab because this is a departure from the designs.

We removed the live-updating "current time taken" because:
1) We cannot trust the backend timings since they come intermittently and do not represent user experience
2) Measuring in the client only is tricky because the time the user experiences is different than the time the search takes due to preloading
3) An additional moving element on the page made it feel more jumpy and and visually noisy and the "current time taken" is not really very valuable information. 

We do still show the total time the search took once it has completed though.

PR #61657, opened 2024-04-05
Svelte [ResultsIndicator]: Add 'Search Jobs' message to popover

jasonhawkharris:
This PR introduces a "Search Jobs" suggestion in the popover when necessary. Additionally, it includes the suggestion in the results indicator.

The message displayed in the results indicator: 
Screenshot 2024-04-05 at 1 04 22 PM

The message displayed in the popover:
Screenshot 2024-04-05 at 1 05 50 PM

Both of these messages will appear simultaneously when the popover is active. Although this simultaneous display shouldn't pose an issue, I'd appreciate thoughts on this approach.

The message will display under the following conditions:

The search is completed with error.
The search is in a loading state.

For reference, here's how it looks when the message does not display: 
Screenshot 2024-04-05 at 1 07 58 PM

## Test plan
Adjust existing unit tests.

### DIFF
--- a/client/web-sveltekit/src/lib/search/resultsIndicator/SuggestedAction.svelte
+++ b/client/web-sveltekit/src/lib/search/resultsIndicator/SuggestedAction.svelte
@@ -49,6 +49,12 @@
                 </small>
             </div>
         {/if}
+        <!--
+        TODO: @jasonhawkharris - When we implement search jobs,
+        we can change the link so that it points to where a user
+        can actually create a search job. We should also change
+        the text of the link when we do so, "Create a search job"
+        -->
         {#if severity === 'error' && !Object.hasOwn(mostSevere, 'suggested')}
             <div class="separator">{CENTER_DOT}</div>
             <div class="search-job-link">
--- a/client/web-sveltekit/src/lib/search/resultsIndicator/SuggestedAction.svelte
+++ b/client/web-sveltekit/src/lib/search/resultsIndicator/SuggestedAction.svelte
@@ -49,6 +49,14 @@
                 </small>
             </div>
         {/if}
+        {#if severity === 'error' && !Object.hasOwn(mostSevere, 'suggested')}
+            <div class="separator">{CENTER_DOT}</div>
+            <div class="search-job-link">
+                <small>
+                    Use <a href="/help/code-search/types/search-jobs">Search Job</a> for background search.
+                </small>
+            </div>
+        {/if}
     </div>
 </div>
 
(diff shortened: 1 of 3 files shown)


client/web-sveltekit/src/lib/search/resultsIndicator/SuggestedAction.svelte
```diff
@@ -49,6 +49,14 @@
                 </small>
             </div>
         {/if}
+        {#if severity === 'error' && !Object.hasOwn(mostSevere, 'suggested')}
+            <div class="separator">{CENTER_DOT}</div>
+            <div class="search-job-link">
+                <small>
+                    Use <a href="/help/code-search/types/search-jobs">Search Job</a> for background search.
```
peterguy, line 56: Suggested alternate wording: `Create a <a>Search Job</a>`. Keep it short and sweet. Placing it next to the alert/warning message gives it enough context to pique curiosity and let the user know that it's a way to address the issue.

peterguy: I didn't realize that the link is to help docs; can we use the link to actually create a Search Job?

jasonhawkharris: Added the following comment for when we get around to implementing search jobs in the sveltekit re-write:

```   
        <!--
        TODO: @jasonhawkharris - When we implement search jobs,
        we can change the link so that it points to where a user
        can actually create a search job. We should also change
        the text of the link when we do so, "Create a search job"
        -->
```

peterguy (approved):
Great first pass!

client/web-sveltekit/src/lib/search/resultsIndicator/SuggestedAction.svelte
```diff
@@ -49,6 +49,20 @@
                 </small>
             </div>
         {/if}
+        <!--
+        TODO: @jasonhawkharris - When we implement search jobs,
+        we can change the link so that it points to where a user
+        can actually create a search job. We should also change
+        the text of the link when we do so, "Create a search job"
+        -->
+        {#if severity === 'error' && !Object.hasOwn(mostSevere, 'suggested')}
```
camdencheek, line 58: Looks like this slipped past in a prior review, but I don't think we should use `Object.hasOwn` because it circumvents the type system. Instead, can we use `mostSevere.suggested !== undefined`?

jasonhawkharris: I'll fix this in a subsequent PR

camdencheek: Perfect, it was unrelated anyways. Thanks!

jasonhawkharris: https://github.com/sourcegraph/sourcegraph/pull/61662

PR #61661, opened 2024-04-05
web: Switch to absolute blame recency calculation

fkling:
[This Slack thread](https://sourcegraph.slack.com/archives/C05EA9KQUTA/p1706906484633939) has a lot more information.

In a nutshell:

- We currently compute the recency color relative to either one year ago, three years ago, or the first commit of the repository.
- That means the colors have different meaning across different repositories.
- Fetching the date of the first commit can be expensive/slow.

This commit changes the method to calculate the color to always use the same time frame. The age of the commit is mapped to a number between `[0, 1)` and `d3-scale-chromatic` is used to 
map that value to a color.
We can control the sensibility of the colors for newer/older commits by changing how we do the mapping. In this specific version I chose the midpoint to be ~ 6 months in the past.

Here are a couple of examples for choosing a different midpoint:

| 3 months | 6 months (current) | 12 months |
|--------|--------|--------|
|  |  |  | 

I don't think there is "the right way" to perform the mapping but intuitively I would say that being able to better differentiate more recent commits is more useful than differentiating older commits. The impact the choice of the midpoint has on this is visible in the screenshots above if you compare the colors for the `Feb 1` and `Mar 4` lines.

Side note: Another benefit of using `d3-scale-chromatic` is that we experiment with [different color scales](https://d3js.org/d3-scale-chromatic/sequential).

## Test plan

Manual testing.


client/wildcard/src/global-styles/colors.scss
```diff
@@ -308,18 +308,6 @@ $violet-09: #5f3dc4;
     --chroma-generic-traceback-fg: #aa0000;
     --chroma-text-whitespace-fg: #bbbbbb;
 
-    // Blame colors
```
eseliger, line 311: Nice side effect that we got rid of another global :)

eseliger (approved):

client/web/src/repo/blob/blameRecency.tsx
```diff
@@ -1,42 +1,35 @@
...
-    const start = Math.min(firstCommitDate ? firstCommitDate.getTime() : THREE_YEARS_AGO, ONE_YEAR_AGO)
+    const age = (Date.now() - commitDate.getTime()) / MILLIS_IN_HOUR
 
-    // Get a value between [0, 1] that represents the recency of the commit in a linear scale
-    const recency = Math.min(Math.max((now - commit.getTime()) / (now - start), 0), 1)
-
-    // Map from the linear scale to the exponential scale
-    const index = STEPS.findIndex(step => recency <= step)
+    // Get a value between [1, 0) that represents the recency of the commit
+    // (1 is most recent, 0 is least recent)
```
camdencheek, line 28: Wouldn't 0 be most recent since when age is 0, recency is 0?

camdencheek (approved):
This is great, thanks!

Also agreed that 6 months is a good balance

client/web/src/repo/blob/blameRecency.tsx
fkling: I probably originally subtracted this value from `1` right here and only added the flag to switch the direction of colors later. I'll clean this up a bit.

PR #61662, opened 2024-04-05
Svelte [bug]: adjust conditional to stay consistent with types

jasonhawkharris:
Using `Object.hasOwn()` circumvents the type system, so we shouldn't use this in conditional statements, in general.

The change is self-explanatory in the diff view.

## Test plan
Manual testing

### DIFF
--- a/client/web-sveltekit/src/lib/search/resultsIndicator/SuggestedAction.svelte
+++ b/client/web-sveltekit/src/lib/search/resultsIndicator/SuggestedAction.svelte
@@ -38,7 +38,7 @@
         {/if}
 
         <!-- completed with suggested items -->
-        {#if done && mostSevere && Object.hasOwn(mostSevere, 'suggested')}
+        {#if done && mostSevere && mostSevere.suggested}
             <div class="separator">{CENTER_DOT}</div>
             <div class="action-badge">
                 <small>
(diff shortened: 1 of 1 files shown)


fkling:
General drive-by feedback: Neither the PR title nor the description tell me anything about what the problem actually here is and why this needs to be changed 😅 
I can only imagine that `Object.hasOwn()` doesn't give you desired result for `mostSevere = { suggested: undefined }`.

camdencheek:
@fkling FWIW, this was just a small thing I noticed in [another PR](https://github.com/sourcegraph/sourcegraph/pull/61657#discussion_r1554255898) that Jason is following up on. So not without context for its target audience (me) 🙂 

My complaint is basically that `Object.hasOwn` circumvents the type system when we could just check whether or not the field is undefined.

fkling:
@camdencheek Got it. I don't know how TypeScript treats `Object.hasOwn` but I assume the "danger" is that it can always return `false` because the provided object can never have that property, but TS wouldn't tell you that.

Even worse though (you probably know that but just for completeness), what I was alluding to in my comment: In TS when a property is defined as 

```ts
interface SomeObject {
    optionalField?: number
}
```

then both of these are valid:

```ts
const obj1: SomeObject = {
    optionalField: undefined
}
const obj2: SomeObject = {}
```

because in both cases `objX.optionalField` would return `undefined`.

**However**, `Object.hasOwn(obj1, 'optionalField')` is `true` and `Object.hasOwn(obj2, 'optonalField`)` is `false`, and that would clearly be a bug.

fkling:
JavaScript is an ... interesting language 😉 I would like to say that it's not too bad once you have built a good mental model, but maybe you shouldn't have to do that in the first place 😄 
Anyways, once you accept that accessing non-existing properties returns `undefined` (instead of erroring) and that `undefined` can also be explicitly assigned, this behavior at least "makes sense".

I don't think there is a need to use something like `Object.hasOwn` in TS outside of user defined type guards.

PR #61663, opened 2024-04-05
Context: remove key words from stopwords

jtibshirani:
While digging into why we perform poorly on "Ownership" and "Changelog"
searches, I noticed we were removing key terms in these queries. This PR adapts
the stopwords list to avoid removing them.

This doesn't immediately fix the problem, as we still don't match on filenames
(like CODEOWNERS or CHANGELOG.md). But it's a step towards the solution.

## Test plan

Adapted unit test.

### DIFF
--- a/internal/search/codycontext/stop_words.go
+++ b/internal/search/codycontext/stop_words.go
@@ -1,6 +1,9 @@
 package codycontext
 
 // A large set of stopwords, taken from the ATIRE list here: https://github.com/igorbrigadir/stopwords/tree/master.
+// Custom modifications:
+// - Remove "changes" to improve changelog queries
+// - Remove "own" to improve ownership queries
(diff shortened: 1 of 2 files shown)


internal/search/codycontext/stop_words.go
```diff
@@ -1,6 +1,9 @@
 package codycontext
 
 // A large set of stopwords, taken from the ATIRE list here: https://github.com/igorbrigadir/stopwords/tree/master.
+// Custom modifications:
```
jtibshirani, line 4: Ideally we'd like to avoid this kind of manual tweak. I'm also looking into an ML-based query rewriting step that could really help here. For example, it could suggest "for this query, you should definitely perform a file filter on CHANGELOG".

stefanhengl (approved):

PR #61664, opened 2024-04-05
Context: match on filename

jtibshirani:
Currently, Cody keyword context only matches on file contents. For many queries
it's critical to match the filename too. The reason we only matched contents is
purely historical -- I just chose the simplest implementation at first.

Examples of when this is important:
* "who owns third party licenses?" should match `third-party-licenses/CODEOWNERS`
* "grafana version deps.bzl" should clearly match `deps.bzl`

This PR extends the searches to match filename too. It doesn't immediately
improve results for most Ownership or Changelog queries, because Zoekt does not
count filename matches towards a result's score. I'll fix this in follow-up work. 

## Test plan

Adapted unit tests. Manually tested using local Cody Web using these queries:
* "tell me about admin/repo/perforce.md"
* "who owns third party licenses?"
* "grafana version deps.bzl"

Reran [context-eval](https://github.com/sourcegraph/context-eval) and made sure there is no regression.

### DIFF
--- a/internal/codycontext/context.go
+++ b/internal/codycontext/context.go
@@ -3,7 +3,6 @@ package codycontext
 import (
 	"context"
 	"fmt"
-	"strconv"
 	"strings"
 	"sync"
 
@@ -268,8 +267,8 @@ func (c *CodyContextClient) getKeywordContext(ctx context.Context, args GetConte
 		regexEscapedRepoNames[i] = regexp.QuoteMeta(string(repo.Name))
 	}
 
-	textQuery := fmt.Sprintf(`repo:^%s$ %s content:%s`, query.UnionRegExps(regexEscapedRepoNames), textFileFilter, strconv.Quote(args.Query))
-	codeQuery := fmt.Sprintf(`repo:^%s$ -%s content:%s`, query.UnionRegExps(regexEscapedRepoNames), textFileFilter, strconv.Quote(args.Query))
+	textQuery := fmt.Sprintf(`repo:^%s$ type:file type:path %s %s`, query.UnionRegExps(regexEscapedRepoNames), textFileFilter, args.Query)
+	codeQuery := fmt.Sprintf(`repo:^%s$ type:file type:path -%s %s`, query.UnionRegExps(regexEscapedRepoNames), textFileFilter, args.Query)
 
 	doSearch := func(ctx context.Context, query string, limit int) ([]FileChunkContext, error) {
 		if limit == 0 {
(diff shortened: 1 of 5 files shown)


internal/codycontext/context.go
```diff
@@ -268,8 +268,8 @@ func (c *CodyContextClient) getKeywordContext(ctx context.Context, args GetConte
 		regexEscapedRepoNames[i] = regexp.QuoteMeta(string(repo.Name))
 	}
 
-	textQuery := fmt.Sprintf(`repo:^%s$ %s content:%s`, query.UnionRegExps(regexEscapedRepoNames), textFileFilter, strconv.Quote(args.Query))
-	codeQuery := fmt.Sprintf(`repo:^%s$ -%s content:%s`, query.UnionRegExps(regexEscapedRepoNames), textFileFilter, strconv.Quote(args.Query))
+	textQuery := fmt.Sprintf(`repo:^%s$ type:file type:path %s %s`, query.UnionRegExps(regexEscapedRepoNames), textFileFilter, strconv.Quote(args.Query))
```
jtibshirani, line 271: I believe we need to specify `type` directly here, since by default we could return repos too, which we are not prepared to handle.

internal/codycontext/context.go
```diff
@@ -268,8 +267,8 @@ func (c *CodyContextClient) getKeywordContext(ctx context.Context, args GetConte
 		regexEscapedRepoNames[i] = regexp.QuoteMeta(string(repo.Name))
 	}
 
-	textQuery := fmt.Sprintf(`repo:^%s$ %s content:%s`, query.UnionRegExps(regexEscapedRepoNames), textFileFilter, strconv.Quote(args.Query))
-	codeQuery := fmt.Sprintf(`repo:^%s$ -%s content:%s`, query.UnionRegExps(regexEscapedRepoNames), textFileFilter, strconv.Quote(args.Query))
+	textQuery := fmt.Sprintf(`repo:^%s$ type:file type:path %s %s`, query.UnionRegExps(regexEscapedRepoNames), textFileFilter, args.Query)
```
jtibshirani, line 270: We need to specify `type` here explicitly, since by default we return repo matches too, which we are not prepared to handle.

internal/codycontext/context.go
```diff
@@ -344,7 +343,15 @@ func (c *CodyContextClient) getKeywordContext(ctx context.Context, args GetConte
...
 	if len(fm.ChunkMatches) == 0 {
-		return nil
+		// If this is a filename-only match, we return the first 20 lines of the file.
+		return []FileChunkContext{{
+			RepoName:  fm.Repo.Name,
+			RepoID:    fm.Repo.ID,
+			CommitID:  fm.CommitID,
+			Path:      fm.Path,
+			StartLine: 0,
+			EndLine:   20,
```
stefanhengl, line 353: Why 20? Is this to catch imports?

jtibshirani: I just chose 20 to match the number of lines we return for content matches. It seemed simplest to keep it consistent, and I didn't see a big downside.

PR #61670, opened 2024-04-07
gitserver: Don't reclone repos every 45 days

eseliger:
This PR simplifies the janitor so that it doesn't have to run clones out of band.

As a side-effect, this will prevent us to go over the configured concurrency for janitorial reasons. There are still a few other reasons but those will get eliminated later.

We used to reclone repos every 45 days, regardless of state. This is from the old dotcom, gitserver-is-only-a-cache days. In todays logic, this doesn't make a ton of sense since on most instances, all repos are expected to always be in a servable state.

Why don't we need this?

Many of the repos we keep don't actually need GC at all, so recloning is actually much more expensive for those. Especially for customers with many repos (and dotcom), it is expected that the majority of repos will get little to no changes.

In addition, we want to start generating additional git data structures like commitgraphs and file change bitmaps. Generating those is generally not free and they will be lost during a reclone.

So instead we now only try to detect when a repo is truly broken and will then delete it, so that the uncloned repos prioritizer will reclone them afterwards.

## Test plan

Adjusted tests for the janitor, and will monitor dotcom for a few days to see what impact this has.
If repos get too slow or the gc operations take too long, we can either revert or take additional measures.

### DIFF
--- a/cmd/gitserver/internal/cleanup.go
+++ b/cmd/gitserver/internal/cleanup.go
@@ -50,7 +50,7 @@ type JanitorConfig struct {
 	DisableDeleteReposOnWrongShard bool
 }
 
-func NewJanitor(ctx context.Context, cfg JanitorConfig, db database.DB, fs gitserverfs.FS, rcf *wrexec.RecordingCommandFactory, cloneRepo cloneRepoFunc, logger log.Logger) goroutine.BackgroundRoutine {
+func NewJanitor(ctx context.Context, cfg JanitorConfig, db database.DB, fs gitserverfs.FS, rcf *wrexec.RecordingCommandFactory, logger log.Logger) goroutine.BackgroundRoutine {
 	return goroutine.NewPeriodicGoroutine(
 		actor.WithInternalActor(ctx),
 		goroutine.HandlerFunc(func(ctx context.Context) error {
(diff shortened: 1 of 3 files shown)


eseliger:
* **#61670** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/61670?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

cmd/gitserver/internal/cleanup.go
```diff
@@ -878,18 +836,35 @@ func gitIsNonBareBestEffort(rcf *wrexec.RecordingCommandFactory, repoName api.Re
...
+func gitGC(ctx context.Context, logger log.Logger, backend git.GitBackend, rcf *wrexec.RecordingCommandFactory, repoName api.RepoName, dir common.GitDir) error {
 	cmd := exec.Command("git", "-c", "gc.auto=1", "-c", "gc.autoDetach=false", "gc", "--auto")
 	dir.Set(cmd)
 	wrappedCmd := rcf.WrapWithRepoName(context.Background(), log.NoOp(), repoName, cmd)
 	err := wrappedCmd.Run()
 	if err != nil {
+		if gclog, readErr := os.ReadFile(dir.Path("gc.log")); readErr == nil && len(gclog) > 0 {
+			// gc failed most likely.
+			logger.Error("git gc failed", log.String("repo", string(dir)), log.Error(err), log.String("gc.log", string(gclog)))
+			existing, err := backend.Config().Get(ctx, gitConfigGCFailed)
```
mmanela, line 853: What does this backend config thing do?

mmanela (approved):
One non-blocking question

cmd/gitserver/internal/cleanup.go
eseliger: If `gitConfigGCFailed` is not present, we set it to the current time. That'll let us identify the first time this has occurred, so we can let it have some more attempts before we consider the repo borked.

mmanela: ah I see this is setting on the actual config file in the repo, interesting

PR #61672, opened 2024-04-08
gitserver: Don't trigger clones in hot requests

eseliger:
This removes the enqueueing of clones during git API requests for all but dotcom.
This didn't serve a good purpose, the repos are all expected to be cloned in a steady state, and uncloned repos are regularly prioritized for cloning.
Cloning repos out-of-band in here caused large spikes in clone jobs on dotcom occasionally.
Also, since the queue was in-memory, we lost those requests across restarts.

Since for sourcegraph.com, we don't manage all repos by the scheduler, we have to keep some path here, and we do that now by asking the repo updater scheduler to consider a clone (through repo update) for that repo, but on the priority and concurrency of the repo fetch scheduler instead of out of band now.

This will help make the scheduler fully own the decision making about fetches and clones.

## Test plan

E2E and integration test suites still work, adjusted the clone test and it's still passing.


cmd/gitserver/internal/server_grpc.go
```diff
@@ -1185,17 +1186,37 @@ func (gs *grpcServer) ResolveRevision(ctx context.Context, req *proto.ResolveRev
...
+			gs.logger.Debug("not cloning on demand as DisableAutoGitUpdates is set")
+		} else {
+			_, err := repoupdater.DefaultClient.EnqueueRepoUpdate(ctx, repo)
+			if err != nil {
+				return errors.Wrap(err, "failed to enqueue repo clone")
+			}
+		}
+	}
+
+	cloneProgress, cloneInProgress := gs.locker.Status(repo)
```
mmanela, line 1217: Question (non-blocking): I see cloneProgress is a string, what does it contain?

mmanela (approved):

cmd/gitserver/internal/server_grpc.go
eseliger: the last line of git output of the fetch command, usually that's something like

```
Receiving objects:  23% (1534/6460), 11.68 MiB | 23 KiB/s  
```

PR #61681, opened 2024-04-08
search: wrap structural search with repoPagerRob 

stefanhengl:
This is a refactor to use the repo pager for structural search. This will be useful to support
structural search in Search Jobs.

Test plan:
- updated unit test
- manual testing: I compared result counts for a cross-repo structural search query of this PR vs main

### DIFF
--- a/internal/search/job/jobutil/job.go
+++ b/internal/search/job/jobutil/job.go
@@ -408,12 +408,16 @@ func NewFlatJob(searchInputs *search.Inputs, f query.Flat) (job.Job, error) {
 				Features:        *searchInputs.Features,
 			}
 
-			addJob(&structural.SearchJob{
+			structuralSearchJob := &structural.SearchJob{
 				SearcherArgs:     searcherArgs,
 				UseIndex:         f.Index(),
 				ContainsRefGlobs: query.ContainsRefGlobs(f.ToBasic().ToParseTree()),
-				RepoOpts:         repoOptions,
 				BatchRetry:       searchInputs.Protocol == search.Batch,
+			}
+
+			addJob(&repoPagerJob{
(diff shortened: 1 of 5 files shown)


internal/search/job/jobutil/job.go
```diff
@@ -408,12 +408,16 @@ func NewFlatJob(searchInputs *search.Inputs, f query.Flat) (job.Job, error) {
...
-				ContainsRefGlobs: query.ContainsRefGlobs(f.ToBasic().ToParseTree()),
-				RepoOpts:         repoOptions,
-				BatchRetry:       searchInputs.Protocol == search.Batch,
+			structuralSearchJob := &structural.SearchJob{
+				SearcherArgs: searcherArgs,
+				UseIndex:     f.Index(),
+				BatchRetry:   searchInputs.Protocol == search.Batch,
+			}
+
+			addJob(&repoPagerJob{
```
mmanela, line 417: Question (non-blocking): What effect does this have? In other words, what behavior changes by nesting the search job in the repoPagerJob?

stefanhengl: Before, `structuralSearchJob` looped over repos and called `runStructuralSearch` for every page of repos it received from the iterator. Now it delegates the looping to repoPageJob. In other words, we move the loop a level higher in the hierarchy. We will certainly have more spans in our tracing because now we create one per repo page, while before we created one per search.

mmanela (approved):

internal/search/job/jobutil/job.go
mmanela: makes sense

keegancsmith (approved):

PR #61685, opened 2024-04-08
build-tracker: fix nil pointer in getting commit author

Strum355:
This was causing panics in MSP. Unclear why this would ever be nil, but apparently it can be?

## Test plan

curl'd locally with payload from notification service log


dev/build-tracker/build/build.go
```diff
@@ -120,8 +120,8 @@ func (b *Build) IsReleaseBuild() bool {
 	return false
 }
 
-func (b *Build) GetBuildAuthor() Author {
-	var author Author
+func (b *Build) GetBuildAuthor() buildkite.Author {
```
burmudar, line 123: You can delete the struct too?

PR #61697, opened 2024-04-08
sg msp: add interactive version of 'sg msp init' and 'sg msp init-env'

bobheadxi:
As titled - we can now prompt you through service setup. This is a small QOL improvement that makes providing the required parameters a bit easier, as the setup flags experience has been a point of feedback from several MSP adopters.

As a follow-up, we can probably remove the flag-based setup entirely, as it should generally be a human-operator-only setup. Then we can expand the setup to include e.g. resource setup (postgres), and maybe an initial generate step as well.

## Test plan

image


unknwon (approved):
Test plan LGTM 👍

Issue #61560, opened 2024-04-03
Refactor: create appliance service


jdpleiness:
Create new appliance service in sourcegraph mono repo `cmd/appliance` to run the reconciliation loop, control sourcegraph deployment resources, service the maintenance API, and maintenance UI.

Issue #61559, opened 2024-04-03
👀 Tracking: EAP appliance deployment method


jdpleiness:
### 🏗  Development

- Required for initial release. Refactoring work has been split into a separate tracking issue:
  - #61556 
  - https://github.com/sourcegraph/sourcegraph-operator/issues/70
  - https://github.com/sourcegraph/sourcegraph-operator/issues/79
  - https://github.com/sourcegraph/sourcegraph-operator/issues/31
  - https://github.com/sourcegraph/sourcegraph-operator/issues/76
  - https://github.com/sourcegraph/sourcegraph-operator/issues/60
  - https://github.com/sourcegraph/sourcegraph-operator/issues/81
  - https://github.com/sourcegraph/sourcegraph-operator/issues/64

- Upgrade Sourcegraph deployments:

- Product integration and features (can be added later):
  -  https://github.com/sourcegraph/sourcegraph-operator/issues/59
  - https://github.com/sourcegraph/sourcegraph-operator/issues/61
  - https://github.com/sourcegraph/sourcegraph-operator/issues/77
  - https://github.com/sourcegraph/sourcegraph-operator/issues/75
  - https://github.com/sourcegraph/sourcegraph-operator/issues/74

jdpleiness:
Moved to linear roadmap: see https://linear.app/sourcegraph/roadmap/o-sourcegraph-appliance-9d5308681bb7/all

Issue #61556, opened 2024-04-03
👀 Tracking: refactor appliance to not use CRDs


jdpleiness:
## Issue
During the appliance demo at Merge, it was communicated that some of our customers would either not be able to or not be comfortable using Custom Resource Definitions (CRDs) in their Kubernetes clusters.

Based on this feedback, the decision was made to refactor the current appliance architecture and remove the usage of CRDs. 

This issue serves as a sub-tracking issue for the work related to this decision and how it affects the overall appliance architecture.

### ✍  Design

- [x] Come up with rough draft of how the new appliance architecture will work.
- [ ] Document out reasoning behind the changes in design based on the feedback from key stakeholders.

### 🏗  Refactoring

Some code from the original project can be reused with only minor changes.

- Migrate k8s resource code that can possibly be reused from the [sourcegraph-operator](https://github.com/sourcegraph/sourcegraph-operator) to the Sourcegraph monorepo:
   - #61414 
   - #61430 
   - #61432 
   - #61435 
   - #61474 
   - #61477 
   - #61479 
   - #61496 
   - #61501 
   - #61505 
   - #61346 
   - #61384 
   - #61329 

- Migrate Sourcegraph Service definitions. These require some changes as references to the CRD to allow users to override fields with customer parameters is no longer possible.
  - https://github.com/sourcegraph/sourcegraph-operator/issues/82
  - https://github.com/sourcegraph/sourcegraph-operator/issues/39
  - https://github.com/sourcegraph/sourcegraph-operator/issues/83
  - https://github.com/sourcegraph/sourcegraph-operator/issues/38
  - https://github.com/sourcegraph/sourcegraph-operator/issues/50
  - https://github.com/sourcegraph/sourcegraph-operator/issues/49
  - https://github.com/sourcegraph/sourcegraph-operator/issues/45
  - https://github.com/sourcegraph/sourcegraph-operator/issues/42
  - https://github.com/sourcegraph/sourcegraph-operator/issues/41
  - https://github.com/sourcegraph/sourcegraph-operator/issues/40
  - https://github.com/sourcegraph/sourcegraph-operator/issues/48
  - https://github.com/sourcegraph/sourcegraph-operator/issues/43
  - https://github.com/sourcegraph/sourcegraph-operator/issues/46
  - https://github.com/sourcegraph/sourcegraph-operator/issues/47
  - https://github.com/sourcegraph/sourcegraph-operator/issues/44

[...]

PR #61706, opened 2024-04-08
cmd/appliance: Add appliance service boilerplate / stub the service

jdpleiness:
This PR introduces a new Sourcegraph service, or at least the boilerplate needed for one, called `appliance`.

The appliance service is part of an on-going effort to simplify our deployments by providing a platform for automations that can assist in deploying, upgrading, and scaling Sourcegraph.

The first aim of appliance is to provide advanced health checking in a Kubernetes environment. This will be then expanded on to managing deployments, upgrades, scaling, and advanced DR options in the future. 

Ultimately, appliance will be an *optional* service that our customers can opt-out of deploying.

---

Why a new service? If we use or [docs](https://sourcegraph.com/docs/dev/background-information/architecture/introducing_a_new_service) as a guide:

- This service will in it's nature need to be able to manage other Sourcegraph services. This alone excludes it from being added to an existing service.
- This is new logic, and is very heavily focused on the Sourcegraph deployment aspect which excludes it from being coupled to another service.
- This service has unique requirements that it needs access to the Kubernetes API in certain cases. While this can be scoped tightly, being able to disable this service while still running other Sourcegraph services is useful.

Also from our docs, "[Services have additional overhead for us and users that is easy to forget](https://sourcegraph.com/docs/dev/background-information/architecture/introducing_a_new_service#services-have-additional-overhead-for-us-and-users-that-is-easy-to-forget)":

- Absolutely! This service is being written by the Release team, so we're in a great position to add it to the proper deployments when the time comes 😏 

---

Currently:
* This is largely a skeleton for now
* This won't be included in releases until it's ready
* It builds
* It runs

More background information around the reasoning, design, and scope can be found below:
* [RFC 818](https://docs.google.com/document/d/1jZzvobpV4f9sHOdUyduNtyTMmJ8ThsZGjMRvJ9VvEq8/edit#heading=h.8nozctbmjyhv)
* [RFC 792](https://docs.google.com/document/d/1SGjaw16-k_Dvm1bWeN9qaM_Kvee1XqShdU5XXML0iwI/edit#)
* closes [#61560](https://github.com/sourcegraph/sourcegraph/issues/61560)
* [#61556](https://github.com/sourcegraph/sourcegraph/issues/61556)
* [#61559](https://github.com/sourcegraph/sourcegraph/issues/61559)

## Test plan

CI pass, tested locally


cmd/appliance/main.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"github.com/sourcegraph/sourcegraph/cmd/appliance/shared"
+	"github.com/sourcegraph/sourcegraph/internal/sanitycheck"
+	"github.com/sourcegraph/sourcegraph/internal/service/svcmain"
+)
+
+func main() {
+	sanitycheck.Pass()
+	svcmain.SingleServiceMain(shared.Service)
```
eseliger, line 11: Is appliance something that runs _as part of_ sourcegraph? My understanding is that it would be sort of a sidecar.

If so, I think we should use `SingleServiceMainWithoutConf` here instead. (see docstring for that)

internal/appliance/spec.go
eseliger, line 1: Can this live in cmd/appliance? I don't see it imported elsewhere.

cmd/appliance/shared/main.go
```diff
@@ -0,0 +1,56 @@
...
+		logger.Fatal("unable to start manager", log.Error(err))
+	}
+
+	// Mark health server as ready
+	ready()
+
+	logger.Info("starting manager")
+	if err := mgr.Start(shutdownOnSignal(ctx)); err != nil {
+		logger.Fatal("problem running manager", log.Error(err))
+	}
```
Chickensoupwithrice, line 33: Should we return the `err` here or is logging it enough?

Chickensoupwithrice (approved):
Looks like a great start! :)

DaedalusG (approved):
LGTM, although I'd love a README with basics about how to run etc

internal/appliance/spec.go
jdpleiness: It's not at the moment, but future code in `internal/appliance` will use it heavily. I think if anything it could have been left out of this PR and merged in the next, seeing it's not really doing anything atm but helping stub out the configuration.

michaellzc (approved):

jhchabran (approved):
SGTM at this stage, two little things: 

- When you want this to get shipped, you'll want to mention the image in https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@8ff72b9db0cf3533b849112e936002f5003ee75a/-/blob/dev/ci/images/images.go?L102
- When you'll want the images to get built, you'll need to add the `oci_*` rules :) 

Also, I think it'd be nice if we had a README in there explaining purpose/owners/where to get help

cmd/appliance/main.go
jdpleiness: Yeah you're spot on. I switched it out 🙏

PR #61707, opened 2024-04-08
V2t src/search

dadlerj:
Since this is a larger and more impactful one, I've tested it pretty thoroughly. I'm a bit confused about the dynamic filters (it seems like we have both new and old versions of these, some of which live in this directory, and some in /client/branded, making it hard to test everything) and search aggregations (which are also spread across multiple files). But overall I feel good.

Context:

* https://github.com/sourcegraph/sourcegraph/pull/60767
* https://github.com/sourcegraph/sourcegraph/pull/60768
* https://github.com/sourcegraph/sourcegraph/pull/60788
* https://github.com/sourcegraph/sourcegraph/pull/60789
* https://github.com/sourcegraph/sourcegraph/pull/60803
* https://github.com/sourcegraph/sourcegraph/pull/60812
* https://github.com/sourcegraph/sourcegraph/pull/60850
* https://github.com/sourcegraph/sourcegraph/pull/60851
* https://github.com/sourcegraph/sourcegraph/pull/60939
* https://github.com/sourcegraph/sourcegraph/pull/60971
* https://github.com/sourcegraph/sourcegraph/pull/61205
* https://github.com/sourcegraph/sourcegraph/pull/61457
* https://github.com/sourcegraph/sourcegraph/pull/61503
* https://github.com/sourcegraph/sourcegraph/pull/61504
* https://github.com/sourcegraph/sourcegraph/pull/61506
* https://github.com/sourcegraph/sourcegraph/pull/61507
* https://github.com/sourcegraph/sourcegraph/pull/61516

## Test plan

`sg start`
visit page

### DIFF
--- a/client/web/src/search/input/V2SearchInput.tsx
+++ b/client/web/src/search/input/V2SearchInput.tsx
@@ -23,6 +23,7 @@ import type { SearchContextProps, SubmitSearchParameters } from '@sourcegraph/sh
 import { FILTERS, FilterType } from '@sourcegraph/shared/src/search/query/filters'
 import { resolveFilterMemoized } from '@sourcegraph/shared/src/search/query/utils'
 import { useTemporarySetting } from '@sourcegraph/shared/src/settings/temporary'
+import { TelemetryV2Props } from '@sourcegraph/shared/src/telemetry'
 import type { TelemetryProps } from '@sourcegraph/shared/src/telemetry/telemetryService'
 
 import { createSuggestionsSource, type SuggestionsSourceConfig } from './suggestions'
@@ -94,6 +95,7 @@ export interface V2SearchInputProps
     extends Omit<CodeMirrorQueryInputWrapperProps, 'suggestionSource' | 'extensions' | 'placeholder'>,
         Pick<SearchContextProps, 'selectedSearchContextSpec'>,
         TelemetryProps,
+        TelemetryV2Props,
         Pick<SuggestionsSourceConfig, 'authenticatedUser' | 'isSourcegraphDotCom'> {
     submitSearch(parameters: Partial<SubmitSearchParameters>): void
 }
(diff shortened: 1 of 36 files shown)


client/web/src/search/input/V2SearchInput.tsx
```diff
@@ -171,6 +174,20 @@ export const V2SearchInput: FC<PropsWithChildren<V2SearchInputProps>> = ({
                     type: option.kind,
                     source,
                 })
+                switch (action.type) {
```
vovakulikov, line 177: Can this be simplified to generic name, like this?

```ts
 telemetryRecorder.recordEvent(`search.input.${action.type}`, 'select')
```

vovakulikov (approved):

client/web/src/search/input/V2SearchInput.tsx
dadlerj: Yes, we definitely could do that :). You may have noticed that I am, at times, choosing the more verbose way of writing out event names. This is intentional (though maybe also a violation of our repo norms, and I can avoid it in the future). The goal is to make it easy to detect where events are being logged in the future—I can run a simple search for an event feature and find the call site, instead of having to guess if/where the string was interpolated.

Let me know if you would recommend simplifying in this case!

dadlerj: I'm going to merge this now just to avoid merge conflict risk and if you think I should change this please just DM me on slack!

vovakulikov: Makes sense to me, had no problem with merging this one, thanks for the reply.

vovakulikov: One thing to add, (not a blocker and no immediate action required)
At the moment we operate with strings for event names. The problem with this 
that the only way to find event name usage is through finding strings across the repository. I think it would be easier for us if we had even names enum, something like this

```
enum EventName {
   SearchInputSelect = 'search.input.select',
   SearchInputAdd = 'search.input.add',
   ...
   ...
}

// Later in the consumer code
telemetryRecorder.recordEvent(EventName. SearchInputSelect, 'select')
```

After this, we could easy 
- find usage of events 
- rename event names if it's needed 

As I said, I think it's okay for now to work with strings; after we're done with migration, I would think about it.

PR #61708, opened 2024-04-08
vscsyncer: introduce syncer wrapper which calculates latencies for all operations

### LINKED ISSUE #61692: Instrument VCSSyncers for call count and latency

ggilmore:
Closes #61692 

## Test plan

Created the following Grafana screenshot using `sg start monitoring`:

screencapture-sourcegraph-test-3443-debug-grafana-d-gitserver-git-server-2024-04-11-13_46_51


ggilmore:
* **#61708** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/61708?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

monitoring/definitions/git_server.go
```diff
@@ -43,6 +120,17 @@ func GitServer() *monitoring.Dashboard {
 				Multi: true,
 			},
 			grpcMethodVariable,
+			{
+				Label: "VCS Syncer Kind",
+				Name:  vcsSyncerVariableName,
+				OptionsLabelValues: monitoring.ContainerVariableOptionsLabelValues{
+					Query:         "vcssyncer_fetch_duration_seconds_bucket",
+					LabelName:     "type",
+					ExampleOption: "jvm",
```
ggilmore, line 129: Fun fact! If this value is set to `git`, this completely breaks out templating in non-intuitve ways (starts complaining about illegal characters like `$` in metrics like disk_size_remaining where that was never a problem before). I will file an issue about this.

cmd/gitserver/internal/vcssyncer/git.go
```diff
@@ -33,11 +33,12 @@ type gitRepoSyncer struct {
 func NewGitRepoSyncer(
 	logger log.Logger,
 	r *wrexec.RecordingCommandFactory,
-	getRemoteURLSource func(ctx context.Context, name api.RepoName) (RemoteURLSource, error)) *gitRepoSyncer {
-	return &gitRepoSyncer{
+	getRemoteURLSource func(ctx context.Context, name api.RepoName) (RemoteURLSource, error)) VCSSyncer {
```
eseliger, line 36: Idea: keep the return type here `*gitRepoSyncer` and let the caller of `NewGitRepoSyncer` wrap it in an observability layer if they are interested in that.
Observability seems like an add-on to me that not everyone might be interested in (hypothetically), and wrapping it at this layer couples the two things very strictly, while they could live separately.

For example, in our test suite for `GitRepoSyncer`, we don't care about this behavior.

cmd/gitserver/internal/vcssyncer/instrumented_syncer.go
```diff
@@ -0,0 +1,107 @@
...
+
+	return &instrumentedSyncer{
+		base:               syncer,
+		formattedTypeLabel: typ,
+	}
+}
+
+// instrumentedSyncer wraps a VCSSyncer and records metrics for each method call.
+type instrumentedSyncer struct {
+	base               VCSSyncer
```
eseliger, line 57: idea: if you embed VCSSyncer directly, you don't have to do things like 

```
func (i *instrumentedSyncer) Type() string {
	return i.base.Type()
}
```

.. but you would lose the certainty that you correctly overwrote all methods. So no strong opinion here, just something I noticed while reading the diff.

cmd/gitserver/internal/vcssyncer/instrumented_syncer.go
```diff
@@ -0,0 +1,107 @@
...
+
+	err := i.base.IsCloneable(ctx, repoName)
+	if err != nil {
+		succeeded = false
+	}
+
+	metricIsCloneableDuration.WithLabelValues(i.formattedTypeLabel, strconv.FormatBool(succeeded)).Observe(time.Since(start).Seconds())
+
+	return err
+}
```
eseliger, line 77: I wonder if this is a little less err prone to accidentally returning the wrong error? 

```suggestion
func (i *instrumentedSyncer) IsCloneable(ctx context.Context, repoName api.RepoName) (err error) {
	start := time.Now()
	defer func() {
	  	metricIsCloneableDuration.WithLabelValues(i.formattedTypeLabel, strconv.FormatBool(err == nil)).Observe(time.Since(start).Seconds())
	}()
	
	return i.base.IsCloneable(ctx, repoName)
}
```

monitoring/definitions/git_server.go
eseliger: Very fun indeed 🙃

eseliger (approved):
Super nit: This reads to me as if this wasn't actually clones, do we need the quotes? 

Screenshot 2024-04-11 at 19 57 10@2x

Question: What is "Value" here? 

Screenshot 2024-04-11 at 19 58 26@2x

Good stuff! added a few non-blocking comments

cmd/gitserver/internal/vcssyncer/instrumented_syncer.go
ggilmore: Good thought! Fixed.

cmd/gitserver/internal/vcssyncer/git.go
ggilmore: Hmm, so you prefer exposing the instrumented syncer as a new type? I liked the idea of not changing the interface at all since I can't think of a case where we'd _want_ to opt-out of instrumentation outside of the vcssyncer package

eseliger: I would probably wrap the GitSyncer in `NewVCSSyncer` and not in the constructor here, but no blocker. That'll let us create a regular one for testing etc

ggilmore: Ah, I was typing this idea up as you posted this: 

> I would probably wrap the GitSyncer in NewVCSSyncer and not in the constructor her

Issue #61313, opened 2024-03-20
Markdown Rendering Issue with Blockquotes and Numbered Lists


gabtorre:
- **Sourcegraph version:** 5.3

When rendering Markdown, blockquotes inserted within numbered lists break the continuity of the list. Instead of continuing the sequence, the list restarts or breaks following a blockquote.

#### Steps to reproduce:

Create a Markdown document.
Insert a numbered list with at least two items.
Include a blockquote (using > syntax) after the first item and before the second.
Render the Markdown to observe the issue.

#### Expected behavior:

![Screenshot 2024-03-20 at 4 03 58 PM](https://github.com/sourcegraph/sourcegraph/assets/69164745/26833db5-6a7f-4cfd-9d9b-8b746f61e155)

#### Actual behavior:
![Screenshot 2024-03-20 at 4 04 23 PM](https://github.com/sourcegraph/sourcegraph/assets/69164745/b77e4d96-b0bc-4700-bff3-e0757eee58f0)

If you would like immediate help on this, please email support@sourcegraph.com (you can still create the issue, but there are no [SLAs](https://sourcegraph.com/support/) on issues like there are for support requests).

raznar:
Affected by [customer 578](https://github.com/sourcegraph/accounts/issues/578)

PR #61709, opened 2024-04-08
Allow start attribute on ordered lists in Markdown

gabtorre:
Closes: https://github.com/sourcegraph/sourcegraph/issues/61313

Allows the `start` attribute on `ol` elements. This will enable specifying a starting value for ordered lists in rendered markdown. e.g. `<ol start="2">`
## Test plan

Tested locally

Before:
314762633-b77e4d96-b0bc-4700-bff3-e0757eee58f0

After:
Screenshot 2024-04-08 at 4 16 29 PM

### DIFF
--- a/internal/markdown/markdown.go
+++ b/internal/markdown/markdown.go
@@ -41,6 +41,7 @@ func Render(content string) (string, error) {
 		policy.AllowAttrs("srcset", "src", "type", "media", "width", "height", "sizes").OnElements("source")
 		policy.AllowAttrs("playsinline", "muted", "autoplay", "loop", "controls", "width", "height", "poster", "src").OnElements("video")
 		policy.AllowAttrs("src", "kind", "srclang", "default", "label").OnElements("track")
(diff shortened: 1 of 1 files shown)


fkling:
Just double checking here: Have you tried indenting the blockquote by several spaces (e.g. 2, 4, etc)? (I don't know if that works but I know that some elements can be made part of the list item this way).

gabtorre:
@fkling three spaces do make the element part of the first list item. This change would display the file the same way it does on GitHub.

Here's an example of this: 
https://github.com/sourcegraph/sourcegraph/blob/7432727f701ddaa2de4f5716bee753e81418f48a/client/vscode/README.md#adding-it-to-your-workspaces-recommended-extensions

https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@7432727f701ddaa2de4f5716bee753e81418f48a/-/blob/client/vscode/README.md#adding-it-to-your-workspaces-recommended-extensions

vovakulikov (approved):

PR #61712, opened 2024-04-09
scip-ctags: better error propagation

jtibshirani:

### DIFF
--- a/docker-images/syntax-highlighter/crates/syntax-analysis/src/ctags.rs
+++ b/docker-images/syntax-highlighter/crates/syntax-analysis/src/ctags.rs
@@ -197,16 +197,12 @@ pub fn generate_tags<W: std::io::Write>(
     let (root_scope, _) = match get_globals(parser, file_data) {
         Ok(vals) => vals,
         Err(err) => {
-            // TODO: Not sure I want to keep this or not
-            #[cfg(debug_assertions)]
-            if true {
-                panic!("Could not parse file: {}", err);
-            }
-
-            let _ = err;
-
-            return None;
-        }
+            Reply::Error {
+                message: err.to_string(),
+                fatal: false,
+            }.write(buf_writer);
+            return None
+        },
     };
 
     let mut scope_deduplicator = HashMap::new();
@@ -242,15 +238,7 @@ pub fn ctags_runner<R: Read, W: Write>(
             break;
         }
 
-        let request = serde_json::from_str::<Request>(&line);
-        let request = match request {
-            Ok(request) => request,
-            Err(_) => {
-                eprintln!("Could not parse request: {}", line);
-                continue;
-            }
-        };
-
+        let request = serde_json::from_str::<Request>(&line)?;
         match request {
             Request::GenerateTags { filename, size } => {
                 let mut file_data = vec![0; size];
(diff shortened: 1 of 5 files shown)


docker-images/syntax-highlighter/crates/syntax-analysis/src/ctags.rs
```diff
@@ -197,16 +197,12 @@ pub fn generate_tags<W: std::io::Write>(
     let (root_scope, _) = match get_globals(parser, file_data) {
         Ok(vals) => vals,
         Err(err) => {
-            // TODO: Not sure I want to keep this or not
-            #[cfg(debug_assertions)]
```
jtibshirani, line 201: I removed this since I think we should always surface errors instead of silently squashing them.

jtibshirani:
Highlighting that I chose to make "unsupported language" a non-fatal error. In terms of user impact, this means we won't fail indexing an entire Zoekt repo if we can't parse a single file due to its language.

Background: this happens sometimes because we can misroute files. Zoekt performs precise language detection, so it classifies [iamjs.pl](https://github.com/github-linguist/linguist/blob/4ac734c15a96f9e16fd12330d0cb8de82274f700/test/fixtures/Data/Modelines/iamjs.pl#L2) as Javascript and passes it to scip-ctags. However, scip-ctags thinks this is Perl, which it can't handle. (We should fix this too, but it's been a problem for a while, both with universal-ctags and scip-ctags!)

docker-images/syntax-highlighter/crates/syntax-analysis/src/ctags.rs
```diff
@@ -242,31 +238,29 @@ pub fn ctags_runner<R: Read, W: Write>(
...
-        let request = serde_json::from_str::<Request>(&line);
-        let request = match request {
-            Ok(request) => request,
-            Err(_) => {
-                eprintln!("Could not parse request: {}", line);
-                continue;
-            }
-        };
-
+        let request = serde_json::from_str::<Request>(&line)?;
```
varungandhi-src, line 241: I think receiving a malformed request should be a non-fatal error, but the `?` will cause a return from a function and hence be treated as a fatal error.

varungandhi-src (changes requested):
Thanks for the PR. ~Looks directionally OK, but it seems a bit odd that we would start treating malformed requests as fatal errors -- is that for matching the behavior of universal-ctags?~

docker-images/syntax-highlighter/crates/syntax-analysis/src/ctags.rs
varungandhi-src: I saw you wrote this in the PR description:

> Fatal: user command is invalid or incorrectly formatted, I/O errors. This indicates a big issue in how we're using ctags and we should fail hard.

I suppose it's also OK to do it this way. We can change this if it is a problem in practice.

varungandhi-src (approved):

docker-images/syntax-highlighter/crates/syntax-analysis/src/ctags.rs
jtibshirani: Right, and this matches the behavior of universal-ctags. Here's an example:

```
~/code/sourcegraph % universal-ctags --_interactive                                    
{"_type": "program", "name": "Universal Ctags", "version": "6.0.0"}
julie
{"_type": "error", "message": "invalid json", "fatal": true}
```

PR #61715, opened 2024-04-09
Adding an extra field for the tokenizer computed tokens for Anthropic API 

arafatkatze:
[Slack Thread
](https://sourcegraph.slack.com/archives/C05AGQYD528/p1712071741350299)

I am doing this so as to compute the average offset in the computed tokens with the usage report from messages api for how many tokens were used. 

The Bigquery query we can use to interpret this 

```
SELECT 
    name,
    JSON_EXTRACT(metadata, '$.model') AS model_name,
    JSON_EXTRACT(metadata, '$.completion_token_count') AS token_count,
    JSON_EXTRACT(metadata, '$.completion_character_count') AS character_count,
    -- Add more JSON_EXTRACT lines here for other fields you're interested in
FROM 
    telligentsourcegraph.cody_gateway.events
LIMIT 50;

```

Interpretting and computing this data for a few days can tell use the average offset which can help us understand if we can use the old tokenizer for abuse detection or do I need to implement a new one. 

image

## Test plan
Checking the Bigquery outputs

### DIFF
--- a/cmd/cody-gateway/internal/httpapi/completions/anthropic.go
+++ b/cmd/cody-gateway/internal/httpapi/completions/anthropic.go
@@ -176,13 +176,18 @@ func (a *AnthropicHandlerMethods) transformRequest(r *http.Request) {
 
 func (a *AnthropicHandlerMethods) parseResponseAndUsage(logger log.Logger, reqBody anthropicRequest, r io.Reader) (promptUsage, completionUsage usageStats) {
 	var err error
+
+	// Setting a default -1 value so that in case of errors the tokenizer computed tokens don't impact the data
+	completionUsage.tokenizerTokens = -1
+	promptUsage.tokenizerTokens = -1
+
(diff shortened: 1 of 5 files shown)


cmd/cody-gateway/internal/httpapi/completions/anthropic.go
```diff
@@ -198,6 +204,7 @@ func (a *AnthropicHandlerMethods) parseResponseAndUsage(logger log.Logger, reqBo
 			logger.Error("failed to count tokens in Anthropic response", log.Error(err))
 		} else {
 			completionUsage.tokens = len(tokens)
+			completionUsage.tokenizerTokens = completionUsage.tokens
```
rafax, line 207: It seems like we never tokenize the completion, so I think we can leave this unset?

rafax (approved):

PR #61719, opened 2024-04-09
symbols: canceling ParseRequest does not spam logs

keegancsmith:
We are seeing lots of log spam related to cancelation of parse requests. We haven't fully tracked this down yet, but the behaviour of spamming logs in this case hides the real issue.

This adjusts the implementation in two places:
- We introduce a parseCanceled metric so we can still track this. This is used instead of parseFailed in case of cancelation. In this case we also do not emit the "Closing failed parser" even though we are. This is due to log spam.
- We introduce an ErrorFilter just for handleParseRequest that will elide Canceled from the logs. This is called per file. Note: we keep the error logging for canceled in the high level Parse handler which runs across all files.

Test Plan: go test. Unfortunetly that is as good as it gets for now, as I investigate the root cause further I will test this code path (and only merge once I have manually triggered and tested this code path).

### DIFF
--- a/cmd/symbols/parser/BUILD.bazel
+++ b/cmd/symbols/parser/BUILD.bazel
@@ -1,3 +1,4 @@
+load("//dev:go_defs.bzl", "go_test")
 load("@io_bazel_rules_go//go:def.bzl", "go_library")
 
 go_library(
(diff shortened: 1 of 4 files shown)


jtibshirani (approved):
Makes sense to me. Hopefully this will get us closer to root causing the cancellations!

PR #61720, opened 2024-04-09
code-search: fix vertical alignment in search results list

BolajiOlajide:
[Context](https://sourcegraph.slack.com/archives/C05MHAP318B/p1712667167930699)

When viewing the search result page for files, the icon and path aren't properly aligned. This fixes that.

| Before  |
|---|
| CleanShot 2024-04-09 at 14 08 26@2x  |
| After  |
| CleanShot 2024-04-09 at 14 08 16@2x  |

## Test plan

Manual testing

### DIFF
--- a/client/web-sveltekit/src/routes/search/RepoRev.svelte
+++ b/client/web-sveltekit/src/routes/search/RepoRev.svelte
@@ -17,22 +17,31 @@
     }
 </script>
 
-<CodeHostIcon repository={repoName} />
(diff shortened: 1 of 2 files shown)


vovakulikov:
We had a convo with @camdencheek about this. It looks like that align-items is correct here  because it aligns repo name text and file path text. For code host icon I think we should set align-self: center

camdencheek (approved):

PR #61721, opened 2024-04-09
Add modal to remind users to connect external accounts

### LINKED ISSUE #60616: Add UX for asking users to connect their external identities
We want to encourage users to connect required external accounts after login. 

Currently, a user can log in to sourcegraph, lands on /search and starts searching, but they won't get results. That is because without their external identity connected, they will not have any permissions synced and attached to their user account.

First step will be to design what the UX should be like here.

pjlast:
Closes #60616 

Adds a modal that reminds users to connect external accounts for permissions syncing.

The modal does not show accounts that the user has already connected, and once dismissed, the modal won't be shown again unless a new auth provider is added.

The modal also only shows auth providers that would have an effect on permissions syncing. If no authorization is configured for any code host connections, then no modal will be shown.

See the attached Loom for a demo

https://www.loom.com/share/cb5f219bb9f3436fa1bcc1f70d603e67

Corresponding docs entry: https://github.com/sourcegraph/docs/pull/282

## Test plan

Lots of manual testing and added a unit test for the resolver. Probably needs a few more tests


cmd/frontend/graphqlbackend/authz.go
```diff
@@ -172,3 +173,9 @@ type PermissionsSyncingStatsResolver interface {
 	UsersWithStalePermissions(ctx context.Context) (int32, error)
 	ReposWithStalePermissions(ctx context.Context) (int32, error)
 }
+
+type AuthzProviderResolver interface {
+	URN() string
+	ServiceType() string
+	ServiceID() string
+}
```
eseliger, line 181: the external accounts page uses this query here: https://sourcegraph.sourcegraph.com/-/editor?remote_url=github.com%2Fsourcegraph%2Fsourcegraph&branch=es%2Fhoney-ms&file=client%2Fweb%2Fsrc%2Fuser%2Fsettings%2Fbackend.tsx&editor=VSCode&version=2.2.16&start_row=49&start_col=0&end_row=68&end_col=0

Why do we need a new resolver here, or what can that one not provide?

client/web/src/LegacyLayout.tsx
```diff
@@ -272,6 +285,14 @@ export const LegacyLayout: FC<LegacyLayoutProps> = props => {
                     userHistory={userHistory}
                 />
             )}
+            {props.authenticatedUser && externalAccountsModalVisible && (
```
eseliger, line 288: IMO this new modal is a net-positive improvement, but to be absolutely sure, should we add something to disable the feature?

client/web/src/external-account-modal/ExternalAccountsModal.tsx
```diff
@@ -0,0 +1,192 @@
...
+            isOpen={isModalOpen}
+            onDismiss={onDismiss}
+            className={styles.modal}
+            position="center"
+        >
+            <div className={styles.title}>
+                <BrandLogo variant="symbol" isLightTheme={props.isLightTheme} disableSymbolSpin={true} />
+                <div>
+                    <H2>Sourcegraph setup: permissions & security</H2>
+                    <Text>Connect external identity providers to access private repositories.</Text>
```
eseliger, line 171: maybe? 

```suggestion
                    <Text>Connect external identities to your account to access private repositories.</Text>
```

eseliger (commented):
left a few questions inline, will continue review tmr

cmd/frontend/graphqlbackend/authz.graphql
```diff
@@ -120,6 +120,24 @@ enum CancelPermissionsSyncJobResultMessage {
...
+    """
+    urn: String!
+    """
+    The service type of the authz provider.
+    """
+    serviceType: String!
+    """
+    The ID of the service the authz provider is associated with, e.g. https://example.gitlab.com
+    """
+    serviceID: String!
```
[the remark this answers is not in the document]
pjlast: Nope, there's nothing here that isn't available to the user in the OAuth flow anyways

client/web/src/LegacyLayout.tsx
```diff
@@ -134,6 +137,16 @@ export const LegacyLayout: FC<LegacyLayoutProps> = props => {
...
 
+    const [seenAuthzProviders, setSeenAuthzProviders] = useTemporarySetting('user.seenAuthProviders')
+
+    useEffect(() => {
+        const externalAccountsModalVisible = shouldShowExternalAccountsModal(
+            window.context.authProviders,
+            seenAuthzProviders
+        )
+        setExternalAccountsModalVisible(externalAccountsModalVisible)
+    }, [seenAuthzProviders])
```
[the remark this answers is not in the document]
pjlast: So I don't like this either, but because the Apollo lib uses React hooks, and React hooks can't be called conditionally, it leads to extra GraphQL queries being made when we could determine earlier that we're not going to render the modal.

Perhaps we can pull out the inner component of the modal into a sub-component? And then at least encapsulate all the logic into one file.

client/web/src/external-account-modal/ExternalAccountsModal.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+        if (
+            !seenAuthProviders.find(
+                seenProvider =>
+                    seenProvider.serviceType === activeProvider.serviceType &&
+                    seenProvider.clientID === activeProvider.clientID
+            )
+        ) {
+            return true
+        }
```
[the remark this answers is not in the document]
pjlast: What would we do in the case of the username ones? 🤔

cmd/frontend/graphqlbackend/authz.go
pjlast: The external accounts page lists all auth providers. We're interested in only Authz providers, i.e. auth providers for which there is also a corresponding code host connection that has `authorization` enabled.

client/web/src/LegacyLayout.tsx
pjlast: I'm not in favour of adding yet another "but have you enabled this setting" thing. I can be convinced otherwise, but my gut feeling is that I don't think it is necessary.

eseliger: I totally hear you :/ I'm just scared we might be missing some edge-case in the initial implementation and that some customer will run into issues with it where we unnecessarily bother every user. For example, what happens with explicit perms? 

It could be some temporary stopgap kinda thing?

cmd/frontend/graphqlbackend/authz.go
eseliger: Aha!
I wonder if we could perhaps add a field to the resolver that's linked that indicates "required for authz" or so 🤔 That way, we could also add an info icon on the account security page or so, and it's mostly a backend concern what actually "is required for authz"

client/web/src/external-account-modal/ExternalAccountsModal.tsx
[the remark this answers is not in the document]
eseliger: That's above my understanding of how this works today 😆 do you need to connect anything there at all?

client/web/src/LegacyLayout.tsx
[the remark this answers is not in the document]
eseliger: Yep I think that's the react-y way to go :)

cmd/frontend/graphqlbackend/authz.go
mmanela: @eseliger  @pjlast  Should we still though default this to enabled and just leave the FF as an option to turn this setting off?

client/web/src/LegacyLayout.tsx
mmanela: @eseliger @pjlast Should we still though default this to enabled and just leave the FF as an option to turn this setting off?

client/web/src/LegacyLayout.tsx
```diff
@@ -272,6 +273,13 @@ export const LegacyLayout: FC<LegacyLayoutProps> = props => {
                     userHistory={userHistory}
                 />
             )}
+            {props.authenticatedUser && (
```
mmanela, line 276: Do we need to do anything with the new svelte app? Or even let code search team know?

client/web/src/external-account-modal/ExternalAccountsModal.tsx
eseliger: Thoughts?

eseliger (approved):
Approving to unblock for release cut today, I’ll play with it a little more later

PR #61722, opened 2024-04-09
search jobs: add support for structural search

stefanhengl:
This adds support for structural search queries to Search Jobs. The changes are isolated to the Search Jobs framework.

Limitations:

- No AND/OR queries, because structural search doesn't support `Basic` queries.

This means

supported:
```
if ... patternType:structural
```

not supported
```
if ... OR switch ... patternType:structural
```

Notes:
- [Structural search relies on `bufio.Scanner` to read stdout](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@2ab15741c105937beba202785002ff31d0e8f5d0/-/blob/cmd/searcher/internal/search/search_structural.go?L411). Potentially long lines in the output of comby might cause a search job to fail. This limitation is not isolated to Search Jobs, but Search Jobs amplifies the issue because we fail the job if one (sub) search fails. In my tests it was very easy to hit that limit. If we merge this PR we should make the buffer size configurable as a follow-up.

Test plan:
- Updated unit test

### DIFF
--- a/internal/search/job/jobutil/exhaustive_job_test.go
+++ b/internal/search/job/jobutil/exhaustive_job_test.go
@@ -30,14 +30,16 @@ func TestNewExhaustive(t *testing.T) {
 	}
 
 	cases := []struct {
-		Name      string
-		Query     string
-		WantPager autogold.Value
-		WantJob   autogold.Value
+		Name       string
+		Query      string
+		WantPager  autogold.Value
+		WantJob    autogold.Value
+		SearchType query.SearchType
 	}{
 		{
-			Name:  "case sensitive lang match",
-			Query: `type:file index:no lang:cpp content case:yes`,
+			Name:       "case sensitive lang match",
+			Query:      `type:file index:no lang:cpp content case:yes`,
+			SearchType: query.SearchTypeLiteral,
 			WantPager: autogold.Expect(`
 (REPOPAGER
   (containsRefGlobs . false)
@@ -60,8 +62,9 @@ func TestNewExhaustive(t *testing.T) {
 `),
 		},
 		{
-			Name:  "glob",
-			Query: `type:file index:no repo:foo rev:*refs/heads/dev* content`,
+			Name:       "glob",
+			Query:      `type:file index:no repo:foo rev:*refs/heads/dev* content`,
+			SearchType: query.SearchTypeLiteral,
 			WantPager: autogold.Expect(`
 (REPOPAGER
   (containsRefGlobs . true)
@@ -85,8 +88,9 @@ func TestNewExhaustive(t *testing.T) {
 `),
 		},
 		{
-			Name:  "only pattern",
-			Query: "type:file index:no content",
+			Name:       "only pattern",
+			Query:      "type:file index:no content",
+			SearchType: query.SearchTypeLiteral,
 			WantPager: autogold.Expect(`
 (REPOPAGER
   (containsRefGlobs . false)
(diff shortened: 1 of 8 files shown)


internal/search/job/jobutil/exhaustive_job_test.go
```diff
@@ -30,14 +30,16 @@ func TestNewExhaustive(t *testing.T) {
...
+		WantPager  autogold.Value
+		WantJob    autogold.Value
+		SearchType query.SearchType
 	}{
 		{
-			Name:  "case sensitive lang match",
-			Query: `type:file index:no lang:cpp content case:yes`,
+			Name:       "case sensitive lang match",
+			Query:      `type:file index:no lang:cpp content case:yes`,
+			SearchType: query.SearchTypeLiteral,
```
stefanhengl, line 42: We should use SearchTypeKeyword instead or at least vary types a bit. I will send a separate PR for this to keep the changes here minimal.

internal/search/job/jobutil/exhaustive_job_test.go
```diff
@@ -157,32 +163,34 @@ func TestNewExhaustive(t *testing.T) {
...
+			SearchType: query.SearchTypeRegex,
 			WantPager: autogold.Expect(`
 (REPOPAGER
   (containsRefGlobs . false)
   (repoOpts.useIndex . no)
   (PARTIALREPOS
     (SEARCHERTEXTSEARCH
       (useFullDeadline . true)
-      (patternInfo . TextPatternInfo{"foo.*bar",nopath,filematchlimit:1000000})
+      (patternInfo . TextPatternInfo{/foo.*bar/,nopath,filematchlimit:1000000})
```
jtibshirani, line 176: Can you explain why it changed from "foo.*bar" -> /foo.*bar/ ? Usually you don't see the /.../ syntax when the patterntype is regexp.

jtibshirani:
> In my tests it was very easy to hit that limit.

Will increasing the buffer size a bit solve the whole? Or are there other roadblocks to making this feature usable? Motivation for asking: we want to avoid investing too much in structural search, so it's good to know how much work is remaining on this.

jtibshirani (commented):
Will increasing the buffer size a bit solve the whole problem? Or are there other roadblocks to making this feature usable? Motivation for asking: we want to avoid investing too much in structural search, so it's good to know how much work is remaining on this.

stefanhengl:
> Will increasing the buffer size a bit solve the whole problem? 

In my testing I increased the buffer size by 10x and the jobs that failed before completed without error. I don't know if there is buffer size that solves the whole problem. My guess is it is a moving target.

> Or are there other roadblocks to making this feature usable? Motivation for asking: we want to avoid investing too much in structural search, so it's good to know how much work is remaining on this.

While testing, I found some oddities here and there, for example

1. Discrepancy between indexed and unindexed result counts for structural search. I am still investigating but I will probably open a ticket for that
2. Sometimes lang: filters are not respected

None of these are related to Search Jobs, I could reproduce the issues with pre structural search.

If we are ok with the limitation of no AND/OR, then there is no further work. However, Search Jobs "unlocks" structural search in the sense that you can now run it across your codebase without timeouts. This means we might get more bug reports about structural search.

internal/search/job/jobutil/exhaustive_job_test.go
stefanhengl: There was a mistake in the test. It was using SearchTypeLiteral although the query is regex. The "/" are just how regex pattern nodes are printed, see [here](https://github.com/sourcegraph/sourcegraph/blob/b93f340eecdb62c310aaa2520841beb11540efb9/cmd/searcher/protocol/query.go#L121)

stefanhengl:
Based on the limitations mentioned above and our current policy not to invest further into structural search, we have decided to not add support for structural search.

PR #61723, opened 2024-04-09
svelte: Redesign file header

### LINKED ISSUE #61630: Add "download file" action

### LINKED ISSUE #61631: Add "view on codehost" action

### LINKED ISSUE #61652: Svelte: Implement file header
The file header is now pretty stable in the designs.

image

The workflow actions toolbar adjusts according to available space or mode, it can also wrap. Cody Explain won't be available as an action item until later.

image

Figma:
- https://www.figma.com/file/EPgl0qVbqGSumvf52brhda/Code-Search?type=design&node-id=1947%3A37160&mode=design&t=rs6WkWU36T6PfNDq-1
- https://www.figma.com/file/EPgl0qVbqGSumvf52brhda/Code-Search?type=design&node-id=2591%3A75731&mode=design&t=rs6WkWU36T6PfNDq-1
- https://www.figma.com/file/EPgl0qVbqGSumvf52brhda/Code-Search?type=design&node-id=2730%3A226742&mode=design&t=rs6WkWU36T6PfNDq-1

fkling:
Closes #61652
Closes #61630
Closes #61631

This PR implements the redesigned file header. Specifically it:

- Updates the font and spacing of the file path
- Adds the file icon to the file name
- Updates the styles of the file action buttons/links
- Adds an overflow menu for less important actions
- Adds an accessible "view mode" switcher
- Adds file meta information (lines and size)

NOTE: Since blame view is not supported yet I've removed that option for now, but it's easy to add back.

NOTE 2: As pointed out in #61652, I intentionally excluded some parts in this first iteration.

| Formatted with open menu | Code |
|--------|--------|
|  |  | 
| Note: The screenshot shows the menu entry as "Raw download" but I've changed it to "View raw" ||

Unlike the React version, "View raw" opens the file in a new tab instead of directly downloading it. That's something we can iterate on in the future.

I've also adjusted removed the icons from the headers on the repo home page and the directory page, to be more aligned with the file page.

## Test plan

- Tested file actions (open on codehost, permalink, view raw, line wrapping)
- Tested view mode changes and switcher UI for "formatted" files (e.g. markdown) files, and code files.
- Visual inspection of repo home and directory pages.
- New playwright tests for file page


client/web-sveltekit/src/lib/repo/Actions.svelte
```diff
@@ -0,0 +1,59 @@
...
+
+        .more {
+            align-self: center;
+        }
+
+        @container fileheader (width < 50cqw) {
+            :global([data-action-label]) {
+                display: none;
+            }
+        }
```
fkling, line 57: This is a very simple way to hide the button/link labels when the container has a certain size. It's not ideal since it doesn't take the length of the file path into account, but again, it's a simple CSS based solution.

client/web-sveltekit/src/lib/repo/FileHeader.svelte
```diff
@@ -13,39 +13,40 @@
...
-            {/each}
-        </span>
+                {#if path}
+                    <a href={path}>{name}</a>
+                {:else}
+                    <slot name="icon" />
+                    {name}
+                {/if}
+            </span>
+        {/each}
```
fkling, line 33: I should add another comment here: I restructured this so that a path like

```
/very/long/path/file.ts
```

is wrapped like this

```
/very/long/path
/file.ts
```

I.e. the slash before the segment will always be wrapped *with* the segment.

client/web-sveltekit/src/lib/wildcard/menu/MenuButton.svelte
```diff
@@ -0,0 +1,14 @@
+<script lang="ts">
+    import type { HTMLButtonAttributes } from 'svelte/elements'
+    import { getContext } from './DropdownMenu.svelte'
+
+    type $$Props = HTMLButtonAttributes
+
+    const { item } = getContext()
```
camdencheek, line 7: I think this is unrelated to the PR, but why do we communicate this via context?

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte
```diff
@@ -52,29 +61,86 @@
         // Update selected position as soon as blob is loaded
         selectedPosition = SourcegraphURL.from($page.url).lineRange
     }
+
+    function changeViewMode({ detail: viewMode }: { detail: ViewMode }) {
+        console.log('changeViewMode', viewMode)
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte:66]: intentional?

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte
```diff
@@ -52,29 +61,86 @@
...
+        switch (viewMode) {
+            case ViewMode.Code: {
+                const url = SourcegraphURL.from($page.url)
+                if (isFormatted) {
+                    url.setSearchParameter('view', 'raw')
+                } else {
+                    url.deleteSearchParameter('view')
+                }
+                goto(url.toString(), { replaceState: true, keepFocus: true })
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte:75]: One thing I noticed: clicking `Code` triggers a navigation, which means we jump into full-page mode. That was a little surprising to me, especially since the button moved.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte
```diff
@@ -52,29 +61,86 @@
...
+            <div class="actions">
+                <Actions>
+                    {#if blob}
+                        <OpenInCodeHostAction data={blob} />
+                    {/if}
+                    <Permalink {commitID} />
+                    <svelte:fragment slot="menu">
+                        <MenuLink href="{repoURL}/-/raw/{filePath}" target="_blank">
+                            <Icon svgPath={mdiFileEyeOutline} inline /> View raw
+                        </MenuLink>
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte:110]: Unrelated to this PR: I just noticed we already have a raw file endpoint. RE: our previous conversation about caching, can we just use that for fetching file contents?

camdencheek (approved):

client/web-sveltekit/src/lib/dom.ts
```diff
@@ -16,6 +16,70 @@ import * as uuid from 'uuid'
...
+ * @param selector The selector to match the sibling against.
+ * @param wrap Whether to wrap around to the last sibling if no match is found.
+ * @returns The previous sibling element that matches the selector, or null if no match is found.
+ */
+export function previousSibling<T extends Element>(element: Element, selector: string, wrap = false): T | null {
+    let sibling = element.previousElementSibling
+    while (true) {
+        while (sibling) {
+            if (sibling.matches(selector)) {
+                return sibling as T
```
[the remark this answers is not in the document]
fkling: Yes, but I think it's "normal" with DOM APIs. E.g. if you use something like `querySelector` to find a specific element, or access `event.target` in an event handler, the results are often cast to a specific element type.
Likewise here we assume that the caller knows that the selector matches only a specific type of element.
It's not great, but I guess that's what we have :(

client/web-sveltekit/src/lib/repo/Permalink.svelte
```diff
@@ -12,7 +16,15 @@
 </script>
 
 {#if href}
-    <Tooltip tooltip="Permalink (with full Git commit SHA)">
-        <a {href}><Icon svgPath={mdiLink} inline /></a>
+    <Tooltip tooltip="Permalink (with full git commit SHA)">
+        <a {href}><Icon svgPath={mdiLink} inline /> <span data-action-label>Permalink</span></a>
```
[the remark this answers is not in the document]
fkling: Yes. Happy to change this in another PR but I'd also like to put this up for discussion. I really like links ;)

client/web-sveltekit/src/lib/dom.ts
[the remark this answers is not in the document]
camdencheek: Makes sense 👍 

Wondering then if we should move the casting to the caller though to force them to be explicit that they know the type rather than assuming they know that passing a type parameter means they're asserting that their selector can only return elements of that type.

client/web-sveltekit/src/lib/repo/Actions.svelte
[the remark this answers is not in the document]
fkling: You are right, I think I need to change what the container is supposed to be. I tried various things so maybe things got mixed up a bit.

client/web-sveltekit/src/lib/wildcard/menu/DropdownMenu.svelte
```diff
@@ -16,15 +16,15 @@
...
-    import type { Writable } from 'svelte/store'
+    import { writable, type Writable } from 'svelte/store'
 
-    interface $$Props extends HTMLButtonAttributes {
-        open: Writable<boolean>
+    type $$Props = {
+        open?: Writable<boolean>
         triggerButtonClass: string
-    }
+    } & HTMLButtonAttributes
```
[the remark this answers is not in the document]
fkling: Moving from `interface` to `type`? In my editor setup I get warnings that `$$Props` cannot be overwritten or something when it's an interface. Honestly I don't know why. This doesn't trigger the error but also works as expected 🤷🏻‍♂️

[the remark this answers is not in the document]
camdencheek: Oh cool. I also get that error. Will start using `type` 😄

client/web-sveltekit/src/lib/wildcard/menu/MenuButton.svelte
fkling: `DropdownMenu.svelte` uses melt UI to create some builders for menu items, submenus, etc and they are specific to the menu that's created. Context is the simplest way to pass those builders to any descendants.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte
Comment (fkling) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte:75]: Yep, I noticed that too. Might have to make an exception when changing the view.

client/web-sveltekit/src/lib/repo/Actions.svelte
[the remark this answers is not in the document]
fkling: But to clarify, it will never work perfectly, i.e. there is not way to determine with CSS alone whether or not there is enough space for the labels. But at least we can make it relative to the size of the file header.
I had tried a JS based approach but it was kind of broken, so I thought before I spent more time on this specifically I rather get some approximation working and we'll revisit this another time.

client/web-sveltekit/src/lib/wildcard/menu/MenuButton.svelte
camdencheek: Ah, got it. Thanks!

client/web-sveltekit/src/lib/repo/Actions.svelte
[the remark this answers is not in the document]
fkling: We can do that with container queries. What I meant is that ideally the labels would be hidden as soon as the path starts to wrap. That's not something we can do in CSS alone.

client/web-sveltekit/src/lib/wildcard/menu/MenuButton.svelte
fkling: Slot props would be a bit round about. The callsite doesn't need to worry about the inner workings/relation between `DropdownMenu` and `MenuButton`/`MenuLink`.

client/web-sveltekit/src/lib/repo/Actions.svelte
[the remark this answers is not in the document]
camdencheek: Oooh, gotcha. Thanks for clarifying 👍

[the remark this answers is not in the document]
fkling: As a quick update: I've misunderstood some aspects of container queries and it doesn't work like I expected it to. The main issue is that by creating a "containment context", the container looses the ability to compute its own size by its contents, which is useful in a flex box context. But I'll experiment a bit with different settings.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte
Comment (fkling) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte:75]: I don't know why but that seems to only happen when I click on the view mode buttons after a fresh page load. When I navigate to/from another file it doesn't happen (and it shouldn't according to the code). So that "to the top scrolling" behavior seems to coming from somewhere else. We should just remove full page scrolling.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte
Comment (fkling) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte:110]: I don't see why not.

fkling:
OK, I completely changed how we determine the space available for actions. It didn't seem there was an easy way to make this work with container queries, even just approximating it, due to how creating containment contexts impact the size calculation of the container.

Instead I want back to a JS based approach. Due to how different styles can impact the size and layout of an element it seems difficult to find a universally applicable approach. So with the API added here it's the callers responsibility to define/provide styles that allow the utility to measure the element's size. I hope the comments and the example explain the approach well enough.

Additionally I removed the `Actions.svelte` class and moved the actions logic into the file header instead. The reason being that there didn't seem to be a way to prevent leaking spacing settings (e.g. gaps, padding) between the two elements (either the actions component needs to know about the padding the file header applies or the file header needs to know which gap the actions component uses). To me that was an indication that this should really be a single component.

taiyab:
Beautiful work @fkling!

PR #61731, opened 2024-04-09
monitoring: add golang monitoring for zoekt

keegancsmith:
Noticed this omission when I was wondering if we had goroutine leaks. Our other services define this.

I added a simple way to indicate the container name in title since this is the first service we added which needs this.

Test Plan: go test. Copy paste generated query into grafana explore on dotcom.

### DIFF
--- a/doc/admin/observability/alerts.md
+++ b/doc/admin/observability/alerts.md
@@ -6570,6 +6570,126 @@ Generated query for warning alert: `max((max by (name) (container_oom_events_tot
 
 <br />
 
+## zoekt: go_goroutines
+
+<p class="subtitle">maximum active goroutines</p>
+
+**Descriptions**
+
+- <span class="badge badge-warning">warning</span> zoekt: 10000+ maximum active goroutines for 10m0s
+
+**Next steps**
+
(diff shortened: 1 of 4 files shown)


doc/admin/observability/alerts.md
```diff
@@ -6570,6 +6570,126 @@ Generated query for warning alert: `max((max by (name) (container_oom_events_tot
 
 <br />
 
+## zoekt: go_goroutines
+
+<p class="subtitle">maximum active goroutines</p>
+
+**Descriptions**
+
+- <span class="badge badge-warning">warning</span> zoekt: 10000+ maximum active goroutines for 10m0s
```
stefanhengl, line 6579: 10k is arbitrary but high enough to detect a leak? Just curious, what is our current level of goroutines for Zoekt?

stefanhengl (approved):

doc/admin/observability/alerts.md
keegancsmith: I reckon high enough. given its a static value we don't want to be close to our current peaks. This is dotcom
image

PR #61732, opened 2024-04-09
symbols: avoid spamming handleParseRequest on context cancelation

keegancsmith:
Understanding how Parse works (and its callers) with respect to goroutines and channels is a sure path to the asylum. This is a short term fix around how it behaves when the context is canceled.

From what I can tell we can have a large number of parse requests go through handleParseRequest even once the context has been canceled, leading to large amounts of logging of errors which all amount to context canceled. This is a short term fix which drains the channel if we know we are just gonna fail our call to handleParseRequest.

Test Plan: go test. Didn't manual test, but the code change is minor enough that I trust it (famous last words).

### DIFF
--- a/cmd/symbols/parser/parser.go
+++ b/cmd/symbols/parser/parser.go
@@ -115,6 +115,12 @@ func (p *parser) Parse(ctx context.Context, args search.SymbolsParameters, paths
 					break
 				}
 
+				// Drain channel if our context has been canceled. Otherwise
+				// we just logspam on handleParseRequest.
+				if ctx.Err() != nil {
+					continue
+				}
+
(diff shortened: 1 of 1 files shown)


jtibshirani (commented):
Do you see this as complementing https://github.com/sourcegraph/sourcegraph/pull/61719? Maybe it's not strictly needed but it's a good practice to check context cancellation?

keegancsmith:
Yup is a complement to it. To be honest if I could understand how all the goroutines and channels fit together properly I would do some sort of break out of the loop. Instead this essentially just avoids calling that function if we know it won't run due to context cancellation.

jtibshirani (approved):

PR #61736, opened 2024-04-09
web: Fix blame recency color "direction"

fkling:
#61661 changed how "blame recency" is calculated in the client. In a last minute change I flipped the meaning of the boolean parameter but forgot to update the callsite. That's why the colors are currently the opposite (most recent = gray, least recent = purpel).

## Test plan

Trivial change.

### DIFF
--- a/client/web/src/repo/blob/codemirror/blame-decorations.tsx
+++ b/client/web/src/repo/blob/codemirror/blame-decorations.tsx
@@ -270,7 +270,7 @@ class RecencyMarker extends GutterMarker {
             if (this.hunk.startLine === this.line) {
                 dom.classList.add('border-top')
             }
-            dom.style.backgroundColor = getBlameRecencyColor(new Date(this.hunk.author.date), !this.darkTheme)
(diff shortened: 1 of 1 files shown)


camdencheek (approved):

camdencheek:
@fkling this should be live now, but I'm still seeing the backwards colors

fkling:
@camdencheek I see the colors in the right direction 🤔 

Screenshot 2024-04-11 at 20 44 18

camdencheek:
@fkling I confirmed that they were backwards for me, but on a hunch, I switched from light mode to dark mode and back. Now it's fixed.

fkling:
Strange. No idea why it wouldn't pick it up correctly otherwise.

eseliger:
That screenshot looks correct to me 🤔 or am I misunderstanding that dark means most recent?

camdencheek:
Dark does mean most recent. In [this comment](https://github.com/sourcegraph/sourcegraph/pull/61736#issuecomment-2050291924), you can see that Jan 2 is light, even though it's the most recent line. 

Turning light mode off and back on seemed to fix it though, which is the screenshot in [this comment](https://github.com/sourcegraph/sourcegraph/pull/61736#issuecomment-2050308328)

PR #61737, opened 2024-04-09
rbac: check for undefined permissions in PermissionsList

BolajiOlajide:
#60795 introduced the concept of permissions that only exists on `Dotcom`, however one thing we didn't take into consideration is how the frontend displays the role.

Turns out we get `allPermissions` from the generated `graphql-operations.ts` file which is statically generated at build time and doesn't know about permissions that only exist on `dotcom`. 

## Test plan

Manual testing?

Added a storybook also.


camdencheek (approved):

camdencheek:
Tested locally that the `Create Role` button now works 👍 

CleanShot 2024-04-09 at 12 40 59@2x

PR #61742, opened 2024-04-09
database: Format metadata as multiline string

eseliger:
We have a suspicion that this might help fix query insights aggregations. If it doesn't, we can revert this PR safely.

## Test plan

All DB related tests are still passing.


camdencheek (approved):
LGTM! Let's give it a try

PR #61746, opened 2024-04-09
Svelte: Implements the RepoPopover component 

jasonhawkharris:
This PR implements the `RepoPopover` component. 

**With header**
Screenshot 2024-04-10 at 1 57 15 PM
Screenshot 2024-04-10 at 1 57 48 PM

**Without header**
Screenshot 2024-04-10 at 1 57 28 PM
Screenshot 2024-04-10 at 1 58 00 PM

There are a few differences in terms of the design and implementation:

1. License information is not something we currently process on the backend, so I have omitted it here. 
2. Similarly, number of forks is also not something we currently process on the backend, so I have omitted it here as well. We can explore adding both of these in a subsequent PR, but I think this will suffice for an initial implementation.
3. Since I've left these out, I have taken the liberty of putting the star count on the right of the `div`, and the primary language on the left. I would appreciate feedback on whether folks think this is the best solution for an MVP, or if a different layout might work better.

Some other notes:
- This only implements the `RepoPopover` component - it is not yet integrated into the flow of the web app. I will do that integration in a subsequent PR. For now, I'm primarily looking for feedback on the design, implementation, and whether either can be improved.
- In general, I am wary of very large components. In this initial PR, I am submitting the entire component in one file. I'd welcome thoughts on potentially splitting this into smaller components within the `RepoPopover` directory.

## Test plan
Manual testing and visual verification


client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -0,0 +1,327 @@
...
+        <div class="title-and-commit">
+            <div class="title">
+                <small>Last Commit</small>
+            </div>
+            <div class="commit-and-number">
+                <div class="commit">
+                    <!--
+                        Since the popover has a fixed width, we use the lodash
+                        truncate() function to truncate the subject to 30 characters.
+                        We do this instead of using text-overflow: ellipsis because
```
camdencheek, line 124: What about [`text-overflow: clip` or `text-overflow: ""`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow)? The problem with truncating is that characters aren't always the same width. As an extreme example, "﷽" is a single character.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -0,0 +1,327 @@
...
+            case ExternalServiceKind.GITOLITE:
+                return mdiGit
+            case ExternalServiceKind.AZUREDEVOPS:
+                return mdiMicrosoftAzureDevops
+            case ExternalServiceKind.AWSCODECOMMIT:
+                return mdiAws
+            default:
+                return mdiSourceMerge
+        }
+    }
```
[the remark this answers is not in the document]
jasonhawkharris: @camdencheek It does exist, but not on the svelte web app yet. I figured I'd go ahead and add it here in anticipation of deprecating the other frontend at some point.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -0,0 +1,312 @@
...
+
+    .repo-stats {
+        color: var(--text-muted);
+        display: flex;
+        flex-flow: row nowrap;
+        justify-content: space-between;
+        align-items: center;
+        margin-bottom: 0.5rem;
+        margin-top: 0.5rem;
+        padding: 0rem 0.75rem;
```
camdencheek, line 297: In our other sections, we use a consistent padding. Why not do the same here?

```suggestion
        padding: 0.75rem;
```

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -0,0 +1,312 @@
...
+        }
+    }
+
+    .divider {
+        border-bottom: 1px solid var(--border-color);
+        width: 100%;
+    }
+
+    .description-and-tags {
+        // border: 1px dotted white;
```
camdencheek, line 204: There are a couple of stray commented lines

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -0,0 +1,312 @@
...
+<style lang="scss">
+    .root {
+        border: 1px solid var(--border-color);
+        border-radius: var(--popover-border-radius);
+        width: 400px;
+    }
+
+    .header {
+        display: flex;
+        flex-flow: row-nowrap;
```
camdencheek, line 159: Firefox says `row-nowrap` is not a valid CSS property

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -0,0 +1,312 @@
...
+            .title {
+                color: var(--text-muted);
+            }
+
+            .commit-and-number {
+                display: flex;
+                flex-flow: row nowrap;
+                align-items: center;
+                justify-content: flex-end;
+                width: 200px;
```
camdencheek, line 254: I know the designs have this set to 200px, but we've got like another 100px of space so it feels wasteful to not show more of the commit message. Thoughts on whether we should increase it?

camdencheek (approved):
General comment: I think the styling could be simplified a bit by leaning on padding/flexbox/gap rather than adding margins to each individual component. Certainly not required, but could be a useful exercise.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
jasonhawkharris: Yeah this is a bad Cody completion. I'll fix this one up.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
jasonhawkharris: I went back and forth on this but I think I agree. I'll add more space.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -0,0 +1,312 @@
...
+    <div class="divider" />
+    <div class="repo-stats">
+        <div class="stats">
+            <div class="stat"><small>{repo.language}</small></div>
+        </div>
+        <div class="stat">
+            <Icon svgPath={mdiStarOutline} size={16} style="margin-right: 0.15rem;" />
+            <small>{formatNumber(repo.stars)}</small>
+        </div>
+    </div>
```
fkling, line 147: Overall this component seems to be doing a lot possibly unnecessary element nesting which makes it quite "verbose".

This here is one example. I think this could be written as

```suggestion
    <div class="footer">
        <small>{repo.language}</small>
        <div class="star">
            <Icon svgPath={mdiStarOutline} size={16} style="margin-right: 0.15rem;" />
            <small>{formatNumber(repo.stars)}</small>
        </div>
    </div>
```

(and this can be even more reduced if you use the `RepoStar` component).

In general it's good to use as few DOM/HTML elements as possible.

I suggest to go through the template and remove unnecessary wrappers, especially around single elements.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -0,0 +1,312 @@
...
+</div>
+
+<style lang="scss">
+    .root {
+        border: 1px solid var(--border-color);
+        border-radius: var(--popover-border-radius);
+        width: 400px;
```
fkling, line 154: Maybe the `width` should be controlled by the caller of this component.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -0,0 +1,312 @@
...
+            margin-top: 0.5rem;
+        }
+
+        .tag {
+            text-align: center;
+            background-color: var(--subtle-bg);
+            border-radius: 1rem;
+            color: var(--primary);
+            font-family: var(--monospace-font-family);
+            margin-right: 0.5rem;
```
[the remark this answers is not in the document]
fkling: ^ Yes, you can use `gap: 0.5rem` instead of `gap: 0.5rem 0rem` and achieve the same effect.

In fact `gap` could be use in a lot of other places here instead of margin.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -0,0 +1,312 @@
...
+                <div class="separator">{CENTER_DOT}</div>
+
+                {#if commitDate}
+                    <div class="commit-date">
+                        <small>{formatDistanceToNow(commitDate, { addSuffix: false })}</small>
+                    </div>
+                {/if}
+            </div>
+        </div>
+    </div>
```
fkling, line 137: Looking at the design, I think this should be structured differently. I see `Last commit` and the actual commit info as two cells in a row. So you have this general structure:

```html
<div class="last-comit">
  <div class="heading">Last commit</div>
  <div class="commit">...</div>
</div>
```

With flex box `.commit` can then use as much space as possible and it's two lines would be perfectly aligned.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -0,0 +1,312 @@
...
+        .stats {
+            display: flex;
+            flex-flow: row nowrap;
+            align-content: center;
+            font-size: 1rem;
+            padding: 0rem;
+
+            .stat {
+                align-self: center;
+                margin-right: 1rem;
```
fkling, line 308: Why is this margin needed?

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -0,0 +1,312 @@
...
+        color: var(--text-muted);
+        display: flex;
+        flex-flow: row nowrap;
+        justify-content: space-between;
+        align-items: center;
+        margin-bottom: 0.5rem;
+        margin-top: 0.5rem;
+        padding: 0rem 0.75rem;
+
+        .stats {
```
fkling, line 299: `.stats` is currently only used with a single child element:

```html
<div class="stats">
  <div class="stat"><small>{repo.language}</small></div>
</div>
```

See my component above on reducing HTML nesting, but that aside, this styling seems unnecessary for a single element.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -0,0 +1,312 @@
...
+                {/if}
+            </div>
+        </div>
+    </div>
+    <div class="divider" />
+    <div class="repo-stats">
+        <div class="stats">
+            <div class="stat"><small>{repo.language}</small></div>
+        </div>
+        <div class="stat">
```
fkling, line 143: There are no styles associated with this specific class.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -0,0 +1,312 @@
...
+    import { capitalize } from 'lodash'
+
+    import Avatar from '$lib/Avatar.svelte'
+    import { ExternalServiceKind } from '$lib/graphql-types'
+    import Icon from '$lib/Icon.svelte'
+
+    import { RepoPopoverFields } from './RepoPopover.gql'
+
+    export let repo: RepoPopoverFields
+    export let withHeader: Boolean = false
```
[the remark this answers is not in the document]
jasonhawkharris: Whoops. Cody got me again. thanks!

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
jasonhawkharris: Curious why you say that. I was thinking it should be uniform, since it's a popover, and the data is fairly consistent in terms of how much size it takes up.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
jasonhawkharris: It was needed when there were more stats below, but not anymore. I'll remove it. I probably should have used gap anyways.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
jasonhawkharris: Good idea!

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -0,0 +1,312 @@
...
+            </div>
+        </div>
+        <div class="divider" />
+    {/if}
+    <div class="description-and-tags">
+        <div class="description">{repo.description}</div>
+        <div class="tags">
+            {#if repo.tags.nodes.length > 0}
+                {#each repo.tags.nodes as tag}
+                    <div class="tag"><small>{tag.name}</small></div>
```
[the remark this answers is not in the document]
jasonhawkharris: Will address this in the similar comment above.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
jasonhawkharris: I'm going to leave this as is for the time being and revisit when I do subsequent PR work.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -0,0 +1,312 @@
...
+<div class="root">
+    {#if withHeader}
+        <div class="header">
+            <div class="icon-name-access">
+                <!-- @TODO: We need to use our customer's logo here, not the code host's -->
+                <!--Icon svgPath={mdiGitlab} /-->
+                <h4 class="repo-name">{repo.name}</h4>
+                <div class="access">
+                    <small>{repo.isPrivate ? 'Private' : 'Public'}</small>
+                </div>
```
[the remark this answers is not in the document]
jasonhawkharris: I'll make the issue.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.gql
```diff
@@ -0,0 +1,31 @@
+fragment RepoPopoverFields on Repository {
```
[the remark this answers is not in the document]
jasonhawkharris: I'll make an issue for this so we can track it.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
jasonhawkharris: done!

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.gql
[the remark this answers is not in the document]
jasonhawkharris: https://github.com/sourcegraph/sourcegraph/issues/61792

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -0,0 +1,258 @@
...
+        </div>
+        <div class="divider" />
+    {/if}
+
+    <div class="description-and-tags">
+        <div class="description">{repo.description}</div>
+        <div class="tags">
+            {#if repo.tags.nodes.length > 0}
+                {#each repo.tags.nodes as tag}
+                    <div class="tag"><small>{tag.name}</small></div>
```
fkling, line 68: Can possibly be flattened.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -0,0 +1,258 @@
...
+            {/if}
+        </div>
+    </div>
+
+    <div class="divider" />
+
+    <div class="last-commit">
+        <div class="heading">
+            <small>Last Commit</small>
+        </div>
```
fkling, line 79: Can possibly be flattened.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -0,0 +1,258 @@
...
+                        >
+                    </div>
+                {/if}
+            </div>
+            <div class="author">
+                {#if avatar}
+                    <Avatar {avatar} --avatar-size="1.0rem" />
+                {/if}
+                <small>{author}</small>
+                <small>{CENTER_DOT}</small>
```
fkling, line 104: This could be `<small>{author} {CENTER_DOT}</small>`.

fkling (commented):
Leaving a couple of comments for a follow up PR.

Looking at this again, seeing how often `<small>` is used, I think it might be better to just adjust the font size in CSS. But not sure.

PR #61752, opened 2024-04-10
streaming: only call flush if writes succeeded

keegancsmith:
This is a speculative change to reduce the chances of misuse of the streaming http event writer (SSE). We are seeing panics in attribution search since it is trying to write events once the request has finished.

From the stack traces it seems the http package safely handles calls to Write without panicing but does not do the same for Flush. This change also seems more correct.

Note: this does not address the root cause of the panics, but should make it far less likely to occur.

Test Plan: go test

Part of #60439


stefanhengl (approved):

cbart (approved):
Thanks Keegan!

PR #61761, opened 2024-04-10
msp: GCP Monitoring Dashboard

jac:
Collects all of the Alert Policies MSP creates into a single GCP Monitoring Dashboard to provide service owners a single pane of glass overview.
image
Logically groups alerts into collapsible sections
image

## Test plan

unit tests
tested by applying to msp-testbed


unknwon (approved):
Side question: is Terraform smart enough to know that when we add a new weight in the future, it would not destroy all off current and re-create from scratch?

dev/managedservicesplatform/stacks/monitoring/dashboard.go
```diff
@@ -0,0 +1,127 @@
...
+	return nil
+}
+
+func generateDashboard(serviceID, envID string, alertGroups map[string][]monitoringalertpolicy.MonitoringAlertPolicy) dashboard {
+	dashboard := dashboard{
+		DisplayName: fmt.Sprintf("MSP Alerts - %s-%s", serviceID, envID),
+		MosaicLayout: mosaicLayout{
+			Columns: dashboardColumns,
+		},
+	}
```
bobheadxi, line 75: nit - I think we can just initialize this at the very end on return, as `dashboard` is never referenced again until the end

dev/managedservicesplatform/stacks/monitoring/dashboard.go
```diff
@@ -0,0 +1,127 @@
...
+	Name string `json:"name"`
+}
+
+type collapsibleGroup struct {
+	Collapsed bool `json:"collapsed"`
+}
+
+const dashboardColumns = 48 // GCP default
+const widgetWidth = 24
+const widgetHeight = 16
```
bobheadxi, line 48: nit - move  this down closer to `generateDashboard` where it is used

dev/managedservicesplatform/stacks/monitoring/dashboard.go
```diff
@@ -0,0 +1,127 @@
...
+	"fmt"
+
+	"github.com/hashicorp/terraform-cdk-go/cdktf"
+	"github.com/sourcegraph/managed-services-platform-cdktf/gen/google/monitoringalertpolicy"
+	"github.com/sourcegraph/managed-services-platform-cdktf/gen/google/monitoringdashboard"
+	"github.com/sourcegraph/sourcegraph/dev/managedservicesplatform/internal/resourceid"
+	"github.com/sourcegraph/sourcegraph/lib/pointers"
+)
+
+type dashboard struct {
```
bobheadxi, line 14: Can you add a docstring link with the schema of this structure?

bobheadxi (approved):
Screenshots look _amazing_ , thank you!

bobheadxi:
> Side question: is Terraform smart enough to know that when we add a new weight in the future, it would not destroy all off current and re-create from scratch?

Do you mean, if we add a new panel/row? I think it shouldn't be a destructive change, but we can see

on that note - @jac , is there a way we can indicate to users that they should not edit this generated dashboard? e.g. indicate in the title or a description somewhere?

jac:
> > Side question: is Terraform smart enough to know that when we add a new weight in the future, it would not destroy all off current and re-create from scratch?
> 
> Do you mean, if we add a new panel/row? I think it shouldn't be a destructive change, but we can see
> 
> on that note - @jac , is there a way we can indicate to users that they should not edit this generated dashboard? e.g. indicate in the title or a description somewhere?

GCP just renders the parsed json definition of the dashboard. Modifying the json will cause GCP to update it in place.

I can add a banner to the top mentioning that it is an autogenerated dashboard and to please not edit it

jac:
here is what the banner looks like
image

I also sorted the sections as the order was indeterminate before

PR #21670, opened 2021-06-02
doc: Document oobmigration startup check

efritz:
Documenting changes in https://github.com/sourcegraph/sourcegraph/pull/20967 separately to ease review.

PR #61770, opened 2024-04-11
Fix OOB migrations progress updates 

### LINKED ISSUE #61350: Post v5.0.0 OOB migrations defined in earlier versions not marked complete
- **Sourcegraph version:** Instances initialized at or after v5.0.0
- **Platform information:** observed in docker-compose and AMI deployments

#### Steps to reproduce:

1. Initialize an instance in v5.0.0 or after
2. View the admin `updates` or `migrations` page 

#### Expected behavior:
These pages should only show warnings about OOB migrations not complete and or blocking. Instances initialized after `v5.0.0` should show previous OOB migrations as completed. For instance an OOB migration defined and required to be completed between `v4.1.0` and `v4.5.1` should be marked as completed on init in v5.0.0

#### Actual behavior:
Instances initialized at or after v5.0.0 always show a set of previously defined migrations as incomplete
Screenshot 2024-03-22 at 10 55 02 AM
Screenshot 2024-03-22 at 10 55 20 AM
^^ screenshots from a fresh instance initialized on v5.0.0

DaedalusG:
closes #61350

Solution to fix bug #61350 introduced in https://github.com/sourcegraph/sourcegraph/pull/48114/files

Previously instance would display many migrations as needing to be applied erroneously. It used to be that `progress` on a migration would be set to `1` in v4.5.1 and as such instances initialized or existent in that version wouldn't display migrations as needed. I've now fixed the check, instead of setting progress we just won' display oob migrations hat are deprecated before the `first_version` value in the `versions` table of `pgsql`

If I set the version to v4.5.0
```
sourcegraph=# UPDATE versions
SET first_version = '4.5.0'
WHERE service = 'frontend';
UPDATE 1
```
Screenshot 2024-04-25 at 10 53 11 PM

Then v3.3.0
Screenshot 2024-04-25 at 10 54 27 PM

I've also fixed the logic for the warnings displayed in the siteAdmin page such that they will no longer display warnings for migrations with progress 0 which were deprecated prior to the current running version. It shouldnt be possible to have a state where Sourcegraph starts with such migrations incomplete.

Planning a follow up PR to add some info I learned about these systems to our dev docs later.

Please review the boolean logic as its an error prone area

## Test plan

 Tested this by running it locally and changing the `first_version` value in the `versions table
 ```
 sourcegraph=# UPDATE versions
SET first_version = '3.3.0'
WHERE service = 'frontend';
UPDATE 1
```

### CODE REVIEW

File: internal/oobmigration/runner.go [hunk 1]
```diff
@@ -202,6 +202,16 @@ func (r *Runner) Start(currentVersion Version) {
 			return false
 		}
 
+		// If a currentVersion is at or after the deprecated version, the migration is considered
+		// deprecated and has been completed.
+		if migration.Deprecated != nil {
+			if CompareVersions(currentVersion, *migration.Deprecated) == VersionOrderAfter {
+				// current version on or after deprecation version
+				r.store.UpdateProgress(r.ctx, migration.ID, 1)
```
Chickensoupwithrice, line 210: I'm fairly certain this `r` has no reference in this function, and you'll get a nil pointer error in worker when trying to run this.

DaedalusG: Do we not get the runner from outer context?
```
func (r *Runner) Start(currentVersion Version) {
```
I was switching the progress between 0 and 1 during testing. I did get a nil pointer before introducing `if migration.Deprecated != nil {`

Chickensoupwithrice: I don't think so :/ In my testing I was getting nil pointer references to that `r`. 

I think the closure doesn't include that reference, but I'm not entirely sure why. I mention this in the footnote on my comment in https://github.com/sourcegraph/sourcegraph/issues/61350

Then again, if it's running on yours maybe I was doing something else that was causing problems, I didn't run this PR.

Chickensoupwithrice: I'm full of shit, I ran the PR and it worked flawlessly. 
Marking resolved.

Chickensoupwithrice (approved):
Ran locally, and it works! 

Worth noting that setting progress to 1 overrides invalid setting on the frontend, so they appear green rather than invalid, which is totally acceptable imo :)

client/web/src/site-admin/SiteAdminMigrationsPage.tsx
```diff
@@ -417,11 +417,15 @@ export const isInvalidForVersion = (migration: OutOfBandMigrationFields, version
         return migration.progress !== 0 && !migration.nonDestructive
     }
 
+    // TODO Ensure check for invalid migration ranges
```
eseliger, line 420: is this TODO still valid?

client/web/src/site-admin/SiteAdminMigrationsPage.tsx
```diff
@@ -417,11 +417,15 @@ export const isInvalidForVersion = (migration: OutOfBandMigrationFields, version
...
+
     if (migration.deprecated) {
         // Migrations only store major/minor version components
         const deprecated = parseVersion(`${migration.deprecated}.0`)
         if (deprecated && version.major === deprecated.major && version.minor >= deprecated.minor) {
             return migration.progress !== 1
+        } else if (deprecated && version.major > deprecated.major) {
```
eseliger, line 427: don't we also need to check for minor here? Deprecations can happen in either.

internal/oobmigration/runner.go
eseliger, line 210: Instead of writing a 1 here, should we run the progress query instead? That way we'll never obfuscate any migrations that hadn't actually finished.

internal/oobmigration/runner.go
```diff
@@ -238,6 +248,9 @@ func (r *Runner) startInternal(shouldRunMigration func(m Migration) bool) {
 			if !ok {
 				continue
 			}
+
+			// TODO: Refresh Progress Before Migrating
```
eseliger, line 252: is this TODO still valid?

client/web/src/site-admin/SiteAdminMigrationsPage.tsx
Chickensoupwithrice: Minor version is checked on line 425

eseliger: god, you're right - nvm me

client/web/src/site-admin/SiteAdminMigrationsPage.tsx
DaedalusG: Yea, this still needs review, I was thinking we shouldn't display deprecated migrations and currently it does.

internal/oobmigration/runner.go
DaedalusG: Yea that approach is probably better, although in theory it shouldn't be possible to start a frontend if theres a incomplete and deprecated migration: https://github.com/sourcegraph/sourcegraph/pull/21670/files

internal/oobmigration/runner.go
DaedalusG: No this one can go

cmd/frontend/graphqlbackend/site.go
```diff
@@ -355,8 +355,9 @@ func (r *siteResolver) UpgradeReadiness(ctx context.Context) (*upgradeReadinessR
 	}
 
 	return &upgradeReadinessResolver{
-		logger: r.logger.Scoped("upgradeReadiness"),
-		db:     r.db,
+		logger:  r.logger.Scoped("upgradeReadiness"),
+		db:      r.db,
+		version: r.ProductVersion(),
```
DaedalusG, line 360: This is set later in the `upgradeReadinessResolver.Init()` method, but I believe there are cases where RequiredOutOfBandMigrations may be called before the Init method gives the instance a version value. It should be okay because Init will overwrite the version when called. Right now the version is inherited from the siteResolver which just sets version by the version set in frontend. cc: @unknwon

cmd/frontend/graphqlbackend/site.go
```diff
@@ -507,13 +508,15 @@ func (r *upgradeReadinessResolver) SchemaDrift(ctx context.Context) ([]*schemaDr
...
 }
 
-// isRequiredOutOfBandMigration returns true if a OOB migration is deprecated not
-// after the given version and not yet completed.
-func isRequiredOutOfBandMigration(version oobmigration.Version, m oobmigration.Migration) bool {
-	if m.Deprecated == nil {
+// isRequiredOutOfBandMigration returns true if an OOB migration will be deprecated in the latest version and has not progressed to completion.
+func isRequiredOutOfBandMigration(currentVersion, latestVersion oobmigration.Version, m oobmigration.Migration) bool {
+	// If the migration is not marked as deprecated, or was deprecated before the current product version, it is not required.
+	if m.Deprecated == nil || oobmigration.CompareVersions(*m.Deprecated, currentVersion) == oobmigration.VersionOrderBefore {
```
DaedalusG, line 514: This is where the scary warnings were coming from, in dev where currentVersion is `0.0.0+dev` you'll still see warnings for deprecated migrations, but I've tested by arbitrarily setting the `currentVersion` and it will only display the desired migrations

DaedalusG: i.e. we were always returning true for any deprecated migrations which were incomplete and deprecated before the current version. This is preupgrade warning, there shouldn't exist a state where the instance is running and deprecated migrations are incomplete and deprecated prior to the current version.

cmd/frontend/graphqlbackend/site.go
```diff
@@ -524,7 +527,15 @@ func (r *upgradeReadinessResolver) RequiredOutOfBandMigrations(ctx context.Conte
 	if !updateStatus.HasUpdate() {
 		return nil, nil
 	}
-	version, _, ok := oobmigration.NewVersionAndPatchFromString(updateStatus.UpdateVersion)
+
+	// The current sourcegraph product version
+	currentVersion, _, ok := oobmigration.NewVersionAndPatchFromString(r.version)
```
DaedalusG, line 532: Please check this close, I think the resolver version here is essentially the productVersion inherited by the siteResolver

cmd/worker/internal/migrations/init.go
```diff
@@ -56,21 +57,36 @@ func (m *migrator) Routines(startupCtx context.Context, observationCtx *observat
...
+	currentVersion, err := currentVersion(observationCtx.Logger)
 	if err != nil {
 		return nil, err
 	}
 
+	firstVersionString, _, err := upgradestore.New(db).GetFirstServiceVersion(startupCtx)
+	if err != nil {
+		return nil, err
+	}
+
```
DaedalusG, line 69: Is there a cleaner way to do this?

internal/oobmigration/runner.go
```diff
@@ -195,13 +195,18 @@ func (r *Runner) UpdateDirection(ctx context.Context, ids []int, applyReverse bo
...
-func (r *Runner) Start(currentVersion Version) {
+func (r *Runner) Start(currentVersion, firstVersion Version) {
 	r.startInternal(func(migration Migration) bool {
 		if CompareVersions(currentVersion, migration.Introduced) == VersionOrderBefore {
 			// current version before migration introduction
 			return false
 		}
 
+		if CompareVersions(firstVersion, *migration.Deprecated) == VersionOrderAfter {
+			// instance initialized after migration deprecation
```
DaedalusG, line 206: I had to filter the Sourcegraph init version quite a ways through the setup of the migration runner to get to this check. I don't think we were running these anyway, but better safe than sorry.

client/web/src/site-admin/SiteAdminMigrationsPage.story.tsx
```diff
@@ -26,7 +26,7 @@ export const InvalidPreIntroduction: StoryFn = () => (
         {props => (
             <SiteAdminMigrationsPage
                 {...props}
-                fetchAllMigrations={(): Observable<OutOfBandMigrationFields[]> => of(migrations)}
+                fetchRelevantMigrations={(): Observable<OutOfBandMigrationFields[]> => of(migrations)}
```
unknwon, line 29: nit: I think this is more accurate, but "relevent" seemingly works too 😂 
```suggestion
                fetchUndeprecatedMigrations={(): Observable<OutOfBandMigrationFields[]> => of(migrations)}
```

client/web/src/site-admin/SiteAdminMigrationsPage.tsx
```diff
@@ -127,6 +127,8 @@ export const SiteAdminMigrationsPage: React.FunctionComponent<
         [migrationsOrError]
     )
 
+    // TODO: we currently just display ll migrations here. We should filter out migrations which have a deprecated value before the firstVersion of the Sourcegraph instance.
```
unknwon, line 130:
```suggestion
    // TODO: we currently just display all migrations here. We should filter out migrations which have a deprecated value before the firstVersion of the Sourcegraph instance.
```

cmd/frontend/graphqlbackend/oobmigrations.go
```diff
@@ -32,7 +34,10 @@ func (r *schemaResolver) OutOfBandMigrationByID(ctx context.Context, id graphql.
 }
 
 // OutOfBandMigrations resolves all registered single out-of-band migrations.
-func (r *schemaResolver) OutOfBandMigrations(ctx context.Context) ([]*outOfBandMigrationResolver, error) {
+// If filterDeprecatedBeforeFirstVersion is set to true, migrations deprecated before the init version of sourcegraph will be filtered out.
+func (r *schemaResolver) OutOfBandMigrations(ctx context.Context, args *struct {
+	FilterDeprecatedBeforeFirstVersion *bool
```
unknwon, line 39: nit: not a native speaker, but i think "exclude" captures the intention better, as "filter" could be seen as "filter only for ...". Alternatively, "FilterOutDeprecatedBeforeFirstVersion" is equally better.
```suggestion
	ExcludeDeprecatedBeforeFirstVersion *bool
```

cmd/frontend/graphqlbackend/oobmigrations.go
```diff
@@ -43,6 +48,33 @@ func (r *schemaResolver) OutOfBandMigrations(ctx context.Context) ([]*outOfBandM
...
+			for _, migration := range migrations {
+				// If the migration has no deprecated value, append it to the list.
+				if migration.Deprecated == nil {
+					filtered = append(filtered, migration)
+					continue
+				}
+
+				// If the migration is deprecated before the first version of sourcegraph, don't append it to the list.
+				if oobmigration.CompareVersions(firstVersion, *migration.Deprecated) != oobmigration.VersionOrderAfter {
+					filtered = append(filtered, migration)
```
unknwon, line 71: I understand it would be redundant in the current form of this function/loop, but adding this `continue` now would give ourselves an easier life in the future. Because at a later point, we would have to think really hard that "this did not have continue was because we shouldn't, or it did not matter?"

```suggestion
					filtered = append(filtered, migration)
					continue
```

cmd/frontend/graphqlbackend/site.go
```diff
@@ -508,24 +508,43 @@ func (r *upgradeReadinessResolver) SchemaDrift(ctx context.Context) ([]*schemaDr
...
-// isRequiredOutOfBandMigration returns true if a OOB migration is deprecated not
-// after the given version and not yet completed.
-func isRequiredOutOfBandMigration(version oobmigration.Version, m oobmigration.Migration) bool {
-	if m.Deprecated == nil {
+// isRequiredOutOfBandMigration returns true if an OOB migration will be deprecated in the latest version and has not progressed to completion.
+func isRequiredOutOfBandMigration(currentVersion, latestVersion oobmigration.Version, m oobmigration.Migration) bool {
+	// If current version is dev, no deprecated migrations are required.
+	if currentVersion.Dev && m.Deprecated != nil {
 		return false
 	}
```
unknwon, line 516: Based on the code comment, it sounds like we shouldn't check for `m.Deprecated` if `Dev` is true (the "kill switch").
```suggestion
	if currentVersion.Dev {
		return false
	}
```

unknwon (approved):
LGTM, but not an authoritative approval.

client/web/src/site-admin/backend.ts
```diff
@@ -690,6 +692,42 @@ export function fetchAllOutOfBandMigrations(): Observable<OutOfBandMigrationFiel
     )
 }
 
+/**
+ * Fetches out-of-band migrations which have not been deprecated or were not deprecated before the Sourcegraph init version.
+ */
+export function fetchRelevantOutOfBandMigrations(): Observable<OutOfBandMigrationFields[]> {
```
camdencheek, line 698: Unsolicited drive-by 🙂:

This looks pretty much exactly the same as the query above, except with the added `FilterDeprecatedBeforeFirstVersion` option. 

I'd recommend we just add a boolean parameter to the query above to keep the fragments and stuff in sync.

Something like this:

```
export function fetchOutOfBandMigrations(onlyRelevant: boolean): Observable<OutOfBandMigrationFields[]> {
    return requestGraphQL<OutOfBandMigrationsResult, OutOfBandMigrationsVariables>(
        gql`
            query OutOfBandMigrations($onlyRelevant: Boolean = false) {
                outOfBandMigrations(FilterDeprecatedBeforeFirstVersion: $onlyRelevant) {
                    ...OutOfBandMigrationFields
                }
            }
            fragment OutOfBandMigrationFields on OutOfBandMigration {
                id
                team
                component
                description
                introduced
                deprecated
                progress
                created
                lastUpdated
                nonDestructive
                applyReverse
                errors {
                    message
                    created
                }
            }
        `, 
        { onlyRelevant },
    ).pipe(
        map(dataOrThrowErrors),
        map(data => data.outOfBandMigrations)
    )
}
```

client/web/src/site-admin/SiteAdminMigrationsPage.tsx
DaedalusG: Leftover comment removing

cmd/frontend/graphqlbackend/oobmigrations.go
DaedalusG: I agree this is a better title 👍🏼

client/web/src/site-admin/backend.ts
DaedalusG: This is much better yea, implemented 👍🏼

PR #61777, opened 2024-04-11
pipeline: more pipeline config fixes

burmudar:
Sooooooooooooo. Backport builds are still failing see https://buildkite.com/sourcegraph/sourcegraph/builds/268392

The base branch is empty when it's a merge into a branch - when a PR for a backport gets merged into the backport branch ... In that case - what we actually want to do:

1. Diff with the previous commit on _this_ branch
2. NOT DIFF WITH origin/main ... right?

## Test plan
CI and attempting a merge into _this_ branch


Strum355 (approved):
There's only so many ways to do it, one of ems gotta be right lol

Issue #61629, opened 2024-04-05
Add "open in editor" action


fkling:
```[tasklist]
### Tasks
- [x] Implement new button
- [x] Add automated tests
- [ ] Implement popover to configure editor
```

bahrmichael:
Recap of what the OpenInEditorActionItem.tsx is currently doing (i.e. the things we need to do in the svelte version):

- Retrieving settings from the backend, and error handling (request fails or settings are invalid)
- Popover to change the editor integration (and its related race conditions)
- Handling multiple editors (e.g. Show a button with the editor icon for each registered editor )
- Constructing the URL that opens the file in the IDE
- Not showing the dropdown on small viewports, since we assume they won't have an IDE on their phones anyway

Open questions:
- Can we drop the code to migrate legacy settings? It's from September '22.

bahrmichael:
Here's where the button would be added (once that PR is merged): https://github.com/sourcegraph/sourcegraph/pull/61723/files#diff-21381efc7ce3a33522c1ffcc88edefa0ff8273c5a9b9d2c7447bd15adc4a6736R104-R109

camdencheek:
@bahrmichael can this be closed after #61782?

bahrmichael:
Will move the remaining work to a new task 👍

PR #61782, opened 2024-04-11
feat: add button to open a file in the editor for the webapp rewrite

bahrmichael:
For https://github.com/sourcegraph/sourcegraph/issues/61629

This is the first iteration of a button to open the currently viewed file in your IDE. I reduced the scope of this task, and will follow up with more PRs. See https://github.com/sourcegraph/sourcegraph/issues/61629#issuecomment-2047686677.

How it looks when you've configured an editor:

Screenshot 2024-04-11 at 13 08 19

How it looks when you did not configure an editor:

Screenshot 2024-04-15 at 13 20 14

Some utility files and icons have been copied from the old webapp. 

## Test plan

My plan is to get feedback on this work, review the scoping and what is important to implement now, and then write tests depending on how much coverage we need. There are no automated tests (beyond what was copied) yet, but it manual testing was good.


client/web-sveltekit/src/lib/repo/open-in-editor/EditorIcon.svelte
```diff
@@ -0,0 +1,11 @@
+<script lang="ts">
+    import type { Editor } from '$lib/repo/open-in-editor/editors'
+
+    export let editor: Editor
+
+    $: svg = `/images/editors/${editor.id}.svg`
+</script>
+
+{#if editor}
+    <img class:icon-inline={true} height={24} width={24} src={svg} alt={editor.name} />
```
bahrmichael, line 10: FYI: I couldn't go with the Icon component, because it takes the `svgPath`, but we have the file path here. I'm not confident about svgs, but noticed that the ones used here seem to have a different structure than what Icon would expect.

client/web-sveltekit/src/lib/repo/open-in-editor/build-url.ts
bahrmichael, line 1: FYI: This file was copied from the web project.

client/web-sveltekit/src/lib/repo/open-in-editor/editor-settings.ts
bahrmichael, line 1: FYI: This file was copied from the web project.

client/web-sveltekit/src/lib/utils/url.test.ts
bahrmichael, line 1: FYI: This file was copied from the web project.

client/web-sveltekit/src/lib/utils/url.ts
bahrmichael, line 1: FYI: This file was copied from the web project.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte
```diff
@@ -102,6 +103,11 @@
     <FileHeader>
         <FileIcon slot="icon" file={blob} inline />
         <svelte:fragment slot="actions">
+            {#await data.externalServiceType}
```
Comment (bahrmichael) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte:106]: I'm not confident about my use of async data here, so feel free to suggest improvements. My idea here is that because we have to wait for the query to resolve, we can show a placeholder. Since the query resolves faster (with reliable internet) than a human would click this button, this is fine to do. The buggy behavior when the serviceType is not known yet has a reasonably small chance to happen.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte
```diff
@@ -102,6 +103,9 @@
     <FileHeader>
         <FileIcon slot="icon" file={blob} inline />
         <svelte:fragment slot="actions">
+            {#await data.externalServiceType then externalServiceType}
```
Comment (bahrmichael) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte:106]: My assumption is that we don't need to display an error message. Firstly I expect much more to break when we can't resolve the repo info, and secondly an error message in place of the editor button doesn't leave a lot of space to inform users what they should do. In that case I feel like just dropping the feature until the errors resolve should be fine.

fkling:
I won't have time to give this a thorough review today, so just a couple of high level points:

- Do we need to copy the business logic? It seems to me we could just import it. Similar with the URL builders which are actually already importing (see `src/lib/web.ts` I think).
- I don't think the images will work this way in production because the static files are not copied to where the `img` element is trying to access them. We either need the correct image path prefix or reference the images differently. But I'd have to take a closer look to give suggestions for that.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte
Comment (camdencheek) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte:106]: Will this log an error to the console? Ideally, we shouldn't just silently drop errors or we might just never know we broke something

client/web-sveltekit/src/lib/repo/open-in-editor/EditorIcon.svelte
```diff
@@ -0,0 +1,11 @@
+<script lang="ts">
+    import type { Editor } from '$lib/web'
+
+    export let editor: Editor
+
+    $: svg = `/images/editors/${editor.id}.svg`
+</script>
+
+{#if editor}
+    <img class:icon-inline={true} height={24} width={24} src={svg} alt={editor.name} />
```
fkling: I don't think this would work in production because static assets are placed in a different folder. However, moving the image files to `src/lib/...` and just importing them might work. See https://kit.svelte.dev/docs/images#vite-s-built-in-handling . The imported value will be a URL to the image.

client/web-sveltekit/src/lib/repo/open-in-editor/EditorIcon.svelte
```diff
@@ -0,0 +1,11 @@
+<script lang="ts">
+    import type { Editor } from '$lib/web'
+
+    export let editor: Editor
+
+    $: svg = `/images/editors/${editor.id}.svg`
+</script>
+
+{#if editor}
```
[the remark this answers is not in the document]
bahrmichael: Dropped this when applying https://github.com/sourcegraph/sourcegraph/pull/61782#discussion_r1563226361.

client/web-sveltekit/src/lib/repo/open-in-editor/OpenInEditor.svelte
```diff
@@ -0,0 +1,62 @@
...
+    let openInEditor = $settings?.openInEditor
+
+    const editorSettingsErrorMessage = getEditorSettingsErrorMessage(openInEditor)
+    const editorIds = (openInEditor as EditorSettings | undefined)?.editorIds ?? []
+    const editors = !editorSettingsErrorMessage ? editorIds.map(getEditor) : undefined
+
+    const sourcegraphBaseURL = SourcegraphURL.from($page.url).toString();
+
+    const { repoName, filePath, position, range } = parseBrowserRepoURL(window.location.href)
+    const start = position || range?.start
```
fkling: These should all be reactive otherwise they will only be computed once when the component is created. Instead of `window.location.href` you should use `$page.url` otherwise that code won't rerun when the URL changes.
```suggestion
    $: editorSettingsErrorMessage = getEditorSettingsErrorMessage(openInEditor)
    $: editorIds = (openInEditor as EditorSettings | undefined)?.editorIds ?? []
    $: editors = !editorSettingsErrorMessage ? editorIds.map(getEditor) : undefined

    $: sourcegraphBaseURL = SourcegraphURL.from($page.url).toString();

    $: { repoName, filePath, position, range } = parseBrowserRepoURL($page.url)
    $: start = position ?? range?.start
```

client/web-sveltekit/src/lib/repo/open-in-editor/EditorIcon.svelte
```diff
@@ -0,0 +1,11 @@
...
+    import type { Editor } from '$lib/web'
+    export let editorId: Editor['id']
+
+    let Icon: any;
+    $: import(`./editors/${editorId}.svelte`).then(i => Icon = i.default);
+</script>
+
+ {#if Icon}
+    <Icon />
+ {/if}
```
fkling, line 11: I'm wondering whether this works:

```suggestion
</script>

 {#await import(`./editors/${editorId}.svelte`) then {default: Icon}}
    <Icon />
 {/await}
```

(or something like that)

client/web-sveltekit/src/lib/repo/open-in-editor/OpenInEditor.svelte
```diff
@@ -0,0 +1,62 @@
...
+    let openInEditor = $settings?.openInEditor
+
+    const editorSettingsErrorMessage = getEditorSettingsErrorMessage(openInEditor)
+    const editorIds = openInEditor?.editorIds ?? []
+    const editors = !editorSettingsErrorMessage ? editorIds.map(getEditor) : undefined
+
+    const sourcegraphBaseURL = SourcegraphURL.from($page.url).toString();
+
+    const { repoName, filePath, position, range } = parseBrowserRepoURL(window.location.href)
+    const start = position ?? range?.start
```
fkling: These should all be reactive otherwise the code only runs once when the component is created. You should also use `$page.url` instead of `window.location.href`, otherwise the switching to a new URL won't update the variables.

```suggestion
    let openInEditor = $settings?.openInEditor

    $: editorSettingsErrorMessage = getEditorSettingsErrorMessage(openInEditor)
    $: editorIds = openInEditor?.editorIds ?? []
    $: editors = !editorSettingsErrorMessage ? editorIds.map(getEditor) : undefined

    $: sourcegraphBaseURL = SourcegraphURL.from($page.url).toString();

    $: { repoName, filePath, position, range } = parseBrowserRepoURL($page.url)
    $: start = position ?? range?.start
```

client/web-sveltekit/src/lib/repo/open-in-editor/OpenInEditor.svelte
```diff
@@ -0,0 +1,62 @@
...
+</script>
+
+{#if editors}
+    {#each editors as e, i}
```
fkling, line 26: `editor` instead of `e` would be more readable (no Go conventions here ;) )

client/web-sveltekit/src/lib/repo/open-in-editor/OpenInEditor.svelte
```diff
@@ -0,0 +1,62 @@
...
+    import { getEditorSettingsErrorMessage } from './build-url'
+    import Tooltip from '$lib/Tooltip.svelte'
+    import EditorIcon from '$lib/repo/open-in-editor/EditorIcon.svelte'
+    import { settings } from '$lib/stores'
+    import { mdiCodeBraces } from '@mdi/js'
+    import Icon from '$lib/Icon.svelte';
+    import {SourcegraphURL} from '$lib/common';
+    import {page} from '$app/stores';
+
+    export let externalServiceType: string = ''
```
fkling, line 11: Does this have to be a string? I think we have GraphQL types for those, or not?

client/web-sveltekit/src/lib/repo/open-in-editor/OpenInEditor.svelte
```diff
@@ -0,0 +1,62 @@
...
+
+    let openInEditor = $settings?.openInEditor
+
+    const editorSettingsErrorMessage = getEditorSettingsErrorMessage(openInEditor)
+    const editorIds = (openInEditor as EditorSettings | undefined)?.editorIds ?? []
+    const editors = !editorSettingsErrorMessage ? editorIds.map(getEditor) : undefined
+
+    const sourcegraphBaseURL = SourcegraphURL.from($page.url).toString();
```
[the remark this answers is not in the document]
bahrmichael: Good catch! It was indeed the URL with the path. I'm adding a fix.

client/web-sveltekit/src/lib/repo/open-in-editor/build-url.ts
fkling, line 1: We can still use the function from `client/web/`. If you are concerned about the link to `/users/settings`: Clicking it will simply load the React app, so that's not a problem.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/page.gql
```diff
@@ -27,6 +27,14 @@ query BlobPageQuery($repoName: String!, $revspec: String!, $path: String!) {
...
 }
 
+query RepoInfoQuery($repoName: String!) {
+    repository(name: $repoName) {
+        externalRepository {
+            serviceType
+        }
+    }
+}
+
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/page.gql:37]: I wouldn't create a separate query for this. We want to keep the number of queries to a minimum and conceptually this is part of the repository data that we are fetching anyways when loading repository information.

Instead you can do the following:

Create a fragment here:

```gql
fragment BlobPage_ResolvedRevision on Repository {
  externalRepository {
    serviceType
  }
}
```

and include that in the `ResolvedRevision` fragment in `src/routes/[...repo=reporev]/layout.gql` (see `RepoPage_ResolvedRevision` as an example).

Then in `+page.ts` you do this instead:

```js
externalServiceType: parent()
    .then(({resolvedRevision}) => resolvedRevision.externalRepository?.serviceType),
```

client/web-sveltekit/src/lib/repo/open-in-editor/EditorIcon.svelte
bahrmichael: It does! I like that :)

client/web-sveltekit/src/lib/repo/open-in-editor/OpenInEditor.svelte
bahrmichael: Destructuring needs another pair of brackets around the statement, but apart from that everything works!

```
$: ({ repoName, filePath, position, range } = parseBrowserRepoURL($page.url.toString()))
```

client/web-sveltekit/src/lib/repo/open-in-editor/OpenInEditor.svelte
bahrmichael: :D

client/web-sveltekit/src/lib/repo/open-in-editor/build-url.ts
bahrmichael: My main issue was with the markdown that the client/web function produces. We don't have markdown rendering in our tooltips, so it looked a bit ugly. When I added markdown support, I noticed that the styling of the tooltip got a bit wonky and more importantly I noticed that the Tooltip doesn't stay open when I hover from the button to the tooltip, rendering its link useless. Instead of digging deeper here, I thought it's a cleaner solution to give the button the href action.

client/web-sveltekit/src/lib/repo/open-in-editor/EditorIcon.svelte
bahrmichael: Thanks! I wasn't sure what the best way to handle svgs in svelte was. That makes much more sense now :)

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/page.gql
Comment (bahrmichael) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/page.gql:37]: Makes sense, I'm adding that :)

client/web-sveltekit/src/lib/repo/open-in-editor/OpenInEditor.svelte
```diff
@@ -0,0 +1,61 @@
...
+                        buildRepoBaseNameAndPath(repoName, externalServiceType, filePath),
+                        start,
+                        openInEditor,
+                        sourcegraphBaseURL,
+                        i
+                    ).toString()}
+                    target="_blank"
+                    rel="noopener noreferrer"
+                >
+                    <EditorIcon editorId={e.id} />
```
fkling, line 40: Forgot this earlier: We should somehow fallback to a default icon if we don't have one for the configured editor, otherwise nothing will be displayed when the actions are in "compact" mode.

bahrmichael: Looking at fallbacks and error handling I feel like a script based approach is the way to go. Here's an example of what I came up with:

```svelte
<script lang="ts">
    import type {Editor} from '$lib/web'
    import DefaultEditorIcon from './DefaultEditorIcon.svelte';
    import type {SvelteComponent} from 'svelte';

    export let editorId: Editor['id']

    let Icon: typeof SvelteComponent | typeof DefaultEditorIcon;
    $: import(`./editors/${editorId}.svelte`)
        .then(i => Icon = i.default)
        .catch((err) => {
            console.error(err);
            Icon = DefaultEditorIcon
        })
</script>
```

If we were to do it in the component body, it would look something like the version below. That version has an error because `error` is unhandled, and I don't feel like adding a `console.error` (or a telemetry call) into the component body would be the right way to go.

```svelte
{#await import(`./editors/${editorId}.svelte`) then {default: Icon}}
    <Icon />
{:catch error}
    <DefaultEditorIcon />
{/await}
```

What do you think? Is there another way to do this within the Svelte body?

fkling: Just having `{:catch}` (without variable name) is valid too.

fkling: Alternatively, we do have a list of known editor IDs, right? We could test against that list before we try to fetch the icon. Maybe we could even generate that list from the list of files at build time.

client/web-sveltekit/src/lib/repo/open-in-editor/OpenInEditor.svelte
bahrmichael: Nope, I'm adding the type from GraphQL 👍

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte
Comment (bahrmichael) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte]: I've updated this so it won't swallow the error anymore. I opted for a very prominent alert, as an error here indicates that we broke an API.

client/web-sveltekit/src/lib/repo/open-in-editor/build-url.ts
bahrmichael: I gave the Popover component a quick spin, and it looks like I'd have to do a significant amount of CSS to align the look and feel to the other actions. In favor of consistent behavior I would stick with the current version of the Tooltip. It may also feel odd that some tooltips of the actions would close on hover and others don't. If you don't have a strong opinion here I would keep the current approach.

https://github.com/sourcegraph/sourcegraph/assets/1830132/386b4f21-d842-4fbf-8470-831c4068acd5

camdencheek:
FYI, even though the UI is confusing and suggests that your branch needs to be up to date with `main` to merge, you don't actually. Merge will be blocked if there are conflicts, but right now the only reason it's blocked is it doesn't have an approving review (which I'll leave to Felix)

CleanShot 2024-04-15 at 11 57 52@2x

client/web-sveltekit/src/lib/repo/open-in-editor/OpenInEditor.svelte
```diff
@@ -0,0 +1,61 @@
...
+    import Tooltip from '$lib/Tooltip.svelte'
+    import EditorIcon from '$lib/repo/open-in-editor/EditorIcon.svelte'
+    import {settings} from '$lib/stores'
+    import {page} from '$app/stores';
+    import type {ExternalRepository} from '$lib/graphql-types';
+    import DefaultEditorIcon from '$lib/repo/open-in-editor/DefaultEditorIcon.svelte';
+
+    export let externalServiceType: ExternalRepository['serviceType'] = ''
+
+    let openInEditor = $settings?.openInEditor
```
fkling, line 13: This should also be reactive since `settings` is a store and `$settings` might change over time. I don't believe it does right now but it would be a subtle bug otherwise.

```suggestion
    $: openInEditor = $settings?.openInEditor
```

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte
```diff
@@ -102,6 +103,15 @@
...
         <svelte:fragment slot="actions">
+            {#await data.externalServiceType}
+                <OpenInEditor />
+            {:then externalServiceType}
+                <OpenInEditor {externalServiceType} />
+            {:catch error}
+                <Alert variant="danger">
+                    Unable to fetch external service type: {error.message}
+                </Alert>
+            {/await}
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte:114]: I don't think this is necessary though I also acknowledge that it's not obvious why this can't error. If `parent` fails to resolves we will likely render an error elsewhere.

Regardless rendering the alert here probably doesn't look good either way. As a precaution we could instead catch any error in the loader and log it in the console (see below).

Eventually we will collect such errors with sentry (I think).

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.ts
```diff
@@ -45,5 +45,8 @@ export const load: PageLoad = ({ parent, params, url }) => {
                       .then(mapOrThrow(result => result.data?.repository?.commit?.diff.fileDiffs.nodes[0] ?? null)),
               }
             : null,
+        externalServiceType: parent().then(
+            ({ resolvedRevision }) => resolvedRevision.repo?.externalRepository?.serviceType
+        ),
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.ts:50]:
```suggestion
        externalServiceType: parent().then(
            ({ resolvedRevision }) => resolvedRevision.repo?.externalRepository?.serviceType
        )
        .catch(error => {
          console.error("Failed to fetch repository data:", error)
          return null
        }),
```

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte
```diff
@@ -102,6 +103,15 @@
     <FileHeader>
         <FileIcon slot="icon" file={blob} inline />
         <svelte:fragment slot="actions">
+            {#await data.externalServiceType}
+                <OpenInEditor />
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte:107]: What would happen if someone clicks the button while the data is not available yet? Can that even happen?

fkling (approved):
Some minor comments inside, overall looks good!

fkling:
FWIW, here is an example of how images can be used by importing as "URL"s. That would also be an acceptable approach instead of converting them all to Svelte files, especially if you can use the existing files.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte
Comment (bahrmichael) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte:114]: Applied your suggestion so that we get those errors through sentry. That means I can reduce this part to just an `await ... then ...` and an inner `if`. Looks clean to me.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte
Comment (bahrmichael) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte:107]: Not applicable anymore after https://github.com/sourcegraph/sourcegraph/pull/61782#discussion_r1570429089

bahrmichael:
> FWIW, here is an example of how images can be used by importing as "URL"s. That would also be an acceptable approach instead of converting them all to Svelte files, especially if you can use the existing files.
> 
> https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@756a245cee6a1ee8b89f6a7df946cc9e28cb2e8e/-/blob/client/web-sveltekit/src/lib/images.ts?L1-3

That's a nice approach!

However, since we're rebuilding the blob page (and the old OpenInEditorAction is only used there), I'm pushing for duplicating the images so we can later more easily remove the old ones. As far as I see the editor images are not used anywhere else. This keeps cross-client dependencies lower.

Screenshot 2024-04-18 at 12 20 26

fkling:
Thank you for the tests!

bahrmichael:
There were some weird UI Review differences that I decided to accept even they look a bit wrong. I have no idea what could cause those.

Screenshot 2024-04-18 at 14 26 06

PR #61788, opened 2024-04-11
bazel: migrate dind dockerfile to rules_oci

Strum355:
From @willdollman's ask:
> when updating the underlying dind image, we update the [upstream image hash here](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@5352138e5ed0e6ab96c95417d90a3d65f28aa769/-/blob/docker-images/dind/Dockerfile?L1). But I don’t see how to get the output of our Dockerfile onto Docker Hub - [oci_deps.bzl](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@5352138e5ed0e6ab96c95417d90a3d65f28aa769/-/blob/dev/oci_deps.bzl?L221-225) and the [BUILD.bazel](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@5352138e5ed0e6ab96c95417d90a3d65f28aa769/-/blob/docker-images/dind/BUILD.bazel?L6) reference the image that’s already on Dockerhub and push to the same image name, which seems to be a circular dependency.
> tl;dr there doesn’t seem to be a way to use CI to push an updated dind image

To address this, Im migrating from the Dockerfile + build.sh approach to Bazel-native rules_oci approach (integrating with our existing ./dev/ci/push_all.sh script).

To remove the docker-{buildx,compose} files, given that we cant run `rm <file>` with rules_oci, I take the approach of how this is actually implemented in the OCI spec, using [whiteout files](https://github.com/opencontainers/image-spec/blob/main/layer.md#whiteouts).

## Test plan

`BUILDKITE_COMMIT=deadbeef VERSION=6.9.0 bazel run --stamp --workspace_status_command=./dev/bazel_stamp_vars.sh //docker-images/dind:image_tarball` to build&stamp&load the oci image

Then [dive](https://github.com/wagoodman/dive) to inspect the image to make sure it removed files as expected
image


docker-images/dind/BUILD.bazel
```diff
@@ -1,9 +1,43 @@
...
+        "org.opencontainers.image.documentation=https://sourcegraph.com/docs/",
+        "com.sourcegraph.github.url=https://github.com/sourcegraph/sourcegraph/commit/COMMIT_SHA",
+    ],
+)
+
+pkg_tar(
+    name = "tar_whiteout",
+    srcs = [],
+    empty_files = [
+        # https://github.com/opencontainers/image-spec/blob/main/layer.md#whiteouts
```
burmudar, line 28:
```suggestion
        # we can't run rm <file> so we use whiteout files to do it - see the following link for more info
        # https://github.com/opencontainers/image-spec/blob/main/layer.md#whiteouts
```

willdollman (approved):
lgtm thanks!

PR #61802, opened 2024-04-11
sg: install postgres 12 instead of 15

Strum355:
https://github.com/sourcegraph/sourcegraph/pull/48807 changed `sg setup` to install postgres 12 instead of 15. Now that https://github.com/sourcegraph/sourcegraph/pull/61741 is merged, we must be running postgres 12 locally so that pg_dump will function correctly

## Test plan

@burmudar tested `brew install postgresql@12`

### DIFF
--- a/dev/sg/dependencies/mac.go
+++ b/dev/sg/dependencies/mac.go
@@ -165,9 +165,9 @@ If you've installed PostgreSQL with Homebrew that should be the case.
 If you used another method, make sure psql is available.`,
 				Check: checkAction(check.Combine(
 					check.InPath("psql"),
-					check.CompareSemanticVersion("psql", "psql --version", ">= 15.0"),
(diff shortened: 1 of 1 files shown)


dev/sg/dependencies/mac.go
```diff
@@ -183,15 +183,15 @@ If you used another method, make sure psql is available.`,
 					}
 					return check.PostgresConnection(ctx)
 				},
-				Description: `Sourcegraph requires the PostgreSQL database (v15+) to be running.
+				Description: `Sourcegraph requires the PostgreSQL database (v12) to be running.
```
burmudar, line 186: We have an ubuntu deps as well where this would need to be installed 🤔

burmudar (approved):
Can do ubuntu in a follow up

PR #61805, opened 2024-04-11
Sveltekit: implement new avatar design

jasonhawkharris:
This adjusts the design parameters for the Avatar component when no avatarURL is available. 

[Figma](https://www.figma.com/file/EPgl0qVbqGSumvf52brhda/Code-Search?type=design&node-id=326-3404&mode=design&t=AJLHXqlTLf95fMdd-0)

Dark
Screenshot 2024-04-11 at 2 12 33 PM

Light
Screenshot 2024-04-11 at 2 11 23 PM

## Test plan
Manual testing with stories on both light and dark mode.

### DIFF
--- a/client/web-sveltekit/src/lib/Avatar.svelte
+++ b/client/web-sveltekit/src/lib/Avatar.svelte
@@ -42,8 +42,9 @@
 <style lang="scss">
     span {
         z-index: 1;
-        color: var(--white);
+        color: var(--gray-06);
         font-size: calc(var(--size) * 0.5);
+        font-weight: 300;
     }
 
     img,
@@ -63,7 +64,7 @@
         align-items: center;
         justify-content: center;
         position: relative;
-        background: linear-gradient(to bottom, var(--logo-purple), var(--logo-orange));
+        background: var(--gray-03);
     }
 
     div::after {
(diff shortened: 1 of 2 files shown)


client/web-sveltekit/src/lib/Avatar.svelte
```diff
@@ -42,8 +42,9 @@
 <style lang="scss">
     span {
         z-index: 1;
-        color: var(--white);
+        color: var(--gray-06);
```
camdencheek, line 45: Here, we should use the semantic color `var(--text-muted)`. You can find that in the designs here:

CleanShot 2024-04-11 at 13 55 58@2x

client/web-sveltekit/src/lib/Avatar.svelte
```diff
@@ -63,7 +64,7 @@
         align-items: center;
         justify-content: center;
         position: relative;
-        background: linear-gradient(to bottom, var(--logo-purple), var(--logo-orange));
+        background: var(--gray-03);
```
camdencheek, line 67: Oops, I have no idea what happened to my second comment, but this should also be `var(--secondary)`

camdencheek (approved):
Approved pending color fix

Issue #61653, opened 2024-04-05
Svelte: Implement blame switcher and file metadata


taiyab:
We can get started on implementing the Code/Blame switcher and file metadata header.

Should be sticky to the top of that panel.

Happy to recreate (and somewhat update according to what we feel is missing/unclear) the version of the actual Blame view we already have in React. I'd rather pair with one of you to update any visuals directly in code also.

<img width="767" alt="image" src="https://github.com/sourcegraph/sourcegraph/assets/5337876/ae5fea14-2939-4e5f-8bc4-d571b7b5ce8f">

Figma: https://www.figma.com/file/EPgl0qVbqGSumvf52brhda/Code-Search?type=design&node-id=1947%3A37160&mode=design&t=rs6WkWU36T6PfNDq-1

Issue #60425, opened 2024-02-12
Svelte prototype: add blame view


camdencheek:
This should be relatively easy because most of the logic is in a CodeMirror extension which accepts a function to generate the UI.

taiyab:
https://github.com/sourcegraph/sourcegraph/issues/61653

camdencheek:
Leaving this open for two followups:
- [ ] Popover implementation (like in React app). Should probably consider if we want to implement it the same way.
- [ ] Playwright tests

fkling:
> Popover implementation (like in React app). Should probably consider if we want to implement it the same way.

Personally I'm not a fan of the popover. It feels to heavy for something that should be small piece of information. I don't really know what added value it brings:

- If we want to make the full commit message visible we can add a `title` attribute or a simple popover.
- Reblaming at/before the commit can be done by adding button directly to the blame information.

PR #61809, opened 2024-04-11
Svelte: implement basic git blame view

camdencheek:
This implements a basic blame view in the Svelte blob view. It is a simplified version of what we have in the React webapp, and notably does not include the popover. I think this is still usable as-is, and the PR was getting large, so I cut it off here, but I will update the issue with the followups.

The blame extension itself was pretty straightforward to add to `CodeMirrorBlob.svelte`. 

[Loom recording with quick demo and walkthrough of some of the prerequisite refactors](https://www.loom.com/share/408ce293e6a84e06aaf6080bb76fc30b?sid=335c4b67-9304-4d61-b115-f9f955ecede2)

Closes https://github.com/sourcegraph/sourcegraph/issues/61653
Implements most of https://github.com/sourcegraph/sourcegraph/issues/60425

## Test plan

Manual testing. See loom video.


client/web-sveltekit/src/lib/Avatar.gql
```diff
@@ -1,16 +1,19 @@
 fragment Avatar_User on User {
+    __typename
```
camdencheek, line 2: We were relying on this in the component, but we had a default case in the switch statement that let us fall back to just an empty string for name.

client/web-sveltekit/src/lib/CodeMirrorBlob.svelte
```diff
@@ -44,6 +44,7 @@
             fontSize: 'var(--code-font-size)',
         },
         '.cm-content': {
+            padding: 0,
```
camdencheek, line 47: It seems codemirror has some default padding that looks pretty bad when blame view is enabled.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.ts
```diff
@@ -45,5 +49,17 @@ export const load: PageLoad = ({ parent, params, url }) => {
                       .then(mapOrThrow(result => result.data?.repository?.commit?.diff.fileDiffs.nodes[0] ?? null)),
               }
             : null,
+        blameHunks: isBlame
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.ts:52]: In theory, this lets us preload blame. However, since the blame switcher is not a link, preloading doesn't just happen automatically for the normal case.

client/web-sveltekit/src/routes/search/PersonSearchResult.svelte
```diff
@@ -23,7 +23,12 @@
 <SearchResult>
     <div slot="title">
         <Avatar
-            avatar={{ displayName, username: result.user?.username ?? '', avatarURL: result.user?.avatarURL ?? null }}
+            avatar={{
+                __typename: 'User',
```
camdencheek, line 27: Added `User` to make this well-typed even though it's not coming from GraphQL

client/web/src/repo/blame/useBlameHunks.ts
```diff
@@ -42,7 +42,7 @@ export interface BlameHunk {
                 | undefined
                 | null
                 | {
-                      username: string | null
+                      username: string
```
camdencheek, line 45: Just narrowed the type, which is guaranteed by the backend to be non-null.

client/web-sveltekit/src/lib/CodeMirrorBlob.svelte
```diff
@@ -250,6 +275,8 @@
 <style lang="scss">
     .root {
         display: contents;
+        --blame-decoration-width: 400px;
+        --blame-recency-width: 4px;
```
camdencheek, line 279: Is there a better place for this? I was struggling to find the right place to define these vars.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.ts
```diff
@@ -1,15 +1,44 @@
...
+            concatMap(externalURLs =>
+                blameHunks.pipe(
+                    map(blameHunks => ({
+                        externalURLs,
+                        current: blameHunks,
+                    }))
+                )
+            )
+        )
+    }
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.ts:41]: Apologies for this mess. The final `blameData` depends on the resolved revision and the external URLs. But for performance, we want to load those asynchronously, and only if we actually need the blame data at all. I think we can significantly reduce the complexity here by passing in ExternalURLs separately, and maybe fetching them as part of the resolution step, but this PR was on the verge of getting out of hand.

client/web/src/repo/blame/hooks.ts
camdencheek, line 1: I moved the react-specific stuff into this file. Everything shared is in the sibling `shared.ts` file.

client/web/src/repo/blame/shared.ts
```diff
@@ -0,0 +1,187 @@
...
+            },
+            onerror(event) {
+                subscriber.error(event)
+            },
+        }).then(
+            () => subscriber.complete(),
+            error => subscriber.error(error)
+        )
+    })
+}
```
camdencheek, line 176: The diff doesn't do a good job of showing this, but as part of the refactoring to split out the GraphQL data from the stream data, I split this into three functions:
- `fetchRawBlameHunks` returns an observable which directly mirrors what is coming back from the API
- `fetchBlameHunks` does some lightweight processing on that stream, mapping the data to the `BlameHunk` type, aggregating it into an observable array, and adding event throttling
- `fetchBlameHunksMemoized` adds memoization to the `fetchBlameHunks` observable. It is the only exported function. 

The data from GraphQL gets added by the caller because how we fetch that depends on whether we're in React or Svelte.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.ts
```diff
@@ -45,5 +74,6 @@ export const load: PageLoad = ({ parent, params, url }) => {
                       .then(mapOrThrow(result => result.data?.repository?.commit?.diff.fileDiffs.nodes[0] ?? null)),
               }
             : null,
+        blameData,
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.ts:77]: In theory, this will be preloaded. However, since the only way to get to the blame view is through the switcher, which is not a link, it does not get preloaded. @fkling is there an easy way to make the switcher preload even though it's a button?

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte
```diff
@@ -74,7 +77,9 @@
                 break
             }
             case ViewMode.Blame:
-                // TODO: Implement
+                const url = SourcegraphURL.from($page.url)
+                url.setSearchParameter('view', 'blame')
+                goto(url.toString(), { replaceState: true, keepFocus: true })
```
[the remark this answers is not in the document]
Comment (camdencheek) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte:82]: Yeah, I was torn on how I expect this to work.

I guess it comes down to: if I share a link to a file, do I expect the user I share it with to view the blame column open if I had it open, or do I expect it to have the state the opening user last used? I _think_ I lean towards the latter, but I don't have a strong opinion. 

If we do put it in the URL, is there a best-practice way to make a URL param "sticky"? If doing it manually, we'd need to make links in every other part of the page know about the param, right?

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.ts
```diff
@@ -1,15 +1,44 @@
...
     const client = getGraphQLClient()
     const { repoName, revision = '' } = parseRepoRevision(params.repo)
     const resolvedRevision = resolveRevision(parent, revision)
+    const isBlame = url.searchParams.get('view') === 'blame'
+
+    var blameData: Observable<BlameHunkData> | undefined = undefined
+    if (isBlame) {
+        const externalURLs = client
+            .query(BlobPageExternalURLs, { repoName })
+            .then(mapOrThrow(result => result.data?.repository?.externalURLs))
```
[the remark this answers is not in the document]
Comment (camdencheek) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.ts]: Done!

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.ts
Comment (camdencheek) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.ts:73]: Aha! That's what I was looking for

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.ts
```diff
@@ -16,24 +16,27 @@ export const load: PageLoad = ({ parent, params, url }) => {
     const resolvedRevision = resolveRevision(parent, revision)
     const isBlame = url.searchParams.get('view') === 'blame'
 
-    var blameData: Observable<BlameHunkData> | undefined = undefined
+    // Create a BehaviorSubject so preloading does not create a subscriberless observable
+    const blameData = new BehaviorSubject<BlameHunkData>({ current: undefined, externalURLs: undefined })
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.ts:20]: @fkling mind taking a look at this commit in which I added preloading for blame data? In particular, my choice to use a `BehaviorSubject` here. When I was just using the normal observable, the page data was preloaded, but no network request was made for the blame data until I acutally clicked the button. I think that was because, when prelaoding, there is nothing subscribing to the observer. So using a `BehaviorSubject` means whenever we _do_ load the page, we can replay the last value on the stream. Does that logic all make sense?

Comment (fkling) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.ts:20]: Yes, that's essentially what we do for the search stream too.

fkling (approved):
Looks all good to me!

Issue #61622, opened 2024-04-05
Svelte: Implement revision selector


taiyab:
We have this newly designed revision selector that needs implementing.

![Image](https://github.com/sourcegraph/sourcegraph/assets/5337876/2412ca4a-3316-461f-abcd-65c1d43d889d)

Prototype:
https://www.figma.com/proto/EPgl0qVbqGSumvf52brhda/Code-Search?page-id=1397%3A21372&type=design&node-id=2581-76243&viewport=120%2C-238%2C0.3&t=ByI0ebF2Mc6wJ8o8-9&scaling=min-zoom&starting-point-node-id=2581%3A76243&show-proto-sidebar=1

Designs:
- https://www.figma.com/file/EPgl0qVbqGSumvf52brhda/Code-Search?type=design&node-id=2581%3A76980&mode=design&t=7EDaEAnwAaYNLDHD-1
- https://www.figma.com/file/EPgl0qVbqGSumvf52brhda/Code-Search?type=design&node-id=2639%3A58372&mode=design&t=7EDaEAnwAaYNLDHD-1

PR #61817, opened 2024-04-11
[Svelte] Add refs picker to the repository page

vovakulikov:
Closes https://github.com/sourcegraph/sourcegraph/issues/61622

Note that designs are not final, this is just a version that I could come up with a bit faster since most of the UI components haven't been implemented by new specifications from Figma. (This should be good enough for the milestone 2 checkpoint but let me know if something important is missing)

| Modes | Image |
| ---- | ---- |
| Branch | Screenshot 2024-04-12 at 15 25 38 |
| Tag | Screenshot 2024-04-12 at 15 25 50 |
| Commit | Screenshot 2024-04-12 at 15 26 04 |

## Todo 
- [x] Add basic picker functionality 
- [x] Add combobox-like navigation to the picker suggestions
- [x] Support dark theme
- [x] Fix problems with melt-ui and our own popover functionality overlap 
- [ ] Add keyboard shortcuts for switching between tabs
- [ ] Add highlight words matching logic (most likely will ship it as separate PR) 

## Test plan
- Go to the blob or repository page
- Check branch, tag, and commit pickers


client/shared/src/util/url.ts
```diff
@@ -430,3 +430,92 @@ export function parseRepoRevision(repoRevision: string): ParsedRepoRevision {
...
     }
 }
+
+/**
+ * Replaces the revision in the given URL, or adds one if there is not already
+ * one.
+ *
+ * @param href The URL whose revision should be replaced.
+ */
+export function replaceRevisionInURL(href: string, newRevision: string): string {
```
vovakulikov, line 440: Just moved this from web utils since I have to use the same logic for revision pickers in svelte directory

client/web-sveltekit/src/lib/TabPanel.svelte
```diff
@@ -20,7 +20,13 @@
 </script>
 
 {#if selected}
-    <div id="{context.id}-panel-{id}" aria-labelledby={tabId} role="tabpanel" tabindex={selected ? 0 : -1}>
+    <div
+        id="{context.id}-panel-{id}"
+        aria-labelledby={tabId}
+        role="tabpanel"
+        tabindex={selected ? 0 : -1}
+        data-tab-panel
```
vovakulikov, line 28: Expose this div tab panel element so we can style it from the consumer level.

client/web-sveltekit/src/lib/Tabs.svelte
```diff
@@ -96,7 +96,8 @@
         letter-spacing: normal;
         margin: 0;
         min-height: 2rem;
-        padding: 0 0.25rem;
+        padding: 0.25rem 0.5rem;
+        border-radius: 0.125rem;
```
vovakulikov, line 100: Increase click area on tabs button to simplify mouse aiming

client/web-sveltekit/src/lib/wildcard/Badge.module.scss
```diff
@@ -100,14 +101,23 @@
     --badge-border: 1px solid var(--secondary);
 }
 
+.link {
```
vovakulikov, line 104: Adding new badge mode based on picker designs

client/web-sveltekit/src/lib/wildcard/Badge.svelte
```diff
@@ -39,7 +40,7 @@
 
 <!--TODO: support non-branded badges -->
 <slot name="custom" class={cls}>
-    <span class={cls}>
+    <span class={cls} data-badge>
```
vovakulikov, line 43: Expose this for consumer styling, had to do it because I have to set text overflow rules which won't work if I set them on consumer wrapper of the badge component

client/web-sveltekit/src/lib/wildcard/Input.svelte
```diff
@@ -0,0 +1,88 @@
+<script lang="ts">
```
vovakulikov, line 1: Input standalone component, we currently have a few inputs UI already and my first intention was just to share input styles and not have a separate component for it, but Input also contains some logic about autofocus and additional loading UI which is needed for picker UI, so I created this. I haven't migrated other inputs to this component but will do in a follow-up PRs.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryBranchesPicker.svelte
```diff
@@ -0,0 +1,239 @@
+<script context="module" lang="ts">
```
Comment (vovakulikov) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryBranchesPicker.svelte:1]: Most of the three pickers have the same logic about fetching, errors, and suggestion rendering. 
I wanted to come up with some generic API to build your own picker/combobox faster next time but
I found implementation too complex to have it in this PR. So far these three pickers are different components
but later I think I would be able to come up with compound components API like we have in react world but in svelte

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryBranchesPicker.svelte
```diff
@@ -0,0 +1,239 @@
...
+    export let onSelect: (branch: RepositoryBranch) => void
+
+    const {
+        elements: { menu, input, option },
+        states: { inputValue },
+    } = createCombobox<any>({
+        portal: null,
+        forceVisible: true,
+        scrollAlignment: 'nearest',
+        closeOnOutsideClick: false,
```
Comment (vovakulikov) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryBranchesPicker.svelte:35]: Melt UI combobox uses their own floating panel instance which we can't disable so I try to turn off as much as possible features like close on outside click to be able to allow our own Popover UI handle this. 
Otherwise outside click won't work in a first run since I suspect that melt ui stops propagation of events if they catch outside click event.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryBranchesPicker.svelte
```diff
@@ -0,0 +1,239 @@
...
+
+    <footer class="footer">
+        <a href={`${repoURL}/-/branches`}>
+            See all commits
+            {#if !$repositoryBranches.error && $repositoryBranches.value}({$repositoryBranches.value.totalCount}){/if}
+        </a>
+    </footer>
+</div>
+
+<style lang="scss">
```
Comment (vovakulikov) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryBranchesPicker.svelte:130]: Styles mostly are the same but with some small differences because each picker suggestion item has slightly different styles and layout (commit, branch, tags), this will be unified as soon as I figure out generic API for pickers.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryRevPicker.gql
```diff
@@ -0,0 +1,70 @@
+# Used to query data for repository branches and tags
+# NOTE: the query which is going to use this fragment is supposed
+# to have query and type (branch or tag) variables
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryRevPicker.gql:3]: Maybe instead of depending on the wrapping query to define the variables, instead we create a `fragment PickerGitRef on GitRef`. Then, we move the `gitRefs(first: 15 ...) { nodes { ...PickerGitRef } }` into the query in `layout.gql` where the variables are defined. I expect `PickerGitRef` is probably an easier type to work with than the connection object anyways

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryBranchesPicker.svelte
Comment (vovakulikov) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryBranchesPicker.svelte]: Was thinking about this but didn't do it since I wasn't sure about generic types in svelte components. I did it now by your example of svelte composition, thanks a lot for the good examples.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryRevPicker.gql
Comment (vovakulikov) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryRevPicker.gql:3]: True, I'm worried that in this case totalCount isn't forced by the query since it lives outside of the nodes field (and actually lives on the top level of the commit connection).

vovakulikov:
Thanks for the review @fkling @camdencheek @taiyab 

I think I covered most of the big comments about compositions and design review that we had this morning with @taiyab. Whenever you have a minute, take a look. It's ready for the second round. Thanks.

taiyab:
As an aside: We should not forget the popups on hover and the likes in the next round of updates on this.

vovakulikov:
@taiyab I got captured leftovers in the follow-up issue (including sub tooltip with commit entry)

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryRevPickerItem.svelte
```diff
@@ -0,0 +1,86 @@
...
+            overflow: hidden;
+            text-overflow: ellipsis;
+        }
+    }
+
+    .timestamp {
+        padding-right: 0.75rem;
+        color: var(--text-muted);
+    }
+</style>
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryRevPickerItem.svelte:86]: It seems this can be simplified a bit so that you don't need to use global overwrites (and/or add `data-` attributes to other components):

```suggestion
<span class="title">
    <slot name="title">
        <span class="no-shrink"><Icon svgPath={iconPath} inline /></span>
        <Badge variant="link">{label}</Badge>
    </slot>
</span>
<span class="author">
    {#if author}
        <span class="no-shrink"><Avatar avatar={author.person} /></span>
        <span class="text">{author.person.displayName}</span>
    {/if}
</span>
<span class="timestamp text">
    {#if author}
        <Timestamp date={author.date} strict />
    {/if}
</span>

<style lang="scss">
    .title,
    .author {
        min-width: 0;
        --avatar-size: 1.5rem;

        display: flex;
        gap: 0.5rem;
    }

    .no-shrink {
        flex-shrink: 0;
    }

    .text {
        white-space: nowrap;
        overflow: hidden;
        text-overflow: ellipsis;
    }

    .title {
        padding-left: 1rem;
        color: var(--icon-muted);

        // Icon
        :global([data-highlighted]) &,
        :global([data-picker-suggestions-list-item]):hover {
            color: var(--icon-color);
        }

        // Badge
        :global(span) {
            overflow: hidden;
            text-overflow: ellipsis;
        }
    }

    .timestamp {
        padding-right: 0.75rem;
        color: var(--text-muted);
    }
</style>
```

Note that I wasn't able to get the ellipsis working by wrapping the badge (i.e. `<span class="text"><Badge .../></span>`). It appears that because badges have `display: inline-block` that somehow prevents `text-overflow`, etc set on a parent from having any affect (it works if it's directly set on the element which is what `:global(span)` does.

Given that "odd" behavior I would suggest that `Badge` should have a prop to control overflow behavior.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryRevPickerItem.svelte
```diff
@@ -0,0 +1,86 @@
...
+    .title {
+        padding-left: 1rem;
+
+        // Branch icon
+        :global(svg) {
+            flex-shrink: 0;
+            color: var(--icon-muted);
+
+            :global([data-highlighted]) &,
+            :global([data-picker-suggestions-list-item]):hover {
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryRevPickerItem.svelte:70]: Is this missing a `&`?

```suggestion
            :global([data-picker-suggestions-list-item]):hover & {
```

fkling (approved):
Left a couple of suggestions inside, looks great overall!

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryRevPickerItem.svelte
Comment (fkling) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryRevPickerItem.svelte:86]: Alternatively, maybe `Badge` should inherit `text-overflow and `overflow` properties:

```css
text-overflow: inherit
overflow: inherit
```

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/layout.gql
```diff
@@ -23,6 +23,18 @@ query LastCommitQuery($repoName: String!, $revspec: String!, $filePath: String!)
     }
 }
 
+query RepositoryGitRefs($repo: ID!, $query: String, $type: GitRefType!) {
+    node(id: $repo) {
```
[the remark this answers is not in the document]
Comment (fkling) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/layout.gql:27]: I've been using `respository` consistently throughout the repo pages. Renamed repository shouldn't be an issue since you are using the repo ID that, essentially, comes from `repositoryRedirect(name: $repoName)`, i.e. we already used the name to get the repo ID.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.ts
```diff
@@ -83,5 +84,61 @@ export const load: LayoutLoad = ({ parent, params }) => {
...
+                    .then(
+                        mapOrThrow(({ data, error }) => {
+                            if (data?.node?.__typename !== 'Repository' || !data?.node?.gitRefs) {
+                                throw new Error(error?.message)
+                            }
+
+                            return data.node.gitRefs
+                        })
+                    )
+            ),
```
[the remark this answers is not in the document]
Comment (vovakulikov) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.ts:106]: I was worried about repo name and ID arguments but just noticed that we already rely on repoName in other queries like last commit, so I assume it should be safe, will simplify by your example here, thanks for suggestions

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryRevPickerItem.svelte
Comment (vovakulikov) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryRevPickerItem.svelte:86]: There are a few suggestions, so I will comment on them one by one

- `timestamp text` or `<class1>, <class2>`, there is a term mixin in BEM notation, but this isn't the case for them. I would like to avoid class mixing if it's in one file because this requires you to unwrap them manually rather than just having a plain class in which you will see all "direct" styles 

- Yeah, `text-overflow` doesn't work on block-like elements (even if they're inline-block, or flex-child blocks) 

- `Alternatively, maybe Badge should inherit text-overflow and overflow` properties:` This might be a problem when parents have it and we don't want to have it for badge directly; Imagine we have badge + text and we want that text shrinks with ... if it doesn't fit, with text-overflow on the parent it will affect badge as well. So it would be like `<badge-te...> <tex...>` 

- `I would suggest that Badge should have a prop to control overflow behaviour.` Seems reasonable but a bit concerned that we invent here a bit of CSS responsibility, don't have good arguments here so don't want to fight over this prop

- About the last utility `.text` class, this is close to my first point. I would like to avoid having any mixed utility classes, even per one file, because it requires you to sum up all styles in your head to figure out how a block would look. `> :global(*)` This would set text-overflow to all children on the first level which causes a problem with multiple truncated text blocks like `<badge-te...> <tex...>` 

The primary reason I used global styles here is that we override the commit item layout through the slot but still want to preserve all styles from the `RepositoryRevPickerItem` file. But I agree that having just :global(span) could be too much relying on sharable components layout structure. 

So I added more specific `data-attributes` to root elements of sharable components so it would be easy for us to understand how we style them in consumers. But this only makes sense on simple (one-element) components like icon or badged, all other elements shouldn't have

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryRevPickerItem.svelte
Comment (vovakulikov) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryRevPickerItem.svelte:70]: Yes, my bad, I guess it worked because melt-ui observes hover state itself, thanks for catching anyway.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryRevPickerItem.svelte
Comment (fkling) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryRevPickerItem.svelte]: I don't want to start a big discussion in a GitHub thread, this is probably discussed better in person. Just to address some concerns:

> I would like to avoid having any mixed utility classes, even per one file, because it requires you to sum up all styles in your head to figure out how a block would look.

Is the concern about utility classes or about using multiple classes on the same element? In my suggestion there is only one element with two classes, `timestamp text`. You could change it to 

```
<span class="timestamp text">
    <span class="text">
        {#if author}
          <Timestamp date={author.date} strict />
        {/if}
    </span>
</span>
```

if mixing multiple classes is your main concern. But if that means elements should only have a single class then I think we are unnecessarily limiting ourselves.

If the concern is about utility classes in general... I don't know, I think a class with a useful name is ... useful. Having these just in a single file also limits the scope somewhat. For example, once you know what `.text` (which should maybe be called `.overflow-ellipsis` instead) does, you know how any element that has this class is styled.
On the other hand, if every element needs to have its dedicated class with its dedicated style, I have to look at the style definition for each element to understand how it's styled.

Ultimately it seems we have two ways of writing styles:

```scss
// combined selectors
.foo, .bar, .baz {
    // properties shared by all of these
}

.foo {
    // properties specific to foo
}

// utility class
.utility {
    // properties that should apply to multiple elements
}

.foo {
  // properties specific to foo
}
```

In both cases I need to look at multiple definitions to get understanding oft the final style of an element. The advantage of the second one, IMO, is that I can use my existing knowledge about `.utility` when I see it on another element.

> `:global(*)` This would set text-overflow to all children on the first level which causes a problem with multiple truncated text blocks like `<badge-te...> <tex...>`

In this specific case you control the children of `.text` so you could ensure that it only has one child.

---

The main reason my suggestion was to avoid having to reach into other components. As soon as we do that we are risking issues with specificity again, and we are making assumptions about the internal structure for a component. That's not a problem for components that expect to be styled, but might be for something like `<Timestamp />`.

As case in point you added this comment:

```
        // Timestamp uses tooltip wrapper element with display:contents
        // override this behavior since we have to overflow text within
        // trigger text
```

Maybe we implement tooltips differently at some point in which case this might not work anymore. And especially in this case there is no need to reach into `<Timestamp />` at all. Setting the text truncation stuff on the parent element will already achieve the desired effect.

Comment (vovakulikov) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryRevPickerItem.svelte]: I have a follow-up issue for this UI, let's discuss this in person and I will address ongoing concerns there, anyway like to have a convo about styles though)

PR #61825, opened 2024-04-12
chore: update readme for svelte dev

bahrmichael:
Updating the Readme based on some things I had to figure out.

## Test plan

Not applicable, because it's only changes to the readme

### DIFF
--- a/client/web-sveltekit/README.md
+++ b/client/web-sveltekit/README.md
@@ -25,7 +25,7 @@ https://sourcegraph.com by default (can be overwritten via the
 If you're a Sourcegraph employee you should run this command to use the right auth instance:
 
 ```bash
-SOURCEGRAPH_API_URL=https://sourcegraph.sourcegraph.com pnpm run dev:dotcom
+SOURCEGRAPH_API_URL=https://sourcegraph.sourcegraph.com pnpm run dev
 ```
 
 ### Using code from `@sourcegraph/*`
--- a/client/web-sveltekit/README.md
+++ b/client/web-sveltekit/README.md
@@ -72,11 +72,6 @@ because it also validates imported modules from other packages, and we are not
 explicitly marking type-only imports with `type` in other parts of the code
 base (which is required by this package).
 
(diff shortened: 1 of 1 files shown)


client/web-sveltekit/README.md
```diff
@@ -22,6 +22,12 @@ signin/signout are proxied to an actual Sourcegraph instance,
 https://sourcegraph.com by default (can be overwritten via the
 `SOURCEGRAPH_API_URL` environment variable.
 
+If you're a Sourcegraph employee you should run this command to use the right auth instance:
+
+```bash
+SOURCEGRAPH_API_URL=https://sourcegraph.sourcegraph.com pnpm run dev:dotcom
```
fkling, line 28: We are not really making use of vite's environment variables, so this should suffice

```suggestion
SOURCEGRAPH_API_URL=https://sourcegraph.sourcegraph.com pnpm run dev
```

It wouldn't be correct anyways because you wouldn't want to connect to a customer instance with dotcom semantics.

client/web-sveltekit/README.md
```diff
@@ -66,6 +72,11 @@ because it also validates imported modules from other packages, and we are not
 explicitly marking type-only imports with `type` in other parts of the code
 base (which is required by this package).
 
+### Regenerate GraphQL TS files
+
+When you edited the `*.gql` files, you will need to compile the changes into the `*.gql.ts` files. You can generate
+them by running `pnpm run build`.
+
```
fkling, line 79: This could be a bit confusing and might depend on how one is working with the prototype. E.g. if you run `pnpm run dev` then the TS files will be automatically regenerated.
I don't know in which situations you'd need the TS files be updated but are not running a dev build or production build already.

In which situation did you need this?

bahrmichael: Hm, I thought that it didn't update when running `pnpm run dev` but when I checked now it worked as expected. I'll cut this part.

fkling (approved):

Issue #61822, opened 2024-04-12
Add more useful spans to traces for uploading/ingesting SCIP indexes


varungandhi-src:
Right now, we get spans like the following:

<img src="https://uploads.linear.app/15b96a4d-ffc0-481a-ba9c-9d840020f7bc/a8e8c5b8-bcca-4358-afe7-6f49fb79ec86/2b1f11aa-1109-4b0c-bab7-5a5391e1b732?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzE1Yjk2YTRkLWZmYzAtNDgxYS1iYTljLTlkODQwMDIwZjdiYy9hOGU4YzViOC1iY2NhLTQzNTgtYWZlNy02ZjQ5ZmI3OWVjODYvMmIxZjExYWEtMTEwOS00YjBjLWJhYjctNWE1MzkxZTFiNzMyIiwiaWF0IjoxNzEyODk4NjYzLCJleHAiOjMzMjgzNDU4NjYzfQ.ShsRcsGGvtHKrCjtZHM7o0Jpwj_IgEgf9h3H4tUlhv8 " alt="CleanShot 2024-04-12 at 13.09.18@2x.png" width="1080" height="676" />

After the insert-metadata step, it's unclear how much time is spent inserting documents vs package references etc. We should record some intermediate spans for those.

PR #61826, opened 2024-04-12
codeintel: Add fine-grained spans for uploads

varungandhi-src:
Fixes https://github.com/sourcegraph/sourcegraph/issues/61822
Fixes ENG-23610

Mostly, this PR splits up the code involved in writing data to various tables into
smaller functions, each of which has its own span for traces. The patch respects
the changes made by Christoph for syntactic indexes.

With this patch, we can see more fine-grained information in Jaeger's statistics view
when traces are enabled.

CleanShot 2024-04-12 at 19 22 37@2x

## Test plan

Manually performed some uploads. Covered by existing tests.

### DIFF
--- a/internal/codeintel/uploads/internal/background/processor/scip.go
+++ b/internal/codeintel/uploads/internal/background/processor/scip.go
@@ -368,7 +368,7 @@ func writeSCIPDocuments(
 		var scipWriter lsifstore.SCIPWriter
 
 		if upload.Indexer == shared.SyntacticIndexer {
-			scipWriter, err = tx.NewSyntacticSCIPWriter(ctx, upload.ID)
+			scipWriter, err = tx.NewSyntacticSCIPWriter(upload.ID)
 		} else {
 			scipWriter, err = tx.NewPreciseSCIPWriter(ctx, upload.ID)
 		}
--- a/internal/codeintel/uploads/internal/lsifstore/insert.go
+++ b/internal/codeintel/uploads/internal/lsifstore/insert.go
@@ -5,20 +5,19 @@ import (
 	"context"
 	"crypto/sha256"
 	"encoding/hex"
-	"sort"
-	"sync/atomic"
-
 	"github.com/keegancsmith/sqlf"
 	"github.com/lib/pq"
 	"github.com/sourcegraph/log"
 	"go.opentelemetry.io/otel/attribute"
 	"google.golang.org/protobuf/proto"
+	"sort"
 
 	"github.com/sourcegraph/scip/bindings/go/scip"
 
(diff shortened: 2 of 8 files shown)


internal/codeintel/uploads/internal/lsifstore/insert.go
```diff
@@ -403,11 +488,10 @@ func (s *scipWriter) writeSymbols(ctx context.Context, documents []bufferedDocum
 			); err != nil {
 				return err
 			}
-
-			atomic.AddUint32(&s.insertedSymbolsCount, 1)
+			*insertCount = *insertCount + 1
+			s.insertedSymbolsCount = s.insertedSymbolsCount + 1
```
varungandhi-src, line 492: It looks like this atomic operation was added accidentally in the original PR which introduced it. https://github.com/sourcegraph/sourcegraph/commit/4e3eebd8dc580b42f39515642b8c63d379d79d38

In that code, this loop was present in a closure, and that closure was passed to a single-threaded batch inserter. I'm guessing during development, it was passed to a concurrent inserter (and hence the atomic operation was used inside the loop). https://sourcegraph.com/github.com/sourcegraph/sourcegraph@4e3eebd8dc580b42f39515642b8c63d379d79d38/-/blob/enterprise/internal/codeintel/uploads/internal/lsifstore/data_write.go?L296-298

No other state on the `scipWriter` struct is protected by any synchronization primitives, and even this field is not read atomically in other places. So the atomic operation seems to purely be a mistake.

internal/codeintel/uploads/internal/background/processor/scip.go
```diff
@@ -368,7 +368,7 @@ func writeSCIPDocuments(
 		var scipWriter lsifstore.SCIPWriter
 
 		if upload.Indexer == shared.SyntacticIndexer {
-			scipWriter, err = tx.NewSyntacticSCIPWriter(ctx, upload.ID)
+			scipWriter, err = tx.NewSyntacticSCIPWriter(upload.ID)
```
keynmol, line 371: Just started reviewing. To confirm - you are removing context here because operations themselves take context parameters?

Is this different from Precise writer, which has context still?

varungandhi-src: > To confirm - you are removing context here because operations themselves take context parameters?

Yes, Context should be passed to method individually. It was an unused parameter for this function, so I removed it.

Precise writer constructs some temporary tables during initialization, and database functions require a Context, that's why it's needed there.

keynmol: Thanks, makes sense

keynmol (approved):

mmanela (approved):

PR #61832, opened 2024-04-12
graphql: expose indexed search endpoints

stefanhengl:
This is part of the effort to enable admins to drain a Zoekt instance, which will help to downscale Zoekt clusters without downtime.

The idea is that admins can query our GraphQL API to get a list of available endpoints. In another PR, I will allow admins to list the endpoints they want to drain.

The API is modelled after the [`gitservers`](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@c66a73ade256777122c65667d2723d25e0f3368a/-/blob/cmd/frontend/graphqlbackend/schema.graphql?L2148) endpoint, which returns similar data. Like `gitservers` we might want to return more data in the future.

Test plan:
manual testing
Screenshot 2024-04-12 at 11 37 03


cmd/frontend/graphqlbackend/search.go
```diff
@@ -63,3 +66,36 @@ type searchResolver struct {
...
+type indexedSearchInstance struct {
+	address string
+}
+
+func (i *indexedSearchInstance) Address() string {
+	return i.address
+}
+
+func (i *indexedSearchInstance) ID() graphql.ID {
+	return marshalGitserverID(i.address)
```
mmanela, line 79: Does this make sense, it will give each an id prefixed with `GitserverInstance`

mmanela (approved):
One small comment

cmd/frontend/graphqlbackend/search.go
```diff
@@ -63,3 +66,36 @@ type searchResolver struct {
...
+}
+
+func (r *schemaResolver) IndexedSearchInstances(ctx context.Context) (graphqlutil.SliceConnectionResolver[*indexedSearchInstance], error) {
+	// 🚨 SECURITY: Site admins only.
+	if err := auth.CheckCurrentUserIsSiteAdmin(ctx, r.db); err != nil {
+		return nil, err
+	}
+
+	indexers := search.Indexers()
+	eps, err := indexers.Map.Endpoints()
```
keegancsmith, line 89: I can't remember, but does this have a stable ordering? IE should we sort it first?

cmd/frontend/graphqlbackend/search.go
keegancsmith: yeah we need to change the prefix here.

keegancsmith (approved):

jtibshirani (approved):
Looks good to me!

PR #61833, opened 2024-04-12
search: support draining Zoekt instances

stefanhengl:
This change lets us set an ENV (`INDEXED_SEARCH_DRAIN_SERVERS`) 
on frontend with the Zoekt endpoints we want to drain.
As a consequence all repos indexed by those instances will be
assigned to other instances. Search still targets all instances. After
an instance has been fully drained, we can remove it without user
impact.

Here is how the process would look like:

1. Use #61832 to get a list of endpoints
2. Decide which endpoints to drain. Then set ENV on frontend and restart.
3. Monitor Zoekt dashboard to see when all indexes have been migrated. Depending on the number and size of repos, this can take minutes,  hours or even days.
4. Remove empty instance
5. Unset ENV

Test plan:
- New unit test
- manual testing

As soon as the ENV is set, indexes are assigned to the "green" instance. Once the repos have been indexed on "green", they are dropped by "yellow". As one can see in the graph, there is a short period during which repos might appear on both instances. We already have a [deduper](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@405150fba2710a58e88120f412503366e0af8706/-/blob/internal/search/backend/horizontal.go?L60-62) in place which takes care of that.

Screenshot 2024-04-12 at 14 38 34


internal/search/env.go
```diff
@@ -91,11 +93,27 @@ func ListAllIndexed(ctx context.Context, zs zoekt.Searcher) (*ZoektAllIndexed, e
...
 func Indexers() *backend.Indexers {
 	indexersOnce.Do(func() {
+		toDrainEnv := strings.Split(os.Getenv("INDEXED_SEARCH_DRAIN_SERVERS"), ",")
+		toDrain := make(map[string]struct{}, len(toDrainEnv))
+		for _, d := range toDrainEnv {
+			toDrain[d] = struct{}{}
+		}
+
 		indexers = &backend.Indexers{
 			Map: endpoint.ConfBased(func(conns conftypes.ServiceConnections) []string {
```
keegancsmith, line 103: This coupling feels a bit strange to me. You teach Indexers about the drained nodes, but then over here you need to ensure Map does not include them. Additionally there is a call site on Indexers.Map which would need to resolve the full set (see `resolveIndexserver`).

I think if you want to stick to this approach, you need to rename this from Map to be something clearer. Alternatively there is a simpler approach which doesn't require modification to backend.Indexers. Instead implement a `backend.EndpointMap` which supports draining. IE it can wrap an endpoint.Map such that:
- It returns all endpoints in `Endpoints()`
- It excludes drained hosts from the consistent map (`Get()`).

internal/search/env.go
```diff
@@ -91,11 +93,27 @@ func ListAllIndexed(ctx context.Context, zs zoekt.Searcher) (*ZoektAllIndexed, e
...
-				return conns.Zoekts
+				// Remove endpoints we want to drain from the list of endpoints. As a
+				// consequence no repos will be assigned to the removed endpoints.
+				var connsZoekts []string
+				for _, z := range conns.Zoekts {
+					if _, ok := toDrain[z]; ok {
```
keegancsmith, line 108: is this sort of check robust enough? Inside of the backend package we use a more lenient check (see `findEndpoint` function).

internal/search/backend/indexers.go
```diff
@@ -87,6 +99,36 @@ func (c *Indexers) ReposSubset(ctx context.Context, hostname string, indexed zoe
 	return subset, nil
 }
 
+// filterReposPendingIndexing returns the subset of repos that are not yet
+// indexed by another host.
+func (c *Indexers) filterReposPendingIndexing(ctx context.Context, indexed zoekt.ReposMap, repos []types.MinimalRepo) ([]types.MinimalRepo, error) {
```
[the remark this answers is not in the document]
keegancsmith: As it is written now, Map is a consistent hash which maps repo names to host it needs to live on. If your hostname appears in the drain list, it is removed from that Map, which then means the repo will be mapped to another hostname.

The algorithm in ReposSubset supports rebalancing. It does this checking all repos on an instance it should no longer have. It will only remove them from the instance if the destination host has indexed the repo. This makes us smoothly handle growing the cluster. Using this same idea we can now slowly drain an endpoint.

jtibshirani:
> As one can see in the graph, there is a short period during which repos might appear on both instances.

Could you explain this part for my understanding? This is just a short interim period where we're waiting to drop the old index, and should only be a few minutes max?

stefanhengl:
> Could you explain this part for my understanding? This is just a short interim period where we're waiting to drop the old index, and should only be a few minutes max?

Yes, exactly. While the new host doesn't report a repo as indexed, the old host will hold on to it. The timeline is like this

instance 0 is empty
instance 1 hosts index "X"
we want to drain instance 1

1. instance 0 asks frontend what to index and reports what it has indexed
2. frontend responds with "X"
3. instance 0 starts indexing "X"
4. instance 0 indexed "X"
5. instance 1 asks frontend what to index
6. frontend responds to instance 1 with an empty list 
7. instance 1 drops "X" 

The time between step 4 and 7 is the time we have "X" on both instances. This should be in the order of minutes.

internal/search/env.go
stefanhengl: It should be robust, because we expect toDrain to contain the endpoints and not the hostnames.

internal/search/env.go
stefanhengl: > Additionally there is a call site on Indexers.Map which would need to resolve the full set (see resolveIndexserver).

I think that one is acceptable. It is used for the admin UI only and will eventually reflect the truth.

stefanhengl:
> Another minor thing. Should this be an env or a site config. For example if env we need to make sure we set it for everyone that does search indexing decisions which is frontend and frontend-internal. If site config it may be easier to communicate.

The way it is designed right now we need a restart either way, so it feels better to use an ENV for that. AFAIK we are encouraged not to add things to site config that require a restart. Frontend-internal is a good point, but we don't seem to deploy that on Cloud. Is that specific to .com?

stefanhengl:
@keegancsmith PTAL. I have tried other approaches this week but I believe this change is still a good compromise. It introduces some coupling between `indexer` and `env` which other approaches don't require, but, on the other hand, it is very explicit and requires minimal changes. I would say we go ahead an try this out.

internal/search/env.go
```diff
@@ -91,11 +93,27 @@ func ListAllIndexed(ctx context.Context, zs zoekt.Searcher) (*ZoektAllIndexed, e
 
 func Indexers() *backend.Indexers {
 	indexersOnce.Do(func() {
+		toDrainEnv := strings.Split(os.Getenv("INDEXED_SEARCH_DRAIN_SERVERS"), ",")
```
keegancsmith, line 96: FYI if this variable is empty it will return a slice of length 1 with the empty string

internal/search/env.go
```diff
@@ -91,11 +93,27 @@ func ListAllIndexed(ctx context.Context, zs zoekt.Searcher) (*ZoektAllIndexed, e
...
+		toDrain := make(map[string]struct{}, len(toDrainEnv))
+		for _, d := range toDrainEnv {
+			toDrain[d] = struct{}{}
+		}
```
keegancsmith, line 100: might simplify, but we have collection.NewSet these days.

PR #61834, opened 2024-04-12
appliance/rec: add basic reconcile loop

jdpleiness:
Add basic reconciler setup for the appliance.

## Test plan

Tested locally for now, more testing coming in following PRs.


internal/appliance/reconcile.go
```diff
@@ -0,0 +1,107 @@
...
+			ResourceVersion: &applianceSpec.ResourceVersion,
+		})
+
+		if err != nil && client.IgnoreNotFound(err) != nil {
+			reqLog.Error(err, "failed to delete sourcegraph appliance spec")
+			return ctrl.Result{}, err
+		}
+
+		return ctrl.Result{}, nil
+	}
```
craigfurman, line 62: Why do we need this block? By default, kubernetes will delete any owned resources when the owner (the configmap) is deleted, so we might not need to do anything in reaction to deletion events 🤔

jdpleiness: Good call. I think I was unsure initially about the ownership tree while spec'ing this out. As it is now, I don't think we need it 👍

PR #61843, opened 2024-04-12
search: refactor anonymous interfaces

stefanhengl:
This refactors the anonymous interfaces defined in `searchIndexerServer` to be named types.

Anonymous interfaces don't support code navigation, at least not in Goland, and it was very tedious to use CMD+F the whole time.

Test plan:
pure refactor => CI


jtibshirani (approved):
I really like this sort of refactor.

Aside: I also find it hard to "go to definition" when we pass functions around, which we do a lot in the symbols code.

stefanhengl:
I had to force push because the base branch changed.

PR #61846, opened 2024-04-12
Remove wolfi base image rebuilds on main

willdollman:
In https://github.com/sourcegraph/sourcegraph/pull/61845 we removed base images being build in CI - but only on PR pipelines. This also removes base image builds from the `main` pipeline as well.

Now that we're using bazel to build base images on the fly, there's no longer a need to build these base images in CI.

## Test plan

- CI

### DIFF
--- a/dev/ci/internal/ci/pipeline.go
+++ b/dev/ci/internal/ci/pipeline.go
@@ -145,7 +145,7 @@ func GeneratePipeline(c Config) (*bk.Pipeline, error) {
 		ops.Merge(securityOps)
 
 		// Wolfi package and apko lock check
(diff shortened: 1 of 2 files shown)


dev/ci/internal/ci/pipeline.go
```diff
@@ -324,16 +324,13 @@ func GeneratePipeline(c Config) (*bk.Pipeline, error) {
 		))
 
 		// Wolfi package and base images
-		packageOps, baseImageOps, apkoOps := addWolfiOps(c)
+		packageOps, _, apkoOps := addWolfiOps(c)
```
Strum355, line 327:
```suggestion
		packageOps, apkoOps := addWolfiOps(c)
```

PR #60106, opened 2024-02-02
Cody context: handle empty queries

jtibshirani:
If a cody context query contains only stopwords, then query processing returns
a list of empty terms, causing a panic. Now, we just return an empty query
which matches no files.

Fixes #60090

## Test plan

Added new unit test

PR #61848, opened 2024-04-12
Context: return no results for stopwords query

jtibshirani:
In #60106, we fixed a bug where the context search threw an error when the query
was composed entirely of stopwords. However, the queries were still returning
results (based on filename match). This means that for a stopwords-only query
like "what's going on?" we return a fairly random set of files, which confuses
the LLM.

Now we make sure to return an empty search, which returns no results.

## Test plan

Added a new unit test. Tested Cody web locally.


jtibshirani:
Another bug I found playing around with `patterntype:codycontext` in the search bar. Here's what results used to look like:
Screenshot 2024-04-12 at 1 37 22 PM

mmanela (approved):

PR #61849, opened 2024-04-12
Context: push limiting and merging into search job

jtibshirani:
To test and debug Cody context, it's really useful to be able to see the search
results directly. This PR makes it possible to test context in the main
Sourcegraph search bar by adding `patterntype:codycontext`.

We already have this patterntype, so the main work here is to push important
logic from the `getCodyContext` GraphQL call down into the job logic.
Specifically, it moves the logic to search code and text separately, limit the
number of files returned, and combine the results. It also moves the "Cody
ignore" filtering logic.

## Test plan

Adapted existing unit tests and added new test `job_test.go` focused on job
logic.

Manual testing:
* Reran benchmarks against golden queries and checked results didn't change.
* Tested Cody Web and search bar locally


internal/search/codycontext/job.go
```diff
@@ -2,48 +2,164 @@ package codycontext
...
+	wg.Go(func() response {
+		return j.doSearch(ctx, clients, j.textJob, j.textCount)
+	})
+	wg.Go(func() response {
+		return j.doSearch(ctx, clients, j.codeJob, j.codeCount)
+	})
+	responses := wg.Wait()
+
+	for _, r := range responses {
+		stream.Send(streaming.SearchEvent{
```
jtibshirani, line 110: I'm not sending any `Stats` here. I think this is fine for now because the `getCodyContext` GraphQL API doesn't care about stats, and we don't have any direct users of `patterntype:codycontext` (it's just exposed for testing).

internal/search/query/fields.go
```diff
@@ -33,6 +33,10 @@ const (
 	FieldTimeout   = "timeout"
 	FieldCombyRule = "rule"
 	FieldSelect    = "select"
+
+	// Experimental fields for Cody, internal use only and not exposed to users
```
jtibshirani, line 37: It's not ideal that we have these and need to expose them. In the future, I'd love if we could automatically diversity across filetypes and not rely on hardcoded limits. Then we could remove 'code count' and 'text count' from this API (and the GraphQL one).

internal/codycontext/context.go
```diff
@@ -267,84 +257,59 @@ func (c *CodyContextClient) getKeywordContext(ctx context.Context, args GetConte
...
-				if fm, ok := res.(*result.FileMatch); ok {
-					collected = append(collected, filter(fileMatchToContextMatches(fm))...)
-					if len(collected) >= limit {
-						cancel()
-						return
-					}
-				}
+		for _, res := range e.Results {
+			if fm, ok := res.(*result.FileMatch); ok {
+				collected = append(collected, filter(fileMatchToContextMatches(fm))...)
```
jtibshirani, line 295: Note: we still perform the snippet selection here in getCodyContext API. This is trickier to push down into the job. And we're still working on finalizing the best strategy (should we really just be selecting the top chunk from each file?) So it's nice to be able to explore all the matching snippets in the search results.

cmd/frontend/internal/codycontext/context.go
```diff
@@ -271,84 +261,59 @@ func (c *CodyContextClient) getKeywordContext(ctx context.Context, args GetConte
...
-				if fm, ok := res.(*result.FileMatch); ok {
-					collected = append(collected, filter(fileMatchToContextMatches(fm))...)
-					if len(collected) >= limit {
-						cancel()
-						return
-					}
-				}
+		for _, res := range e.Results {
+			if fm, ok := res.(*result.FileMatch); ok {
+				collected = append(collected, filter(fileMatchToContextMatches(fm))...)
```
jtibshirani, line 299: Oops, I just realized I changed the Cody ignore filter behavior. Now, it's possible to filter away all returned results, so nothing is returned. I need to push the filter down into the job logic too. Or revise the strategy somehow.

internal/search/codycontext/job.go
stefanhengl: Just curious, couldn't the stats be helpful for debugging? Could they help to figure out that we didn't find a match because we skipped shards or files?

internal/search/codycontext/job.go
```diff
@@ -26,27 +46,123 @@ func NewSearchJob(plan query.Plan, newJob func(query.Basic) (job.Job, error)) (j
...
+}()
+
 type searchJob struct {
-	child    job.Job
+	codeJob   job.Job
+	codeCount int
+
+	textJob   job.Job
+	textCount int
+
```
stefanhengl, line 95: Did you consider using NewParallelJob instead, handing it textJob and codeJob as children?

internal/search/query/fields.go
stefanhengl: Alternatively you could hide it in the query parameters and then use `inputs.Features` which is available in the job layer. We could use a custom string format or json like we do for `ZoektSearchOptionsOverride`. Thinking about it, check out the Zoekt tab in the "Developer Settings".

stefanhengl: image

cmd/frontend/internal/codycontext/context.go
keegancsmith: For future follow-ups, but I think technically we can convert the cody ignore filters into a part of the query. But yeah we need to push this down. Out of interest do we just search with default count limits?

keegancsmith (approved):

internal/search/codycontext/job.go
```diff
@@ -26,27 +46,123 @@ func NewSearchJob(plan query.Plan, newJob func(query.Basic) (job.Job, error)) (j
 		return nil, err
 	}
 
-	child, err := newJob(q.query)
+	params := q.query.Parameters
+	patterns := q.patterns
+
+	codeQuery := q.query.MapParameters(append(params, query.Parameter{Field: query.FieldFile, Value: textFileFilter, Negated: true}))
```
mmanela, line 52: Why the negated on the file field?

internal/search/codycontext/job.go
```diff
@@ -26,27 +46,123 @@ func NewSearchJob(plan query.Plan, newJob func(query.Basic) (job.Job, error)) (j
...
+	textCount = DefaultTextResultsCount
+
+	// We can ignore errors because the values were already validated during query parsing
+	parameters.VisitParameter(query.FieldCodyCodeCount, func(value string, negated bool, annotation query.Annotation) {
+		codeCount, _ = strconv.Atoi(value)
+	})
+	parameters.VisitParameter(query.FieldCodyTextCount, func(value string, negated bool, annotation query.Annotation) {
+		textCount, _ = strconv.Atoi(value)
+	})
+	return
```
mmanela, line 78: Shouldn't this be `return codeCount, textCount`?

internal/search/codycontext/job.go
```diff
@@ -26,27 +46,123 @@ func NewSearchJob(plan query.Plan, newJob func(query.Basic) (job.Job, error)) (j
...
+	var (
+		mu        sync.Mutex
+		collected result.Matches
+	)
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	stream := streaming.StreamFunc(func(e streaming.SearchEvent) {
+		mu.Lock()
```
mmanela, line 142: Question (non-blocking): Why do we use locks instead of leveraging channels for synchronization?  (More of a learning go lang question)

internal/search/query/fields.go
jtibshirani: Thanks for the suggestion, this indeed turned out nicer. Now you can't modify these parameters from the search bar, but that seems fine.

internal/search/codycontext/job.go
jtibshirani: Indeed they could be! Let's consider this for a follow-up as this PR is already substantial.

internal/search/codycontext/job.go
jtibshirani: I'm using named return parameters, and in this case Go doesn't force you to return them explicitly. I like this syntax since it draws attention to the fact we're using named return params.

internal/search/codycontext/job.go
jtibshirani: I believe channels would work as well! I just found this approach simplest, since it's not important to optimize this collection code.

internal/search/codycontext/job.go
mmanela: oh nice, just learned a new feature :)

internal/search/codycontext/job.go
jtibshirani: I tried this out, and while it helped cut down on testing, it has the downside of not being able to control how the final results are combined. So I'd like to keep the current implementation, but we can keep it in mind in the future.

cmd/frontend/internal/codycontext/context.go
jtibshirani: Note: I tried to address all other comments but still need to address this before merging!

jtibshirani:
Updates since last review:
* Pass code and text limits through `search.Features` to avoid exposing new search parameters
* Push Cody ignore logic down into job, and add related tests. This required a bit of refactoring and clean-up in the Cody GraphQL code. Nothing about the "Cody ignore" behavior should change, I just moved the logic.

I'm happy with how it turned out and plan to merge. If you have other comments/ questions, very happy to address those in a follow-up.

cmd/frontend/internal/codycontext/context.go
jtibshirani: This turned out okay. I'll also catch up with the Cody Strat team to make sure they're aware of the change. And to share the idea of modelling things as regular repo and file filters instead of having this custom "postfiltering" logic.

Issue #60726, opened 2024-02-23
Empty logs produced when honeycomb is disabled


eseliger:
![Image](https://github.com/sourcegraph/sourcegraph/assets/19534377/84dc7f96-e192-42ce-bd1a-bf7dc9241361)

The noop event doesn't collect any attributes.

PR #61854, opened 2024-04-12
honey: replace no-op event with new NonSendingEvent that stores fields in memory

ggilmore:
Fixes: https://github.com/sourcegraph/sourcegraph/issues/60726

---

When honeycomb wasn't enabled, we use a noop Event type everywhere. However, there are some places where we'd like to inspect or log the event regardless of whether or not honeycomb is enabled, like the following code in gitserver.

https://github.com/sourcegraph/sourcegraph/blob/ee9199762853c10e6a737f341ceb0808ceca20ea/cmd/gitserver/internal/git/gitcli/command.go#L280-L286

Our use of the noop type, resulted in unexpected behavior - such as the `ev.Fields` in the log line not getting populated (https://github.com/sourcegraph/sourcegraph/issues/60726). 

This PR fixes this issue by replacing the noop event type a new NonSendingEvent which still allows `Fields` to be aggregated in memory and later retrieved with `Fields()`. This ensures that use cases like the logging one above still work as expected.

## Test plan

Unit tests

Also saw the following logs when I ran sourcegraph with `SRC_DEVELOPMENT=false SRC_LOG_FORMAT=json sg start` (notice that `ev.Fields` is no-longer null).

```
[    gitserver-0] {"SeverityText":"WARN","Timestamp":1712965049496465000,"InstrumentationScope":"gitserver","Caller":"gitcli/command.go:286","Function":"github.com/sourcegraph/sourcegraph/cmd/gitserver/internal/git/gitcli.(*cmdReader).trace","Body":"Long exec request","Resource":{"service.name":"gitserver","service.version":"0.0.0+dev","service.instance.id":"127.0.0.1:3178"},"TraceId":"db86930d642952fd9b2dd3df0ec7e380","SpanId":"e41abbebd5b8c832","Attributes":{"ev.Fields":{"cmd":"config","args":"[git config sourcegraph.type git]","duration_ms":"6","exit_status":"0","user_time":"1.957ms","traceID":"db86930d642952fd9b2dd3df0ec7e380","trace":"https://sourcegraph.test:3443/-/debug/jaeger/trace/db86930d642952fd9b2dd3df0ec7e380","repo":"github.com/sourcegraph/deploy-sourcegraph","actor":"0","cmd_duration_ms":"6","system_time":"2.311ms"}}}
```


internal/honey/event.go
```diff
@@ -21,9 +21,6 @@ type Event interface {
 	AddField(key string, val any)
 	// AddAttributes adds each otel/attribute key-value field to this event.
 	AddAttributes([]attribute.KeyValue)
-	// Add adds a complex type to the event. For structs, it adds each exported field.
-	// For maps, it adds each key/value. Add will error on all other types.
-	Add(data any) error
```
ggilmore, line 26: We don't use the Add function anywhere in the codebase. Simply removing this from the interface means that we don't have to re-implement this unnecessarily.

internal/honey/non_sending.go
```diff
@@ -0,0 +1,70 @@
...
+// For nonSendingEvent, this is always an empty string.
+func (e *nonSendingEvent) Dataset() string {
+	return ""
+}
+
+// AddField adds a single key-value pair to this event.
+func (e *nonSendingEvent) AddField(key string, val any) {
+	if e.fields == nil {
+		e.fields = make(map[string]any)
+	}
```
eseliger, line 32: does the event need to be concurrency safe? would the real event be safe?

eseliger (approved):

internal/honey/non_sending.go
ggilmore: Ah, that's a good call out. Funnily enough, the event type from libhoney is threadsafe - but I think our eventwrapper isn't 🙃 

I'll wrap a mutex around this anyway.

internal/honey/non_sending_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+		var wg sync.WaitGroup
+		for i := 0; i < 1000; i++ {
+			wg.Add(1)
+			go func(i int) {
+				defer wg.Done()
+				event.AddField(fmt.Sprintf("key%d", i), i)
+			}(i)
+		}
+		wg.Wait()
```
varungandhi-src, line 42: This kind of code would be slightly simpler with our `conc` library.

```
		wg := conc.NewWaitGroup()
		for i := 0; i < 1000; i++ {
			wg.Go(func() {
				event.AddField(fmt.Sprintf("key%d", i), i)
			})
		}
		wg.Wait()

```

PR #61872, opened 2024-04-15
Gateway: Increase response token limit

umpox:
## Description

Part of https://github.com/sourcegraph/cody/issues/3648

This PR:
- Increases the maximum output limits in the Cody Gateway from `1000` to `4000` tokens
- This is paired with https://github.com/sourcegraph/cody/pull/3797, where we will increase this limit (currently for dotcom only, but enterprise in the future)

## Test plan

1. Run https://github.com/sourcegraph/sourcegraph/pull/61872 locally
2. Trigger long context responses (e.g. "Show me 300 lines of typescript code")
3. Expect that everything functions as normal


cmd/cody-gateway/internal/httpapi/completions/anthropicmessages_test.go
```diff
@@ -23,16 +23,16 @@ func TestIsFlaggedAnthropicMessagesRequest(t *testing.T) {
 		PromptTokenFlaggingLimit:       18000,
 		PromptTokenBlockingLimit:       20000,
 		MaxTokensToSample:              0, // Not used within isFlaggedRequest.
-		MaxTokensToSampleFlaggingLimit: 1000,
-		ResponseTokenBlockingLimit:     1000,
+		MaxTokensToSampleFlaggingLimit: 4000,
```
umpox, line 26: @chrsmith My understanding is that we're not really flagging anything with `MaxTokensToSampleFlaggingLimit` right now, as the limit is the same as the max tokens (1000)

Given we'll be increasing the max tokens to 4000, do you think we should set this to a different value? I want to make sure we don't have any concerns around abuse when increase this limit

philipp-spiess (commented):
Looks good from my side but I think it makes sense to have @chrsmith stamp this.

I also made sure there are no env flag overwrites for the max tokens yet: https://github.com/sourcegraph/infrastructure/blob/main/cody-gateway/envs/prod/cloudrun/main.tf

abeatrix (commented):
Do we need to increase the response timeout limit to accommodate this change?

chrsmith (approved):
After carefully reviewing the changes, double-checking the LLM provider docs, our flagging logic, etc. This all looks good to me. 🎉 :shipit: 🚀!

However, I would also say this PR is "dangerous". No need to make any changes, but I'll tag you on another PR to fix the "problems" introduced here.

My concern is that this PR is only changing the default values for some settings. But these new, higher token limits aren't optional. The Cody product simply _won't work_ if the flagging/prompt max token limits are too low.

In other words, this is a required change in expected behavior. And relying on the default settings is extremely fragile. So IMO, the correct change would be to _explicitly_ set these values in the `sourcegraph/infrastructure` configuration. So that it is crystal clear what these values will be in production, and why they need to be set the way they are.

(But changing the defaults via this PR makes sense too, since we still want things to work when running locally as well.)

Also, another reason I want to pair this with a PR in the `sourcegraph/infrastructure` repo is so that I can finally document all of these poorly named and confusing settings. (TBH I didn't quite grok them all before looking at this PR, and now that I do I want to avoid paging in this context ever again.... I'm looking at you `MAX_TOKENS_TO_SAMPLE`, `MAX_TOKENS_TO_SAMPLE_FLAGGING_LIMIT`, and `RESPONSE_TOKEN_BLOCKING_LIMIT`!)

> Do we need to increase the response timeout limit to accommodate this change?

Yes? My understanding is that the time it takes to serve the request will scale linearly with the number of tokens in the response. So by setting the limits to 4x what they were previously, we can expect the long tail to be ~4x as long.

Though unfortunately, I don't know where (or even if) we set a timeout/deadline for the HTTP request made to upstream handlers.

We just pass along the same `context.Context` in `upstream.go` [here](https://github.com/sourcegraph/sourcegraph/blob/main/cmd/cody-gateway/internal/httpapi/completions/upstream.go#L283). So we'd need to trace the callstack and HTTP handler middleware to figure out where any timeouts get added to the request.

(Read: we should probably do that in a follow-up PR after we can do some testing after we roll out this change.)

abeatrix:
> Though unfortunately, I don't know where (or even if) we set a timeout/deadline for the HTTP request made to upstream handlers.

@chrsmith Could this be the right place? Right now it's set to 2mins in  https://sourcegraph.com/github.com/sourcegraph/sourcegraph@main/-/blob/internal/completions/httpapi/handler.go?L39-41

REF: https://github.com/sourcegraph/sourcegraph/pull/59875

cmd/cody-gateway/internal/httpapi/completions/anthropicmessages_test.go
chrsmith: I spent a long while looking through the code, and where "max tokens" is used and how it relates to LLM provider APIs... and I think the answer is that it does make sense to keep these as "separate" configuration knobs, because in theory there could be reasons why we might want to have different limits for the max sample size, flagging limit, and blocking limit.

... but for practical purposes, just keeping these as the same value is fine. When I get to that TBD PR to set the configuration values in the `main.tf` file (rather than rely on the default values), I'll add some obnoxiously long comments to explain the difference. (Since it certainly didn't make sense to me until I dug into it.)

umpox:
@chrsmith Thanks for taking a detailed look!

> IMO, the correct change would be to explicitly set these values in the sourcegraph/infrastructure configuration. So that it is crystal clear what these values will be in production, and why they need to be set the way they are.

I agree, I guess it's been pretty fragile for a while so this would be a good change. I know you said you'll add a follow-up PR with detailed comments for each of these vars so I'll leave this one for you. I wasn't aware of how we use flagging etc internally so it'd be super helpful for the future. 

@abeatrix Thanks for finding that ❤️  I've updated it to 4x the current limit. That feels like a really generous limit right now so we can keep an eye out for any long running requests. We also don't do anything dynamic based on the max values set here so we could consider updating that if we notice any issues

PR #61873, opened 2024-04-15
Change link URL and label, and display only for Pro

vdavid:
Closes https://github.com/sourcegraph/self-serve-cody/issues/510

This is what it used to look like:
Before

This is what it looks like now for Pro users:

After - Pro users

And for Free users:

After - Free users

## Test plan

Checked out https://sourcegraph.test:3443/cody/manage with both a Pro and a Free user, to see the diff. See the screenshots above.


thenamankumar (approved):
LGTM

PR #61875, opened 2024-04-15
sg: add `application/vnd.oci.image.manifest.v1+json` to accepted image types for `ops update-images`

jac:
Docker image type has changed. Update the accepted types to fix the `sg ops update-images` command by including `application/vnd.oci.image.manifest.v1+json`
## Test plan

tested locally command ran successfully


Strum355 (approved):
Danke good sir

PR #61876, opened 2024-04-15
Log events in a non-blocking way

vdavid:
We got this from a customer ([source](https://sourcegraph.slack.com/archives/C05SZB829D0/p1713020068425919))

> Hi! 
> 
> I have tried without success quite a few times to log in to start using Cody, but the onboarding process is stuck on the first hurdle.
> 
> Clicking get started does nothing and a 403 error is in the console. 
> 
> I've disabled all my plugins and tried multiple browsers, and nothing seems to work. 
> 
> I can get to the dashboard by editing the html and deleting the modal, but when I try to log in from my Rider IDE, it just comes here, and I can't log in.
> 
> Can you help?

We suspect it's due to failing to dismiss the dialog until we emit some telemetry event, which is being blocked at the DNS level or similar.

## Test plan

Add `console.log` after the logging code and see whether it gets triggered after this change, plus check if there's no race condition by navigating away from the page shortly after submitting the form (while having the `ab-hubspot-form-workpersonal-to-handraiser` ff enabled. Done this, got the console log, and experienced no page reload:

CleanShot 2024-04-15 at 13 12 57@2x

### DIFF
--- a/client/web/src/cody/onboarding/CodyOnboarding.tsx
+++ b/client/web/src/cody/onboarding/CodyOnboarding.tsx
@@ -308,10 +308,13 @@ function PurposeStep({
         if (choice) {
             const metadata = { variant: 'treatment' }
 
-            eventLogger.log(EventName.CODY_HANDRAISER_TEST_ENROLLMENT, metadata, metadata)
-            telemetryRecorder.recordEvent('cody.onboarding.purpose', 'select', {
-                metadata: { onboardingCall: choice.checked ? 1 : 0 },
-            })
+            void new Promise(resolve => {
+                eventLogger.log(EventName.CODY_HANDRAISER_TEST_ENROLLMENT, metadata, metadata)
+                telemetryRecorder.recordEvent('cody.onboarding.purpose', 'select', {
+                    metadata: { onboardingCall: choice.checked ? 1 : 0 },
+                })
+                resolve(undefined)
+            }).then(() => {})
         }
     }
 
--- a/client/web/src/cody/onboarding/CodyOnboarding.tsx
+++ b/client/web/src/cody/onboarding/CodyOnboarding.tsx
@@ -314,7 +314,11 @@ function PurposeStep({
                     metadata: { onboardingCall: choice.checked ? 1 : 0 },
                 })
                 resolve(undefined)
-            }).then(() => {}).catch(() => { /* Swallow errors */})
+            })
+                .then(() => {})
+                .catch(() => {
+                    /* Swallow errors */
+                })
         }
     }
 
@@ -331,11 +335,25 @@ function PurposeStep({
                     formId="19f34edd-1a98-4fc9-9b2b-c1edca727720"
                     onFormSubmitted={() => {
                         onNext()
-                        telemetryRecorder.recordEvent('cody.onboarding.qualificationSurvey', 'succeeded', {})
+                        void new Promise(resolve => {
+                            eventLogger.log(EventName.CODY_ONBOARDING_FORM_SUBMITTED)
+                            telemetryRecorder.recordEvent('cody.onboarding.qualificationSurvey', 'succeeded', {})
+                        })
+                            .then(() => {})
+                            .catch(() => {
+                                /* Swallow errors */
+                            })
                     }}
                     onFormLoadError={() => {
                         onNext()
-                        telemetryRecorder.recordEvent('cody.onboarding.qualificationSurvey', 'failed', {})
+                        void new Promise(resolve => {
+                            eventLogger.log(EventName.CODY_ONBOARDING_FORM_LOAD_ERRORED)
+                            telemetryRecorder.recordEvent('cody.onboarding.qualificationSurvey', 'failed', {})
+                        })
+                            .then(() => {})
+                            .catch(() => {
+                                /* Swallow errors */
+                            })
                     }}
                     userId={authenticatedUser.id}
                     userEmail={primaryEmail}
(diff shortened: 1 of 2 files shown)


client/web/src/cody/onboarding/CodyOnboarding.tsx
```diff
@@ -308,10 +308,13 @@ function PurposeStep({
         if (choice) {
             const metadata = { variant: 'treatment' }
 
-            eventLogger.log(EventName.CODY_HANDRAISER_TEST_ENROLLMENT, metadata, metadata)
-            telemetryRecorder.recordEvent('cody.onboarding.purpose', 'select', {
-                metadata: { onboardingCall: choice.checked ? 1 : 0 },
-            })
+            void new Promise(resolve => {
```
thenamankumar, line 311: I think what we need here is a `try` & `catch` block to handle the suspected runtime error and now the promise. 

Moreover, if we can verify that indeed one of the two loggers is failing during runtime, then IMO it will be better to put the `try` & `catch` inside the func definition and not at the call site.

vdavid:
I’ve paired with @Naman on this and his thoughts were:
- It’s unlikely that either of the logging calls (old and new) are blocking indefinitely.
- It’s more likely that they throw an error, which we don’t catch.
- It feels a bit wrong for these logging calls to throw errors as we’re calling them from various places. It's true, but I guess, at least this gives the caller a way to cope with logging errors.

As a result:
- I still kept the promise because the internals of the event logger look pretty sync so I’m not sure if it truly can’t be blocking. But I added a ().catch() block after the promise to swallow any errors.
- As a byproduct, I’ve also added logging to two other events to aid debugging.

client/web/src/cody/onboarding/CodyOnboarding.tsx
vdavid: Done, thanks! I ended up wrapping the two new logging calls in promises with `.catch(()=>{})` for the same reason I did for `onFormSubmit`. Can't hurt.

thenamankumar:
I recommend testing this hypothesis by manually raising an error from inside the submit handler promise and then testing if the submit success callback is still called.

client/web/src/cody/onboarding/CodyOnboarding.tsx
```diff
@@ -308,10 +308,17 @@ function PurposeStep({
         if (choice) {
             const metadata = { variant: 'treatment' }
 
-            eventLogger.log(EventName.CODY_HANDRAISER_TEST_ENROLLMENT, metadata, metadata)
```
chrsmith, line 311: ⚠️ we should put a `try/catch` block around this entire function, since it's possible loading `choice` will fail.

```ts
const handleFormSubmit = (form: HTMLFormElement): void => {
    // We dismiss the HubSpot form regardless of any errors, as to not block
    // the getting started experience.
    try {   
        const choice = form[0].querySelector<HTMLInputElement>('input[name="cody_form_hand_raiser"]')
        if (choice) {
            const metadata = { variant: 'treatment' }
        ...
    } catch (ex) {
        console.error("error handling form submit:", ex)
    }
}
```

chrsmith (changes requested):
Some minor notes. But I think the problem is here:
https://community.hubspot.com/t5/APIs-Integrations/Form-submission-returning-403-SOLVED/m-p/231465

Which I think Kevin tracked down but I haven't caught up to it on Slack, so maybe this PR is out of date.

Anyways, the key thing we needed to pay attention to was the "403 error" from the bug report. Since none of our logging/telemetry stuff would serve a 403 error response.

chrsmith (approved):
Update #2, after spending more time looking for the underlying problem.

---

I don't think this PR will actually fix anything, but at the very least having more logging would help. So feel free to submit after making the suggested cleanups.

As far as what the root problem is, I still haven't found a clear answer.

I think the most important part is that the initial user reported seeing a "403 error" in their JavaScript console. It seems extremely unlikely that our telemetry/logging stack would return that error... so I suspect it is coming from HubSpot.

Here's my theory:

> HubSpot renders its form (which is probably an iframe or something), and exposes event handlers that we tie into. (e.g. `onSubmit`.
>
> [The documentation](https://legacydocs.hubspot.com/docs/methods/forms/advanced_form_options)(?) doesn't do a great job explaining this. But I'm guessing that the the user clicks submit on the HubSpot form, and the failure comes when processing the request. (And _before_ HubSpot passes control back to us, invoking our supplied `onSubmitHandler`.)
>
> You can see various forum posts and things about HubSpot form submissions returning 403s if there are validation errors. Or ways to add more aggressive validation logic, e.g.:
> - https://rschu.me/prevent-the-submit-of-any-hubspot-forms-after-a-successful-validation-5e1f230b14f1
> - https://gist.github.com/DuaneNielsen/78a25225bcb125bfd009ea877f218f20
>
> Anyways, I'm guessing that HubSpot is failing for some reason and we just aren't _displaying_ the "validation errors" in the HubSpot form. (Because we haven't configured that.) So no matter what validation logic we put on our-end, the error is being emitted and swallowed by HubSpot. (e.g. I expect the validation error to be `submissionErrors.OUT_OF_DATE` or similar, from that [advanced_form_options docs](https://legacydocs.hubspot.com/docs/methods/forms/advanced_form_options) page.

If my theory is right, then the best we can do is simply update the validation error CSS/display information in the way we configure the embedded form. So that if there is one of these unexpected errors, the end user will at least see that _something_ has gone wrong. (Because our telemetry/logging wouldn't observe it.)

vdavid:
We've removed `?isPro=true` from the last place where we had it (I couldn't find the PR that did it, but I couldn't find the isPro link in the current codebase, plus couldn't repro [this flow](https://sourcegraph.slack.com/archives/C05SZB829D0/p1713203584160869?thread_ts=1713020068.425919&cid=C05SZB829D0) anymore, so looks like this is fixed. The added logging doesn't seem too helpful, after all, so just closing this PR.

PR #61880, opened 2024-04-15
highlight: share log attributes with trace and handle canceled

keegancsmith:
While investigating a large number of errors in highlighting I noticed the vast majority where canceled. This commit adds special handling for it and in the process does a few other things to the observability here:

- Log duration in these errors. This is useful signal to understand if a slow request is being canceled, or we have bad debouncing in a client.
- Re-use the tracers logger. This has many more fields and avoids duplication.
- Downgrade cancel error to warning.
- Include error problem in logs.
- Factor out the snippet log field creation.

Note I didn't include snippet in the cancel log since it's unlikely to be related to the code being highlighted. Easy to add in if we find that is incorrect.

Test Plan: CI will exercise this code path. Otherwise I will follow-up on dotcom.

### DIFF
--- a/internal/highlight/highlight.go
+++ b/internal/highlight/highlight.go
@@ -317,8 +317,6 @@ func Code(ctx context.Context, p Params) (response *HighlightedCode, aborted boo
 		return Mocks.Code(p)
 	}
 
-	logger := log.Scoped("highlight")
-
 	p.Filepath = normalizeFilepath(p.Filepath)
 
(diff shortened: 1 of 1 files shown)


keegancsmith:
@camdencheek any idea why you only emit errors to honeycomb? I am guessing error logging/metrics/traces would be useful here? https://github.com/sourcegraph/sourcegraph/blob/be908b2135b35d3afc5e36434a08a98212657187/internal/highlight/highlight.go#L54

mmanela (approved):

PR #61881, opened 2024-04-15
Build our own caddy image

willdollman:
We currently just retag the upstream Caddy image where we [use it in deploy-sourcegraph-docker](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/deploy-sourcegraph-docker@814afb2762cd86d21ce4cf270b9b8855204d6f91/-/blob/docker-compose/docker-compose.yaml?L72#tab=history). The upstream is slow to patch go module vulns, and the update process is different to the rest of our images.

Build a caddy container from scratch here to avoid having a separate update process, and to make use of the latest patched version of caddy from the wolfi repo.

## Test plan

- Added container structure test
- CI


wolfi-images/repo.bzl
```diff
@@ -40,6 +40,9 @@ def apko_translate_locks():
 # See the following commit to see what is being removed from WORKSPACE by this repository rule:
 # https://github.com/sourcegraph/sourcegraph/pull/60785/commits/041fb7a177c8f9004a973306b2e045a25e64fc68
 def _wolfi_lockfiles(rctx):
+    # Used to invalidate this repository when any lockfiles change.
+    rctx.path(rctx.path("wolfi-images"))
+
```
willdollman, line 45: @Strum355 this works in some basic testing but not sure if there's a cleaner way to handle this. The issue I ran into was after adding the caddy image's yaml and lock.json files, the new lockfile wasn't in the dependency list -> repo wasn't invalidated -> repo isn't added to list -> GOTO 1.

Strum355: You can use `rctx.watch("wolfi-images")` for this now that we're on bazel 7.1

willdollman:
Tested the `rctx.watch_tree()` change by creating a new yaml + lockfile and running `bazel build @mynewfile_apko_lock//:contents`. It works, and without the change it fails.

Strum355 (approved):
love a streamlined process

PR #61883, opened 2024-04-15
rbac: make user permissions additive

BolajiOlajide:
Before this PR, RBAC permissions were automatically added to the User role and devs would have to explicitly declare permission namespaces that shouldn't be added to the user role.
This PR flips this logic by ensuring devs would explicitly have to specify permissions to be added to the User namespace.

[Context](https://app.incident.io/sourcegraph/incidents/292?source=slack_bookmark)

## Test plan

- Manual testing
- Add unit test

### DIFF
--- a/cmd/frontend/internal/bg/update_permissions.go
+++ b/cmd/frontend/internal/bg/update_permissions.go
@@ -53,21 +53,15 @@ func UpdatePermissions(ctx context.Context, logger log.Logger, db database.DB) {
 				return errors.Wrap(err, "creating new permissions")
 			}
 
-			roles := []types.SystemRole{types.SiteAdministratorSystemRole, types.UserSystemRole}
-			excludedRoles := collections.NewSet[rtypes.PermissionNamespace](rbacSchema.ExcludeFromUserRole...)
+			permissionsToIncludeForUserRole := collections.NewSet[rtypes.PermissionNamespace](rbacSchema.IncludeInUserRole...)
 			for _, permission := range permissions {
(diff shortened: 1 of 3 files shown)


internal/rbac/schema.yaml
```diff
@@ -17,5 +17,7 @@ namespaces:
     actions:
       - READ
       - WRITE
-excludeFromUserRole:
-  - OWNERSHIP
+includeInUserRole:
```
eseliger, line 20: Naming suggestion: `userDefaultRoles` conveys IMO better what this is used for, so you don't just blindly copy a new one here thinking that'd be required.

cmd/frontend/internal/bg/update_permissions.go
eseliger, line 1: No test needed to be updated for this?

eseliger (approved):

cmd/frontend/internal/bg/update_permissions.go
BolajiOlajide: Oh! I'm adding a test definitely. That's why it's in draft status. Sorry I shared early lol, but good to get early feedback hehe.

eseliger: Oh lol sorry didn't see that, I followed the incident.io link 😆

BolajiOlajide: No worries at all. Appreciate the feedback.

eseliger, line 1: No test needed to be adjusted?

eseliger (approved):

evict (approved):
🔒 👍

PR #61898, opened 2024-04-16
frontend/auth: fix and test telemetry V2 user assignment

bobheadxi:
Some pretty gnarly code out here, but the gist is to fix some missing user IDs from the V2 telemetry that this massive `GetAndSaveUser` thing reports, caused by a lack of user ID in the context.

The core changes are:

1. Parameterize some `GetAndSaveUser` dependencies (logger, telemetry recorder) to make testing easier (improve visibility of logs on failures etc). The change cascades a bit but should be harmless.
2. Generate a `telemetryCtx` with user actor inserted at various point so that telemetry recorder can pick it up. This is a separate context to avoid unexpected auth security problems.
3. Add tests that assert users are (or aren't) included in telemetry events recorded by `GetAndSaveUser`

## Test plan

New and existing tests pass


cmd/frontend/auth/user.go
```diff
@@ -313,21 +329,27 @@ func GetAndSaveUser(ctx context.Context, db database.DB, op GetAndSaveUserOp) (n
...
-		),
-		PrivateMetadata: map[string]any{"serviceType": acct.AccountSpec.ServiceType},
-	})
+	defer func() {
+		action := telemetry.ActionSucceeded
+		if err != nil { // check final error
+			action = telemetry.ActionFailed
+		}
+		serviceVariant, _ := extsvc.VariantValueOf(acct.AccountSpec.ServiceType)
+		recorder.Record(telemetryCtx, telemetryV2UserSignUpFeatureName, action, &telemetry.EventParameters{
```
bobheadxi, line 338: There are more ways to error out of this function - this makes use record the event only at the very end, and accurately represent if the operation succeeded or not

cmd/frontend/auth/user_test.go
```diff
@@ -494,6 +506,33 @@ func TestGetAndSaveUser(t *testing.T) {
...
+
+							// Scenarios where we should not have any user found
+							case oc.mock.getByUsernameErr != nil ||
+								oc.mock.getByVerifiedEmailErr != nil ||
+								oc.mock.externalAccountUpdateErr != nil:
+								assert.Nil(t, ev.GetUser(),
+									"Event '%s#%s' should not have user ID, found: %s",
+									ev.GetFeature(), ev.GetAction(), ev.GetUser().GetUserId())
+							}
+						}
```
bobheadxi, line 535: An awful lot of exit paths here. Tried to be as specific as possible, but these scenarios seem reasonable to me so far

cmd/frontend/auth/user.go
```diff
@@ -77,13 +77,19 @@ const telemetryV2UserSignUpFeatureName = "externalAuthSignup"
 // 🚨 SECURITY: The safeErrMsg is an error message that can be shown to unauthenticated users to
 // describe the problem. The err may contain sensitive information and should only be written to the
 // server error logs, not to the HTTP response to shown to unauthenticated users.
-func GetAndSaveUser(ctx context.Context, db database.DB, op GetAndSaveUserOp) (newUserCreated bool, userID int32, safeErrMsg string, err error) {
+func GetAndSaveUser(
+	ctx context.Context,
+	logger sglog.Logger, // accept logger for scoping from various callsites, and readability of test logs
```
bobheadxi, line 82: Huge improvement - if test fails a propagated logger can use `t.Log` to render its output

cmd/frontend/auth/user.go
```diff
@@ -77,13 +77,19 @@ const telemetryV2UserSignUpFeatureName = "externalAuthSignup"
...
+	db database.DB,
+	tr *telemetry.EventRecorder, // accept explicit recorder for ease of testing
```
bobheadxi, line 84: Easier to use than mocking our `DB` megainterface

bobheadxi:
@eseliger only tagged you here since you were in this recently and maybe I could get a sanity check - no worries if you don't want to though 😁

cmd/frontend/internal/auth/openidconnect/user.go
```diff
@@ -28,7 +31,19 @@ type ExternalAccountData struct {
 // getOrCreateUser gets or creates a user account based on the OpenID Connect token. It returns the
 // authenticated actor if successful; otherwise it returns a friendly error message (safeErrMsg)
 // that is safe to display to users, and a non-nil err with lower-level error details.
-func getOrCreateUser(ctx context.Context, db database.DB, p *Provider, token *oauth2.Token, idToken *oidc.IDToken, userInfo *oidc.UserInfo, claims *userClaims, usernamePrefix string, userCreateEventProperties telemetry.EventMetadata, hubSpotProps *hubspot.ContactProperties) (newUserCreated bool, _ *actor.Actor, safeErrMsg string, err error) {
+func getOrCreateUser(
+	ctx context.Context,
+	logger log.Logger,
```
eseliger, line 36: jay, another property 😆

cmd/frontend/auth/user_test.go
eseliger: I wonder if it would be a bit easier to let the test table descriptions indicate `shouldHaveUserInLog bool` or something like that?

cmd/frontend/auth/user.go
```diff
@@ -242,13 +247,24 @@ func GetAndSaveUser(ctx context.Context, db database.DB, op GetAndSaveUserOp) (n
 
 		return user.ID, true, true, "", nil
 	}()
+
+	// Annotate the telemetry ctx with the user ID we found, irrespective of
+	// successful auth, and avoid clobbering any existing authenticated actor in
+	// the context - this is only used for telemetry.
+	//
+	// 🚨 SECURITY: Only use telemetryCtx for telemetry operations.
```
eseliger, line 255: I wonder if there could be something like

```go
func WithActorOverride(int32) TelemetryOptFunc

func (r*recorder) Record([...], opts TelemetryOptFunc...) [...]
```

that would allow us to do this more explicitly than creating a context here which we are scared of might be used for something else. Not a blocker for this PR, maybe some future improvement we could make?

eseliger (approved):
Change seems good,  I trust you that this megafunction is not exploding 😆 .. and acknowledge that we at some point have to untangle this mess 😀

cmd/frontend/auth/user.go
```diff
@@ -242,13 +247,24 @@ func GetAndSaveUser(ctx context.Context, db database.DB, op GetAndSaveUserOp) (n
...
+
 	if err != nil {
 		// Retain legacy event logging format since there are references to it
-		ctx = teestore.WithoutV1(ctx)
+		telemetryCtx = teestore.WithoutV1(telemetryCtx)
 
 		// New event - most external services have an exstvc.Variant, so add that as safe metadata
 		serviceVariant, _ := extsvc.VariantValueOf(acct.AccountSpec.ServiceType)
-		recorder.Record(ctx, telemetryV2UserSignUpFeatureName, telemetry.ActionFailed, &telemetry.EventParameters{
+		recorder.Record(telemetryCtx, telemetryV2UserSignUpFeatureName, telemetry.ActionFailed, &telemetry.EventParameters{
```
dadlerj, line 267: Should we (1) pull the `defer`-ed function that handles both success and failure cases up in this function so that it's created _before_ the big closure (or immediately after), and (2) remove this `recorder.Record` call altogether (since then it would be duplicative I believe? That would also ensure that the various error cases in the block below this one (line 291–325 in this branch) result in a telemetry event being recorded.

dadlerj (commented):
Thank you so much for this!

cmd/frontend/auth/user.go
bobheadxi: This is a great idea, taking it for a sping

cmd/frontend/auth/user_test.go
bobheadxi: I thought about that, but:

1. This kind of requires me (and everyone else) to have a stronger understanding of the bajillion cases here than I really have (or anyone else might have)
2. Makes the already-complex test case descriptions even more complex

Having a summary here of, "these errors result in a user attached to the event, these errors do not" is also not too bad for readability from my end, and helps demonstrate what's happening

I did end up adding an `assert.Failf` here if there is a scenario that is not covered - is that too jank?

eseliger: No should be fine - thanks for checking!

cmd/frontend/internal/auth/openidconnect/user.go
bobheadxi: On the bright side, this is actually a good use case for logger propagation because the function at the end of the trail is called from many callsites - logger propagation gives things more context

cmd/frontend/auth/user.go
bobheadxi: I'm a bit hesitant to add this because I'm worried it'll confuse callers, and also at the end of the day it all boils down to who we are acting as in the context - for the telemetry case, if you end up reporting telemetry in a particular user context that should be an explicit context and not some additional parameter. WDYT?

cmd/frontend/auth/user.go
bobheadxi: Did this in 009c69f6bf4526e9419bad1bb7952d253fbcea0e and according to test cases, it works great 😁

cmd/frontend/auth/user.go
eseliger: I guess this is quite an edge-case where when the request starts no actor is set but during execution it'll be changed - so probably no broad need for this. 
I am mostly scared that we create a new context here but then are scared and have to add this comment that someone might be using it wrong.

Maybe we can encapsulate it in a function here?

```go
func logForActor(ctx context.Context, recorder Recorder, actorID int32, event ...) {
ctx = actor.WithActor()
recorder.Log(event)
}
```

bobheadxi: I mention this in a new docstring but we need to finalize the actor immediately after the closure, and then log the event in the closure based on the other final return values, to avoid having it be clobbered by the returned ID (which may not be the ID of interest it seems). We could save this to another `telemetryUserID` but I'm not sure that is significantly better 😰

eseliger: gotcha - yep no blocker :)

cmd/frontend/auth/user.go
```diff
@@ -242,27 +246,61 @@ func GetAndSaveUser(ctx context.Context, db database.DB, op GetAndSaveUserOp) (n
...
+		privateMetadata := map[string]any{"serviceType": acct.AccountSpec.ServiceType}
+
+		// Include safe err if there is one for maybe-useful diagnostics
+		if len(safeErrMsg) > 0 {
+			privateMetadata["safeErrMsg"] = safeErrMsg
+		}
+
+		// Record our V2 event.
+		recorder.Record(telemetryCtx, telemetryV2UserSignUpFeatureName, action, &telemetry.EventParameters{
+			Version: 2, // We've significantly refactored telemetryV2UserSignUpFeatureName occurrences
```
unknwon, line 282: qq: What does this comment mean in this context exactly?

cmd/frontend/auth/user_test.go
```diff
@@ -494,6 +504,42 @@ func TestGetAndSaveUser(t *testing.T) {
...
+
+							// Scenarios where we should not have any user found
+							case oc.mock.getByUsernameErr != nil ||
+								oc.mock.getByVerifiedEmailErr != nil ||
+								oc.mock.externalAccountUpdateErr != nil ||
+								oc.mock.createWithExternalAccountErr != nil ||
+								!c.op.CreateIfNotExist:
+								assert.Nil(t, ev.GetUser(),
+									"Event '%s#%s' should not have user ID, found: %s",
+									ev.GetFeature(), ev.GetAction(), ev.GetUser().GetUserId())
```
unknwon, line 536: Doesn't seem entirely clear why certain DB-level mocks exists should imply the behavior of the caller 😂 but I trust you have done the best you can do here.

cmd/frontend/internal/auth/saml/user.go
```diff
@@ -81,6 +84,8 @@ func readAuthnResponse(p *provider, encodedResp string) (*authnResponseInfo, err
 // authenticated actor if successful; otherwise it returns an friendly error message (safeErrMsg)
 // that is safe to display to users, and a non-nil err with lower-level error details.
 func getOrCreateUser(ctx context.Context, db database.DB, allowSignup bool, info *authnResponseInfo) (newUserCreated bool, _ *actor.Actor, safeErrMsg string, err error) {
+	logger := log.Scoped("saml") // TODO: propagate logger from callers
```
unknwon, line 87: Is this TODO meant for this PR?

unknwon (approved):

cmd/frontend/auth/user.go
bobheadxi: the version field is a utility indicator that indicates the way an event has been recorded has changed significantly, as has happened in this PR. For example, this scenario covers more situations better

unknwon: Oh... I see, it's version of the event type.

cmd/frontend/auth/user_test.go
bobheadxi: In auth errors, we don't tell the caller who the user is, but for telemetry we still want to attach the event to the user that failed

bobheadxi: a little more discussion here: https://github.com/sourcegraph/sourcegraph/pull/61898#discussion_r1566759305

cmd/frontend/internal/auth/saml/user.go
bobheadxi: I got lazy after doing some propagation... was hoping to slip this in unnoticed, and the next person will do this 😛 😉

unknwon: Haha, fair.

cmd/frontend/auth/user_test.go
unknwon: Ah yeah, that's because we are testing against the presentation layer directly... but refactoring this function is just too scary...

bobheadxi: > but refactoring this function is just too scary...

Yeah that's the gist of it 😅 😰

cmd/frontend/auth/user.go
```diff
@@ -242,27 +246,61 @@ func GetAndSaveUser(ctx context.Context, db database.DB, op GetAndSaveUserOp) (n
...
+		telemetryCtx = sgactor.WithActor(telemetryCtx, actor.FromUser(userID))
+	}
+	// We handle all V2 telemetry related to GetAndSaveUser within this defer
+	// closure, to ensure we cover all exit paths correctly after the other mega
+	// closure above.
+	defer func() {
+		action := telemetry.ActionSucceeded
+		if err != nil { // check final error
+			action = telemetry.ActionFailed
+		}
```
akalia25, line 267: Hey @bobheadxi, we've noticed a **significant** increase in our `externalAuthSignup` `succeeded` event, where it is firing almost 5M times a day on v2 telemetry. Is it possible that this line of code is misfiring because on the `externalAuth` event we have a `userId` `null`?

bobheadxi: We discussed sync earlier - the spike is called by an increase in the old version of this event (`parameters.version < 2`), prompted by what seems to be a migration of a large customer to Cloud. It should be resolved in an upcoming Cloud release, as this fix only landed fairly recently - in the meantime we can drop events from this feature with `parameters.version < 2`

PR #61903, opened 2024-04-16
svcmain: Simplify service setup

eseliger:
Since we switched away from single binary, we always pass down a single service. So we can simplify this logic a little bit.

Test plan:

sg start still works and the E2E / integration test suites are still passing.


vdavid (approved):
Oh, that's nice and tidy.
I see no problems with the code 👍

unknwon (approved):

PR #61904, opened 2024-04-16
Remove unnecessary dotcom checks

eseliger:
These checks for dotcom mode are no longer necessary, so let's remove the number of code paths.

Test plan:

Test suites are still passing, also code review.

### DIFF
--- a/cmd/frontend/graphqlbackend/org.go
+++ b/cmd/frontend/graphqlbackend/org.go
@@ -313,7 +313,6 @@ func (o *OrgResolver) BatchChanges(ctx context.Context, args *ListBatchChangesAr
 func (r *schemaResolver) CreateOrganization(ctx context.Context, args *struct {
 	Name        string
 	DisplayName *string
-	StatsID     *string
 },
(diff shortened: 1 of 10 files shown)


cmd/frontend/graphqlbackend/org.go
```diff
@@ -443,12 +434,6 @@ func (r *schemaResolver) AddUserToOrganization(ctx context.Context, args *struct
 		return nil, err
 	}
 
-	// 🚨 SECURITY: Do not allow direct add on Cloud unless the site admin is a member of the org
```
eseliger, line 446: @evict this seems ok, does it?

We used to only allow admins that are _also_ members of an org add a member to said org.

That prevented admins to gain access to an orgs private code in v1 cloud, which no longer exists.

evict: Makes sense!

evict (approved):
🧹

unknwon (commented):
Thank you 😁

PR #61910, opened 2024-04-16
Use time.Time in ContributorCount Git API

eseliger:
This uses our internal package to parse dates, instead of passing the date onto git directly. This makes for a cleaner API and the same date parsing across the product.

Test plan:

Tested the contributor count page by hand.


camdencheek (approved):
Nice!

PR #61932, opened 2024-04-16
Appliance reconciliation helpers

craigfurman:
Based on https://github.com/sourcegraph/sourcegraph/pull/61834. Thsi PR should auto-rebase onto main when that is merged.

This is a bit of an odd PR, it'll need some explaining. It's probably even stranger than I realise, since this is my first contribution to this project. I'm of course very open to feedback, and this might be the wrong approach entirely. Let's go commit-by-commit (this will hopefully make more sense than reading the whole diff):

**appliance: set pseudo-CRD name values**

When the appliance deploys a ConfigMap to a given k8s namespace, we want to provision resources in that namespace. In a CRD-driven operator, the `Sourcegraph` object would be that CRD, so more-deeply-nested code could just take the namespace from that. But in this appliance, the Sourcegraph object that we pass around is just some deserialised config (from the ConfigMap). It still has its Namespace field due to having been scaffolded by Kubebuilder, but it's always empty string, which is confusing. This commit sets the namespace on the config object. If we ever introduce support for CRDs, hopefully most code won't have to change.

**appliance: receive and store requested version**

The admin flows we've been talking about as a team decouple appliance installation from SG upgrades. Therefore a given appliance version, in general, needs to be able to deploy some range of supported SG versions, in order to avoid "surprise" unrequested upgrades when the appliance self-updates (another feature that we'd been discussing). Therefore, the Sourcegraph config object (derived from the ConfigMap), should contain a version field. We can imagine this being injected down by the appliance web app.

In general, a given SG service might have different resource sets across different versions, so the operator might need to do branching logic depending on what version is requested.

More speculatively: service-specific upgrade logic _might_ want to know the previously deployed version, e.g. to orchestrate DB MVUs. The Sourcegraph config object contains a version in its Status, an artifact of when this operator was to be CRD-driven. Similar to the commit above, we can use this field in go, but must drive it using a mechanism other than k8s Status updates. This commit makes use of a ConfigMap annotation to actually persist the last version, since ConfigMaps don't have Status fields in k8s.

**appliance: add some standard labels to default deployment**

Notably the version, which might be useful for debugging. Label keys
lifted from the helm chart.

This is definitely a bit speculative, and to be honest I'm quite unsure how similar we want the helm and appliance configs to be. Maybe we want to take the opportunity to remove some of these labels? I'm really not sure, and only made this commit to start a conversation! We can remove it if wanted.

**appliance: deploy blobstore deployment using new reconciliation logic**

In Kubernetes, when an owned object (e.g. one of our Deployments) changes, an event is sent to the controller _for the owning resource_ (our ConfigMap). If the controller then updates the same object in response _unconditionally_, this starts an infinite tight loop. There must be conditions in which the controller returns without updating the resource. Therefore, we need to write logic to detect when the object doesn't need updating.

We do want to reconcile away out-of-band updates to certain fields (e.g. the container command), but probably not for all of them. A common example is Deployment.spec.replicas. Often, this field is de-facto managed by a horizontal pod autoscaler (HPA). We do not want to clobber HPA scaling events back every time they occur. This commit introduces a generic createOrUpdateObject function that will create a kubernetes object if it doesn't exist, and as one of its parameters, takes a function that returns a "masked" object from one that already exists in the cluster. This masked object is hashed, and that hash is compared to one calculated from the object the reconciler thinks should be in the cluster. In this way we can only update objects with salient changes.

I've illustrated this by using the new function for the Blobstore Deployment, to get feedback on the idea before adding/changing more resources.

---

## Test plan

The appliance is still incomplete and very new, and useful testing of such glue/deployment code is a tricky topic. I can think of a few possibilities, which aren't mutually exclusive - we can potentially do all/some of them:

1. @jdpleiness (IIRC) mentioned golden testing (a.k.a snapshot testing). We can separate the functions that generate objects from the ones that deploy them to clusters, and wrap golden tests around those.
2. Similar to above, but using Kubernetes `envtest` to assert against the contents of an actual k8s control plane, in-memory, but not backed by controllers (so no container runtime is needed).
3. Smoke testing against actual kubernetes deployments. This should be doable in a CI pipeline using k3d or similar

I propose that we split out an issues in https://github.com/sourcegraph/sourcegraph/issues/61556 or elsewhere to discuss these ideas, rather than make them part of this PR. No appliance code is actually run anywhere, yet.


craigfurman:
@jdpleiness I've updated this based on what we talked about yesterday. The testing piece is large enough that I reckon we should split it. I'll mark this as ready once your branch is merged, and this is rebased on main. Holler at me if you have any feedback before then!

internal/appliance/kubernetes.go
```diff
@@ -0,0 +1,100 @@
...
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
```
jdpleiness, line 14:
```suggestion
	"context"
	"crypto/sha256"
	"encoding/hex"
	"encoding/json"

	
	kerrors "k8s.io/apimachinery/pkg/api/errors"
	"k8s.io/apimachinery/pkg/types"
	ctrl "sigs.k8s.io/controller-runtime"
	"sigs.k8s.io/controller-runtime/pkg/client"
	"sigs.k8s.io/controller-runtime/pkg/log"
	
	"github.com/sourcegraph/sourcegraph/lib/errors"
```
Super small nit

jdpleiness (approved):

internal/appliance/kubernetes.go
craigfurman: Oh sure, no worries, accepted. I'll configure my editor to do this on auto-format. Neovim users might be interested to see how I overengineered exactly this a couple of years ago 😂 

https://github.com/craigfurman/ansible-home/blob/28e26492f3e11d1783273f851cd070bced34b28e/config/home/.config/nvim/lua/lang/go.lua#L12

Then in relevant go projects, a `.envrc` file loaded by `direnv` contains: `export GOPLS_SETTINGS_LOCAL=github.com/sourcegraph`

PR #61933, opened 2024-04-16
Automatically check for updates to packaged dependencies

willdollman:
Automatically update wolfi package dependencies.

Currently our packaged dependencies in [`wolfi-packages/`](https://github.com/sourcegraph/sourcegraph/tree/main/wolfi-packages) are updated very infrequently, as this requires someone to manually check for an upstream update, validate the new hash, update the config, and open a PR.

After [adding the relevant metadata](https://github.com/sourcegraph/sourcegraph/pull/61935) to each package config, we can use `wolfictl` to automatically check for updates and open relevant PRs. This runs in a daily Github action.

- [x] Add upstream sources to all wolfi packages so that new versions can be automatically identified using `wolfictl`
- [x] Create periodic github action to open update PRs

## Links for reviewers

- [Example of action running](https://github.com/sourcegraph/sourcegraph/actions/runs/8725592189/job/23938910606)
  - Several package updates failed - these is fixed in https://github.com/sourcegraph/sourcegraph/pull/61981
- [PRs opened by this action](https://github.com/sourcegraph/sourcegraph/pulls?q=is:pr+is:open+label:security-auto-update/packages)
  - Unfortunately there's no way to customise the contents of these PRs. We could potentially submit an upstream PR to change this.

## Test plan

- [x] Manually run `wolfictl update` locally to test `update:` metadata
- [x] Manually trigger github action to ensure it works as expected


.github/workflows/wolfictl-update.yml
```diff
@@ -0,0 +1,79 @@
...
+
+      # NOTE: The main wolfictl binary has a bug which prevents us from using it
+      # Once https://github.com/wolfi-dev/wolfictl/pull/763 is merged, we can remove the following which
+      # builds wolfictl from a patched fork.
+
+      # Build and install a fork of wolfictl
+      - name: Check out code from external repository
+        uses: actions/checkout@v4
+        with:
+          repository: 'willdollman/wolfictl'
```
willdollman, line 30: Discussed that we're using a personal fork with @evict - we think this is ok for now due to the comments, but happy to get other opinions.

.github/workflows/wolfictl-update.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+         go-version: '1.22.2'
+
+      - name: Build binary
+        run: |
+          cd $GITHUB_WORKSPACE/wolfictl
+          ls -al ./
+          go build -o wolfictl ./
```
burmudar, line 44: Just thinking out loud but wondering if we can do `GOBIN=<the path> go install willdoman/wolfictl@the-patch` and then to the whole adding to the path? That simplifies this to a few steps 🤔

.github/workflows/wolfictl-update.yml
```diff
@@ -0,0 +1,79 @@
+name: 'wolfictl-update'
+
+on:
+  schedule:
+    # everyday at 0:00 UTC
+    - cron:  0 0 * * *
```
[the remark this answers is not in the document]
willdollman: cronlife

.github/workflows/wolfictl-update.yml
willdollman: `go install` doesn't support remote references afaik, but this is a good idea - I've simplified installation based on this idea. Thanks!

.github/workflows/wolfictl-update.yml
```diff
@@ -0,0 +1,79 @@
...
+
+      - name: Add wolfictl to path
+        run: |
+          TMP=$(mktemp -d)
+          echo "Temp dir: $TMP"
+          cd $GITHUB_WORKSPACE/wolfictl
+          ls -al ./
+          mv ./wolfictl $TMP/
+          ls -al $TMP
+          echo "$TMP" >> $GITHUB_PATH
```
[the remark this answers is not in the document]
willdollman: Tidied up

burmudar (approved):

willdollman:
Tested workflow after applying review feedback: https://github.com/sourcegraph/sourcegraph/actions/runs/8735785292

PR #61934, opened 2024-04-16
telemetry: unexport internal type, add SafeMetadataKey

bobheadxi:
I noticed the `telemetry.constString` type was wrongly exported in a previous PR (https://github.com/sourcegraph/sourcegraph/pull/61134). This type is intentionally unexported to prevent the exact use case the PR used the exported type for - this scenario (list keys and include it in metadata) is exactly the scenario where something bad might unexpectedly show up in our telemetry (an unexpected Redis key is matched, an unexpected value is provided to be stored, etc - the cause of multiple data leakage incidents in the past)

However, given the format I was unable to quickly come up with a better alternative. Moving it to private metadata is also not possible, as because judging from the description of #61134 we need this to be exported from on-prem as a business need.

For now:

1. Add loud warnings on the various intentionally-internal types to not export them
2. Add a constructor `telemetry.SafeMetadataKey(...)` to support the current use case, with warnings - callers more or less "sign off" on owning problems here if there are any. In the future if we have any regex strategy we want to apply we can add them to these constructors as well.
3. Add CODEOWNERS to make sure changes to telemetry do not go without adequate review

## Test plan

CI

### DIFF
--- a/internal/telemetrygateway/CODEOWNERS
+++ b/internal/telemetrygateway/CODEOWNERS
@@ -1,0 +1,2 @@
+**/* @sourcegraph/core-services
+**/* @sourcegraph/data-team
--- a/cmd/frontend/internal/bg/store_token_usage_in_postgres.go
+++ b/cmd/frontend/internal/bg/store_token_usage_in_postgres.go
@@ -32,8 +32,7 @@ func storeTokenUsageinDb(ctx context.Context, db database.DB) error {
 	}
 	convertedTokenUsageData := make(telemetry.EventMetadata)
 	for key, value := range tokenUsageData {
-		castedKey := telemetry.ConstString(key) // Cast the key to telemetry.ConstString
-		convertedTokenUsageData[castedKey] = value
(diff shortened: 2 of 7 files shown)


dadlerj (approved):
Thanks Robert

akalia25 (approved):
Thanks Robert!

internal/telemetrygateway/CODEOWNERS
eseliger, line 1: Is this a new feature in GitHub?

GitHub used to honor only exactly one Codeowners file 🤔

bobheadxi: good question... I saw that there was a subdir one in https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/third-party-licenses/CODEOWNERS?L1

bobheadxi: Ah, I think I actually want `CODENOTIFY` - got the names mixed up 😆

PR #61938, opened 2024-04-16
gitserver: Migrate Diff to gRPC call

eseliger:
This PR migrates the Diff method to use a proper gRPC method. I decided to call it RawDiff as it streams back the raw diff format.

Later, we can decide to also add a structured endpoint, but for now we'll let the caller decide what to do with the diff string. There's probably a use-case for both versions.

Closes https://github.com/sourcegraph/sourcegraph/issues/61868

Test plan:

Added tests across all the layers, verified locally that the diff view works.


cmd/gitserver/internal/git/iface.go
```diff
@@ -80,13 +79,31 @@ type GitBackend interface {
 	// If no commit exists in the history of revspec before time, an empty
 	// commitID is returned.
 	RevAtTime(ctx context.Context, revspec string, time time.Time) (api.CommitID, error)
+	// ReadDiff returns the raw git diff for the given range.
+	// Diffs returned from this function will have the following settings applied:
+	// - 3 lines of context
+	// - No a/ b/ prefixes
+	// - Rename detection
+	// If either base or head don't exist, a RevisionNotFoundError is returned.
+	ReadDiff(ctx context.Context, base string, head string, typ GitDiffComparisonType, paths ...string) (io.ReadCloser, error)
```
mmanela, line 88: You name things RawDiff everywhere else, why here do we switch to `ReadDiff`.

cmd/gitserver/internal/server_grpc.go
```diff
@@ -1310,6 +1310,81 @@ func (gs *grpcServer) ListRefs(req *proto.ListRefsRequest, ss proto.GitserverSer
...
+			return s.Err()
+		}
+		gs.svc.LogIfCorrupt(ctx, repoName, err)
+		// TODO: Better error checking.
+		return err
+	}
+	defer r.Close()
+
+	w := streamio.NewWriter(func(p []byte) error {
+		return ss.Send(&proto.RawDiffResponse{Chunk: p})
```
mmanela, line 1381: What is in each chunk? Is this determined by git?

internal/gitserver/v1/gitserver.proto
```diff
@@ -177,6 +177,51 @@ service GitserverService {
...
+  // is returned.
+  //
+  // If the given repo is not cloned, it will be enqueued for cloning and a
+  // NotFound error will be returned, with a RepoNotFoundPayload in the details.
+  rpc RawDiff(RawDiffRequest) returns (stream RawDiffResponse) {
+    option idempotency_level = NO_SIDE_EFFECTS;
+  }
+}
+
+message RawDiffRequest {
```
mmanela, line 196: So RawDiff means just a stream of the raw git bytes of the diff and you are thinking in the future we could have a structure diff where we parse it first. 

Is there a use-case where someone would not want us to pre-parse it into an object model?

cmd/gitserver/internal/git/iface.go
eseliger: ehm.. yeah I must've messed that up somewhere on the way - will align them

Thanks for catching!

cmd/gitserver/internal/server_grpc.go
eseliger: Arbitary, the chunking is determined by the streamio package that decides how big the individual chunks are. The sole reason for that is that we don't have to send 250MB in one proto message which is not just failing, but also will require 250MB of memory to marshal.

The caller should receive all of the packets and stitch them together at will. 
There's a streamio.Reader equivalent that'll do that for you on the other end of the wire

internal/gitserver/v1/gitserver.proto
eseliger: Yep -- we have a feature similar to https://patch-diff.githubusercontent.com/raw/sourcegraph/sourcegraph/pull/61938.diff in batch changes for example. I.e., something out of our control is really just interested in the pure patch.

pjlast (approved):

Issue #60739, opened 2024-02-23
Code Insights: change the behavior of drilldown


camdencheek:
When drilling down into a code insights data point, we generate a diff search for the difference between that data point and the point before it. This is not ideal for many reasons:
1) It is somewhat confusing semantics. You don't see what the data point represents, you only see the diff
2) Many code insights searches cannot actually be represented as a diff search
3) The query rewriting has been the source of a bunch of different bugs
4) Diff search is slow

Instead, we want to change the behavior to drill down into a search at the commit that data point represents.
```[tasklist]
### Related issues
- [ ] https://github.com/sourcegraph/sourcegraph/issues/60686
- [ ] https://github.com/sourcegraph/sourcegraph/issues/58906
```

camdencheek:
I started looking into this today because I was hopeful it was going to be a relatively easy change. 

_Narrator_: It will not be a relatively easy change.

Attempting to capture what I've found so far below:

- The goal is to make clicking on an insights data point generate a search query that yields the set of results that correspond to the data point that is clicked. So if the data point shows 30, our search should have 30 results.
- This is different from the current behavior, which attempts to generate a diff search that yields the _difference_ in the results between a data point and its previous data point.
- In order to generate a single query that yields all the results represented, we would need to generate a query that looks like `(repo:repo1@abcde OR repo:repo2@bcdef) needle` because we need to query each repo in the set at a specific revision corresponding to that point in time, and we do not support anything like `rev:head.at(1996-06-28)` (though that would be really cool!)
- We generate the diff query for each point on the backend with [PointDiffQuery](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@4a87b94e63fe9be68457b37cdc7c56b941406f55/-/blob/internal/insights/query/querybuilder/builder.go?L145-145)
- The data we use to generate the diff query ultimately comes from the [series_points](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@4a87b94e63fe9be68457b37cdc7c56b941406f55/-/blob/internal/database/schema.codeinsights.md?L15-31) table, which stores points at timestamps partitioned by repo, but does _not_ store the commit that was searched along with it.

So, to implement this change, we would need to:
[...]

PR #61953, opened 2024-04-16
Code Insights: use point-in-time search for drilldown

camdencheek:
This implements a change in Code Insights drilldown behavior to show the set of results represented by that point rather than a diff search between the timestamps of the two points. This is built on top of the new `rev:at.time()` filter. 

Closes https://github.com/sourcegraph/sourcegraph/issues/60739

## Test plan

Copied all tests we already had for diffQuery. Manual end-to-end test which is shown in the attached video.

https://github.com/sourcegraph/sourcegraph/assets/12631702/4e60e768-e0b7-40f9-869c-a1105a0259b6


internal/insights/query/querybuilder/builder.go
```diff
@@ -20,6 +20,10 @@ import (
 // by fully parsing and reconstructing the tree, and does **not** overwrite user supplied values for the default fields.
 // This also converts count:all to count:99999999.
 func withDefaults(inputQuery BasicQuery, defaults searchquery.Parameters) (BasicQuery, error) {
+	if len(defaults) == 0 {
+		return inputQuery, nil
+	}
```
camdencheek, line 25: No need to do a parse/stringify round trip if there are no defaults to apply. This is actually meaningful because parsing a `rev:` query without adding the `repo:` parts is invalid, which causes errors.

internal/insights/query/querybuilder/builder_test.go
```diff
@@ -913,3 +907,117 @@ func TestPointDiffQuery(t *testing.T) {
 		})
 	}
 }
+
+func TestPointTimeQuery(t *testing.T) {
```
camdencheek, line 911: Copied directly from `TestPointDiffQuery`

peterguy (approved):
tested locally; looks great

internal/insights/query/querybuilder/builder.go
```diff
@@ -219,6 +223,69 @@ func PointDiffQuery(diffInfo PointDiffQueryOpts) (BasicQuery, error) {
 	return BasicQuery(query), nil
 }
 
+func PointInTimeQuery(diffInfo PointDiffQueryOpts) (BasicQuery, error) {
```
vovakulikov, line 226: I don't think it's a regression since this code works in the same way as the diffQuery resolver (maybe it happened before, but demo staging works fine, not sure), but it looks like now it's possible to break the query that already has a repo filter. 

I checked this on demo staging and this works in the following way
- If the insight's original query has a `repo:` filter with specific repositories we just add post-process filters to the original repo filters
- if the insight's original query has a `repo:.*` filter when we add some post-process filters we just use these filters as repo: filter with no `repo:.*` in the result query

After this PR, I think the second case breaks; if you add repository post-filters to the insight with `repo:*` in the repo query, the result query will be `repo:* repo:<post-process-repo-filter-value>` which suddenly breaks the result for some reason. 

To clarify when I'm talking about post-process filters I mean these filters form UI under the filter button in the insight card on the dashboard 

Screenshot 2024-04-23 at 22 39 24

cc @camdencheek

vovakulikov (commented):
@camdencheek my apologies that I couldn't check this PR earlier, left some comments after manual testing, maybe I checked it wrongly but worth double-checking with you I think.

PR #61958, opened 2024-04-17
gitserver: Convert ContributorCounts to gRPC

eseliger:
This PR migrates the ContributorCounts client API to gRPC. The method was previously untested, so I added a bunch of tests there.

Test plan:

New tests on all the involved layers, no existing ones broke.


cmd/gitserver/internal/git/iface.go
```diff
@@ -175,3 +180,14 @@ type RefIterator interface {
 	// Close releases resources associated with the iterator.
 	Close() error
 }
+
+// ContributorCountsOpts are options for the ContributorCounts method.
+type ContributorCountsOpts struct {
+	// If set, only count commits that are in the given range.
+	Range string
```
pjlast, line 187: Example of range format?

pjlast (approved):

cmd/gitserver/internal/git/iface.go
```diff
@@ -175,3 +180,14 @@ type RefIterator interface {
 	// Close releases resources associated with the iterator.
 	Close() error
 }
+
+// ContributorCountsOpts are options for the ContributorCounts method.
+type ContributorCountsOpts struct {
+	// If set, only count commits that are in the given range.
```
eseliger, line 186:
```suggestion
	// If set, only count commits that are in the given range.
	// Range can contain:
	// - A commit hash or ref name, which includes that commit/ref and all of the parents
	// - A range of two commits/hashes, separated by either .. or ... notation: A..B, A...B
```

cmd/gitserver/internal/git/iface.go
eseliger: Added!

PR #61960, opened 2024-04-17
Cody Gateway: Block request if user limit is 1

vdavid:
- Closes https://github.com/sourcegraph/self-serve-cody/issues/713

Just added the blocking logic.

Considered adding a different error message (`return nil, errors.Wrap(NoAccessError{}, "blocked")`) because it'd help us debug, but it seems to make more sense to make these responses less helpful in this case.

Update: now it's just a "-1" change and some comments.

Update: reverted to the "if 1 then let's say it's 0" logic

[Related Slack thread](https://sourcegraph.slack.com/archives/C06062P5TS5/p1713483012837119?thread_ts=1712856401.820429&cid=C06062P5TS5).

## Test plan

1. Make sure the tests are green (done)
2. Make sure the change is applied properly, by adding an automated test [here](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/cmd/cody-gateway/internal/limiter/limiter_test.go?L143) that triggers the new behavior.
3. Dig a bit in the code (or perhaps consult Joe, might be quicker) to understand how the concurrency limiter works, because it internally uses the static limiter that this PR changes. If the concurrency limiter is guaranteed to never use the static limiter with limit=1, then great, nothing to do. If it does, then the static limiter probably needs to identify the case when the concurrency limiter uses it (could be through an existing or a new argument) and disable the new behavior in those cases.
4. Make sure the limiting works as intended by manually reproducing the limit == 1 and limit != 1 cases (clues from Joe for running Cody Gateway and SSC at the same time without clashing on port `9992`: "On both sides, it should be just a matter of changing the PORT env var in your .envrc/sg.config.yaml , changing on the CG side is probably less side effect since changing on SAMS/SSC side would need to change some URLs (also in .envrc)"


chrsmith (approved):
Looks good to me. 🤞

vdavid:
@chrsmith, unfortunately, not to CI. The concurrency limiter also [relies on the static limiter](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/cmd/cody-gateway/internal/actor/limits.go?L78), and with the changed logic, it breaks in our tests, and I suspect that it'll also break in production in ways that I find difficult to predict.

With this finding, I'd rather go with my original solution. Also, `NoAccessError` does make sense to me, and that's what we [already do](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/cmd/cody-gateway/internal/limiter/limiter.go?L85) in the case of limit == 0.

vdavid:
Update:
I ran out of time and can't test this one.

What I'd do is:
1. Make sure the tests are green (done)
2. Make sure the change is applied properly, by adding an automated test [here](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/cmd/cody-gateway/internal/limiter/limiter_test.go?L143) that triggers the new behavior.
3. Dig a bit in the code (or perhaps consult Joe, might be quicker) to understand how the concurrency limiter works, because it internally uses the static limiter that this PR changes. If the concurrency limiter is guaranteed to never use the static limiter with limit=1, then great, nothing to do. If it does, then the static limiter probably needs to identify the case when the concurrency limiter uses it (could be through an existing or a new argument) and disable the new behavior in those cases.
4. Make sure the limiting works as intended by manually reproducing the limit == 1 and limit != 1 cases (clues from Joe for running Cody Gateway and SSC at the same time without clashing on port `9992`: "On both sides, it should be just a matter of changing the PORT env var in your .envrc/sg.config.yaml , changing on the CG side is probably less side effect since changing on SAMS/SSC side would need to change some URLs (also in .envrc)"

PR #61968, opened 2024-04-17
Create a shared Cody Ignore dataset

taras-yemets:
To ensure consistency in how Cody context filters are handled across
both the server and clients, this PR:
- moves existing enterprise Cody context filter test cases to a separate JSON file
- creates NPM package exporting a JSON file containing the copy of the mentioned test cases
  - dataset file is generated on the `prepublishOnly` step and is ignored by Git
  - package is excluded from the workspace to ensure the published version of the test dataset is used 
- refactors Cody Web tests to consume test cases from the NPM package.

More context in [this](https://sourcegraph.slack.com/archives/C06Q9R098N4/p1712802266879379) thread.

TODO:
- [ ] Release `1.0.0` version after PR is approved

Follow-up work:
- [ ] Add GitHub action commenting on the PR changing the original test dataset suggesting to publish a new package version

## Test plan
- CI


keegancsmith (commented):
Convention in go to put test data in a sub directory called `testdata`.

From slack DM about motivation for this change:

> ... publish npm package with fixtures ...

I'd consider removing this step and instead just directly fetching the test fixture from github. It has a very simple API to download a file? You can then fetch the file by the sourcegraph version tag.

taras-yemets:
Thank you, @keegancsmith! 

> > ... publish npm package with fixtures ...
> 
> I'd consider removing this step and instead just directly fetching the test fixture from github. It has a very simple API to download a file? You can then fetch the file by the sourcegraph version tag.

Using this approach, if we conduct tests on this shared dataset in Cody clients' CI, we might encounter situations where the CI fails due to unrelated code changes. I previously suggested this approach, but [it was not supported](https://github.com/sourcegraph/cody/pull/3860#pullrequestreview-2009987005) with it.

keegancsmith:
> > I'd consider removing this step and instead just directly fetching the test fixture from github. It has a very simple API to download a file? You can then fetch the file by the sourcegraph version tag.
> 
> Using this approach, if we conduct tests on this shared dataset in Cody clients' CI, we might encounter situations where the CI fails due to unrelated code changes. I previously suggested this approach, but [it was not supported](https://github.com/sourcegraph/cody/pull/3860#pullrequestreview-2009987005) with it.

In that PR you mention you want to use HEAD. I agree that this is not correct, you do not want outside changes to affect your CI. You should use commits or even better sourcegraph version tags. In that case it won't have the issue dominik raises.

The point about NPM and renovate is a good one. However, I would double check that we have automation to publish packages from sourcegraph automatically. I imagine this is quite complicated but I may be mistaken. So the only real reason to use NPM would then be renovate bot picking this up automatically in client repos once you manually publish. This seems reasonable.

My gut tells me that the approach you took is the one with less moving parts. Things like automatic updates don't seem that important to me, since changing things like the ignore file format is a big deal, and will involve manual work in the respective client repos.

So I would just double check again, since the main objection was sound about using an mutable fixture source instead of immutable.

taras-yemets:
> In that PR you mention you want to use HEAD. I agree that this is not correct, you do not want outside changes to affect your CI. You should use commits or even better sourcegraph version tags.

Both in [PR](https://github.com/sourcegraph/cody/pull/3860/files#diff-e8dc4ca3d61cc6c44d4a2319cb3e4b05d72436bfee3086dc2b5e9d4fbf579f0cR101) and in [Slack](https://sourcegraph.slack.com/archives/C06Q9R098N4/p1713448706268309?thread_ts=1712802266.879379&cid=C06Q9R098N4) I suggested using tags (maybe I should have been more explicit about it). My understanding is that the packaging approach is preferable for the clients team. 

> I would double check that we have automation to publish packages from sourcegraph automatically. I imagine this is quite complicated

That's my feeling too. At the current stage, we are considering publishing new versions manually.

taras-yemets:
@dominiccooney, @valerybugakov, can you please give some early feedback on the approach (in particular, [package](https://github.com/sourcegraph/sourcegraph/pull/61968/files#diff-ab3e7e3af8200a1d6d8824b5a683ef3ab37e6c7527f08a6ee6b3f7de6a2fa9a5) structure and its [usage](https://github.com/sourcegraph/sourcegraph/pull/61968/files#diff-962f6cbc5831ec1afffaac0baba242d32910a0d719fd5264eca9b8bf7a6d0117))? Would it be convenient to integrate with Cody clients codebases in your opinion?

client/cody-context-filters-test-dataset/dataset.json
```diff
@@ -0,0 +1,717 @@
+{
+  "testCases": [
+    {
+      "name": "Cody context filters are not defined",
```
valerybugakov, line 4: Can we remove the dataset duplication to ensure they won't be updated independently? Can we copy it during the `pnpm generate` step from the go package?

client/cody-context-filters-test-dataset/dataset.json
```diff
@@ -0,0 +1,717 @@
...
+      "description": "Any repo should be included.",
+      "includeByDefault": true,
+      "includeUnknown": false,
```
valerybugakov, line 7: Are these properties required for the backend tests? AFAIU these are the only possible values based on the feature spec.

valerybugakov (commented):
The approach looks good to me. Left a couple of questions inline.

client/cody-context-filters-test-dataset/dataset.json
taras-yemets: I added these fields mostly for clients use. I agree that we have a single (spec-defined) way to treat defaults and unknowns (`true` and `false` respectively).
WDYT, @dominiccooney?

valerybugakov: We might need them in the future if the spec changes, but they are redundant now. I would remove them to avoid confusion, as seeing them made me wonder if we have documented cases where these properties can have other values.

dominiccooney (approved):
Thank you!

client/cody-context-filters-test-dataset/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "version": "0.0.1",
+  "license": "Apache-2.0",
+  "description": "Cody context filters test dataset",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/sourcegraph/sourcegraph.git",
+    "directory": "client/cody-context-filters-test-dataset"
+  },
+  "scripts": {
+    "prepublishOnly": "node scripts/generate.js"
```
valerybugakov, line 12: 🧠

valerybugakov (approved):
Beautiful! Thank you 🙌

PR #61984, opened 2024-04-17
Remove path verification for archive reader

pjlast:
Removes path verification for ArchiveReader calls since there are edge cases that we're not handling correctly.

## Test plan

Manual tests that `src batch preview` works now.


eseliger (approved):
I'm assuming tests will fail now

pjlast:
@eseliger yep! Removed the tests as well. Will see if pipeline passes

ggilmore (approved):

PR #61989, opened 2024-04-17
Svelte [RepoPopover]: Instantiate the RepoPopover result across the web-app

jasonhawkharris:
This commit includes optimizations and improvements for the RepoPopover, in addition to instantiating the component on the Repo search results page. Not much has changed on the design from the implementation PR. Here is a video to show some live interaction: 

https://github.com/sourcegraph/sourcegraph/assets/62355966/534a8119-fc83-4eeb-bab9-0ce8b7800524

1. @fkling I'd like to get your thoughts on how I've pulled the data in for this. @camdencheek mentioned you might have thoughts on this approach. 

## Test plan
manual, visual test, storybooks, unit tests.

### DIFF
--- a/client/web-sveltekit/src/lib/Popover.svelte
+++ b/client/web-sveltekit/src/lib/Popover.svelte
@@ -9,11 +9,13 @@
      * Show the popover when hovering over the trigger.
      */
     export let showOnHover = false
+    export let delay: number | null = null
 
     let isOpen = false
     let trigger: HTMLElement | null
     let target: HTMLElement | undefined
     let popoverContainer: HTMLElement | null
+    let timeoutId: NodeJS.Timeout | undefined
 
     function toggle(open?: boolean): void {
         isOpen = open === undefined ? !isOpen : open
@@ -33,10 +35,21 @@
         trigger = node
 
         function handleMouseEnterTrigger(): void {
-            isOpen = true
+            if (delay) {
+                timeoutId = setTimeout(() => {
+                    isOpen = true
+                }, delay)
+            } else {
+                isOpen = true
+            }
         }
 
         function handleMouseLeaveTrigger(event: MouseEvent): void {
+            // We have to clear the timeout here, otherwise it would be triggered
+            // even if the mouse leaves the trigger before the delay is over.
+            if (delay) {
+                clearTimeout(timeoutId)
+            }
             // It should be possible to move the mouse from the trigger to the popover without closing it
             if (event.relatedTarget && !popoverContainer?.contains(event.relatedTarget as Node)) {
                 isOpen = false
--- a/client/web-sveltekit/src/lib/Popover.svelte
+++ b/client/web-sveltekit/src/lib/Popover.svelte
@@ -1,8 +1,8 @@
 <script lang="ts">
     import type { Placement } from '@floating-ui/dom'
+    import type { Action } from 'svelte/action'
 
     import { popover, onClickOutside, portal } from './dom'
-    import type { Action } from 'svelte/action'
 
     export let placement: Placement = 'bottom'
     /**
@@ -106,9 +106,9 @@
         font-size: 0.875rem;
         background-clip: padding-box;
         background-color: var(--dropdown-bg);
+        color: var(--body-color);
+        box-shadow: var(--dropdown-shadow);
         border: 1px solid var(--dropdown-border-color);
         border-radius: var(--popover-border-radius);
-        color: var(--body-color);
-        box-shadow: var(--popover-shadow);
     }
 </style>
--- a/client/web-sveltekit/src/lib/Popover.svelte
+++ b/client/web-sveltekit/src/lib/Popover.svelte
@@ -80,6 +80,7 @@
 <slot {toggle} {registerTrigger} {registerTarget} />
 {#if trigger && isOpen}
     <div
+        use:registerPopoverContainer
         use:portal
         use:onClickOutside
         use:registerPopoverContainer
@@ -108,7 +109,7 @@
         background-color: var(--dropdown-bg);
         color: var(--body-color);
         box-shadow: var(--dropdown-shadow);
-        border: 1px solid var(--dropdown-border-color);
+        border: 0;
         border-radius: var(--popover-border-radius);
     }
 </style>
--- a/client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
+++ b/client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
@@ -180,14 +180,6 @@ For example:
             flex-flow: row wrap;
             gap: 0.5rem 0.5rem;
             justify-content: flex-start;
-
-            small {
-                background-color: var(--subtle-bg);
-                border-radius: 1rem;
-                color: var(--primary);
-                font-family: var(--monospace-font-family);
-                padding: 0.15rem 0.35rem;
-            }
         }
     }
 
--- a/client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
+++ b/client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
@@ -78,7 +78,6 @@ For example:
             <div class="commit-info">
                 <div class="commit">
                     <small class="subject">{commit.subject}</small>
-                    <small>{CENTER_DOT}</small>
                     <small class="commit-number"
                         ><a href={commit.canonicalURL} target="_blank">
                             {commit.abbreviatedOID}
--- a/client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
+++ b/client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
@@ -16,7 +16,6 @@ For example:
 
     import Avatar from '$lib/Avatar.svelte'
     import Icon from '$lib/Icon.svelte'
-    import LoadingSpinner from '$lib/LoadingSpinner.svelte'
     import Timestamp from '$lib/Timestamp.svelte'
 
     import RepoStars from '../RepoStars.svelte'
@@ -36,6 +35,14 @@ For example:
         return numStr.substring(numStr.length - length)
     }
 
+    function formatRepoName(repoName: string): string {
+        const slashes = repoName.split('/')
+        let repo = slashes.pop()
+        let org = slashes.pop()
+        return `${org} / ${repo}`
+    }
+
+    $: firstFiveTags = repo.tags.nodes.slice(0, 5)
     $: subject = repo.commit?.subject
     $: url = repo.commit?.canonicalURL
     $: author = repo.commit?.author.person.name
@@ -52,7 +59,9 @@ For example:
             <div class="icon-name-access">
                 <!-- @TODO: We need to use our customer's logo here. mdiAlienOutline is a place holder-->
                 <Icon svgPath={mdiAlienOutline} --color="var(--primary)" />
-                <h4 class="repo-name">{repo.name}</h4>
+                <div>
+                    <h4 class="repo-name">{formatRepoName(repo.name)}</h4>
+                </div>
                 <small>{repo.isPrivate ? 'Private' : 'Public'}</small>
             </div>
             <div class="code-host">
@@ -63,16 +72,18 @@ For example:
         <div class="divider" />
     {/if}
 
-    <div class="description-and-tags">
-        <div class="description">
-            {repo.description}
+    {#if repo.description || firstFiveTags.length > 0}
+        <div class="description-and-tags">
+            <div class="description">
+                {repo.description}
+            </div>
+            <div class="tags">
+                {#each firstFiveTags as tag}
+                    <small>{tag.name}</small>
+                {/each}
+            </div>
         </div>
-        <div class="tags">
-            {#each repo.tags.nodes as tag}
-                <small>{tag.name}</small>
-            {/each}
-        </div>
-    </div>
+    {/if}
 
     <div class="divider" />
 
@@ -107,8 +118,6 @@ For example:
 
 <style lang="scss">
     .root {
-        border: 1px solid var(--border-color);
-        border-radius: var(--popover-border-radius);
         width: 480px;
     }
 
@@ -164,9 +173,6 @@ For example:
 
         .description {
             padding: 0rem;
-            text-overflow: ellipsis;
-            overflow: hidden;
-            white-space: nowrap;
         }
 
         .tags {
@@ -176,14 +182,15 @@ For example:
             flex-flow: row wrap;
             gap: 0.5rem 0.5rem;
             justify-content: flex-start;
-            margin-top: 0.5rem;
+            margin-top: 1rem;
 
             small {
                 background-color: var(--subtle-bg);
                 border-radius: 1rem;
                 color: var(--primary);
                 font-family: var(--monospace-font-family);
-                padding: 0.25rem 0.5rem;
+                padding: 0.15rem 0.35rem;
+                font-size: 10px;
             }
         }
     }
@@ -225,6 +232,7 @@ For example:
                 .commit-number {
                     color: var(--text-muted);
                     align-self: center;
+                    font-family: var(--monospace-font-family);
                 }
             }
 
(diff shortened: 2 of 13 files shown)


dev/update-embeddings-list/tsconfig.json
```diff
@@ -37,7 +37,7 @@
     // "types": [],                                      /* Specify type package names to be included without being referenced in a source file. */
     // "allowUmdGlobalAccess": true,                     /* Allow accessing UMD globals from modules. */
     // "moduleSuffixes": [],                             /* List of file name suffixes to search when resolving a module. */
-    // "allowImportingTsExtensions": true,               /* Allow imports to include TypeScript file extensions. Requires '--moduleResolution bundler' and either '--noEmit' or '--emitDeclarationOnly' to be set. */
+    "allowImportingTsExtensions": true,               /* Allow imports to include TypeScript file extensions. Requires '--moduleResolution bundler' and either '--noEmit' or '--emitDeclarationOnly' to be set. */
```
camdencheek, line 40: Why did this change?

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -20,94 +21,94 @@ For example:
...
-    $: author = repo.commit?.author.person.name
-    $: commitDate = repo.commit?.author.date
-    $: avatar = repo.commit?.author.person
-    $: codeHostKind = repo.externalServices.nodes[0].kind
+    function formatRepoName(repoName: string): string {
+        const slashes = repoName.split('/')
+        let repo = slashes.pop()
+        let org = slashes.pop()
+        return `${org} / ${repo}`
+    }
```
camdencheek, line 44: What happens for repos that are not named like `github.com/sourcegraph/sourcegraph`? e.g., many enterprise customers have repos named just like `myrepo`

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -20,94 +21,94 @@ For example:
...
+
+    $: commit = repo.commit
+    $: author = commit?.author
+    $: avatar = author?.person
+    $: access = repo.isPrivate ? 'Private' : 'Public'
+    $: codeHostKind = capitalize(repo.externalServices.nodes[0].kind)
     $: codeHostIcon = getIconPathForCodeHost(codeHostKind)
+    $: abbreviatedCommitSHA = truncateCommitNumber(commit?.oid, 6)
```
camdencheek, line 52: naming nit: "commit number" doesn't reall mean anything. The terms "commit OID" or "commit SHA" are more precise

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -20,94 +21,94 @@ For example:
...
-                <!-- @TODO: We need to use our customer's logo here, not the code host's -->
-                <!--Icon svgPath={mdiGitlab} /-->
-                <h4 class="repo-name">{repo.name}</h4>
-                <div class="access">
-                    <small>{repo.isPrivate ? 'Private' : 'Public'}</small>
-                </div>
+            <div class="left">
+                <Icon svgPath={orgSVGPath} --color="var(--primary)" />
+                <h4>{formatRepoName(repo.name)}</h4>
+                <small>{access}</small>
```
camdencheek, line 61: nit: I prefer the inlined version of `access` here. I don't really know what the variable `access` means, so I have to go look up its definition. And the definition is small enough to just put in the markup directly.

client/web-sveltekit/src/routes/search/+page.svelte
```diff
@@ -35,6 +35,7 @@
         queryFromURL={data.queryFromURL}
         {queryState}
         selectedFilters={data.queryFilters}
+        fetchRepoPopoverData={data.fetchRepoPopoverInfo}
```
camdencheek, line 38: Do we still need this?

client/web-sveltekit/src/lib/Popover.svelte
```diff
@@ -1,14 +1,20 @@
...
     /**
      * Show the popover when hovering over the trigger.
      */
     export let showOnHover = false
+    /**
+     * If true, popover content will use the border set by Popover. If false,
+     * the popover will not have a border, and the border should be set in the
+     * child component where the content is rendered.
+     */
+    export let useDefaultBorder = true
```
camdencheek, line 17: cc @fkling  on this one. Should we just get rid of the default border entirely? IMO a general popover component shouldn't handle styling at all.

fkling: That's the question, isn't it? Whether `Popover` is general or not. IMO the "general" version would be to use the `popover` action directly.
If we say we don't want to use the border color, wouldn't we have to make  the same decision for any of the other styling (e.g. box-shadow)?

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
fkling: We already have a `displayRepoName` function that's used across the product to produce a "nice" repo name.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -20,94 +21,94 @@ For example:
...
 
     const CENTER_DOT = '\u00B7' // interpunct
 
-    function truncateCommitNumber(numStr: string, length: number) {
+    function truncateCommitNumber(numStr: string | undefined, length: number): string | null {
+        if (!numStr) {
+            return null
+        }
         return numStr.substring(numStr.length - length)
     }
```
fkling, line 37: You can request the short OID from GraphQL. It's the `abbreviatedOid` field (or something like that).

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -20,94 +21,94 @@ For example:
...
         <div class="header">
-            <div class="icon-name-access">
-                <!-- @TODO: We need to use our customer's logo here, not the code host's -->
-                <!--Icon svgPath={mdiGitlab} /-->
-                <h4 class="repo-name">{repo.name}</h4>
-                <div class="access">
-                    <small>{repo.isPrivate ? 'Private' : 'Public'}</small>
-                </div>
+            <div class="left">
+                <Icon svgPath={orgSVGPath} --color="var(--primary)" />
```
fkling, line 59: I think the name `--color` was recently changed by Vova.

client/web-sveltekit/src/routes/search/+page.svelte
fkling: It doesn't seem used but I'd actually prefer it if use it (for now), and have the `search/+page.ts` loader provide `fetchRepoPopoverInfo`, not the global layout.

client/web-sveltekit/src/routes/search/RepoRev.svelte
```diff
@@ -1,12 +1,16 @@
...
+    import Popover from '$lib/Popover.svelte'
+    import RepoPopover from '$lib/repo/RepoPopover/RepoPopover.svelte'
     import CodeHostIcon from '$lib/search/CodeHostIcon.svelte'
     import { displayRepoName } from '$lib/shared'
 
     export let repoName: string
     export let rev: string | undefined
     export let highlights: [number, number][] = []
 
+    $: repoPopoverData = $page.data.fetchRepoPopoverInfo(repoName)
```
fkling, line 13: Seeing how this component is specific to the search page, `fetchRepoPopoverInfo` should (for now) be passed as a prop.

Should that function be provided by the global layout in the future then we'd have to extend the type of `$page.data` with the function so that we get some type safety.

dev/update-embeddings-list/tsconfig.json
fkling: Probably from the previous attempt to import `x.gql.ts`.

client/web-sveltekit/src/routes/+layout.ts
```diff
@@ -43,6 +43,13 @@ export const load: LayoutLoad = async ({ fetch }) => {
         settings,
         featureFlags: result.data.evaluatedFeatureFlags,
         globalSiteAlerts: globalSiteAlerts.then(result => result.data?.site),
+        fetchRepoPopoverInfo: async (name: string) => {
+            const response = await client.query(RepoPopoverQuery, { repoName: name }, { fetch, requestPolicy: 'network-only' })
+            if (!response.data || response.error) {
+                throw new Error(`Failed to fetch repo info: ${response.error}`)
+            }
+            return response.data.repository
+        },
```
fkling, line 52: I'd prefer it if that function was provided by the search page loader  (for now).

client/web-sveltekit/src/routes/+layout.ts
```diff
@@ -43,6 +43,13 @@ export const load: LayoutLoad = async ({ fetch }) => {
         settings,
         featureFlags: result.data.evaluatedFeatureFlags,
         globalSiteAlerts: globalSiteAlerts.then(result => result.data?.site),
+        fetchRepoPopoverInfo: async (name: string) => {
+            const response = await client.query(RepoPopoverQuery, { repoName: name }, { fetch, requestPolicy: 'network-only' })
```
fkling, line 47: Do we want to avoid caching because we always want to show the latest information (e.g. latest commit)? If so, leaving a comment to explain that might be useful.

client/web-sveltekit/src/routes/search/RepoRev.svelte
fkling: Yes, I understand, but this specific component (`RepoRev`) seems to be specific to the search page.

camdencheek: Ah, got it. Yep, makes sense

fkling: We will eventually have to revisit how we want to work/handle "on demand" data queries that are used across multiple pages, but maybe not right now 😉

dev/update-embeddings-list/tsconfig.json
jasonhawkharris: Yeah I just need to change it back now that it's working.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -20,94 +21,94 @@ For example:
...
-                {/each}
-            {/if}
+        <div class="description">
+            {repo.description}
         </div>
+        {#if repo.topics.length}
+            <div class="tags">
+                {#each repo.topics as topic}
+                    <small>{topic}</small>
+                {/each}
```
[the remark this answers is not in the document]
jasonhawkharris: Didn't know we had that component.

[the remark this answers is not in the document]
jasonhawkharris: When I plug in the badge component, here's what I get: 
Screenshot 2024-04-24 at 9 28 47 AM

I'm all for using standardized components but I think this one needs a little work. I'll follow up with an issue and assign it to myself.

client/web-sveltekit/src/lib/Popover.svelte
jasonhawkharris: I prefer it this way, personally. How about this, it looks like I'll be working with Popover's for the forseeable future. I'll update the component this way and try it out during development and see how it feels. It's a trivial change and we can always change it back if we want.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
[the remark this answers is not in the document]
camdencheek: In general, even if it looks bad, we should use the standardized component. If the standardized component needs updated, we should make it work for that. Badges should be consistent everywhere in the webapp

client/web-sveltekit/src/routes/search/RepoRev.svelte
jasonhawkharris: @camdencheek @fkling I just pushed a version that fetches the popover data inside of the RepoRev component. This seems to be the simplest solution for all the issues we're having at the moment. Whatever system we come up with in the future for on demand data, it will be easy to remove it from here and put it where it needs to go. 

Is that enough to resolve this for now? I can make a follow up issue to address how we handle on-demand data.

client/web-sveltekit/src/routes/+layout.ts
jasonhawkharris: TLDR: I've changed this back to how I initially had it, loading the data from the RepoRev component: See my comment here and let me know what you think: https://github.com/sourcegraph/sourcegraph/pull/61989#discussion_r1588033738

camdencheek:
Can we increase the hover delay slightly? It's pretty annoying to have the hovercard show up just when moving the mouse across the screen.

https://github.com/sourcegraph/sourcegraph/assets/12631702/9c3c264b-5228-4b10-99cf-9c4431a7fdfd

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -131,35 +118,38 @@ For example:
         align-items: center;
         padding: 0.5rem 0.75rem;
         background-color: var(--subtle-bg);
+        border-top-left-radius: 3px;
+        border-top-right-radius: 3px;
```
camdencheek, line 122: Why do we use absolute values here and not `var(--popover-border-radius)`? Also, why does this need to be set here when the radius is set on the root already?

client/web-sveltekit/src/routes/search/+page.ts
```diff
@@ -142,6 +142,7 @@ export const load: PageLoad = ({ url, depends }) => {
             searchStream,
             queryFilters,
             queryFromURL: query,
+            // fetchRepoPopoverData,
```
camdencheek, line 145: This intended to still be commented otu?

client/web-sveltekit/src/routes/search/RepoRev.svelte
```diff
@@ -1,12 +1,28 @@
...
     export let highlights: [number, number][] = []
 
+    const fetchPopoverInfo = async (repo: string) => {
+        const client = getGraphQLClient()
+        const popoverInfo = await client.query(RepoPopoverQuery, { repoName })
+        if (popoverInfo.data) {
+            return popoverInfo.data.repository
+        }
+        console.error('Failed to fetch popover info for', repo)
+        throw new Error('Failed to fetch popover info')
```
camdencheek, line 22: Looks like we ignore any information about the error that is provided by `popoverInfo.error`

camdencheek (commented):
screenshot-2024-05-13_15-28-34@2x

client/web-sveltekit/src/routes/search/RepoRev.svelte
```diff
@@ -21,12 +37,21 @@
...
-            {/if}
-        </a>
+        <Popover showOnHover let:registerTrigger placement="bottom-start" useDefaultBorder={false}>
+            <a class="repo-link" {href} use:highlightRanges={{ ranges: highlights }} use:registerTrigger>
+                {displayRepoName(repoName)}
+                {#if rev}
+                    <small class="rev"> @ {rev}</small>
+                {/if}
+            </a>
+            <div slot="content">
```
camdencheek, line 47: We can use `<svelte:fragment slot="content">` here to avoid introducing another div

camdencheek (approved):

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
jasonhawkharris: Because the radius needs to be different for the bottom two corners. I can use --popover-border-radius though.

client/web-sveltekit/src/routes/search/RepoRev.svelte
jasonhawkharris: Just curious... what's the problem with introducing another div here?

PR #62005, opened 2024-04-18
search: support draining Zoekt instances

keegancsmith:
This is an alternative implementation to https://github.com/sourcegraph/sourcegraph/pull/61833

This PR implements support for draining a zoekt replica via including its hostname in the comma-separated environment variable `INDEXED_SEARCH_DRAIN_SERVERS` on sourcegraph-frontend.

The way this functionality is implemented is via adjusting the endpoint map we use when making assignment of repos. We still report the hostname as part of the list of endpoints. However, the endpoint is left out of the consistent hash which maps the repositories to endpoints.

Our interactions with zoekt are already designed to do smooth rebalancing when the set of endpoints changes. We have logic to only remove repos from a replica once its new endpoint has it, and we support deduplication of search results across endpoints.

Test Plan: Stefan manually tested it.

Co-authored-by: @stefanhengl


stefanhengl (approved):

stefanhengl:
Here I move the repos back and forth between two Zoekt instances by setting either ` INDEXED_SEARCH_DRAIN_SERVERS` to either  `localhost:3070` or `localhost:3071`.

image

jtibshirani (commented):
Nice, this turned out quite clean.

PR #62011, opened 2024-04-18
feat: code insights language stats performance improvements

bahrmichael:
Closes https://github.com/sourcegraph/customer/issues/2962

This PR improves the performance of language stats insights by **50-70%**.

We achieved this by
- Traversing the file tree of a repository with a concurrency of up to 4 (default) against gitserver
- ~Forcing a depth-first-search to better utilize the cache that each directory-step uses~ (see below)
- Excluding `*.lock` files from the language stats insights

Furthermore we're
- Raising the timeout for language stats insights from 3 to 5 minutes
- Making the concurrency and timeout configurable

## New environment variables

* `GET_INVENTORY_TIMEOUT`: number, defines the number of minutes that the getInventory call has before timing out
* `GET_INVENTORY_GIT_SERVER_CONCURRENCY`: number, defines the maximum number of concurrent requests from the getInventory operation to the gitserver

## Test results

Screenshot 2024-04-23 at 16 15 16

## Why only 4 concurrent requests to gitserver

As you can see in the test results above, we started to get diminishing returns after 4 concurrent requests. This is probably a good default for customers, where they can get performance improvements by default, and have the ability to tweak it further if they have very powerful gitserver instances.

When we trace the language stats requests, we can see that increasing the concurrency further leads to longer response times from gitserver, which then sets off the improvement we get from concurrent requests.

## No depth first search

It turns out that the DFS used in https://github.com/sourcegraph/sourcegraph/pull/62011/commits/e3ef818aac365c6ed380bc32cb5a06f1d7418bf5 is 5-10% slower than the original approach. When the request is cancelled, only the files in the currently checked directory (not the files within subdirectories) need to be re-checked, which is an acceptable trade-off.

## Excluding .lock files

From my understanding, lock files are not valuable when it comes to understanding the language of a repository. [To keep compatibility with the linguist library](https://github.com/go-enry/go-enry/issues/132), go-envry doesn't flag all generated files as such.

In a future change we can enhance the pattern matching and make it configurable so that customers can tweak which files should be checked for their large repositories. Adding that right now may be premature optimization.

## Test plan

The affected code paths are covered by existing tests. I've added some unit tests for a newly added function. Furthermore I've done extensive manual testing.


cmd/frontend/internal/inventory/entries.go
```diff
@@ -61,39 +74,67 @@ func (c *Context) tree(ctx context.Context, tree fs.FileInfo, buf []byte) (inv I
...
+					results <- treeIteratorResult{i, Inventory{Languages: []Lang{lang}}, err}
+				}
+				results <- treeIteratorResult{i, Inventory{Languages: []Lang{lang}}, nil}
+
+			case e.Mode().IsDir(): // subtree
+				subtreeInv, err := c.tree(ctx, e, buf, sem)
+				if err != nil {
+					results <- treeIteratorResult{i, subtreeInv, err}
+				}
+				results <- treeIteratorResult{i, subtreeInv, nil}
```
camdencheek, line 110: No need to check `err` since we just pass the result along anyways.

```suggestion
				results <- treeIteratorResult{i, subtreeInv, err}
```

cmd/frontend/internal/inventory/entries.go
```diff
@@ -61,39 +74,67 @@ func (c *Context) tree(ctx context.Context, tree fs.FileInfo, buf []byte) (inv I
 	}
 
 	entries, err := c.ReadTree(ctx, tree.Name())
+	sem.Release(1)
 	if err != nil {
 		return Inventory{}, err
 	}
 	invs := make([]Inventory, len(entries))
+	results := make(chan treeIteratorResult, len(entries)) // Buffer the channel to the number of entries
+	var wg sync.WaitGroup
```
camdencheek, line 83: Certainly not required, but am going to ~shamelessly shill~ recommend checking out [conc](https://github.com/sourcegraph/conc)

cmd/frontend/internal/inventory/entries.go
```diff
@@ -61,39 +74,67 @@ func (c *Context) tree(ctx context.Context, tree fs.FileInfo, buf []byte) (inv I
...
 	return Sum(invs), nil
 }
 
 // file computes the inventory of a single file. It caches the result.
-func (c *Context) file(ctx context.Context, file fs.FileInfo, buf []byte) (inv Inventory, err error) {
+func (c *Context) file(ctx context.Context, file fs.FileInfo, buf []byte, sem *semaphore.Weighted) (inv Inventory, err error) {
 	// Get and set from the cache.
+	if err := sem.Acquire(ctx, 1); err != nil {
+		return Inventory{}, err
+	}
```
camdencheek, line 137: We probably do not need to acquire from the semaphore until after we checked the cache. The semaphore just needs to bound gitserver requests

cmd/frontend/internal/inventory/entries.go
```diff
@@ -108,6 +149,7 @@ func (c *Context) file(ctx context.Context, file fs.FileInfo, buf []byte) (inv I
 	}
 
 	lang, err := getLang(ctx, file, buf, c.NewFileReader)
+	sem.Release(1)
```
camdencheek, line 152: style nit: where appropriate, using `defer sem.Release(1)` right after acquire is preferable.

As a very concrete example of why this is helpful, you acquire the semaphore above, then check the cache and return early if there is a cache hit. In the case of a cache hit, `sem.Release(1)` is never called, and we will eventually exhaust our semaphore permanently and never make any progress.

cmd/frontend/internal/inventory/entries.go
```diff
@@ -61,39 +74,67 @@ func (c *Context) tree(ctx context.Context, tree fs.FileInfo, buf []byte) (inv I
...
-			// Don't individually cache files that we found during tree traversal. The hit rate for
-			// those cache entries is likely to be much lower than cache entries for files whose
-			// inventory was directly requested.
-			lang, err := getLang(ctx, e, buf, c.NewFileReader)
-			if err != nil {
-				return Inventory{}, errors.Wrapf(err, "inventory file %q", e.Name())
-			}
-			invs[i] = Inventory{Languages: []Lang{lang}}
+		wg.Add(1)
+		go func(i int, e os.FileInfo) {
```
camdencheek, line 87: Fun fact: in the latest version of go, passing in `i` and `e` is no longer required and it's safe to capture those variables in the goroutine closure. See [here](https://go.dev/blog/loopvar-preview) for details.

camdencheek (commented):
This looks like a great approach! And the numbers quoted are promising.

cmd/frontend/internal/inventory/entries.go
bahrmichael: I found that too many parallel requests lead to failures in redis. That's why I extended the semaphore around the cache access as well.

cmd/frontend/internal/inventory/entries.go
bahrmichael: Good catch!

cmd/frontend/internal/inventory/entries.go
bahrmichael: Nice! Thank you :)

cmd/frontend/internal/inventory/entries.go
bahrmichael: I'm using `iter.MapErr` now, let me know if my usage is wrong :)

cmd/frontend/internal/inventory/inventory.go
```diff
@@ -39,13 +40,38 @@ type Lang struct {
...
+		matched, err := regexp.MatchString(pattern, name)
+		if err != nil {
+			return false, err
+		}
+		if matched {
+			return matched, nil
+		}
+	}
+	return false, nil
+}
```
camdencheek, line 56: Why was this added?

cmd/frontend/backend/inventory.go
```diff
@@ -40,22 +58,50 @@ func InventoryContext(logger log.Logger, repo api.RepoName, gsClient gitserver.C
...
 			return gsClient.ReadDir(ctx, repo, commitID, path, false)
 		},
 		NewFileReader: func(ctx context.Context, path string) (io.ReadCloser, error) {
-			return gsClient.NewFileReader(ctx, repo, commitID, path)
+			trc, ctx := trace.New(ctx, "NewFileReader waits for semaphore")
+			err := gitServerSemaphore.Acquire(ctx, 1)
+			if err != nil {
+				return nil, err
+			}
+			trc.End()
```
camdencheek, line 85: Should we end this span when the reader finishes rather than when we return a reader? e.g. in `releaseSemaphore`?

cmd/frontend/backend/inventory.go
```diff
@@ -40,22 +58,50 @@ func InventoryContext(logger log.Logger, repo api.RepoName, gsClient gitserver.C
...
-			// to avoid sequential tree traversal calls.
+			trc, ctx := trace.New(ctx, "ReadTree waits for semaphore")
+			err := gitServerSemaphore.Acquire(ctx, 1)
+			trc.End()
+			if err != nil {
+				return nil, err
+			}
+			defer gitServerSemaphore.Release(1)
+			// Using recurse=true is not possible, because we're likely to run into the following error:
+			// "Maximum call stack size exceeded"
```
camdencheek, line 76: That's pretty surprising to me -- why is the stack size so large? Does recurse also return the target so we end up in an infinite recursion?

camdencheek (approved):

cmd/frontend/backend/inventory.go
bahrmichael: I haven't gone much deeper than trying this out and seeing the error. My first intention was "it's just going through a lot of folders, and that's breaking it", but now I feel more like this tree call PLUS all the other tree calls that I still did at that time are leading to the call stack error. I can try again with just one ReadDir request later.

cmd/frontend/backend/inventory.go
bahrmichael: Yes!

cmd/frontend/internal/inventory/inventory.go
bahrmichael: See "Excluding .lock files" in the PR description for the reason to exclude this file.

This particular method is here because I started out making the filename patterns user configurable. After we talked about environment variables I cut that back, but left this method in place. In the trade-off between just writing what you need or optimizing for future use cases, I'm leaning slightly to the latter (which means keeping this code). Happy to change my mind though :)

cmd/frontend/internal/inventory/inventory.go
```diff
@@ -91,7 +119,9 @@ func getLang(ctx context.Context, file fs.FileInfo, buf []byte, getFileReader fu
 	}
 	lang.Name = matchedLang
 
+	trc, _ := trace.New(ctx, "CountLines")
 	lineCount, byteCount, err := countLines(rc, buf)
+	trc.End()
```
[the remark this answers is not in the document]
bahrmichael: Since our spans now properly surround the filereader we can drop the ones here alltogether. I will instead add an log even to the trace that tells us which path the evaluation is taking for which file. Since the path with `countLines` could be optimized with a gitserver side LOC count that should help us get a better understanding of the potential performance gain there. Same for files where enry may be unsure, but we can make a confident call that a certain file type should be treated as a particular language.

cmd/frontend/backend/inventory.go
bahrmichael: I took another stab at this, and it was indeed about recursive recursion breaking the thing.

The numbers aren't too promising. When running with a recursive call (and skipping the recursion in the inventory), we get between 0.8% and 9.5% performance improvement, without a conclusive pointer what kind of repos actually benefit from this. I'll update the comment with a reference to this, but wouldn't attempt a further improvement here until we hear back from customers.

Screenshot 2024-04-24 at 10 34 14

cmd/frontend/internal/inventory/BUILD.bazel
```diff
@@ -32,3 +35,18 @@ go_test(
...
     ],
 )
+
+go_test(
+    name = "entries_test",
+    srcs = [
+        "entries_test.go",
+    ],
+    embed = [":inventory"],
+    race = "off",
```
bahrmichael, line 45: I updated entries.go to use concurrency via conc's iter.MapErr. This means that we're racing various tasks. Our bazel tests by default don't like racing. Unless we want to do a major refactor of the code, we need to allow racing here. From what I can see in the code this should be safe.

camdencheek: Oh, that's a problem. If the race detector triggers, that _almost always_ indicates a bug. It has a very low false positive rate. We should fix the race conditions.

cmd/frontend/internal/inventory/inventory.go
camdencheek: > lock files are not valuable when it comes to understanding the language of a repository

Inventory is not about understanding _the_ language, it's about understanding the _distribution_ of languages. And lockfiles contribute to that. For example, I would be surprised if a repo that contains a `syntax-highlighter.lock.json` did not list JSON in its languages. 

I think we should remove this addition since we already have exclusion rules elsewhere in Sourcegraph, and adding another invisible and non-configurable layer seems likely to add to user confusion about this feature.

camdencheek: Also, a note that regex should be pre-compiled when the pattern is known at build time. Right now, this re-compiles the regex on every file, which is quite expensive. I wouldn't be surprised to see this show up pretty heavily on a profile.

Issue #61627, opened 2024-04-05
Svelte: Implement streamlined repo header


taiyab:
I've gone ahead and updated the repo header in the designs.

We should try and implement it as shown.

## Now
![Image](https://github.com/sourcegraph/sourcegraph/assets/5337876/17efd1cd-8afe-4683-9fd6-a2915d829ac1)

## Proposed

![Image](https://github.com/sourcegraph/sourcegraph/assets/5337876/1d4e1b72-d938-4627-8cee-8b55cc87a666)

![Image](https://github.com/sourcegraph/sourcegraph/assets/5337876/deebfb34-e6d5-464b-b1d6-49a946b07ee8)

We need to make sure it still accommodates for all these settings:
<img width="396" alt="image" src="https://github.com/sourcegraph/sourcegraph/assets/5337876/d2d2637f-8a2f-4bee-83d4-1b86baf7251a">

Prototype:
https://www.figma.com/proto/EPgl0qVbqGSumvf52brhda/Code-Search?page-id=1397%3A21372&type=design&node-id=1947-37160&viewport=-754%2C355%2C1&t=33wVR7ZPnEzFq6W7-9&scaling=min-zoom&starting-point-node-id=1947%3A37160&show-proto-sidebar=1

Designs:
https://www.figma.com/file/EPgl0qVbqGSumvf52brhda/Code-Search?type=design&node-id=1947%3A37160&mode=design&t=otqoTqJC6RRLuWXb-1

https://www.figma.com/file/EPgl0qVbqGSumvf52brhda/Code-Search?type=design&node-id=2690%3A79770&mode=design&t=otqoTqJC6RRLuWXb-1

PR #62012, opened 2024-04-18
Svelte: Add new generic global navigation UI

vovakulikov:
Closes https://github.com/sourcegraph/sourcegraph/issues/61627

This PR adds a new navigation UI layout that is able to render any arbitrary UI from the children's level. This allows us to render different global navigation menu UI based on context 

- Global product nav on the home page
- Search bar on the search result page
- Repository navigation on the repo and blob UI pages

| Page | Screenshot |
| -------- | -------- |
| Search results with search bar | Screenshot 2024-04-18 at 16 11 23 |
| Repository page with repo nav | Screenshot 2024-04-18 at 16 12 03 |
| With opened sidebar nav | Screenshot 2024-04-18 at 16 12 33 |

## Test plan
- Go to the search home page, it should be plain navigation that we had prior to this PR
- Go to the search result home, check that navigation rendered search box instead of product navigation (product navigation is still available through sidebar navigation)
- Go to the repository page, and check that the main navigation renders the repository menu instead of the product menu


camdencheek:
Looks really nice! Could you rebase this on `main` to pull in the scrolling changes Felix merged? Tried to review, but looks like there are some changes that meaningfully affect this PR

camdencheek:
Fine as a followup: could we add more clear hierarchy in the hamburger popover?

screenshot-2024-04-18_16-30-09@2x

client/web-sveltekit/src/lib/dom.ts
```diff
@@ -303,9 +305,19 @@ export const classNames: Action<HTMLElement, string | string[]> = (node, classes
...
+export const portal: Action<HTMLElement, { container?: HTMLElement | null } | undefined> = (target, options) => {
+    const root = options?.container ?? window.document.body
+
+    root.appendChild(target)
+
+    console.log('INITIAL CALL', options)
+
     return {
+        update(options) {
+            console.log('UPDATE', options)
```
camdencheek, line 317: intentional console.log?

client/web-sveltekit/src/lib/navigation/GlobalHeader.svelte
```diff
@@ -0,0 +1,207 @@
...
+                }
+                100% {
+                    transform: rotate(180deg) scale(1);
+                }
+            }
+
+            @media (prefers-reduced-motion: no-preference) {
+                animation: spin 0.5s ease-in-out 1;
+            }
+        }
```
camdencheek, line 144: Am I understanding this right that the logo should spin on hover? I can't seeem to get that working in my browser (firefox)

client/web-sveltekit/src/lib/navigation/GlobalHeader.svelte
```diff
@@ -0,0 +1,207 @@
...
+        flex: 1;
+        display: flex;
+        align-self: stretch;
+        min-width: 0;
+
+        &-list {
+            display: flex;
+            gap: 1rem;
+            padding: 0;
+            margin: -0.5rem 0 -0.5rem 0;
```
camdencheek, line 157: Why do we need negative margins here?

client/web-sveltekit/src/lib/navigation/GlobalHeader.svelte
```diff
@@ -0,0 +1,207 @@
...
+
+        p:last-child {
+            margin-bottom: 0;
+        }
+    }
+
+    // Custom menu with sidebar navigation controls styles
+    .menu-button {
+        border: none;
+        padding: 0.35rem 0.35rem;
```
camdencheek, line 190:
```suggestion
        padding: 0.35rem;
```

client/web-sveltekit/src/lib/navigation/GlobalHeader.svelte
```diff
@@ -0,0 +1,207 @@
...
+        border-radius: var(--border-radius);
+        display: flex;
+        align-items: center;
+        background-color: transparent;
+        margin-right: -0.5rem;
```
camdencheek, line 195: Rather than using negative margins to cancel out the flexbox gap, can we just put the menu button and the logo in their own div so we can control the spacing between them better? Or maybe just move the menu button outside of the flex div and have the remainder of the header flex?

client/web-sveltekit/src/lib/navigation/GlobalSidebarNavigation.svelte
```diff
@@ -0,0 +1,153 @@
...
+                height: 1rem;
+                color: var(--icon-color);
+            }
+        }
+
+        header {
+            display: flex;
+            gap: 0.25rem;
+            align-items: center;
+            padding: 0.67rem 0.5rem 0.5rem 0.5rem;
```
camdencheek, line 100: Why is the top padding unbalanced here?

client/web-sveltekit/src/lib/search/input/SearchInput.svelte
```diff
@@ -293,6 +295,12 @@
                 display: block;
             }
         }
+
+        &.compat {
+            padding: 0.25rem;
+            margin: -0.25rem;
+            width: calc(100% + 0.5rem);
```
camdencheek, line 302: Instead of negating the additional padding with negative margin in header mode, can we instead just only add it in the other modes?

camdencheek (approved):

client/web-sveltekit/src/lib/dom.ts
```diff
@@ -247,13 +247,15 @@ interface ComputeFitAttributes {
 export const computeFit: Action<HTMLElement, void, ComputeFitAttributes> = node => {
     // Holds the cumulative width of all elements up to element i.
     const widths: number[] = [0]
+    const parentRight = node.getBoundingClientRect().left
```
fkling, line 250: Despite what I side during our call, this is not actually the right side of the parent, it's the left 😅 
I suggest we rename this do something more neutral, `offset` and add a comment explaining the intention behind it.

```suggestion
    // Used to compute the "end" of each child relative to the parent container.
    // This ensures that the logic here still works when the parent node is moved
    // or when it is scrolled inside a scroll container.
    const offset = node.getBoundingClientRect().left
```

client/web-sveltekit/src/lib/dom.ts
```diff
@@ -247,13 +247,15 @@ interface ComputeFitAttributes {
...
 
     for (let i = 0; i < node.children.length; i++) {
-        widths[i + 1] = node.children[i].getBoundingClientRect().right
+        widths[i + 1] = node.children[i].getBoundingClientRect().right - parentRight
     }
 
     function compute(): void {
-        const right = node.getBoundingClientRect().right
+        const rootRect = node.getBoundingClientRect()
+        const right = rootRect.right - rootRect.left
```
fkling, line 258: We can probably just use `width` here.

```suggestion
        const nodeWidth = node.getBoundingClientRect().width
```

client/web-sveltekit/src/lib/dom.ts
```diff
@@ -303,9 +305,19 @@ export const classNames: Action<HTMLElement, string | string[]> = (node, classes
 /**
  * An action to move the attached element to the end of the document body.
```
fkling, line 306: This comment needs to be updated to reflect the new behavior of the this action/function.

client/web-sveltekit/src/lib/navigation/GlobalHeader.svelte
```diff
@@ -0,0 +1,207 @@
+<script context="module" lang="ts">
+    import { type Writable, writable } from 'svelte/store'
+
+    export enum NavigationMode {
+        PlainNavigation = 'global',
+        WithCustomContent = 'with-custom-content',
+    }
+
+    export const navigationModeStore = writable<NavigationMode | `${NavigationMode}`>(NavigationMode.PlainNavigation)
+    export const navigationExtensionElement: Writable<HTMLElement | null> = writable(null)
```
fkling, line 10: In theory, anything that imports this store could assign a new node to this, which we obviously do not want.
You could declare the writable store as local and export a read-only version of it:

```ts
import { readonly } form 'svelte/store'

const extensionElement: Writable<HTMLElement | null> = writable(null)
export const navigationExtensionElement = readonly(extensionElement)
```

Also using the store like this means that `GlobalHeader` can only be instantiated once, which I guess is not that surprising. But in this case I would probably keep it inside `src/routes/`, since it's not supposed to be used anywhere else.

client/web-sveltekit/src/lib/navigation/GlobalHeader.svelte
```diff
@@ -0,0 +1,207 @@
...
+    const isDevOrS2 =
+        (browser && window.location.hostname === 'localhost') ||
+        window.location.hostname === 'sourcegraph.sourcegraph.com'
+
+    $: withCustomContent = $navigationModeStore === NavigationMode.WithCustomContent
+</script>
+
+<header class="root">
+    {#if isSidebarNavigationOpen}
+        <GlobalSidebarNavigation onClose={() => setSideNavigationState(false)} />
```
fkling, line 46: Nit: You could just use

```suggestion
        <GlobalSidebarNavigation onClose={() => isSidebarNavigationOpen = false} />
```
and remove the additional function declaration.

client/web-sveltekit/src/lib/navigation/GlobalSidebarNavigation.svelte
```diff
@@ -0,0 +1,153 @@
...
+
+            &:hover {
+                background-color: var(--secondary-2);
+            }
+
+            :global([data-icon]) {
+                width: 1rem;
+                height: 1rem;
+                color: var(--icon-color);
+            }
```
fkling, line 93: `Icon` has a `--size` CSS variable that you can just set. Some for `color`. You don't have to use `global` here:

```suggestion
--size: 1rem;
--color: var(--icon-color)
```

We should rename those variables to be specific to the component though, e.g. `--icon-size`.
 
The same goes for all other places where you are targeting the `Icon` component.

client/web-sveltekit/src/lib/navigation/GlobalSidebarNavigation.svelte
```diff
@@ -0,0 +1,153 @@
...
+            margin: 0;
+            list-style: none;
+            display: flex;
+            flex-direction: column;
+            gap: 0.25rem;
+            flex-grow: 1;
+
+            li[aria-current='page'] > :global(a) {
+                background-color: var(--secondary-2);
+            }
```
fkling, line 127: Why do we not style the `li` element instead?

client/web-sveltekit/src/lib/navigation/GlobalSidebarNavigation.svelte
```diff
@@ -0,0 +1,153 @@
...
+                font-size: 1rem;
+
+                :global([data-icon]) {
+                    color: var(--icon-color);
+                }
+
+                &:hover {
+                    background-color: var(--secondary-2);
+                }
+            }
```
fkling, line 144: Is the style for the list version really that different? If it's not possible to make it work with components own style then I'd either

1. Add a `variant` prop to the component
2. Make the component unstyled and expose the stylable elements via `data-*` props.

client/web-sveltekit/src/lib/navigation/GlobalSidebarNavigation.svelte
```diff
@@ -0,0 +1,153 @@
...
+
+            &.sub-list {
```
fkling, line 146: This could also just be `> ul`.

client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte
```diff
@@ -38,9 +40,10 @@
         { path: '/-/settings', icon: mdiCog, title: 'Settings' },
     ]
 
-    let visibleNavEntries = navEntries.length
-    $: navEntriesToShow = visibleNavEntries === navEntries.length ? navEntries : navEntries.slice(0, visibleNavEntries)
-    $: overflowMenu = visibleNavEntries !== navEntries.length ? navEntries.slice(visibleNavEntries) : []
+    const visibleNavEntries: Writable<number> = writable(navEntries.length)
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte:43]: Why did you turn this into a store? It doesn't look like it has to be one.

client/web-sveltekit/src/lib/navigation/GlobalHeader.svelte
vovakulikov: @camdencheek can you check that your reduce-motion setting is off? I checked it in my Firefox 125.0.1 (64-bit) and it works. Also notice that only Sourcegraph logo in the global header spins and logo in the sidebar menu doesn't

client/web-sveltekit/src/lib/navigation/GlobalHeader.svelte
vovakulikov: Here is the structure of this element
```html
<header>
  <logo/>
 
  <navigation-container>
     <plain-list-navigation/>
  <navigation-container/>
</header
```

So this element with negative margins (plain-list-navigation) represents a global navigation list in which we have navigation tabs, since `navigation-container` has some paddings (because this element also represents a container for any arbitrary UI, kind of portal) and we want to ensure that this generic content will look good even if the consumer doesn't add paddings themselves. 

The reason we use negative margins because `plain-list-navigation` UI have tabs with border at the bottom for selected active tab and this border should be at the bottom edge visually. 

Screenshot 2024-04-19 at 15 54 06

client/web-sveltekit/src/lib/navigation/GlobalHeader.svelte
camdencheek: Oh! Sure enough. I had no idea I had that reduce motion enabled. Works great when I turned that off. Thanks for the pointer

client/web-sveltekit/src/lib/navigation/GlobalHeader.svelte
vovakulikov: I understand the concern about the negative margin. Usually, I think this is better to keep the layout div (any generic wrappers) count as minimal as possible. I don't want to confuse anyone with margins as well, so I added a div wrapper for the logo button and logo icon link.

client/web-sveltekit/src/lib/navigation/GlobalSidebarNavigation.svelte
vovakulikov: Note that when you open the sidebar menu, the menu-button icon or Sourcegraph Icon don't move a bit (no UI jumps between close/open states); this non-standard padding is exactly for this. But just realised that probably I can ensure the same height and paddings to avoid jumps rather than calc paddings there.

client/web-sveltekit/src/lib/navigation/GlobalHeader.svelte
vovakulikov: I was thinking about making the element stream read-only, but I thought that having an intermediate stream was too much for a non-existing problem (I have no problem adding it, though). 

> But in this case I would probably keep it inside src/routes/, since it's not supposed to be used anywhere else.

This is what I wanted to discuss at some point, it looks like our routes directories are growing fast and already have a lot of files, sometimes navigation there isn't that straightforward (pages make a lot of sense but components usually do not) so I moved it to the lib but I agree lib is for reusable components? (but should we change it maybe?)

client/web-sveltekit/src/lib/navigation/GlobalSidebarNavigation.svelte
vovakulikov: >Icon has a --size CSS variable that you can just set. Some for color. You don't have to use global here:

I don't think I see it in the Icon component (did you confuse it with Avatar component?), I can add it, though, if you think this would be better.

client/web-sveltekit/src/lib/navigation/GlobalSidebarNavigation.svelte
vovakulikov: There is no reason, changed it to li element

client/web-sveltekit/src/lib/navigation/GlobalSidebarNavigation.svelte
vovakulikov: A few questions to clarify 
- > Is the style for the list version really that different? 

different from what component?

> Add a variant prop to the component

- to which component do you suggest to add variant?

My point here is we probably should be ok with not sharing everything as much as possible, layout here is relatively simple I'm afraid if we try to reuse and override some parts of list component it would be more complex rather than simple list in place.

client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte
Comment (vovakulikov) [reply client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte:43]: Yeah, I previously did some experiments with extracting it in a separate component to generalise the approach but it's too much for this PR so I rolled back but forgot about the visible entries stream, thanks for spotting.

vovakulikov:
I fixed most of the comments and rebased the main (the diff got bigger because the main branch wasn't formatted so I did it in this PR) merging this one, let me know if anything big was left.

client/web-sveltekit/src/lib/navigation/GlobalHeader.svelte
fkling: > but I thought that having an intermediate stream was too much for a non-existing problem

It doesn't create an intermediate stream. It returns a new object with the same `subscribe` method but without mutation methods:

https://sourcegraph.com/github.com/sveltejs/svelte@b91a67ba6845c5d37dbb6f2fda8a0d057f828dab/-/blob/packages/svelte/src/store/index.js?L193-206

> This is what I wanted to discuss at some point, it looks like our routes directories are growing fast and already have a lot of files, sometimes navigation there isn't that straightforward (pages make a lot of sense but components usually do not) so I moved it to the lib but I agree lib is for reusable components? (but should we change it maybe?)

I haven't arrived at a good pattern yet. On one hand I'd prefer it if components that are used by a single page are located where the page is. On the other hand the specific nature of the `src/routes/` directory makes it impossible to have more meaningful code organization there.

I'd be fine if all components lived in `src/lib/` if we follow specific patterns to indicate shared components. For example the `src/lib/search/` directory could have a `shared/` subfolder and only modules/components inside this folder are allowed to be used by other parts of the app.

client/web-sveltekit/src/lib/navigation/GlobalSidebarNavigation.svelte
fkling: Whoops, it accepts a size property.

client/web-sveltekit/src/lib/navigation/GlobalSidebarNavigation.svelte
fkling: I meant the `MainNavigationLink` component. It seems that the `:global(a)` is supposed to overwrite the styles of that component.

client/web-sveltekit/src/lib/search/input/SearchInput.svelte
camdencheek: My suggestion was to just not claim that extra space when we're in header mode. With the negative margins, the negative margin is tightly coupled to the size of the padding we add. No big deal though, no need to follow up, thanks for the explanation 🙂

PR #62017, opened 2024-04-18
Support setting clientID for calls from Cody Gateway to sourcegraph.com

rafax:
Add a header that we can use to identify Cody Gateway requests to sourcegraph.com - [context](https://sourcegraph.slack.com/archives/C05497E9MDW/p1713453541038879?thread_ts=1713431784.192999&cid=C05497E9MDW).

## Test plan

- local testing

### DIFF
--- a/cmd/cody-gateway/internal/dotcom/dotcom.go
+++ b/cmd/cody-gateway/internal/dotcom/dotcom.go
@@ -25,7 +25,7 @@ import (
 //	println(resp.GetDotcom().ProductSubscriptionByAccessToken.LlmProxyAccess.Enabled)
 //
 // The client generator automatically ensures we're up-to-date with the GraphQL schema.
-func NewClient(endpoint, token string) graphql.Client {
+func NewClient(endpoint, token, clientID string) graphql.Client {
 	return &tracedClient{graphql.NewClient(endpoint, &http.Client{
 		Transport: &tokenAuthTransport{
 			token:   token,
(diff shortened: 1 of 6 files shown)


cmd/cody-gateway/internal/dotcom/dotcom.go
```diff
@@ -82,5 +83,8 @@ func (t *tokenAuthTransport) RoundTrip(req *http.Request) (*http.Response, error
 	req.URL.RawQuery = req.Context().Value(contextKeyOp).(string)
 
 	req.Header.Set("Authorization", fmt.Sprintf("token %s", t.token))
+	if t.clientID != "" {
+		req.Header.Set("X-Sourcegraph-Client-ID", t.clientID)
+	}
```
evict, line 88: Can we also supply a better user-agent header? Right now it's pretty generic and it will be helpful for many metrics in Cloudflare's dashboard.

evict (approved):

cmd/cody-gateway/internal/dotcom/dotcom.go
rafax: Yes, what do you have in mind - `X-User-Agent: Cody-Gateway` ? Including env (dev/prod)? Including version?

evict: That would be great! I'd omit the `X-` so we keep the industry standard: `Cody-Gateway/env (version)` something like that?

rafax: Done, PTAL

PR #62018, opened 2024-04-18
chore: Completely sunset qdrant

eseliger:
This removes qdrant from this codebase entirely.
All the docker images, dependencies, (dead) usage in code.

My understanding is that we don't use this feature and never properly rolled it out.

Test plan:

CI passes and code review from owners.


jtibshirani (approved):
This looks good to me. I have a feeling qdrant will make a comeback down the line :) But it really feels like the right call for now to clean this up.

Qdrant shouldn't be enabled anywhere. But it was previously enabled on S2, and it'd be good to clean up that config. (Actually, why is there any embeddings config on S2?)

```
 "embeddings": {
    "dimensions": 1536,
    "endpoint": "https://cody-gateway.sourcegraph.com/v1/embeddings",
    "model": "openai/text-embedding-ada-002",
    "provider": "sourcegraph",
    "qdrant": {
      "enable": true
    }
  },
```

willdollman:
Merged main to bring in [this fix](https://github.com/sourcegraph/sourcegraph/pull/62033) for the CI package issue, and reverted my temporary workaround. Buildkite is passing so good to go!

PR #62023, opened 2024-04-18
svelte: Fix playwright tests

fkling:
Running playwright tests on the production build has been broken for some tests for two reasons:

- In the production build the app has knowledege about which routes are enabled. It assumes none are and so any navigation event results in the page reloading. This can have some unexpected effects on the state of the application.
- `this.page.addInitScript` returns a promise that has to be awaited. Otherwise we cannot be sure that it has executed before the other steps of the test. I don't know why that wasn't a problem in the development build.

To solve the first issue I added a new "preview" build which is identical to the production build except that every route is "enabled" and served normally. I've updated the playwright config to use that build instead. It can also be used "manually" by running `pnpm build:preview`.

For anyone whose curious about this: `dev || import.meta.env.MODE === 'preview'` is being complete removed from the build output.

## Test plan

`pnpm test` runs without error

### DIFF
--- a/client/web-sveltekit/package.json
+++ b/client/web-sveltekit/package.json
@@ -6,6 +6,7 @@
     "dev:dotcom": "vite dev --mode=dotcom",
     "dev:enterprise": "DEPLOY_TYPE=dev vite build --watch",
     "build": "vite build",
+    "build:preview": "vite build --mode=preview",
     "build:watch": "vite build --watch",
     "preview": "vite preview",
     "test": "playwright test",
--- a/client/web-sveltekit/src/routes/layout.spec.ts
+++ b/client/web-sveltekit/src/routes/layout.spec.ts
@@ -10,10 +10,12 @@ test('has sign in button', async ({ page }) => {
     await expect(page).toHaveURL('/sign-in')
 })
 
-test('has experimental web app toggle button', async ({ page }) => {
+test('has experimental opt out popover', async ({ sg, page }) => {
+    sg.signIn({ username: 'test' })
+
     await page.goto('/')
-    await page.getByLabel('Leave experimental web app').click()
-    await expect(page).toHaveURL(/\?feat=-web-next/)
+    await page.getByText('Experimental').hover()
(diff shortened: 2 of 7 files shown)


client/web-sveltekit/src/routes/layout.spec.ts
```diff
@@ -10,10 +10,12 @@ test('has sign in button', async ({ page }) => {
...
 
-test('has experimental web app toggle button', async ({ page }) => {
+test('has experimental opt out popover', async ({ sg, page }) => {
+    sg.signIn({ username: 'test' })
+
     await page.goto('/')
-    await page.getByLabel('Leave experimental web app').click()
-    await expect(page).toHaveURL(/\?feat=-web-next/)
+    await page.getByText('Experimental').hover()
+    await expect(page.getByRole('link', { name: 'opt out' })).toBeVisible()
```
fkling, line 18: The implementation of this has changed.

client/web-sveltekit/package.json
```diff
@@ -6,6 +6,7 @@
     "dev:dotcom": "vite dev --mode=dotcom",
     "dev:enterprise": "DEPLOY_TYPE=dev vite build --watch",
     "build": "vite build",
+    "build:preview": "vite build --mode=preview",
     "build:watch": "vite build --watch",
     "preview": "vite preview",
```
camdencheek, line 11: what's the difference between `pnpm build:preview` and `pnpm preview`?

client/web-sveltekit/src/testing/integration.ts
```diff
@@ -119,8 +119,8 @@ class Sourcegraph {
      * the search results being received. The returned function will wait for the search results
      * page to be "ready" by waiting for the "Filter results" heading to be visible.
      */
-    public mockSearchStream(): MockSearchStream {
-        this.page.addInitScript(function () {
+    public async mockSearchStream(): Promise<MockSearchStream> {
+        await this.page.addInitScript(function () {
```
camdencheek, line 123: good catch! Thanks for the fix 😄

camdencheek (approved):

PR #62024, opened 2024-04-18
svelte: Remove full page scrolling of repo pages

fkling:
This commit removes the full page scrolling behavior for repo root, directory and file pages. The implementation had several issues.

In the new implementation the scroll container is CodeMirror itself. It provides the following behavior:

- Show the top of the file when navigating to a new file.
- Scroll the selected line into view when opening a URL containing a position.
- Do not scroll when selecting a line (currently broken).
- Restore previous scroll position when navigating backward or forward in the browser history.
- Toggling the blame view maintains the current line view (well, almost but that's as best as we can do at the moment).

Additional changes in this commit:

- Removed logic to enable/disable full page scrolling.
- Remove CSS for making elements `sticky`.
- Update "scroll selected line into view" logic to prevent scrolling when restoring the previous scroll position.
- Update `codemirror/view` to include a recent fix for a scroll related bug.

NOTE: Due to these changes we lost scroll restoration for repo root and directory pages. This will be fixed in a separate PR.

## Test plan

`pnpm test` passes

Manual testing:

- Open file page in new tab -> file is scrolled to the top
- Open file page with selected line in new tab -> selected line is scrolled into view
- Navigate to new file via file tree -> new file is scrolled to the top
- Scroll file, navigate to new file and back -> scroll position is restored
- Scroll file, navigate back, navigate forward -> scroll position is restored

- Preview panel appears to be working (can scroll and navigate between matches)
- Directory page is scrollable
- Repo root page is scrollable

### DIFF
--- a/client/web-sveltekit/src/lib/CodeMirrorBlob.svelte
+++ b/client/web-sveltekit/src/lib/CodeMirrorBlob.svelte
@@ -28,8 +28,6 @@
         languages: string[]
     }
 
-    const extensionsCompartment = new Compartment()
-
     const defaultTheme = EditorView.theme({
         '&': {
             height: '100%',
@@ -42,6 +40,7 @@
             lineHeight: '1rem',
             fontFamily: 'var(--code-font-family)',
             fontSize: 'var(--code-font-size)',
+            overflow: 'auto',
         },
         '.cm-content': {
             padding: 0,
@@ -106,20 +105,12 @@
         defaultTheme,
         linkify,
     ]
-
-    function configureSyntaxHighlighting(content: string, lsif: string): Extension {
-        return lsif ? syntaxHighlight.of({ content, lsif }) : []
-    }
-
-    function configureMiscSettings({ wrapLines }: { wrapLines: boolean }): Extension {
-        return [wrapLines ? EditorView.lineWrapping : []]
-    }
 </script>
 
 <script lang="ts">
     import '$lib/highlight.scss'
 
-    import { Compartment, EditorState, type Extension } from '@codemirror/state'
+    import { EditorState, type Extension } from '@codemirror/state'
     import { EditorView } from '@codemirror/view'
     import { createEventDispatcher, onMount } from 'svelte'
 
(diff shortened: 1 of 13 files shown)


client/web-sveltekit/src/lib/CodeMirrorBlob.svelte
```diff
@@ -152,21 +151,34 @@
...
+        selectableLineNumbers: null,
+        syntaxHighlighting: null,
+        lineWrapping: null,
+        temporaryTooltip,
+        codeIntelExtension: null,
+        staticExtensions,
+        staticHighlightExtension: null,
+        blameDataExtension: null,
+        blameColumnExtension: null,
+    })
```
fkling, line 178: I had introduced this helper function when refactoring the search input and it seems to make sense to use it here too to get more fine grained control over updating individual extensions.

client/web/src/repo/blob/codemirror/linenumbers.ts
fkling, line 1: The changes in this file make it possible to set the initial selected line without scrolling it into view. It was necessary to do this otherwise the selected line would be scrolled into view after restoring the scroll position (most likely because `requestAnimationFrame` schedules the update to be run after the scroll restoration).

client/web-sveltekit/src/lib/codemirror/utils.ts
```diff
@@ -1,59 +1,122 @@
...
+export function restoreScrollSnapshot(view: EditorView, snapshot: ScrollSnapshot): void {
+    if (snapshot.scrollTop !== undefined) {
+        // We are using request measure here to ensure that the DOM has been updated/painted
+        // before updating the scroll position.
+        view.requestMeasure({
+            read() {
+                return null
+            },
+            write(_measure, view) {
+                if (snapshot.scrollTop !== undefined) {
```
camdencheek, line 116: Why do we check if `scrollTop` is undefined again here? Does this request happen asynchronously?

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte
```diff
@@ -23,17 +23,31 @@
...
+        restore(data) {
+            initialScrollPosition = data
+        },
+    }
+
     const combinedBlobData = createBlobDataHandler()
-    let selectedPosition: LineOrPositionOrRange | null = null
     const lineWrap = writable<boolean>(false)
+    let cmblob: CodeMirrorBlob | null = null
+    let blob: Awaited<PageData['blob']> | null = null
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte:47]: TIL about `Awaited`

client/web-sveltekit/src/lib/codemirror/utils.ts
fkling: It seems TypeScript's type refinement is lost across function boundaries.

fkling: Oh but I think that's only because it's a property on a mutable object. I can assign the value to local variable. That would probably be clearer.

camdencheek:
I noticed horizontal scroll for blob view scrolls the whole codemirror blob

https://share.cleanshot.com/fp9KhR0T

camdencheek (approved):
Looks great! Thanks Felix!

fkling:
> I noticed horizontal scroll for blob view scrolls the whole codemirror blob. I think this is unrelated to this PR (also see it on s2), but wanted to point it out in case it was an easy fix with the new changes

Looks like that only happens when the blame view is enabled. I think we "unfix" the line number gutter when it's enabled, so it scrolls, but I don't know anymore why we do this.

PR #62026, opened 2024-04-18
Context: avoid panic on stopwords query

jtibshirani:
In #61848, we tried to fix an issue when the query contained only stopwords. In
this case, we returned all content. This fix introduced a panic by returning a
nil search job, which is anot allowed by the our job framework.

Now, we return a `noopJob` which returns no results. This is the same approach
we use [when an AND/ OR query has no operands](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/internal/search/job/jobutil/job.go?L1042).

## Test plan

My testing was clearly lacking in the previous fix!! This time, I've added new
end-to-end GraphQL tests using `patterntype:codycontext`. Also tested manually
and checked there are no errors or panics in logs.

### DIFF
--- a/internal/search/codycontext/job.go
+++ b/internal/search/codycontext/job.go
@@ -20,10 +20,14 @@ func NewSearchJob(plan query.Plan, newJob func(query.Basic) (job.Job, error)) (j
 	basicQuery := plan[0].ToParseTree()
 	q, err := queryStringToKeywordQuery(query.StringHuman(basicQuery))
 
-	// If there are no patterns left, this query was entirely composed of
-	// stopwords, so we return no results.
-	if err != nil || len(q.patterns) == 0 {
+	if err != nil {
 		return nil, err
+	}
+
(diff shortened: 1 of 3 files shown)


internal/search/codycontext/job.go
```diff
@@ -70,3 +74,18 @@ func (j *searchJob) MapChildren(fn job.MapFunc) job.Job {
 	cp.child = job.Map(j.child, fn)
 	return &cp
 }
+
+func newNoopJob() *noopJob {
```
jtibshirani, line 78: We already have `NewNoopJob` in `jobutil`, but using it would introduce an import cycle. I opted to create a new private one since it's really simple, and it's nice to keep this Cody context logic very separate.

jtibshirani:
Thanks for the speedy review! I'm going to merge + backport as I want to make the patch release Monday.

PR #62027, opened 2024-04-19
Graph: Support JSX syntax highlighting

mmanela:
Fixes ENG-20458

Adds support to highlight JSX files.  Adds new SCM file for jsx.

image

## Test plan
- Added unit test
- Manual testing


docker-images/syntax-highlighter/crates/syntax-analysis/queries/jsx/highlights.scm
```diff
@@ -0,0 +1,2 @@
+;; This file inherits from javascript/highlights.scm
+(jsx_attribute (property_identifier) @identifier.attribute)
```
mmanela, line 2: @varungandhi-src This is the only thing we were specifically highlighting in our tsx file so I kept this the same here

varungandhi-src (changes requested):
For the source of truth, we rely on languages.yml in Linguist, because language detection going forward is based on go-enry, which in turn is based on Linguist. Yes, there is some old code relying on language detection from syntect, but we want to rely on that less with time.

https://sourcegraph.com/github.com/github-linguist/linguist/-/blob/lib/linguist/languages.yml

You'll notice that TSX and TypeScript are listed as separate languages there, whereas the 'jsx' file extension is present on JavaScript. So we should not introduce a new language/ParserId for JSX here, we should instead:
1. Add the JSX-specific query to the JavaScript queries
2. Make sure that `.jsx` files are getting mapped to `JavaScript` in the language detection code in our Go wrapper code for go-enry in `package languages`.

When `JavaScript` (or `javascript`) is passed as the language from the Go code, we should use the correct Tree-sitter grammar automatically.

internal/highlight/language.go
```diff
@@ -95,7 +95,6 @@ var highlightConfig = syntaxHighlightConfig{
 
 var baseHighlightConfig = syntaxHighlightConfig{
 	Extensions: map[string]string{
-		"jsx":  "jsx", // default `getLanguage()` helper doesn't handle JSX
```
mmanela, line 98: This is not needed, since go-enry properly returns jsx as javascript which is what we want

varungandhi-src:
I noticed that the SCIP kind was copied from tsx/highlights.scm and it was incorrect there too. So I've fixed the SCIP kinds (see scip.proto for descriptions https://github.com/sourcegraph/scip/blob/main/scip.proto#L582-L583). It looks like `@identifier.attribute` is actually being misused in a bunch of queries, we should fix those later.

Updated screenshots in light mode and dark mode.

CleanShot 2024-04-22 at 12 00 54@2x
CleanShot 2024-04-22 at 11 58 54@2x

internal/highlight/language_test.go
```diff
@@ -155,3 +155,36 @@ func TestGetLanguageFromContent(t *testing.T) {
...
 	}
 }
+
+func TestGetLanguageFromPath(t *testing.T) {
+	type testCase struct {
+		Filename string
+		Expected string
+	}
+
+	cases := []testCase{
```
varungandhi-src, line 165:
```suggestion
	// JSX is mapped to javascript but TSX is mapped to 'tsx' language
	// because Linguist contains a separate language for TSX but not JSX
	cases := []testCase{
```

varungandhi-src (approved):

PR #62039, opened 2024-04-19
Add a screen to check if user is logged in with the same account on web as cody client

thenamankumar:
Context: https://sourcegraph.slack.com/archives/C05AGQYD528/p1713500117867789?thread_ts=1713416841.706729&cid=C05AGQYD528

VS Code PR: https://github.com/sourcegraph/cody/pull/3868

E2E Demo: https://www.loom.com/share/7b525676583b4ad6aa38972a8138f4cd?sid=a24403de-8c0f-4bba-b343-2f9214651402

The VS Code PR adds a query parameter `cody_client_user=<username>` while redirecting to `/cody/manage`. 

In this PR, we check for the query param on the web and verify if the user is logged in with the same account. If not, we redirect them to the switch account screen. The screen gives them a disclaimer and option to either sign out and switch to the correct account or to continue with the one they are logged in right now.

Screenshot:
image

## Test plan
- checkout the vs code [PR](https://github.com/sourcegraph/cody/pull/3868) locally, build the extension and connect to the local sg instance
- signing to account A in the VS Code extension.
- got to sourcegraph.test:3443, signout and sign in into account B.
- Go to Cody in VS Code, click on account -> manage account -> redirect
- You should be redirected to the switch account page. 
- Test both continue and signout actions. 
- After signout on web, signin with account A. 
- Go to Cody in VS Code again, click on account -> manage account -> redirect
- This time you should land at `/cody/manage` and the switch account screen should not be visible.


toolmantim (approved):
Nice! 👍🏼 from a design POV

unknwon (approved):
Demo looks great!

x-post a nit on the wording:

CleanShot 2024-04-22 at 14 00 35@2x

client/web/src/cody/management/CodyManagementPage.tsx
```diff
@@ -57,15 +57,23 @@ export const CodyManagementPage: React.FunctionComponent<CodyManagementPageProps
...
     const parameters = useSearchParameters()
 
     const utm_source = parameters.get('utm_source')
-
     useEffect(() => {
         eventLogger.log(EventName.CODY_MANAGEMENT_PAGE_VIEWED, { utm_source })
         telemetryRecorder.recordEvent('cody.management', 'view')
     }, [utm_source, telemetryRecorder])
 
+    const codyClientUser = parameters.get('cody_client_user')
```
chrsmith, line 69: Could you add a doc comment here to explain what this is all about? It isn't overly confusing since `/cody/switch-account` provides a strong hint. But it would be very useful to call out:

- The URL query param is added by the VS Code extension. (And hopefully other extensions in the future.)
- We do this to solve the problem where a use is logged into their IDE as a different user account than their browser. (Leading to confusion regarding why one is Pro but the other isn't, etc.)

client/web/src/cody/management/CodyManagementPage.tsx
```diff
@@ -99,6 +105,10 @@ export const CodyManagementPage: React.FunctionComponent<CodyManagementPageProps
         telemetryRecorder.recordEvent('cody.management.upgradeToProCTA', 'click')
     }, [telemetryRecorder])
 
+    if (accountSwitchRequired) {
+        return null
```
chrsmith, line 109: ⚠️ Is this safe? Suppose for whatever reason the automatic navigation didn't take effect. The user would then be presented with a blank page and no explanation.

Would it be safer to just do something like:

```jsx
return <><p>Different account detected. Switching to disambiguation page...</p><>
```

client/web/src/cody/switch-account/CodySwitchAccountPage.tsx
```diff
@@ -0,0 +1,82 @@
...
+        telemetryRecorder.recordEvent('cody.switch-account', 'view')
+    }, [telemetryRecorder])
+
+    const navigate = useNavigate()
+    const { username = '' } = useParams()
+
+    const accountSwitchNotRequired = !username || !authenticatedUser || authenticatedUser.username === username
+    useEffect(() => {
+        if (accountSwitchNotRequired) {
+            navigate(PageRoutes.CodyManagement)
```
chrsmith, line 36: This seems a little dangerous, having two different pages potentially redirect to each other.

Because this requires that the check is identical, perhaps we should tease out the `accountSwitchNotRequired`/`accountSwitchRequired` logic into a separate function? So we can use the same thing in both pages?

client/web/src/cody/switch-account/CodySwitchAccountPage.tsx
```diff
@@ -0,0 +1,82 @@
...
+        }
+    }, [accountSwitchNotRequired, navigate])
+
+    if (accountSwitchNotRequired || !authenticatedUser) {
+        return null
```
chrsmith, line 41: Same feedback as before. Do we need to worry about the navigation/redirect not taking effect and just rendering `null` leading to the user being "stuck"?

chrsmith (approved):
The solution looks good to me, and I think this is going to be SUPER helpful for reducing user confusion when logged in as different accounts!

Just a few nitpicks and questions about the best way to land this change. (e.g. having a single location where we put the logic for "user does/does not need to redirect".) Because if we aren't careful, this can break down in some unfortunate ways (maybe?)

chenkc805:
@thenamankumar @chrsmith what's holding us back from deploying this? Looks like everyone approved

unknwon:
@thenamankumar - Have you communicated with @chrsmith through other channel regarding his feedback? (e.g. Slack DM, and if so, appreciate next time you can explicitly mention that 🙏 ) Otherwise, ignoring his feedback and question does not seem like a health code review culture to build upon.

client/web/src/cody/management/CodyManagementPage.tsx
thenamankumar: Updating here: https://github.com/sourcegraph/sourcegraph/pull/62247

client/web/src/cody/management/CodyManagementPage.tsx
thenamankumar: We do not need to worry about navigation.

client/web/src/cody/switch-account/CodySwitchAccountPage.tsx
thenamankumar: They both have the opposite & exclusive logic to get redirected. Only one can be true. The logic is well tested. Need not worry.

client/web/src/cody/switch-account/CodySwitchAccountPage.tsx
thenamankumar: Same response as above: https://github.com/sourcegraph/sourcegraph/pull/62039#discussion_r1583364559

thenamankumar:
Thanks, @unknwon, for being vigilant here. Due to mey be some GitHub bug, I did not see any of Chris's comments at the time I merged the PR.  Maybe because my browser tab wasn't updated in a long time. 

I have addressed all his concerns now.

PR #62044, opened 2024-04-19
Pin cURL version and update packages for release

willdollman:
* Pin curl and libcurl to 8.6.0-r0 due to a regression introduced in 8.7.1-r0 that affects gitserver
  * Pinning an older version of curl introduces CVE-2024-2398. This is rated Medium (though trivy reports it as high 🤔) and not exploitable  
* Update all other package to latest versions

cc @eseliger 

## Test plan

- Buildkite CI

### DIFF
--- a/wolfi-images/sourcegraph-template.yaml
+++ b/wolfi-images/sourcegraph-template.yaml
@@ -16,9 +16,12 @@ contents:
 
     # TODO: Dev tools - remove in future release
     - busybox
-    - curl
+    - curl=8.6.0-r0 # TODO(will): Remove pin after resolving git fetch bug
     - wget
     - bind-tools
+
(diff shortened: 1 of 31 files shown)


wolfi-images/sourcegraph-template.yaml
```diff
@@ -16,10 +16,13 @@ contents:
...
     # TODO: Dev tools - remove in future release
     - busybox
-    - curl
+    - curl=8.6.0-r0 # TODO(will): Remove pin after resolving git fetch bug
     - wget
     - bind-tools
 
+    # TODO(will): Remove explicit add of libcurl after resolving git fetch bug
+    - libcurl-openssl4=8.6.0-r0
+
```
willdollman, line 25: Pin curl and libcurl versions

PR #62046, opened 2024-04-19
Publish a cloud-mi2 base image for use in the cloud repo

willdollman:
We no longer publish this image either, and it's used by the cloud team for several of their containers.

## Test plan

- CI


michaellzc (approved):
ty

PR #62047, opened 2024-04-19
Pin curl packages and update base images

willdollman:
@eseliger and I [identified a regression](https://sourcegraph.slack.com/archives/C1JH2BEHZ/p1713493749089869) in gitserver caused by libcurl 8.7.1, which was not present in 8.6.0.

To fix, pin the curl and libcurl packages to an older version. Although this reintroduces a medium-severity CVE, but this is an acceptable tradeoff due to its medium severity and lack of exploitability.

This PR also updates all the base images to pull in updated packages.

## Test plan

- CI
- Check `curl --version` output

### DIFF
--- a/wolfi-images/sourcegraph-base.yaml
+++ b/wolfi-images/sourcegraph-base.yaml
@@ -16,9 +16,12 @@ contents:
 
     # TODO: Dev tools - remove in future release
     - busybox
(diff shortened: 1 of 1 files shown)


eseliger (approved):
Trust 😬

PR #62048, opened 2024-04-19
Require minimum compatible Cody client version if Cody Ignore enabled

taras-yemets:
Part of https://github.com/sourcegraph/sourcegraph/issues/61913

We enforce clients' minimum required version to ensure that Cody clients respect the `"cody.contextFilters"` site config parameter (more context in [this](https://sourcegraph.slack.com/archives/C02UC4WUX1Q/p1713272704235389) thread).

To achieve that, we modify the completions HTTP handler to require client name and version query params in the request if `"cody.contextFilters"` are defined in the site config.
We do not modify completions GraphQL query as it's not used by Cody clients and is deprecated (https://github.com/sourcegraph/sourcegraph/pull/61381).

The completion handler returns an error in the following cases:
- either client name or version is not present in query params ("old" Cody clients versions without Cody Ignore support)
- client doesn't have Cody Ignore implementation and thus is not supported (clients other than vscode/jetbrains/web)
- client is supported (vscode/jetbrains/web), but its version doesn't satisfy the version constraint. 

Follow-up work on the Cody clients side (vscode/jetbrains/web):
- [ ] add support for client name and version query params
- [ ] provide descriptive message to users about handle `ClientCodyIgnoreCompatibilityError` error
- [ ] release new versions of the IDE extension before the May release so that more users can update and get descriptive error message

## Test plan
- CI
- Added unit tests
- Tested in Postman. The screenshots below are for `/.api/completions/stream` endpoint, but it works the same for the `/.api/completions/code` endpoint.

| Name | Screenshot |
| -- | -- |
| Both client name and version missing | Screenshot 2024-04-19 at 18 32 34|
| Unsupported client | Screenshot 2024-04-19 at 18 33 31|  
| Supported client, but missing version | Screenshot 2024-04-19 at 18 32 28|
| Supported client, version doesn't satisfy constraint | Screenshot 2024-04-19 at 18 32 18Screenshot 2024-04-19 at 18 32 55Screenshot 2024-04-19 at 18 33 06|
| Supported client and version | Screenshot 2024-04-19 at 18 32 10Screenshot 2024-04-19 at 18 33 21|
| Web, version is not required |Screenshot 2024-04-19 at 18 42 04 Screenshot 2024-04-19 at 18 33 58|


internal/completions/types/types.go
```diff
@@ -178,6 +178,21 @@ const (
...
+	CodyClientWeb       CodyClientName = "web"
+	CodyClientVscode    CodyClientName = "vscode"
+	CodyClientJetbrains CodyClientName = "jetbrains"
+)
+
+type CodyClientVersionConstraint string
+
+const (
+	VscodeVersionConstraint    CodyClientVersionConstraint = "> 1.14.0"
+	JetbrainsVersionConstraint CodyClientVersionConstraint = "> 5.5.5"
```
taras-yemets, line 193: TODO: replace with versions with actual Cody Ignore support

internal/completions/httpapi/handler.go
```diff
@@ -62,6 +69,7 @@ func newCompletionsHandler(
 	responseHandler := newSwitchingResponseHandler(logger, db, feature)
 
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		fmt.Printf("!!!!! newCompletionsHandler request: %+v\n", r)
```
chrsmith, line 72: Nit: I assume you just had this for debugging :)

internal/completions/httpapi/handler.go
```diff
@@ -40,6 +41,12 @@ import (
 // being cancelled as DeadlineExceeded.
 const maxRequestDuration = 8 * time.Minute
 
+// versionConstraintErrorPrefix value is used to identify specific errors in the Cody clients codebases.
+// When changing its value be sure to update the clients code.
+const versionConstraintErrorPrefix = "UnsupportedClient"
+
+var codyClientNotSupportedError = errors.New(fmt.Sprintf("%s: please use one of the supported clients: %s, %s.", versionConstraintErrorPrefix, types.CodyClientVscode, types.CodyClientJetbrains))
```
chrsmith, line 48: Stylistic nit-pick: Having a package-level variable that isn't a string literal seems a off, IMHO.

What do you think about something like this, instead:

```go
// codyIgnoreRequiredClients defines the only clients that are able to send requests
// to the completions API. (This "check" is done by looking at a URL-query parameter,
// and so it cannot be used for security. And only confirming the client is at
// a certain version number.)
const codyIgnoreRequiredClients = []string{
     types.CodyClientVscode,
     types.CodyClientsJetbrains,
}

// assertClientCanUseCodyIgnore checks that the calling client is one on the
// allow list of those able to use the Cody Ignore feature. (Since other clients,
// or say, `curl`, are not providing the necessary Cody Ignore-specific metadata.)
func assertClientCanUseCodyIgnore(callingClient string) error {
    ...
}

File: internal/completions/httpapi/handler.go [hunk 4]
```diff
@@ -92,6 +100,47 @@ func newCompletionsHandler(
 			return
 		}
 
+		isDotcom := dotcom.SourcegraphDotComMode()
+		if !isDotcom && conf.SiteConfig().CodyContextFilters != nil {
+			clientName := r.URL.Query().Get("client-name")
+			if clientName == "" {
+				http.Error(w, fmt.Sprintf("%s: \"client-name\" query param is required.", versionConstraintErrorPrefix), http.StatusNotAcceptable)
```
chrsmith, line 107: Nit:

IMHO, it would help readability to move the `fmt.Sprintf(...)` call to a local variable. Also, as a very unimportant and personal style preference, you can replace the `\"` with just using `%q`. It's not a big change, but to some broken individuals, is easier on the eyes.

```go
msg := fmt.Sprintf("%s: %q query param is required.", "client-name", versionConstraintErrorPrefix)
http.Error(w, msg, http.StatusNotAcceptable)
```

chrsmith, line 107: 👍 for using `StatusNotAcceptable`!

#TIL that 406 Status Not Acceptable is _exactly_ the right status code for this situation.

For those not in the know, like me: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406
> The HyperText Transfer Protocol (HTTP) 406 Not Acceptable client error response code indicates that the server cannot produce a response matching the list of acceptable values defined in the request's proactive [content negotiation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation) headers, and that the server is unwilling to supply a default representation.
> 
> Proactive content negotiation headers include:
> 
> [Accept](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept)
> [Accept-Encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding)
> [Accept-Language](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language)

internal/completions/httpapi/handler.go
```diff
@@ -92,6 +100,47 @@ func newCompletionsHandler(
...
+				return
+			}
+			cvc, err := newClientVersionConstraint(types.CodyClientName(clientName))
```
chrsmith, line 110: Not to split hairs, but since in both cases a non-`nil` error is being surfaced to the user. Why not just return a specialized `interface ErrorWithHTTPStatusCode {...}`. Or have the function return a 3-tuple, including the `errStatusCode`, and `err`. So you just need to call `http.Error(w, err.Error(), errStatusCode)`.

Not a big deal, just a suggestion to (ever so marginally, and _maybe_) improve the code. Feel free to totally ignore.

internal/completions/httpapi/handler.go
```diff
@@ -92,6 +100,47 @@ func newCompletionsHandler(
...
+					http.Error(w, err.Error(), http.StatusNotAcceptable)
+					return
+				}
+
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
+			if cvc != nil {
+				clientVersion := r.URL.Query().Get("client-version")
+				if clientVersion == "" {
```
chrsmith, line 122: This isn't a lot of code, but it seems unfortunate that we are duplicating out "no seriously, the URL parameter is required" check. What do you think about just having a more specialized `func getRequiredURLQueryParam(name string) (string, errorWithStatusCode)` type of thing?

(This is where IMHO, Golang's http package can get a little clunky. Since needing to call `http.Error(w, ...) \n return` can get a little annoying/tedious for things like this.)

internal/completions/httpapi/handler.go
chrsmith: Actually, looking at the number of times you call 

```go
http.Error(w, x, y)
return
```

It seems like you should put all of this validation logic into a single function, so that you can simplify things to just: `validateRequest(req *http.Request) (int, error)`

And have the contract be, if given a non-nil error it is something that would be suitable for surfacing to the end user. IMHO that would clean things up considerably.

internal/completions/httpapi/handler.go
```diff
@@ -433,3 +481,21 @@ func newNonStreamingResponseHandler(logger log.Logger, db database.DB, feature t
 		_, _ = w.Write(completionBytes)
 	}
 }
+
+type clientVersionConstraint struct {
```
chrsmith, line 485: 👍 for encapsulating the logic here into a specific type. Though could you please add a doc comment to explain how this fits into things?

Yes, the type name is very clear. (🙏) but currently we are only using this for a very narrow use-case. So maybe it could be confused with a more general, required invariant in the codebase? (Feel free to push back, just a thought.)

internal/completions/types/types.go
chrsmith: Yes! We should definitely not have a generic name like `VersionConstraint` not be paired with an explicit comment stating why this constraint is needed, and in which circumstances. Because we do support VS Code clients much older than "1.14.0", right? And this is just for Cody Ignore, specifically?

chrsmith (commented):
As far as solving problem we have today, this seems reasonable. But my hope is that we can have a more robust, generalized solution in the future where we do not specify the "API version and client" in a URL query parameter. (As that's IMHO, clunky, compared to a more standard `Accept` header.)

The code/approach seems fine. All of my PR feedback is just nitpick things about how we can improve or simplify the code. (So please don't take it as criticism, just me trying to help us land the highest-quality PR possible. And if you don't think my suggestions are strictly improvements, feel free to push back or ignore.)

The main thing I would ask you to change though, is just adding more doc comments and clarity around the scope of what these client/version checks are for. We aren't adding a "general" invariant, that we will need to maintain across the entire API surface area. This is instead specific to just the Cody Ignore feature. (Since how we approach solving one vs. the other is different.)

taras-yemets:
Very helpful feedback, @chrsmith! Thank you!

I addressed it in the following way.
1. Moved all the client Cody Ignore compatibility check logic to a separate function.
2. Client version constraints are defined as package-level constants.
I left client name constants in the `types` package in case we want to introduce similar query params on the other endpoints. 
3. Added unit tests.

Can you please review the updated implementation?

internal/completions/httpapi/handler.go
```diff
@@ -433,3 +441,93 @@ func newNonStreamingResponseHandler(logger log.Logger, db database.DB, feature t
...
 }
+
+type clientCodyIgnoreCompatibilityError struct {
+	reason     string
+	statusCode int
+}
+
+// clientCodyIgnoreCompatibilityErrorPrefix value is used to identify specific errors in the Cody clients codebases.
+// When changing its value be sure to update the clients code.
+const clientCodyIgnoreCompatibilityErrorPrefix = "ClientCodyIgnoreCompatibilityError"
```
chrsmith, line 452: Super-minor personal preference nit-pick:

For readability, it's usually an improvement to group all of the `const` and `var` declarations within a Go file. And try to have some sort of "ordering". Like basic types at the top, and more sophisticated types with lots of methods below that. (There aren't any hard and fast rules, just things that may or may not make the code easier to read.)

With that in mind, I'd suggest moving this definition and associated doc comment below to line 461 or so, so you group all of the constants related to these compatibility checks together.

It also has the nice side-effect of having the `func (e *...) Error() string {` definition immediately follow the `type ... struct` definition of `clientCodyIngoreCompatibilityError`. (But yes, defining the `const clientCodyIgnoreCompatibilityErrorPrefix` _before_ it is used in that `Error()` function is "good, more readable" too. 🤷.)

Ideally we'd be able to remove the need to create a package-level `clientCodyIgnoreCompatibilityErrorPrefix` entirely. e.g. can we somehow obviate the need for it by relying on `errors.Is(...)` instead?

internal/completions/httpapi/handler.go
```diff
@@ -433,3 +441,93 @@ func newNonStreamingResponseHandler(logger log.Logger, db database.DB, feature t
...
+// clientCodyIgnoreVersionConstraint represents client version constraint following the semver spec.
+type clientCodyIgnoreVersionConstraint string
+
+const (
+	vscodeCodyIgnoreVersionConstraint    clientCodyIgnoreVersionConstraint = "> 1.14.0"
+	jetbrainsCodyIgnoreVersionConstraint clientCodyIgnoreVersionConstraint = "> 5.5.5"
+)
+
+func checkClientCodyIgnoreCompatibility(r *http.Request) *clientCodyIgnoreCompatibilityError {
+	// If Cody context filters are not defined on the instance, we do not restrict client version
```
chrsmith, line 467: This is overly pedantic, but comments are there to aid in the code's readability. So to maximize that, they should be complete sentences and use proper punctuation.

```suggestion
	// If Cody context filters are not defined on the instance, we do not restrict client version.
	// Because the site hasn't configured Cody Ignore, no need to enforce it.	
```

internal/completions/httpapi/handler.go
```diff
@@ -433,3 +441,93 @@ func newNonStreamingResponseHandler(logger log.Logger, db database.DB, feature t
...
+
+func (e *clientCodyIgnoreCompatibilityError) Error() string {
+	return fmt.Sprintf("%s: %s", clientCodyIgnoreCompatibilityErrorPrefix, e.reason)
+}
+
+// clientCodyIgnoreVersionConstraint represents client version constraint following the semver spec.
+type clientCodyIgnoreVersionConstraint string
+
+const (
+	vscodeCodyIgnoreVersionConstraint    clientCodyIgnoreVersionConstraint = "> 1.14.0"
```
chrsmith, line 462: Suggestion: This is fine as-is, but do we have "client types" defined as an enum somewhere else? If we do, e.g. `type codyClient string`, then perhaps it would be more useful to make this a property of _that_ instead?

e.g.

```go
func (c codyClientEnum) CodyIgnoreVersionConstraint() string {
   switch c {
       case "jet-brains":
            return "> 5.5.5"
       ...
       default:
            return ""
}
```

This isn't a big deal, but introducing a couple of top-level constants like this seems a little fragile. Since I imagine we'd need to extend it with other "supported" clients in the future, as they add support for VS Code.

So in the spirit of minimizing the number of places we'd need to add the logic, we could just have the core check be:

```go
versionReq := clientType.CodyIgnoreVersionConstraint()
if versionReq == "" {
   logger.Warn("Client has no Cody Ignore version constraint check. Ignoring.")
} else {
   ...
}
```

internal/completions/httpapi/handler.go
```diff
@@ -433,3 +441,93 @@ func newNonStreamingResponseHandler(logger log.Logger, db database.DB, feature t
...
+	jetbrainsCodyIgnoreVersionConstraint clientCodyIgnoreVersionConstraint = "> 5.5.5"
+)
+
+func checkClientCodyIgnoreCompatibility(r *http.Request) *clientCodyIgnoreCompatibilityError {
```
chrsmith, line 466: ⚠️ I would suggest adding a doc comment here explaining the purpose of this function. It is _mostly_ clear from the name, but IMHO it would still be helpful for the next person trying to make sense of the code.

In particular, calling out that a non-nil `clientCodyIgnoreCompatibilityError` implies that the HTTP request _should be rejected_ with the error response from the returned `error`. And that any reason strings _are_ safe to be surfaced to the end user.

internal/completions/httpapi/handler_test.go
```diff
@@ -0,0 +1,161 @@
...
+			q: url.Values{
+				"client-name": []string{"sublime"},
+			},
+			want: &clientCodyIgnoreCompatibilityError{
+				reason:     fmt.Sprintf("please use one of the supported clients: %s, %s.", types.CodyClientVscode, types.CodyClientJetbrains),
+				statusCode: http.StatusNotAcceptable,
+			},
+		},
+		{
+			name: "version doesn't follow semver spec",
```
chrsmith, line 84: ⚠️ Could you add another "malformed version" test? e.g. is "1.0" or just "1" allowed by the sem-ver parsing logic? Or must it _always_ contain all three parts of the version? And if there is a discrepancy, do we want to be that strict / flexible? (Seems like something we would reasonably expect a client to maybe be fast and loose with. So having a unit test to confirm that we don't panic or anything is sufficient. We probably don't actually care if it is or is not allowed, as long as the version check that we do perform makes sense.)

internal/completions/httpapi/handler_test.go
```diff
@@ -0,0 +1,161 @@
...
+				"client-name":    []string{string(types.CodyClientJetbrains)},
+				"client-version": []string{"1.14.0"},
+			},
+			want: &clientCodyIgnoreCompatibilityError{
+				reason:     fmt.Sprintf("Cody for %s version \"1.14.0\" doesn't match version constraint \"%s\"", types.CodyClientJetbrains, jetbrainsCodyIgnoreVersionConstraint),
+				statusCode: http.StatusNotAcceptable,
+			},
+		},
+		{
+			name: "jetbrains: version matches constraint",
```
chrsmith, line 129: ⚠️ Could you add some more unit tests to just cover all of the "positive cases" here?

e.g.

- ✅ "major version bump" (6.0.0)
- TODO: "minor version bump" 5.6.0
- TODO: "revision bump" 5.5.6

And similarly, that we have negative cases for all of the above, too. (We may, but it isn't super-clear at a glance.)

internal/completions/httpapi/handler.go
```diff
@@ -92,6 +93,14 @@ func newCompletionsHandler(
 			return
 		}
 
+		isDotcom := dotcom.SourcegraphDotComMode()
```
chrsmith, line 96: ⚠️ Could you please add a doc comment explaining what this block of code is for? For people coming across this code later, it may not be clear what a "cody ignore compatibility" check is for. And if it is actually important.

Since this is registered in a very top-level handler (all completions), it's worth erring on the side of clarity IMO.

internal/completions/httpapi/handler.go
```diff
@@ -92,6 +93,14 @@ func newCompletionsHandler(
...
+		if !isDotcom {
+			err := checkClientCodyIgnoreCompatibility(r)
+			if err != nil {
```
chrsmith, line 99: Stylistic nit-pick. This is arguably cleaner, since the name of the function is relatively short. But please reject if you do not think this is strictly an improvement.

```suggestion
			if err := checkClientCodyIgnoreCompatibility(r); err != nil {
```

internal/completions/httpapi/handler.go
chrsmith: Also, updated the comment for clarity.

chrsmith (approved):
Thanks for making the change, IMHO the codes a lot easier to understand.

There are still some minor ways we can improve readability. But overall the code looks good. All of my feedback is optional except for those flagged with a ⚠️ emoji. (Feel free to push back on those too, but those are things I personally would be more insistent upon.)

You may still want an LGTM from some more more familiar with the Cody Enterprise, server-side piece though. As I might have missed something important.

internal/completions/httpapi/handler.go
```diff
@@ -92,6 +93,14 @@ func newCompletionsHandler(
...
+				http.Error(w, err.Error(), err.statusCode)
```
unknwon, line 100: missing a `return` here?

internal/completions/httpapi/handler.go
unknwon: side nit: if those consts are only being used once... don't see why they have to be declared at the package-level, i.e. just use the literals within the exact places (and context) that are being used right now.

internal/completions/httpapi/handler.go
```diff
@@ -433,3 +441,93 @@ func newNonStreamingResponseHandler(logger log.Logger, db database.DB, feature t
...
+	switch clientName {
+	case types.CodyClientWeb:
+		return nil
+	case types.CodyClientVscode:
+		cvc = clientVersionConstraint{client: clientName, constraint: vscodeCodyIgnoreVersionConstraint}
+	case types.CodyClientJetbrains:
+		cvc = clientVersionConstraint{client: clientName, constraint: jetbrainsCodyIgnoreVersionConstraint}
+	default:
+		return &clientCodyIgnoreCompatibilityError{
+			reason:     fmt.Sprintf("please use one of the supported clients: %s, %s.", types.CodyClientVscode, types.CodyClientJetbrains),
```
unknwon, line 495: Should mention "web" as well?

internal/completions/httpapi/handler.go
```diff
@@ -433,3 +441,93 @@ func newNonStreamingResponseHandler(logger log.Logger, db database.DB, feature t
...
+		return &clientCodyIgnoreCompatibilityError{
+			reason:     "\"client-version\" query param is required.",
+			statusCode: http.StatusNotAcceptable,
+		}
+	}
+
+	c, err := semver.NewConstraint(string(cvc.constraint))
+	if err != nil {
+		return &clientCodyIgnoreCompatibilityError{
+			reason:     fmt.Sprintf("Cody for %s version constraint \"%s\" doesn't follow semver spec.", cvc.client, cvc.constraint),
```
unknwon, line 511: FYI `%q` adds quotes for you.

```suggestion
			reason:     fmt.Sprintf("Cody for %s version constraint %q doesn't follow semver spec.", cvc.client, cvc.constraint),
```

unknwon (commented):
Complement to Chris's review, and a potential bug.

internal/completions/httpapi/handler.go
taras-yemets: @unknwon, these constants are used in tests, but we can use literals there too 👍🏻

taras-yemets: >do we have "client types" defined as an enum somewhere else?

@chrsmith, I looked for some kind of "client type" before ([search](https://sourcegraph.com/search?q=(context:global+r:%5Egithub%5C.com/sourcegraph/sourcegraph$+lang:Go+vscode+OR+jet.*brains)+-file:_test%5C.go$&patternType=regexp&sm=0)) and was surprised not to find any 🤷🏻‍♂️ 
I'll follow @unknwon's suggesting and inline these values.

taras-yemets: @unknwon, I removed these constants and inlined version constraints. Given that these constraints won't change often, is it fine if we inline them in tests too ([search](https://sourcegraph.com/search?q=context:global+r:%5Egithub%5C.com/sourcegraph/sourcegraph$@ty/cody-ignore-enforce-min-supported-client-version+path:handler+1.14.1&patternType=keyword&sm=0))?

internal/completions/httpapi/handler_test.go
taras-yemets: extended test cases in https://github.com/sourcegraph/sourcegraph/pull/62048/commits/ab6317c6b0cbfb18c9e6eaaac0df50327b04856c

internal/completions/httpapi/handler_test.go
taras-yemets: addressed in https://github.com/sourcegraph/sourcegraph/pull/62048/commits/ab6317c6b0cbfb18c9e6eaaac0df50327b04856c

internal/completions/httpapi/handler.go
taras-yemets: Moved the `const` declaration closer to where it's used (https://github.com/sourcegraph/sourcegraph/pull/62048/commits/4be90497429770a0ccabaa859652479414ca1adb). Though we technically don't need a `const` here (we can just use a literal), I still decided to use it mostly for documentation sake: I want to stress enough the need to change the client code in case this prefix changes. WDYT?

internal/completions/httpapi/handler.go
unknwon: I think it's fine to inline them / use literals in this case. 👍

unknwon (approved):

internal/completions/httpapi/handler.go
taras-yemets: addressed in https://github.com/sourcegraph/sourcegraph/pull/62048/commits/8fb753df0b9a9689142177ca4e9fb39f1e756cc6

internal/completions/httpapi/handler.go
taras-yemets: addressed in https://github.com/sourcegraph/sourcegraph/pull/62048/commits/8fb753df0b9a9689142177ca4e9fb39f1e756cc6

internal/completions/httpapi/handler.go
```diff
@@ -442,3 +452,89 @@ func newNonStreamingResponseHandler(logger log.Logger, db database.DB, feature t
 		_, _ = w.Write(completionBytes)
 	}
 }
+
+type clientCodyIgnoreCompatibilityError struct {
```
chrsmith, line 456: Minor nit: Naming is hard, but IMHO the `client` prefix here is unnecessary. And just `codyIgnoreCompatibilityError` is sufficient. What do you think?

internal/completions/httpapi/handler.go
```diff
@@ -442,3 +452,89 @@ func newNonStreamingResponseHandler(logger log.Logger, db database.DB, feature t
...
+}
+
+func (e *clientCodyIgnoreCompatibilityError) Error() string {
+	// prefix value is used to identify specific errors in the Cody clients codebases.
+	// When changing its value be sure to update the clients code.
+	const prefix = "ClientCodyIgnoreCompatibilityError"
+	return fmt.Sprintf("%s: %s", prefix, e.reason)
+}
+
+// checkClientCodyIgnoreCompatibility checks if the client version respects Cody context filters (a.k.a. Cody Ignore) defined in the site config.
```
chrsmith, line 468: Unimportant ultra-pedantic nitpick you should just ignore because you aren't a broken person like me:

The "a.k.a." looks wrong to me. However, I cannot say for certain what is "correct"... maybe "AKA", just because it's an acronym and shorter?

- The AP style guide, AFAICT has you write "aka". https://twitter.com/APStylebook/status/565239073892466689?lang=en
- Wikipedia's style has "a.k.a. or AKA" but never "aka". https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style/Abbreviations

Maybe we use one more consistently in our codebase? Hard to say for sure.
https://github.com/search?q=org%3Asourcegraph%20AKA&type=code
https://github.com/search?q=org%3Asourcegraph+%22a.k.a.%22&type=code

internal/completions/httpapi/handler.go
```diff
@@ -442,3 +452,89 @@ func newNonStreamingResponseHandler(logger log.Logger, db database.DB, feature t
...
+	clientName := types.CodyClientName(r.URL.Query().Get("client-name"))
+	if clientName == "" {
+		return &clientCodyIgnoreCompatibilityError{
+			reason:     "\"client-name\" query param is required.",
+			statusCode: http.StatusNotAcceptable,
+		}
+	}
+
+	// clientVersionConstraint defines the minimum client version required to support Cody Ignore.
+	type clientVersionConstraint struct {
```
chrsmith, line 487: 👍 Out of principle, scoping identifiers as narrowly as possible is "the best thing". So literally having a type that only exists within the context of a function is (IMHO) a nice thing to see in the narrow situations where you can get away with that.

internal/completions/httpapi/handler.go
```diff
@@ -442,3 +452,89 @@ func newNonStreamingResponseHandler(logger log.Logger, db database.DB, feature t
...
+// A non-nil clientCodyIgnoreCompatibilityError implies that the HTTP request should be rejected with the error text and code from the returned error.
+// Error text is safe to be surfaced to end user.
+func checkClientCodyIgnoreCompatibility(r *http.Request) *clientCodyIgnoreCompatibilityError {
+	// If Cody context filters are not defined on the instance, we do not restrict client version.
+	// Because the site hasn't configured Cody Ignore, no need to enforce it.
+	if conf.SiteConfig().CodyContextFilters == nil {
+		return nil
+	}
+
+	clientName := types.CodyClientName(r.URL.Query().Get("client-name"))
```
chrsmith, line 478: Suggestion: Since you make this same check twice, it may be nice to have a small function. (I'd feel strongly about this if you had the logic duplicated 3+ times. But since it's only 2x, it isn't that big of a deal.)

```go
getRequiredQueryParam := func(name string) (string, error) {
    value := r.URL.Query().Get(name)
    if value == "" {
        return "", &clientCodyIgnoreCompatibilityError{
			reason:     fmt.Sprintf("%q query param is required", name),
			statusCode: http.StatusNotAcceptable,
		}
    }
    return value, nil
}
```

chrsmith (approved):
Thanks for making the suggested changes. I know it was a lot of work, but the PR that we have here seems (IMO) a lot cleaner, easier to follow, etc.! :shipit: 🚀

taras-yemets:
@chrsmith, @unknwon,  
Thanks for your reviews! I learn a lot from you!

internal/completions/httpapi/handler.go
taras-yemets: Addressed in https://github.com/sourcegraph/sourcegraph/pull/62048/commits/2956d79d31df0e0c8566f338e0f5c44c72b7edd3

internal/completions/httpapi/handler.go
taras-yemets: Since it's only repeated 2 times I'll leave it as it is.

internal/completions/httpapi/handler.go
taras-yemets: Addressed in https://github.com/sourcegraph/sourcegraph/pull/62048/commits/98025fe8d628b41b9402fb86141be45ce6582404

unknwon:
Nice work and let's ship! 🚀

PR #62058, opened 2024-04-19
WIP: Move SSC frontend to the monorepo

vdavid:
Little progress with moving the SSC front end to the monorepo.

Caveats / WIP parts:
- Three major components missing
- Need to replace Tailwind styling with CSS (Cody or ChatGPT should help)
- Need to replace icons with Wildcard icons
- Need to import Stripe through `@sourcegraph/cody-plg` rather than directly
- "refetch" needs to be mocked
- Absolutely untested

## Test plan

Move first, then make sure it works with the mocked back end. Once it works, we can merge this branch behind a feature flag if needed. Without a feature flag, it might be confusing: even if we don't publish the link anywhere, someone might stumble upon it.
But we could also just continue by connecting it to the back end.

### DIFF
--- a/client/cody-plg/package.json
+++ b/client/cody-plg/package.json
@@ -1,0 +1,19 @@
+{
+  "private": true,
+  "name": "@sourcegraph/cody-plg",
+  "version": "0.0.7",
+  "description": "Cody PLG library",
+  "license": "Apache-2.0",
+  "main": "src/index.ts",
+  "sideEffects": false,
+  "dependencies": {
+    "@stripe/react-stripe-js": "^2.4.0",
+    "@stripe/stripe-js": "^2.3.0"
+  },
+  "scripts": {
+    "build": "tsc --build && cp -R src/* dist/",
+    "lint": "pnpm run lint:js",
+    "lint:js": "eslint --cache '**/*.[tj]s?(x)'",
+    "test": "vitest"
+  }
+}
--- a/client/web/src/cody/plan/CodyPlanPage.tsx
+++ b/client/web/src/cody/plan/CodyPlanPage.tsx
@@ -1,153 +1,202 @@
-import React, { useCallback, useEffect } from 'react'
+'use client'
 
-import { mdiCreditCardOutline } from '@mdi/js'
-import classNames from 'classnames'
-import { useNavigate } from 'react-router-dom'
+import { InvoiceHistoryCard } from './InvoiceHistoryCard.tsx' // TODO: Move it
+import { NewSignupCard } from './NewSignupCard.tsx' // TODO: Move it
+import { SubscriptionDetailsCard } from './SubscriptionDetailsCard.tsx' // TODO: Move it
 
-import { useQuery } from '@sourcegraph/http-client'
-import type { TelemetryV2Props } from '@sourcegraph/shared/src/telemetry'
-import { ButtonLink, H1, H2, Icon, Link, PageHeader, Text, useSearchParameters } from '@sourcegraph/wildcard'
+import * as stripeJs from '@stripe/stripe-js' // TODO: Export these from @sourcegraph/cody-plg, and import that here
+import { Elements } from '@stripe/react-stripe-js' // TODO: Export from @sourcegraph/cody-plg, and import it here
+import { type FunctionComponent, useState, useEffect } from 'react'
+import { LoadingPage } from './LoadingPage'
+import { BackIcon } from './BackIcon'
+import { H1, Link, Text } from '@sourcegraph/wildcard'
+import { useTemporarySetting } from '@sourcegraph/shared/src/settings/temporary'
 
-import type { AuthenticatedUser } from '../../auth'
-import { Page } from '../../components/Page'
-import { PageTitle } from '../../components/PageTitle'
-import {
-    type UserCodyPlanResult,
-    type UserCodyPlanVariables,
-    CodySubscriptionPlan,
-} from '../../graphql-operations'
-import { CodyProIcon, DashboardIcon } from '../components/CodyIcon'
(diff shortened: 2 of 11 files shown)


client/cody-plg/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "private": true,
+  "name": "@sourcegraph/cody-plg",
```
chrsmith, line 3: I'd suggest we try to avoid using the phrase "Cody PLG" **wherever** possible. It just isn't a good, descriptive name. (It's a strategy, not an actual product or component.)

Instead, I'd suggest we use something like `cody-pro`. That is, it is functionality specific to the Cody Pro SKU of the Cody Product. (Separate from, say, `Cody-enterprise`.) That however, is still misleading, because "Cody Free" is another SKU too... but it's much closer to `cody-plg` than `Cody-enterprise`.


Regardless, I'd love to avoid putting `cody-plg` into a top level folder name if we can avoid it.

client/web/src/cody/plan/CodyPlanPage.tsx
```diff
@@ -0,0 +1,202 @@
...
+// NOTE: Call loadStripe outside a component’s render to avoid recreating the object.
+const stripePromise = stripeJs.loadStripe(publishableKey)
+
+/**
+ * Subscription page.
+ *
+ * Figma Design:
+ * https://www.figma.com/file/FMSdn1oKccJRHQPgf7053o/Cody-GA?type=design&node-id=3496-3651&mode=design&t=uuQfSmhMZEoVRBIG-0
+ */
+export const CodyPlanPage: FunctionComponent = () => {
```
chrsmith, line 31: What is the distinction between `client/web/src/cody/...tsx` and `client/cody-plg/package.json`?

Would we want this to be in `client/web/src/cody-plg` (or `cody-pro`) because it is specific to the non-Enterprise queue? (I have no idea how the code is organized here, so perhaps this is a dumb question.)

cmd/frontend/internal/app/ui/router.go
```diff
@@ -145,9 +145,10 @@ func InitRouter(db database.DB) {
 		{path: "/search/cody", name: "cody-search", title: "Search (Cody)", index: false},
 		{path: "/app/coming-soon", name: "app-coming-soon", title: "Coming soon", index: false},
 		{path: "/app/auth/callback", name: "app-auth-callback", title: "Auth callback", index: false},
-		{path: "/cody/manage", name: "cody", title: "Cody Manage", index: false},
-		{path: "/cody/subscription", name: "cody", title: "Cody Pricing", index: false},
-		{path: "/cody/chat", name: "cody", title: "Cody", index: false},
+		{path: "/cody/manage", name: "cody-manage", title: "Cody Manage", index: false},
```
chrsmith, line 148: Would you mind adding some doc comments here, to clarify what things are "in use today" vs. "part of our temporary moving the SSC-specific UI into dotcom"?

It will very likely be a source of confusion, e.g. why _some_ Cody pages are behind a feature flag, and yet others aren't. So trying to head that off and make that clear during this (hopefully brief 🤞) transition period would be helpful.

chrsmith (commented):
Some minor suggestions about how to make sure we can minimize confusion during the transition.

client/cody-plg/package.json
chrsmith: Thinking about this a little bit more, I still don't like having `cody-plg` in the code, but this probably is the best name.

Since we do _not_ want to hard-code the UI based on the SKU. (Since e.g. suppose we have an Enterprise feature that we later want to make available to Cody Pro. That shouldn't require refactoring the codebase! Only just changing some flags.)

However, this PR and the current set of UI elements we are migrating are _not_ about features per-say. But really, the crux of the entire "Cody Pro" experience. So yeah, `cody-plg` is probably the best term to use. #namingishard

chrsmith (changes requested):
Hey @vdavid, so I spent the day poking at the sg/sg repo and trying to get things wired in. And I think we should start with a different approach for the "first step" in migrating the SSC UI. (I'll send it out for review shortly.)

There are a few things that I think we want to do differently.

**We shouldn't fuss with a separate NPM package for now**

While we may want to do this eventually, I don't think we are ready to try and have an NPM package separate the "Cody Pro" and "Cody Enterprise" UI. Because they are _already_ deeply intertwined, and kinda broken. So adding an NPM package in the mix would just add overhead without actually improving the situation.

For example, the `[cody/subscription](https://sourcegraph.com/cody/subscription)` page shows the Cody SKUs side by side. (Which is something we'd want on the "Cody Pro" site, but not "Cody Enterprise"... however if you try view that same page on S2 (https://sourcegraph.sourcegraph.com/cody/subscription) you get a nice 500 error.

Similarly, the "Cody Dashboard" page acts/behaves totally differently for Cody Pro and Cody Enterprise, too.

So as far as migrating the ~3 UI pages from the SSC UI into the monorepo, I think we shouldn't try to isolate them in their own NPM package. (As it doesn't really provide any benefit.)

And instead, try to get the basics right about which pages should serve a 404 vs. be displayed. Clearly delineating any SKU-specific functionality or APIs, etc.

**Is a "site config setting" preferable to "temporary setting"?**

I may totally be wrong, but from what I could find about "[Temporary Settings](https://docs.sourcegraph.com/dev/background-information/web/temporary_settings)", I don't know if they are the right fit for what we are trying to enable. (And neither is a feature flag for that matter.)

Just having a boolean site configuration option, e.g. `ssc.embeddedUIEnabled` seemed like the least complicated option. As well as being more easily accessible (via `window.context`) rather than a React hook. (Maybe?)

vdavid:
Closing this as obsolete

PR #62066, opened 2024-04-20
KeywordSearch: Honor default search pattern type if present

### LINKED ISSUE #61937: 'defaultPatternType' setting is no longer respected
- **Sourcegraph version:** v5.3.0+ (was working as of 5.2.4)
- **Platform information:** N/A

#### Steps to reproduce:

1. Add `"search.defaultPatternType": "regexp",` to user settings 
2. Go to Search Home
3. Notice that the regexp icon is not colored in
4. Perform a search and notice that `patternType=keyword` is appended to the URL and the search term is not treated as regexp
5. Add `"search.defaultPatternType": "literal",` or `"search.defaultPatternType": "regexp",` to Global Settings in Site Admin
6. Perform a search and notice that the pattern type still defaults to keyword search and the search pattern specified in settings is not respected

#### Expected behavior:
The search pattern specified in user settings and global settings will be set as the default

#### Actual behavior:
Search pattern is always defaulted to keyword search

mmanela:
Fixes #61937 

It looks like we hardcoded keyword if the feature was enabled. This change honors an existing default

 

## Test plan
- Manual testing of different permutations of default and no-default settings

### DIFF
--- a/client/web/src/stores/navbarSearchQueryState.test.ts
+++ b/client/web/src/stores/navbarSearchQueryState.test.ts
@@ -157,7 +157,7 @@ describe('navbar query state', () => {
             expect(useNavbarQueryState.getState()).toHaveProperty('searchPatternType', SearchPatternType.regexp)
         })
 
-        it('prefers keyword search over user settings if keyword search is enabled', () => {
+        it('honors default user settings even if keyword search is enabled', () => {
             setQueryStateFromURL(parseSearchURL(''))
             setQueryStateFromSettings({
                 subjects: [],
--- a/client/web/src/stores/navbarSearchQueryState.test.ts
+++ b/client/web/src/stores/navbarSearchQueryState.test.ts
@@ -162,7 +162,7 @@ describe('navbar query state', () => {
             setQueryStateFromSettings({
                 subjects: [],
                 final: {
-                    'search.defaultPatternType': SearchPatternType.standard,
+                    'search.defaultPatternType': SearchPatternType.regexp,
                     experimentalFeatures: {
                         keywordSearch: true,
                     },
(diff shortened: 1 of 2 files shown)


client/web/src/util/settings.ts
```diff
@@ -61,12 +61,14 @@ export function defaultSearchModeFromSettings(settingsCascade: SettingsCascadeOr
...
-        return SearchPatternType.keyword
-    }
-
-    return getFromSettings(settingsCascade, 'search.defaultPatternType')
+    const defaultPatternType: SearchPatternType | undefined = getFromSettings(
+        settingsCascade,
+        'search.defaultPatternType'
+    )
+    // When the 'keyword search' language update is enabled, default to the 'keyword' patterntype if none set
+    return isKeywordSearchEnabled(settingsCascade)
```
mmanela, line 69: @jtibshirani  I think the right fix is to just return default and not do any keyword search check here and assume no patternType means keyword server side, but that may have to wait until we GA the feature

client/web/src/stores/navbarSearchQueryState.test.ts
```diff
@@ -162,15 +162,15 @@ describe('navbar query state', () => {
             setQueryStateFromSettings({
```
mmanela, line 162: @stefanhengl  @jtibshirani  So, this behavior that I am "fixing" seems to have been explicitly added based on this unit test.  What was the rational for keyword to trump the default setting the user explicitly configured?

jtibshirani: I believe this was a mistake on my part in the original PR (https://github.com/sourcegraph/sourcegraph/pull/59779), and later we added a new test codifying the behavior. I'm +1 to change this, it is always incorrect for us to ignore user settings.

client/web/src/util/settings.ts
jtibshirani: Indeed that's the best approach, but agree we need to wait!

Tiny suggestion about this: I find chained `?` and `??` harder to read, maybe we could be boring and use an if statement.

jtibshirani (approved):

PR #62074, opened 2024-04-22
chore: Bump dependencies of progress-bot

eseliger:
We have a few dependabot alerts with High severity all from this subpackage that we never seem to upgrade. This is a very well isolated package, so there should be little risk from upgrading.

Test plan:

Still compiles.


burmudar (approved):

evict (approved):
Thank you!

PR #62087, opened 2024-04-22
Fixing cache methods to use the right method to retrieve key value stores

arafatkatze:
Fixing some methods for cache selection of key values stores for redis. 

## Test plan
- I tested it on the instance


camdencheek (approved):
Should we also do this for `FIFOList`, which is the other caller of `kv()`? Then we can get rid of `kv()` entirely.

arafatkatze:
I merged this now because I really wanted to see something to test on S2 tomorrow morning but I agree with you camden.

I made an issue https://github.com/sourcegraph/sourcegraph/issues/62091 because I am too strapped to pick this up now

Issue #62057, opened 2024-04-19
Cody Gateway: Update Tokenizer


abeatrix:
CONTEXT: https://sourcegraph.slack.com/archives/C05AGQYD528/p1713491201142189

Cody's client has recently updated to use a `cl100k_base` tokenizer for their tokenization process, while Cody Gateway is using the tokenizer for Claude 2.

To optimize performance and ensure compatibility, we should update the tokenizer on the Cody Gateway side to `cl100k_base` encoding, aligning it with the client-side configuration, in the hope of streamlining the tokenization process and eliminating potential discrepancies between Cody Gateway and clients.

PR #62088, opened 2024-04-22
Cody PLG: Update tokenizer encoding 

abeatrix:
CLOSE https://github.com/sourcegraph/sourcegraph/issues/62057
BACKGROUND CONTEXT: https://sourcegraph.slack.com/archives/C05AGQYD528/p1713491201142189

Cody's client has recently standardized the usage of the `cl100k_base` tokenizer for their token counting process, while Cody Gateway is using different tokenizers for different models, for example: 
- Claude 2 tokenizer for Claude 2 and Claude 3 models
- `cl100k_base` tokenizer for the OpenAI models that we currently support
- no tokenizer is used for firework models

This introduces mismatches between the token count returned by the client and gateway. To optimize performance and ensure compatibility, this PR updates the tokenizer used on the server to cl100k_base encoding, aligning it with the client-side configuration, in the hope of streamlining the tokenization process and eliminating potential discrepancies between Cody Gateway and clients.

- Add [tiktoken_loader](https://github.com/pkoukk/tiktoken-go-loader) as a dependency for embedding encodings to avoid making additional network requests at run time. 
  - This also addresses an issue detailed in https://sourcegraph.slack.com/archives/C04MYFW01NV/p1713828488854649?thread_ts=1713828011.479819&cid=C04MYFW01NV where the current implementation of the Tokenizer does not work in a sandbox environment
- The same tokenizer will be used for all models, including firework models that did not have a tokenizer before.

## Test plan

Updated existing tests to use the updated tokenizer.


cmd/cody-gateway/internal/httpapi/completions/anthropicmessages_test.go
```diff
@@ -156,7 +156,7 @@ func TestIsFlaggedAnthropicMessagesRequest(t *testing.T) {
 		require.True(t, result.IsFlagged())
 		require.True(t, result.shouldBlock)
 		require.Contains(t, result.reasons, "high_prompt_token_count")
-		require.Equal(t, result.promptTokenCount, validPreambleTokens+4+cfg.PromptTokenBlockingLimit+4)
```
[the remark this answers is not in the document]
abeatrix: @arafatkatze can you clarify where this is from and if changing it to +3 looks correct to you

arafatkatze (approved):

cmd/cody-gateway/internal/httpapi/completions/anthropicmessages_test.go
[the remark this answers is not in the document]
chrsmith: Regardless of "where", the +3 or +4 is specific to the tokenizer. So this test is essentially confirming that the prompt counts are "stable" and will break if we ever unintentionally change the tokenizer used.

So I don't know if it is worth tracking down where those 6x tokens come from. But I would guess this line of now deleted code had something to do with it:

```go
numTokens += 3 // every reply is primed with <|start|>assistant<|message|>
```

internal/completions/tokenizer/tokenizer.go
```diff
@@ -41,26 +42,16 @@ type tiktokenTokenizer struct {
...
+// NumTokenizeFromMessages returns the number of tokens in a list of messages.
 func (t *tiktokenTokenizer) NumTokenizeFromMessages(messages []types.Message) (int, error) {
-	if t.modelFamily != GPT {
-		return 0, errors.Newf("tiktoken NumTokenizeFromMessages is only support for OpenAI GPT family of models")
-	}
 	numTokens := 0
-	tokensPerMessage := 3
 	for _, message := range messages {
-		numTokens += tokensPerMessage
-		numTokens += len(t.tk.Encode(message.Speaker, nil, nil))
```
chrsmith, line 60: ⚠️ What is the purpose of `types.Message::Speaker`. And are we sure we don't want to include it in our token counts now?

internal/completions/tokenizer/tokenizer.go
```diff
@@ -69,49 +60,34 @@ func modelFamilyFromString(model string) ModelFamily {
 	switch {
 	case strings.Contains(model, AnthropicModel):
 		switch {
-		// Selects for claude 2, claude 2.1 and claude instant models
```
chrsmith, line 72: Since this PR is removing the `claude.{go, json}` and only using tiktoken... could it be that this function is going to be problematic? e.g. callers of `tokenzier.modelFamilyFromString` will likely do the wrong thing, now that we have removed support for Claude tokenizers?

It seems like we should understand the callers of this function, and try to obviate/delete it if possible.

internal/completions/tokenizer/tokenizer.go
```diff
@@ -69,49 +60,34 @@ func modelFamilyFromString(model string) ModelFamily {
...
 		return UnknownModel
 	}
 }
 
-// NewTokenizer returns a Tokenizer instance based on the provided model.
+// NewTokenizer returns a cl100k_base Tokenizer instance for all models, use for abuse detection only.
+//
+// NOTE: The tokenizer must match the tokenizer used by the Cody clients to ensure consistency across clients,
+// and the Cody clients have standardized on using the cl100k_base tokenizer for token counting.
 func NewTokenizer(model string) (Tokenizer, error) {
```
chrsmith, line 80: For clarity, why don't we call this `NewCL100kBaseTokenizer()`?

Also, what purposes does the `model` parameter serve, if the specific tokenizer we return is fixed? Isn't that something that is no longer relevant to this more "streamlined" design? (Note that I'm not sure if this will break any of the Cody Enterprise stuff that @arafatkatze was working on recently.)

cmd/cody-gateway/internal/httpapi/completions/fireworks.go
```diff
@@ -125,9 +125,15 @@ func (f *FireworksHandlerMethods) validateRequest(_ context.Context, _ log.Logge
 	return nil
 }
 
-func (f *FireworksHandlerMethods) shouldFlagRequest(_ context.Context, _ log.Logger, _ fireworksRequest) (*flaggingResult, error) {
-	// TODO[#61278]: Add missing request validation for all LLM providers in Cody Gateway.
-	return nil, nil
+func (f *FireworksHandlerMethods) shouldFlagRequest(ctx context.Context, logger log.Logger, req fireworksRequest) (*flaggingResult, error) {
+	result, err := isFlaggedRequest(
```
chrsmith, line 129: Could you actually revert the changes here in `fireworks.go`? It's a good change to make, but I'd suggest we "update the tokenizer encoding" in this PR, and put "enable Flagging for Fireworks-routed requests" in a second PR. (So we can more easily rollback that change if something goes awry, instead of making bot changes in one PR.)

internal/completions/tokenizer/tokenizer.go
```diff
@@ -69,49 +60,34 @@ func modelFamilyFromString(model string) ModelFamily {
...
-}
-
-func newOpenAITokenizer(model string, modelFamily ModelFamily) (*tiktokenTokenizer, error) {
-	// Remove "azure" or "openai" prefix from the model string
+	// Remove "azure" or "openai" prefix from the model string if any
 	model = strings.NewReplacer(AzureModel+"/", "", OpenAIModel+"/", "").Replace(model)
+	modelFamily := modelFamilyFromString(model)
 
-	tkm, err := tiktoken.EncodingForModel(model)
+	// Use the offline loader to avoid downloading the encoding at runtime
```
chrsmith, line 85: For readability, comments in code should be complete sentences. Meaning starting with a capital letter and ending with a period. (If the comment is only 2-3 words, don't worry about it.)

But the comment on line 81 needs a period, too.

```suggestion
	// Use the offline loader to avoid downloading the encoding at runtime.
```

chrsmith (approved):
Overall this PR looks good, but there are a few important things to track down.

> Cody's client has recently standardized the usage of the cl100k_base tokenizer for their token counting process, while Cody Gateway is using different tokenizers for different models...

We've talked about this before, and this is a good change to make. Although it's worth calling out explicitly that while this means that the client/server will "agree" on token counts, it is not "accurate" because different LLMs and models may use different tokenizers.

So the token count we compute during the flagging process is NOT expected to be authoritative. And if we want to surface token counts for, say, LLM costs analysis, we will need to use the actual token count returned from the LLM response. Or do the proper LLM x model x tokenizer mapping. (And not just use `cl100k_base` for everything.

Anyways, while the change conceptually looks good, I'm worried that this PR isn't covering all the bases. e.g. we still have the `modelFamily` field on `tiktokenTokenizer`, despite it likely no longer being correct or relevant?

Or if there are any invariants that are required by the code outside of Cody Gateway that uses the `tokenizer` package.

internal/completions/tokenizer/tokenizer.go
arafatkatze: The model parameter is not needed in this PR and I agree with that. 

AS for the PR working I tested it with my changes and my case of using my own keys and it works fine.

internal/completions/tokenizer/tokenizer.go
arafatkatze: Absolutely the callers of these things don't really need to do the model differentiation now. At the time of writing this the idea was that I will be implementing a separate tokenizer for claude 2 and claude 3 family of models which is not needed now.

internal/completions/tokenizer/tokenizer.go
abeatrix: On yjr client side we are not counting token used by the `role` / `speaker` ([link to code](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/lib/shared/src/token/counter.ts?L161-163)), that's why I'm removing it from the tokenizer on the Cody Gateway side as well. The token used by each "speaker" (user, assistant, system) would be 1 anyway, so if we want to add it back, we won't need to encode the speaker string to get the token count if I understand correctly?

internal/completions/tokenizer/tokenizer.go
abeatrix: I didn't remove any of the helper functions because I wasn't sure how much of these @arafatkatze will reuse, but now that I got confirmation, I will delete them from this PR. Thanks for the reviews, you all!

internal/completions/tokenizer/tokenizer.go
abeatrix: removed!

abeatrix:
This PR will be followed up by @arafatkatze to implement getting actual token count returned from the LLM response for different models

PR #62090, opened 2024-04-22
Fix icons color

vovakulikov:
Regression after https://github.com/sourcegraph/sourcegraph/pull/62012

In https://github.com/sourcegraph/sourcegraph/pull/62012 I changed Icon.svelte CSS variable API but didn't make a proper migration, so in this fix, I just bring back old variable support. 

## A few words about naming

So originally we have `--color` variable to set the color for `Icon.svelte` but this seemed a bit too generic so I wanted to rename it to `--icon-color` but we already have this variable as a design token (too bad, we should revisit our CSS color variables it looks a bit chaotic at the moment), eventually I ended up with `--icon-fill-color`, not ideal but doesn't overlap with other css variables.

## Test plan
- Check that icons have proper colors in different context (file tree extension icons, global nav icons color, etc.)

### DIFF
--- a/client/web-sveltekit/src/lib/Icon.svelte
+++ b/client/web-sveltekit/src/lib/Icon.svelte
@@ -27,14 +27,13 @@
 
 <style lang="scss">
     :root {
-        --icon-size: 1rem;
-        --icon-fill-color: inherit;
+        --icon-size: 1.5rem;
     }
 
     svg {
         width: var(--icon-size);
         height: var(--icon-size);
-        color: var(--icon-fill-color, inherit);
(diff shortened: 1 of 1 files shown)


client/web-sveltekit/src/lib/Icon.svelte
```diff
@@ -27,14 +27,13 @@
 
 <style lang="scss">
     :root {
-        --icon-size: 1rem;
-        --icon-fill-color: inherit;
+        --icon-size: 1.5rem;
```
camdencheek, line 30: Why did you change the default icon size here?

camdencheek (approved):

PR #62095, opened 2024-04-22
Simplify path mapping of p4-fusion package

eseliger:
This PR is a precursor to a second PR that will drop the shell layer in between gitserver and p4-fusion. We will be able to track command outputs and resource requirements separately, so there should be no need for this additional layer of complexity that could potentially mess with process groups and exit code forwarding etc.

Test plan:

This just bumps the package and simplifies the path, a follow-up PR will make use of this new epoch.


willdollman (approved):
👍 
You could also update `p4-fusion.yaml` for consistency, but it's unused so up to you.

PR #62106, opened 2024-04-23
feat(sg/release): allow auto versioning on release

Chickensoupwithrice:
This adds `--version auto` functionality to `sg` for performing releases. Resolves 

One concern I have about this is that it hard codes the sourcegraph endpoint specifically, but we use `sg release create` in various other repos. Currently we only use `sg release` to deploy sourcegraph (alongside deploy-* where we would instead want to query the latest public release and use that number, rather than generate a new one), but this sets the wrong precedent. 

I don't see a clear path towards calling this in a `release.yaml`, and hard coding specific git repo seems like a bad idea too. Welcome to suggestions. 

## Test plan

Manually tested locally:

image


dev/sg/internal/release/release.go
```diff
@@ -199,9 +201,20 @@ func newReleaseRunnerFromCliContext(cctx *cli.Context) (*releaseRunner, error) {
 
 	workdir := cctx.String("workdir")
 	pretend := cctx.Bool("pretend")
-	// Normalize the version string, to prevent issues where this was given with the wrong convention
-	// which requires a full rebuild.
-	version := fmt.Sprintf("v%s", strings.TrimPrefix(cctx.String("version"), "v"))
+	var version string
+	if cctx.String("version") == "auto" {
+		resp, err := http.Post("https://releaseregistry.sourcegraph.com/v1/releases/sourcegraph/next/5.3", "", nil)
```
craigfurman, line 206: I'm new to release registry / sg-release, so I might be missing context, but: why do we hardcode the minor version here? If the patch version is the awkward thing to determine, perhaps we ask the user to specify the major.minor, e.g. instead of `sg release create --version auto`, something like: `sg release create --next 5.3`?

More of a question than a blocker!

BolajiOlajide: Yeah, we can't hardcode the `<major>.<minor>` version for the release registry endpoint.

CleanShot 2024-04-23 at 12 04 13@2x

Could we use the Github endpoint to fetch the latest release and parse the `tag_name` to get the `<major>.<minor>` version to use.

```
curl https://api.github.com/repos/sourcegraph/sourcegraph/releases/latest | jq '.tag_name'
```

BolajiOlajide (changes requested):

dev/sg/internal/release/release.go
craigfurman: @BolajiOlajide is it a safe assumption that we are always building a release on the latest minor series? Do we always roll forward, or do we maintain multiple patch series? e.g. backport a patch to older, but still maintained, minor series

BolajiOlajide: This is true for patch release bumps (monthly / patch releases).

craigfurman: 💡 right, so when (e.g.) 5.4.PATCH comes out, there will be no more 5.3 releases?

BolajiOlajide: Correct. And for that, we'd have to specify the version explicitly and not use the `auto` flag.

Chickensoupwithrice: Good catch! I re-worked this such that we don't hardcode the `<major>.<minor>` in the URL, but still call out to `releaseregistry` rather than Github Releases. However, it's using latest id instead of latest release time, which may be a problem still? Currently the way the `releaseregistry` is seeded, newer IDs correspond to newer releases, but this assumption is not necessarily true?

Chickensoupwithrice:
Updated to no longer hardcode `<major>.<minor>` in the URL requesting the next version. 

Tested locally again:

image

dev/sg/internal/release/release.go
```diff
@@ -190,6 +191,45 @@ var Command = &cli.Command{
...
+	}
+	latestVersionBody, _ := io.ReadAll(latestVersion.Body)
+	latestVersion.Body.Close()
+	var versions []ReleaseInfo
+	jsonErr := json.Unmarshal(latestVersionBody, &versions)
+	if jsonErr != nil {
+		return "", errors.New("Could not parse ReleaseInfo json")
+	}
+
+	newestVersion := versions[0].Version
```
craigfurman, line 218: non-blocking note: we should probably check the length of versions first, but tbh the resultant panic if it is zero-length will be pretty obvious too, so no big deal!

craigfurman (approved):
🚀

dev/sg/internal/release/release.go
```diff
@@ -188,6 +191,45 @@ var Command = &cli.Command{
...
+	ID         int    `json:"id"`
+	Name       string `json:"name"`
+	Public     bool   `json:"public"`
+	CreatedAt  string `json:"created_at"`
+	PromotedAt string `json:"promoted_at"`
+	Version    string `json:"version"`
+	GitSha     string `json:"git_sha"`
+}
+
+func DetermineMinorVersion() (string, error) {
```
BolajiOlajide, line 204: Tiny nit: This function isn't used anywhere else so no need to export it, same with the `ReleaseInfo` struct.

dev/sg/internal/release/release.go
```diff
@@ -188,6 +191,45 @@ var Command = &cli.Command{
...
+	latestVersion, err := http.Get("https://releaseregistry.sourcegraph.com/v1/releases/sourcegraph")
+	if err != nil {
+		fmt.Print(err)
```
BolajiOlajide, line 207: We don't need this `fmt.Print` since we're returning the error already.

I'm guessing this was left there during debugging.

dev/sg/internal/release/release.go
```diff
@@ -188,6 +191,45 @@ var Command = &cli.Command{
...
+		return "", errors.New("Could not get latest release version")
+	}
+	latestVersionBody, _ := io.ReadAll(latestVersion.Body)
```
BolajiOlajide, line 210: How about we check the status code of the response before reading the body, that way if the status code isn't 200, we can return err.

Lastly, it might not be a bad idea to `defer latestVersion.Body.Close()` so it runs as a cleanup.

dev/sg/internal/release/release.go
BolajiOlajide: Yeah, +1 to what craig said.

If the version specified doesn't exist, the API returns `nil` so it's better to have a `len(versions) == 0` check.

dev/sg/internal/release/release.go
```diff
@@ -188,6 +191,45 @@ var Command = &cli.Command{
...
+	latestVersion.Body.Close()
+	var versions []ReleaseInfo
+	jsonErr := json.Unmarshal(latestVersionBody, &versions)
+	if jsonErr != nil {
+		return "", errors.New("Could not parse ReleaseInfo json")
+	}
+
+	newestVersion := versions[0].Version
```
BolajiOlajide, line 218: Also, to be safe, I'd pass the `.Version` with `semver.New` or `semver.Parse` (Can't remember the API signature), then that way we can get the patch and minor version by calling

`.Patch`
or
`.Minor`

BolajiOlajide (changes requested):

dev/sg/internal/release/release.go
Chickensoupwithrice: Is status code errors not part of `err`?

BolajiOlajide: No. This will return an error if there was an error with making the request, but it won't necessarily throw an error if the request returns an error code . For example, it won't return an error because the request 404'd.

You'd explicitly need to check the .StatusCode field of the http response.

Sorry the formatting of my message is bad, i'm typing from my mobile.

craigfurman: What bolaji said is right: If you get a legal HTTP response back, err will be nil. That err is basically for transport errors only

dev/sg/internal/release/release.go
```diff
@@ -201,24 +203,27 @@ type ReleaseInfo struct {
...
 
-	newestVersion := versions[0].Version
-	patchIndex := strings.LastIndex(newestVersion, ".")
-	minorVersion := strings.TrimPrefix(newestVersion[:patchIndex], "v")
-	url := fmt.Sprintf("https://releaseregistry.sourcegraph.com/v1/releases/sourcegraph/next/%s", minorVersion)
+	if len(versions[0].Version) == 0 {
+		return "", errors.New("Empty version number detected")
+	}
+	newestVersion := semver.MustParse(strings.TrimPrefix(versions[0].Version, "v"))
+	url := fmt.Sprintf("https://releaseregistry.sourcegraph.com/v1/releases/sourcegraph/next/%s", strconv.FormatInt(newestVersion.Major(), 10) + "." + strconv.FormatInt(newestVersion.Minor(), 10))
```
Chickensoupwithrice, line 226: I'm not sure this is much better than what I was doing previously, but at least it's type checked?

dev/sg/internal/release/release.go
```diff
@@ -188,6 +193,66 @@ var Command = &cli.Command{
...
+	PromotedAt string `json:"promoted_at"`
+	Version    string `json:"version"`
+	GitSha     string `json:"git_sha"`
+}
+
+// Determines latest major.minor.patch version number by hitting the releaseregistry
+// Is only called when --version auto is passed to the sg release command
+// Should *only* be called for patch releases for the monorepo!
+// returns the new patch number for the latest minor version, in the form of "major.minor.patch"
+func determineMinorVersion(ctx context.Context) (string, error) {
```
BolajiOlajide, line 212:
```suggestion
func determineNextReleaseVersion(ctx context.Context) (string, error) {
```

BolajiOlajide (approved):

dev/sg/internal/release/release.go
jhchabran: > 5.4.PATCH comes out, there will be no more 5.3 releases 

We might have to release an emergency patch though, so we don't want to cut one nor we plan to, but we may have to.

dev/sg/internal/release/release.go
```diff
@@ -188,6 +193,66 @@ var Command = &cli.Command{
...
+	ID         int    `json:"id"`
+	Name       string `json:"name"`
+	Public     bool   `json:"public"`
+	CreatedAt  string `json:"created_at"`
+	PromotedAt string `json:"promoted_at"`
+	Version    string `json:"version"`
+	GitSha     string `json:"git_sha"`
+}
+
+// Determines latest major.minor.patch version number by hitting the releaseregistry
```
jhchabran, line 208: nit: godoc conventions are to start the comment with the function name. 

```suggestion
// determineNextReleaseVersion determines latest major.minor.patch version number by hitting the releaseregistry
```

dev/sg/internal/release/release.go
```diff
@@ -188,6 +193,66 @@ var Command = &cli.Command{
...
+// Is only called when --version auto is passed to the sg release command
+// Should *only* be called for patch releases for the monorepo!
+// returns the new patch number for the latest minor version, in the form of "major.minor.patch"
+func determineNextReleaseVersion(ctx context.Context) (string, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, "https://releaseregistry.sourcegraph.com/v1/releases/sourcegraph", nil)
```
jhchabran, line 213: nit: extracting this to a constant will open up the path for a later PR to use a flag to swap the endpoint. Why wouldn't you want to use the `dev` variant of the deployment when iterating locally?

dev/sg/internal/release/release.go
```diff
@@ -188,6 +193,66 @@ var Command = &cli.Command{
...
+	}
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return "", errors.Wrap(err, "Could not get response from releaseregistry")
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 200 {
+		return "", errors.New("Releaseregistry did not return statuscode 200")
```
jhchabran, line 225: This error message won't help to understand what's wrong, because we're not sharing what was the status code. We should log the status code and the body if any.

dev/sg/internal/release/release.go
```diff
@@ -188,6 +193,66 @@ var Command = &cli.Command{
...
+		return "", errors.Wrap(err, "Could not create POST request")
+	}
+	resp, err = http.DefaultClient.Do(req)
+	if err != nil {
+		return "", errors.New("Could not automatically determine new version number")
+	}
+	if resp.StatusCode != 200 {
+		return "", errors.New("Releaseregistry did not return statuscode 200")
+	}
+	bodyBytes, _ := io.ReadAll(resp.Body)
```
jhchabran, line 250: Why skip handling the error here?

dev/sg/internal/release/release.go
```diff
@@ -188,6 +193,66 @@ var Command = &cli.Command{
...
+	req, err = http.NewRequestWithContext(ctx, http.MethodPost, url, nil)
+	if err != nil {
+		return "", errors.Wrap(err, "Could not create POST request")
+	}
+	resp, err = http.DefaultClient.Do(req)
+	if err != nil {
+		return "", errors.New("Could not automatically determine new version number")
+	}
+	if resp.StatusCode != 200 {
+		return "", errors.New("Releaseregistry did not return statuscode 200")
```
jhchabran, line 248: Same

dev/sg/internal/release/release.go
jhchabran, line 225:
```suggestion
		return "", errors.Newf("API error, got status %d", resp.StatusCode)
```

jhchabran (changes requested):
Thanks! Comments are minor, and easy to fix but I believe that the status code error should at least be fixed before merging this.

jhchabran (approved):

PR #62107, opened 2024-04-23
gitserver: Implement Stat and ReadDir in gRPC API

eseliger:
This PR migrates the calls to git ls-tree to gRPC.

Since ReadDir and Stat used the same git command under the hood, I migrated those two together.

Stat and ReadDir are now simply moved to the backend with the usual layering we did so far.

`ListDirectoryChildren` effectively was a `ReadDir` with some post-request processing into a map, so I moved that logic to the only caller instead and made it just use `ReadDir`.

`LsFiles` uses the `ls-files` command which is meant for worktrees and is not a plumbing command so I tried to best replicate its use-cases using `ReadDir` instead.
There were three use-cases for `LsFiles` so far:
1. Exposing all files in the repo from GQL for the fuzzy finder (this is a terrible API as it'll create potentially humongous responses but that's not a fault of our API), and I made it use ReadDir with recurse: true instead and I filter out dir entries.
2. The own indexer used it to go over all the files, I made that use `ReadDir` instead as well.
3. In inference, it was used to get files matching certain patterns. Git `ls-tree` does not support globbing while `ls-files` does, so we can not do that anymore. BUT: the logic already did filtering in-app, it just used that to reduce the result set to iterate over. So we decided that for now we will go with a ReadDir and then later we can try to find a better replacement here.

Closes https://github.com/sourcegraph/sourcegraph/issues/47420

Closes https://github.com/sourcegraph/sourcegraph/issues/60413

Closes https://github.com/sourcegraph/sourcegraph/issues/60414

## Test plan

Verified that inference still returns the same results. Also browsed files, opened file trees etc from the web app. Existing E2E test suites also still pass.


eseliger:
* **#62107** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62107?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

cmd/gitserver/internal/git/gitcli/odb.go
```diff
@@ -276,3 +284,325 @@ func (g *gitCLIBackend) FirstEverCommit(ctx context.Context) (api.CommitID, erro
...
+	if path != "" {
+		// Trailing slash is necessary to ls-tree under the dir (not just
+		// to list the dir's tree entry in its parent dir).
+		path = filepath.Clean(rel(path)) + "/"
+	}
+
+	return g.lsTree(ctx, commit, path, recursive)
+}
+
+func (g *gitCLIBackend) ReadDirPatterns(ctx context.Context, commit api.CommitID, pathspecs ...gitdomain.Pathspec) (git.ReadDirIterator, error) {
```
eseliger, line 343: Turns out ls-tree doesn't expand patterns and ls-files does, so this is NOT a viable replacement rn.

eseliger: This _does_ pass all tests though.. scary.

internal/codeintel/autoindexing/internal/inference/service.go
```diff
@@ -274,17 +274,29 @@ func (s *Service) resolvePaths(
...
 	}
 
-	// Ideally we can pass the globs we explicitly filter by below
-	paths, err := invocationContext.gitService.LsFiles(ctx, invocationContext.repo, invocationContext.commit, pathspecs...)
+	// Ideally we can pass the pathspecs from flattenPatterns here and avoid returning
+	// all files in the tree, so we don't need to filter as much further down.
+	// This requires implementing either globbing support in the ReadDir call,
+	// or that the pathpatterns are supported by ReadDir (and thus, in git ls-tree)
+	// which is not the case today.
+	fds, err := invocationContext.gitService.ReadDir(ctx, invocationContext.repo, api.CommitID(invocationContext.commit), "", true)
```
eseliger, line 287: this is the tradeoff we talked about, we send more paths here again and filter on more of them in-app instead of in git.

cmd/gitserver/internal/git/gitcli/odb.go
```diff
@@ -276,3 +284,285 @@ func (g *gitCLIBackend) FirstEverCommit(ctx context.Context) (api.CommitID, erro
...
+
+func (g *gitCLIBackend) ReadDir(ctx context.Context, commit api.CommitID, path string, recursive bool) (git.ReadDirIterator, error) {
+	if err := checkSpecArgSafety(string(commit)); err != nil {
+		return nil, err
+	}
+
+	if path != "" {
+		// Trailing slash is necessary to ls-tree under the dir (not just
+		// to list the dir's tree entry in its parent dir).
+		path = filepath.Clean(rel(path)) + "/"
```
ggilmore, line 341: Nit, why not use the os path separator?

```suggestion
		path = filepath.Clean(rel(path)) + string(os.PathSeparator)
```

eseliger: On windows, this will be `\` which git might not interpret correctly I believe 👀 (although, I straight up copied that from the existing implementation in the client)

cmd/gitserver/internal/git/gitcli/odb_test.go
```diff
@@ -421,3 +425,397 @@ func TestGitCLIBackend_GetBehindAhead(t *testing.T) {
...
+		fis := make([]fs.FileInfo, 0)
+		for {
+			fi, err := it.Next()
+			if err == io.EOF {
+				break
+			}
+			require.NoError(t, err)
+			fis = append(fis, fi)
+		}
+		require.Empty(t, cmp.Diff([]fs.FileInfo{
```
ggilmore, line 787: nit: I think it's a bit odd to use both `require` and `cmp` on the same assertion TBH

cmd/gitserver/internal/git/gitcli/odb_test.go
```diff
@@ -421,3 +425,397 @@ func TestGitCLIBackend_GetBehindAhead(t *testing.T) {
...
+			&fileutil.FileInfo{
```
ggilmore, line 788: I see that we're not doing any sort of sorting. In the interface definition for the function, I don't think we specify that we emit paths an any particular order - is that correct? (if we don't care about the order, we can make that more obvious in the test by adding a Sort() call)

cmd/gitserver/internal/git/gitcli/odb_test.go
eseliger: require can unfortunately not be modified in it's behavior and some fields of the fileutil.FileInfo are quite volatile so I excluded them in the comparison using cmp to make it easier to adjust tests in the future if we change anything here.
Does that make sense?

cmd/gitserver/internal/git/gitcli/odb_test.go
eseliger: Hmm yea good question, it is ordered by Git, so it's not entirely random, but I am not sure if git reorders the items before emitting, or if it orders at insertion time into a tree object and when you hand-craft a tree entry in the ODB it wouldn't be ordered the same way.

cmd/gitserver/internal/server_grpc.go
```diff
@@ -1520,6 +1520,171 @@ func (gs *grpcServer) ChangedFiles(req *proto.ChangedFilesRequest, ss proto.Gits
...
+	}
+
+	sendFunc := func(fis []*proto.FileInfo) error {
+		return ss.Send(&proto.ReadDirResponse{FileInfo: fis})
+	}
+
+	// We use a chunker here to make sure we don't send too large gRPC messages.
+	// For repos with thousands or even millions of files, sending them all in one
+	// message would be very slow, but sending them all in individual messages
+	// would be slow either, so we chunk them instead.
```
ggilmore, line 1650: Grammar error 

```suggestion
// would also be slow, so we chunk them instead.
```

cmd/gitserver/internal/server_grpc.go
```diff
@@ -1520,6 +1520,171 @@ func (gs *grpcServer) ChangedFiles(req *proto.ChangedFilesRequest, ss proto.Gits
...
+			})
+			if err != nil {
+				return err
+			}
+			return s.Err()
+		}
+		gs.svc.LogIfCorrupt(ctx, repoName, err)
+		return err
+	}
+
```
ggilmore, line 1642: Can you add a defer for closing the iterator here? I feel like it would clean up a ton of the error handling logic below:

```go
defer func() {
   closeErr := it.Close()
   if closeErr == nil {
      return
   }


   if err != nil {
      err = errors.Append(err, closeErr)
      return 
   }

   err = closeErr
   return
}

```

go.mod
```diff
@@ -111,7 +111,7 @@ require (
 	github.com/gogo/protobuf v1.3.2
 	github.com/golang-jwt/jwt/v4 v4.5.0
 	github.com/golang/gddo v0.0.0-20210115222349-20d68f94ee1f
-	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da
+	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
```
ggilmore, line 114: Intentional?

eseliger: Side-effect because we no longer cache ls-tree stuff in the client package

cmd/gitserver/internal/git/iface.go
```diff
@@ -243,3 +257,12 @@ type ContributorCountsOpts struct {
 	// (e.g., "foo/bar/").
 	Path string
 }
+
+// ReadDirIterator is an iterator for the contents of a directory.
+type ReadDirIterator interface {
```
ggilmore, line 262: You should document that the caller is expected to close the iterator.

ggilmore (approved):

cmd/gitserver/internal/server_grpc.go
eseliger: ah yes that seems nicer thanks!

eseliger:
### Merge activity

* **May 10, 6:58 PM EDT**: @eseliger merged this pull request with [Graphite](https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62107).

PR #62108, opened 2024-04-23
msp: add support for "private networking"

bobheadxi:
Implements the enablement of the [MSP "private networking"]() mechanism via MSP service spec.

Closes https://github.com/sourcegraph/managed-services/issues/1224


bobheadxi:
Leaving in draft state for now until we get a chance to dig into https://linear.app/sourcegraph/issue/CORE-110 and decide on a final mechanism. The existing setup is fairly trivial Terraform so should be fine to configure by hand for now

PR #62110, opened 2024-04-23
cmd/msp-example: remove gorm

bobheadxi:
gorm is a heavy-ish dependency and not _actually_ used anywhere in the monorepo - I mostly added this to ensure we had comprehensive test environment coverage for SAMS and SSC use cases, which do use gorm in separate repositories.

Both of the above services now have their own long-lived dev environments - we no longer need such a comprehensive example, and a simple ping and query will do for msp-example.

## Test plan

Add `sg.config.overwrite.yaml` to override the database connection with local dev:

```yaml
  msp-example:
    env:
      STATELESS_MODE: false
      PGDSN: 'user=sourcegraph password=sourcegraph database=sourcegraph'
```

Start msp-example:

```sh
sg run msp-example
```

Observe logs reach the point where the Postgres is connected to successfully (then fails due to other things, but that is okay):

```
[    msp-example] INFO service example/example.go:47 starting service
[    msp-example] INFO service example/example.go:53 postgresql connection success
```


eseliger (approved):
<3

unknwon (approved):
Good move 😁

PR #62114, opened 2024-04-23
symbols: use concurrency safe map for repoToSize in rockskip

keegancsmith:
We are getting "fatal error: concurrent map writes" in production when enabling rockskip at large customers. The root cause is the repoToSize map which is read for every search request (and written every "cache miss").

This was the easiest change to make. We should probably also be single flighting this request, but that can be changed at another time.

Additionally fixed another bug where in the cache miss case we would do `size :=` and therefore not update size from 0, which then does a sqliteSearch. This means on the very first request to a rockskip repo we would always do sqliteSearch which is also a waste of resources.

Test Plan: Stefan manually testing :)

### DIFF
--- a/cmd/symbols/shared/setup.go
+++ b/cmd/symbols/shared/setup.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"net/http"
 	"strings"
+	"sync"
 
 	"github.com/sourcegraph/go-ctags"
 	"github.com/sourcegraph/log"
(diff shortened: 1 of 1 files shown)


stefanhengl:
Manual testing:
- I ran a local instance and triggered Rockskip indexing for 2 repos. I checked the db and confirmed that both repos were indexed as expected
- The logs showed no panic or error (other than the expected error that indexing is still in progress).
- I browsed at an unindexed commit and confirmed that the symbol sidebar is working and that the sqlite cache `/tmp/symbols-cache` stayed empty.

stefanhengl (approved):
LGTM!

I left a comment about how I tested.

keegancsmith:
Thanks!

Issue #61942, opened 2024-04-16
Svelte: mouse passing over hoverable token at top of page opens popover


camdencheek:
Moving the mouse over a symbol at the top of the page opens the popover, blocking the "Blame" button. This happens even if I don't stop on the symbol. It does not happen elsewhere on the page, only seemingly when moving the mouse outside the editor div.

https://github.com/sourcegraph/sourcegraph/assets/12631702/4bbef064-8583-4ef0-93af-871a5623e031

PR #62124, opened 2024-04-23
blob: Properly hide popovers when the mouse leaves the file area

fkling:
Fixes https://github.com/sourcegraph/sourcegraph/issues/61942

There are currently two situations where the popover isn't properly closed:

1. The popover extends outside of the file view boundaries and the mouse leaves the file view via the popover.
2. The hovered symbol is at the edge of the file view and the mouse leaves the file view moving over the symbol.

The first case is annoying becaues the popover might cover important information and it's not possible to close the popover by moving outside of it.

The second case is annoying because it can cause the popover to appear when moving outside of the file view (towards the top or bottom), which is just surprising at the best case, but can hide important information in the worst case.

This affects both the React and SvelteKit versions.

The solution implemented in this commit is to explicitly listen to `mouseleave` events and hide the popover accordingly.

Before and after video: 

https://github.com/sourcegraph/sourcegraph/assets/179026/a71982b7-6a32-4b62-be17-0863ce76d48b

Fixes https://github.com/sourcegraph/sourcegraph/issues/61942

## Test plan

Manual testing.

### DIFF
--- a/client/web/src/repo/blob/codemirror/codeintel/hover.ts
+++ b/client/web/src/repo/blob/codemirror/codeintel/hover.ts
@@ -1,7 +1,7 @@
 import type { Extension } from '@codemirror/state'
 import { type EditorView, type PluginValue, ViewPlugin, getTooltip, type Tooltip } from '@codemirror/view'
-import { from, fromEvent, Subscription } from 'rxjs'
-import { debounceTime, filter, map, scan, tap } from 'rxjs/operators'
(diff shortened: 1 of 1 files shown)


camdencheek:
Fixes https://github.com/sourcegraph/sourcegraph/issues/61942

camdencheek (approved):
LGTM! Tried it out and it feels much better. Thanks for the fix!

Issue #61636, opened 2024-04-05
Resubmitting the same query is broken


fkling:
When submitting the same query again we are not rerunning the query (likely due to caching) but we are adding a new history entry.

Instead we should rerun the query and we should not add a new history entry.

fkling:
On a high level the problem is that SvelteKit is not rerunning the loader because the URL doesn't change.

Originally this was solved by *invalidating* the data when the query was submitted. This was done by registering a custom dependency in the loader that contains the query from the URL and call `invalidate` with the same value.

This is currently broken because we use different values for generating the dependency "key": In the loader the query contains the `context:` filter, and in the page/input it doesn't. 

Maybe it's enough to just fix that. At the same time the fact that we call `invalidate` as well as `goto` seems weird. Maybe we can do that more selectively (only call invalidate when the query hasn't changed).

PR #62132, opened 2024-04-23
Svelte: invalidate all loaders when a query is submitted

camdencheek:
Fixes https://github.com/sourcegraph/sourcegraph/issues/61636

### DIFF
--- a/client/web-sveltekit/src/lib/search/input/SearchInput.svelte
+++ b/client/web-sveltekit/src/lib/search/input/SearchInput.svelte
@@ -8,7 +8,7 @@
     import { EditorView } from '@codemirror/view'
     import { mdiCodeBrackets, mdiFormatLetterCase, mdiRegex } from '@mdi/js'
 
-    import { goto, invalidate } from '$app/navigation'
+    import { goto } from '$app/navigation'
     import {
         type Option,
         type Action,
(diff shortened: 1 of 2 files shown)


client/web-sveltekit/src/lib/search/input/SearchInput.svelte
```diff
@@ -180,10 +180,7 @@
     }
 
     async function submitQuery(state: QueryState): Promise<void> {
-        // This ensures that the same query can be resubmitted from the search input. Without
-        // this, SvelteKit will not re-run the loader because the URL hasn't changed.
-        await invalidate(`query:${state.query}--${state.caseSensitive}`)
-        void goto(getQueryURL(state))
+        void goto(getQueryURL(state), { invalidateAll: true })
```
camdencheek, line 183: @fkling Am I thinking about this too simply? When we manually submit a query, can we just invalidate the page data?

fkling: I think that will also invalidate the root layout loader, i.e. will refetch user authentication information, settings, etc. That's probably not what we want.

camdencheek:
Closing in favor of https://github.com/sourcegraph/sourcegraph/pull/62184

PR #62134, opened 2024-04-23
sg/msp: warn user if sg-msp lockfile differs from current version

bobheadxi:
In https://github.com/sourcegraph/managed-services/pull/1288 I'm introducing granular, per-category version locks for https://github.com/sourcegraph/managed-services/issues/599. This change adds warnings that are shown to the user on version mismatches.

This may cause some friction with users, but for now, let's just deal with this on a case-by-case basis - our primary goal right now is to build a process for https://github.com/sourcegraph/managed-services/issues/599 by introducing more granular version locks.

Requires https://github.com/sourcegraph/sourcegraph/pull/62176 so that we can access the same version format as the one used in our lockfiles.

Details on our locking strategy is here: https://sourcegraph.notion.site/Deploying-new-versions-of-MSP-1808e7e45bd54f419dd93af542d99238#58dabe4992754ca18ed39bc212ccbbba

## Test plan

```
sg msp generate -all
```

image


unknwon (approved):

dev/sg/msp/helpers.go
```diff
@@ -152,7 +153,44 @@ func syncEnvironmentWorkspaces(c *cli.Context, tfc *terraformcloud.Client, servi
...
+			out.WriteWarningf("Unable to determine locked 'sg' version for category %q: %s",
+				category, err.Error())
+		} else if lockedSgVersion != c.version {
+			out.WriteWarningf(`Lockfile for category %q declares 'sg' version %q, you are using %q - generated outputs may differ from what is expected.
+If there is a diff in the generated output, try running the following:`,
+				category, lockedSgVersion, c.version)
+			_ = out.WriteCode("bash", fmt.Sprintf(
+				"sg update -release %q &&\n  SG_SKIP_AUTO_UPDATE=true sg msp generate -all -category %q",
+				lockedSgVersion, string(category)))
+		}
```
bobheadxi, line 186: This depends on https://github.com/sourcegraph/sourcegraph/pull/62189

PR #62136, opened 2024-04-23
lib/background: upgrade `Routine` interface with context and errors

unknwon:
This PR is a result/followup of the improvements we've made in the [SAMS repo](https://github.com/sourcegraph/sourcegraph-accounts/pull/199) that allows call sites to pass down a context (primarily to indicate deadline, and of course, cancellation if desired) and collects the error returned from `background.Routine`s `Stop` method.

Note that I did not adopt returning error from `Stop` method because I realize in monorepo, the more common (and arguably the desired) pattern is to hang on the call of `Start` method until `Stop` is called, so it is meaningless to collect errors from `Start` methods as return values anyway, and doing that would also complicate the design and semantics more than necessary.

All usages of the the `background.Routine` and `background.CombinedRoutines` are updated, I DID NOT try to interpret the code logic and make anything better other than fixing compile and test errors.

The only file that contains the core change is the [`lib/background/background.go`](https://github.com/sourcegraph/sourcegraph/pull/62136/files#diff-65c3228388620e91f8c22d91c18faac3f985fc67d64b08612df18fa7c04fafcd).

## Test plan

CI


lib/background/background.go
```diff
@@ -89,19 +120,28 @@ func exitAfterSignals(signals <-chan os.Signal, numSignals int) {
...
 }
 
-// CombinedRoutine is a list of routines which are started and stopped in unison.
+// CombinedRoutine is a list of routines which are started and stopped in
+// unison.
 type CombinedRoutine []Routine
 
-func (r CombinedRoutine) Start() {
+func (rs CombinedRoutine) Name() string {
+	return "combined"
```
eseliger, line 128: this feels a bit lazy, shouldn't it return the names of all inner routines? otherwise you can't really track this down to anything, every combined routine will simply be called "combined"

lib/background/background.go
```diff
@@ -112,24 +152,41 @@ func (r CombinedRoutine) Stop() {
...
+		err := r[i].Stop(ctx)
+		if err != nil {
+			stopErr = errors.Append(stopErr, errors.Wrapf(err, "stop routine %q", r[i].Name()))
+		}
 	}
+	return stopErr
 }
 
 // NoopRoutine does nothing for start or stop.
 func NoopRoutine() Routine {
```
eseliger, line 171: should we let the caller pass a name here?

lib/background/background.go
```diff
@@ -6,43 +6,47 @@ import (
...
 }
 
-// Monitor will start the given background routines in their own
-// goroutine. If the given context is canceled or a signal is received, the Stop
-// method of each routine will be called. This method blocks until the Stop methods
-// of each routine have returned. Two signals will cause the app to shutdown
+// Monitor will start the given background routines in their own goroutine. If
+// the given context is canceled or a signal is received, the Stop method of
+// each routine will be called. This method blocks until the Stop methods of
+// each routine have returned. Two signals will cause the app to shut down
```
eseliger, line 31: Not sure if this is from a formatted but meta comment: The change of line breaks makes this diff kinda hard to read 🙈

lib/background/background.go
```diff
@@ -6,43 +6,47 @@ import (
...
 // immediately.
-func Monitor(ctx context.Context, routines ...Routine) {
+func Monitor(ctx context.Context, routines ...Routine) error {
```
eseliger, line 33: Monitor is meant to be called as the last thing in an application as it calls os.Exit, I'm not sure if the `error` here wouldn't be misleading 🙈

internal/httpserver/server.go
```diff
@@ -64,17 +68,18 @@ func (s *server) Start() {
...
 
-func (s *server) Stop() {
+func (s *server) Stop(ctx context.Context) error {
 	s.once.Do(func() {
 		if s.preShutdownPause > 0 {
 			time.Sleep(s.preShutdownPause)
 		}
 
-		ctx, cancel := context.WithTimeout(context.Background(), goroutine.GracefulShutdownTimeout)
+		ctx, cancel := context.WithTimeout(ctx, goroutine.GracefulShutdownTimeout)
```
eseliger, line 77: This is now potentially overwriting the timeout that is passed, that feels a bit counter intuitive 🤔

cmd/cody-gateway/internal/actor/source.go
```diff
@@ -247,6 +251,7 @@ func (s *redisLockedBackgroundRoutine) Stop() {
 
 	s.logger.Info("Background sync successfully stopped",
 		log.Duration("elapsed", time.Since(start)))
+	return stopErr
```
eseliger, line 254: feels a bit strange to say successful and then potentially return an error 😬

internal/repos/sync_worker_test.go
```diff
@@ -64,9 +65,18 @@ func TestSyncWorkerPlumbing(t *testing.T) {
...
+		require.NoError(t, err)
+	}()
+	defer func() {
+		err := resetter.Stop(ctx)
+		require.NoError(t, err)
+	}()
+	defer func() {
+		err := worker.Stop(ctx)
+		require.NoError(t, err)
+	}()
```
eseliger, line 79: I think this could leave behind running routines that could cross-contaminate test executions, the right thing to use when this can error is probably t.Cleanup instead of defer

lib/background/background.go
```diff
@@ -52,12 +56,39 @@ func startAll(wg *sync.WaitGroup, routines ...Routine) {
...
+// each running goroutine with the given waitgroup. It waits for all routines to
+// stop or the context to be canceled.
+func stopAll(ctx context.Context, wg *sync.WaitGroup, routines ...Routine) error {
+	var stopErrs error
+	var stopErrsLock sync.Mutex
 	for _, r := range routines {
-		t := r
 		wg.Add(1)
-		Go(func() { defer wg.Done(); t.Stop() })
+		go func(routine Routine) {
```
eseliger, line 66: Why the shift away from `Go` to `go` here? Mostly curious.

eseliger (commented):
A couple questions before I feel comfortable with this:

> (primarily to indicate deadline, and of course, cancellation if desired)

I am not too sure about the cancellation part here, what does cancelling a stop mean? I don't think most processes are fully reversible. You would probably need to continue a full stop and then call start again, but that seems overly complex.

For deadline, we have a hardcoded variable called `goroutine.GracefulShutdownTimeout` that implementors should ideally stick to. It feels a bit random to me that Start does not take a context (I think for good reasons, the factory function for the routine should handle that) but stop does, that could cause wildly different contexts to be passed with different dependencies stored on the context, like an internal actor or else.
Do you think documenting that you should stick to that graceful shutdown period where possible, vs documenting that you should respect the context deadline would be much different?
Right now, we have two ways here now, which will realistically never be consolidated if this PR doesn't do it so wondering if we could simplify here 😬 

---

The Name() method seems to often return not very useful information. How about instead we make a NamedRoutine interface that has Name() and embeds Routine and when you want to implement name, you can implement Name() but otherwise we return nothing? That seems cleaner than relying on the current hard-coded values that provide just about as much value as `""` or `<undefined>` 

---

A few more inline questions

cmd/cody-gateway/internal/actor/source.go
bobheadxi: I think this error message was always bad (oops), the only thing this shutdown should care about is the redis-locking part, and in this case releasing the lock was successful

lib/background/background.go
bobheadxi: I think the way it has always worked is that if a signal is caught, this function _can_ return, but on multiple signals or reaching a deadline, this function can choose to exit immediately. In that world and the new stop-error semantic, we can still have a case where `Monitor` never calls `os.Exit` directly, and instead returns an indicator of whether the shutdown was clean or not. I think that is the new desired behaviour, and the caller can then report an error indicating shutdown was unclean

lib/background/background.go
bobheadxi: +1 to reporting all named routines

internal/httpserver/server.go
bobheadxi: I actually couldn't find a lot online about this, isn't this the case with _every_ context, the child could always be overwriting a parent `WithTimeout` at any point it uses `WithTimeout`?

bobheadxi: By the new semantics though I think this is the correct behaviour, and the user of `httpserver` should stop with an appropriate timeout/deadline of their choice (which IMO makes more sense than the magic timeout that exists here right now)

lib/managedservicesplatform/runtime/service.go
```diff
@@ -137,6 +137,9 @@ func Start[
 		log.Int("port", ctr.Port),
 		log.Bool("msp", ctr.MSP),
 		log.Bool("sentry", sentryEnabled))
-	background.Monitor(ctx, routine)
+	err = background.Monitor(ctx, routine)
+	if err != nil {
+		startLogger.Error("error monitoring service routine", log.Error(err))
+	}
```
bobheadxi, line 143: Does this only error on shutdown? If so, should we indicate so in the docstring + error logs?

bobheadxi (approved):
Overall LGTM, minor questions + Erik's comments :) Thank you for doing this!

unknwon:
@eseliger:
> 
> > (primarily to indicate deadline, and of course, cancellation if desired)
> 
> I am not too sure about the cancellation part here, what does cancelling a stop mean? I don't think most processes are fully reversible. You would probably need to continue a full stop and then call start again, but that seems overly complex.

"Cancellation" is just happen to be the term that Go context's package uses, which IMO it just means caller wants to signal that "stop what you're doing and walk away". How exactly routines needs to be stopped, is not a concern of the caller, nor does caller imply things must be reversed. In other words, I think it is in routines' interests to do graceful shutdown whenever possible (either it's abort DB transaction, clean up temp files, or simply exit the loop, etc.).

Deadline is also just a form of cancellation, with a (generally longer) grace period.

> For deadline, we have a hardcoded variable called `goroutine.GracefulShutdownTimeout` that implementors should ideally stick to. It feels a bit random to me that Start does not take a context (I think for good reasons, the factory function for the routine should handle that) but stop does, that could cause wildly different contexts to be passed with different dependencies stored on the context, like an internal actor or else. Do you think documenting that you should stick to that graceful shutdown period where possible, vs documenting that you should respect the context deadline would be much different? Right now, we have two ways here now, which will realistically never be consolidated if this PR doesn't do it so wondering if we could simplify here 😬

The approach of sticking to `goroutine.GracefulShutdownTimeout` would only work consistently and reliably within the context of monorepo, where we know what we are doing. Now, the `lib/background` is clearly aiming to be more generically helpful package, there are two major cons of it:
1. It is not a safe assumption anymore, the needs and constraints of services can be vastly different, maybe longer, maybe shorter.
2. A hardcoded timeout more or less restricts that routines cannot/should not be nested, otherwise, it's subtle and not-obvious-illusion that nested routines are still having the same timeout, which in practice, let's say 50/50 true.

By nested routines, not the best illustration but here it is 😂 

```
routine 1/
├─ routine 1-1/
├─ routine 1-2/
│  ├─ routine 1-2-1
│  ├─ routine 1-2-2
│  ├─ routine 1-2-3
├─ routine 1-2
```

"routine 1" wants to stop within 30 seconds, "routine 1-2" needs to some work before its sub-routines can be shutdown, say, that takes 5 seconds, which realistically only left 25 seconds for "routine 1-2-{1,2,3}" to shutdown (not 30 seconds).

It is not a problem for routines can be stopped freely, but for routines that are absolutely critical to "save the work" before quitting, getting the wrong timeout is fatal.

> The Name() method seems to often return not very useful information. How about instead we make a NamedRoutine interface that has Name() and embeds Routine and when you want to implement name, you can implement Name() but otherwise we return nothing? That seems cleaner than relying on the current hard-coded values that provide just about as much value as `""` or `<undefined>`

@bobheadxi asked the same question and proposed this same idea in the original PR, please see [my response for why I did not adopt](https://github.com/sourcegraph/sourcegraph-accounts/pull/199#issuecomment-2073377317) and let me know WDYT (reply here)!

lib/background/background.go
unknwon: You caught me 😂 I was lazy, pushed changes in https://github.com/sourcegraph/sourcegraph/pull/62136/commits/9ec582833ad7b42ed17d79c8410e855730f63c68

lib/background/background.go
unknwon: Updated in https://github.com/sourcegraph/sourcegraph/pull/62136/commits/a10d07b2b32e3bb3796987a907aa2291d62aaade

lib/background/background.go
unknwon: 🥹 sorry yes it's basically just re-formatted.

lib/background/background.go
unknwon: To add on Robert's comment: While making this PR compile, I actually do notice that although it is often the last thing being called, many places are just returning `nil` right after calling it, but now it can actually return any real `error` now.

One step back, the reason the error is propagated to caller is because I do not think we should assume how caller what to handle the error, or where to print/send the error.

internal/httpserver/server.go
unknwon: The way I understand how `context.WithTimeout` (or in general, how `context.With`) works is that it always creates a new context (with the given timeout). In this particular case, it shouldn't change any semantic meaning, i.e. whoever gets the `ctx` on the left still have the same `goroutine.GracefulShutdownTimeout` as the timeout as before.

cmd/cody-gateway/internal/actor/source.go
unknwon: It feels strange for sure, but as stated in the PR description, I DID NOT try to interpret the code logic and make anything better other than fixing compile and test errors. 😁 If we want to improve the case here, I'd rather doing a follow-up, since that is technically changing some behavior who knows may or may not rely on it.

internal/repos/sync_worker_test.go
unknwon: Updated in https://github.com/sourcegraph/sourcegraph/pull/62136/commits/4b522ac37479b3644574bcccc64a06881eb52f26

lib/background/background.go
unknwon: Good eye, lost during refactoring 😀 Fixed in https://github.com/sourcegraph/sourcegraph/pull/62136/commits/750c011373c6256a7b70d05b4a682251da8293de

lib/managedservicesplatform/runtime/service.go
unknwon: Push a change in https://github.com/sourcegraph/sourcegraph/pull/62136/commits/fef863d7e0836037b713422c2605f2891713facd, PTAL!

unknwon:
@eseliger - thanks for the review! I should have addressed all the comments, PTAL!

internal/httpserver/server.go
bobheadxi: > whoever gets the ctx on the left still have the same goroutine.GracefulShutdownTimeout as the timeout as before.

I think that's where the confusion is:

1. Caller calls `foobar`, telling it the deadline is `goroutine.GracefulShutdownTimeout - 30s` for example
2. `foobar` uses `WithTimeout` and _actually_ has a deadline of `goroutine.GracefulShutdownTimeout`

internal/debugserver/debug.go
```diff
@@ -84,7 +84,7 @@ type Dumper interface {
 // Any extra endpoints supplied will be registered via their own declared path.
 func NewServerRoutine(ready <-chan struct{}, extra ...Endpoint) goroutine.BackgroundRoutine {
 	if addr == "" {
-		return goroutine.NoopRoutine()
+		return goroutine.NoopRoutine("noop server")
```
bobheadxi, line 87: Do we have a naming convention? Should this be `noopServer`?

internal/httpserver/server.go
```diff
@@ -51,6 +51,10 @@ func newServer(httpServer *http.Server, makeListener func() (net.Listener, error
 	return s
 }
 
+func (s *server) Name() string {
+	return "HTTP server"
```
bobheadxi, line 55: Hmm I guess the name is pretty free-form 😁

lib/background/background.go
```diff
@@ -2,47 +2,55 @@ package background
...
-// Routine represents a component of a binary that consists of a long
-// running process with a graceful shutdown mechanism.
+// Routine represents a background process that consists of a long-running
+// process with a graceful shutdown mechanism.
 type Routine interface {
-	// Start begins the long-running process. This routine may also implement
-	// a Stop method that should signal this process the application is going
-	// to shut down.
+	// Name returns the name of the routine.
+	Name() string
```
bobheadxi, line 18: From the current examples it looks like this is a free-form, human-readable name, let's document that here?

bobheadxi (approved):

internal/httpserver/server.go
unknwon: Got it, we could do something similar to SAMS pub/sub routine, where we can check the parent deadline, and apply whoever is shorter here?

CleanShot 2024-05-22 at 17 00 12@2x

(example here only checks whether parent imposes a deadline and leave a 5-second buffer)

lib/background/background.go
unknwon: Sounds good, updated in https://github.com/sourcegraph/sourcegraph/pull/62136/commits/1bd294db35c7c93bc80ab10677faba59a0437659

internal/httpserver/server.go
bobheadxi: IDK how we would enforce it - not saying everything needs updating in this PR, but I do think "what timeout is actually enforced" is tricky

bobheadxi: WDYT @eseliger ?

unknwon:
Thanks again for the reviews! Merging as-is (3 other repos are eagerly waiting to import rather than replicate the same logic proposed here 😅), happy to address post-merge feedback!

PR #62137, opened 2024-04-24
managedservicesplatform: add generalized 'HA' toggle for Redis, Cloud SQL

bobheadxi:
This change adds an explicit 'HA' toggle for Redis and Cloud SQL, as part of investigation into https://github.com/sourcegraph/managed-services/issues/311:

- `redis.highAvailability`: sets the standard HA mode. Right now, we do this by default to preserve existing behaviour - a follow-up to this PR would be to explicitly set the HA mode on production services, then make this default `false`.
- `postgreSQL.highAvailability`: enables regional mode, and also point-in-time-recovery as required. This could be quite expensive - we have ~$200/mo of Cloud SQL expenses, mostly in CPU before discounts, so this is projected to ~double that bill, but would be the simplest and most reliable way to maintain uptime in the event of a zonal failover.

The plan is not necessarily to immediately make use of `postgreSQL.highAvailability` but have the option open, as the configuration is trivial - we could still develop a manual failover process to adhere to requirements outlined in https://github.com/sourcegraph/managed-services/issues/311

Preparing a summary here: https://www.notion.so/sourcegraph/MSP-service-availability-655e89d164b24727803f5e5a603226d8?pvs=4

## Test plan

`sg msp generate -all` has no diff


dev/managedservicesplatform/internal/resource/cloudsql/cloudsql.go
```diff
@@ -106,8 +111,14 @@ func New(scope constructs.Construct, id resourceid.ID, config Config) (*Output,
...
-				Enabled:                     pointers.Ptr(true),
-				PointInTimeRecoveryEnabled:  pointers.Ptr(false), // PITR uses a lot of resources and is cumbersome to use
+				Enabled: pointers.Ptr(true),
+				PointInTimeRecoveryEnabled: pointers.Ptr(func() bool {
+					// PITR uses a lot of resources and is cumbersome to use -
+					// only enable it for services that require HA, since it is
+					// required for regional deployments:
+					// https://cloud.google.com/sql/docs/mysql/high-availability#backups-and-restores
+					return pointers.DerefZero(config.Spec.HighAvailability)
+				}()),
```
bobheadxi, line 121: Actually this link is for MySQL, the Postgres page doesn't specifically call out PITR requirement: https://cloud.google.com/sql/docs/postgres/high-availability#backups-and-restores

TODO look into this a bit more

jac: The Terraform/REST config guides to enabling HA do enable pitr 🤔 
https://cloud.google.com/sql/docs/postgres/configure-ha#terraform

unknwon (approved):
Looks good other than the TODO-comment that always exists

dev/managedservicesplatform/internal/resource/cloudsql/cloudsql.go
bobheadxi: I'll take that as confirmation it's required 😆 The docstring on PITR in the Terraform provider indicates something similar

Issue #61974, opened 2024-04-17
Context: handle qualified symbol names


jtibshirani:
Currently, if your search has a qualified symbol name like `indexData.scoreFile` or `grf::causal_forest`, we can fail to find results. This is likely because we only tokenize on whitespace.

Here's [an example using `patterntype:codycontext`](https://sourcegraph.sourcegraph.com/search?q=context:global+repo:sourcegraph/zoekt+indexData.scoreFile&patternType=codycontext&sm=0.) that returns no results, when it should find relevant snippets.

[Original Slack thread](https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1713221410344119?thread_ts=1713221347.773559&cid=C04MSD3DP5L)

Relates to #61571

/cc @sourcegraph/search-platform

PR #62142, opened 2024-04-24
context: handle qualified symbol names

stefanhengl:
Relates to #61974 

With this change we split patterns on common namespace separators like `::` or `.`. 

Test plan:
- updated unit tests
- manual testing
image


internal/search/codycontext/query_transformer_test.go
```diff
@@ -9,6 +10,12 @@ import (
 )
 
 func TestTransformPattern(t *testing.T) {
+	old := maxTransformedPatterns
+	defer func() {
+		maxTransformedPatterns = old
+	}()
+	maxTransformedPatterns = math.MaxInt
```
keegancsmith, line 17: minor: should we consider this being an input to transformPattern rather than switching it to a var?

internal/search/codycontext/term_utils.go
```diff
@@ -11,6 +11,20 @@ func removePunctuation(input string) string {
 	return strings.TrimFunc(input, unicode.IsPunct)
 }
 
+var separators = map[rune]struct{}{
+	':': {}, // C++, Rust
+	'.': {}, // Java, Go, Python
```
keegancsmith, line 16: should we also consider `->`, although that isn't a single rune.

keegancsmith (approved):

internal/search/codycontext/query_transformer_test.go
stefanhengl: I moved the cropping out of the function.

keegancsmith (approved):

internal/search/codycontext/query_transformer_test.go
```diff
@@ -27,6 +27,11 @@ func TestTransformPattern(t *testing.T) {
 		"computing",
 		"own", // key terms should not be removed, even if they are common
 		"!?",  // punctuation-only token should be removed
+		"grf::causal_forest",
```
jtibshirani, line 30: Nice tests. This seems good to me -- we split on "qualifier characters" like `::`, but we don't touch punctuation in the middle of names like underscores.

jtibshirani (approved):
This looks good to me. I have a slight concern around splitting on `.` -- this will cause us to split up URLs, which is likely not desirable? But for now this seems like the right trade-off, and we can always add more nuanced tokenization in the future.

Before merging, could you run the context eval (https://github.com/sourcegraph/context-eval) to ensure we don't regress anything? Happy to help get set up with this. Also, let's add a couple of C++ examples to that to capture this use case!

stefanhengl:
> Before merging, could you run the context eval (https://github.com/sourcegraph/context-eval) to ensure we don't regress anything? Happy to help get set up with this. Also, let's add a couple of C++ examples to that to capture this use case!

```
Running evaluation...

Breakdown by class:
Find symbol     7/9
Find string     2/2
Explain concept 4/5
Check dependency        1/2
Find logic      16/22
Gather information      6/9
Changelog       0/2
Ownership       2/2

Breakdown by file type:
TSX     4/5
Golang  18/23
Typescript      8/13
Markdown        5/8
Graphql 1/1
Yaml    0/1
Codeowners      2/2

Combined recall 38/53
```

Numbers are identical to the [latest snapshot](https://github.com/sourcegraph/context-eval/blob/main/golden_queries/snapshots/enterprise_keyword/2024-04-23_8edfc0f4.out)

PR #62154, opened 2024-04-24
appliance: simplify k8s resource factories

craigfurman:
This commit removes most of the code associated with configuring defaults for k8s resources, while removing none of those sensible defaults.

Remove builder functions in favour of just manipulating the struct. This:

```
obj, err := NewObj(
  name,
  WithFieldFoo(foo),
  WithFieldBar(bar),
)
```

Becomes:

```
obj := NewObj(name)
obj.Foo = foo
obj.Bar = bar
```

No error is returned because we don't have to assume that a general interface for builders can return an error. If a code path that creates a field (e.g. foo or bar above) can return an error, handle and propagate that error as normal - it is the exception, not the rule.

No unit tests are needed for a function that returns a struct with sensible defaults. Only the indirection of the builders necessitated these tests. These functions now have no behavior, and asserting that such a struct equals some value is just double-entry book-keeping. The golden tests should catch regressions in resource configs generated by the appliance.

Utility functions in the maps package, like MergePreservingExistingKeys, may well be useful for some resource configs. These are left in place for those use cases, although none have arisen yet.

## Test plan

This PR replaces the need for unit testing builders, with reliance on the golden test mechanism introduced by previous PRs to the appliance package.


jdpleiness (approved):
Nuke it 💣 

Just for some history before we nuke and for anyone looking back at this PR, the idea behind the helper functions was to "some day" add validation and make this more friendly for devs that might not be as familiar with K8s but still need to create a resource with some sensible defaults. However, most these functions didn't add any defaults or do any kind of validation, so the abstraction did fall flat.

With the golden testing at the manifest level, and these no longer being builders, I think we're "golden" 😏

craigfurman:
Nice one 😁 

I'll expand this to cover all resources, not just deployments, pods, and containers, then rebase on main once my other branch merges.

PR #62156, opened 2024-04-24
Svelte: add gap between line numbers and content

camdencheek:
Tiny style thing that I think was making the blob view look bad. Note that, because blame view is rendered inside the content rather than as a gutter (which I think is very weird still), the padding is replaced by the blame view when it's enabled. I think it looks fine, it's just weird.

## Test plan

Before:
screenshot-2024-04-24_10-14-36@2x

After:
screenshot-2024-04-24_10-14-06@2x


client/web-sveltekit/src/lib/CodeMirrorBlob.svelte
```diff
@@ -61,7 +61,7 @@
             color: 'var(--line-number-color)',
         },
         '.cm-line': {
-            paddingLeft: '0',
+            paddingLeft: '1ex',
```
camdencheek, line 64: Using `1ex` to make it a one-character-width  space

vovakulikov (approved):

PR #62157, opened 2024-04-24
Search: fix case-sensitive queries with lang filters

jtibshirani:
This PR fixes a regression where language filters might not match when case
sensitivity is enabled. Before, we just appended `(?i)` to the file filters to
indicate case insensitivity, but still passed a case-sensitive file query to
Zoekt (called `case_file_regex`). Now, we make sure that the Zoekt queries are
also case insensitive so there is no conflict in how to interpret them.

Co-authored-by: Matthew Manela <mmanela@users.noreply.github.com>

## Test plan

Adapted unit tests. TODO: add an end-to-end test exercising the original bug!

### DIFF
--- a/internal/search/zoekt/query.go
+++ b/internal/search/zoekt/query.go
@@ -52,27 +52,23 @@ func QueryToZoektQuery(b query.Basic, resultTypes result.Types, feat *search.Fea
 			and = append(and, &zoekt.Not{Child: filter})
 		}
 	} else {
-		filesInclude = append(filesInclude, mapSlice(langInclude, query.LangToFileRegexp)...)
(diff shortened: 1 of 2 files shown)


dev/gqltest/search_test.go
```diff
@@ -260,6 +260,27 @@ func testSearchClient(t *testing.T, client searchClient) {
...
+		}
+
+		// Ensure some results are returned
+		if len(results.Results) == 0 {
+			t.Fatal("Want non-zero results but got none")
+		}
+
+		// Make sure we only got .go files
+		for _, r := range results.Results {
+			if !strings.Contains(r.File.Name, ".go") {
```
mmanela, line 278: should this technically match insensitively in case a file has a  `.Go` extension?

mmanela (approved):

Issue #44881, opened 2022-11-29
Commit filter on UI doesn't allow search using commit hash


rakeshjosh2003:
Currently, the commit filter on a file being searched allows one to search the commits based on commit headlines and not commit hash. I have tested this internally and replicated the behavior. The search should allow searching with a commit hash.
Attaching a short video for reference. Thank you.

https://sourcegraph2020.lightning.force.com/lightning/r/Account/0013t00001XicE2AAJ/view

rakeshjosh2003:
https://user-images.githubusercontent.com/102702631/204430252-e6e9fc51-cabd-45bc-9c2f-7767f5f12fcd.mov

rakeshjosh2003:
https://user-images.githubusercontent.com/102702631/204431135-21bb9196-8705-4ce1-bfb5-01e266adafcf.mov

camdencheek:
Oh whoops, this seems to be about using the commit hash in a search, not about using the rev picker. Reopening

Issue #18792, opened 2021-03-02
Searching commits should support results on commit SHAs


chrislovett03:
When using the `Search commits.. ` field in the attached view - you can't search for the commit id itself. Users would like the ability to search both messages **and** commit sha in this field. 

<img width="908" alt="Screen Shot 2021-03-01 at 3 58 01 PM" src="https://user-images.githubusercontent.com/12041606/109683163-16ca7f80-7b4d-11eb-851e-af95f2909720.png">

christinelovett:
https://app.hubspot.com/contacts/2762526/company/578600789

rvantonder:
>  Via @umpox: It seems to me that sending a git commit hash in query should be the best way to return this results, in this case it’d be the backends responsibility to update the logic that queries ancestors.

![image](https://user-images.githubusercontent.com/888624/109703583-dec63b00-7b52-11eb-8b07-a21c8c9c4248.png)

If I follow correctly, commit ID here refers to commit sha (or short sha). Any reason we couldn't just look at the entered input, check if it looks like a hash (detect `if (value.match(/^[\dA-Fa-f]+$/) && value.length > 6)`) and then query the backend at this sha, like we see in your screenie ^ and update the list here? Am I missing something?

christinelovett:
Thanks for naming correction @rvantonder - updated :-)

umpox:
@rvantonder So I assume you mean that we could update `revision` with the commit sha and clear out `query` to return a specific commit.

My thoughts are that this feels a bit of a potential code smell, we'll no longer be able to directly map the users input to the `query` field, we'll have to add this logic to detect the type of their query and if it's a commit sha handle the mutation variables different.

I'd also say that this is no longer 'searching', more selecting a specific commit. If you only remember the first 3 characters of a commit sha (i.e. you haven't just copy-pasted it) I think you should still be able to view results and select the correct one from also viewing the message.

cc: @sourcegraph/frontend-platform Interested to hear your thoughts here

pdubroy:
I don't think the frontend is the right place to do the query interpretation. Specifically, I'd be concerned about the following scenarios:

- Normal text that happens to look like a SHA, e.g. "deface", "feed01".
- Other hexadecimal strings that could appear in comments, e.g. any kind of ID

I'd worry that to handle these properly, we might have to execute multiple queries from the frontend.

rvantonder:
[...]

PR #62160, opened 2024-04-24
Svelte: allow picking by commit hash

camdencheek:
This updates the revision picker to support searching for commits by hash. This has been a long-standing product gap in the React webapp, and the same problem exists in the Svelte webapp. 

I did not fix this in the React webapp as well because our pagination logic is very tied to GraphQL pagination, and I couldn't come up with an easy way to add an item to the list that is outside of a paginated list of nodes.

Fixes https://github.com/sourcegraph/sourcegraph/issues/18792 in the Svelte webapp
Fixes https://github.com/sourcegraph/sourcegraph/issues/44881 in the Svelte webapp

## Test plan

https://github.com/sourcegraph/sourcegraph/assets/12631702/d2ea52e7-eded-42cb-a874-341002af8e15

### DIFF
--- a/client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryRevPicker.gql
+++ b/client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryRevPicker.gql
@@ -22,25 +22,13 @@ fragment RepositoryGitRefs on Repository {
     }
 }
 
-# Used to query data for repository git commits
-# NOTE: the query which is going to use this fragment is supposed
-# to have query and anchor revision variables
-fragment RepositoryGitCommits on Repository {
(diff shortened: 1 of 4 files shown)


client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryRevPicker.gql
```diff
@@ -22,25 +22,13 @@ fragment RepositoryGitRefs on Repository {
     }
 }
 
-# Used to query data for repository git commits
-# NOTE: the query which is going to use this fragment is supposed
-# to have query and anchor revision variables
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryRevPicker.gql:27]: This hoists the variable stuff all up into the query and leaves the fragment to just be the type for the commit.

vovakulikov (approved):

PR #62161, opened 2024-04-24
workflows: remove sg-msp variant from sg publish

bobheadxi:
We no longer require a build variant, and MSP is now included by default since https://github.com/sourcegraph/sourcegraph/pull/59087.

Remaining reference to the variant will be removed in https://github.com/sourcegraph/managed-services/pull/1288. Do not merge this until https://github.com/sourcegraph/managed-services/pull/1288 is merged.

## Test plan

https://github.com/sourcegraph/managed-services/pull/1288


jhchabran (approved):
Approving so it's ready when you need to merge this :)

unknwon (approved):

PR #62162, opened 2024-04-24
Context: allow embeddings only for testing

jtibshirani:
As we're evaluating the quality of Cody's context, it's really useful to be
able to compare against Enterprise embeddings. However, we completely disabled
Enterprise embeddings when we turned them off for customers.

This PR adds an environment variable `SRC_EMBEDDINGS_ALLOWED` that lets us
enable and test embeddings locally. It is not advertised and should never be
enabled in production use.

## Test plan

Local testing using the [context eval repo](https://github.com/sourcegraph/context-eval). The README now contains instructions
for how to run locally using `SRC_EMBEDDINGS_ALLOWED=true sg start`.

### DIFF
--- a/internal/conf/computed.go
+++ b/internal/conf/computed.go
@@ -13,7 +13,6 @@ import (
 	"github.com/sourcegraph/sourcegraph/internal/conf/confdefaults"
 	"github.com/sourcegraph/sourcegraph/internal/conf/conftypes"
 	"github.com/sourcegraph/sourcegraph/internal/conf/deploy"
-	"github.com/sourcegraph/sourcegraph/internal/dotcom"
 	"github.com/sourcegraph/sourcegraph/internal/hashutil"
 	"github.com/sourcegraph/sourcegraph/internal/license"
 	srccli "github.com/sourcegraph/sourcegraph/internal/src-cli"
@@ -916,8 +915,8 @@ func GetEmbeddingsConfig(siteConfig schema.SiteConfiguration) *conftypes.Embeddi
 		return nil
 	}
 
-	// Only allow embeddings on dotcom
-	if !dotcom.SourcegraphDotComMode() {
+	// Only allow embeddings as part of evaluating context quality.
+	if !ForceAllowEmbeddings() {
 		return nil
 	}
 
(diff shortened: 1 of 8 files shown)


internal/conf/computed.go
```diff
@@ -916,8 +915,8 @@ func GetEmbeddingsConfig(siteConfig schema.SiteConfiguration) *conftypes.Embeddi
 		return nil
 	}
 
-	// Only allow embeddings on dotcom
-	if !dotcom.SourcegraphDotComMode() {
+	// Only allow embeddings as part of evaluating context quality.
+	if !ForceAllowEmbeddings() {
```
jtibshirani, line 919: This cleans up something that was previously confusing. We actually don't ever want embeddings to be enabled for dot com, and have explicitly disabled them (https://github.com/sourcegraph/deploy-sourcegraph-cloud/pull/18537)

jtibshirani:
I'm not 100% confident in this approach, very open to feedback on how to do this.

internal/conf/embeddings.go
```diff
@@ -0,0 +1,19 @@
+package conf
+
+import (
+	"strconv"
+
+	"github.com/sourcegraph/sourcegraph/internal/env"
+)
+
+var embeddingsAllowed, _ = strconv.ParseBool(env.Get(
+	"SRC_EMBEDDINGS_ALLOWED",
```
rafax, line 10: Not sure if we have prior art for this, but if this is really internal-only/unsafe, maybe we should call it `SRC_INTERNAL_EMBEDDINGS_ALLOWED` or `SRC_UNSAFE_EMBEDDINGS_ALLOWED` (to stop people from poking around and killing their instance)? But, up to you.

mmanela (approved):

internal/conf/computed.go
rafax: I think there was a (short) period of time where we allowed Embeddings on .com (while they were used by PLG), but removed them from upcoming Sourcegraph release? But, your code makes sense.

rafax (approved):

internal/conf/embeddings.go
jtibshirani: Good idea, I renamed this to `SRC_INTERNAL_EMBEDDINGS_ALLOWED`.

PR #62163, opened 2024-04-24
Fixing OpenAI TokenUsage to replace internally implemented token counter with openai response

arafatkatze:
Thanks to @rafax 
image

Next Steps:
- Try to make it work in Azure too(It didn't work in my tiny experiments)
- If you can't make something similar work in azure then write a separate tokenizer for azure

## Test plan
Tested locally with the entire debugger environment

### DIFF
--- a/internal/completions/client/openai/openai.go
+++ b/internal/completions/client/openai/openai.go
@@ -104,6 +104,8 @@ func (c *openAIChatCompletionStreamClient) Stream(
 	dec := NewDecoder(resp.Body)
 	var content string
 	var ev types.CompletionResponse
+	var promptTokens, completionTokens int
+
 	for dec.Scan() {
 		if ctx.Err() != nil && ctx.Err() == context.Canceled {
 			return nil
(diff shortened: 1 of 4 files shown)


internal/completions/client/openai/openai.go
```diff
@@ -120,6 +122,14 @@ func (c *openAIChatCompletionStreamClient) Stream(
 			return errors.Errorf("failed to decode event payload: %w - body: %s", err, string(data))
 		}
 
+		// These are only included in the last message, so we're not worried about overwriting
+		if event.Usage.PromptTokens > 0 {
```
abeatrix, line 126: I know you said we don't need to worry about it being overwritten, but maybe we can make it safer by allowing this to be updated if the promptToken var doesn't have any value?

(Go noob here so feel free to ignore me if it's not needed)

abeatrix (approved):
The code lgtm since it's just name changing, and I'm having troubles connecting to my local cody gateway instance to verify this change, so maybe @rafax can help with verifying? 

If not, do you mind sharing a loom with me on the steps I need to take to test this please 🙇🏻‍♀️

internal/completions/client/openai/openai.go
arafatkatze: I am just maintaining consistency with what rafal did [here](https://sourcegraph.com/-/editor?remote_url=https%3A%2F%2Fgithub.com%2Fsourcegraph%2Fsourcegraph.git&branch=openai-tokenusage-update&file=cmd%2Fcody-gateway%2Finternal%2Fhttpapi%2Fcompletions%2Fopenai.go&editor=VSCode&version=2.2.16&start_row=217&start_col=12&end_row=224&end_col=2) and we know that works from many months of data so I wanna keep the same logic

arafatkatze:
Hey @abeatrix 
Sorry I had a very busy last week because of some team priority item shifts so I totally forgot about this issue.

But I realize that debugging this can be super helpful so I made a [video](https://www.loom.com/share/439cc42b4d9d4660b9d27c51fee6b8b5?sid=1e5d84e6-cc38-4a47-952f-0727e4ed1e49)

Issue #61689, opened 2024-04-08
gitserver: Move FirstEverCommit to new gRPC pattern


eseliger:
Note: This is known to be very slow. Maybe figure out if we can get rid of it instead. It will perform poorly for large repos.

PR #62158, opened 2024-04-24
gitserver: grpc: create exhaustive logger for repository service

ggilmore:
This PR builds on https://github.com/sourcegraph/sourcegraph/pull/61270, and adds the request logging funcionality to the repository grpc server in gitserver.

## Test plan

Local testing:

Ran `env SRC_GITSERVER_EXHAUSTIVE_LOGGING_ENABLED=true SRC_LOG_SCOPE_LEVEL=gitserver.gRPCRequestLogger=debug sg start` locally, and saw the following output from the logger in my terminal (note that specifically the repository service is mentioned):

```
[    gitserver-0] DEBUG gitserver.gRPCRequestLogger internal/server_grpc_logger.go:46 Handled FetchRepository RPC {"server": "gitserver.v1.GitserverRepositoryService", "method": "FetchRepository", "status": "OK", "traceID": "67f1aa837367fa7ea72b164e70a67419", "duration": "1.068438833s", "request": {"repoName": "github.com/sgtest/megarepo"}}
```

PR #62165, opened 2024-04-24
gitserver: port FirstEverCommit functionality from client to git cli backend

ggilmore:
Part of [#61689](https://github.com/sourcegraph/sourcegraph/issues/61689)

## Overview

This PR adapts the existing FirstEverCommit logic from the gitserver client to our new git cli backend. This is the first step as part of a set of stacked PRs to port the functionality to gRPC.

Existing `func (c *clientImplementor) FirstEverCommit()` implementation for reference: 

https://github.com/sourcegraph/sourcegraph/blob/412654f098492856f32a6c561eda471800924d71/internal/gitserver/commands.go#L1734-L1757

## Note

One of the existing tests that I ported called out that code insights depended on the formatting of a specific error string in order to detect if the repository was empty.

(Test in question):

https://github.com/sourcegraph/sourcegraph/blob/948848bca6e87e84878a9c1a9222d36734103fa2/internal/gitserver/commands_test.go#L721-L729

(String checking logic in code insights):

https://github.com/sourcegraph/sourcegraph/blob/2443801b4ea4a4d48a8badf3c34ed2761af65db4/internal/insights/gitserver/first_commit.go#L18-L38

This is quite brittle, as any unrelated formatting change with our errors could change the semantics here. I decided to create a new sentinel error for this condition, `gitdomain.RepoEmptyErr`, to 1) better state our intent and 2) make this check more robust against minor formatting changes in our error libraries. When we generate the gRPC bindings, we'll need to create a new protobuf message for this error and write `ToProto/FromProto` functions, etc.

## Test plan

Adapted existing unit tests

### DIFF
--- a/cmd/gitserver/internal/git/gitcli/odb.go
+++ b/cmd/gitserver/internal/git/gitcli/odb.go
@@ -187,3 +187,39 @@ func (g *gitCLIBackend) getBlobOID(ctx context.Context, commit api.CommitID, pat
 	}
 	return api.CommitID(fields[2]), nil
 }
+
+func (g *gitCLIBackend) FirstEverCommit(ctx context.Context) (api.CommitID, error) {
+	rc, err := g.NewCommand(ctx, WithArguments("rev-list", "--reverse", "--date-order", "--max-parents=0", "HEAD"))
+	if err != nil {
+		return "", err
+	}
+	defer rc.Close()
+
+	out, err := io.ReadAll(rc)
+	if err != nil {
+		var cmdFailedErr *CommandFailedError
+		if errors.As(err, &cmdFailedErr) {
+			if cmdFailedErr.ExitStatus == 129 && bytes.Contains(cmdFailedErr.Stderr, []byte(revListUsageString)) {
+				// If the error is due to an empty repository, return a sentinel error.
+				e := &gitdomain.RevisionNotFoundError{
+					Repo: g.repoName,
+					Spec: "HEAD",
+				}
+				return "", e
+			}
+		}
+
+		return "", errors.Wrap(err, "git rev-list command failed")
+	}
+
+	lines := bytes.TrimSpace(out)
+	tokens := bytes.SplitN(lines, []byte("\n"), 2)
+	if len(tokens) == 0 {
+		return "", errors.New("FirstEverCommit returned no revisions")
+	}
+	first := tokens[0]
+	id := api.CommitID(bytes.TrimSpace(first))
+	return id, nil
+}
+
+const revListUsageString = `usage: git rev-list [<options>] <commit>... [--] [<path>...]`
--- a/cmd/gitserver/internal/git/gitcli/odb_test.go
+++ b/cmd/gitserver/internal/git/gitcli/odb_test.go
@@ -2,12 +2,15 @@ package gitcli
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
 	"runtime"
 	"testing"
+	"time"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/stretchr/testify/require"
 
 	"github.com/sourcegraph/sourcegraph/cmd/gitserver/internal/git"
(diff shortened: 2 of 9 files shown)


ggilmore:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62165?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#62165** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62165?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#62158** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62158?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

cmd/gitserver/internal/git/gitcli/odb.go
```diff
@@ -187,3 +188,42 @@ func (g *gitCLIBackend) getBlobOID(ctx context.Context, commit api.CommitID, pat
...
+		return nil, err
+	}
+	defer rc.Close()
+
+	out, err := io.ReadAll(rc)
+	if err != nil {
+		var cmdFailedErr *CommandFailedError
+		if errors.As(err, &cmdFailedErr) {
+			if cmdFailedErr.ExitStatus == 129 && bytes.Contains(cmdFailedErr.Stderr, []byte(revListUsageString)) {
+				// If the error is due to an empty repository, return a sentinel error.
```
ggilmore, line 206: This exit condition comes from https://sourcegraph.com/github.com/git/git@bf995e7a4f94a9388aa8042dc9e338f3fcb75496/-/blob/builtin/rev-list.c?L701-706 in the source I believe.

cmd/gitserver/internal/git/observability.go
```diff
@@ -369,6 +369,17 @@ func (b *observableBackend) ContributorCounts(ctx context.Context, opt Contribut
...
 
+func (b *observableBackend) FirstEverCommit(ctx context.Context) (_ *gitdomain.Commit, err error) {
+	ctx, _, endObservation := b.operations.firstEverCommit.With(ctx, &err, observation.Args{})
+	defer endObservation(1, observation.Args{})
+
+	concurrentOps.WithLabelValues("FirstEverCommit").Inc()
+	defer concurrentOps.WithLabelValues("FirstEverCommit").Dec()
+
+	return b.backend.FirstEverCommit(ctx)
+
```
eseliger, line 380:
```suggestion
```

internal/insights/gitserver/first_commit.go
```diff
@@ -18,6 +18,10 @@ var (
 const emptyRepoErrMessage = `git command [rev-list --reverse --date-order --max-parents=0 HEAD] failed (output: ""): exit status 129`
 
 func isFirstCommitEmptyRepoError(err error) bool {
+	if gitdomain.IsRepoEmptyErr(err) {
+		return true
+	}
+
 	if strings.Contains(err.Error(), emptyRepoErrMessage) {
 		return true
```
eseliger, line 26: Should we remove this, after the gRPC API PR is merged? Once the client makes use of the gRPC API there's no way this could still be returned.

cmd/gitserver/internal/git/gitcli/odb.go
eseliger: strange that it'd print the usage 😬 but thanks for the link, very helpful!

cmd/gitserver/internal/git/gitcli/odb_test.go
```diff
@@ -253,3 +258,68 @@ func TestRepository_GetCommit(t *testing.T) {
 		}, c)
 	})
 }
+
+func TestRepository_FirstEverCommit(t *testing.T) {
+	ctx := actor.WithActor(context.Background(), &actor.Actor{
+		UID: 1,
+	})
```
eseliger, line 265: I don't think we make any checks for the actor, is this needed? No dependency on sg actors in this package would be great, less coupling.

cmd/gitserver/internal/git/gitcli/odb.go
```diff
@@ -187,3 +188,42 @@ func (g *gitCLIBackend) getBlobOID(ctx context.Context, commit api.CommitID, pat
...
+				return nil, &gitdomain.RepositoryEmptyError{Repo: g.repoName}
+			}
+		}
+		return nil, errors.Wrap(err, fmt.Sprintf("git command %v failed (output: %q)", args, out))
```
eseliger, line 210: I think you can trust the command helpers to return relevant information instead of adding output and args here again - that'll let us encapsulate potential error redaction as well.
At least that's how I think the other commands treat this.

cmd/gitserver/internal/git/gitcli/odb.go
```diff
@@ -187,3 +188,42 @@ func (g *gitCLIBackend) getBlobOID(ctx context.Context, commit api.CommitID, pat
...
+		return nil, errors.New("FirstEverCommit returned no revisions")
+	}
+	first := tokens[0]
+	id := api.CommitID(bytes.TrimSpace(first))
+
+	c, err := g.GetCommit(ctx, id, false)
+	if err != nil {
+		return nil, errors.Wrapf(err, "getting commit information for commit %q", id)
+	}
+
```
eseliger, line 225: I wonder if, from an API perspective, we should internally just return the `api.CommitID` (for maximum composability) and then the presentation layer can make the second call to `GetCommit` if it's interested in that?

cmd/gitserver/internal/git/iface.go
```diff
@@ -95,6 +95,9 @@ type GitBackend interface {
 	// No new usages of it should be introduced and once the migration is done we will
 	// remove this method.
 	Exec(ctx context.Context, args ...string) (io.ReadCloser, error)
+
+	// FirstEverCommit returns the first commit ever made to the repository.
+	FirstEverCommit(ctx context.Context) (*gitdomain.Commit, error)
```
eseliger, line 100: We should document the well-known error type we return here for empty repo (or revnotfound if we decide to go with that)

eseliger (commented):
> This is quite brittle, as any unrelated formatting change with our errors could change the semantics here. I decided to create a new sentinel error for this condition, gitdomain.RepoEmptyErr, to 1) better state our intent and 2) make this check more robust against minor formatting changes in our error libraries. When we generate the gRPC bindings, we'll need to create a new protobuf message for this error and write ToProto/FromProto functions, etc.

Strongly agree that no caller should depend on any git error message. We want to fully own the process of potentially switching to go-git, libgit2, or any self-made implementations that will yield different error strings. 

One question - in alignment with gRPC saying ["there should not be too many different error codes, because callers will most likely not add more than a few checks"](https://cloud.google.com/apis/design/errors#error_codes), should we try to keep the error types in check as well, and repurpose the `RevisionNotFoundError` here?
I think it's technically a correct error, since FirstEverCommit would try to look at HEAD, HEAD resolves to no branch yet, so HEAD is not found?

internal/insights/gitserver/first_commit.go
ggilmore: https://github.com/sourcegraph/sourcegraph/pull/62302

ggilmore:
### Merge activity

* **May 1, 12:14 AM EDT**: @ggilmore merged this pull request with [Graphite](https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62165).

PR #62170, opened 2024-04-24
Svelte: render search alerts

camdencheek:
Before this, we were completely ignoring any alerts that come from the search backend. This fixes that.

## Test plan

Added simple playwright test that ensures the alert is visible if it is sent on the stream.

screenshot-2024-04-24_15-06-51@2x


fkling (approved):
Thank you!

PR #62172, opened 2024-04-24
appliance: add appliance skeleton

jdpleiness:
Add appliance service skeleton with configuration and tie into gRPC stub from #62171 

## Test plan

Tested gRPC endpoints locally. As these are just stubs and will certainly be changed before shipping, real e2e and integration testing will be added at that time.


jdpleiness:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62172?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#62172** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62172?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#62171** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62171?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

 Join @jdpleiness and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

internal/appliance/appliance.go
```diff
@@ -0,0 +1,93 @@
...
+
+const (
+	StageUnknown         Stage = "unknown"
+	StageIdle            Stage = "idle"
+	StageInstall         Stage = "install"
+	StageInstalling      Stage = "installing"
+	StageUpgrading       Stage = "upgrading"
+	StageWaitingForAdmin Stage = "waitingForAdmin"
+	StageRefresh         Stage = "refresh"
+)
```
jdpleiness, line 44: These are a carry-over from our POC and can be adjusted as we go, but they did work okay in the POC

internal/appliance/appliance.go
```diff
@@ -0,0 +1,93 @@
...
+				"deploy": "sourcegraph",
+			},
+			Annotations: map[string]string{
+				// required annotation for our controller filter.
+				"appliance.sourcegraph.com/managed": "true",
+			},
+		},
+		Immutable: pointers.Ptr(false),
+		BinaryData: map[string][]byte{
+			"spec": spec,
```
jdpleiness, line 88: This will likely need to be changed as well. I think we could keep this in a configmap in the binary as is, but there might be a better way as well 🤔

cmd/appliance/shared/config.go
```diff
@@ -1,20 +1,42 @@
...
+	if err != nil {
+		if home := homedir.HomeDir(); home != "" {
+			kubeConfig = c.Get("KUBECONFIG", filepath.Join(home, ".kube", "config"), "Absolute path to the kubeconfig file.")
+		}
+
+		k8sConfig, err = clientcmd.BuildConfigFromFlags("", kubeConfig)
+		if err != nil {
+			c.AddError(errors.New("could not create kubernetes client config"))
+		}
+	}
```
craigfurman, line 34: non-blocking note: It would be nice if we can log that we're trying to fall back from in-cluster config to kubeconfig, but we might not even have a logger at this stage. If this is hard, disregard.

internal/appliance/reconcile.go
```diff
@@ -55,13 +55,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	defer r.Recorder.Event(&applianceSpec, "Normal", "ReconcileFinished", "Reconcile finished.")
 
 	// TODO place holder code until we get the configmap spec'd out and working'
-	data, ok := applianceSpec.Data["spec"]
+	data, ok := applianceSpec.BinaryData["spec"]
```
craigfurman, line 58: Was about to ask why, then realised this is probably just to avoid converting back in yaml.Unmarshal, right? If so, all good

internal/appliance/grpc.go
```diff
@@ -6,17 +6,14 @@ import (
...
 
-func (s *GRPCServer) GetApplianceVersion(ctx context.Context, request *pb.GetApplianceVersionRequest) (*pb.GetApplianceVersionResponse, error) {
-	return nil, nil
+	return &pb.GetApplianceVersionResponse{Version: version.String()}, nil
 }
 
-func (s *GRPCServer) GetApplianceStage(ctx context.Context, request *pb.GetApplianceStageRequest) (*pb.GetApplianceStageResponse, error) {
-	return nil, nil
+func (a *Appliance) GetApplianceStage(ctx context.Context, request *pb.GetApplianceStageRequest) (*pb.GetApplianceStageResponse, error) {
+	stage := a.GetCurrentStage(ctx)
```
craigfurman, line 16: What exactly is a stage here? I'm not super knowledgable about gRPC, but I can't find that term in its docs, so I'm guessing it's part of the SG service framework? `grep -i stage doc/dev` doesn't help me though 🤔

internal/appliance/appliance.go
```diff
@@ -0,0 +1,93 @@
...
+	configMap := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			Labels: map[string]string{
+				"deploy": "sourcegraph",
+			},
+			Annotations: map[string]string{
+				// required annotation for our controller filter.
+				"appliance.sourcegraph.com/managed": "true",
```
craigfurman, line 83: There's a constant for this key in reconcile.go

internal/appliance/appliance.go
```diff
@@ -0,0 +1,93 @@
...
+
+func (a *Appliance) GetCurrentVersion(ctx context.Context) semver.Version {
+	return a.version
+}
+
+func (a *Appliance) GetCurrentStage(ctx context.Context) Stage {
+	return a.status.stage
+}
+
+func (a *Appliance) CreateConfigMap(ctx context.Context, name, namespace string) (*corev1.ConfigMap, error) {
```
craigfurman, line 68: This isn't used AFAICT, and no worries if we're just setting up a skeleton to fill in later, but I figured this might be a good opportunity to discuss how this thing might work! 😁 

My initial guess is: on startup, the appliance checks for a configmap. If it doesn't exist, we know we're doing initial setup, and need to guide the user through setting up initial config. If it does exist, we know there's already an SG deployment. I'd imagine that in the early days, there's nothing for the appliance to do if SG already exists, but of course, we'll add upgrade facilities and some pane of glass for o11y / health-checking eventually too (without duplicating the existing dashboard pages of course).

Basically, the configmap can be the sole source of state for the appliance (unless we discover the need for more). WDYT?

internal/appliance/appliance.go
```diff
@@ -0,0 +1,93 @@
...
+	pb "github.com/sourcegraph/sourcegraph/internal/appliance/v1"
+	"github.com/sourcegraph/sourcegraph/lib/pointers"
+)
+
+type Appliance struct {
+	client client.Client
+
+	version     semver.Version
+	status      Status
+	sourcegraph Sourcegraph
```
craigfurman, line 21: This isn't created anywhere here, but similar to another note, no worries since this is just a skeleton. One thing that comes to mind, is for first installation, we'll need to set a version on the Spec. I guess each appliance revision can come hardcoded with some version(s) of SG that it can support installing?

Looking ahead a bit to upgrades, something we jotted down in https://docs.google.com/document/d/1He4TUxR81_kcakGPQgn7tumGMXhcfko7I9QzEYMhX6E/edit:

> A given Appliance version can deploy a certain range of older SG resource sets

This is really just me thinking out loud trying to develop a shared understanding of what we might do and doesn't really pertain to this PR, sorry! 😂

craigfurman (approved):
Similar to last one, not sure if you were looking for ✅ given that this is a draft, dismiss if not!

internal/appliance/reconcile.go
jdpleiness: Yup, it was just easier to serialize it into the configmap this way 😆

internal/appliance/grpc.go
craigfurman: Ah, I just noticed the stage constants. You even left a note on them. I suppose, we think the appliance has a single global state at any one time, and it may be useful to know that state from deep within the code, possibly in order to mutex certain operations, e.g. upgrades? Sounds plausible if so. I'd like to learn more about the state machine aspects of the PoC, but also, I don't think we need to plan these impl details up-front. They can just fall out naturally from implementation. I'm aware of the irony of saying that after I speculated wildly on that impl 😂

internal/appliance/appliance.go
craigfurman: 👍 

Started chatting about the stage in https://github.com/sourcegraph/sourcegraph/pull/62172#discussion_r1579471862.

internal/appliance/appliance.go
jdpleiness: It's used in `shared.go` on [line 38](https://github.com/sourcegraph/sourcegraph/pull/62172/files#diff-7904122b44df84f35eb349f9f903c862131ac80cad55b3f301b5cfda16dbefb6R38) as part of the gRPC server.

My thoughts are to also use this to keep track of state via `Stage` for the appliance itself (during the initial install, setup, etc... before we have a configmap in place) and any other API/frontend handlers we may have going forward aside from our backend gRPC endpoint.

PR #62169, opened 2024-04-24
gitserver: grpc: create FirstEverCommit RPC method and server implementation

ggilmore:
Part of #61689 

This PR builds on https://github.com/sourcegraph/sourcegraph/pull/62165, and adds the FirstEverCommit new gRPC method along with a corresponding server implementation. The client implementation will be in a future PR.

## Test plan

- Unit tests

PR #62173, opened 2024-04-24
gitserver: grpc: switch clientImplementor to use new dedicated gRPC method for FirstEverCommit

ggilmore:
Closes https://github.com/sourcegraph/sourcegraph/issues/61689

This PR switches the gitserver client implementation to use the new gRPC FirstCommitEver() method introduced in https://github.com/sourcegraph/sourcegraph/pull/62169 instead of shoving it through the `exec()` endpoint.

## Test plan

Unit tests

### DIFF
--- a/internal/insights/gitserver/first_commit.go
+++ b/internal/insights/gitserver/first_commit.go
@@ -33,11 +33,7 @@ func isFirstCommitEmptyRepoError(err error) bool {
 }
 
 func GitFirstEverCommit(ctx context.Context, gitserverClient gitserver.Client, repoName api.RepoName) (*gitdomain.Commit, error) {
-	commit, err := gitserverClient.FirstEverCommit(ctx, repoName)
-	if err != nil && isFirstCommitEmptyRepoError(err) {
-		return nil, errors.Wrap(EmptyRepoErr, err.Error())
-	}
-	return commit, err
+	return gitserverClient.FirstEverCommit(ctx, repoName)
 }
 
 func NewCachedGitFirstEverCommit() *CachedGitFirstEverCommit {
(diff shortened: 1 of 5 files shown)


eseliger (approved):

pjlast (approved):

internal/insights/gitserver/first_commit.go
```diff
@@ -33,17 +33,28 @@ func isFirstCommitEmptyRepoError(err error) bool {
...
+			return nil, errors.Wrap(EmptyRepoErr, err.Error())
+		}
+
+		return nil, errors.Wrap(err, "fetching SHA of first commit")
 	}
+
+	commit, err := gitserverClient.GetCommit(ctx, repoName, id)
+	if err != nil {
+		return nil, errors.Wrap(err, "fetching first commit information")
+	}
```
eseliger, line 48: Ah maybe I've been a bit unclear here, I meant the gRPC server could make those two calls to the gitcli package if we want the full commit returned from the API, vs. making those two calls in the gitcli package itself.

So the proto would return `message FirstEverCommitResponse { GitCommit commit = 1; }`, and this logic in here would happen in `server_grpc.go`.

Not a blocker though.

eseliger (approved):

internal/insights/gitserver/first_commit.go
ggilmore: Ah, I see - I misunderstood what you meant there...

I think that though that there is value in not returning the full commit information if we don't need it? I can see a usecase for only needing the sha (e.g. as part of a future call to diff between the original commit and current head of something). 

I'm inclined to leave things like this (although now that I think about it, we could use another strategy like field masks to still accomplish something similar while returning the whole commit). 

Thoughts?

eseliger: I think masks are a neat way to achieve that - but we shouldn't implement them until we could optimize a bunch here for a certain caller. 

Since the only caller of this is insights right now, I am somewhat a fan of making the API as useful for them as possible, and returning the commit right away to avoid a round trip, but I have no strong opinion here - you decide :)

ggilmore:
### Merge activity

* **May 1, 12:35 AM EDT**: [Graphite](https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62173) rebased this pull request as part of a merge.

PR #62175, opened 2024-04-24
Code insights: followup from #62011

camdencheek:
This is just a couple of followups from #62011 that I had asked for. This is @bahrmichael's branch that I am just pushing up to get reviewed and merged. Consider it reviewed and approved by me, feel free to just stamp.

This:
- Removes the lockfile ignoring logic that was added in #62011 
- Fixes a race condition on a shared buffer by allocating a new buffer for each file

I'm not concerned about the perf of the allocation because it is a fixed-size allocation for each file (we process the file in a streaming manner) and there is already a network request for each file.

## Test plan

Bazel tests pass after re-enabling race detection.


bahrmichael (approved):
I like the code

PR #62177, opened 2024-04-25
chore: Break dependency of internal/trace on conf

eseliger:
This PR adds an initializer function that will be called from the svcmain package through the call to tracer.Init.

The reasoning behind this is that it's very easy to accidentally use a package that uses conf internally from a service like Cody Gateway, Appliance, Migrator, or other MSP services and just because we don't have any config source for the trace ID should not let the process stall entirely. To make it absolutely clear that a dependency is safe to use from a conf perspective, this change indicates that well by dropping the dependency on it entirely and making it another packages concern to pass this type down.

Test plan:

All tests for Sourcegraph are still passing.


internal/trace/init.go
```diff
@@ -0,0 +1,12 @@
+package trace
+
+type URLRenderer func(traceID string) string
+
+var urlRenderer URLRenderer
+
+// Init initializes the trace package with a URL renderer. The passed function
+// will be used to take a traceID and turn it into an environment specific URL
+// that can be used to make debugging a little easier.
+func Init(ur URLRenderer) {
```
bobheadxi, line 10: If we want to break free from the `Init` convention, we can do a "register logger"-style setup where its expected consumers of this package can change the callback whenever they want to enable/disable behaviour

```suggestion
func RegisterURLRenderer(ur URLRenderer) {
```

internal/trace/url.go
```diff
@@ -1,48 +1,14 @@
...
-	if cachedURLTemplateStr != tracing.UrlTemplate {
-		cachedURLTemplateStr = tracing.UrlTemplate
-		cachedURLTemplate, cachedURLTemplateErr = template.New("traceURL").Parse(tracing.UrlTemplate)
-	}
-
-	if cachedURLTemplateErr != nil {
-		// We contribute a validator on tracer package init, so safe to no-op here
-		return ""
+	if urlRenderer == nil {
+		return "<url renderer not configured see trace.Init>"
```
bobheadxi, line 10: I think giving a more complete path to the undefined variable should be sufficient for someone to trace down what went wrong:

```suggestion
		return "<internal/trace.urlRenderer not configured>"
```

cmd/symbols/shared/main.go
```diff
@@ -97,7 +97,7 @@ func Main(ctx context.Context, observationCtx *observation.Context, ready servic
 	}, handleStatus)
 
 	handler = handlePanic(logger, handler)
-	handler = trace.HTTPMiddleware(logger, handler, conf.DefaultClient())
+	handler = trace.HTTPMiddleware(logger, handler)
```
bobheadxi, line 100: Nice 😁

bobheadxi:
Is there a "bazel linter" for preventing certain packages from ever importing something in CI?

eseliger:
> Is there a "bazel linter" for preventing certain packages from ever importing something in CI?

Maybe, the problem is that many packages import conf, but then maybe something you're using isn't using conf so it would prevent more than necessary :( Ideally, we'll just get rid of conf.Get and use the querier interface instead

eseliger:
@bobheadxi addressed feedback, PTAL :)

internal/trace/url.go
```diff
@@ -1,48 +1,14 @@
...
-		cachedURLTemplateStr = tracing.UrlTemplate
-		cachedURLTemplate, cachedURLTemplateErr = template.New("traceURL").Parse(tracing.UrlTemplate)
-	}
-
-	if cachedURLTemplateErr != nil {
-		// We contribute a validator on tracer package init, so safe to no-op here
-		return ""
+	if urlRenderer == nil {
+		return "<internal/trace.urlRenderer not configured>"
 	}
```
bobheadxi, line 11: do we need to claim the new `urlRendererMu` here on usage?

bobheadxi (approved):
Overall LGTM except for comment about mutex 😁 thank you for doing this!

internal/trace/url.go
eseliger: good catch! added

PR #62178, opened 2024-04-25
gitserver: Improve signal handling and maxRSS reporting

eseliger:
This PR makes a change to how we spawn git processes, it puts them in separate process groups. This will allow us to track maxRSS reliably, and also lets us handle context cancellations a bit nicer; we can now 
- trigger a SIGTERM first for a more graceful shutdown
- later send a SIGKILL if that doesn't stop the process
- and make sure subprocesses are also always properly killed, by killing the group and not just the parent process

We've done a bit of back and forth here on Slack before on whether this has any other unintended side-effects, but the parent ID is untouched so the worst that could happen is that someone kills the process group where gitserver runs and the git processes aren't properly shut down by that. The pipes will break though so that should propagate quite okay.
In a containerized environment, this will not matter as all children of the init process will be reaped, and they remain children of that.

## Test plan

Integration and E2E tests pass.

### DIFF
--- a/cmd/gitserver/internal/git/gitcli/exec_test.go
+++ b/cmd/gitserver/internal/git/gitcli/exec_test.go
@@ -34,7 +34,7 @@ func TestIsAllowedGitCmd(t *testing.T) {
 		{"commit", "--file=relative/path"},
 	}
 
-	logger := logtest.Scoped(t)
+	logger := logtest.NoOp(t)
 	for _, args := range isAllowed {
 		t.Run("", func(t *testing.T) {
 			if !IsAllowedGitCmd(logger, args, "/fake/path") {
@@ -76,7 +76,7 @@ func TestIsAllowedDiffGitCmd(t *testing.T) {
 		{args: []string{"diff", "a1c0f7d19f6e9eb76facc67c1c22c07bb2ad39c4...c70f79c26526ba74f38ecff2e1e686fc3e2bdcdd"}, pass: true},
 	}
 
-	logger := logtest.Scoped(t)
+	logger := logtest.NoOp(t)
 	for _, cmd := range allowed {
 		t.Run(fmt.Sprintf("%s returns %t", strings.Join(cmd.args, " "), cmd.pass), func(t *testing.T) {
 			assert.Equal(t, cmd.pass, IsAllowedGitCmd(logger, cmd.args, "/foo/baz"))
--- a/cmd/gitserver/internal/git/gitcli/refs_test.go
+++ b/cmd/gitserver/internal/git/gitcli/refs_test.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"runtime"
 	"testing"
+	"time"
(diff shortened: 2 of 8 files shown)


eseliger:
* **#62178** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62178?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

cmd/gitserver/internal/git/gitcli/refs_test.go
```diff
@@ -399,6 +400,8 @@ func TestGitCLIBackend_ListRefs_GoroutineLeak(t *testing.T) {
 	// Don't complete reading all the output, instead, bail and close the reader.
 	require.NoError(t, it.Close())
 
+	time.Sleep(time.Millisecond)
```
eseliger, line 403: This makes the test a little more resilient to flakes because it gives the associated goroutines one more chance to get scheduled.

cmd/gitserver/internal/git/gitcli/refs_test.go
```diff
@@ -345,7 +346,7 @@ func TestGitCLIBackend_ListRefs(t *testing.T) {
 		require.Error(t, err)
 		require.True(t, errors.Is(err, context.Canceled), "unexpected error: %v", err)
 
-		require.NoError(t, it.Close())
+		require.True(t, errors.Is(it.Close(), context.Canceled), "unexpected error: %v", err)
```
eseliger, line 349: I think I like this behavior better, what do people think?

That makes sure that when you don't consume all the read data you still get the command error.

cmd/gitserver/internal/git/gitcli/exec_test.go
```diff
@@ -34,7 +34,7 @@ func TestIsAllowedGitCmd(t *testing.T) {
 		{"commit", "--file=relative/path"},
 	}
 
-	logger := logtest.Scoped(t)
+	logger := logtest.NoOp(t)
```
eseliger, line 37: Reduces noise in tests

cmd/gitserver/internal/git/gitcli/refs_test.go
ggilmore: Why not something more than enough, like 500 ms?

```suggestion
time.Sleep(500 * time.Millisecond)
```

ggilmore (approved):
Thanks for spearheading this! I think it would be nice to eventually hook this up to traces and request logs as well

cmd/gitserver/internal/git/gitcli/refs_test.go
eseliger: It would make the tests much slower to run, they're pretty fast right now and 500ms is quite a lot :/ If this flakes, we can go to 10ms or so first, and see? But so far I haven't been able to get it to flake locally.

PR #62180, opened 2024-04-25
Load and pass API token to SMEGA

rafax:
PR 2/3 for https://github.com/sourcegraph/embeddings-api/issues/51 - load API token from GSM and pass it to SMEGA API. Must be merged after infrastructure changes

Other PRs:
1/3 -> https://github.com/sourcegraph/infrastructure/pull/6000
3/3 - https://github.com/sourcegraph/embeddings-api/pull/53

## Test plan

- tested locally
- e2e tests


cmd/cody-gateway/internal/httpapi/embeddings/sourcegraph.go
```diff
@@ -15,19 +15,22 @@ import (
 )
 
 const modelDimensions = 768
+const inferenceSecretHeaderName = "x-smega-inference-auth"
```
janhartman, line 18: nit: maybe format constants by grouping them into one `const ( )` block?

janhartman (approved):

shivasurya (approved):

cmd/cody-gateway/internal/httpapi/embeddings/sourcegraph.go
rafax: done

cmd/cody-gateway/internal/httpapi/embeddings/sourcegraph.go
```diff
@@ -56,6 +61,7 @@ func (s sourcegraphClient) GenerateEmbeddings(ctx context.Context, request codyg
 	if err != nil {
 		return nil, 0, err
 	}
+	req.Header.Set(inferenceSecretHeaderName, s.apiToken)
```
chrsmith, line 64: A few lines earlier we call `s.json.Marshal(tritonRequest)`. But is the "new" name of the embeddings API SMEGA? Or should we rename this to something more generic like `generateEmbeddingsRequest`?

cmd/cody-gateway/internal/httpapi/embeddings/sourcegraph.go
```diff
@@ -14,20 +14,25 @@ import (
 	"github.com/sourcegraph/sourcegraph/internal/httpcli"
 )
 
-const modelDimensions = 768
+const (
+	modelDimensions           = 768
+	inferenceSecretHeaderName = "x-smega-inference-auth"
```
chrsmith, line 19: JFYI, it is no longer required to prefix custom headers with `X-`.

> Custom proprietary headers have historically been used with an X- prefix, but this convention was deprecated in June 2012 because of the inconveniences it caused when nonstandard fields became standard in [RFC 6648](https://datatracker.ietf.org/doc/html/rfc6648)
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

chrsmith (approved):
This all seems legit. Though I'll admit I don't know anything about SMEGA or what you are trying to do here 😬 . But it looks like we may want to do some general code cleanup / refactoring to the names, since it looks like we use "triton" still in a few places?

(Not important enough to block your PR, but IMHO that's something we'd definitely want to do so that you don't need to know the oral history of the codebase in order to safely make changes.)

cmd/cody-gateway/internal/httpapi/embeddings/sourcegraph.go
chrsmith: Feel free to keep this as-is if it is your personal preference.

IMHO, as a personal preference, I'd go with the "up-to-date, but inconsistent". Since for this particular API call, it's consistent for all of its HTTP headers :D However, we may want to rename it to `Sourcegraph-SMEGA-infrence-auth`, to be consistent in another dimension. (Since this is a Sourcegraph-specific thing? Or is SMEGA a 3rd party app/tool?

cmd/cody-gateway/internal/httpapi/embeddings/sourcegraph.go
rafax: Renamed now

cmd/cody-gateway/internal/httpapi/embeddings/sourcegraph.go
rafax: Fair, it's not like we're passing other X- headers here, so might as well go with something modern ;)

PR #62181, opened 2024-04-25
appliance: helm-compare utility

craigfurman:
Included README in the new package tells the story.

## Test plan

This is a development utility that helps interpret test fixtures, so it's in aid of testing, but is admittedly not itself tested.


internal/appliance/dev/compare-helm/compare-helm.go
```diff
@@ -0,0 +1,156 @@
...
+		goldenBytes, err := yaml.Marshal(unmatchedGolden)
+		must(err)
+		_, err = sortedGoldenFile.Write(goldenBytes)
+		must(err)
+	}
+
+	must(sortedHelmResourceFile.Close())
+	must(sortedGoldenFile.Close())
+
+	diffCmd := exec.Command("diff", helmTemplateOutputPath, sortedGoldenPath)
```
craigfurman, line 100: Tested with GNU diff. There's probably a go diff package I could use, but this was very convenient.

jdpleiness:
Ooo this is ✨nice✨

My only thought would be to use something like https://github.com/fatih/color so we can drop the dependency on colordiff, but that probably makes this a bit harder. 

Adding a screen shot for easy reference:
CleanShot 2024-04-25 at 09 48 22@2x

craigfurman:
Thanks @jdpleiness! Re: colors, initially, I wanted to rely on GNU diff's flag `--color=auto` to use color output when stdout is a tty. When I didn't see colours in my terminal, I assumed there was some shenanigans going on with `go run`, like putting a fifo in front of stdout. This isn't actually true, our terminal pty stdout is forwarded to our executable, which forwards it to diff. What was really happening, is that `--color=never` is the default, and I have diff aliased to `diff --color=auto` somewhere in my shell config 😂 

I've pushed a commit that passes --color=auto, and removes mention of colordiff. The trade-off is that this now explicitly depends on GNU diff (for us mac users, from diffutils in homebrew). Unlike coreutils, there's no PATH manipulation you have to do - brew's diff will automatically shadow BSD diff under all standard homebrew configs.

If we bring in a color library, I'd rather also bring in a diff library to replace exec'ing diff - but since this is a dev tool for one team I think that might be overengineering and I'm inclined to stop here. WDYT?

craigfurman:
P.S. I'll actually quickly make a `--no-color` flag, in case Mac diff die-hards still want to use this, and can always pipe into colordiff

jdpleiness:
>but since this is a dev tool for one team I think that might be overengineering and I'm inclined to stop here. 

Haha nah I completely agree 👍

PR #62184, opened 2024-04-25
Svelte: fix resubmit query

camdencheek:
Fixes https://github.com/sourcegraph/sourcegraph/issues/61636

## Test plan

Manually tested that hitting enter in the query bar re-submits the query. Verified that the only request made is a search request and `EditTemporarySettings`, which seems to be run after ever search request for reasons unbeknownst to me.

### DIFF
--- a/client/web-sveltekit/src/routes/search/+page.ts
+++ b/client/web-sveltekit/src/routes/search/+page.ts
@@ -84,7 +84,6 @@ const streamManager = browser ? new CachingStreamManager() : new NonCachingStrea
 
 export const load: PageLoad = ({ url, depends }) => {
     const hasQuery = url.searchParams.has('q')
-    const caseSensitiveURL = url.searchParams.get('case') === 'yes'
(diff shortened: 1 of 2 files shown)


client/web-sveltekit/src/routes/search/+page.ts
```diff
@@ -97,9 +96,7 @@ export const load: PageLoad = ({ url, depends }) => {
             caseSensitive,
             filters: queryFilters,
         } = parsedQuery
-        // Necessary for allowing to submit the same query again
-        // FIXME: This is not correct
-        depends(`query:${query}--${caseSensitiveURL}`)
+        depends(`search:${url}`)
```
camdencheek, line 99: If we use the passed-in `fetch` for search streams, I don't think we need this since we could just invalidate the stream URL, which is automatically registered as a dependency. However, we don't, so for now just depending on the search URL string to contain all the necessary data to uniquely identify a search seems good enough to me.

fkling (approved):
We store the search history in temporary settings ;)

PR #62187, opened 2024-04-25
svelte: First iteration of history panel/diff view redesign

fkling:
This commit implements some parts of the history panel and diff view redesign. In particular it

- Updates the file header to provide an option to close the diff view and to show meta information about the diff.
- Updates the history panel design
- Adds icons to bottom panel tabs
- Updates the hunk view

| Before | After |
|--------|--------|
|  |  |

TODO, likely in a follow up PR (I'll create an issue before merging):

- [x] Add additional actions for viewing the whole file and browsing the code at a specific commit.
- [ ] Use custom comparison icon.
- [ ] Add "commit icon" to beginning of each row.

## Test plan

Manual testing.

### DIFF
--- a/client/web-sveltekit/src/lib/TabPanel.svelte
+++ b/client/web-sveltekit/src/lib/TabPanel.svelte
@@ -5,6 +5,10 @@
     import { type TabsContext, KEY } from './Tabs.svelte'
 
     export let title: string
+    /**
+     * SVG path for the icon to display in the tab header.
+     */
+    export let icon: string | undefined = undefined
 
     const context = getContext<TabsContext>(KEY)
     const id = uuid.v4()
@@ -13,6 +17,7 @@
         context.register({
             id: tabId,
             title,
+            icon,
         })
     )
     $: selectedId = context.selectedTabID
--- a/client/web-sveltekit/src/lib/Tabs.svelte
+++ b/client/web-sveltekit/src/lib/Tabs.svelte
@@ -18,6 +18,7 @@
     import { createEventDispatcher, setContext } from 'svelte'
     import { derived, writable, type Readable, type Writable, type Unsubscriber } from 'svelte/store'
     import * as uuid from 'uuid'
+
     import Icon from './Icon.svelte'
 
     /**
@@ -54,7 +55,7 @@
     })
 
     function selectTab(event: MouseEvent) {
-        const index = (event.target as HTMLElement).id.match(/\d+$/)?.[0]
+        const index = (event.target as HTMLElement).closest('[role="tab"]')?.id.match(/\d+$/)?.[0]
         if (index) {
             $selectedTab = $selectedTab === +index && toggable ? null : +index
             dispatch('select', $selectedTab)
@@ -72,9 +73,10 @@
                 tabindex={$selectedTab === index ? 0 : -1}
                 role="tab"
                 on:click={selectTab}
-                data-tab-title={tab.title}
                 data-tab
-                >{#if tab.icon}<Icon svgPath={tab.icon} aria-hidden inline /> {/if}{tab.title}</button
+                >{#if tab.icon}<Icon svgPath={tab.icon} aria-hidden inline /> {/if}<span data-tab-title={tab.title}
+                    >{tab.title}</span
+                ></button
             >
         {/each}
     </div>
@@ -89,41 +91,46 @@
     }
 
     .tabs-header {
+        --icon-fill-color: var(--header-icon-color);
+
         display: flex;
+        align-items: stretch;
         justify-content: var(--align-tabs, center);
         gap: var(--tabs-gap, 0);
     }
 
-    button {
+    [role='tab'] {
+        all: unset;
+
         cursor: pointer;
-        border: none;
-        background: none;
-        letter-spacing: normal;
-        margin: 0;
-        min-height: 2rem;
-        padding: 0.5rem;
-        border-radius: 0.125rem;
-        color: var(--body-color);
-        text-transform: none;
+        color: var(--text-body);
+        padding: 0.5rem 0.75rem;
         white-space: nowrap;
+
         border-bottom: 2px solid transparent;
 
-        &[aria-selected='true'],
-        &:hover {
-            color: var(--body-color);
-            background-color: var(--color-bg-2);
-        }
-
         &[aria-selected='true'] {
-            font-weight: 500;
+            border-color: var(--brand-secondary);
+            font-weight: 700;
+            color: var(--text-title);
         }
 
-        &::before {
-            content: attr(data-tab-title);
-            display: block;
-            font-weight: 500;
-            height: 0;
-            visibility: hidden;
+        &:hover {
+            border-color: var(--border-color-2);
+        }
+
+        span {
+            display: inline-block;
+
+            // Hidden rendering of the bold tab title to prevent
+            // shifting when the tab is selected.
+            &::before {
+                content: attr(data-tab-title);
+                display: block;
+                font-weight: 700;
+                height: 0;
+                visibility: hidden;
+            }
         }
     }
 </style>
--- a/client/web-sveltekit/src/lib/Tabs.svelte
+++ b/client/web-sveltekit/src/lib/Tabs.svelte
@@ -2,6 +2,7 @@
     export interface Tab {
         id: string
         title: string
+        icon?: string
     }
 
     export interface TabsContext {
@@ -17,6 +18,7 @@
     import { createEventDispatcher, setContext } from 'svelte'
     import { derived, writable, type Readable, type Writable, type Unsubscriber } from 'svelte/store'
     import * as uuid from 'uuid'
+    import Icon from './Icon.svelte'
 
     /**
      * The index of the tab that should be selected by default.
@@ -71,7 +73,8 @@
                 role="tab"
                 on:click={selectTab}
                 data-tab-title={tab.title}
-                data-tab>{tab.title}</button
+                data-tab
+                >{#if tab.icon}<Icon svgPath={tab.icon} aria-hidden inline /> {/if}{tab.title}</button
             >
         {/each}
     </div>
@@ -95,7 +98,6 @@
         cursor: pointer;
         border: none;
         background: none;
-        align-items: center;
         letter-spacing: normal;
         margin: 0;
         min-height: 2rem;
@@ -103,9 +105,6 @@
         border-radius: 0.125rem;
         color: var(--body-color);
         text-transform: none;
-        display: inline-flex;
-        flex-direction: column;
-        justify-content: center;
         white-space: nowrap;
         border-bottom: 2px solid transparent;
 
--- a/client/web-sveltekit/src/lib/Tabs.svelte
+++ b/client/web-sveltekit/src/lib/Tabs.svelte
@@ -109,14 +109,18 @@
 
         border-bottom: 2px solid transparent;
 
+        &:hover {
+            border-color: var(--border-color-2);
+        }
+
         &[aria-selected='true'] {
             border-color: var(--brand-secondary);
             font-weight: 700;
             color: var(--text-title);
-        }
 
-        &:hover {
-            border-color: var(--border-color-2);
+            &:hover {
+                border-color: var(--brand-secondary);
+            }
         }
 
         span {
--- a/client/web-sveltekit/src/lib/repo/HistoryPanel.svelte
+++ b/client/web-sveltekit/src/lib/repo/HistoryPanel.svelte
@@ -9,13 +9,14 @@
     import { tick } from 'svelte'
 
     import { page } from '$app/stores'
-    import { scrollIntoView } from '$lib/actions'
+    import { scrollIntoViewOnMount } from '$lib/dom'
     import Avatar from '$lib/Avatar.svelte'
     import Icon from '$lib/Icon.svelte'
     import LoadingSpinner from '$lib/LoadingSpinner.svelte'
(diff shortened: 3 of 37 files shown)


client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.ts
```diff
@@ -96,66 +93,63 @@ export const load: LayoutLoad = async ({ parent, params, url }) => {
...
-                          if (!nextResult.data?.repository?.commit?.blob?.lsif) {
-                              return nextResult
-                          }
-
-                          const previousNodes =
-                              previousResult.data?.repository?.commit?.blob?.lsif?.references?.nodes ?? []
-                          const nextNodes = nextResult.data?.repository?.commit?.blob?.lsif?.references?.nodes ?? []
+        // We are not extracting the selected position from the URL because that creates a dependency
+        // on the full URL, which causes this loader to be re-executed on every URL change.
+        getReferenceStore: (lineOrPosition: LineOrPositionOrRange & { line: number }) =>
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.ts:98]: This is actually an important change. As the comment explains, I had to change this from fetching the data when `load` is called to fetching it on demand. Without this change there is a dependency on the whole URL so if any part of the URL changes, the loader is called again.

Maybe this shouldn't be an issue but (a) this interferes with the history panel and (b) it seems unnecessary/wasteful.

Regarding (a): This causes the reference panel to be recreated, which causes the selected history entry to be scrolled into the center.

So in general there are two ways to solve this:
- Make components work better with unchanged loader data.
- Only run the data loader when necessary.

Either way I think we'll have to reevaluate how to handle this kind of "optional" data in loaders, specifically in layout data loaders.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/ReferencePanelCodeExcerpt.svelte
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/ReferencePanelCodeExcerpt.svelte:1]: Only formatting changes.

client/web-sveltekit/src/lib/CodeExcerpt.stories.svelte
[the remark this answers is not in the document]
camdencheek: We should really get all our prettier settings in sync 🤔

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.ts
Comment (camdencheek) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.ts:98]: Could we use [untrack](https://kit.svelte.dev/docs/load#rerunning-load-functions-untracking-dependencies) here?

camdencheek:
A few notes for followup, ignore at will to get this shipped.

It would be nice if we reserved space for the X so the layout didn't shift on select.

https://github.com/sourcegraph/sourcegraph/assets/12631702/da012fda-28ac-498c-9c67-40ddb02974cb

This icon looks like the [google translate icon](https://translate.google.com/), which makes me think it's clickable and will switch the two, but it does not:
screenshot-2024-04-25_16-57-37@2x

The selected background color is the same as the new default avatar background color, making it disappear:

screenshot-2024-04-25_16-59-03@2x

Ideally, even if we focus the changed hunks, the rest of the file is still viewable in diff view.

screenshot-2024-04-25_17-02-03@2x

Why the very subtle rounded corners on the tabs? Makes it look a bit like we have tiny artefacts in the render.

screenshot-2024-04-25_17-05-12@2x

camdencheek (approved):
Looks great! Thanks for cranking it out 🙂

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.ts
Comment (fkling) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.ts:98]: Ah, I forgot about this. I'll try that.

client/web-sveltekit/src/lib/CodeExcerpt.stories.svelte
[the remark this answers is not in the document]
fkling: In this case I forgot to run prettier and the CI doesn't enforce it for svelte files (yet).

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.ts
Comment (fkling) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.ts:98]: Thinking again about this, `untrack` wouldn't help either. What we really want here is a dependency on the position query parameter in the URL, e.g. `L1`. However, these parameter names are not statically known so we have to inspect *all* query parameters to find the one we need. And that's what's problematic because it means we have a dependency on all other query params even if they are not relevant to this layout.

Doing something like `?position=L1` instead would solve this problem.

fkling:
Moved quite a bit of code around and updated the tab panel design. This is how it looks know:

client/web-sveltekit/src/lib/TabPanel.svelte
```diff
@@ -5,6 +5,10 @@
     import { type TabsContext, KEY } from './Tabs.svelte'
 
     export let title: string
+    /**
+     * SVG path for the icon to display in the tab header.
+     */
+    export let icon: string | undefined = undefined
```
vovakulikov, line 11: I probably need a slot for this because in revision picker tabs, I will need a suffix-like icon for KBD shortcuts. 
Not a blocker, and I will probably add this later myself, the question I have that slots are not simple render functions and can't be passed through context register, curios about your thought how you would do it.

client/web-sveltekit/src/lib/repo/HistoryPanel.svelte
```diff
@@ -59,40 +56,50 @@
...
 </script>
 
 <Scroller bind:this={scroller} margin={200} on:more={loadMore}>
     {#if history}
         <table>
             {#each history.nodes as commit (commit.id)}
-                {@const selected = commit.abbreviatedOID === selectedRev}
-                <tr class:selected use:scrollIntoView={selected}>
+                {@const selected = commit.abbreviatedOID === selectedRev || commit.oid === selectedRev}
+                <tr class:selected use:scrollIntoViewOnMount={selected}>
```
vovakulikov, line 73: Kind of surprised that Svelte doesn't support conditional directives out of the box.

client/web-sveltekit/src/lib/repo/HistoryPanel.svelte
```diff
@@ -59,40 +56,50 @@
...
                     <td>
-                        <Avatar avatar={commit.author.person} />&nbsp;
-                        {commit.author.person.displayName}
+                        <Badge variant="link"><a href={commit.canonicalURL}>{commit.abbreviatedOID}</a></Badge>
                     </td>
                     <td class="subject">
                         {#if enableInlineDiffs}
-                            <a href="?rev={commit.abbreviatedOID}">{commit.subject}</a>
+                            <a href="?rev={commit.oid}&diff=1">{commit.subject}</a>
```
vovakulikov, line 79: Not a block either, but have you thought about having helpers for any things that add anything like query parameters to the URL? It would simplify later refactoring and understanding on exact set of query parameters that we use.

client/web-sveltekit/src/lib/utils/assert.ts
```diff
@@ -0,0 +1,5 @@
+export function assertNonNullable<T>(value: T | null | undefined, message: string): asserts value is T {
```
vovakulikov, line 1: Minor: I remember that resize panels also have some custom assert utils, probably should combine them

client/web-sveltekit/src/lib/Tabs.svelte
```diff
@@ -86,45 +91,46 @@
...
+        &[aria-selected='true'] {
+            border-color: var(--brand-secondary);
+            font-weight: 700;
+            color: var(--text-title);
         }
 
-        &[aria-selected='true'] {
-            font-weight: 500;
+        &:hover {
+            border-color: var(--border-color-2);
```
vovakulikov, line 119: Hover on selected items shouldn't cause changing border-bottom color.

client/web-sveltekit/src/lib/Tabs.svelte
```diff
@@ -86,45 +91,46 @@
...
+        --icon-fill-color: var(--header-icon-color);
+
         display: flex;
+        align-items: stretch;
         justify-content: var(--align-tabs, center);
         gap: var(--tabs-gap, 0);
     }
 
-    button {
+    [role='tab'] {
```
vovakulikov, line 102: Since we're changing styles around the tabs, I would check the revision picker tabs UI; I remember that I had to do a few additional paddings to the selected tab tabs because of how we present selected tab. Since the selected tab no longer has a grey background but a border probably additional vertical spacing in the revisions tab is unnecessary

taiyab:
@camdencheek 

I can sort most of your proposed visual changes once this is merged.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte
```diff
@@ -111,15 +117,25 @@
     $: sidebarWidth = `max(320px, ${$sidebarSize * 100}%)`
 
     // The observable query to fetch references (due to infinite scrolling)
-    $: referenceQuery = data.references
+    $: sgURL = SourcegraphURL.from($page.url)
+    $: selectedLine = sgURL.lineRange
+    $: referenceQuery =
```
Comment (vovakulikov) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte:122]: Have you tried to move all those references specific to the ReferencesPanel component? 
It looks like this component gets bigger and bigger and I don't think we should have everything 
on the top level here, or is there any reason for this?

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.ts
Comment (vovakulikov) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.ts:98]: I don't believe I fully understand how usage tracking works in loaders but can we pass a stream with URL updates instead of using url from loader params and sub to this stream and track only params that we need? 
This probably would solve the problem of triggering loader and triggering infinityQuery on any other param change? 

Not a blocker, just because I think loading on demand actually makes more sense to me when it comes to something like reference panel.

client/web-sveltekit/src/lib/repo/HistoryPanel.svelte
```diff
@@ -110,7 +117,7 @@
     }
 
     td {
```
vovakulikov, line 119: I would check author column, it looks like alignment between avatar and text is a little bit off

client/web-sveltekit/src/lib/repo/HistoryPanel.svelte
```diff
@@ -59,40 +56,50 @@
...
-                {@const selected = commit.abbreviatedOID === selectedRev}
-                <tr class:selected use:scrollIntoView={selected}>
+                {@const selected = commit.abbreviatedOID === selectedRev || commit.oid === selectedRev}
+                <tr class:selected use:scrollIntoViewOnMount={selected}>
                     <td>
-                        <Avatar avatar={commit.author.person} />&nbsp;
-                        {commit.author.person.displayName}
+                        <Badge variant="link"><a href={commit.canonicalURL}>{commit.abbreviatedOID}</a></Badge>
                     </td>
                     <td class="subject">
```
vovakulikov, line 77: UX observation, I didn't notice in first two times that link in the history item opens diff, probably it makes sense (or at least consider) to move this to the separate action as well. But not sure that 3 action in each row will make thing simple here 🤔

vovakulikov (commented):
Left some minor comments, thanks, great improvement

vovakulikov:
@taiyab Since we're talking about UI here, I don't know why, but our old (react version) tab styles look a bit off in svelte UI. Maybe because we no longer use our bright colours that often and now purple tabs are very highlighted on this background. I liked our raw version a bit more. 

Just my opinionated feedback, feel free to ignore.

client/web-sveltekit/src/lib/repo/HistoryPanel.svelte
fkling: Not sure what you mean :D

client/web-sveltekit/src/lib/repo/HistoryPanel.svelte
fkling: Yes, I think everything will be rolled into the `SourcegraphURL` class eventually. But at the same time I like that you can immediately see what's going on here.

client/web-sveltekit/src/lib/Tabs.svelte
fkling: Yeah I checked it, it seems fine to me.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte
Comment (fkling) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte:122]: I wanted to keep everything related to syncing the "view" and "data" state out of the component. I don't think the reference panel itself should know much about this. But I agree that the current situation isn't ideal either.

taiyab:
> @taiyab Since we're talking about UI here, I don't know why, but our old (react version) tab styles look a bit off in svelte UI. Maybe because we no longer use our bright colours that often and now purple tabs are very highlighted on this background. I liked our raw version a bit more.
> 
> Just my opinionated feedback, feel free to ignore.

I wholeheartedly agree with this. Once we're done with a one-over on the file view, I'll do a full set of new tab designs as a proposal for the new version of the design system and then replace them all throughout here.

PR #62189, opened 2024-04-25
sg/update: allow users to specify a release with '-release'

bobheadxi:
MSP pins `sg` version per environment category to control rollout of changes: https://github.com/sourcegraph/sourcegraph/pull/62134

This allows users to update to a specified release (instead of always `latest`) with a new `-release` flag on `sg update`:

```
sg update -release ${release_name}
```

Example usage: https://github.com/sourcegraph/sourcegraph/pull/62134#discussion_r1580114111, details on how `sg msp` version pinning works: https://sourcegraph.notion.site/Deploying-new-versions-of-MSP-1808e7e45bd54f419dd93af542d99238#58dabe4992754ca18ed39bc212ccbbba

## Test plan

Install specified release:

```sh
$ go build -o ./sg ./dev/sg && ./sg install -f -p=false
$ sg version
unknown
$ sg update -r 2024-04-24-16-44-3623ecb2               
⚠️ Installing user-specified release "2024-04-24-16-44-3623ecb2" - 'sg' auto-updates might update your 'sg' installation anyway, set 'SG_SKIP_AUTO_UPDATE=false' to disable auto-updates.
✅ sg has been updated to "2024-04-24-16-44-3623ecb2"!
To see what's new, run 'sg version changelog'.
$ sg version
3623ecb2cfcec81390e37a49540dabb6f67aaa99
```

Default behaviour (latest):

```sh
$ go build -o ./sg ./dev/sg && ./sg install -f -p=false
$ sg version
unknown
$ sg update 
✅ sg has been updated to "latest"!
To see what's new, run 'sg version changelog'.
$ sg version
2024-04-25-15-45-d52b6411
```

### DIFF
--- a/dev/sg/sg_update.go
+++ b/dev/sg/sg_update.go
@@ -23,20 +23,49 @@ import (
 	"github.com/sourcegraph/sourcegraph/lib/output"
 )
 
+const sgReleases = "https://github.com/sourcegraph/sg/releases"
+
 var updateCommand = &cli.Command{
 	Name:  "update",
 	Usage: "Update local sg installation",
-	Description: `Update local sg installation with the latest changes. To see what's new, run:
+	Description: fmt.Sprintf(`Update local sg installation with the latest changes. To see what's new, run:
 
-    sg version changelog -next`,
+    sg version changelog -next
+
+A custom release from %s can be installed with the '-release' flag.`, sgReleases),
 	Category: category.Util,
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:    "release",
+			Aliases: []string{"r"},
+			Usage:   fmt.Sprintf("The name of the release in %s to update to", sgReleases),
+			Value:   "latest",
+		},
+	},
 	Action: func(cmd *cli.Context) error {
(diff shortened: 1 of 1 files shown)


unknwon (approved):
Assuming it works. What about switching between test and prod, for example, would two `sg` version overwrites each other with `sg update --release`?

bobheadxi:
> Assuming it works. What about switching between test and prod, for example, would two sg version overwrites each other with sg update --release?

Yes 😎 This update happens in-place

unknwon:
> > Assuming it works. What about switching between test and prod, for example, would two sg version overwrites each other with sg update --release?
> 
> Yes 😎 This update happens in-place

Wouldn't it be a pain that when test and prod releases differs, every time I want to generate test and prod, I need to `sg update ...`? 😂

bobheadxi:
> Wouldn't it be a pain that when test and prod releases differs, every time I want to generate test and prod, I need to sg update ...? 😂

The lag time should be a fairly small window, as we shouldn't let `external` fall too far behind. In https://github.com/sourcegraph/sourcegraph/pull/62134, the warning also says you only need to do this if there is a diff in the output - so this is mostly a "I really need this now, here's a one-liner to get it"

jhchabran (approved):

Issue #61635, opened 2024-04-05
Add telemetry/logging


fkling:
It would be nice if we could get some basic logging working before make this available to customers, even if it's opt-in, but it's a stretch goal.

PR #62190, opened 2024-04-25
Svelte: Add basic telemetry to svelte implementation

vovakulikov:
Closes https://github.com/sourcegraph/sourcegraph/issues/61635

## Context
This PR moves the telemetry events out of the web package to the shared package. This allows using this telemetry service in the `web-sveltekit` package. Also, the wrapper over the original telemetry adds a special `isSveltePrototype: true` parameter to make a difference later between the original event and the svelte events. 

Second thing, this PR adds only a subset of the original event; if you see that some event is missing, please leave a note. Currently, this adds these events

```
    ViewHomePage = 'Home',
    ViewSearchResultsPage = 'SearchResults',
    ViewRepositoryPage = 'Repository',
    ViewBlobPage = 'Blob',

    SearchSubmit = 'SearchSubmitted',
    SearchResultClick = 'SearchResultClicked',
    ShowHistoryPanel = 'ShowHistoryPanel',
    HideHistoryPanel = 'HideHistoryPanel',
    CodeCopied = 'CodeCopied',
    GoToCodeHost = 'GoToCodeHostClicked',
    GitBlameEnabled = 'GitBlameEnabled',
    SelectSearchFilter = 'SearchFiltersSelectFilter',
```

## For reviewers 
Diff got a lot of files due to the telemetry service moving. Check the commit history for easier review. 

## Next steps
Note that this PR uses old telemetry V1 due to needs around migration amplitude dashboards to new v2 telemetry events. I have a follow up here and I aim to bring new telemetry in June release https://github.com/sourcegraph/sourcegraph/issues/62191

## Test plan
- Manually check that page view event and action-based event send actual telemetry events


client/web-sveltekit/src/lib/telemetry.ts
```diff
@@ -0,0 +1,43 @@
...
+    public override log(eventName: string, eventProperties?: any, publicArgument?: any): void {
+        super.log(
+            eventName,
+            { ...eventProperties, isSveltePrototype: true },
+            { ...publicArgument, isSveltePrototype: true }
+        )
+    }
+}
+
+export const SVELTE_LOGGER = new SvelteTelemetry()
```
camdencheek, line 17: Should this be part of [the context](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@32c8ff7436543e3e158d81ba5cddd33dfaf24acf/-/blob/client/web-sveltekit/src/lib/stores.ts?L14-19) rather than a global?

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/ReferencePanelCodeExcerpt.svelte
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/ReferencePanelCodeExcerpt.svelte:1]: We should really get our formatters all in sync 😄

camdencheek (approved):

client/web-sveltekit/src/lib/telemetry.ts
vovakulikov: I was thinking about this but then I couldn't come up with reasons why it shouldn't be a global singleton. There might be an argument that this could be better for mocking in tests but I can say that there are no big differences in `vi.mock` vs context mocking, curious about what @fkling thinks

vovakulikov:
To make sure we have time to test it on staging and with amplitude I'm going ahead and merging this, will be happy to change anything later next week if something shows up

PR #62193, opened 2024-04-25
Context: always return code results first

jtibshirani:
In #62162, we pushed some Cody context logic down into the 'search job'. The
tests I added assume that the results are returned in a consistent order. This
is nice behavior and makes testing and debugging easier.

I mistakenly thought that `pool.NewWithResults` would return results in the
behavior goroutines were originally submitted. Oops! Now we use two separate
wait groups and ensure we return results in a consistent order.

## Test plan
Test are currently flaking ([example](https://buildkite.com/sourcegraph/sourcegraph/builds/270930#018f1761-0549-4b0d-91af-bc91242506a5)), this PR fixes them.

### DIFF
--- a/internal/search/codycontext/job.go
+++ b/internal/search/codycontext/job.go
@@ -119,15 +119,19 @@ func (j *searchJob) Run(ctx context.Context, clients job.RuntimeClients, stream 
 	_, ctx, stream, finish := job.StartSpan(ctx, stream, j)
 	defer func() { finish(alert, err) }()
 
-	wg := pool.NewWithResults[response]()
-	wg.Go(func() response {
-		return j.doSearch(ctx, clients, j.textJob, j.textCount)
-	})
-	wg.Go(func() response {
+	codeGroup := pool.NewWithResults[response]()
+	codeGroup.Go(func() response {
 		return j.doSearch(ctx, clients, j.codeJob, j.codeCount)
 	})
-	responses := wg.Wait()
 
+	textGroup := pool.NewWithResults[response]()
+	textGroup.Go(func() response {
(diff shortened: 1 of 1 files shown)


internal/search/codycontext/job.go
```diff
@@ -119,15 +119,19 @@ func (j *searchJob) Run(ctx context.Context, clients job.RuntimeClients, stream
 	_, ctx, stream, finish := job.StartSpan(ctx, stream, j)
 	defer func() { finish(alert, err) }()
 
-	wg := pool.NewWithResults[response]()
-	wg.Go(func() response {
-		return j.doSearch(ctx, clients, j.textJob, j.textCount)
-	})
-	wg.Go(func() response {
+	codeGroup := pool.NewWithResults[response]()
```
jtibshirani, line 122: Maybe it's unusual to use a result pool for a single result, but I found it very clean/ convenient.

camdencheek:
> I mistakenly thought that pool.NewWithResults would return results in the
behavior goroutines were originally submitted. Oops!

Common mistake! Common enough that I actually [recently pushed a change](https://github.com/sourcegraph/conc/commit/8427ccd11dde5f56ccacfbbd184f42ce865d2f0d) to `conc` to make that the default behavior. Which reminds me, we should really pull in the latest version.

internal/search/codycontext/job.go
camdencheek: Unusual, but I've also found myself doing it before. Maybe adding a `Promise` primitive or something would feel less weird 🤔

camdencheek (approved):

jtibshirani:
Thanks for the review! I'll merge to unbreak the build, then I can refactor this once we pull in the latest conc.

PR #62194, opened 2024-04-25
sg/rfc: add -s for filtering by status(es)

bobheadxi:
We do this in the printer, where we have more control over interpreting the file titles.

Context: https://sourcegraph.slack.com/archives/C015P61EXV1/p1714085557621709?thread_ts=1714070620.515829&cid=C015P61EXV1

## Test plan

```
sg rfc list -s approved -s implemented -s done
```

Note that this only covers public RFCs, you need `-private` as well to get the private RFCs.

image

There are some recent RFCs at the bottom, so the format is still used :)

### DIFF
--- a/dev/sg/internal/rfc/rfc.go
+++ b/dev/sg/internal/rfc/rfc.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"slices"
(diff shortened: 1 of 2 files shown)


ErikaRS (approved):
Thanks!

PR #62196, opened 2024-04-26
Svelte: fix file header links

camdencheek:
This fixes some issues with the breadcrumbs in the file header:
1) The links did not preserve the current revision
2) The links assumed that if the current page was a blob, all parent breadcrumbs are also blobs. Everything but the last element in the path is guaranteed to be a tree.

## Test plan

Added a playwright test that we link back to a tree page correctly. Manually tested the current revision stuff.


fkling (approved):
Oh, what a big mistake from me, thank you for fixing this!

PR #62197, opened 2024-04-26
rockskip: skip non-utf8 paths

stefanhengl:
We have seen errors on S2 for inserting non-utf-8 data to the symbols table. `path` and `name` (symbol name) are the only user controlled fields. We already check for non-utf-8 content in go-ctags, so that leaves `path`, which is not required to be utf-8 compatible.

With this change we simply skip the path, which means that symbols within that path will not exist in the index. This is somewhat in line with ctags, where we skip indexing non-utf-8 content.

I believe this is an edge case and it is acceptable to skip those files to keep the code simple.

Alternatives:
- We could store arbitrary bytes `bytea` instead of `text` in Postgres.

Test plan:
CI

### DIFF
--- a/internal/rockskip/index.go
+++ b/internal/rockskip/index.go
@@ -138,7 +138,7 @@ func (s *Service) Index(ctx context.Context, repo, givenCommit string) (err erro
 		addedPaths := []string{}
 		for _, pathStatus := range entry.PathStatuses {
 			if !utf8.ValidString(pathStatus.Path) {
-				s.logger.Error(
+				s.logger.Warn(
 					"Rockskip skipping file due to path not being utf-8 encoded",
 					log.String("repo", repo),
 					log.String("path", pathStatus.Path),
--- a/internal/rockskip/index.go
+++ b/internal/rockskip/index.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"fmt"
 	"time"
+	"unicode/utf8"
 
 	"github.com/amit7itz/goset"
 	pg "github.com/lib/pq"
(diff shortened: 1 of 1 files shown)


internal/rockskip/index.go
```diff
@@ -136,6 +137,15 @@ func (s *Service) Index(ctx context.Context, repo, givenCommit string) (err erro
 		deletedPaths := []string{}
 		addedPaths := []string{}
 		for _, pathStatus := range entry.PathStatuses {
+			if !utf8.ValidString(pathStatus.Path) {
```
mmanela, line 140: Can we document somewhere this behavior?

mmanela (approved):

keegancsmith (approved):

internal/rockskip/index.go
```diff
@@ -136,6 +137,15 @@ func (s *Service) Index(ctx context.Context, repo, givenCommit string) (err erro
...
+				s.logger.Error(
+					"Rockskip skipping file due to path not being utf-8 encoded",
+					log.String("repo", repo),
+					log.String("path", pathStatus.Path),
```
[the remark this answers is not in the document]
stefanhengl: zap replaces invalid bytes with the [unicode replace character](https://en.wikipedia.org/wiki/Specials_(Unicode_block)#Replacement_character)

internal/rockskip/index.go
stefanhengl: I will add something to the docs

stefanhengl:
> Looks good! How hard would it be to add a unit test? It'd be nice to start to chip away at getting Rockskip under test!

Technically it's not hard, but it would need a small refactor to break up the long indexing function into smaller chunks and given the limited test coverage and risk to introduce bugs, I feel it's not justified in this case. Indexing is covered by a small test case already, so we have that at least.

I think the next step should be to introduce a very detailed integration test to catch any regression. Once we have that we can do bigger changes with more confidence.

PR #62200, opened 2024-04-26
appliance: deploy repo-updater

craigfurman:
Closes https://github.com/sourcegraph/sourcegraph-operator/issues/44

## Test plan

Golden tests included.

PR #62201, opened 2024-04-26
appliance: de-dup common service features

craigfurman:
Create a "common" config struct+interface that can be embedded in service config, to get various facilities for free. The aim is to reduce the prospect for errors given how much service config the appliance will soon handle.

See changes to development.md for a full explanation of what I'm trying to do here.

## Test plan

Tests included.

### DIFF
--- a/internal/appliance/blobstore.go
+++ b/internal/appliance/blobstore.go
@@ -114,27 +114,22 @@ func buildBlobstoreDeployment(sg *Sourcegraph) appsv1.Deployment {
 		},
 	}
 
-	defaultContainer := container.NewContainer(name, sg.Spec.Blobstore)
+	defaultContainer := container.NewContainer(name, sg.Spec.Blobstore, corev1.ResourceRequirements{
+		Requests: corev1.ResourceList{
+			corev1.ResourceCPU:    resource.MustParse("1"),
+			corev1.ResourceMemory: resource.MustParse("500M"),
+		},
+		Limits: corev1.ResourceList{
+			corev1.ResourceCPU:    resource.MustParse("1"),
+			corev1.ResourceMemory: resource.MustParse("500M"),
+		},
+	})
 
 	// TODO: https://github.com/sourcegraph/sourcegraph/issues/62076
 	defaultContainer.Image = "index.docker.io/sourcegraph/blobstore:5.3.2@sha256:d625be1eefe61cc42f94498e3c588bf212c4159c8b20c519db84eae4ff715efa"
 
 	defaultContainer.Ports = containerPorts
 	defaultContainer.VolumeMounts = containerVolumeMounts
-
-	// default resources
-	if container.HasNoConfiguredResources(defaultContainer) {
-		defaultContainer.Resources = corev1.ResourceRequirements{
-			Requests: corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("1"),
-				corev1.ResourceMemory: resource.MustParse("500M"),
-			},
-			Limits: corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("1"),
-				corev1.ResourceMemory: resource.MustParse("500M"),
-			},
-		}
-	}
 
 	podVolumes := []corev1.Volume{
 		{
--- a/internal/appliance/blobstore.go
+++ b/internal/appliance/blobstore.go
@@ -15,7 +15,6 @@ import (
 	"github.com/sourcegraph/sourcegraph/internal/k8s/resource/pvc"
 	"github.com/sourcegraph/sourcegraph/internal/k8s/resource/service"
 	"github.com/sourcegraph/sourcegraph/lib/errors"
-	"github.com/sourcegraph/sourcegraph/lib/pointers"
 )
 
 func (r *Reconciler) reconcileBlobstore(ctx context.Context, sg *Sourcegraph, owner client.Object) error {
@@ -70,13 +69,13 @@ func (r *Reconciler) reconcileBlobstorePersistentVolumeClaims(ctx context.Contex
 		return err
 	}
 
-	return reconcileBlobStoreObject(ctx, r, &p, &corev1.PersistentVolumeClaim{}, sg, owner)
+	return reconcileObject(ctx, r, sg.Spec.Blobstore, &p, &corev1.PersistentVolumeClaim{}, sg, owner)
 }
 
 func buildBlobstoreService(sg *Sourcegraph) corev1.Service {
 	name := "blobstore"
 
-	s := service.NewService(name, sg.Namespace)
+	s := service.NewService(name, sg.Namespace, sg.Spec.Blobstore)
 	s.Spec.Ports = []corev1.ServicePort{
 		{
 			Name:       name,
@@ -93,7 +92,7 @@ func buildBlobstoreService(sg *Sourcegraph) corev1.Service {
 
 func (r *Reconciler) reconcileBlobstoreServices(ctx context.Context, sg *Sourcegraph, owner client.Object) error {
 	s := buildBlobstoreService(sg)
-	return reconcileBlobStoreObject(ctx, r, &s, &corev1.Service{}, sg, owner)
+	return reconcileObject(ctx, r, sg.Spec.Blobstore, &s, &corev1.Service{}, sg, owner)
 }
 
 func buildBlobstoreDeployment(sg *Sourcegraph) appsv1.Deployment {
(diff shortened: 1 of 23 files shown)


craigfurman:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62201?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#62201** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62201?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#62200** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62200?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

 Join @craigfurman and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

internal/appliance/blobstore.go
```diff
@@ -104,38 +103,6 @@ func buildBlobstoreDeployment(sg *Sourcegraph) appsv1.Deployment {
...
-				corev1.ResourceCPU:    pointers.DerefZero(reqCPU),
-				corev1.ResourceMemory: pointers.DerefZero(reqMem),
-			},
-		}
-	}
-
-	if sg.Spec.LocalDevMode {
-		containerResources = corev1.ResourceRequirements{}
-	}
-
```
craigfurman, line 138: This change is not like the rest of this PR, and I'm happy to split it out to discuss further if it's more controversial than the rest.

Basically: I don't think it's a good idea for k8s configs to set default resources, for a few reasons. First, fears about noisy neighbours from unlimited pods: Pods with containers without resources specified are given [BestEffort QoS](https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/#create-a-pod-that-gets-assigned-a-qos-class-of-besteffort). It's hard to truly mitigate noisy neighbours in kubernetes at the best of times, but in my experience, unlimited/besteffort containers are not as much of an invitation to resource contention as they look! 

In terms of scheduling/requests: these pods will be first in line for eviction if non-besteffort pods want to be scheduled.

In terms of CPU saturation: containers with CPU requests will be allocated total CPU in the ratio of these requests, and BestEffort containers will be starved.

In terms of memory saturation: this one can be a bit non-deterministic admittedly, but an unlimited-memory container that uses a high amount of the total memory available on the node is likely to be targetted first by the kernel OOM killer.

Default resources are no real replacement (IMO) for the cluster/workload-specific practice of monitoring resource usage and setting requests (and sometimes limits) appropriate over time.

That (lengthy!) first justification for this change is an argument as to why the _downside_ of not specifying resources by default is oversold IMO. The second justification is about the upside of not specifying resources by default: we avoid either harming performance by under-speccing resources, or costing users money by reserving resources we don't need.

I'm happy to talk about this further!

Since it would be unused if we do indeed remove the concept of default non-dev resources, I've also removed the LocalDevMode config option.

internal/appliance/config/config.go
```diff
@@ -0,0 +1,26 @@
+package config
+
+import corev1 "k8s.io/api/core/v1"
+
+type StandardComponent interface {
```
craigfurman, line 5: This is in a subpackage because both internal/appliance and the packages under k8s/resources need it. Arguably the k8s/resources packages should be moved under appliance, because they are only used by it. Maybe it also makes sense to move all the config in spec.go into this config subpackage too. These are 2-way doors and we don't have to get it right the first time.

internal/appliance/blobstore.go
jdpleiness: Since we've got an opportunity for somewhat of a fresh start with this deployment, I'm open to giving this a try! It seems like if it doesn't work out for some reason we can revert back, but your justification is pretty compelling 😃 

One concern that comes to mind though is how does this effect auto-scaling setups like GKE autopilot? 

These are mostly carry-overs from our Helm chart, but even there most of these are based on best guesses as to the resource allocations for each service. This goes back to a broader issue we face around having better insights into resource usage for each service we have though.

>Since it would be unused if we do indeed remove the concept of default non-dev resources, I've also removed the LocalDevMode config option.

That is a nice little bonus and reduces our complexity for local dev users as well.

internal/appliance/config/config.go
jdpleiness: > Arguably the k8s/resources packages should be moved under appliance, because they are only used by it. 

I was thinking the same thing, then read the next line of your comment 😆

internal/appliance/config/config.go
```diff
@@ -0,0 +1,26 @@
...
+	Disableable
+	GetResources() map[string]*corev1.ResourceRequirements
+	GetServiceAccountAnnotations() map[string]string
+	PrometheusPort() *int
+}
+
+type Disableable interface {
```
jdpleiness, line 12: Ahh now I see why the previous PR was adding disabled to Repo Updater 🤔 ✅

jdpleiness (approved):

internal/appliance/config/config.go
craigfurman: It's more like the other way around - this piece of code looks the way it does because of the assumptions I made in the previous PR. See https://github.com/sourcegraph/sourcegraph/pull/62200#discussion_r1581091505, can revise it as needed.

internal/appliance/blobstore.go
craigfurman: > One concern that comes to mind though is how does this effect auto-scaling setups like GKE autopilot?

Good question - the short answer is, not particularly well! This turned into a bit of a saga, and the tl;dr is that while writing it, I think I've reversed my recommendation here, or at least become less sure.

Let's start with HPAs, and build up to autopilot (since autopilot is a node-level autoscaler). HPAs (and similar things like keda, off the top of my head) need some saturation metric to base scaling decisions on. Without CPU/memory _requests_ set, HPAs can't operate (on CPU/memory metrics at least).

At the cluster level, a node autoscaler _could_ operate on the actual linux-reported CPU and memory pressure, but from a k8s PoV, assuming sensible resource requests (not always a valid assumption), it's better to scale the node pool out when requests are near-saturated, not actual CPU/memory, because no further pods can be scheduled otherwise. The GKE autopilot docs indicate that it integrates with pod requests, not just linux resource saturation metrics.

Note that I was _not_ advocating for running k8s clusters in production in which every workload is BestEffort - total anarchy! 😂 

A workflow I've seen used before is:

1. Add some new service with BestEffort QoS, leave it a week.
2. Decide on resource requests based on the mean utilisation, and limits based on the sort of spikes we saw and want to accommodate.
3. Rely on continuous monitoring to bump these resources if they become unrealistic, e.g. if the service gets more traffic and utilization scales linearly.

Certainly, developers making a generic helm chart for something like PostgreSQL could probably not reasonably specify any particular static set of k8s requests / limits - any such recommendation would have to be guided by benchmarking. In that case, if I were making such a helm chart, I'd write some version of my first message. Any default resources would be arbitrary ("how long is a piece of string?")

My initial proposal delegates responsibility fully to cluster operators. We can do clever things like only deploy HPAs if resources are configured, but there's still a risk that folks will leave BestEffort configured long-term and have a bad time (throttling, evictions, or worse - SG acting as a noisy neighbour!)

But... now that I read your message and think again... maybe we can do a bit better?

> a broader issue we face around having better insights into resource usage for each service

What if we:

1. Undertook work to understand how much memory / CPU each of our services take for given traffic levels. Hopefully benchmarking will reveal an approximately linear relationship.
2. Tried to figure out median / mean / p75 traffic to SG services on our managed cloud environments (or just median / mean resource utilization directly)
3. Used this to inform reasonable guesses for default resources for each service

We still risk over/under provisioning, but if we do that, an advantage is that we can configure autoscalers by default, which hopefully could lead to better user experience if their workloads do become saturated and have the option of scaling out rather than just being killed (BestEffort QoS). Further out, we can even use the appliance pane-of-glass features to recommend vertical container resource scaling to operators (rather than leaning solely on HPA).

This is a lot more work, but feels better than punting the entire problem to admins.

Anyway, sorry for the massive message, and the U-turn! But better to talk it out than not, perhaps.

jdpleiness: >Undertook work to understand how much memory / CPU each of our services take for given traffic levels. Hopefully benchmarking will reveal an approximately linear relationship.

So we do have something like this that we worked on a few years ago: [ref arch test](https://github.com/sourcegraph/reference-architecture-test) but I don't think we went down to the service level and was more of a load test than anything else. 

The defaults we have in the Helm chart _are_ decent baselines though in _some_ cases where teams have given us guidelines about baseline memory and CPU usage. In other cases we set a min/max based on past experience or educated guess and let administrators tweak it to their workloads from there.

>Tried to figure out median / mean / p75 traffic to SG services on our managed cloud environments (or just median / mean resource utilization directly)

This would be the better approach imo. and would give us the most accurate results.

jdpleiness: Sounds good! 

> It could be a neat appliance feature, a single toggle to template resources: null to every service

👍

craigfurman: OK, I think we can close the loop on this! 

re: 1 (HPAs): I added a line to the epic outline (https://docs.google.com/document/d/1He4TUxR81_kcakGPQgn7tumGMXhcfko7I9QzEYMhX6E/edit) since there's a little too much uncertainty to write an issue just yet. Feel free to flesh it out a bit of course, and/or chat further.

I didn't create issues for 2 or 3 either. For 2, we decided to keep it in our back pocket. For 3 (LocalDevMode), let's see what falls out of the appliance work. I just pushed a commit to here that restores default resources. An example of explicitly passing `null` resources can be seen in internal/appliance/testdata/sg/repo-updater-with-no-resources.yaml:

```yaml
  # service
  repoUpdater:
    resources:
      # named containers
      repo-updater: null
```

This might be useful for the appliance, which will need knowledge of all the services and their containers anyway, and can template `null` (unlimited / BestEffort QoS) resources everywhere with some "dev mode" toggle, if we wish. 🙏

PR #62203, opened 2024-04-26
Exclude files for getCodyContext keyword search

thenamankumar:
Exclude `.map` and `.min.js` files from keyword search behind getCodyContext. 

## Test plan

Tested manually with `getCodyContext` API for react repo.

Normal search gets results from `scheduler.profiling.min.js`
CleanShot 2024-04-29 at 11 27 41@2x

But `getCodyContext` does not:
CleanShot 2024-04-29 at 11 28 47@2x

Before:
CleanShot 2024-04-29 at 11 52 43@2x

### DIFF
--- a/cmd/frontend/internal/codycontext/context.go
+++ b/cmd/frontend/internal/codycontext/context.go
@@ -250,7 +250,6 @@ func getKeywordContextExcludeFilePathsQuery() string {
 	var excludeFilePaths = []string{
 		"\\.min\\.js$",
 		"\\.map$",
-		"dist\\",
 	}
 
 	filters := []string{}
@@ -261,7 +260,7 @@ func getKeywordContextExcludeFilePathsQuery() string {
 	return strings.Join(filters, " ")
 }
 
-var keywordContextExcludeFilePaths = getKeywordContextExcludeFilePathsQuery()
+var keywordContextExcludeFilePathsQuery = getKeywordContextExcludeFilePathsQuery()
 
 // getKeywordContext uses keyword search to find relevant bits of context for Cody
 func (c *CodyContextClient) getKeywordContext(ctx context.Context, args GetContextArgs, matcher FileMatcher) (_ []FileChunkContext, err error) {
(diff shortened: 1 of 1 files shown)


cmd/frontend/internal/codycontext/context.go
```diff
@@ -245,6 +246,22 @@ func (c *CodyContextClient) getEmbeddingsContext(ctx context.Context, args GetCo
...
+
+	filters := []string{}
+	for _, filePath := range excludeFilePaths {
+		filters = append(filters, fmt.Sprintf("-file:%v", filePath))
+	}
+
+	return strings.Join(filters, " ")
+}
+
+var keywordContextExcludeFilePathsQuery = getKeywordContextExcludeFilePathsQuery()
```
keegancsmith, line 263: minor: we don't need to compute this as `init()` time in go. It is cheap enough to just compute this map each time rather than add it to the critical path of starting your program.

keegancsmith (approved):

keegancsmith:
Can be for follow-up, but interested in @jtibshirani's opinion on the right place to do this exclusion in our stack? For example right now its really nice that we can do `patternType:codycontext` and we can compare results. A part of me thinks those exclusion rules should play a part for that query, but at the same time its also really nice to have full control on the query we execute when experimenting with it.

cmd/frontend/internal/codycontext/context.go
thenamankumar: updated

jtibshirani:
I'd prefer if we implemented all important logic (including exclude rules like this) at the Cody context search job level. That way we can test the behavior exactly using `patterntype:codycontext`. I agree it feels a little weird to push this down... ideally this would be implemented as improved ranking (like always rank vendored and generated content last), and that would make a lot of sense to do at a low level.

thenamankumar:
> I'd prefer if we implemented all important logic (including exclude rules like this) at the Cody context search job level. That way we can test the behavior exactly using `patterntype:codycontext`. I agree it feels a little weird to push this down... ideally this would be implemented as improved ranking (like always rank vendored and generated content last), and that would make a lot of sense to do at a low level.

@jtibshirani can you point me to the search job code and where this logic will be better placed? I can work on moving it.

jtibshirani:
Thanks, here's a good entrypoint: https://github.com/sourcegraph/sourcegraph/blob/676f7c2a5c111d69332f6e1d3f7b256d93610080/internal/search/codycontext/job.go#L107

PR #62206, opened 2024-04-26
Cody Web: add required "client-name" param to completions endpoint

taras-yemets:
https://github.com/sourcegraph/sourcegraph/pull/62048 introduced the minimum supported client version constraint for Cody clients in case Cody Ignore is enabled.

To ensure Cody Web client satisfies this constraint, this PR adds the "client-name" query param to the completions endpoint. The "client-version" param is not required for the web client (web client is of the same version as server) so we omit it. 

## Test plan
- CI
- Tested manually
  - Open chat panel and submit a question
  - Ensure completions endpoint request has the "client-name" query param set to "web"
  - Ensure Cody responds with no error (in particular, look for [this](https://github.com/sourcegraph/sourcegraph/blob/cd900e2de5733c16c20683dba5642e04eabbe047/internal/completions/httpapi/handler.go#L497-L503) error)

### DIFF
--- a/client/cody-shared/src/sourcegraph-api/completions/client.ts
+++ b/client/cody-shared/src/sourcegraph-api/completions/client.ts
@@ -15,7 +15,7 @@ export interface CompletionLogger {
 
 export type CompletionsClientConfig = Pick<
     ConfigurationWithAccessToken,
-    'serverEndpoint' | 'accessToken' | 'debugEnable' | 'customHeaders'
+    'serverEndpoint' | 'accessToken' | 'debugEnable' | 'customHeaders' | 'clientName' | 'clientVersion'
 >
 
 /**
@@ -35,7 +35,19 @@ export abstract class SourcegraphCompletionsClient {
     }
 
     protected get completionsEndpoint(): string {
-        return new URL('/.api/completions/stream', this.config.serverEndpoint).href
+        const url = new URL('/.api/completions/stream', this.config.serverEndpoint)
+
+        // Sourcegraph >=5.4 instances require client name and version params on the completions endpoint to ensure
+        // client supports Cody Ignore functionality. See https://github.com/sourcegraph/sourcegraph/pull/62048.
+        const { clientName, clientVersion } = this.config
+        if (clientName) {
+            url.searchParams.set('client-name', clientName)
+            if (clientVersion) {
+                url.searchParams.set('client-version', clientVersion)
+            }
+        }
+
+        return url.href
     }
 
     protected sendEvents(events: Event[], cb: CompletionCallbacks): void {
--- a/client/cody-shared/src/sourcegraph-api/completions/client.ts
+++ b/client/cody-shared/src/sourcegraph-api/completions/client.ts
@@ -15,7 +15,7 @@ export interface CompletionLogger {
 
 export type CompletionsClientConfig = Pick<
     ConfigurationWithAccessToken,
-    'serverEndpoint' | 'accessToken' | 'debugEnable' | 'customHeaders' | 'clientName' | 'clientVersion'
+    'serverEndpoint' | 'accessToken' | 'debugEnable' | 'customHeaders'
 >
 
 /**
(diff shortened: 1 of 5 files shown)


client/cody-shared/src/sourcegraph-api/completions/client.ts
```diff
@@ -35,7 +35,19 @@ export abstract class SourcegraphCompletionsClient {
...
+
+        // Sourcegraph >=5.4 instances require client name and version params on the completions endpoint to ensure
+        // client supports Cody Ignore functionality. See https://github.com/sourcegraph/sourcegraph/pull/62048.
+        const { clientName, clientVersion } = this.config
+        if (clientName) {
+            url.searchParams.set('client-name', clientName)
+            if (clientVersion) {
+                url.searchParams.set('client-version', clientVersion)
+            }
+        }
```
philipp-spiess, line 48: Any particular reason of why this is in the search params and not the JSON payload (IIRC the upstream API will ignore additional fields so that should be fine) or a HTTP header (it's usually where you'd expect this data e.g the `user-agent` header for web)?

philipp-spiess (approved):
LGTM just one question on why we chose search params

taras-yemets:
>❓ Question to reviewers
sourcegraph/sourcegraph [uses local](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/sourcegraph$+%22@sourcegraph/cody-shared%22+path:package.json&patternType=keyword&sm=0) "@sourcegraph/cody-shared" package". [So does sourcegraph/cody](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/cody+%22@sourcegraph/cody-shared%22+path:package.json&patternType=keyword&sm=0).
[SourcegraphCompletionsClient](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/sourcegraph$+SourcegraphCompletionsClient&patternType=keyword&sm=0) is used only for Cody Web.
Given that, should we just hardcode ?client-name=web in the completionsEndpoint instead of adding clientName and clientVersion to the configuration interface and passing these values as config when creating SourcegraphCompletionsClient instance?

WDYT, @philipp-spiess?

client/cody-shared/src/sourcegraph-api/completions/client.ts
philipp-spiess: Sorry I meant user-agent as a _precedence_ for how this is usually done on the web. Agree that we should not use user-agent itself for this. It's fine do use the param for now I think, soon the new Cody API will solve all these cosmetic inconsistencies anyways. 🙂

philipp-spiess:
>Given that, should we just hardcode ?client-name=web in the completionsEndpoint instead of adding clientName and clientVersion to the configuration interface and passing these values as config when creating SourcegraphCompletionsClient instance?

Sounds good yep

taras-yemets:
Refactored to hardcode client name in the completions endpoint as the latter is used only for Cody Web https://github.com/sourcegraph/sourcegraph/pull/62206/commits/3ae3be398dde3d275ccd2474edf8dad48962ec8d

client/cody-shared/src/sourcegraph-api/completions/client.ts
```diff
@@ -35,7 +35,14 @@ export abstract class SourcegraphCompletionsClient {
     }
 
     protected get completionsEndpoint(): string {
-        return new URL('/.api/completions/stream', this.config.serverEndpoint).href
+        const url = new URL('/.api/completions/stream', this.config.serverEndpoint)
+
+        // Sourcegraph >=5.4 instances require client name and version params on the completions endpoint to ensure client supports Cody Ignore functionality.
+        // Ensure client name is always set to "web" for Cody Web. Client version is not required for Cody Web as it aligns with server version.
+        // See https://github.com/sourcegraph/sourcegraph/pull/62048.
+        url.searchParams.set('client-name', 'web')
```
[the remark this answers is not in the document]
taras-yemets: Thanks for your review!
I don't have a strong opinion on query param vs header for this case. We already use the API version query param to check client version compatibility, so I added another check in this manner.
https://github.com/sourcegraph/sourcegraph/blob/d10d4f0922015ed786a9dd20a2e1b994f38c2a84/internal/completions/httpapi/handler.go#L84-L93  

We are migrating to the new Cody API soon, so this logic will likely be changed anyway.

PR #62213, opened 2024-04-26
sg/msp: architecture diagrams

jac:
Use d2 library to generate architecture diagrams for MSP Services

You can see the generated diagrams in the handbook PR
https://github.com/sourcegraph/handbook/pull/8919

For the first iteration of the diagrams I didn't include tooltips or links to the resources but that is something that can be done in a follow-up pr
## Test plan

locally tested


unknwon (approved):
🤞

dev/managedservicesplatform/operationdocs/diagram/diagram.go
```diff
@@ -0,0 +1,268 @@
...
+
+		rgraph, err = addDirectedConnection(rgraph, cloudrun, redis, "")
+		if err != nil {
+			return errors.Wrap(err, "failed to add connection from cloudrun to redis")
+		}
+		graph = rgraph
+	}
+
+	if env.Resources != nil && env.Resources.BigQueryDataset != nil {
+		bgraph, bigquery, err := bigquery(graph, env)
```
bobheadxi, line 144: nit - can we add a verb to all the constructors? for example:

```suggestion
		bgraph, bigquery, err := newBigQueryNode(graph, env)
```

Here, the var `bigquery` will shadow the function `bigquery`, and with the large amount of code here, it gets difficult to read

dev/managedservicesplatform/operationdocs/diagram/diagram.go
```diff
@@ -0,0 +1,268 @@
...
+		if err != nil {
+			return errors.Wrap(err, "failed to generate bigquery")
+		}
+
+		bgraph, err = addDirectedConnection(bgraph, cloudrun, bigquery, "")
+		if err != nil {
+			return errors.Wrap(err, "failed to add a connection from cloudrun to bigquery")
+		}
+		graph = bgraph
```
bobheadxi, line 153: since we return on error, I think this (and all equivalent callsites) can just assign directly to the final `graph`

```suggestion
		grah, err = addDirectedConnection(bgraph, cloudrun, bigquery, "")
		if err != nil {
			return errors.Wrap(err, "failed to add a connection from cloudrun to bigquery")
		}
```

dev/managedservicesplatform/operationdocs/internal/markdown/inline.go
```diff
@@ -68,6 +68,31 @@ func Linkf(text, url string, vars ...any) string {
...
 }
 
+// Image generates a markdown image.
+func Image(text, url string) string {
+	// some urls params are not escaped properly, let's fix that magically
+	parsedUrl, err := neturl.Parse(url)
+	// we are intentionally ignoring the error here
+	// if the user supplied an invalid url, it may or may not be intentional
+	// either way, not this method's problem
+	if err == nil {
```
bobheadxi, line 78: we still want it to "break" somewhat though - maybe we can put it in the output markdown? for example set `url` to `ERROR: %q`

dev/managedservicesplatform/operationdocs/diagram/msp.go
```diff
@@ -0,0 +1,142 @@
...
+func monitoring(graph *d2graph.Graph, _ *spec.EnvironmentSpec) (*d2graph.Graph, string, error) {
+	return CreateWithIcon(graph, "Monitoring", CloudMonitoring)
+}
+
+func opsgenie(graph *d2graph.Graph, _ *spec.EnvironmentSpec) (*d2graph.Graph, string, error) {
+	return CreateWithIcon(graph, "Opsgenie", Opsgenie)
+}
+
+func postgres(graph *d2graph.Graph, env *spec.EnvironmentSpec) (*d2graph.Graph, string, error) {
+	graph, key, err := CreateContainer(graph, "Postgres", CloudSQL)
```
bobheadxi, line 108:
```suggestion
	graph, key, err := CreateContainer(graph, "Cloud SQL (Postgres)", CloudSQL)
```

bobheadxi: Reason being in console etc the naming is `Cloud SQL` - we want to make it easy to retain a mapping from diagram to actual infra elements

dev/managedservicesplatform/operationdocs/diagram/components.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/sourcegraph/sourcegraph/lib/pointers"
+	"oss.terrastruct.com/d2/d2graph"
+	"oss.terrastruct.com/d2/d2oracle"
+)
+
+type Icon string
+
+var (
+	//go:embed assets/internet
+	Internet Icon
```
bobheadxi, line 17: Let's add a docstring here about why we need to do this (support rendering in Notion + portability in general), and add a `README` to `assets/` that documents where to find these nice icons and how to use them

dev/managedservicesplatform/operationdocs/diagram/diagram.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	if env.Resources != nil && env.Resources.PostgreSQL != nil {
```
bobheadxi, line 156: For this and Redis, I think we need to add a VPC network node:

```
Cloud Run --> VPC Network --> Redis
Cloud Run --> VPC Network --> Cloud SQL
```

this is a pretty important distinction IMO to indicate traffic does not route externally, and for both of the above, neither Redis nor Cloud SQL are accessible from public internet at all (both require a tunnelling mechanism)

dev/managedservicesplatform/operationdocs/internal/markdown/inline.go
bobheadxi: Aha, yeah I think we should 😆

dev/managedservicesplatform/operationdocs/diagram/msp.go
```diff
@@ -0,0 +1,146 @@
...
+func newSentryNode(graph *d2graph.Graph, _ *spec.EnvironmentSpec) (*d2graph.Graph, string, error) {
+	return createWithIcon(graph, "Sentry", Sentry)
+}
+
+func newTraceNode(graph *d2graph.Graph, _ *spec.EnvironmentSpec) (*d2graph.Graph, string, error) {
+	return createWithIcon(graph, "Cloud Trace", CloudTrace)
+}
+
+func newVPCNode(graph *d2graph.Graph, _ *spec.EnvironmentSpec) (*d2graph.Graph, string, error) {
+	return createWithIcon(graph, "VPC", VPC)
```
bobheadxi, line 145:
```suggestion
	return createWithIcon(graph, "VPC Network", VPC)
```

following the naming here: https://cloud.google.com/vpc/docs/vpc

dev/managedservicesplatform/operationdocs/diagram/msp.go
```diff
@@ -0,0 +1,146 @@
...
+
+func newSentryNode(graph *d2graph.Graph, _ *spec.EnvironmentSpec) (*d2graph.Graph, string, error) {
+	return createWithIcon(graph, "Sentry", Sentry)
+}
+
+func newTraceNode(graph *d2graph.Graph, _ *spec.EnvironmentSpec) (*d2graph.Graph, string, error) {
+	return createWithIcon(graph, "Cloud Trace", CloudTrace)
+}
+
+func newVPCNode(graph *d2graph.Graph, _ *spec.EnvironmentSpec) (*d2graph.Graph, string, error) {
```
bobheadxi, line 144: Do we need to include the unused variables still? I think we can just add them when needed

dev/managedservicesplatform/operationdocs/diagram/components.go
```diff
@@ -0,0 +1,108 @@
...
+// We need to embed the icons in our output svg
+// rather than using urls to externally hosted svgs
+// as _most_ sites block requests to external content for security reasons
+//
+// Note: when adding a new icon ensure there is no trailing whitespace/newline
+type Icon string
+
+var (
+	//go:embed assets/internet
+	Internet Icon
```
bobheadxi, line 23: Idea - should we move this `components.go` into `assets`, maybe named `assets.go`? Then callsites in `diagrams` would look like:

```go
return createWithIcon(graph, "Cloud Trace", assets.CloudTrace)
```

which reads nice

bobheadxi (approved):
Nice, LGTM other than some last code organization suggestions + naming of the VPC node. Reviewed the output based on https://deploy-preview-8919--sourcegraph-handbook.netlify.app/departments/engineering/managed-services/sourcegraph-accounts/#prod-architecture-diagram

This is great work, and some very nice-looking diagrams, thank you!

PR #62214, opened 2024-04-26
telemetry/resolvers, internal/telemetry, telemetry-gateway: validate feature/action

bobheadxi:
As suggested by @willdollman [here](https://sourcegraph.slack.com/archives/C1JH2BEHZ/p1713950660542809?thread_ts=1713950290.063969&cid=C1JH2BEHZ), we should do some more robust validation when telemetry is submitted by clients, to reduce the noise that ends up in BigQuery.

~This is not needed (yet) in backend or telemetry-gateway because those are from trusted sources.~ This is also added to telemetry-gateway and backend.

## Test plan

Unit tests (thanks Cody!)


unknwon (approved):

lib/telemetrygateway/v1/v1.go
```diff
@@ -4,11 +4,45 @@
...
+// - Not contain any whitespace
+var featureActionRegex = regexp.MustCompile(`^[a-z][a-zA-Z-\.]+$`)
+
+const featureActionMaxLength = 64
+
+// ValidateEventFeatureAction validates the given feature and action names. It
+// should be used where event features and actions are provided by a client.
+func ValidateEventFeatureAction(feature, action string) error {
+	if feature == "" || action == "" {
+		return errors.New("'feature', action must both be provided")
```
willdollman, line 33:
```suggestion
		return errors.New("'feature', 'action' must both be provided")
```

willdollman (approved):

lib/telemetrygateway/v1/v1.go
bobheadxi: Nice catch, thanks!

bobheadxi:
FYI for reviewers, I ended up:

- adding the validation to telemetry-gateway and backend SDK for completeness, since they already check for empty feature/action
- allowing numbers

PR #62221, opened 2024-04-26
chore: update SAMS SDK

unknwon:
Sister PR of https://github.com/sourcegraph/sourcegraph-accounts-sdk-go/pull/16

## Test plan

CI

### DIFF
--- a/cmd/telemetry-gateway/service/config.go
+++ b/cmd/telemetry-gateway/service/config.go
@@ -2,6 +2,7 @@ package service
 
 import (
 	sams "github.com/sourcegraph/sourcegraph-accounts-sdk-go"
+
 	"github.com/sourcegraph/sourcegraph/lib/managedservicesplatform/runtime"
 )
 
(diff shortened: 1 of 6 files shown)


cmd/telemetry-gateway/service/config.go
```diff
@@ -16,7 +17,13 @@ type Config struct {
 		StreamPublishConcurrency int
 	}
 
-	SAMS sams.ClientV1ConnConfig
+	SAMS SAMS
+}
+
+type SAMS struct {
```
bobheadxi, line 23:
```suggestion
type SAMSConfig struct {
```

bobheadxi (approved):
thank you!

cmd/telemetry-gateway/service/config.go
unknwon: Ah, this is not a own `config` package, good catch!

PR #62222, opened 2024-04-26
scip-ctags: solidify snapshot tests

jtibshirani:
While using the snapshot tests to check the output on a large file, I noticed
some tags were missing from the output. It turns out we accidentally save the
buffer of the writer into the snapshot, as opposed to the actual written
data. 

## Test plan
Test-only change


varungandhi-src (approved):
Good catch, thanks!

PR #62224, opened 2024-04-27
Add new ssc.useEmbeddedUI site configuration setting

chrsmith:
Adds a new ssc.useEmbeddedUI site config value. If set along with `SOURCEGRAPHDOTCOM_MODE`, it controls whether or not a new "Subscription Management Page" is available in the UI. (On `/cody/manage/subscription`.)

This is essentially a spiritual successor to @vdavid 's PR https://github.com/sourcegraph/sourcegraph/pull/62058. The objective is to start migrating the SSC UI pages into the monorepo. So as a first step, this PR just tries to get the mechanism by which we can enable/develop these new pages right.

After doing my best to research the options, I think (and may be totally wrong) that a site configuration setting is the most straight-forward option. We could make a "feature flag" or "temporary setting" work as well, but the site-config option seemed to be the easiest to grok.

As far as what that new `ssc.useEmbeddedUI` option controls, I just stubbed out a new `ManageSubscriptionPage` page. Served from `/cody/manage/subscription`, and not to be confused by `/cody/subscription` or `/cody/manage`. (And yes, we may want to revisit our URL layout as what we have today isn't particularly clear.)

## Test plan

Ran locally and confirmed we show a 404 if the site config value isn't set.

image

Otherwise, you are greeted with this beautiful sight:

image


schema/site.schema.json
```diff
@@ -17,6 +17,11 @@
       "default": "accounts.sourcegraph.com",
       "description": "The hostname of SAMS instance to connect."
     },
+    "ssc.useEmbeddedUI": {
```
chrsmith, line 20: Actually, rather than having several properties with the "ssc" prefix, we should probably just bite the bullet and have an `object` bundling everything. e.g.

```json
...
"codyProConfig": {
      "type": "object",
      "properties": {
          "useEmbeddedUI": {
          "type": "boolean",
          "default": false,
          }
     }
}
...
```

This way we can isolate all of the Cody Pro-specific configuration settings, such as the Stripe publishable key, URLs to the SSC backend, etc.

client/web/src/cody/management/subscription/ManageSubscriptionPage.tsx
```diff
@@ -0,0 +1,67 @@
...
+interface ManageSubscriptionPageProps extends TelemetryV2Props {
+    authenticatedUser: AuthenticatedUser | null
+}
+
+export const ManageSubscriptionPage: React.FunctionComponent<ManageSubscriptionPageProps> = ({
+    authenticatedUser,
+    telemetryRecorder,
+}) => {
+    useEffect(() => {
+        telemetryRecorder.recordEvent('cody.new-subscription-checkout', 'view')
```
thenamankumar, line 27: Event name should be updated to `cody.manage-subscription`, `view`.

client/web/src/cody/management/subscription/ManageSubscriptionPage.tsx
```diff
@@ -0,0 +1,67 @@
...
+                        </div>
+                    </PageHeader.Heading>
+                </PageHeader>
+
+                <div className={classNames('p-4 border bg-1 mt-4', styles.container)}>
+                    <div className="d-flex justify-content-between align-items-center border-bottom pb-3">
+                        <div>
+                            <H2>My subscription</H2>
+                            <Text className="text-muted mb-0">
+                                Hello {authenticatedUser?.displayName}. Depending on whether or not you have a Cody Pro
```
thenamankumar, line 52: For the upcoming PR implementing what is described in the text here. I recommend to isolate stripe checkout into a separate page and redirect there if no subscription if found on this page.

client/web/src/cody/util.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Returns whether or not to use the embedded Cody Pro subscription management UI,
+ * or direct users to https://accounts.sourcegraph.com
+ */
+export function showEmbeddedCodyProUI(): boolean {
+    return window?.context?.useSelfServeCodyEmbeddedUI === true
```
thenamankumar, line 6: We should not use the site config here but rather use feature flag, which should be simple and also provide us the ability to turn on this feature for specific internal users or all sourcegraph teammates for testing on prod before releasing it to the wider audience. 

I do not think that is possible with site config, or at least as simple as calling `useFeatureFlag`.

You should be able to simply `useFeatureFlag` hook to evaluate the feature flag. It is equally easy to do it at the backend as well.

Enable/Disable a feature flag is also much easy than updating the site config.

thenamankumar (changes requested):
The overall scaffolding looks good. Have left 2 comments. Highly recommend using feature flag to control the visibility of this feature rather than the site config.

client/web/src/cody/management/subscription/ManageSubscriptionPage.tsx
chrsmith: Sounds good. But what should we do if we have the same URL (`/manage/subscription`) perform different actions? e.g. if you don't already have a subscription we show the "new subscription flow", and if you do the "manage existing subscription" functionality.

I assume there is no problem with us emitting multiple telemetry events in that case? e.g.

```
ManageSubscriptionPage emits {cody.manage-subscription, view}
... which renders NewSubscriptionComponent, which emits {cody.new-subscription, view}
```

client/web/src/cody/management/subscription/ManageSubscriptionPage.tsx
chrsmith: Gotcha, so `manage/subscription` would redirect to `manage/subscription/new` if the user doesn't have a subscription? And otherwise show the "Update existing subscription" flow?

For reference, I think the latest [Figma designs are here](https://www.figma.com/file/FMSdn1oKccJRHQPgf7053o/Cody-PLG-GA?type=design&node-id=4042-4927&mode=design&t=WGtu3w1Iq7T3ZHCL-0). (But I am not certain.)

client/web/src/cody/util.ts
chrsmith: I see the benefits of a feature flag, and having more granular control for rolling things out. But I'm not quite convinced we should start with it. Because we'll need to wire site configuration data to the frontend anyways (for the "Stripe publishable key"). So needing both site configuration _and_ a feature flag to access the new UI seems redundant.

> Just "for now". When we are closer to actually shipping the new functionality, I would totally be onboard with removing `useEmbeddedUI` and requiring a feature flag. But it will be a little bit until we are ready to start testing this outside of `localhost`.)


For reference, this is the current shape of the `dotcom.codyProConfig` as I continue hacking on things.

```json
  "dotcom": {
    ...
    "codyProConfig": {
      "useEmbeddedUI": true,
      "stripePublishableKey": "pk_test_0b1ei45h7ypIEeAkXYKBU059",
      "samsBackendOrigin": "http://localhost:19992",
      "sscBackendOrigin": "http://localhost:19992"
    },
    ...
    // Existing SAMS client credentials, for contacting the SSC backend.
    // Since this is only applicable to the Cody Pro functionality, we can
    // maybe replace the "sams." prefix, with "samsCreds" and put the
    // values in the "dotcom.codyProConfig" block. WDYT?
    "sams.server": "http://localhost:19992",
    "sams.clientID": "sams_cid_...",
    "sams.clientSecret": "sams_cs_..."
    ...
  },
```

thenamankumar: `useEmbeddedUI` in the site config will not allow us to enable this for internal users before we release this to all the users. Feature flags have a purpose and they are not even difficult to use. It is only one line of code. 

We should focus on doing the things right way and not take shortcuts. I again highly recommend to use feature flag for this. If you need further help in understanding how to implement that, please follow this guide from Sourcegraph docs: https://docs.sourcegraph.com/dev/how-to/use_feature_flags#implement-a-feature-flag

client/web/src/cody/management/subscription/ManageSubscriptionPage.tsx
```diff
@@ -0,0 +1,67 @@
...
+    const { data, error: dataLoadError } = useQuery<UserCodyPlanResult, UserCodyPlanVariables>(USER_CODY_PLAN, {})
+    if (dataLoadError) {
+        throw dataLoadError
+    }
+
+    return (
+        <>
+            <Page className={classNames('d-flex flex-column')}>
+                <PageTitle title="New Subscription" />
+                <PageHeader className="mb-4 mt-4">
```
vdavid, line 39: Hmm, does this work? What provides the tailwind classes? 🤔 I don't think we have tailwind in the monorepo, do we?

client/web/src/cody/management/subscription/ManageSubscriptionPage.tsx
vdavid: Yeah, I agree with Naman, it'd be the same or better UX and better devX to have the two features on two separate pages.

client/web/src/cody/util.ts
vdavid: Sounds like a great case for a feature flag indeed, and a feature flag is a lot comfier to set and toggle (for testing) than a site config item.

chrsmith:
I've made a bit more progress on porting the overall SSC UI over, and will be closing this out in favor of the newer changes I have with all the things working end-to-end.

The first PR just adds the site configuration, here: https://github.com/sourcegraph/sourcegraph/pull/62255. I appreciate the feedback Naman, but I still don't think it's worth wiring through both the site configuration and a feature flag. But hopefully I'm able to get things ported over quickly enough that it's a moot point. (That is, replacing the config with a feature flag for rollout, and then have it enabled globally by default so we can remove the setting/flag all together.)

client/web/src/routes.constants.ts
```diff
@@ -39,8 +39,16 @@ export enum PageRoutes {
     CodySearch = '/search/cody',
     Cody = '/cody',
     CodyChat = '/cody/chat',
+
+    // The CodyManagement page is labeled as the "Dashboard" page.
     CodyManagement = '/cody/manage',
+    // The CodySubscriptions page is a side-by-side comparison of the
+    // Free, Pro, and Enterprise tiers.
```
vdavid, line 46: I'm unsure if these comments are helpful. This one is inaccurate, as in a mobile view, it's not a side-by-side comparison. Maybe if the comments capture the _function_ of the page rather than its layout, it could be more accurate and future-proof?

client/web/src/routes.constants.ts
```diff
@@ -39,8 +39,16 @@ export enum PageRoutes {
...
     CodySubscription = '/cody/subscription',
+    // The CodyManageSubscription page is where a user will either manage their
+    // existing Cody Pro subscription, or sign up for a new Cody Pro subscription.
```
vdavid, line 49: Note that this might just get inaccurate as soon as we split the two pages.

PR #62228, opened 2024-04-29
ci: remove chromatic

burmudar:
Since we're not renewing our chromatic this removes it from CI.

There are still some references in Storybooks but I think someone from frontend can remove that. @valerybugakov I see your name on the commits - donno if you want to take that on and remove it there too?
## Test plan
CI


Strum355 (approved):
Praise the lord

PR #62231, opened 2024-04-29
Cody Ignore: feature-flag context filters until Cody IDE clients support is added

taras-yemets:
This PR:
1. Doesn't let admin define "cody.contextFilters" in the site config as a temporary measure until Cody Ignore is supported in Cody for VSCode and JetBrains.  
Server-side Cody Ignore support is going to be shipped in the May 6th Sourcegraph release (tentatively `v5.4`). To ensure that Cody clients support Cody Ignore, we implement a minimum supported client version check (https://github.com/sourcegraph/sourcegraph/pull/62048). 
Cody Ignore support for Cody IDE clients (VSCode and JetBrains) will be shipped later in May but before the next Sourcegraph release. Before that, Cody Ignore can't be used as the bigger part of context gathering is happening on a client.

As Cody Ignore rules are configured in the site config (https://github.com/sourcegraph/sourcegraph/pull/61101), we want to prevent admins from setting them until the IDE support is shipped. To achieve that, we introduce the `cody-context-filters-enabled` feature flag. 
  - If `cody-context-filters-enabled` is not set or is set to `false`, site admins won't be able to save the site config if the latter has the "cody.contextFilters" field defined. Remote context filtering won't work either.
Screenshot 2024-04-29 at 13 59 56

  - If `cody-context-filters-enabled` is set to `true`, admins can set "cody.contextFilters" in the site config, and the remote context will be filtered according to these rules.
  
⚠️ As the [custom validator is created at the init time](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@a384a9d6a41024bd6492ace974a61a87c4528da0/-/blob/cmd/frontend/internal/codycontext/init.go?L42:8&popover=pinned), server restart is required after changing the feature flag value.

When Cody Ignroe functionality is developed in Cody clients, we are going to remove the `cody-context-filters-enabled` condition in a patch release. Customers who don't upgrade to this patch release will be able to use Cody Ignore by enabling the `cody-context-filters-enabled` feature flag.
More context can be found in [this](https://sourcegraph.slack.com/archives/C06Q9R098N4/p1714115458010179) thread.

2. Кemoves redundant `codycontext` package exports.

## Test plan
- Updated tests
- CI


cmd/frontend/internal/codycontext/init.go
```diff
@@ -0,0 +1,59 @@
...
+// TODO: remove this check after `CodyContextFilters` support is added to the IDE clients.
+func maybeValidateSiteConfig(ctx context.Context, db database.DB) error {
+	enabled, err := checkFeatureFlagEnabled(ctx, db)
+	if err != nil {
+		return err
+	}
+	if !enabled {
+		conf.ContributeValidator(func(confQuerier conftypes.SiteConfigQuerier) (problems conf.Problems) {
+			if confQuerier.SiteConfig().CodyContextFilters != nil {
+				problems = append(problems, conf.NewSiteProblem("\"cody.contextFilters\" param can't be set as it is not supported by Cody IDE clients (VSCode and JetBrains) yet. For information on when IDE support will be available, please visit our documentation: https://sourcegraph.com/docs/cody/capabilities/ignore-context#cody-ignore."))
```
taras-yemets, line 44: @aramaraju, WDYT about the error text?

willdollman, line 44: nit
```suggestion
				problems = append(problems, conf.NewSiteProblem("\"cody.contextFilters\" param can't be set as it is not supported by Cody IDE clients (VS Code and JetBrains) yet. For information on when IDE support will be available, please visit our documentation: https://sourcegraph.com/docs/cody/capabilities/ignore-context#cody-ignore."))
```

willdollman:
This lgtm from a security perspective but I'll hold off approving in favour of somebody with more code context.

cmd/frontend/internal/codycontext/init.go
aramaraju: LGTM, thank you Taras!

cmd/frontend/internal/codycontext/init.go
```diff
@@ -0,0 +1,59 @@
...
+// by Cody IDE clients (VSCode and JetBrains) yet.
+//
+// TODO: remove this check after `CodyContextFilters` support is added to the IDE clients.
+func maybeValidateSiteConfig(ctx context.Context, db database.DB) error {
+	enabled, err := checkFeatureFlagEnabled(ctx, db)
+	if err != nil {
+		return err
+	}
+	if !enabled {
+		conf.ContributeValidator(func(confQuerier conftypes.SiteConfigQuerier) (problems conf.Problems) {
```
slimsag, line 42: idea: If you move the `checkFeatureFlagEnabled` call right inside of this `ContributeValiator` callback function code block.. then it will check the feature flag each time before rendering the `problems` - meaning a server restart will not be required by admins.

slimsag (approved):
LGTM, I left a suggestion inline that I think could improve the UX but looks good!

cmd/frontend/internal/codycontext/init.go
taras-yemets: Thanks for your review, @slimsag! 

Problems returned by `ContributeValidator` callback functions are site (or external service) config validation problems. The non-nil error returned by `checkFeatureFlagEnabled` is not a site config problem actually and cannot be fixed by site admins (it'll be exposed in the admin UI but admins can't do anything about it).
If we move this check inside the `ContributeValidator` callback function, maybe we should return some user-friendly text like "Failed to get \"cody-context-filters-enabled\" feature flag value." if this error occurs, what do you think?

taras-yemets: Merged as-is to unblock clients ([thread](https://sourcegraph.slack.com/archives/C05MW2TMYAV/p1714481385612999)). Will address your comment in a follow-up PR.

slimsag: Good question! So you are just wondering about the error `checkFeatureFlagEnabled` might return?

I think if there is an error, it's probably unlikely to happen, so we could just log it and return no problems instead?

As long as the log line has some unique text we could easily search for in the logs later, then we can debug if anything is going wrong. Otherwise, admins can't do anything really to fix it, so no point in presenting them it I think.

taras-yemets: Addressed in https://github.com/sourcegraph/sourcegraph/pull/62329

PR #62234, opened 2024-04-29
[Svelte]: Add Search Input focus shortcut

vovakulikov:
Closes https://github.com/sourcegraph/sourcegraph/issues/61995

@fkling Not sure if this is a known problem with code mirror focus, but sometimes I can catch that focus doesn't preserve cursor input (meaning that focusing the input leads to cursor to be on the 0 position even if the input has some content) 

## Test plan
- Check that Search Input can be opened with `/` keyboard on the home, result and repository/blob pages


vovakulikov:
@fkling I think I found the right API with the cursor position dispatch fix

PR #62237, opened 2024-04-29
release: fix regression in release requirements check

BolajiOlajide:
The update to requirements check introduced a regression where excluded requirements would short-circuit the function for requirement checking and return a nil error.

This allowed the `createRelease` function to be executed even when requirements were missing.

## Test plan

* Manual testing
CleanShot 2024-04-29 at 14 59 49@2x


craigfurman (approved):
Good catch!

jhchabran (approved):
Thank you for fixing my dumb mistakes 🙏

PR #62253, opened 2024-04-29
remove unused get-cody page code

dadlerj:
After https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/commit/b3dc2f5783ad725f3cb408d111ab43b7ac8367d6?visible=1 it doesn't seem like this page is needed at all anymore. I noticed this while updating telemetry from v1 -> v2 and instead of moving on, it was quick to just delete the unused stuff.

## Test plan

tested locally


dadlerj:
I also have no idea why tests are failing here, but it runs just fine for me locally 🤔

vdavid (approved):

PR #62254, opened 2024-04-29
Code monitors: add some levers for tuning resource usage

camdencheek:
This adds a couple of configuration options which will allow us to tune the cost of code monitors in the event that we are seeing issues. It does not change the default values.

## Test plan

Tested that the conf validation works and that the intervals actually updated.

### DIFF
--- a/internal/codemonitors/background/workers.go
+++ b/internal/codemonitors/background/workers.go
@@ -33,7 +33,7 @@ func newTriggerQueryRunner(ctx context.Context, observationCtx *observation.Cont
 	options := workerutil.WorkerOptions{
 		Name:                 "code_monitors_trigger_jobs_worker",
 		Description:          "runs trigger queries for code monitors",
-		NumHandlers:          4,
+		NumHandlers:          conf.CodeMonitors().Concurrency,
 		Interval:             5 * time.Second,
 		HeartbeatInterval:    15 * time.Second,
 		Metrics:              metrics.workerMetrics,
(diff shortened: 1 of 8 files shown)


internal/codemonitors/background/workers.go
```diff
@@ -33,7 +33,7 @@ func newTriggerQueryRunner(ctx context.Context, observationCtx *observation.Cont
 	options := workerutil.WorkerOptions{
 		Name:                 "code_monitors_trigger_jobs_worker",
 		Description:          "runs trigger queries for code monitors",
-		NumHandlers:          4,
+		NumHandlers:          conf.CodeMonitors().Concurrency,
```
camdencheek, line 36: Change #1: allow configuring the number of code monitors that can run at the same time. This will effectively increase the poll interval if there are many code monitors.

internal/codemonitors/background/workers.go
```diff
@@ -185,7 +185,7 @@ func (r *queryRunner) Handle(ctx context.Context, logger log.Logger, triggerJob
 
 	// Log next_run and latest_result to table cm_queries.
 	newLatestResult := latestResultTime(q.LatestResult, results, searchErr)
-	err = cm.SetQueryTriggerNextRun(ctx, q.ID, cm.Clock()().Add(5*time.Minute), newLatestResult.UTC())
+	err = cm.SetQueryTriggerNextRun(ctx, q.ID, cm.Clock()().Add(conf.CodeMonitors().PollInterval), newLatestResult.UTC())
```
camdencheek, line 188: Change #2: allow configuring the poll interval of code monitors. This will let us tune down resource usage at the cost of increased latency.f

bahrmichael (approved):

PR #62255, opened 2024-04-29
Add new `dotcom.codyProConfig` site configuration object

chrsmith:
The "Cody Pro UI" is going to be served from sourcegraph.com instead of accounts.sourcegraph.com/cody. There is some nuance to the reason why we are making this change, but in short it is so that it will be easier to cleanly separate the "Cody Pro SKU" from the sourcegraph.com Sourcegraph instance down the road. (And instead, power the same "Sourcegraph UI" website from a different backend, via a compatible API.)

This PR is the first step towards that effort, updating the Sourcegraph site configuration schema to have a new `dotcom.codyProConfig` configuration object. Since Cody Pro for the time being is only available on Sourcegraph.com, it makes sense to put it there.

This also wires the new site configuration data through to the frontend as well, via the `window.context` / `context.ts`. However, we don't want to make _all_ Cody Pro configuration data because it will eventually contain secrets. So we instead wire a subset of those values via a `FrontendCodyProConfig` type.

We finally then export a new, but unused `showEmbeddedCodyProUi(): boolean` function. Subsequent PRs that add the UI pages currently served on accounts.sourcegraph.com/cody will use that function to gate registering new route handlers. (So when developing locally, `/cody/manage/team` will display the new page, but in production it will show a 404 error.)

When we are closer to shipping the embedded UI by default, we'll likely remove this setting entirely, and instead use an opt-in feature flag in order to roll out the new experience more gradually. But for now, the expectation is that `showEmbeddedCodyProUI` will only be set by @vdavid and I when running Sourcegraph locally. Since it will be a little while until we have enough of the SSC UI functionality migrated over for it to make sense to enable the embedded UI more granularly.

## Test plan

I've tested these changes locally (with other commits stacked on top). Since this just adds optional configuration values that are not required, its functionally a no-op.


schema/site.schema.json
```diff
@@ -2234,6 +2234,32 @@
...
+            },
+            "samsBackendOrigin": {
+              "type": "string",
+              "default": "",
+              "description": "Origin of the SAMS backend. (Must match the SAMS OIDC registration in auth.providers.)"
+            },
+            "sscBackendOrigin": {
+              "type": "string",
+              "default": "",
+              "description": "Origin of the Self-serve Cody backend."
```
vdavid, line 2260: Same here, I'll need guidance or an example when I try to use this setting, otherwise I probably need to find usages in the code etc, which seems messy.

schema/codypro_util.go
```diff
@@ -0,0 +1,21 @@
+package schema
+
+// FrontendCodyProConfig defines the subset of Cody Pro-specific configuration data
+// that we make available to the Sourcegraph frontend. (So any sensitive or irrelevant
+// site configuration data doesn't blindly get exposed to the web frontend.)
+//
+// Changes to this type need to be mirrored in `jscontext.ts`.
```
vdavid, line 7: Clarifying the path seems more helpful because we have two more `jscontext.ts` files in the repo:
```suggestion
// Changes to this type need to be mirrored in `client/web/src/cody/jscontext.ts`.
```

schema/BUILD.bazel
```diff
@@ -63,6 +63,7 @@ go_library(
     srcs = [
         "bitbucket_server_util.go",
         "bitbucketcloud_util.go",
+        "codypro_util.go",
```
vdavid, line 66: Very nit, how about we call this:
```suggestion
        "cody_pro_util.go",
```
Otherwise, the word "codypro" will trigger my spell checker each time, which would be slightly annoying.

vdavid (approved):

schema/site.schema.json
evict: I agree with @vdavid that we should add an example. I also think we need to have some validation here. From a security POV we should be very strict, as an invalid input here can lead to SSRF. For instance we should make sure the scheme is always `https` and ends with a trailing `/`. Something like:

```
func validateURL(inputURL string) (bool, error) {
    parsedURL, err := url.Parse(inputURL)
    if err != nil {
        return false, err
    }

    if parsedURL.Scheme != "https" {
        return false, fmt.Errorf("invalid URL scheme: %s", parsedURL.Scheme)
    }

    // make sure nothing can be appended to the hostname
    if parsedURL.Path[len(parsedURL.Path)-1] != '/' {
        return false, fmt.Errorf("path does not end with a slash")
    }
    
    return true
```
    
This code fits in `schema/codypro_util.go`. We make sure that the URL has `https` and the trailing slash to make sure people can not add additional parts to the domain name (to make SSRF happen still). An alternative for the trailing slash check would be to add `:443` at the end.

evict (approved):

schema/site.schema.json
evict: I agree with @vdavid that we should add an example. I also think we need to have some validation here. From a security POV we should be very strict, as an invalid input here can lead to SSRF. For instance we should make sure the scheme is always `https` and ends with a trailing `/`. Something like:

```
func validateURL(inputURL string) (bool, error) {
    parsedURL, err := url.Parse(inputURL)
    if err != nil {
        return false, err
    }

    if parsedURL.Scheme != "https" {
        return false, fmt.Errorf("invalid URL scheme: %s", parsedURL.Scheme)
    }

    // make sure nothing can be appended to the hostname
    if parsedURL.Path[len(parsedURL.Path)-1] != '/' {
        return false, fmt.Errorf("path does not end with a slash")
    }
    
    return true
```
    
This code fits in `schema/codypro_util.go`. We make sure that the URL has `https` and the trailing slash to make sure people can not add additional parts to the domain name (to make SSRF happen still). An alternative for the trailing slash check would be to add `:443` at the end.

evict (changes requested):

schema/site.schema.json
vdavid: In that case I'd include it in the doc that the scheme and the trailing `/` are _not_ needed, and apply those in the code. Then the validations are not needed.

evict: That also works!

client/web/src/jscontext.ts
```diff
@@ -301,6 +301,12 @@ export interface SourcegraphContext extends Pick<Required<SiteConfiguration>, 'e
         knownRoutes: string[]
         showToggle: boolean
     }
+
+    /** Configuration for Cody Pro-tier functionality, if applicable. */
+    frontendCodyProConfig?: {
+        stripePublishableKey: string
+        useEmbeddedUi: boolean
```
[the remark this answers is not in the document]
chrsmith: Again, I'm not opposed to a feature flag. And if you guys insist upon it, I definitely won't push back.

My only thinking is that it just seems unnecessary. Like if you _already_ need to set the Site config, AND we know a priori that we aren't anywhere close to running this anywhere other than `localhost`, it just seems like exposing the `useEmbeddedUi` as a feature flag is extra overhead with zero benefit. (e.g. as a developer, I need to set the `stripePublishableKey` site config, and also set the `useEmbeddedUi` feature flag.)

🤷 but if I don't get it, I don't get it.

Just to keep things simpler for the commits I have stacked on top of this, I'll merge it as-is. But when I do send out the first PR that adds React components, I'll remove this setting and instead rework it to rely on an `ff-use-cody-pro-embedded-ui` or similar.

schema/site.schema.json
```diff
@@ -2234,6 +2234,32 @@
...
+          "properties": {
+            "useEmbeddedUI": {
+              "type": "boolean",
+              "default": false,
+              "description": "Whether or not to direct users to manage their Cody Pro subscriptions on accounts.sourcegraph.com."
+            },
+            "stripePublishableKey": {
+              "type": "string",
+              "default": null,
+              "description": "Stripe publishable key for use in Stripe Checkout, Stripe Elements. This is not considered a secret."
```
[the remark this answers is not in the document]
chrsmith: I'm inclined to keep this as-is, because its a term of art. i.e. it literally exactly a "Stripe Publishable Key". And so I want to be very precise about what this thing should be. (Since "for use in..." isn't clear.)

https://docs.stripe.com/keys#obtain-api-keys

schema/site.schema.json
```diff
@@ -2234,6 +2234,32 @@
...
+            },
+            "samsBackendOrigin": {
+              "type": "string",
+              "default": "",
+              "description": "Origin of the SAMS backend. (Must match the SAMS OIDC registration in auth.providers.)"
```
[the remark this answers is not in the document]
chrsmith: @vdavid , I too find it helpful to clarify whether the protocol and slash and/or port are needed. But I did some research! And this is _precisely_ what an "Origin" is.

> Web content's origin is defined by the scheme (protocol), hostname (domain), and port of the [URL](https://developer.mozilla.org/en-US/docs/Glossary/URL) used to access it.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin
https://developer.mozilla.org/en-US/docs/Glossary/Origin

So I definitely want to keep the prefix of "Origin of the SAMS backend." (Since <3 origin, my new favorite technical term that I've been missing for so long.)

But I'll add an `"example": ["https://accounts.sourcegraph.com"],` field to the configuration thing, since I totally agree having examples would be useful. (And I'll do that for some of the other settings as well.

client/web/src/jscontext.ts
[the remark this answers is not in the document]
chrsmith: I just removed `useEmbeddedUI` entirely. So there is no more TypeScript portion of this code. (And will add the feature flag when I get there.)

chrsmith:
@evict, @vdavid  PTAL

- Rebased with `main` and force-pushed.
- Removed the `useEmbeddedUI` setting, and its plumbing to the frontend.
- Added examples for the new site config values.
- Registered a new schema validator, reporting errors if the origin is malformed or Stripe Publishable Key doesn't contain an expected prefix.

schema/site.schema.json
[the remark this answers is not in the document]
vdavid: If you had to look it up and I wasn't aware of it, that's already two data points in support for clarifying the term :)

unknwon (commented):
Changes self seems fine, but if it's all meant to be temporary and intentionally not discoverable by other... wouldn't it be easier and simpler to just read from env vars (that are easy to add and easy to remove). Site config, after all, is visible to all customer instances as well (but sure, under the `dotcom` section says enough about why they shouldn't care about it).

PR #62257, opened 2024-04-29
svelte: Add infinity scroll and search to tags page

### LINKED ISSUE #61633: Paginate tags
The GraphQL connection available for this is not ideal: It returns an `endCursor` but the field doesn't accept an `after` cursor parameter. Instead you can only fetch more data by increasing the value of the `first` parameter, resulting in fetching unnecessary data.

fkling:
Closes #61633 

This uses a similar approach as the commits page, except that pagination works a little bit differently.

Note that unlike the React app, every submission adds a new history entry, but also search submission has to be done explicitly. We can iterate on either if desired.

## Test plan

Manual testing.

- Scrolling loads more tags
- Search can be submitted via 'enter' or clicking the submit button
- Going back or forth restores previously loaded data and scroll position

### DIFF
--- a/client/web-sveltekit/src/lib/wildcard/Button.svelte
+++ b/client/web-sveltekit/src/lib/wildcard/Button.svelte
@@ -27,7 +27,7 @@
 
 <slot name="custom" {buttonClass}>
     <!-- $$restProps holds all the additional props that are passed to the component -->
-    <button class={buttonClass} {...$$restProps} on:click|preventDefault>
+    <button class={buttonClass} type="button" {...$$restProps} on:click>
         <slot />
     </button>
 </slot>
--- a/client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/Picker.svelte
+++ b/client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/Picker.svelte
@@ -114,10 +114,10 @@
     .root {
         display: flex;
         flex-direction: column;
-    }
 
-    :global([data-input-container]) {
-        margin: 0.75rem;
+        :global([data-input-container]) {
+            margin: 0.75rem;
+        }
     }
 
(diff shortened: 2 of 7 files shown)


client/web-sveltekit/src/lib/wildcard/Button.svelte
```diff
@@ -27,7 +27,7 @@
 
 <slot name="custom" {buttonClass}>
     <!-- $$restProps holds all the additional props that are passed to the component -->
-    <button class={buttonClass} {...$$restProps} on:click|preventDefault>
```
fkling, line 30: This prevented the form from being submitted by clicking the button as well as pressing enter in the input.
I added `type="button"` as default as an overwritable alternative.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/Picker.svelte
```diff
@@ -114,10 +114,10 @@
     .root {
         display: flex;
         flex-direction: column;
-    }
 
-    :global([data-input-container]) {
-        margin: 0.75rem;
+        :global([data-input-container]) {
+            margin: 0.75rem;
+        }
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/Picker.svelte:120]: This was accidentally applying styles to all Input components.

bahrmichael (approved):

taiyab:
FYI: I'm happy to have full "typical" pagination (numbers 1-n, with back/forward buttons) on these subpages for commits, tags, etc. This means any "sub UI" (e.g. the revision selector) can use infinite scrolling or run a search to find the relevant items.

fkling:
@taiyab For sure. That requires support form the backend though, which we currently don't have. The way we do pagination on the tags page is particularly bad because it means fetching data we already have. But again, we need to update the backend for better pagination.

PR #62258, opened 2024-04-29
Add SSC backend API Proxy

chrsmith:
> ~~ℹ️ This PR is stacked on top of https://github.com/sourcegraph/sourcegraph/pull/62255.~~

This PR adds an HTTP handler that will proxy HTTP requests sent to `http://sourcegraph.com/.api/ssc/proxy/*` to `https://accounts.sourcegraph.com/cody/api/v1/*`.

> **Why?**
> For a variety of reasons we are moving the SSC UI from accounts.sourcegraph.com into sourcegraph.com. It doesn't make a lot of sense to add Cody Pro-specific APIs to the Sourcegraph instance's GraphQL API, since the resolver step would just require us to translate/proxy the HTTP request anyways.
>
> So this "universal API proxy" approach is the simplest way to achieve the intended result. Although if this is too frightening, there are some alternatives we can pursue to make this more palatable...

This is done by introducing a new `ssc.APIProxy` type that implements the `http.Handler` interface. (Pedantically, the term "proxy" is probably generous, as there are any number of features that it does not support and/or it does not do correctly, such as a setting an `X-Forwarded-For` header. We don't need this type to be particularly robust beyond just handling prototypical HTTP requests sending/receiving small JSON payloads.)

So essentially we are just converting:

```diff
# Hypothetical HTTP request if it were JSON
{
    "method": "PATCH",
-   "targetDomain": "https://sourcegraph.com"
+   "targetDomain": "https://accounts.sourcegraph.com"
-   "urlPath": ".ssc/proxy/teams/members",
+   "urlPath": "cody/api/v1/teams/members",
   "headers": {
+       "Authorization": "Bearer {calling user's SAMS access token}",
-       "Cookie": "sgs={sourcegraph-session-cookie}",
   }
   "body": { ... }
}
```

## Key Feature: Setting the `Authorization` Header

The key feature of this API proxy is that it exchanges the credentials from the calling user for their backing SAMS access token. (And if no such external account is available for the calling user, a `401 Unauthorized` response is returned.)

The SSC backend only uses SAMS tokens, and we want to move away from any hard dependencies on sourcegraph.com. Such as needing to contact sourcegraph.com to verify tokens, for _it's view_ of user accounts. And instead, the SSC backend uses SAMS as the source of truth for users.

So in order for this "dotcom to SSC backend proxy" request to work, we need the following conditions to be met:

- The `http.Request`'s `context.Context` contains an authenticated `actor.Actor`. (Which is done via authentication middleware registered elsewhere in the HTTP handler registration.)
- The backing Sourcegraph user has an OIDC external account that is managed by SAMS. (For Sourcegraph.com this is the only registered authentication provider, and all active Sourcegraph.com accounts have a SAMS identity.)
- The credentials associated with that SAMS identity/external account (an `oauth2.Token`) is still valid. (A later PR will maybe add a background process to automatically refresh active user SAMS tokens, but for now we just serve a 401 if it has expired.)

## Additional Notes

While the proxied request URL, body, method, etc. are all user controlled, the _destination_ is not. The `sscBackendOrigin` is a site-level configuration knob. So this shouldn't be susceptible to an SSRF-style attack (I believe).

**Alternative Design**

If having a generic, open-ended proxy is too scary (no judgement), then an alternative would be to just replicate the exact URL routes that we have on the SSC backend in [backend/apis.go here](https://github.com/sourcegraph/self-serve-cody/blob/main/cody/backend/apis.go#L242).

So rather than having an HTTP handler for `.api/ssc/proxy/*`, we would need to maintain a list of more specific endpoints that align with what is offered on the SSC side. e.g.

```go
allowList := []urlRoute{
    {"GET", "/teams/current" },
    {"PATCH", "/teams/current" },
    {"LIST", "/teams/current/members" },
    {"GET", "/team/{teamID}/invites/{inviteID}" },
    {"POST", "/team/{teamID}/invites/{inviteID}" },
    ...
}
```

You can see that this will probably get tedious enough why I'm starting with a generic `/proxy/*` endpoint, and hoping that it is constrained enough that we are not worried about security threats.

## Test plan

Added new unit tests, and verified things are working locally (with some other local commits stacked on top).

### DIFF
--- a/internal/ssc/ssc_proxy.go
+++ b/internal/ssc/ssc_proxy.go
@@ -2,9 +2,10 @@ package ssc
 
 import (
 	"context"
-	"fmt"
 	"io"
 	"net/http"
+	"net/url"
+	"path"
 	"strings"
 	"time"
 
@@ -75,13 +76,27 @@ func (p *APIProxyHandler) buildProxyRequest(sourceReq *http.Request, token strin
 	// Source: ".api/ssc/proxy/" + "teams/current/members"
 	// Proxy :    "cody/api/v1/" + "teams/current/members"
 	sourceURLPath := strings.TrimPrefix(sourceReq.URL.Path, p.URLPrefix)
-	sscURL := fmt.Sprintf(
-		"%s/cody/api/v1/%s?%s",
+	sourceURLPath = path.Clean(sourceURLPath)
+	// Don't allow this sort of URL traversal, since we want to
+	// enforce that the URL contains the SSC prefix.
+	sourceURLPath = strings.Replace(sourceURLPath, "../", "", -1)
+
+	sscURLPath, err := url.JoinPath(
 		p.CodyProConfig.SscBackendOrigin,
-		sourceURLPath,
-		sourceReq.URL.Query().Encode())
-	p.Logger.Info("Building proxy request", log.String("method", sourceReq.Method), log.String("url", sscURL))
-	proxyReq, err := http.NewRequest(sourceReq.Method, sscURL, bodyReader)
+		"cody/api/v1",
+		sourceURLPath)
+	if err != nil {
+		return nil, errors.Wrap(err, "building proxy URL")
+	}
+
+	sscURL, err := url.Parse(sscURLPath)
+	if err != nil {
+		return nil, errors.Wrap(err, "parsing generated proxy URL")
+	}
+	sscURL.RawQuery = sourceReq.URL.Query().Encode()
+
+	p.Logger.Info("Building proxy request", log.String("method", sourceReq.Method), log.String("url", sscURL.String()))
+	proxyReq, err := http.NewRequest(sourceReq.Method, sscURL.String(), bodyReader)
 	if err != nil {
 		return nil, errors.Wrap(err, "creating request")
 	}
@@ -174,7 +189,7 @@ func (p *APIProxyHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// Copy the incomming request and send it to the SSC backend.
+	// Copy the incoming request and send it to the SSC backend.
 	proxyRequest, err := p.buildProxyRequest(r, samsToken.AccessToken)
 	if err != nil {
 		p.Logger.Error("building SSC proxy request", log.Error(err))
--- a/internal/ssc/ssc_proxy.go
+++ b/internal/ssc/ssc_proxy.go
@@ -76,10 +76,9 @@ func (p *APIProxyHandler) buildProxyRequest(sourceReq *http.Request, token strin
 	// Source: ".api/ssc/proxy/" + "teams/current/members"
 	// Proxy :    "cody/api/v1/" + "teams/current/members"
 	sourceURLPath := strings.TrimPrefix(sourceReq.URL.Path, p.URLPrefix)
+	sourceURLPath = "/" + sourceURLPath // Force the path to be rooted.
+	// nosemgrep: resolving the supplied path, to concatenate with the SSC API URL prefix below.
 	sourceURLPath = path.Clean(sourceURLPath)
-	// Don't allow this sort of URL traversal, since we want to
-	// enforce that the URL contains the SSC prefix.
-	sourceURLPath = strings.Replace(sourceURLPath, "../", "", -1)
 
 	sscURLPath, err := url.JoinPath(
 		p.CodyProConfig.SscBackendOrigin,
--- a/internal/ssc/ssc_proxy.go
+++ b/internal/ssc/ssc_proxy.go
@@ -45,8 +45,7 @@ var _ http.Handler = (*APIProxyHandler)(nil)
 
 // getUserIDFromRequest extracts the Sourcegraph User ID from the incomming request,
 // or returns an error suitable for sending to the end user.
-func (p *APIProxyHandler) getUserIDFromRequest(r *http.Request) (int32, error) {
-	ctx := r.Context()
+func (p *APIProxyHandler) getUserIDFromContext(ctx context.Context) (int32, error) {
 	callingActor := actor.FromContext(ctx)
 	if callingActor == nil || !callingActor.IsAuthenticated() {
 		p.Logger.Warn("rejecting request made by unauthenticated Sourcegraph user")
@@ -77,7 +76,7 @@ func (p *APIProxyHandler) buildProxyRequest(sourceReq *http.Request, token strin
 	// Proxy :    "cody/api/v1/" + "teams/current/members"
 	sourceURLPath := strings.TrimPrefix(sourceReq.URL.Path, p.URLPrefix)
 	sscURL := fmt.Sprintf(
-		"%s/cody/api/v1%s?%s",
+		"%s/cody/api/v1/%s?%s",
 		p.CodyProConfig.SscBackendOrigin,
 		sourceURLPath,
 		sourceReq.URL.Query().Encode())
@@ -142,7 +141,7 @@ func (p *APIProxyHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	p.Logger.Info("proxying SSC API request", log.String("url", r.URL.String()))
 
-	sgUserID, err := p.getUserIDFromRequest(r)
+	sgUserID, err := p.getUserIDFromContext(ctx)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusUnauthorized)
 		return
--- a/internal/ssc/ssc_proxy.go
+++ b/internal/ssc/ssc_proxy.go
@@ -1,0 +1,227 @@
+package ssc
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"time"
+
+	"golang.org/x/oauth2"
+
+	"github.com/sourcegraph/log"
+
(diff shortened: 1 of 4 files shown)


vdavid (commented):
The code looks good to me. Thanks for the abundant comments!
You'll need to run `bazel run //:configure` to make CI happy.

Regarding this: "A later PR will maybe add a background process to automatically refresh active user SAMS tokens, but for now we just serve a 401 if it has expired.", I think we definitely want _some_ solution to the problem (don't want to require users to log out and back in again). Could you make sure there's an issue that tracks this and link it to this PR?

vdavid (approved):
Oops, I meant to approve

internal/ssc/ssc_proxy.go
```diff
@@ -0,0 +1,226 @@
...
+	// Source: ".api/ssc/proxy/" + "teams/current/members"
+	// Proxy :    "cody/api/v1/" + "teams/current/members"
+	sourceURLPath := strings.TrimPrefix(sourceReq.URL.Path, p.URLPrefix)
+	sscURL := fmt.Sprintf(
+		"%s/cody/api/v1/%s?%s",
+		p.CodyProConfig.SscBackendOrigin,
+		sourceURLPath,
+		sourceReq.URL.Query().Encode())
+	p.Logger.Info("Building proxy request", log.String("method", sourceReq.Method), log.String("url", sscURL))
+	proxyReq, err := http.NewRequest(sourceReq.Method, sscURL, bodyReader)
```
evict, line 84: I would recommend using `url.JoinPath` here, instead of using `fmt.Sprintf`, and call `path.Clean` before joining the API path with the user controlled bits. This makes sure we always go to `cody/api/v1`. 

```suggestion
	sourceURLPath := strings.TrimPrefix(sourceReq.URL.Path, p.URLPrefix)
	// make sure the path is clean
	apiPath := fmt.Sprintf("/cody/api/v1/%s", path.Clean(sourceURLPath))
	
	fullURL, err := url.JoinPath(baseURL.String(), apiPath)
	if err != nil {
		return nil, errors.Wrap(err, "error constructing fullpath")
	}

	sscURL, err := url.Parse(fullURL)
	if err != nil {
		return "", errors.Wrap(err, "error parsing full path)
	}
	sscURL.RawQuery = sourceReq.Query().Encode()
	proxyReq, err := http.NewRequest(sourceReq.Method, sscURL.String(), bodyReader)
```

internal/ssc/ssc_proxy.go
```diff
@@ -0,0 +1,226 @@
...
+	//
+	// The frontend needs to expect this 401 response, and force the user to
+	// reauthenticate. (Which would then pick up a new SAMS auth token.) Or
+	// we need to have a background process that will periodically refresh the
+	// user's SAMS credentials, so that the refresh and access token for the
+	// user's SAMS account are sufficiently fresh.
+	if samsToken.Expiry.Before(time.Now()) {
+		p.Logger.Warn("the user's SAMS token has expired", log.Time("expiry", samsToken.Expiry))
+		http.Error(w, "Sourcegraph Accounts identity has expired", http.StatusUnauthorized)
+		return
```
evict, line 174: Should we make sure this is UTC?

internal/ssc/ssc_proxy.go
```diff
@@ -0,0 +1,226 @@
...
+
+	// Copy the incomming request and send it to the SSC backend.
+	proxyRequest, err := p.buildProxyRequest(r, samsToken.AccessToken)
+	if err != nil {
+		p.Logger.Error("building SSC proxy request", log.Error(err))
+		http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+		return
+	}
+
+	client := http.DefaultClient
```
evict, line 185: Can we use client from `internal/httpcli/client.go` i.e. `ExternalClient`?

evict:
Looks good! I do think we need to add some additional hardening for the user controlled bits. :+1:

internal/ssc/ssc_proxy.go
```diff
@@ -0,0 +1,226 @@
...
+	// we need to have a background process that will periodically refresh the
+	// user's SAMS credentials, so that the refresh and access token for the
+	// user's SAMS account are sufficiently fresh.
+	if samsToken.Expiry.Before(time.Now()) {
+		p.Logger.Warn("the user's SAMS token has expired", log.Time("expiry", samsToken.Expiry))
+		http.Error(w, "Sourcegraph Accounts identity has expired", http.StatusUnauthorized)
+		return
+	}
+
+	// Copy the incomming request and send it to the SSC backend.
```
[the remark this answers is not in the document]
chrsmith: Fixed. "How sad it is that some people can only think of a single way to spell a word." 😜🙃😭

internal/ssc/ssc_proxy.go
chrsmith: Good catch. I'm relatively new to the monorepo, so I have no idea what the right idioms to use are. 👍

chrsmith:
Rebased on top of https://github.com/sourcegraph/sourcegraph/pull/62255, and addressed PR feedback.

The URL parsing shenanigans now look like this. (So while we use `path.Clean` and `url.JoinPath`, I still need to call `strings.Replace(..., "../", "", -1)` to get the right behavior when given odd looking requests to work right. (See the updates to the unit tests.)

```go
	// Construct the remapped URL on the SSC backend.
	// For example:
	// Source: ".api/ssc/proxy/" + "teams/current/members"
	// Proxy :    "cody/api/v1/" + "teams/current/members"
	sourceURLPath := strings.TrimPrefix(sourceReq.URL.Path, p.URLPrefix)
	sourceURLPath = path.Clean(sourceURLPath)
	// Don't allow this sort of URL traversal, since we want to
	// enforce that the URL contains the SSC prefix.
	sourceURLPath = strings.Replace(sourceURLPath, "../", "", -1)

	sscURLPath, err := url.JoinPath(
		p.CodyProConfig.SscBackendOrigin,
		"cody/api/v1",
		sourceURLPath)
	if err != nil {
		return nil, errors.Wrap(err, "building proxy URL")
	}

	sscURL, err := url.Parse(sscURLPath)
	if err != nil {
		return nil, errors.Wrap(err, "parsing generated proxy URL")
	}
	sscURL.RawQuery = sourceReq.URL.Query().Encode()

	p.Logger.Info("Building proxy request", log.String("method", sourceReq.Method), log.String("url", sscURL.String()))
	proxyReq, err := http.NewRequest(sourceReq.Method, sscURL.String(), bodyReader)
```

internal/ssc/ssc_proxy.go
```diff
@@ -0,0 +1,241 @@
...
+
+	// Construct the remapped URL on the SSC backend.
+	// For example:
+	// Source: ".api/ssc/proxy/" + "teams/current/members"
+	// Proxy :    "cody/api/v1/" + "teams/current/members"
+	sourceURLPath := strings.TrimPrefix(sourceReq.URL.Path, p.URLPrefix)
+	sourceURLPath = path.Clean(sourceURLPath)
+	// Don't allow this sort of URL traversal, since we want to
+	// enforce that the URL contains the SSC prefix.
+	sourceURLPath = strings.Replace(sourceURLPath, "../", "", -1)
```
evict, line 82: This essentially doesn't do anything after the `path.Clean` above.

internal/ssc/ssc_proxy.go
```diff
@@ -0,0 +1,241 @@
...
+		}
+		bodyReader = strings.NewReader(string(bodyBytes))
+	}
+
+	// Construct the remapped URL on the SSC backend.
+	// For example:
+	// Source: ".api/ssc/proxy/" + "teams/current/members"
+	// Proxy :    "cody/api/v1/" + "teams/current/members"
+	sourceURLPath := strings.TrimPrefix(sourceReq.URL.Path, p.URLPrefix)
+	sourceURLPath = path.Clean(sourceURLPath)
```
evict, line 79:
```suggestion
    // nosemgrep: resolving the supplied path, to concatenate with `../` in the next step.
	sourceURLPath = path.Clean(sourceURLPath)
```

internal/ssc/ssc_proxy.go
chrsmith: Here we go. The behavior of `path.Clean` depends on whether or not the path is "rooted".

So this does what we want (and obviates the need for the janky AF `strings.Replace("../")`).

```go
	sourceURLPath := strings.TrimPrefix(sourceReq.URL.Path, p.URLPrefix)
	sourceURLPath = "/" + sourceURLPath
	// nosemgrep: resolving the supplied path, to concatenate with `../` in the next step.
	sourceURLPath = path.Clean(sourceURLPath)
```

internal/ssc/ssc_proxy.go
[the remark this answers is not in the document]
chrsmith: This alert likely isn't wrong. But the code here is a little tricky. Since the later call to `url.JoinPath` calls `path.Clean` under the hood. But we _do_ need to call `path.Clean` _before_ the call to `url.JoinPath`, in order to canonicalize any directory traversal before prefixing the URL path with the "cody/api/v1" prefix.

... it's confusing. I added another unit test to make the expected behavior clearer.

chrsmith:
https://github.com/sourcegraph/sourcegraph/pull/62255 has been merged into `main`. Rebasing on top of `main` with those changes, and force-pushed.

PR #62259, opened 2024-04-30
rockskip: Switch from long-running process to pagination

eseliger:
We want to migrate this call to gRPC. However, it isn't good practice to have to keep this process running for potentially hours, as any server restart will have to make the process start over.
With pagination here, we're not occupying a process slot on gitserver, and rollouts of gitserver should much less affect rockskip, as individual pages can be retried before being returned to the rockskip code.

Test plan:

Existing tests still pass, but not a Rockskip expert so requesting review from the owners.

### DIFF
--- a/cmd/symbols/gitserver/client.go
+++ b/cmd/symbols/gitserver/client.go
@@ -101,8 +101,30 @@ func (c *gitserverClient) LogReverseEach(ctx context.Context, repo string, commi
 	return c.innerClient.LogReverseEach(ctx, repo, commit, n, onLogEntry)
 }
 
+const revListPageSize = 100
+
 func (c *gitserverClient) RevList(ctx context.Context, repo string, commit string, onCommit func(commit string) (shouldContinue bool, err error)) error {
-	return c.innerClient.RevList(ctx, repo, commit, onCommit)
+	nextCursor := commit
+	for {
+		var commits []api.CommitID
+		var err error
+		commits, nextCursor, err = c.innerClient.RevList(ctx, api.RepoName(repo), nextCursor, revListPageSize)
+		if err != nil {
+			return err
+		}
+		for _, c := range commits {
+			shouldContinue, err := onCommit(string(c))
+			if err != nil {
+				return err
+			}
+			if !shouldContinue {
+				return nil
+			}
+		}
+		if nextCursor == "" {
+			return nil
+		}
+	}
 }
 
 var NUL = []byte{0}
--- a/internal/gitserver/client.go
+++ b/internal/gitserver/client.go
@@ -494,9 +494,10 @@ type Client interface {
 	// LogReverseEach runs git log in reverse order and calls the given callback for each entry.
 	LogReverseEach(ctx context.Context, repo string, commit string, n int, onLogEntry func(entry gitdomain.LogEntry) error) error
 
-	// RevList makes a git rev-list call and iterates through the resulting commits, calling the provided
-	// onCommit function for each.
-	RevList(ctx context.Context, repo string, commit string, onCommit func(commit string) (bool, error)) error
+	// RevList makes a git rev-list call and returns up to count commits. if nextCursor
+	// is non-empty, it is used as the starting point for the next call, use it to iterate
+	// to iterate over the whole history.
+	RevList(ctx context.Context, repo api.RepoName, commit string, count int) (_ []api.CommitID, nextCursor string, err error)
 
 	// SystemsInfo returns information about all gitserver instances associated with a Sourcegraph instance.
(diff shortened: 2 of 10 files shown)


eseliger:
* **#62260** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62260?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#62259** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62259?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

stefanhengl (approved):
LGTM!

I focused mainly on Rockskip.

I compared the index of pinned-zoekt for this PR vs. main: Number of symbols and count of ancestors match. I also compared the hops for a couple of symbols and they matched, too.  

Note:
Surprisingly, the reported size of the index changed. However, I reran the index for main and the reported size fluctuates for the same index. It might be connected to the unstable order of symbols but I have to take a closer look to be sure.

internal/rockskip/server_test.go
```diff
@@ -45,50 +47,22 @@ func (mockParser) Parse(path string, bytes []byte) ([]*ctags.Entry, error) {
...
 func TestIndex(t *testing.T) {
-	fatalIfError := func(err error, message string) {
-		if err != nil {
-			t.Fatal(errors.Wrap(err, message))
-		}
-	}
-
-	gitDir, err := os.MkdirTemp("", "rockskip-test-index")
-	fatalIfError(err, "faiMkdirTemp")
+	gitserver.ClientMocks.LocalGitserver = true
```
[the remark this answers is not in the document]
eseliger: No _intended_ functional changes, just adapted the tests to work properly with the adjusted backend without just hard-coding git output here

cmd/symbols/gitserver/client.go
```diff
@@ -101,8 +101,30 @@ func (c *gitserverClient) LogReverseEach(ctx context.Context, repo string, commi
...
+			return err
+		}
+		for _, c := range commits {
+			shouldContinue, err := onCommit(string(c))
+			if err != nil {
+				return err
+			}
+			if !shouldContinue {
+				return nil
+			}
```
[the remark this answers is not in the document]
eseliger: I didn't intend it to be different - what's the change you observed here?

cmd/symbols/gitserver/client.go
```diff
@@ -101,8 +101,30 @@ func (c *gitserverClient) LogReverseEach(ctx context.Context, repo string, commi
...
 
+const revListPageSize = 100
+
 func (c *gitserverClient) RevList(ctx context.Context, repo string, commit string, onCommit func(commit string) (shouldContinue bool, err error)) error {
-	return c.innerClient.RevList(ctx, repo, commit, onCommit)
+	nextCursor := commit
+	for {
+		var commits []api.CommitID
+		var err error
+		commits, nextCursor, err = c.innerClient.RevList(ctx, api.RepoName(repo), nextCursor, revListPageSize)
```
ggilmore, line 111: Can this be an in-line assignment instead of a declaration and then assignment? 


```suggestion
commits, nextCursor, err := c.innerClient.RevList(ctx, api.RepoName(repo), nextCursor, revListPageSize)
```

eseliger: This would create a block-scoped `nextCursor` so it wouldn't update that one - I could do a `nc` and then a second line with `nextCursor = nc` if that's prettier?

ggilmore: Ah, I see. Don't worry about it, I think this looks fine

internal/gitserver/client.go
```diff
@@ -494,9 +494,10 @@ type Client interface {
 	// LogReverseEach runs git log in reverse order and calls the given callback for each entry.
 	LogReverseEach(ctx context.Context, repo string, commit string, n int, onLogEntry func(entry gitdomain.LogEntry) error) error
 
-	// RevList makes a git rev-list call and iterates through the resulting commits, calling the provided
-	// onCommit function for each.
-	RevList(ctx context.Context, repo string, commit string, onCommit func(commit string) (bool, error)) error
+	// RevList makes a git rev-list call and returns up to count commits. if nextCursor
+	// is non-empty, it is used as the starting point for the next call, feed it to commit
```
ggilmore, line 498: I think there is a grammtical issue (feed it to commit). Maybe just "to iterate" like the next line?

```suggestion
	// is non-empty, it is used as the starting point for the next call, use it to iterate
```

ggilmore (approved):

Issue #62100, opened 2024-04-23
gitserver: Move LogReverseEach to new gRPC pattern


eseliger:

eseliger:
Currently blocked by lack of Rockskip integration tests to validate changes to this method.

PR #62260, opened 2024-04-30
rockskip: Replace second long-running process with API

eseliger:
We want to migrate this call to gRPC as well. However, it isn't good practice to have to keep this process running for potentially hours, as any server restart will have to make the process start over.
With one call per commit, we're not occupying a process slot on gitserver, and rollouts of gitserver should much less affect rockskip, as individual commits can be retried before being returned to the rockskip code.

Closes https://github.com/sourcegraph/sourcegraph/issues/62100

Test plan:

Existing tests still pass, I let it index a repo locally and that was fast and I get results for symbols instantly now, but not a Rockskip expert so requesting review from the owners.


eseliger:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62260?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#62260** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62260?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#62259** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62259?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

internal/rockskip/index.go
```diff
@@ -95,213 +96,220 @@ func (s *Service) Index(ctx context.Context, repo, givenCommit string) (err erro
 
 	tasklog.Start("Log")
 	entriesIndexed := 0
-	err = s.git.LogReverseEach(ctx, repo, givenCommit, missingCount, func(entry gitdomain.LogEntry) error {
-		defer tasklog.Continue("Log")
+	for _, missingCommit := range missingCommits {
+		err := func() error {
```
eseliger, line 100: Wrapped for the defers used in this block.

pjlast:
Changes look good to me, but probably worth waiting for someone with a bit more ownership

eseliger:
On hold until we have an integration test suite for rockskip, and until ChangedFiles is merged by Geoffrey which we might be able to use here instead of a second ChangedDiffFiles.

keegancsmith (commented):
I'm not sure how slow the rockskip stuff is once we have the "logentry" of a missing commit, but if its fast then I can imagine this slowing down rockskip a bunch? I see there is quite a lot of instrumentation inside of index.go, do we have any ideas here if rockskip being slow to index is git interactions or if its the work it does. I imagine its rockskip that is slow not git given some of the numbers @stefanhengl mentioned. But just calling that out just in case :)

eseliger:
Thanks for the callout - good angle :) Where can I find those numbers @stefanhengl mentioned?

eseliger:
I decided to take a slightly different stab at this, so here's the new PR https://github.com/sourcegraph/sourcegraph/pull/62734

PR #62252, opened 2024-04-29
gitserver: grpc: port DiffSymbols from client as new GitBackend.ChangedFiles() func

ggilmore:

Issue #60654, opened 2024-02-21
Provide GQL API to list all files and change mode in a given diff


eseliger:
This is needed to build a better diff view where all files changed are shown in a sidebar.

`git diff-tree --no-prefix --format=format: --root -r --name-status 65564433042849da9e25d592f7b42e5d51d61996` seems to be rather quick.

eseliger:
@ggilmore I think your train will close this ticket as well - I'll assign to you and this iteration but feel free to move it back if not appropriate.

PR #62262, opened 2024-04-30
gitserver: grpc: create changed files RPC implementation for gRPC server

ggilmore:
Part of https://github.com/sourcegraph/sourcegraph/issues/60654

This PR implements a new gitserver gRPC method, ChangedFiles that lists the files that changed between two commits alonside their added/modified/deleted status. This PR builds on the functionality exposed in https://github.com/sourcegraph/sourcegraph/pull/62252. 

## Test plan

Unit tests


internal/gitserver/v1/gitserver.proto
```diff
@@ -248,6 +248,16 @@ service GitserverService {
...
     option idempotency_level = NO_SIDE_EFFECTS;
   }
+
+  // ChangedFiles returns the list of files that have been added, modified, or
+  // deleted in the entire repository between the two given <tree-ish> identifiers (e.g., commit, branch, tag).
+  //
+  // If base is omitted, the parent of head is used as the base.
+  //
+  // If either the base or head <tree-ish> id does not exist, an error with a RevisionNotFoundPayload is returned.
+  rpc ChangedFiles(ChangedFilesRequest) returns (stream ChangedFilesResponse) {
```
ggilmore, line 258: Note: I made the decision to make this a server-streaming RPC endpoint. Depending on the commits you choose (such as diff'ing between the root and HEAD of a large repository), the list of changed paths can be humongous. 

I ran a local experiment with the megarepo, and diffing the root and HEAD already took up 34 megabytes. This was enough evidence to convince me that the response needs to be streamed over multiple messages.

internal/gitserver/gitdomain/log.go
```diff
@@ -18,14 +19,72 @@ type PathStatus struct {
...
 type StatusAMD int
 
 const (
-	AddedAMD    StatusAMD = 0
-	ModifiedAMD StatusAMD = 1
-	DeletedAMD  StatusAMD = 2
+	AddedAMD             StatusAMD = 0
+	ModifiedAMD          StatusAMD = 1
+	DeletedAMD           StatusAMD = 2
+	StatusUnspecifiedAMD StatusAMD = 3
```
eseliger, line 46: It would be nice if unspecified could be at the beginning of the enum, I'm sure we'll want to add rename support here later, and then you cannot do something like 
`if s <= RenamedAMD && s > Unspecified` because the unspecified state would be "in the middle" of the enum. 
I don't think this would affect anything in practice, as rockskip (IIRC) doesn't ever store this value, it's just used in-memory to represent the state.

Also, that would make gRPC and Go code numbers align, which is probably helpful in debugging

eseliger (approved):

eseliger (approved):

ggilmore:
### Merge activity

* **May 7, 1:33 PM EDT**: [Graphite](https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62262) rebased this pull request after merging its parent, because this pull request is set to merge when ready.

PR #62264, opened 2024-04-30
chore: Remove redundant loop captures

varungandhi-src:
Go 1.22 changes loop variables to have more sensible
semantics where the variable is not reused across iterations
by the codegen, so we can simplify a bunch of code
working around that counterintuitive behavior.

## Test plan

Covered by existing tests


camdencheek (approved):
Nice! Exciting to see that linter go away 😄 Thank you!

PR #62266, opened 2024-04-30
Add 'internal' case for repositoryQuery

### LINKED ISSUE #61924: Support internal repos on GHE better without having to individually add orgs
There are 3 ways to add `internal` repos to an instance:

- Add the repo by name
- Use `"repositoryQuery": ["public"]`
- Use `"repositoryQuery": ["org:org-name is:internal"]`

However, on instances with many, fluctuating repos and fluctuating orgs, none of these are feasible.

`"public"` uses GitHub's public repos endpoint, which does not return `visibility` or `topics` as part of the JSON result, making it impossible to tell which repos are actually internal and yielding inconsistent search results for repos with topics.

And having to add a new repo or a new org to code host connections every time is extremely tedious.

What we should be able to do instead is add a new `"internal"` option to the `repositoryQuery` special cases that iterates over the users orgs, and lists all the repos for each org using `visibility=internal`

pjlast:
Closes #61924 

Docs update: https://github.com/sourcegraph/docs/pull/286

Adds 'internal' as a special case option for the repositoryQuery field on GitHub code host connections. This query will fetch all orgs the user of the PAT belongs to and construct a search query consisting of `org:org1 org:org2 ... is:internal`.

A search query is used instead of the standard `org/repos` endpoint, because for a large amount of orgs it is much faster to fetch all of the repositories with the search API than to hit the `/repos` endpoint for each individual org.

## Test plan

Test added for `"repositoryQuery": ["internal"]`


internal/repos/github.go
```diff
@@ -807,6 +824,31 @@ func (s *GitHubSource) listAffiliatedPage(ctx context.Context, first int, result
 	}
 }
 
+// Fetches a list of all internal repositories visible to the authenticated user.
```
eseliger, line 827: nit: go style 

```suggestion
// fetchInternal fetches a list of all internal repositories visible to the authenticated user.
```

internal/repos/github.go
```diff
@@ -807,6 +824,31 @@ func (s *GitHubSource) listAffiliatedPage(ctx context.Context, first int, result
...
+	if err != nil {
+		results <- &githubResult{err: err}
+		return
+	}
+
+	var sb strings.Builder
+
+	for _, org := range orgs {
+		sb.WriteString("org:")
+		sb.WriteString(org.Login)
```
eseliger, line 844: I hope those can't contain spaces or colons themselves 😬

internal/repos/github.go
```diff
@@ -807,6 +824,31 @@ func (s *GitHubSource) listAffiliatedPage(ctx context.Context, first int, result
 	}
 }
 
+// Fetches a list of all internal repositories visible to the authenticated user.
+// It first fetches a list of all organizations the user belongs to and then fetches
+// all of the internal repositories belonging to these organizations by using
+// a search query. This leads to much fewer requests to the GitHub API.
+func (s *GitHubSource) listInternal(ctx context.Context, results chan *githubResult) {
+	orgs, err := fetchAll(func(page int) (items []*github.Org, hasNext bool, cost int, err error) {
+		return s.v3Client.GetAuthenticatedUserOrgs(ctx, page)
```
eseliger, line 833: I'm assuming that works with GitHub App installations as well, where only a subset of orgs may be granted access to?

eseliger (approved):
LGTM, we talked through this on a call. In a next step, we want to announce that public will change behavior in a future version and that internal should be added as well, when public was used before.

mmanela:
@pjlast  Can we update changelog too?

eseliger (commented):
Can we update this doc string as well? https://sourcegraph.sourcegraph.com/-/editor?remote_url=github.com%2Fsourcegraph%2Fsourcegraph&branch=main&file=schema%2Fgithub.schema.json&editor=VSCode&version=2.2.16&start_row=159&start_col=6&end_row=159&end_col=6

internal/repos/github.go
pjlast: GitHub App installations are tied to a single org/user. So this wouldn't have worked. I added a special case in case we're dealing with a github app, although it would be more efficient for the admin to add a repositoryQuery with `org:org-name is:internal` than any options the GitHub App installations API gives us.

Actually, perhaps that's what we should do 🤔  return an error asking them to do that

PR #62272, opened 2024-04-30
Fix search box `/` shourtcut

vovakulikov:
Fixes problem which was introduced by https://github.com/sourcegraph/sourcegraph/pull/62234

The original PR doesn't set `allowDefaults` which prevents populating input from entering the symbol which also was registered as a shortcut. Also there is a fix for catching field-like events 

## Test plan
- Check that `/` focuses the search box input on the home, result and blob UI pages
- Check that if the search input already got focus entering `/` leads to the adding `/` to the input value
- Check that if other input or form elements have focus entering `/` doesn't lead to the focusing search box input

### DIFF
--- a/client/web-sveltekit/src/lib/Hotkey.ts
+++ b/client/web-sveltekit/src/lib/Hotkey.ts
@@ -24,6 +24,20 @@ function isElement(t: any): t is Element {
 }
 
 /**
+ * This is an internal function to check if an Element has attributes that indicate it is a content field.
+ * It's exported for testing only.
+ * @param target Element
+ */
+function isContentElement(target: Element): boolean {
+    return (
+        target.getAttribute('contenteditable') === 'true' ||
+        // textarea and input are from the HTML standard, textbox is from svelte
+        ['textarea', 'input', 'textbox'].includes(target.getAttribute('role') ?? '') ||
+        ['INPUT', 'TEXTAREA'].includes(target.tagName)
+    )
+}
+
+/**
  * This function determines if the field that's focussed by the KeyboardEvent is some kind of input.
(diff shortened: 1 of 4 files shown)


bahrmichael (approved):

client/web-sveltekit/src/lib/Hotkey.ts
```diff
@@ -23,6 +23,20 @@ function isElement(t: any): t is Element {
     return t instanceof Element
 }
 
+/**
+ * This is an internal function to check if an Element has attributes that indicate it is a content field.
+ * It's exported for testing only.
+ * @param target Element
+ */
+function isContentElement(target: Element): boolean {
```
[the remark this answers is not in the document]
fkling: I guess that's to keep it consistent with `isContentField`, but maybe that should be renamed too.

client/web-sveltekit/src/lib/search/input/SearchInput.svelte
```diff
@@ -156,7 +156,19 @@
...
     registerHotkey({
         keys: { key: '/' },
-        handler: focus,
+        allowDefault: true,
+        handler: () => {
+            // If input already has focus, pass `/` symbol to the
+            // input element value
+            if (editor && editor.hasFocus) {
+                return true
+            }
```
fkling, line 165: This shouldn't be necessary, should it? The global keyboard handler should ignore the event. If that doesn't work, maybe we need to check whether the event originates *inside* a `contenteditable` element, not just whether the target element is one (but I'm not sure).

vovakulikov: Still don't know why bazel fails on editor has type never problem. If you will see it again let me know probably worth investigating

PR #62278, opened 2024-04-30
[Svelte]: Visual fixes for svelte blame panel

vovakulikov:
Adds a few visual tweaks to the blame panel 
- Fixes padding problem with gutters in blame UI
- Removes blame avatars
- Fixes width for dates (making it monospace)
- Make the blame panel a bit smaller by width (ideally we should make it resizable I think, made it smaller because it takes almost all width on my mid-size screen)

| Before | After |
| -------- | ------- |
| Screenshot 2024-04-30 at 13 53 28 | Screenshot 2024-04-30 at 13 52 48 |

 
## Test plan
- Check git blame panel, make sure it looks ok
- Check other code mirror editors like file preview panel

### DIFF
--- a/client/web-sveltekit/src/lib/CodeMirrorBlob.svelte
+++ b/client/web-sveltekit/src/lib/CodeMirrorBlob.svelte
@@ -56,7 +56,7 @@
             backgroundColor: 'transparent',
         },
         '.cm-gutters': {
-            'background-color': 'transparent',
+            'background-color': 'var(--code-bg)',
             border: 'none',
             color: 'var(--line-number-color)',
         },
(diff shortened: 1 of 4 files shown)


camdencheek (approved):
Thanks!

My only feedback is that I think it looks a little strange that the author name is monospace even though its color and weight match the message

vovakulikov:
> My only feedback is that I think it looks a little strange that the author's name is monospace even though its color and weight match the message

I will play with it later; I think this requires @taiyab touch; in my opinion, the date and author name are the first priority and should be highlighted as much as possible. Monospace, in this case just gives a bit bigger visual font size. Ideally, I think we should change the colours for the date and name to be standard black and not text-muted; another solution might be to colour the whole blame line with the most recent colour as most IDEs do. But again I'll leave it to @taiyab

camdencheek:
Yep, for sure. Certainly don't block on this. I think blame still needs a full design once-over. Thanks!

PR #62281, opened 2024-04-30
appliance: pod template standard config

craigfurman:
User deployments may need to set image pull secrets for private
registries, and/or node selectors/tolerations/affinities for use with
custom node pools. These are supported for all services in the helm
chart. Embed this in standard config so that all services get this for
free.

This is part of a stack, needs rebase when base branch merges.

## Test plan

Golden tests included.


jdpleiness (approved):
I really like this and I've got nothing to add! 🚀

PR #62286, opened 2024-04-30
appliance: default config

craigfurman:
## Test plan

This is a refactor: golden test fixtures have not changed (well, except config hashes, due to the changed structure of the defaults).

### DIFF
--- a/internal/appliance/defaults.go
+++ b/internal/appliance/defaults.go
@@ -1,15 +1,35 @@
 package appliance
 
+import (
+	"k8s.io/utils/ptr"
+
+	"github.com/sourcegraph/sourcegraph/internal/appliance/config"
+)
+
+// Default config.
+//
+// Warning: never extract `ptr.To(thing)` into a package-level variable! If you
+// do this, reconciling a config that overrides a default value for that
+// pointer, will affect the subsequent _default_ for all future resources
+// reconciled.
 func newDefaultConfig() Sourcegraph {
 	return Sourcegraph{
 		Spec: SourcegraphSpec{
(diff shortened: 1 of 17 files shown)


internal/appliance/spec.go
```diff
@@ -301,8 +296,6 @@ type SymbolsSpec struct {
 	Env map[string]string `json:"env,omitempty"`
 }
 
-func (SymbolsSpec) PrometheusPort() *int { return ptr.To(6060) }
```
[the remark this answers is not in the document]
jdpleiness: Yeah I'd be interested to take a closer look at it when you have some time.

internal/appliance/defaults.go
```diff
@@ -0,0 +1,38 @@
...
+
+	"github.com/sourcegraph/sourcegraph/internal/appliance/config"
+)
+
+// Default config.
+//
+// Warning: never extract `ptr.To(thing)` into a package-level variable! If you
+// do this, reconciling a config that overrides a default value for that
+// pointer, will affect the subsequent _default_ for all future resources
+// reconciled.
```
jdpleiness, line 14: This makes sense now that I think about it, but yeah I likely wouldn't have caught it 😬

jdpleiness (approved):

internal/appliance/defaults.go
craigfurman: Yeah, I don't love this comment, as it's a bit arbitrary - maybe we should put that everywhere we handle pointers - but it still surprised me in a golden test, so I figured I'd write this just in case.

PR #62287, opened 2024-04-30
Svelte: simplify file header

camdencheek:
Followup from https://github.com/sourcegraph/sourcegraph/pull/62246/files#r1583375607

Now that we don't style the file name differently than the rest of the path, this component can be simplified a bit.

## Test plan

Quick manual test.
screenshot-2024-04-30_08-15-34@2x

### DIFF
--- a/client/web-sveltekit/src/lib/shared.ts
+++ b/client/web-sveltekit/src/lib/shared.ts
@@ -92,11 +92,3 @@ export function displayRepoName(repoName: string): string {
     }
     return parts.join('/')
 }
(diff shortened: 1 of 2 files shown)


bahrmichael (approved):
LGTM. I slightly recall splitPath being used somewhere when I worked on the OpenInEditor, but if it works then that code is probably somewhere in the old react app.

camdencheek:
I couldn't find any other uses. If we need it again, there's always git history 🤷

PR #62289, opened 2024-04-30
svelte: Add infinity scroll and search to all branches page.

### LINKED ISSUE #61634: Paginate branches
The GraphQL connection available for this is not ideal: It returns an `endCursor` but the field doesn't accept an `after` cursor parameter. Instead you can only fetch more data by increasing the value of the `first` parameter, resulting in fetching unnecessary data.

fkling:
Closes #61634

Similar to #62257. I copied most of the code because I assume we'll redesign those pages anyway, so there is no point spending time to make these components reusable.

I implemented it such that the content is centered and the scrollbar is at the right edge of the screen. I added some padding so that there is some space between edge and content on smaller screens (I made the same change to the tags page).

I also updated the `GitReference` component to use `Badge` instead, which also affects the tags page.

| Before | After |
|--------|--------|
|  |  | 

Note: the table layout is not idea, I assume we'll redesign this anyway.

## Test plan

Manual testing.

- Scrolling/loading
- Back and forth navigation restores position and data
- Search works

### DIFF
--- a/client/web-sveltekit/src/lib/Scroller.svelte
+++ b/client/web-sveltekit/src/lib/Scroller.svelte
@@ -53,7 +53,7 @@
 </script>
 
 <div class="viewport" bind:this={viewport}>
-    <div class="scroller" bind:this={scroller} on:scroll={handleScroll}>
+    <div class="scroller" bind:this={scroller} on:scroll={handleScroll} data-scroller>
         <slot />
     </div>
 </div>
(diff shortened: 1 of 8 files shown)


bahrmichael (approved):

camdencheek:
On a larger screen, should we center the list?

screenshot-2024-04-30_08-47-23@2x

camdencheek (approved):

fkling:
Ah.. I think it was centered but then I made another change and didn't notice this. Will fix.

PR #62298, opened 2024-04-30
Svelte: Fix width of revision picker

taiyab:
# Context
The popover width keeps adjusting depending on the content width. This means it jumps around as you're trying to use it to search.

This resolves it from having varying widths.

## Test plan
Locally tested for visual change.


client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryRevPicker.svelte
```diff
@@ -189,6 +189,7 @@
         padding: 0.75rem;
         min-width: 35rem;
         max-width: 40rem;
+        width: 640px;
```
Comment (vovakulikov) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/RepositoryRevPicker.svelte:192]: I know that this is not enforced at the moment by linters in svelte but we always use `rem` unit for widths and sizes rules. Since rem standard size (which is 16px) can be override by users browsers settings all block sizes also should adjust to this (probably not relevant to spacing where we probably should use px)

vovakulikov (approved):
If we're setting width rule I would remove max/min width above since they are not working anymore, thanks for the change

PR #62308, opened 2024-04-30
Fix minor 'sg lint --help' issue(?)

chrsmith:
This PR is a collection of trivial cleanups. However, I have no idea what I'm doing here, so let me know if something is wrong with my approach.

Motivating problem:

When you run `sg lint --help`, the `format` linter is specified twice.
image

## Test plan

❓ I was able to run `bazel build`, but I couldn't track down how to build and run the new `sg` binary locally. (Any hints? I'll gladly update the repo's `README.md` file.)


dev/sg/VISION.md
```diff
@@ -1,3 +1,5 @@
 # `sg` Vision
 
-This page has moved to [`sg` vision](https://sourcegraph.com/docs/dev/background-information/sg/vision).
+This page has moved to [`sg` vision](https://docs.sourcegraph.com/dev/background-information/sg/vision).
```
chrsmith, line 3: The current link leads to a 404, but the current link has a warning butter bar:

"You're viewing our dev docs, which are in the process of being migrated. Last sync done on April 16th, for up to date content please read them from
[here](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/doc/dev/index.md)
  for the time being".

So should we just move the handbook page into the repo? Since the "vision" page didn't seem to get copied over to the new docs site. (But https://sourcegraph.com/docs/cli/ works.)

bobheadxi (approved):
Nice :)

dev/sg/linters/linters.go
```diff
@@ -191,10 +193,3 @@ func bazelExec(name, args string) *linter {
 func pnpmInstallFilter() pipeline.Pipeline {
 	return pipeline.Filter(func(line []byte) bool { return !bytes.Contains(line, []byte("warning")) })
 }
-
-// disabled can be used to mark a category or check as disabled.
-func disabled(reason string) check.EnableFunc[*repo.State] {
```
chrsmith, line 196: Perhaps ironically, VS Code presented a warning that this function was unused... so perhaps the Go linter isn't hooked up to this file or something?

dev/sg/VISION.md
bobheadxi: I think @jhchabran has something WIP for dev docs

bobheadxi: let's leave this as-is so he can do a broad find-replace for all these links later

dev/sg/sg_lint.go
```diff
@@ -132,7 +131,7 @@ sg lint --help
 		std.Out.WriteNoticef("Running checks from targets: %s", strings.Join(targets, ", "))
 		return runner.Check(cmd.Context, repoState)
 	},
-	Subcommands: lintTargets(append(linters.Targets, linters.Formatting)).Commands(),
+	Subcommands: lintTargets(linters.Targets).Commands(),
```
chrsmith, line 134: I suspect that this is the root cause of the main issue. Since we already have `linters.Formatting` included in the `linters.Targets` slice.

Perhaps that is why `format` is printed twice?

doc/dev/background-information/sg/reference.md
```diff
@@ -1,5 +1,3 @@
-<!-- DO NOT EDIT: generated via: go generate ./dev/sg -->
```
chrsmith, line 1: For the life of me I couldn't figure out how to regenerate this file. Searching for things like `//go:generate` didn't find many hits. And `sg generate` ended up with a wall of errors for seemingly unrelated things.

Happy to update the docs, comments, whatever to update _how_ to generate this file. But `go generate ./dev/sg` no longer works.

bobheadxi: This was (IMO sadly) removed while back - I proposed bringing it back because being able to surface help in a published format/search was nice, but we ended up not doing it. IMO this file can be deleted entirely based on that decision, see discussion here: https://github.com/sourcegraph/sourcegraph/pull/58348

chrsmith: Just to confirm, you are saying the correct way to update this file would be to run:

```zsh
REF_PATH="doc/dev/background-information/sg/reference.md"

echo '<!-- DO NOT EDIT' >  ${REF_PATH}
echo 'Generated via:'   >> ${REF_PATH}
echo 'sg help --full > doc/dev/background-information/sg/reference.md' >> ${REF_PATH}
echo '-->'              >> ${REF_PATH}
sg help --full          >> ${REF_PATH}
```

I've added the changes, which leads to a significant diff FWIW.

dev/sg/VISION.md
chrsmith: Ack. I'll back out this change then.

doc/dev/background-information/sg/reference.md
chrsmith: Cool. I'll just delete this file then.

camdencheek: TIL! @bobheadxi is the expert here, ignore me 👍

PR #62310, opened 2024-04-30
Misc: Add #team-cloud-ops slack webhook to release.yaml

Chickensoupwithrice:
Resolves [Michael's request](https://sourcegraph.slack.com/archives/C05DWT4ANHH/p1714420855027439)

## Test plan

Assumed valid since copy pasted

### DIFF
--- a/release.yaml
+++ b/release.yaml
@@ -31,6 +31,9 @@ requirements:
 
    # bolaji-release-testing
    # https://start.1password.com/open/i?a=HEDEDSLHPBFGRBTKAKJWE23XX4&v=isv4abynddpox72wwbhaamo76e&i=7zjax5rm5hlilbgrzeb257i62i&h=team-sourcegraph.1password.com
+
+   # team-cloud-ops salck webhook url:
+   # https://start.1password.com/open/i?a=HEDEDSLHPBFGRBTKAKJWE23XX4&v=dnrhbauihkhjs5ag6vszsme45a&i=xwwpaakz5iajowr2fruueoevba&h=team-sourcegraph.1password.com
   - name: 'Slack Webhook URL'
     env: SLACK_WEBHOOK_URL
   - name: 'Slack Webhook URL Cloud Ops'
--- a/release.yaml
+++ b/release.yaml
@@ -288,38 +288,22 @@ promoteToPublic:
           echo "Posting slack notification for release"
           tag="{{tag}}"
           changelog_version="${tag//./}"
-          body=$(wget --content-on-error -O- --header="Content-Type: application/json" --post-data '{
-            "type": "mrkdwn",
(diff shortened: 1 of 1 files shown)


release.yaml
```diff
@@ -31,8 +31,13 @@ requirements:
 
    # bolaji-release-testing
    # https://start.1password.com/open/i?a=HEDEDSLHPBFGRBTKAKJWE23XX4&v=isv4abynddpox72wwbhaamo76e&i=7zjax5rm5hlilbgrzeb257i62i&h=team-sourcegraph.1password.com
+
+   # team-cloud-ops salck webhook url:
+   # https://start.1password.com/open/i?a=HEDEDSLHPBFGRBTKAKJWE23XX4&v=dnrhbauihkhjs5ag6vszsme45a&i=xwwpaakz5iajowr2fruueoevba&h=team-sourcegraph.1password.com
   - name: 'Slack Webhook URL'
     env: SLACK_WEBHOOK_URL
+  - name: 'Slack Webhook URL Cloud Ops'
```
BolajiOlajide, line 39: Can this be provided in 1password so whoever is running the release can have access to it?

If it's already in 1password, can you add a link to it here for reference?

release.yaml
```diff
@@ -289,6 +294,26 @@ promoteToPublic:
           }' "$SLACK_WEBHOOK_URL")
           exit_code=$?
 
+          if [ $exit_code != 0 ]; then
```
BolajiOlajide, line 297: Since the step is basically the same as above. How about we add the webhook URLs in an array and just loop through each of them to make the call.

This ensures this YAML isn't unnecessary long and makes it easier to read also.

BolajiOlajide (changes requested):

release.yaml
Chickensoupwithrice: It's added here :) 
https://github.com/sourcegraph/sourcegraph/pull/62310/commits/76e4c7a6902d79abec60bdcc3485094e8eeae42f#diff-a0b29149f10adefb5c44814777ca3aa3c992d81fd92a1a051add31fab99452fcR36

BolajiOlajide: Oh i see it now. It's easy to confuse it with the others, do you mind moving it just above the check for the env var for Cloud ops?

release.yaml
```diff
@@ -31,8 +31,13 @@ requirements:
 
    # bolaji-release-testing
    # https://start.1password.com/open/i?a=HEDEDSLHPBFGRBTKAKJWE23XX4&v=isv4abynddpox72wwbhaamo76e&i=7zjax5rm5hlilbgrzeb257i62i&h=team-sourcegraph.1password.com
```
BolajiOlajide, line 33:
```suggestion
```

I think we can safely remove this, we wouldn't be doing any slack testing for a while.

BolajiOlajide (approved):

michaellzc (approved):

PR #62311, opened 2024-04-30
Move cody packages only used by frontend into cmd/frontend

eseliger:
The httpapi package for completions APIs and the internal/cody and internal/ssc packages are only used in frontend, so moving them there.

This cleans up the massive /internal package a bit further, and is part of a larger effort to make less calls to conf.Get() inside /internal, because those are hard to trace down and can potentially brick services like cody-gateway when they are ever in any call paths.

Test plan:

Just moved code around, and the go compiler doesn't complain.


chrsmith (approved):
Unfortunately I totally just merged some PRs that will add even more merge conflicts for you 😅 . But as far as the intent of this PR, this sounds good to me. I am unaware of any usages of `internal/{cody, ssc}` outside of what we need from the Sourcegraph.com backend (AKA `cmd/frontend`). So AFAIK this is safe to do, as long as the Go compiler and CI/CD is happy.

PR #62314, opened 2024-04-30
Svelte: fix icon sizing

camdencheek:
This fixes an issue with the sizing of the `Icon` component. There was a `size` prop, but it wasn't actually being respected. Additionally, it stops relying on the global `.icon-inline` class.

This changes the `Icon` component to use only CSS variables to set its size. By default, it uses the `--icon-size` variable. If in `inline` mode, it uses the `--icon-inline-size` variable. 

## Test plan

Before:
screenshot-2024-04-30_14-55-40@2x

After:
screenshot-2024-04-30_14-56-19@2x

### DIFF
--- a/client/web-sveltekit/src/lib/Icon.svelte
+++ b/client/web-sveltekit/src/lib/Icon.svelte
@@ -18,21 +18,25 @@
 
     export let svgPath: string
     export let inline: boolean = false
-    export let size: number = 24
 </script>
 
-<svg class:icon-inline={inline} height={size} width={size} viewBox="0 0 24 24" data-icon {...$$restProps}>
+<svg class:icon-inline={inline} viewBox="0 0 24 24" data-icon {...$$restProps}>
     <path d={svgPath} />
 </svg>
 
 <style lang="scss">
     :root {
         --icon-size: 1.5rem;
+        --icon-inline-size: #{(16 / 14)}em;
     }
 
     svg {
         width: var(--icon-size);
         height: var(--icon-size);
+        &.icon-inline {
+            width: var(--icon-inline-size);
+            height: var(--icon-inline-size);
+        }
         color: var(--icon-fill-color, var(--color, inherit));
         fill: currentColor;
     }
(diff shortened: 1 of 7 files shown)


fkling:
> [...] but nothing actually used that class to change the styling.

There is a global `.inline-icon` class that we import from the react app (as part of global styles) so that should work. But I also prefer it to be explicit.

client/web-sveltekit/src/lib/Icon.svelte
```diff
@@ -18,21 +18,25 @@
 
     export let svgPath: string
     export let inline: boolean = false
-    export let size: number = 24
 </script>
 
-<svg class:icon-inline={inline} height={size} width={size} viewBox="0 0 24 24" data-icon {...$$restProps}>
+<svg class:icon-inline={inline} viewBox="0 0 24 24" data-icon {...$$restProps}>
```
fkling, line 23: Are you sure `width` and `height` are not needed (I really don't know 😅 ).

I'd just make sure that all the other icons still look good.

camdencheek: > Any height or width you set for the SVG with CSS will override the height and width attributes on the <svg>.

This was always being overridden by the CSS styling, so it ended up completely unused

camdencheek:
> There is a global .inline-icon class

Ooh, for some reason a search was not finding this. I see it now, thanks!

camdencheek:
also, how did  you find this PR? I didn't request a review yet 😄

fkling:
> also, how did you find this PR? I didn't request a review yet 😄

I'm subscribed to all PRs via the Slack GitHub integration :)

client/web-sveltekit/src/lib/Icon.svelte
```diff
@@ -18,21 +18,23 @@
...
     <path d={svgPath} />
 </svg>
 
 <style lang="scss">
-    :root {
-        --icon-size: 1.5rem;
-    }
+    $iconSize: var(--icon-size, 1.5rem);
+    $iconInlineSize: var(--icon-inline-size, #{(16 / 14)}em);
```
camdencheek, line 29: I moved this off `:root` because I think we should generally avoid allowing things other than the `Icon` component to use the default `--icon-size` variable.

fkling (approved):
I suspect this is not the last time we'll work on icon sizes but at least this reduces the API to just CSS. Thank you!

PR #62316, opened 2024-04-30
svelte: Update some aspects diff view and history panel

### LINKED ISSUE #62290: Fix history panel links and wording for folder pages

fkling:
Closes #62290

"View at commit" on a directory page didn't work at all. I first added support for it but it become clear that it's not obvious how such a view would behave when clicking on a directory entry. So I decided to remove it from this view, for now, and avoid the whole issue.

I don't know whether enabling the history panel features  via `$page.data.X` is a good idea or not, but it seemed better than explicitly testing for the route.

I also moved the "Close x" action to the info header (after talking to Taiyab). No idea whether it should be styled as a link or not but we can iterate on that.

## Test plan

Local visual inspection.


vovakulikov (approved):

vovakulikov:
Merging this one since the branch cut is going to be tomorrow, good to have time to test it on staging as well.

Issue #62306, opened 2024-04-30
Svelte: Remove "See more details" label


taiyab:
# Context

This is our current results summary widget.

<img width="418" alt="CleanShot 2024-04-30 at 19 53 15@2x" src="https://github.com/sourcegraph/sourcegraph/assets/5337876/029943ce-de4d-4128-863c-eb54965f9d79">

We should eliminate the label that says "See more details" to save vertical space. You should only get a notice there once of the conditions are met for an incomplete search, or whatever other states we have for that.

camdencheek:
@taiyab If I understand how we got here correctly, we added the "see more details" because if there was nothing in that space, the page jumps when a search goes from showing a message to showing nothing. I don't think we want the search results to shift when a search completes. Thoughts on how to get around that?

PR #62317, opened 2024-04-30
Svelte: compress progress indicator

camdencheek:
This removes some extra padding from the progress indicator and also simplifies some of the padding so it's symmetrical and doesn't rely on adding padding together to get the right layout. 

## Test plan

Before: 
screenshot-2024-04-30_16-58-19@2x

After: 
screenshot-2024-04-30_16-58-03@2x

### DIFF
--- a/client/web-sveltekit/src/lib/LoadingSpinner.svelte
+++ b/client/web-sveltekit/src/lib/LoadingSpinner.svelte
@@ -8,8 +8,6 @@
 </div>
 
 <style lang="scss">
-    $size: var(--size, 1rem);
-
     .center {
         display: flex;
         flex-direction: column;
@@ -28,12 +26,11 @@
             --loading-spinner-inner-color: var(--white);
         }
 
-        width: $size;
-        height: $size;
+        width: var(--size, 1rem);
+        height: var(--size, 1rem);
         &.inline {
-            $inlineSize: #{(16 / 14)}em;
-            width: $inlineSize;
-            height: $inlineSize;
+            width: #{(16 / 14)}em;
+            height: #{(16 / 14)}em;
 
             vertical-align: bottom;
             display: inline-flex;
--- a/client/web-sveltekit/src/lib/LoadingSpinner.svelte
+++ b/client/web-sveltekit/src/lib/LoadingSpinner.svelte
@@ -4,10 +4,12 @@
 </script>
 
 <div class:center>
-    <div class="loading-spinner" class:icon-inline={inline} aria-label="loading" aria-live="polite" />
+    <div class="loading-spinner" class:inline aria-label="loading" aria-live="polite" />
 </div>
 
 <style lang="scss">
+    $size: var(--size, 1rem);
+
     .center {
         display: flex;
         flex-direction: column;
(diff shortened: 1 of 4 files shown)


fkling:
Maybe you want to tackle the rest of https://github.com/sourcegraph/sourcegraph/issues/62306 as part of this PR?

client/web-sveltekit/src/routes/search/StreamingProgress.svelte
```diff
@@ -163,7 +163,7 @@
     .progress-button {
         border: 1px solid var(--border-color-2);
         border-radius: 4px;
-        margin-left: 0.3rem;
+        padding: 0;
```
camdencheek, line 166: Disable the default button padding so we can set it consistently without having to add paddings together

client/web-sveltekit/src/lib/search/resultsIndicator/ResultsIndicator.svelte
```diff
@@ -59,29 +53,25 @@
...
 <style lang="scss">
     .indicator {
         display: flex;
         flex-flow: row nowrap;
-        justify-content: space-evenly;
+        justify-content: space-between;
         align-items: center;
-        gap: 0.5rem;
-        min-width: 200px;
-        max-width: fit-content;
```
camdencheek, line 73: Got rid of the min-width and max-width because this looks better when the spacing is balanced.

client/web-sveltekit/src/lib/search/resultsIndicator/ResultsIndicator.svelte
```diff
@@ -59,29 +53,25 @@
...
+
+    <Icon svgPath={mdiChevronDown} --icon-size="12px" --color={isError ? 'var(--danger)' : 'var(--text-title)'} />
 </div>
 
 <style lang="scss">
     .indicator {
         display: flex;
         flex-flow: row nowrap;
-        justify-content: space-evenly;
+        justify-content: space-between;
```
camdencheek, line 64: No functional difference, but since we are fitting to content, space-between is more semantically correct

client/web-sveltekit/src/lib/search/resultsIndicator/ResultsIndicator.svelte
```diff
@@ -37,17 +37,11 @@
 </script>
 
 <div class="indicator">
-    <div>
```
camdencheek, line 40: Unnecessary outer div.

client/web-sveltekit/src/lib/LoadingSpinner.svelte
```diff
@@ -26,9 +28,18 @@
...
+        height: $size;
+        &.inline {
+            $inlineSize: #{(16 / 14)}em;
+            width: $inlineSize;
+            height: $inlineSize;
+
+            vertical-align: bottom;
+            display: inline-flex;
+            align-items: center;
+        }
```
camdencheek, line 41: Added the ability to specify the size of the spinner and moved the inline styling into here because the `.icon-inline` styles expect to be styling an SVG.

PR #62322, opened 2024-04-30
Migrate "New Cody Subscription" page

chrsmith:
This PR migrates the "New Cody Subscription" page from accounts.sourcegraph.com to sourcegraph.com, with it being served at `/cody/manage/subscription/new`. (My next step is to add the "Manage Subscription" page at `/cody/manage/subscription`, to support `/cody/manage/team` or other things down the road.)

The React code is pretty much a copy of what is in the `sourcegraph/self-serve-cody` repo. With some sg/sg specific idioms added. (e.g. wrapping the component in `withAuthorizedUser(...)` so it will automatically redirect the user to login if needed.)

A few things of note:

- The page will only get registered IFF the site configuration has the `codyProConfig` setting available. (So this won't be enabled on sourcegraph.com, or be visible on enterprise installations.)
- The page also requires that the calling user has the `cody-enable-embedded-cody-pro-ui` feature flag.
- The `stripe-js` dependency was added to the `client/web/package.json`. My understanding is that because the route handler is loaded lazily, this won't require any new downloads when the browser doesn't need the Stripe bits. But the general thinking is that we will try to isolate the Cody Pro-specific UI into its own NPM package, to better isolate it from the "regular Cody Pro" UI. And any other Cody Enterprise-specific UI that we have in the future. But there is a bit too much ambiguity in what we want there, to do that splitting/refactoring just yet.

## Next Steps

With these changes you can complete the new Cody Pro subscription checkout process. But we will need to update the `/cody/manage` page to support cases where the subscription was _just_ created. (I'll sync up with Rob about maybe having an intersitular page, like we have with `/cody/welcome` on accounts.sourcegraph.com today. Or maybe we'll just add a "toast" notification or something from `/cody/manage`.)

But other than that, the next (and final) step of migrating the Cody Pro UI to sourcegraph.com is just migrating the "Manage Subscription" page..

## Test Plan

I tested things manually, but do plan on adding some proper automated tests for the react bits. However, there isn't much we can do for the checkout experience specifically, since the UI is all owned by Stripe and served within an iframe. (So there isn't really any way to configure Storybook to do that in CI/CD.)


client/web/package.json
```diff
@@ -46,6 +46,8 @@
     "@sourcegraph/schema": "workspace:*",
     "@sourcegraph/shared": "workspace:*",
     "@sourcegraph/telemetry": "^0.11.0",
-    "@sourcegraph/wildcard": "workspace:*"
+    "@sourcegraph/wildcard": "workspace:*",
+    "@stripe/react-stripe-js": "^2.7.0",
+    "@stripe/stripe-js": "^3.3.0"
```
eseliger, line 51: I saw your comment in slack on this, but I still think that we shouldn't add the dependency here, and in /package.json instead.

There are a lot of dependencies in there already that are either web, shared, or something else only, but imported to all of them. 

That prevents that we have to make sure not to have two objects of the same name, but no actual equality because they're from different versions.

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
```diff
@@ -0,0 +1,111 @@
...
+        // createCheckoutSession and let it run async.
+        // (And not `await createCheckoutSession` or `return createCheckoutSession`.)
+        void createCheckoutSession('monthly', showPromoCodeField, customerEmail, setClientSecret, setErrorDetails)
+    }, [customerEmail, showPromoCodeField])
+
+    const embeddedCheckoutOpts /* unexported EmbeddedCheckoutProviderProps.options */ = {
+        clientSecret,
+    }
+    return (
+        <div id="checkout">
```
eseliger, line 38: do we need the `id` property here? I don't see a reference to it, but might be blind. Generally, in react using `refs` is the way to reference a rendered object, where possible.

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
```diff
@@ -0,0 +1,111 @@
...
+            {errorDetails && (
+                <div>
```
eseliger, line 40: Is this div needed for styling purposes? Otherwise, react allows empty (called a fragment) which won't be part of the rendered output, to not clutter the DOM unnecessarily:

```suggestion
                <>
```

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
```diff
@@ -0,0 +1,111 @@
...
+): Promise<void> {
+    // e.g. "https://sourcegraph.com"
+    const origin = window.location.origin
+
+    try {
+        // So the request is kinda made to 2x backends. dotcom's .api/ssc/proxy endpoint will
+        // take care of exchanging the Sourcegraph session credentials for a SAMS access token.
+        // And then proxy the request onto the SSC backend, which will actually create the
+        // checkout session.
+        const resp = await fetch(`${origin}/.api/ssc/proxy/checkout/session`, {
```
eseliger, line 78: I'm surprised we don't have a helper that wraps fetch nicely and handles auth etc, but I couldn't find one in our codebase, so 🤷

client/web/src/cody/management/subscription/new/NewCodyProSubscriptionPage.module.scss
```diff
@@ -0,0 +1,23 @@
...
+        > div {
+            border: none !important;
+            align-items: center;
+        }
+    }
+}
+
+.container {
+    border-radius: 6px;
+    border: 1px solid (--border-color);
```
eseliger, line 17: We have a container component that has a standardized look, could we use that, even if it's slightly different than this one?

client/web/src/cody/management/subscription/new/NewCodyProSubscriptionPage.tsx
```diff
@@ -0,0 +1,86 @@
...
+    telemetryRecorder,
+}) => {
+    useEffect(() => {
+        telemetryRecorder.recordEvent('cody.new-subscription-checkout', 'view')
+    }, [telemetryRecorder])
+
+    // If the user already has a Cody Pro subscription, direct them back to the Cody Management page.
+    const { data, error: dataLoadError } = useQuery<UserCodyPlanResult, UserCodyPlanVariables>(USER_CODY_PLAN, {})
+    if (dataLoadError) {
+        throw dataLoadError
```
eseliger, line 49: do you want to just throw here, or render a nicer error message? The default error handler looks quite dull for a B2C application 😬

client/web/src/cody/management/subscription/new/NewCodyProSubscriptionPage.tsx
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    const stripeElementsAppearance: stripeJs.Appearance = {
+        theme: 'stripe',
+        variables: {
+            colorPrimary: '#00b4d9',
+        },
+    }
+    return (
+        <>
```
eseliger, line 62: This outer fragment seems unnecessary

client/web/src/cody/management/subscription/new/NewCodyProSubscriptionPage.tsx
```diff
@@ -0,0 +1,86 @@
...
+                <PageTitle title="New Subscription" />
+                <PageHeader className="mb-4 mt-4">
+                    <PageHeader.Heading as="h2" styleAs="h1">
+                        <div className="d-inline-flex align-items-center">
+                            <CodyProIcon className="mr-2" /> Give your team Cody Pro
+                        </div>
+                    </PageHeader.Heading>
+                </PageHeader>
+
+                <div className={classNames('p-4 border bg-1 mt-4', styles.container)}>
```
eseliger, line 73: Instead of crafting a Container component here from various classnames, could we use a proper `<Container>`?

client/web/src/cody/management/subscription/new/NewCodyProSubscriptionPage.tsx
```diff
@@ -0,0 +1,86 @@
...
+                    <Elements stripe={stripePromise} options={{ appearance: stripeElementsAppearance }}>
+                        <CodyProCheckoutForm
+                            stripeHandle={stripePromise}
+                            customerEmail={authenticatedUser?.emails[0].email || ''}
```
eseliger, line 77: Is empty string a valid configuration, or should we error out instead?

schema/cody_pro_util.go
eseliger, line 1: I would inline these types in `cmd/frontend/internal/app/jscontext/jscontext.go`, it looks like this code has nothing to do with the schema generator that creates most contents in this package

client/web/src/routes.tsx
```diff
@@ -419,6 +424,23 @@ export const routes: RouteObject[] = [
...
+                render={props => (
+                    <NewCodyProSubscriptionPage
+                        authenticatedUser={props.authenticatedUser}
+                        telemetryRecorder={props.platformContext.telemetryRecorder}
+                    />
+                )}
+                condition={({ isSourcegraphDotCom }) => {
+                    const [embeddedUiEnabled] = useFeatureFlag('cody-enable-embedded-cody-pro-ui', false)
+                    return isSourcegraphDotCom && embeddedUiEnabled
+                }}
```
eseliger, line 440: Does that work? I would think condition is only evaluated once and while the feature flag state hasn't been loaded yet from the server, it would default to `false` (and the second entry in the return array would indicate it's still loading).

eseliger (commented):
Left a few questions and suggestions inline, I don't see any big issues with this change, and love that we get to use Wildcard here now.

client/web/package.json
chrsmith: I am 100% in agreement that we want to move this new dependency into the narrowest scope possible. So if `/package.json` is the right place to put it, awesome!

But for my own understanding, what's the difference between `client/web/package.json` and `/package.json`? Both have the description "The Sourcegraph web app", and I would have assumed that the "more general one" would be in the root of the repository. And that the `package.json` file just for the Sourcegraph web app would be in `client/web`... but I take it that isn't the case?

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
chrsmith: Apparently we don't? The Stripe Checkout examples have the `id="checkout"` attribute. But I couldn't find any reason _why_ in the documentation, and things seem to work just fine without it.

So I'll remove it and if that somehow was "load bearing", we can add it back later. (But I wouldn't be surprised it was to make it easier for Stripe's JS library to know where to put the embedded `iframe` for rendering the embedded UI.)

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
chrsmith: Not needed for styling, will change it to use a fragment. 👍

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
chrsmith: Well, don't worry. We will have a helper that wraps `fetch` nicely and handles auth... since we'll be adding it a follow-up PR as we add more instances where we need to call the SSC backend through this `.api/ssc/proxy` endpoint.

client/web/src/cody/management/subscription/new/NewCodyProSubscriptionPage.module.scss
chrsmith: This was 100% cargo culted from the `CodyManagementPage.module.scss`. There is no reason to use a slightly different one. So I'll actually just remove it entirely, since removing the `.module.scss` file entirely doesn't seem to make any changes, and we only were using the `.container` property.

https://github.com/sourcegraph/sourcegraph/blob/main/client/web/src/cody/management/CodyManagementPage.module.scss#L15

client/web/src/cody/management/subscription/new/NewCodyProSubscriptionPage.tsx
chrsmith: For just getting this PR checked-in so we can iterate on it, I don't really care throwing the error.

However, I am interested in trying to catch and render a nicer error message. Can you point me to a place in the codebase where we have a better error message or something?

client/web/src/cody/management/subscription/new/NewCodyProSubscriptionPage.tsx
chrsmith: Good catch, done.

client/web/src/cody/management/subscription/new/NewCodyProSubscriptionPage.tsx
chrsmith: Empty string is valid configuration. (It just means that we don't auto-populate the email address on the rendered form.)

client/web/package.json
chrsmith: Also, should I ignore this warning? Sounds scary. But then again, I don't quite know what we use the root `package.json` file for.

```
% pnpm install @stripe/sourcegraph-js
 ERR_PNPM_ADDING_TO_ROOT  Running this command will add the dependency to the workspace root, which might not be what you want
- if you really meant it, make it explicit by running this command again with the -w flag (or --workspace-root). If you don't
- want to see this warning anymore, you may set the ignore-workspace-root-check setting to true.
```

vdavid: I'd be curious about the reason for not currently having any non-workspace dependencies in `client/web/package.json`. I don't know about any particular best practice that would oppose it, but I'm sure we have _some_ reason, and our current use case doesn't seem that special compared to what we normally do. Were people just lazy and put everything to `/package.json`? I kinda doubt it.

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
```diff
@@ -0,0 +1,111 @@
...
+    const showPromoCodeField = urlSearchParams.get('showCouponCodeAtCheckout') !== null
+
+    // Issue an API call to the backend asking it to create a new checkout session.
+    // This will update clientSecret/errorDetails asynchronously when the request completes.
+    useEffect(() => {
+        // useEffect will not accept a Promise, so we call
+        // createCheckoutSession and let it run async.
+        // (And not `await createCheckoutSession` or `return createCheckoutSession`.)
+        void createCheckoutSession('monthly', showPromoCodeField, customerEmail, setClientSecret, setErrorDetails)
+    }, [customerEmail, showPromoCodeField])
```
vdavid, line 32: It looks like the `setClientSecret` and `setErrorDetails` dependencies are missing from this `useEffect` call 🤔 
Did you manage to run this? Maybe it only causes a linter error, I'm not sure, my IDE pretty much keeps this list auto-updated.

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+    const embeddedCheckoutOpts /* unexported EmbeddedCheckoutProviderProps.options */ = {
```
vdavid, line 34: Nit: "Opts" is pretty Go-ish. "Options" is idiomatic.

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
vdavid: Also nit: "resp" is pretty Go-ish. "response" is idiomatic. (If you don't believe me, search our front-end codebase, "response" wins by a landslide)

client/web/src/routes.tsx
vdavid: Also, I don't think you can (or even should be allowed to) use React hooks outside of components. Even if React/TS allows you, this looks like an anti-pattern. I'd just rely on the config and leave it at that.

client/web/package.json
eseliger: Having multiple versions of the same dependency is valid in JS land, but can cause issues like `instanceof` checks suddenly failing. It can also cause the bundle to become much bigger because you suddenly import multiple versions.

I do agree with yall that we should probably use packages by workspace and isolate those more, but since we don't do that today, for consistency I think we should add to the root package. I am not sure why pnpm complains about that though, this is how we always did it :/

client/web/src/cody/management/subscription/new/NewCodyProSubscriptionPage.tsx
eseliger: Instead of throwing you can return a just yet another component here that represents the error state; one that we have is the `<ErrorAlert>`, but depending on your use-case you can go completely wild and use all that wildcard has to offer :)

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
eseliger: I think it's fine to leave setters out of there because they are known to be immutable. But that requires to know it's a setState setter so I bet that's why autocomplete for that won't consider it.

schema/cody_pro_util.go
eseliger: Oh I didn't mean literal inlining (sorry - this term is overloaded), I meant just moving the code in this file at the end of jscontext.go :)

client/web/package.json
chrsmith: > I do agree with yall that we should probably use packages by workspace and isolate those more, but since we don't do that today, for consistency I think we should add to the root package ... this is how we always did it :/

Sounds good. I'll defer to your expertise here.

client/web/src/routes.tsx
chrsmith: > Does that work?

It certainly works locally 🤷 . But if we "should" do this, is another question entirely.

I'm happy to remove it. But in case you had forgotten David, I literally added this feature flag because you and Naman insisted upon it. I wanted to have these pages be gated on configuration changes all along. Read: y'all wasted my time.

Relying on the feature flag _within_ the react component/page is insufficient, because unless these conditions are met we don't want the page to even load. (And not, try to load, but report an error or render a blank page instead.)

chrsmith: 🤷 No, just running locally. So it could be that things happen to be "so fast" I'm not seeing this failure. Or, perhaps it is totally broken, but I'm just not noticing because the page reloads to quickly or something?

But I 100% believe you that requiring some sort of async-data within the URL routing is a bad idea. So I'm happy to make this change.

chrsmith:
@vdavid , @eseliger PTAL. In particular, are there any other important things that we should gate on checking this PR in?

If there are minor things I'd be happy to address that in a smaller follow-up PR.

client/web/src/routes.tsx
vdavid: > unless these conditions are met we don't want the page to even load

@chrsmith Why is that? I don't see any harm done if we start loading the page and then show an error or a blank page while this feature is unannounced. Do you see a scenario where UX suffers because you move the check from the routing layer to the Page component?

client/web/src/cody/util.ts
```diff
@@ -1,2 +1,17 @@
...
+ * If false, we rely on the current behavior. Where users are directed to https://accounts.sourcegraph.com/cody
+ * for managing their Cody Pro subscription information.
+ */
+export function useEmbeddedCodyProUi(): boolean {
+    const codyProConfig = window.context.frontendCodyProConfig
+    if (codyProConfig && codyProConfig.stripePublishableKey) {
+        return true
+    }
+    return false
+}
```
vdavid, line 17: Nit: I'd simplify this to:
```suggestion
export const useEmbeddedCodyProUi = (): boolean =>
    !!(window.context.frontendCodyProConfig && window.context.frontendCodyProConfig.stripePublishableKey)
```

vdavid (approved):
Ship it!

client/web/src/routes.tsx
chrsmith: > Do you see a scenario where UX suffers because you move the check from the routing layer to the Page component?

Yes, this is very much impacts the user experience.

If the router doesn't recognize the URL path, then it serves a pretty nice 404 error page. If we pass that responsibility onto the component then all bets are off.

Maybe the component is built "correctly", and renders the same 404 page. So the UX is identical. Maybe it doesn't, so some 404 pages look weird. Or perhaps the component doesn't handle this well at all, and just throws an error.

Ultimately, if we remove the requirement here from the specific components (so it's one fewer thing they need to implement in order to be "correct"), and put that logic into the router. Then the overall system is easier to reason about and maintain.

PR #62323, opened 2024-05-01
svelte: Fix code/blame view for rich HTML files

fkling:
PR #62187 introduced a regression that causes the code and blame views for Markdown files to not show.

This commit fixes this. I renamed the variables to be a bit more clear and removed the check for `showCodeView`. Instead this relies on the template handling all other cases (binary, formatted, etc) first. Showing the file contents is now basically the fallback mode.

## Test plan

Manual testing. Verified that hover cards show up.


camdencheek (approved):
LGTM! The logic looks a little easier to follow too

PR #62328, opened 2024-05-01
Update wording on account mismatch for Cody PLG Manage page

toolmantim:
Small suggested change by @unknwon that slipped through in #62039. Also an update the comment about the URL param now that support for it has been merged to Jetbrains too.

## Test plan

- CI


unknwon (commented):
Thank you :D

PR #62331, opened 2024-05-01
chore: Run bazel configure

varungandhi-src:
Not sure why this wasn't caught by the CI check?
Have we stopped running this?

## Test plan

Run `bazel configure`


varungandhi-src:
Closing; user error. I should've run `sg bazel configure` instead.

PR #62335, opened 2024-05-01
chore: Group upload matching options into struct

varungandhi-src:
There are lots of layers of indirection/plumbing for propagating
the various options/settings for matching uploads. There was also
a mix of terminology - 'toolName' in some places and 'indexer'
in some other places. 'exactPath' in some places and 'rootMustEnclosePath'
in other places. Grouping the various options into a struct provides
a central place to document how the path matching happens,
as well as special arguments such as `shared.SyntacticIndexer`.

This also makes it much easier to understand where the various
upload matching options are used via 'Find usages' in the editor.

As a nice side-effect, the noise of creating o11y-related K-V pairs
is also reduced.

## Test plan

Covered by existing tests

### DIFF
--- a/internal/codeintel/uploads/internal/store/commitgraph.go
+++ b/internal/codeintel/uploads/internal/store/commitgraph.go
@@ -274,9 +274,7 @@ LIMIT %s
 `
 
 // FindClosestCompletedUploads returns the set of uploads that can most accurately answer queries for the given repository, commit, path, and
-// optional indexer. If rootMustEnclosePath is true, then only dumps with a root which is a prefix of path are returned. Otherwise,
-// any dump with a root intersecting the given path is returned.
-// Syntactic indexes are not returned unless the requested indexer is shared.SyntacticIndexer
+// optional indexer.
 //
 // This method should be used when the commit is known to exist in the lsif_nearest_uploads table. If it doesn't, then this method
 // will return no dumps (as the input commit is not reachable from anything with an upload). The nearest uploads table must be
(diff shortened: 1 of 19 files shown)


internal/codeintel/uploads/internal/store/commitgraph.go
```diff
@@ -295,19 +293,13 @@ LIMIT %s
...
-	}})
+func (s *store) FindClosestCompletedUploads(ctx context.Context, opts shared.UploadMatchingOptions) (_ []shared.CompletedUpload, err error) {
+	ctx, trace, endObservation := s.operations.findClosestCompletedUploads.With(ctx, &err, observation.Args{Attrs: opts.Attrs()})
 	defer endObservation(1, observation.Args{})
 
-	conds := makeFindClosestProcessUploadsConditions(path, rootMustEnclosePath, indexer)
+	conds := makeFindClosestProcessUploadsConditions(opts)
 	// TODO(id: completed-state-check) Make sure we only return uploads with state = 'completed' here
-	query := sqlf.Sprintf(findClosestCompletedUploadsQuery, makeVisibleUploadsQuery(repositoryID, commit), sqlf.Join(conds, " AND "))
+	query := sqlf.Sprintf(findClosestCompletedUploadsQuery, makeVisibleUploadsQuery(opts.RepositoryID, opts.Commit), sqlf.Join(conds, " AND "))
```
keynmol, line 302: I think there's some inconsistency here:

1. `makeVisibleUploadsQuery` takes only a subset of fields, which aligns more with principle of least power
2. `makeFindClosestProcessUploadsConditions` - takes entire `opts`, but seems to only use a subset as well

I think it'd be better to convert (2) to take necessary fields as args as well

varungandhi-src: It's because out of the various fields, the tricky one is the Root<->Path matching one, and so being able to do 'Find usages' on that is useful. By having `makeFindClosestProcessUploadsConditions` take the entire `opts`, 'Find usages' directly provides the most useful result for that field.

The repository ID and commit fields aren't that interesting, and are used in a bunch of other places.

That said, I can change the latter if you still think it's better to have consistency here.

keynmol (approved):
Great improvement, 1 small comment, optional to address

PR #62338, opened 2024-05-01
[Svelte]: Improve result indicator layout

vovakulikov:
The main intention to change here is to save vertical space, we already have a few panels and sticky headers in the search result block, and having an additional row in the sticky header above make this panel crowded especially on mid-size screens

| Before | After |
| ------- | ------- |
| Screenshot 2024-05-01 at 14 16 47 |  Screenshot 2024-05-01 at 14 15 52 |

## Test plan
- Check progress result UI on the search result page

### DIFF
--- a/client/web-sveltekit/src/lib/search/resultsIndicator/ResultsIndicator.svelte
+++ b/client/web-sveltekit/src/lib/search/resultsIndicator/ResultsIndicator.svelte
@@ -36,52 +36,66 @@
     $: done = progress.done || state === 'complete'
 </script>
 
-<div class="indicator">
-    <div>
-        {#if loading}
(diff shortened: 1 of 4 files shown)


camdencheek:
Looks like this will conflict with https://github.com/sourcegraph/sourcegraph/pull/62317. Thoughts on how we should handle that?

vovakulikov:
So after the pairing session, we agreed that this PR won't go to the May release, instead, we're going to merge https://github.com/sourcegraph/sourcegraph/pull/62317 and in this PR we split the result button into two buttons, info one with result count and reasons badge and action button which populates the query with additional filter to resolve the problem that info button shows

PR #62339, opened 2024-05-01
ci: don't emit bep for non-test commands

Strum355:
Rebase mistake reintroduced the BEP file in `common` :facepalm: https://github.com/sourcegraph/sourcegraph/commit/b5cf7b8ab4545101048b2abcc0fd8b6a7d832a65

## Test plan

Undoing change reintroduced by mistake


jac (approved):
stamp

PR #62342, opened 2024-05-01
docs: Fix Sourcegraph Secret regex format

willdollman:
Based on the spacing, I suspect this was done by an overly-helpful markdown table editor plugin :)

## Test plan

- Docs - rely on CI


shivasurya (approved):
Thank you 🥇

PR #62346, opened 2024-05-01
Pass go debug flags to zoekt to enable better debugging

mmanela:
When testing a change I hit issues debugging zoekt. @ggilmore showed me how we can plumb the proper debug flags to zoekt.  This seems like it would be generally useful for local development.

## Test plan
- Debugged locally and it all worked!

### DIFF
--- a/sg.config.yaml
+++ b/sg.config.yaml
@@ -513,7 +513,6 @@ commands:
       if [ -n "$DELVE" ]; then
         export GCFLAGS='all=-N -l'
       fi
-      export GCFLAGS='all=-N -l'
       mkdir -p .bin
       export GOBIN="${PWD}/.bin"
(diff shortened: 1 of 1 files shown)


ggilmore (approved):

sg.config.yaml
```diff
@@ -510,11 +510,12 @@ commands:
         -listen "127.0.0.1:$ZOEKT_LISTEN_PORT" \
         -cpu_fraction 0.25
     install: |
+      export GCFLAGS='all=-N -l'
```
jtibshirani, line 513: Maybe we can make this conditional, like we do for other services:

```
      if [ -n "$DELVE" ]; then
        export GCFLAGS='all=-N -l'
      fi
```

keegancsmith (commented):
LGTM

PR #62350, opened 2024-05-01
appliance: gitserver service definition

jdpleiness:
Add GitServer service definition and golden tests to the appliance.

## Test plan

Golden files for testing.

### DIFF
--- a/internal/appliance/gitserver.go
+++ b/internal/appliance/gitserver.go
@@ -56,14 +56,8 @@ func (r *Reconciler) reconcileGitServerStatefulSet(ctx context.Context, sg *Sour
 	// TODO: https://github.com/sourcegraph/sourcegraph/issues/62076
 	ctr.Image = "index.docker.io/sourcegraph/gitserver:5.3.2@sha256:6c6042cf3e5f3f16de9b82e3d4ab1647f8bb924cd315245bd7a3162f5489e8c4"
 
-	ctr.Env = []corev1.EnvVar{
-		container.NewEnvVarSecretKeyRef("REDIS_CACHE_ENDPOINT", "redis-cache", "endpoint"),
-		container.NewEnvVarSecretKeyRef("REDIS_STORE_ENDPOINT", "redis-store", "endpoint"),
-
-		// OTEL_AGENT_HOST must be defined before OTEL_EXPORTER_OTLP_ENDPOINT to substitute the node IP on which the DaemonSet pod instance runs in the latter variable
-		container.NewEnvVarFieldRef("OTEL_AGENT_HOST", "status.hostIP"),
-		{Name: "OTEL_EXPORTER_OTLP_ENDPOINT", Value: "http://$(OTEL_AGENT_HOST):4317"},
-	}
+	ctr.Env = append(ctr.Env, container.EnvVarsRedis()...)
+	ctr.Env = append(ctr.Env, container.EnvVarsOtel()...)
 
 	ctr.Ports = []corev1.ContainerPort{
 		{Name: "rpc", ContainerPort: 3178},
--- a/internal/appliance/gitserver.go
+++ b/internal/appliance/gitserver.go
@@ -63,6 +63,8 @@ func (r *Reconciler) reconcileGitServerStatefulSet(ctx context.Context, sg *Sour
 		{Name: "rpc", ContainerPort: 3178},
 	}
 
+	ctr.Args = []string{"run"}
+
 	ctr.LivenessProbe = &corev1.Probe{
 		ProbeHandler: corev1.ProbeHandler{
 			TCPSocket: &corev1.TCPSocketAction{
@@ -79,7 +81,7 @@ func (r *Reconciler) reconcileGitServerStatefulSet(ctx context.Context, sg *Sour
 	}
 
 	podVolumes := []corev1.Volume{
-		pod.NewVolumeEmptyDir("repos"),
+		{Name: "repos"},
 		pod.NewVolumeEmptyDir("tmpdir"),
 	}
 
--- a/internal/appliance/gitserver.go
+++ b/internal/appliance/gitserver.go
@@ -1,0 +1,135 @@
+package appliance
+
+import (
+	"context"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+	"github.com/sourcegraph/sourcegraph/lib/pointers"
+
+	"github.com/sourcegraph/sourcegraph/internal/k8s/resource/container"
+	"github.com/sourcegraph/sourcegraph/internal/k8s/resource/pod"
+	"github.com/sourcegraph/sourcegraph/internal/k8s/resource/pvc"
+	"github.com/sourcegraph/sourcegraph/internal/k8s/resource/service"
+	"github.com/sourcegraph/sourcegraph/internal/k8s/resource/serviceaccount"
+	"github.com/sourcegraph/sourcegraph/internal/k8s/resource/statefulset"
+)
+
+func (r *Reconciler) reconcileGitServer(ctx context.Context, sg *Sourcegraph, owner client.Object) error {
+	if err := r.reconcileGitServerStatefulSet(ctx, sg, owner); err != nil {
+		return err
+	}
+	if err := r.reconcileGitServerService(ctx, sg, owner); err != nil {
+		return err
+	}
+	if err := r.reconcileGitServerServiceAccount(ctx, sg, owner); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (r *Reconciler) reconcileGitServerStatefulSet(ctx context.Context, sg *Sourcegraph, owner client.Object) error {
+	cfg := sg.Spec.GitServer
+	name := "gitserver"
+
+	ctr := container.NewContainer(name, cfg, corev1.ResourceRequirements{
+		Requests: corev1.ResourceList{
+			corev1.ResourceCPU:    resource.MustParse("4"),
+			corev1.ResourceMemory: resource.MustParse("8Gi"),
+		},
+		Limits: corev1.ResourceList{
+			corev1.ResourceCPU:    resource.MustParse("4"),
+			corev1.ResourceMemory: resource.MustParse("8Gi"),
+		},
+	})
+
+	storageSize, err := resource.ParseQuantity(cfg.StorageSize)
+	if err != nil {
+		return errors.Wrap(err, "parsing storage size")
+	}
+
+	// TODO: https://github.com/sourcegraph/sourcegraph/issues/62076
+	ctr.Image = "index.docker.io/sourcegraph/gitserver:5.3.2@sha256:6c6042cf3e5f3f16de9b82e3d4ab1647f8bb924cd315245bd7a3162f5489e8c4"
+
+	ctr.Env = []corev1.EnvVar{
+		container.NewEnvVarSecretKeyRef("REDIS_CACHE_ENDPOINT", "redis-cache", "endpoint"),
+		container.NewEnvVarSecretKeyRef("REDIS_STORE_ENDPOINT", "redis-store", "endpoint"),
+
+		// OTEL_AGENT_HOST must be defined before OTEL_EXPORTER_OTLP_ENDPOINT to substitute the node IP on which the DaemonSet pod instance runs in the latter variable
+		container.NewEnvVarFieldRef("OTEL_AGENT_HOST", "status.hostIP"),
+		{Name: "OTEL_EXPORTER_OTLP_ENDPOINT", Value: "http://$(OTEL_AGENT_HOST):4317"},
+	}
+
+	ctr.Ports = []corev1.ContainerPort{
+		{Name: "rpc", ContainerPort: 3178},
+	}
+
+	ctr.LivenessProbe = &corev1.Probe{
+		ProbeHandler: corev1.ProbeHandler{
+			TCPSocket: &corev1.TCPSocketAction{
+				Port: intstr.FromString("rpc"),
+			},
+		},
+		TimeoutSeconds:      5,
+		InitialDelaySeconds: 5,
+	}
+
+	ctr.VolumeMounts = []corev1.VolumeMount{
+		{Name: "tmpdir", MountPath: "/tmp"},
+		{Name: "repos", MountPath: "/data/repos"},
+	}
+
+	podVolumes := []corev1.Volume{
+		pod.NewVolumeEmptyDir("repos"),
+		pod.NewVolumeEmptyDir("tmpdir"),
+	}
+
+	if sshSecret := sg.Spec.GitServer.SSHSecret; sshSecret != "" {
+		podVolumes = append(podVolumes, corev1.Volume{
+			Name: "ssh",
+			VolumeSource: corev1.VolumeSource{
+				Secret: &corev1.SecretVolumeSource{
+					SecretName:  sshSecret,
+					DefaultMode: pointers.Ptr[int32](0644),
+				},
+			},
+		})
+	}
+
+	podTemplate := pod.NewPodTemplate(name, cfg)
+	podTemplate.Template.Spec.Containers = []corev1.Container{ctr}
+	podTemplate.Template.Spec.ServiceAccountName = name
+	podTemplate.Template.Spec.Volumes = podVolumes
+
+	sset := statefulset.NewStatefulSet(name, sg.Namespace, sg.Spec.RequestedVersion)
+	sset.Spec.Template = podTemplate.Template
+	sset.Spec.VolumeClaimTemplates = []corev1.PersistentVolumeClaim{
+		pvc.NewPersistentVolumeClaimSpecOnly(storageSize, sg.Spec.StorageClass.Name),
+	}
+
+	return reconcileObject(ctx, r, sg.Spec.GitServer, &sset, &appsv1.StatefulSet{}, sg, owner)
+}
+
+func (r *Reconciler) reconcileGitServerService(ctx context.Context, sg *Sourcegraph, owner client.Object) error {
+	svc := service.NewService("gitserver", sg.Namespace, sg.Spec.GitServer)
+	svc.Spec.Ports = []corev1.ServicePort{
+		{Name: "unused", TargetPort: intstr.FromInt(10811), Port: 10811},
+	}
+	svc.Spec.Selector = map[string]string{
+		"app":  "gitserver",
+		"type": "gitserver",
+	}
+
+	return reconcileObject(ctx, r, sg.Spec.GitServer, &svc, &corev1.Service{}, sg, owner)
+}
+
+func (r *Reconciler) reconcileGitServerServiceAccount(ctx context.Context, sg *Sourcegraph, owner client.Object) error {
+	cfg := sg.Spec.GitServer
+	sa := serviceaccount.NewServiceAccount("gitserver", sg.Namespace, cfg)
+	return reconcileObject(ctx, r, sg.Spec.GitServer, &sa, &corev1.ServiceAccount{}, sg, owner)
+}
--- a/internal/appliance/gitserver.go
+++ b/internal/appliance/gitserver.go
@@ -105,7 +105,7 @@ func (r *Reconciler) reconcileGitServerStatefulSet(ctx context.Context, sg *Sour
 	sset := statefulset.NewStatefulSet(name, sg.Namespace, sg.Spec.RequestedVersion)
 	sset.Spec.Template = podTemplate.Template
 	sset.Spec.VolumeClaimTemplates = []corev1.PersistentVolumeClaim{
-		pvc.NewPersistentVolumeClaimSpecOnly(storageSize, sg.Spec.StorageClass.Name),
+		pvc.NewPersistentVolumeClaim("repos", sg.Namespace, storageSize, sg.Spec.StorageClass.Name),
 	}
 
 	return reconcileObject(ctx, r, sg.Spec.GitServer, &sset, &appsv1.StatefulSet{}, sg, owner)
--- a/internal/appliance/gitserver_test.go
+++ b/internal/appliance/gitserver_test.go
@@ -1,0 +1,25 @@
+package appliance
+
+import "time"
+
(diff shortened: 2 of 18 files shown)


internal/appliance/spec.go
```diff
@@ -114,20 +114,19 @@ type FrontendSpec struct {
...
-
 	// StorageSize defines the requested amount of storage for the PVC.
 	// Default: 200Gi
 	StorageSize string `json:"storageSize,omitempty"`
 
-	// Resources allows for custom resource limits and requests.
-	Resources *corev1.ResourceList `json:"resources,omitempty"`
+	// SSHSecret is the name of existing secret that contains SSH credentials to clone repositories.
+	// This secret generally contains keys such as `id_rsa` (private key) and `known_hosts`.
+	SSHSecret string `json:"sshSecret,omitempty"`
```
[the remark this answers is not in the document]
craigfurman: This is certainly injectable in the helm chart, and has no default there. From the looks of things the user has to create the secret out-of-band: https://docs.sourcegraph.com/admin/deploy/kubernetes/helm#using-ssh-to-clone-repositories

I don't object to keeping it configurable here, but let's add a default in spec.go, e.g. `git-ssh` (not git-ssh-secret because it already has type Secret). We will (hopefully) be operating in a dedicated sourcegraph namespace, so it's unlikely the admin will feel the need to create another git-ssh secret. And if they do, they can namespace it (sg-git-ssh) in this config element 😁

internal/appliance/gitserver_test.go
```diff
@@ -0,0 +1,25 @@
...
+import "time"
+
+func (suite *ApplianceTestSuite) TestDeployGitServer() {
+	for _, tc := range []struct {
+		name string
+	}{
+		{name: "gitserver-default"},
+		{name: "gitserver-with-storage"},
+		{name: "gitserver-with-resources"},
+		{name: "gitserver-with-no-resources"},
```
craigfurman, line 12: FYI, no action needed: for standard features like resources, which gitserver.go contains no logic for (a good thing), I reckon we should create a test file called standard_features_test.go or something like that, and pick random services that happen to use these features to test this in exactly one place. That's not a note for this PR, but an idea for the near future. I might pick that up after this merges.

This PR doesn't include a test for disabling, which for similar reasons, I think is totally fine!

internal/appliance/gitserver.go
```diff
@@ -0,0 +1,135 @@
...
+	ctr.Image = "index.docker.io/sourcegraph/gitserver:5.3.2@sha256:6c6042cf3e5f3f16de9b82e3d4ab1647f8bb924cd315245bd7a3162f5489e8c4"
+
+	ctr.Env = []corev1.EnvVar{
+		container.NewEnvVarSecretKeyRef("REDIS_CACHE_ENDPOINT", "redis-cache", "endpoint"),
+		container.NewEnvVarSecretKeyRef("REDIS_STORE_ENDPOINT", "redis-store", "endpoint"),
+
+		// OTEL_AGENT_HOST must be defined before OTEL_EXPORTER_OTLP_ENDPOINT to substitute the node IP on which the DaemonSet pod instance runs in the latter variable
+		container.NewEnvVarFieldRef("OTEL_AGENT_HOST", "status.hostIP"),
+		{Name: "OTEL_EXPORTER_OTLP_ENDPOINT", Value: "http://$(OTEL_AGENT_HOST):4317"},
+	}
```
craigfurman, line 66: Optional note, non-blocking:

This is turning up in most (every?) service so far. Maybe we extract a function for it so that we can do:

```go
ctr.Env = container.CommonEnvVars()
```

I'm wary of putting this as a default into NewContainer(), unless we're sure that almost everything needs it. This is a loosely-held aesthetic preference, but I think I'd rather have a little repetition through function calls, than have some services trim this default back out 🤔 

Another 2-way door.

internal/appliance/gitserver.go
```diff
@@ -0,0 +1,135 @@
...
+
+	return reconcileObject(ctx, r, sg.Spec.GitServer, &svc, &corev1.Service{}, sg, owner)
+}
+
+func (r *Reconciler) reconcileGitServerServiceAccount(ctx context.Context, sg *Sourcegraph, owner client.Object) error {
+	cfg := sg.Spec.GitServer
+	sa := serviceaccount.NewServiceAccount("gitserver", sg.Namespace, cfg)
+	return reconcileObject(ctx, r, sg.Spec.GitServer, &sa, &corev1.ServiceAccount{}, sg, owner)
+}
```
craigfurman, line 135: Does this service definitely need an SA? In my PR for the symbols service, we were talking about paring these back (https://github.com/sourcegraph/sourcegraph/pull/62270#issuecomment-2085680362), but I don't think we were super clear on what exactly to do next 🤔 

Maybe we create an issue to investigate all services need for / use of SAs (e.g. to access external services via cloud-provider-specific annotations, or to use the k8s API via rolebindings), and keep SAs in place for them?

Nothing has to change in this PR, I'm just thinking about what to write in an issue.

internal/appliance/gitserver.go
```diff
@@ -0,0 +1,135 @@
...
+
+	podTemplate := pod.NewPodTemplate(name, cfg)
+	podTemplate.Template.Spec.Containers = []corev1.Container{ctr}
+	podTemplate.Template.Spec.ServiceAccountName = name
+	podTemplate.Template.Spec.Volumes = podVolumes
+
+	sset := statefulset.NewStatefulSet(name, sg.Namespace, sg.Spec.RequestedVersion)
+	sset.Spec.Template = podTemplate.Template
+	sset.Spec.VolumeClaimTemplates = []corev1.PersistentVolumeClaim{
+		pvc.NewPersistentVolumeClaimSpecOnly(storageSize, sg.Spec.StorageClass.Name),
```
craigfurman, line 112: See corresponding note about this function in the annotated helm diff. This one is my bad, it probably needs fixing in symbols service too 🤦

internal/appliance/gitserver.go
```diff
@@ -0,0 +1,135 @@
...
+		InitialDelaySeconds: 5,
+	}
+
+	ctr.VolumeMounts = []corev1.VolumeMount{
+		{Name: "tmpdir", MountPath: "/tmp"},
+		{Name: "repos", MountPath: "/data/repos"},
+	}
+
+	podVolumes := []corev1.Volume{
+		pod.NewVolumeEmptyDir("repos"),
```
craigfurman, line 88: Back in https://github.com/sourcegraph/sourcegraph/pull/62270#issuecomment-2085020236, I commented that we didn't need an emptyDir volume for something because it would be shadowed by the mounted PVC. I was only half-correct: while the emptyDir part would have been ignored and shadowed by the PVC, we do need a named, typeless volume here so that the statefulset knows which named PVC to mount here:

```suggestion
		{Name: "repos"},
```

Helm equivalent: https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/main/charts/sourcegraph/templates/gitserver/gitserver.StatefulSet.yaml#L101

This does need fixing in symbols too, my bad 🤦

craigfurman (changes requested):
Nice! It's great to see this little framework get used! 🥳 
Feedback/ideas very welcome of course!

Do we want to add a note that this closes https://github.com/sourcegraph/sourcegraph-operator/issues/47?

The comments are mostly non-blocking notes, but I think I found a couple of things in the compare-helm diff that might actually need fixing, so holding ✅ for now 🙏  

Feel free to holler at me for pairing or a chat if any of this doesn't make sense.

The main way I reviewed the functionality of this was reading the output of:

```
  -component gitserver \
  -golden-file internal/appliance/testdata/golden-fixtures/gitserver-default.yaml \
  -helm-template-extra-args '--set gitserver.serviceAccount.create=true'

```

Annotated output 👇 

```diff
2c2
< # helm: ServiceAccount/gitserver
---
> # golden: ServiceAccount/gitserver
5a6,8
>   annotations:
>     appliance.sourcegraph.com/configHash: 83479964f6f3f84d669035db9422f03b10c7543d344f6c222c1d1698eb3acb6a
>   creationTimestamp: "2024-04-19T00:00:00Z"
7,8d9
<     app.kubernetes.io/component: gitserver
<     category: rbac
10a12,14
>   namespace: NORMALIZED_FOR_TESTING
>   resourceVersion: NORMALIZED_FOR_TESTING
>   uid: NORMALIZED_FOR_TESTING
12c16
< # helm: Service/gitserver
---
> # golden: Service/gitserver
17,18c21
<     description: Headless service that provides a stable network identity for the
<       gitserver stateful set.
```

I haven't been porting these descriptions either tbh

```diff
---
>     appliance.sourcegraph.com/configHash: 83479964f6f3f84d669035db9422f03b10c7543d344f6c222c1d1698eb3acb6a
20a24
>   creationTimestamp: "2024-04-19T00:00:00Z"
25d28
<     type: gitserver
26a30,32
>   namespace: NORMALIZED_FOR_TESTING
>   resourceVersion: NORMALIZED_FOR_TESTING
>   uid: NORMALIZED_FOR_TESTING
28c34,40
<   clusterIP: None
---
>   clusterIP: NORMALIZED_FOR_TESTING
>   clusterIPs:
>   - NORMALIZED_FOR_TESTING
>   internalTrafficPolicy: Cluster
>   ipFamilies:
>   - IPv4
>   ipFamilyPolicy: SingleStack
31a44
>     protocol: TCP
35,36d47
<     app.kubernetes.io/instance: release-name
<     app.kubernetes.io/name: sourcegraph
37a49
>   sessionAffinity: None
38a51,52
> status:
>   loadBalancer: {}
40c54
< # helm: StatefulSet/gitserver
---
> # golden: StatefulSet/gitserver
45c59,61
<     description: Stores clones of repositories to perform Git operations.
---
>     appliance.sourcegraph.com/configHash: 83479964f6f3f84d669035db9422f03b10c7543d344f6c222c1d1698eb3acb6a
>   creationTimestamp: "2024-04-19T00:00:00Z"
>   generation: 1
48,49d63
<     app.kubernetes.io/instance: release-name
<     app.kubernetes.io/managed-by: Helm
51c65
<     app.kubernetes.io/version: 5.3.2
---
>     app.kubernetes.io/version: 5.3.9104
53d66
<     helm.sh/chart: sourcegraph-5.3.2
54a68,70
>   namespace: NORMALIZED_FOR_TESTING
>   resourceVersion: NORMALIZED_FOR_TESTING
>   uid: NORMALIZED_FOR_TESTING
55a72,76
>   minReadySeconds: 10
>   persistentVolumeClaimRetentionPolicy:
>     whenDeleted: Retain
>     whenScaled: Retain
>   podManagementPolicy: OrderedReady
61,62d81
<       app.kubernetes.io/instance: release-name
<       app.kubernetes.io/name: sourcegraph
67d85
<         checksum/redis: 63b58e05a2640417d599c4aee6d866cb9063e3a9aa452dc08dbfff836b7781b7
68a87
>       creationTimestamp: null
71,72d89
<         app.kubernetes.io/instance: release-name
<         app.kubernetes.io/name: sourcegraph
74,75c91
<         group: backend
<         type: gitserver
---
>       name: gitserver
77d92
<       affinity: null
79,81c94
<       - args:
<         - run
```

This looks like it might actually be a regression.

```diff
<         env:
---
>       - env:
94a108
>               apiVersion: v1
100a115
>           failureThreshold: 3
101a117,118
>           periodSeconds: 10
>           successThreshold: 1
108a126
>           protocol: TCP
112c130
<             memory: 8G
---
>             memory: 8Gi
115c133
<             memory: 8G
---
>             memory: 8Gi
```

This is actually a slightly different resource sizing (1024 vs 1000). Not that
it matters much, but we may as well squash the diff!

```diff
120a139
>         terminationMessagePath: /dev/termination-log
127c146,148
<       nodeSelector: null
---
>       dnsPolicy: ClusterFirst
>       restartPolicy: Always
>       schedulerName: default-scheduler
132a154
>       serviceAccount: gitserver
134c156
<       tolerations: null
---
>       terminationGracePeriodSeconds: 30
136a159,160
>         name: repos
>       - emptyDir: {}
```

See other notes in this diff and the code about needing to put an untyped, but
named, volume here to match the PVC.

```diff
138d161
<       - name: repos
142,143c165,168
<   - metadata:
<       name: repos
---
```

Ah, I think we do actually need metadata.name to be present in this PVC template
in order for the statefulset controller to match it to a pod volume. I think
this is a bug I introduced in the symbols service PR. Let's get rid of
pvc.NewPersistentVolumeClaimSpecOnly() completely, and use
NewPersistentVolumeClaim(), since we always need the name. I think the rest of
the metadata (e.g. namespace) is probably harmless.

This needs fixing in the symbols service too, since I missed this there! My bad.

```diff
>   - apiVersion: v1
>     kind: PersistentVolumeClaim
>     metadata:
>       creationTimestamp: null
150a176,251
>       volumeMode: Filesystem
>     status:
>       phase: Pending
> status:
>   availableReplicas: 0
>   replicas: 0
> ---
> # golden: ConfigMap/sg
> apiVersion: v1
> data:
>   spec: |
>     spec:
>       requestedVersion: "5.3.9104"
> 
>       blobstore:
>         disabled: true
> 
>       codeInsights:
>         disabled: true
> 
>       codeIntel:
>         disabled: true
> 
>       frontend:
>         disabled: true
> 
>       gitServer: {}
> 
>       indexedSearch:
>         disabled: true
> 
>       indexedSearchIndexer:
>         disabled: true
> 
>       pgsql:
>         disabled: true
> 
>       postgresExporter:
>         disabled: true
> 
>       preciseCodeIntel:
>         disabled: true
> 
>       redisCache:
>         disabled: true
> 
>       redisExporter:
>         disabled: true
> 
>       redisStore:
>         disabled: true
> 
>       repoUpdater:
>         disabled: true
> 
>       searcher:
>         disabled: true
> 
>       symbols:
>         disabled: true
> 
>       syntectServer:
>         disabled: true
> 
>       worker:
>         disabled: true
> kind: ConfigMap
> metadata:
>   annotations:
>     appliance.sourcegraph.com/currentVersion: 5.3.9104
>     appliance.sourcegraph.com/managed: "true"
>   creationTimestamp: "2024-04-19T00:00:00Z"
>   name: sg
>   namespace: NORMALIZED_FOR_TESTING
>   resourceVersion: NORMALIZED_FOR_TESTING
>   uid: NORMALIZED_FOR_TESTING
```

internal/appliance/gitserver.go
jdpleiness: Yeah this is more "luck of the draw" from the services we've been working on and it appears to only be used in 6ish of the services, which is a fair amount, but not all.
I'd rather keep it out as well.

craigfurman: I'm happy leaving this as-is, but actually, WDYT of this idea: extract 2 functions, so that we can do something like:

```go
ctr.Env = append(ctr.Env, container.EnvVarsRedis()...)
ctr.Env = append(ctr.Env, container.EnvVarsOtel()...)
```

Again totally happy to leave it, it's doing no harm 😁

jdpleiness: Yeah I think that's a nice middle ground. Less copypasta as well 👍

internal/appliance/gitserver.go
jdpleiness: Ahh okay, I was wondering a bit about this as well and went back and forth between two a few times 😄

internal/appliance/gitserver.go
jdpleiness: I thought some more about this, and there migtht be some use cases for the SAs that our cloud is using, but I'd have to double check. It makes me skew toward leaving them in for now the time being though.

I think another issue to get a final decision on this would be good though.

internal/appliance/gitserver.go
jdpleiness: I switched this over but it looks as though is defaults to use `emptyDir` if no type is specified.

```shell
125a147,148
>         name: repos
>       - emptyDir: {}
127d149
<       - name: repos
```

internal/appliance/spec.go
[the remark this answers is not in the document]
craigfurman: We talked about this on zoom, empty string has semantic meaning (the git-ssh feature is optional). This note was not relevant! 😁

internal/appliance/gitserver.go
craigfurman: Ah ok, that makes sense. We still need a metadata.name on the PVC though, so that it can be matched to a pod volume (and shadow the empty dir)

craigfurman (approved):
🚀 latest compare-helm diff looks great!

I forgot to comment on the checksum/redis annotation in the diff - in previous PRs we punted this to https://github.com/sourcegraph/sourcegraph/issues/62227, so no action needed IMO.

I can start a branch on top of yours to fix up the symbols service with the same bits we fixed here and remove NewPersistentVolumeClaimSpecOnly.

internal/appliance/testdata/golden-fixtures/gitserver-default.yaml
```diff
@@ -0,0 +1,250 @@
...
+      type: RollingUpdate
+    volumeClaimTemplates:
+    - apiVersion: v1
+      kind: PersistentVolumeClaim
+      metadata:
+        creationTimestamp: null
+        labels:
+          deploy: sourcegraph
+        name: repos
+        namespace: test-appliance-7a0613
```
craigfurman, line 121: Ah, I just realised this will break the build on every subsequent test run. We need to add a normalization for this field in golden_testing.go, like this: 

```patch
diff --git a/internal/appliance/golden_test.go b/internal/appliance/golden_test.go
index a34ae7652e7..ba1de1c7ed2 100644
--- a/internal/appliance/golden_test.go
+++ b/internal/appliance/golden_test.go
@@ -5,6 +5,7 @@ import (
 	"path/filepath"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -64,6 +65,15 @@ func (suite *ApplianceTestSuite) gatherResources(namespace string) []client.Obje
 	suite.Require().NoError(err)
 	for _, obj := range ssets.Items {
 		obj := obj
+
+		pvcts := obj.Spec.VolumeClaimTemplates
+		normalizedPVCTs := make([]corev1.PersistentVolumeClaim, len(pvcts))
+		for i, pvct := range pvcts {
+			pvct.SetNamespace("NORMALIZED_FOR_TESTING")
+			normalizedPVCTs[i] = pvct
+		}
+		obj.Spec.VolumeClaimTemplates = normalizedPVCTs
+
 		obj.SetGroupVersionKind(schema.GroupVersionKind{Group: "apps", Version: "v1", Kind: "StatefulSet"})
 		normalizeObj(&obj)
 		objs = append(objs, &obj)
diff --git a/internal/appliance/testdata/golden-fixtures/gitserver-default.yaml b/internal/appliance/testdata/golden-fixtures/gitserver-default.yaml
index 05d83d6871b..64b9fa39096 100644
--- a/internal/appliance/testdata/golden-fixtures/gitserver-default.yaml
+++ b/internal/appliance/testdata/golden-fixtures/gitserver-default.yaml
@@ -118,7 +118,7 @@ resources:
         labels:
           deploy: sourcegraph
         name: repos
-        namespace: test-appliance-7a0613
+        namespace: NORMALIZED_FOR_TESTING
       spec:
         accessModes:
         - ReadWriteOnce
diff --git a/internal/appliance/testdata/golden-fixtures/gitserver-with-no-resources.yaml b/internal/appliance/testdata/golden-fixtures/gitserver-with-no-resources.yaml
index f5dc953e748..f272d713bf3 100644
--- a/internal/appliance/testdata/golden-fixtures/gitserver-with-no-resources.yaml
+++ b/internal/appliance/testdata/golden-fixtures/gitserver-with-no-resources.yaml
@@ -112,7 +112,7 @@ resources:
         labels:
           deploy: sourcegraph
         name: repos
-        namespace: test-appliance-b5cabb
+        namespace: NORMALIZED_FOR_TESTING
       spec:
         accessModes:
         - ReadWriteOnce
diff --git a/internal/appliance/testdata/golden-fixtures/gitserver-with-resources.yaml b/internal/appliance/testdata/golden-fixtures/gitserver-with-resources.yaml
index 52471a9e1a9..152c8a8319b 100644
--- a/internal/appliance/testdata/golden-fixtures/gitserver-with-resources.yaml
+++ b/internal/appliance/testdata/golden-fixtures/gitserver-with-resources.yaml
@@ -118,7 +118,7 @@ resources:
         labels:
           deploy: sourcegraph
         name: repos
-        namespace: test-appliance-3d0da9
+        namespace: NORMALIZED_FOR_TESTING
       spec:
         accessModes:
         - ReadWriteOnce
diff --git a/internal/appliance/testdata/golden-fixtures/gitserver-with-storage.yaml b/internal/appliance/testdata/golden-fixtures/gitserver-with-storage.yaml
index fc3067ab16a..5663fc92596 100644
--- a/internal/appliance/testdata/golden-fixtures/gitserver-with-storage.yaml
+++ b/internal/appliance/testdata/golden-fixtures/gitserver-with-storage.yaml
@@ -118,7 +118,7 @@ resources:
         labels:
           deploy: sourcegraph
         name: repos
-        namespace: test-appliance-d1ab84
+        namespace: NORMALIZED_FOR_TESTING
       spec:
         accessModes:
         - ReadWriteOnce
diff --git a/internal/appliance/testdata/golden-fixtures/symbols-default.yaml b/internal/appliance/testdata/golden-fixtures/symbols-default.yaml
index 13007991598..c959a82964a 100644
--- a/internal/appliance/testdata/golden-fixtures/symbols-default.yaml
+++ b/internal/appliance/testdata/golden-fixtures/symbols-default.yaml
@@ -137,6 +137,7 @@ resources:
       kind: PersistentVolumeClaim
       metadata:
         creationTimestamp: null
+        namespace: NORMALIZED_FOR_TESTING
       spec:
         accessModes:
         - ReadWriteOnce
diff --git a/internal/appliance/testdata/golden-fixtures/symbols-with-storage.yaml b/internal/appliance/testdata/golden-fixtures/symbols-with-storage.yaml
index 7d058912bdc..1c2967222cc 100644
--- a/internal/appliance/testdata/golden-fixtures/symbols-with-storage.yaml
+++ b/internal/appliance/testdata/golden-fixtures/symbols-with-storage.yaml
@@ -137,6 +137,7 @@ resources:
       kind: PersistentVolumeClaim
       metadata:
         creationTimestamp: null
+        namespace: NORMALIZED_FOR_TESTING
       spec:
         accessModes:
         - ReadWriteOnce

```

[normalize_statefulset_pvct_ns.patch](https://github.com/sourcegraph/sourcegraph/files/15198373/normalize_statefulset_pvct_ns.patch)

jdpleiness:
closes https://github.com/sourcegraph/sourcegraph-operator/issues/47

PR #62351, opened 2024-05-01
Svelte: Font size tweaks for results indicator

taiyab:
Follow up to https://github.com/sourcegraph/sourcegraph/pull/62317

## Test plan
Locally tested for visual changes.

### DIFF
--- a/client/web-sveltekit/src/lib/search/resultsIndicator/ProgressMessage.svelte
+++ b/client/web-sveltekit/src/lib/search/resultsIndicator/ProgressMessage.svelte
@@ -21,7 +21,8 @@
 
 <style lang="scss">
     .progress-message {
-        font-size: var(--font-size-base);
+        font-size: var(--font-size-small);
+        color: var(--text-title);
 
         &.error-text {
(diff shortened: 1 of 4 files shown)


camdencheek (changes requested):
Something about this change caused the button to change size vertically between the loaded and completed states

https://github.com/sourcegraph/sourcegraph/assets/12631702/d8d23766-6288-4617-813d-68fb83ff8e2b

taiyab:
@camdencheek Should be sorted now.

camdencheek:
Looks like there is still a shift when `TimeoutMessage` is shown because we would need to keep the padding in sync between all three things. 

Instead of having to keep the padding in sync between these three different spots, I think it would be better to just [increase the `row-gap`](https://github.com/sourcegraph/sourcegraph/pull/62352) in the flexbox. I made that PR against this branch so you can merge it in if it looks good to you

screenshot-2024-05-01_14-16-29@2x

taiyab:
Merged in your changes.

camdencheek (approved):

PR #62353, opened 2024-05-01
Updates to cookies sent to HubSpot

dadlerj:
Updates to cookies sent to HubSpot on user signups on dotcom. Changing property names and eliminating certain fields.

## Test plan

Low risk change that only affects dotcom. Confirmed it compiled and ran locally. CI.


unknwon:
Is it ready for review? Asking because the PR is in draft mode.

PR #62356, opened 2024-05-01
svelte: Fix user name in "last commit" message

fkling:
Currently we are using `name`, which seems generated. We should try `displayName` first before falling back to `name`.

| Before | After |
|--------|--------|
|  |  | 

## Test plan

Manual testing.

### DIFF
--- a/client/web-sveltekit/src/lib/repo/LastCommit.gql
+++ b/client/web-sveltekit/src/lib/repo/LastCommit.gql
@@ -7,6 +7,8 @@ fragment LastCommitFragment on GitCommit {
         date
         person {
             ...Avatar_Person
+            displayName
+            name
         }
     }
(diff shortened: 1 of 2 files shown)


client/web-sveltekit/src/lib/repo/LastCommit.gql
```diff
@@ -7,6 +7,8 @@ fragment LastCommitFragment on GitCommit {
         date
         person {
             ...Avatar_Person
+            displayName
+            name
```
fkling, line 11: These fields hadn't been explicitly specified. It only worked because `Avatar_Person` includes those fields, but the component should explicitly declare all its dependencies.

camdencheek (approved):
Good catch, thanks!

PR #62357, opened 2024-05-01
svelte: Set max width for repo page

fkling:
I believe we used to do this, maybe there was a regression.

| Before | After |
|--------|--------|
|  |  | 

## Test plan

Manual testing.


camdencheek (approved):
Oh, I think I might have done this on the blob page but forgot about the repo page

PR #62365, opened 2024-05-01
Svelte: Revert code bg changes

taiyab:
## Before
CleanShot 2024-05-01 at 23 48 37@2x

## After
CleanShot 2024-05-01 at 23 46 40@2x

## Test plan
Colors manually tested locally.


fkling (approved):

fkling:
I've merged this because it also affects the diff view (no colored background) and I rather have this in sooner than later 🙂

PR #62367, opened 2024-05-02
dev/linearhooks: add POC

michaellzc:
This PR introduced a new service to handle Linear webhook. The initial use case is to move issues between teams base on labels. See added README.md for more information.

TODO

- [ ] spin up a MSP service at `linearhooks.sgdev.org`
- [ ] ask Tech Ops to create an API key using the machine account
- [ ] configure the webhook in linear

## Test plan

Tested in our linear workspace, and it works

```yaml
spec:
  mover:
    rules:
      - src:
          teamId: MLZC
          labels: [team/cloud]
        dst:
          teamId: FOO
```

Confirmed issues are moved to the desired team

- https://linear.app/sourcegraph/team/FOO/all
- https://linear.app/sourcegraph/team/MLZC/all


michaellzc:
* **#62367** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62367?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @michaellzc and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

dev/linearhooks/internal/lineargql/gqltest/gqltest.go
```diff
@@ -0,0 +1,120 @@
+package gqltest
```
michaellzc, line 1: copy from https://github.com/sourcegraph/controller/tree/main/internal/srcgql/gqltest

michaellzc:
cc @eshamow let me know if you want to pair on deploying the MSP service

dev/linearhooks/internal/lineargql/gen.go
```diff
@@ -0,0 +1,5 @@
+package lineargql
+
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
+
+//counterfeiter:generate -o gqltest github.com/Khan/genqlient/graphql.Client
```
michaellzc, line 5: `counterfeiter` is impossible to run in bazel unfortunately

dev/linearhooks/internal/lineargql/BUILD.bazel
```diff
@@ -0,0 +1,63 @@
...
+    cmd = "$(execpath @com_github_khan_genqlient//:genqlient) $(location genqlient.yaml) && mv dev/linearhooks/internal/lineargql/generated.go $@",
+    tools = ["@com_github_khan_genqlient//:genqlient"],
+)
+
+write_generated_to_source_files(
+    name = "write_genql_yaml",
+    output_files = {"generated.go": "_generated.go"},
+    tags = ["go_generate"],
+    target = ":generate_genql_yaml",
+)
```
michaellzc, line 46: fetch linear graphql schema and generate Go client

dev/linearhooks/config.example.yaml
```diff
@@ -0,0 +1,22 @@
...
+          labels: [team/bar]
+        dst:
+          # teamId is the identifier of the destination team. Issues that match the rule will be moved to this team.
+          # Either identifier or UUID is accepted
+          teamId: BAR
+      - src:
+          teamId: BAR
+          labels: [some-label, team/foo]
+          # defaults to AND
+          op: OR
```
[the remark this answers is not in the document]
michaellzc: feeling iffy about this as well.

The spec is derived from https://github.com/actions/add-to-project where they support logical operators.

One use case I have in mind is

```
src:
  labels: [team/old-team-name, team/new-team-name]
```

of course, one can achieve similar result by just adding another rule entry:

```yml
rules:
  - src:
    labels: [team/old-team-name]
  - src:
    labels: [team/new-team-name]
```

no strong preference here. happy to remove it if it's too confusing.

dev/linearhooks/internal/lineargql/gen.go
michaellzc: I think we do

https://github.com/sourcegraph/sourcegraph/blob/f9eddd3c4503dc025c6222b177cbc0120eae3b31/dev/sg/internal/generate/golang/golang.go#L53-L55

will check with Dev Infra prior to merging

dev/linearhooks/config.example.yaml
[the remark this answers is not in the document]
varungandhi-src: Yeah, I think it's best to remove it for the first pass unless it is strictly necessary.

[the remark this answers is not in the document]
michaellzc: removed

dev/linearhooks/internal/lineargql/gen.go
michaellzc: confirmed it works:

```
⋊> ~/C/s/sourcegraph on 05-01-dev_linearhooks_add_poc ◦ rm dev/linearhooks/internal/lineargql/gqltest/fake_client.go                                                            10:31:57
⋊> ~/C/s/sourcegraph on 05-01-dev_linearhooks_add_poc ⨯ git status                                                                                                              10:32:31
On branch 05-01-dev_linearhooks_add_poc
Your branch is up to date with 'origin/05-01-dev_linearhooks_add_poc'.

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        deleted:    dev/linearhooks/internal/lineargql/gqltest/fake_client.go

no changes added to commit (use "git add" and/or "git commit -a")
⋊> ~/C/s/sourcegraph on 05-01-dev_linearhooks_add_poc ⨯ sg generate go                                                                                                          10:32:37
✅ 3/3 packages generated  ███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████  100%
ℹ️ go generate ./... (excluding doc/cli/references)
ℹ️ goimports -w
ℹ️ go mod tidy
✅ go (15s)
⋊> ~/C/s/sourcegraph on 05-01-dev_linearhooks_add_poc ◦ git status                                                                                                              10:32:57
On branch 05-01-dev_linearhooks_add_poc
Your branch is up to date with 'origin/05-01-dev_linearhooks_add_poc'.

nothing to commit, working tree clean
⋊> ~/C/s/sourcegraph on 05-01-dev_linearhooks_add_poc ◦          
```

dev/linearhooks/internal/handlers/handlers.go
```diff
@@ -0,0 +1,151 @@
...
+		return
+	}
+
+	w.Write([]byte("OK"))
+}
+
+func moveIssue(logger log.Logger, client graphql.Client, rs []RuleSpec, i linearschema.IssueData) error {
+	for _, r := range rs {
+		if r.Src.TeamID == i.Team.ID || r.Src.TeamID == i.Team.Key {
+			if shouldMove(r, i.LabelNames()) {
```
varungandhi-src, line 104: style: It seems weird to have the logic for 'should move or not' be split across different functions (the team check is happening here, and some checks are happening inside `shouldMove`)

Instead, it would be better to fold these into the same function and return the new team name from that.

```
func (r *RuleSet) moveToNewTeam(issueData linearschema.IssueData) (newTeamId *string) {
    if len(s.Src.Labels) == 0 {
        panic("Expected non-empty label set for RuleSet")
    } 
    teamMatch := r.Src.TeamID == issueData.Team.ID || r.Src.TeamID == i.Team.Key
    currentLabels := collections.NewSet(issueData.LabelNames())
    if teamMatch && collections.NewSet(r.Src.Labels).Difference(currentLabels).IsEmpty() {
        return &r.Dst.TeamID
    }
    return nil
}
```

That way, fewer functions have to poke at the internals of RuleSet apart from the methods on RuleSet.

dev/linearhooks/internal/handlers/handlers_test.go
```diff
@@ -0,0 +1,139 @@
...
+	t.Parallel()
+
+	tests := []struct {
+		name   string
+		r      RuleSpec
+		labels []string
+		want   bool
+	}{
+		{
+			name:   "yes",
```
varungandhi-src, line 105: I think the general convention is the name describes the situation that the test is checking, and that the `name` is different for different test cases, right?

dev/linearhooks/internal/handlers/config.go
```diff
@@ -0,0 +1,80 @@
...
+func (s *MoverSpec) Validate() error {
+	var errs errors.MultiError
+	if len(s.Rules) == 0 {
+		errs = errors.Append(errs, errors.New("rules must contain at least one rule"))
+	}
+	for _, r := range s.Rules {
+		if err := r.Validate(); err != nil {
+			errs = errors.Append(errs, err)
+		}
+	}
```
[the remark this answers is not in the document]
michaellzc: this is actually fine. `Rules` is `[]Rule`, hence we can rely on the zero value ;) 

we could exit early, but I prefer keeping the consistent style of `err = errors.Append()`. especially if we need to add more fields with validation in the future

dev/linearhooks/internal/handlers/handlers.go
```diff
@@ -0,0 +1,151 @@
...
+	if v, ok := teamKeyUUIDCache.Load(idOrKey); ok {
+		return v.(string), nil
+	}
+	resp, err := lineargql.GetTeamById(ctx, c, idOrKey)
+	if err != nil {
+		return "", err
+	}
+	teamKeyUUIDCache.Store(idOrKey, resp.Team.Id)
+	return resp.Team.Id, nil
+}
```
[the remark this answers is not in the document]
michaellzc: moved to https://github.com/hashicorp/golang-lru

dev/linearhooks/README.md
```diff
@@ -0,0 +1,29 @@
+# Linear Webhooks
+
+## Development
+
+Make of a copy of the dotenv file. You can retrieve Linear API key and webhook signing secrets from [Linear settings](https://linear.app/sourcegraph/settings/api). If you don't have access, reach out to #wg-linear-trial.
```
varungandhi-src, line 5:
```suggestion
First make a copy of the dotenv file and set the API key and webhook signing secrets in `.env` based on the [Linear API settings](https://linear.app/sourcegraph/settings/api). If you don't have access, reach out to #wg-linear-trial.
```

dev/linearhooks/internal/handlers/handlers.go
```diff
@@ -0,0 +1,160 @@
...
+				return errors.Wrapf(err, "move issue %q to team %q", issue.Identifier, teamUUID)
+			}
+			return nil
+		}
+	}
+	return nil
+}
+
+var (
+	// teamKeyUUIDCache is a cache of team key to team UUID mappings with a 5 minute TTL.
```
varungandhi-src, line 128:
```suggestion
	// teamKeyUUIDCache caches team identifier -> UUID mappings to avoid
	// re-requesting the same data for every label change.
	// The cache has a TTL for eventually correct behavior in case a team is
	// deleted and re-created.
```

I think it would be more useful to elaborate on the why (why have a cache here? why use a TTL?).

Also, it seems like the terms 'identifier' and 'key' are essentially equivalent in the API, so perhaps we should just consistently use 'identifier'?

varungandhi-src (approved):

dev/linearhooks/internal/handlers/handlers.go
michaellzc: linear has not been very consistent on theirs own as well. l'll leave a note of `identifier/key -> UUID` to avoid renaming a bunch of things everywhere.

PR #62368, opened 2024-05-02
gitserver: Implement CommitGraph using Commits

eseliger:
This PR consolidates the client methods CommitGraph and Commits into one, to reduce the surface of APIs we have to migrate to gRPC and then maintain.
To my best understanding, this should be equivalent, just use a different API that returns slightly more information than necessary.

But once we have a baseline of all incoming requests, we will be able to much better optimize specific use-cases.

Closes https://github.com/sourcegraph/sourcegraph/issues/62104

Test plan:

Existing test suites and integration tests are still passing.


eseliger:
* **#62368** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62368?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

internal/codeintel/uploads/internal/commitgraph/commit_graph_test.go
```diff
@@ -187,12 +187,13 @@ func BenchmarkCalculateVisibleUploads(b *testing.B) {
...
 func readBenchmarkCommitGraph() (*gitdomain.CommitGraph, error) {
-	contents, err := readBenchmarkFile(filepath.Join("testdata", customer, "commits.txt.gz"))
+	_, err := readBenchmarkFile(filepath.Join("testdata", customer, "commits.txt.gz"))
 	if err != nil {
 		return nil, err
 	}
 
-	return gitdomain.ParseCommitGraph(strings.Split(string(contents), "\n")), nil
+	return gitdomain.ParseCommitGraph(nil), nil
+	// return gitdomain.ParseCommitGraph(strings.Split(string(contents), "\n")), nil
```
eseliger, line 196: TODO

internal/codeintel/uploads/internal/store/commitgraph_test.go
```diff
@@ -1566,12 +1566,13 @@ func BenchmarkCalculateVisibleUploads(b *testing.B) {
...
 func readBenchmarkCommitGraph() (*gitdomain.CommitGraph, error) {
-	contents, err := readBenchmarkFile("../../../commitgraph/testdata/customer1/commits.txt.gz")
+	_, err := readBenchmarkFile("../../../commitgraph/testdata/customer1/commits.txt.gz")
 	if err != nil {
 		return nil, err
 	}
 
-	return gitdomain.ParseCommitGraph(strings.Split(string(contents), "\n")), nil
+	return gitdomain.ParseCommitGraph(nil), nil
+	// return gitdomain.ParseCommitGraph(strings.Split(string(contents), "\n")), nil
```
eseliger, line 1575: TODO

ggilmore (approved):
You should probably wait on @sourcegraph/team-graph for a more informed review, but this looks okay

internal/codeintel/uploads/internal/commitgraph/git_commit_graph.go
```diff
@@ -0,0 +1,55 @@
+package commitgraph
```
varungandhi-src, line 1: What's the reason for moving this to the `commitgraph` package instead of leaving it under `gitdomain`? It seems broadly useful outside of codeintel, e.g. if we want to show a graph visualization of the commit history in the web UI.

internal/codeintel/uploads/internal/commitgraph/git_commit_graph.go
```diff
@@ -0,0 +1,55 @@
...
+
+func (c *CommitGraph) Graph() map[api.CommitID][]api.CommitID { return c.graph }
+func (c *CommitGraph) Order() []api.CommitID                  { return c.order }
+
+// ParseCommitGraph converts the output of git log into a map from commits to
+// parent commits, and a topological ordering of commits such that parents come
+// before children. If a commit is listed but has no ancestors then its parent
+// slice is empty, but is still present in the map and the ordering. If the
+// ordering is to be correct, the given commits must be ordered with `--topo-order`.
+func ParseCommitGraph(commits []*gitdomain.Commit) *CommitGraph {
```
varungandhi-src, line 21: We should probably mention `CommitsOrderTopoDate` here instead of `--topo-order`.

internal/codeintel/uploads/internal/commitgraph/commit_graph_test.go
```diff
@@ -186,13 +187,27 @@ func BenchmarkCalculateVisibleUploads(b *testing.B) {
...
+		commit := &gitdomain.Commit{
+			ID: api.CommitID(parts[0]),
+		}
+		for _, parent := range parts[1:] {
+			commit.Parents = append(commit.Parents, api.CommitID(parent))
+		}
+		commits = append(commits, commit)
+	}
+
+	return ParseCommitGraph(commits), nil
```
varungandhi-src, line 210: Can we split out some of the logic to avoid having the large struct literals in the test cases? The test case inputs are now arguably harder to read because the commit IDs are single letters, whereas there are many more letters taken up by the struct literals.

internal/codeintel/uploads/internal/commitgraph/git_commit_graph_test.go
```diff
@@ -1,36 +1,39 @@
-package gitdomain
+package commitgraph
```
varungandhi-src, line 1: If we're set on moving `ParseCommitGraph` to the `commitgraph` package, might as well merge the tests here into `commit_graph_test.go` since they're conceptually very similar.

internal/codeintel/uploads/internal/store/commitgraph_test.go
```diff
@@ -156,19 +157,19 @@ func TestCalculateVisibleUploadsNonDefaultBranches(t *testing.T) {
...
-		strings.Join([]string{makeCommit(7), makeCommit(6)}, " "),
-		strings.Join([]string{makeCommit(6), makeCommit(5)}, " "),
-		strings.Join([]string{makeCommit(5), makeCommit(4), makeCommit(9)}, " "),
-		strings.Join([]string{makeCommit(9), makeCommit(8)}, " "),
-		strings.Join([]string{makeCommit(8), makeCommit(2)}, " "),
-		strings.Join([]string{makeCommit(4), makeCommit(3)}, " "),
-		strings.Join([]string{makeCommit(3), makeCommit(2)}, " "),
-		strings.Join([]string{makeCommit(2), makeCommit(1)}, " "),
-		strings.Join([]string{makeCommit(1)}, " "),
+	graph := commitgraph.ParseCommitGraph([]*gitdomain.Commit{
```
varungandhi-src, line 160: Could you introduce a helper function or two here to reduce the test inputs instead of making it bigger with even more struct literals? It seems quite hard to read because of the multiple levels of wrapping.

internal/codeintel/uploads/internal/store/commitgraph_test.go
```diff
@@ -1565,13 +1566,27 @@ func BenchmarkCalculateVisibleUploads(b *testing.B) {
...
+		commit := &gitdomain.Commit{
+			ID: api.CommitID(parts[0]),
+		}
+		for _, parent := range parts[1:] {
+			commit.Parents = append(commit.Parents, api.CommitID(parent))
+		}
+		commits = append(commits, commit)
+	}
+
+	return commitgraph.ParseCommitGraph(commits), nil
```
varungandhi-src, line 1589: This logic seems to be the same in another file `commit_graph_test.go`, maybe it can be moved to a common function somewhere? (Also, maybe we should rename this file to `commit_graph_test.go`?)

varungandhi-src (commented):
Largely LGTM, and I'm glad to see a more structured API rather than what was present before.

That said, some of the tests are much more noisy now with the extra struct wrapping, and it's not super clear why some of the logic was moved from the general gitdomain package to the commitgraph package under codeintel.

internal/codeintel/uploads/internal/commitgraph/git_commit_graph.go
eseliger: This type is currently only used by codeintel use-case here and seems fairly specific to that (I think the term graph is a bit misleading here as it is a linearized version of a graph, not this graph itself).

Now that there's no CommitGraph function on the gitserver.Client interface anymore, it's not returned from the gitserver package at all, so I found it to be misplaced a bit in there. 

Does that make sense to you?

internal/gitserver/commands.go
```diff
@@ -1323,9 +1279,21 @@ func (c *clientImplementor) Stat(ctx context.Context, repo api.RepoName, commit
 	return fi, nil
 }
 
+type CommitsOrder int
+
+const (
+	// CommitsOrderDefault uses the default ordering of git log: in reverse chronological order.
+	CommitsOrderDefault CommitsOrder = iota
```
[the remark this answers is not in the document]
eseliger: No strong preference, one thing that you get for free with `int` "enums" is that the null value has a strict meaning, where in a string enum `""` would need to be special cased. But I am ok with either or.

internal/gitserver/commands.go
```diff
@@ -1323,9 +1279,21 @@ func (c *clientImplementor) Stat(ctx context.Context, repo api.RepoName, commit
...
+	// Show no parents before all of its children are shown, but otherwise show commits in the commit timestamp order.
+	CommitsOrderCommitDate
+	// Show no parents before all of its children are shown, and avoid showing commits on multiple lines of history intermixed.
+	CommitsOrderTopoDate
```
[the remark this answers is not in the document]
eseliger: yep it was just copied, I can add the links

internal/codeintel/uploads/internal/commitgraph/commit_graph_test.go
eseliger: The logic linked here is for parsing the existing benchmark file into commits which I think can't be much less verbose than this, I assume this comment was for further up in the file where we now do

```go
		{ID: "n", Parents: []api.CommitID{"l"}},
		{ID: "m", Parents: []api.CommitID{"k"}},
		{ID: "k", Parents: []api.CommitID{"h"}},
		{ID: "j", Parents: []api.CommitID{"b", "h"}},
```

? I found that to be quite expressive, but I could add some helper that looks like `func makeCommit(id api.CommitID, parents ...api.CommitID) *gitdomain.Commit` if you like that better? 
I find that actually reads a little _less_ well coz now you need to know the order of parameters, but open to do whatever you think you like the best in your codebase.

internal/codeintel/uploads/internal/commitgraph/git_commit_graph_test.go
eseliger: That's fair, I checked coverage and the existing tests in `commit_graph_test.go` cover the same branches so I think I can drop this one.

internal/codeintel/uploads/internal/commitgraph/git_commit_graph.go
varungandhi-src: Understood. We can always move it back if the code needs to be shared elsewhere.

internal/codeintel/uploads/internal/commitgraph/commit_graph_test.go
varungandhi-src: I think each line individually is easier to read with the struct literal, but as a whole, the test case becomes harder-to-understand. It's also harder to compared the test cases, and harder to draw a diagram given an array, because you need to visually scan more characters for line.

I would prefer if the lines for the ancestry were essentially unchanged, and the only difference was that there's an extra wrapping call inside the `ParseCommitGraph` call.

PR #62374, opened 2024-05-02
svelte: Enable repo routes by default

fkling:
(in preparation for the S2 rollout)

This commit enables the following routes via the `web-next-rollout` flag:

- repo home
- blob
- file
- branches
- tags
- commits
- commit

Question: Should we also enable the contributors page by default? It would be the only page left but it didn't receive any attention.

## Test plan

Inspected generated code.


vovakulikov (approved):
I think the Contributors page can wait. It doesn't block any core workflows

PR #62376, opened 2024-05-02
Check if access requests are enabled before redirecting to auth provider

### LINKED ISSUE #62313: Bug: Users cannot make access requests when there is only one auth provider configured
- **Sourcegraph version:** 5.3.11625
- **Platform information:** 

#### Steps to reproduce:

Configure Sourcegraph with only a single OAuth provider (e.g. GitHub)
Navigate to /sign-in as an anonymous user
Observe being immediately redirected to the OAuth provider instead of seeing the Sourcegraph sign-in page

#### Expected behavior:
Users should be able to access the Sourcegraph sign-in page even if there is only a single OAuth provider configured. This allows users to make access requests if signup is disabled

Screenshot 2024-04-30 at 1 37 25 PM

#### Actual behavior:
Users are immediately redirected to the OAuth provider, and can't make access requests.

https://github.com/sourcegraph/sourcegraph/assets/69164745/cc51f727-ba9e-404b-85eb-609fe88d4312

#### Workaround:
Use built-in auth so users can see the login page.

pjlast:
Closes #62313 

This PR adds an additional check to the automatic redirect check that checks if request access is enabled or not, and only redirects if request access is disabled, otherwise there would be no way to request access to the instance.

## Test plan

Verified locally.

### DIFF
--- a/client/web/src/auth/SignInPage.tsx
+++ b/client/web/src/auth/SignInPage.tsx
@@ -86,8 +86,8 @@ export const SignInPage: React.FunctionComponent<React.PropsWithChildren<SignInP
     }
 
     const thirdPartyAuthProviders = nonBuiltinAuthProviders.filter(provider => shouldShowProvider(provider))
-    // If there is only one auth provider that is going to be displayed, we want to redirect to it directly.
-    if (thirdPartyAuthProviders.length === 1 && !builtInAuthProvider) {
+    // If there is only one auth provider that is going to be displayed, and request access is disabled, we want to redirect to the auth provider directly.
+    if (thirdPartyAuthProviders.length === 1 && !builtInAuthProvider && !isRequestAccessAllowed) {
         // Add '?returnTo=' + encodeURIComponent(returnTo) to thirdPartyAuthProviders[0].authenticationURL in a safe way.
         const redirectUrl = new URL(thirdPartyAuthProviders[0].authenticationURL, window.location.href)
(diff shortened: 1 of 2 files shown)


client/web/src/auth/SignInPage.tsx
```diff
@@ -86,8 +86,8 @@ export const SignInPage: React.FunctionComponent<React.PropsWithChildren<SignInP
     }
 
     const thirdPartyAuthProviders = nonBuiltinAuthProviders.filter(provider => shouldShowProvider(provider))
-    // If there is only one auth provider that is going to be displayed, we want to redirect to it directly.
-    if (thirdPartyAuthProviders.length === 1 && !builtInAuthProvider) {
+    // If there is only one auth provider that is going to be displayed, and request access is disabled, we want to redirect to the auth provider directly.
+    if (thirdPartyAuthProviders.length === 1 && !builtInAuthProvider && !isRequestAccessAllowed) {
```
mmanela, line 90: What is `isRequestAccessAllowed` ? What is the request access feature?

pjlast: Whether or not a user can request access to the sourcegraph instance https://docs.sourcegraph.com/admin/config/site_config#auth-accessRequest

eseliger: Added some docs here: https://github.com/sourcegraph/docs/pull/289

PR #62379, opened 2024-05-02
Svelte: Support custom position target for Popover element

vovakulikov:
Prior to this Popover trigger was used as the only option for the popover target, it didn't fit in the revision popover trigger case, when the target should be a group of buttons to avoid visual overlapping, but trigger still should be different button (not both buttons). 

This PR introduces new prop for Popover to split trigger and target elements.

| Before | After |
| -------- | ------- |
| image | Screenshot 2024-05-02 at 13 42 57 |

## Test plan
- Check Repository Rev Popover position

### DIFF
--- a/client/web-sveltekit/src/lib/Popover.svelte
+++ b/client/web-sveltekit/src/lib/Popover.svelte
@@ -12,6 +12,7 @@
 
     let isOpen = false
     let trigger: HTMLElement | null
+    let target: HTMLElement | undefined
     let popoverContainer: HTMLElement | null
 
     function toggle(open?: boolean): void {
@@ -22,6 +23,10 @@
         if (event.detail !== trigger && !trigger?.contains(event.detail)) {
             isOpen = false
         }
+    }
+
+    const registerTarget: Action<HTMLElement> = node => {
+        target = node
     }
 
     const registerTrigger: Action<HTMLElement> = node => {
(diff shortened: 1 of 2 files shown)


client/web-sveltekit/src/lib/Popover.svelte
```diff
@@ -72,12 +77,12 @@
...
 </script>
 
-<slot {toggle} {registerTrigger} />
+<slot {toggle} {registerTrigger} {registerTarget} />
 {#if trigger && isOpen}
     <div
         use:registerPopoverContainer
         use:portal
-        use:popover={{ reference: trigger, options: { placement, shift: { padding: 4 } } }}
+        use:popover={{ reference: target ?? trigger, options: { placement, offset: 3, shift: { padding: 4 } } }}
```
fkling, line 85: What's the `offset` for?

fkling (approved):
Seems reasonable to me. We can expose more placement options via props, but shifting it manually doesn't seem the right solution here.
Please check other occurrences of the popover too.

client/web-sveltekit/src/lib/Popover.svelte
vovakulikov: Offset adds a bit of spacing between the target and the actual tooltip; prior, it was right next to the element, which visually didn't look good. 

Added for now, but ideally we should bring tail middleware to the popover API

vovakulikov:
@fkling what do you mean by "shifting it manually"?
If you're talking about offset setting, I added this because I think all targets should have some space between the target element and popover content, it's not something specific to the revision picker.

vovakulikov:
Merging now, but would be happy to revert `offset` change, checked other 3 popovers, they look good.

fkling:
I meant we can expose `offset` as prop and let callsites use it. Then the revision picker could offset itself to not cover the close button. But that's not a reliable solution because the styling of the close button might change, hence "doesn't seem the right solution here".

vovakulikov:
>Then the revision picker could offset itself to not cover the close button

Offset doesn't solve the overlapping problems in the revision picker; it just adds spacing between the target and popover element. Pure visual change, problem with overlapping was solved with custom target prop.

fkling:
I assumed you could offset it so far to the right so that it doesn't cover the close button. But as I said, not a reliable solution.

PR #62380, opened 2024-05-02
[Svelte]: Fix Tab appearance

vovakulikov:
This PR adds two changes
- Makes that only selected tab has font-weight bold (not sure if it was a regression, let me know if it's not, but it looks like only selected tab should have bold font)
- Uses pseudo-element for border styling, prior to this PR we used border-bottom on element, which visually "collapses" with border at the tabs row, now selected tab border overlaps tabs row border (see picture below)

| Before | After |
| ------ | ------ |
| Screenshot 2024-05-02 at 14 05 25 | Screenshot 2024-05-02 at 14 05 13 | 

## Test plan
- Check the Tabs component


client/web-sveltekit/src/lib/Tabs.svelte
```diff
@@ -103,12 +103,12 @@
     [role='tab'] {
         all: unset;
 
-        cursor: pointer;
+        margin: 0;
         border: none;
+        cursor: pointer;
         background: none;
         align-items: center;
         letter-spacing: normal;
```
fkling, line 111: Do we actually need `margin`, `border`, `background`, `letter-spacing` if we have `all: unset`?

client/web-sveltekit/src/lib/Tabs.svelte
```diff
@@ -118,20 +118,29 @@
...
         &:hover {
             color: var(--text-title);
             background-color: var(--color-bg-2);
         }
 
         &[aria-selected='true'] {
-            border-color: var(--brand-secondary);
+            font-weight: 500;
             color: var(--text-title);
+            border-color: var(--brand-secondary);
```
fkling, line 141: Is this needed?

fkling (approved):
Don't know if it was intentional or not (I guess not). Too many changes to tabs recently 😅

client/web-sveltekit/src/lib/Tabs.svelte
vovakulikov: Good catch, we don't

client/web-sveltekit/src/lib/Tabs.svelte
fkling: At the beginning I just copied a bunch of CSS to make things look ok, it was probably that.

PR #62382, opened 2024-05-02
Svelte: Modify file tree header

taiyab:
Further updates to bring in line with new designs.

## Before
CleanShot 2024-05-02 at 13 40 39@2x

## After
CleanShot 2024-05-02 at 13 40 07@2x

## Test plan
Tested locally.


fkling (commented):
So there is no way to close the sidebar now? If yes then we can remove all of the "sidebar close/open" logic.

taiyab:
@fkling I'm tempted to say yes, but we still might decide to place it elsewhere later on.

fkling:
FWIW, without that button this is impossible to use on mobile. I know we need a proper mobile strategy anyway, but maybe we still don't want to break it completely at this point.

taiyab:
> FWIW, without that button this is impossible to use on mobile. I know we need a proper mobile strategy anyway, but maybe we still don't want to break it completely at this point.

A temporary fix for now is to simply not show the file tree on mobile (they can rely on the file path for nav, of course not ideal).

Once we move the repo header into the sidebar, we'll need to deal with the mobile stuff more robustly at the same time.

taiyab:
I can't get this build to work for the life of me @fkling :D

PR #62383, opened 2024-05-02
svelte: Make several improvements to the commit page

fkling:
This adds some style and functionality improvements to the commit page.

 Styling:
 - Commit and parent SHAs as badges
 - Increased avatar size
 - Proper status badges (added, deleted, moved)
 - Normal font weight for file paths

 Behavior:
 - Add copy button for commit SHAs
 - Add a link to browse files at the commit
 - Add links to open commit at code host
 - Split changed lines into added and removed

Obviously there is more that can and should be done (e.g. adding split view), but it's a start.

| Type | Screenshot |
|--------|--------|
| dotcom/React (for reference) |  |
| Before |  |
| After |  | 

Since I updated the file search results to also use the generalized copy button, I made some layout changes to have it properly aligned to with the path name:

| Before | After |
|--------|--------|
|  |  | 

## Test plan

- Visual inspection
- Click copy buttons
- Click 'browse files' link
- Click 'open on code host' link

### DIFF
--- a/client/wildcard/src/components/Button/Button.module.scss
+++ b/client/wildcard/src/components/Button/Button.module.scss
@@ -132,10 +132,14 @@ input.btn {
     padding: 0;
     background: transparent;
     color: inherit;
-    display: flex;
+    display: inline-flex;
     align-items: center;
     border: none;
     cursor: pointer;
+
+    &.btn.btn-block {
+        display: flex;
+    }
 
     &.focus:not(:disabled):not(:global(.disabled)):not([aria-disabled='true']) {
         box-shadow: 0 0 0 2px var(--primary-2);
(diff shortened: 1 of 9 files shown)


client/wildcard/src/components/Button/Button.module.scss
```diff
@@ -132,11 +132,15 @@ input.btn {
     padding: 0;
     background: transparent;
     color: inherit;
-    display: flex;
```
fkling, line 135: @vovakulikov Do you know why icon buttons are "block" elements by default? It seems by default buttons are "inline". There also doesn't seem to be a way to make icon buttons inline.
Maybe just on oversight? I hope screenshot tests catch any issues that would be caused by this change.

vovakulikov (approved):
I tested locally and haven't found any regressions in the react version around icon buttons; since code host integration is no longer published, I think this should be a relatively safe change. 

Thanks for fixing copy button on the search result!

PR #62384, opened 2024-05-02
Svelte: Fix blame decoration line layout

vovakulikov:
| Before | After |
| ------ | ------ |
|  | Screenshot 2024-05-02 at 15 03 34 |

## Test plan
- Manual testing (here is the localhost link from the screenshot - [link](http://localhost:5173/github.com/sourcegraph/sourcegraph@362cc80c4d004eedf37df341d7de743ea324c90d/-/blob/cmd/frontend/graphqlbackend/testing.go?L79-80&view=blame))


fkling (approved):
I wanted to ask whether this even has to use flex layout, but I guess it does to make `text-overflow` work.

PR #62392, opened 2024-05-02
sg: update `cloud get` and `cloud list` to handle new fields

burmudar:
## Test plan
CI

### DIFF
--- a/dev/sg/internal/cloud/instance.go
+++ b/dev/sg/internal/cloud/instance.go
@@ -14,6 +14,14 @@ import (
 
 var ErrLeaseTimeNotSet error = errors.New("lease time not set")
 
+// EphemeralInstanceType is the instance type for ephemeral instances. An instance is considered ephemeral if it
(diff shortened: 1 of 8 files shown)


dev/sg/internal/cloud/instance.go
```diff
@@ -2,85 +2,234 @@ package cloud
...
-		return "in progress"
-	case cloudapiv1.InstanceStatus_INSTANCE_STATUS_FAILED:
-		return "failed"
-	default:
-		return "unknown"
+func parseStatusReason(reason string) (string, string, error) {
+	if reason == "" {
+		return "", "", nil
 	}
+	parts := strings.Split(reason, ",")
```
filiphaftek, line 158: This is something we will probably change later (to be splitted into more parts - stage/description/gotoAction - but I will let you know about the format)

filiphaftek (approved):
👍

PR #62409, opened 2024-05-02
[Svelte]: Fix popover onHover mode

vovakulikov:
@fkling catch that now popover hides too fast when we use it with on hover mode because of standard non-zero offset. This PR just removes offset when the popover is used with onHover mode

## Test plan
- Check popover appearance manually

### DIFF
--- a/client/web-sveltekit/src/routes/styles.scss
+++ b/client/web-sveltekit/src/routes/styles.scss
@@ -29,6 +29,7 @@ input:focus-visible {
 
 kbd {
     display: inline;
+    height: 1.125rem;
     margin: 0 0.125rem;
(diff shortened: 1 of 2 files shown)


client/web-sveltekit/src/routes/styles.scss
```diff
@@ -29,7 +29,6 @@ input:focus-visible {
 
 kbd {
     display: inline;
-    height: 1.125rem;
```
fkling, line 32: I'd deal with `kbd` in a separate PR, I'm not sure we actually want to remove this (idk). See https://github.com/sourcegraph/sourcegraph/pull/62279#discussion_r1588231880

fkling (approved):

PR #62415, opened 2024-05-03
fix: Correctly filter uploads with commit and path existence checks

varungandhi-src:
Fixes GRAPH-563

The upload filtering code previously did checks whether:
- The commit corresponding to the upload actually existing on gitserver
- The path of interest had an associated SCIP document in that upload.

However, the second code path incorrectly only handled the case when
the first operation was a no-op. Consequently, it was possible to have
the following:
- Some uploads were filtered out due to the path not existing in the
   wrong upload (not itself)
- Some uploads were not filtered out due to the path existing in the
   wrong upload (not itself)

This PR is broken up into three commits.
- The first commits adds a negative test for the bug (I've triggered CI)
- The second commit simply adds comments related to aliasing to help
  reason with the code and better understand the bug.
- The third commit removes all micro-optimization related to in-place filtering,
  the aliasing comments, fixes the bug, and updates the test.

## Test plan

- Added regression tests

### DIFF
--- a/internal/codeintel/codenav/service.go
+++ b/internal/codeintel/codenav/service.go
@@ -671,58 +671,34 @@ func (s *Service) GetClosestCompletedUploadsForBlob(ctx context.Context, opts up
 		return nil, err
 	}
 
-	uploadCandidates := copyUploads(candidates)
-	// Disjoint: {uploadCandidates}, {candidates}
-	trace.AddEvent("TODO Domain Owner",
+	trace.AddEvent("InferClosestUploads",
 		attribute.Int("numCandidates", len(candidates)),
-		attribute.String("candidates", uploadIDsToString(uploadCandidates)))
+		attribute.String("candidates", uploadIDsToString(candidates)))
 
 	commitChecker := NewCommitCache(s.repoStore, s.gitserver)
 	commitChecker.SetResolvableCommit(opts.RepositoryID, opts.Commit)
 
-	candidatesWithCommits, err := filterUploadsWithCommits(ctx, commitChecker, uploadCandidates)
-	// Disjoint: {uploadCandidates, candidatesWithCommits}, {candidates}
+	candidatesWithExistingCommits, err := filterUploadsWithCommits(ctx, commitChecker, candidates)
 	if err != nil {
 		return nil, err
 	}
-	trace.AddEvent("TODO Domain Owner",
-		attribute.Int("numCandidatesWithCommits", len(candidatesWithCommits)),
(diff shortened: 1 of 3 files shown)


internal/codeintel/codenav/service.go
```diff
@@ -697,51 +697,34 @@ func (s *Service) GetClosestCompletedUploadsForBlob(ctx context.Context, opts up
 		return nil, err
 	}
 
-	uploadCandidates := copyUploads(candidates)
```
keynmol, line 700: Can you please clarify (in GH comment) what was the copying for and why you no longer need it?

varungandhi-src: The copying was purely defensive. The earlier filtering function modified the slice in-place, so presumably this was to allow looking at the before/after while debugging the code.

varungandhi-src: Similar defensive copying seen elsewhere: 

https://sourcegraph.com/github.com/sourcegraph/sourcegraph@008d572e1e9c79b28d0feaf48b09b9dffb2f1152/-/blob/internal/codeintel/codenav/service.go?L187-190

(we can remove that in a separate PR)

internal/codeintel/codenav/service_closest_uploads_test.go
```diff
@@ -0,0 +1,115 @@
...
+			matchingOptions:       shared.UploadMatchingOptions{Commit: "C2", Path: "a.c"},
+			finalCheck: func(t *testing.T, filtered []shared.CompletedUpload) {
+				hasUploadForWhichPathCheckFails := func() bool {
+					for _, upload := range filtered {
+						if upload.ID == 23 {
+							return true
+						}
+					}
+					return false
+				}()
```
keynmol, line 51: If I squint, this function (and `isMissingUploadForWhichPathCheckShouldSucceed` below) look like `containsUploadWithID(filtered, 23)` and the negation of that.

Perhaps it will be clear to extract a function like that and simplify the respective assertions?

keynmol: Assuming Go doesn't have anything like `filtered.exists(_.ID == 23)`

keynmol: Or even better - replace `finalCheck` with something like a list of IDs you expect to find in the `filtered` result?

varungandhi-src: Tweaked this to move the common logic outside and use a comment instead of the variable name to indicate what is being tested.

keynmol (approved):

PR #62421, opened 2024-05-03
Svelte: Calm down shadows and strip down blame header

taiyab:
# Context

Shadow fixes and other minor styling adjustments to optimise for smaller screens.

# Before

## Light Mode

CleanShot 2024-05-03 at 15 23 21@2x

CleanShot 2024-05-03 at 15 23 37@2x

## Dark Mode

CleanShot 2024-05-03 at 15 24 13@2x

CleanShot 2024-05-03 at 15 24 00@2x

# After

## Light Mode

image

image

## Dark Mode
image

image

# Test plan
Tested locally for visual changes.

### DIFF
--- a/client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte
+++ b/client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte
@@ -232,7 +232,7 @@
         padding: 0.5rem;
         padding-bottom: 0;
         box-shadow: var(--sidebar-shadow);
-        z-index: 1;
+        z-index: 2;
     }
 
     .main {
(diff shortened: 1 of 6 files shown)


taiyab:
> @taiyab, have you tried to make a sidebar panel render its shadow on top of other panels on the right, too (history panel and action header)?

Ugh, I thought I had already done that. Should be sorted.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte
```diff
@@ -232,7 +232,7 @@
         padding: 0.5rem;
         padding-bottom: 0;
         box-shadow: var(--sidebar-shadow);
-        z-index: 1;
+        z-index: 2;
```
Comment (vovakulikov) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte:235]: I usually don't feel good when we have to set `z-index` if it's not 0 or 1 values. 
I don't feel bad here because I will change it with resize panels PR soon so 
no blockers from me here. But for the future we should avoid having any non 0 or 1 values.

vovakulikov (approved):
I'm happy with code and visual changes, approve this and actually merge to avoid 
conflicts with resizable panels PR, thanks (Felix if you have any feedback let me know I can address them in upcoming panels PR)

PR #62430, opened 2024-05-03
Add Cody Pro upgrade script

vdavid:
To make it less magic, I've transformed @chrsmith's Bash script to Go and added it to the repo, with docs.

## Test plan

Tested it but I have problems with my account so the original cURL commands don't work for me either. Will work with @chrsmith to make this work.

### DIFF
--- a/dev/plg-cody-pro-upgrade/BUILD.bazel
+++ b/dev/plg-cody-pro-upgrade/BUILD.bazel
@@ -1,0 +1,14 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
+
+go_library(
+    name = "plg-cody-pro-upgrade_lib",
+    srcs = ["main.go"],
+    importpath = "github.com/sourcegraph/sourcegraph/dev/plg-cody-pro-upgrade",
+    visibility = ["//visibility:private"],
+)
+
+go_binary(
+    name = "plg-cody-pro-upgrade",
+    embed = [":plg-cody-pro-upgrade_lib"],
+    visibility = ["//visibility:public"],
+)
--- a/dev/plg-cody-pro-upgrade/README.md
+++ b/dev/plg-cody-pro-upgrade/README.md
@@ -1,0 +1,23 @@
+# PLG Cody Pro Upgrade script
+
+We use this script to give Cody Pro to our teammates.
+
+## Usage
+
+Run the script using
+
+```bash
+go run add_member.go [NEW_MEMBER_ACCOUNT_ID] [SAMS_SESSION_COOKIE]
+```
+replacing the placeholders with actual values.
+
+## Adding teammates to Cody Pro
+
+Here is the complete process:
+
+1. Use [Redash](https://redash.sgdev.org/queries/614/source) to get the SAMS ID.
+    - If a user doesn't have a `sams_account_id`, it means they don't have a SAMS account because the user hasn't logged into dotcom since the SAMS rollout on 2024-02-08.
+2. Add the SAMS account to the Sourcegraph team
+    - Someone with `cody-admin` SSC superpowers needs to run this.
+    - Go to https://accounts.sourcegraph.com/ and in Dev Tools → Application → Cookies → https://accounts.sourcegraph.com/, copy the value for the `accounts_session` cookie.
+    - Run `sourcegraph/sourcegraph/dev/plg-cody-pro-upgrade/main.go {SAMS_ID} {SESSION_COOKIE}`.
--- a/dev/plg-cody-pro-upgrade/main.go
+++ b/dev/plg-cody-pro-upgrade/main.go
@@ -1,0 +1,79 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+)
+
+func main() {
+	if len(os.Args) != 3 {
+		fmt.Println("Usage: go run script.go SAMS_ID SAMS_SESSION_COOKIE")
+		os.Exit(1)
+	}
+
+	samsID := os.Args[1]
+	samsSessionCookie := os.Args[2] // Get this from your browser's cookies.
+	sscHost := "https://accounts.sourcegraph.com"
+	teamID := "018da9a6-ec3d-7ab9-9468-44fef461607a" // ID of the unofficial Sourcegraph Cody Pro team.
+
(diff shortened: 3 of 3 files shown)


chrsmith (commented):
This PR is not in an acceptable state, and I strongly suggest you do something more valuable with your time.

I appreciate you thoughtfully trying to this script into the repo.  And trying to ensure it is clear and useful. But whatever time you invested working on this would have been better spent on just shipping the Teams and Invites feature, so a script like this wouldn't be necessary in the first place.

## What would be acceptable, IMO

Given that "adding sourcegraph employees to the company Cody Pro team" isn't really something that is generally useful. (e.g. only two people on Earth have the right credentials to run this script.)

But checking in code artifacts so they can be reused is generally a good thing. However, keep in mind that "It doesn't matter how well you polish the underside of the banister." That is to say, we shouldn't put much work into something that will _at best_ be a temporary workaround for a missing feature.

The original Bash script I shared out was the following:

```zsh
#!/bin/zsh

# add_sams_account_to_sourcegraph_cody_pro_team 018d56d8-6946-7892-bf1d-58e7eecd9eb5
function add_sams_account_to_sourcegraph_cody_pro_team() {
    local NEW_MEMBER_ACCOUNT_ID="$1"
    local SSC_HOST="https://accounts.sourcegraph.com"
    # SAMS session token cribbed from your browser.
    local SAMS_SESSION_COOKIE="xxxxxx"
    # ID of the "official" Sourcegraph Cody Pro team.
    local TEAM_ID="018da9a6-ec3d-7ab9-9468-44fef461607a"

    # Creating Cody user account. This step is required for anybody who
    # hasn't visited https://accounts.sourcegraph.com/cody before. (Don't
    # worry about it, it's just a thing we need to do.)
    echo "Creating Cody user account..."
    curl \
        -X PUT \
        -H "Cookie: accounts_session_v2=${SAMS_SESSION_COOKIE}" \
        -H "Content-Length: 0" \
        "${SSC_HOST}/cody/api/rest/admin/user/${NEW_MEMBER_ACCOUNT_ID}"

    # Add the member to the team.
    echo "Adding member to PLG team..."
    curl \
        -X PUT \
        -H "Cookie: accounts_session_v2=${SAMS_SESSION_COOKIE}" \
        -H "Content-Length: 0" \
        "${SSC_HOST}/cody/api/rest/admin/team/${TEAM_ID}/members/${NEW_MEMBER_ACCOUNT_ID}"
}
```

I've been using it for months now, and it's worked just fine. The raw HTTP body printed to the console are sufficiently clear to the operator. And while manually needing to fetch your SAMS session ID is kinda lame, it isn't a major undertaking.

So IMO, it would be sufficient to just check-in as-is. Though expanding on comments would be helpful. (e.g. mentioning that Redash query to look up a user's SAMS ID.)

But beyond that, I don't think any other investment would really be worth it.

## Why this is unacceptable, again, IMO

What you have here adds a lot more "light but generates no heat". You have turned a 29-line Bash script into 130 lines, but without adding any actual value/benefit to the operator.

In fact, I'd argue that this leaves things in a much worse state than if you had just added the script. (Which also belongs in the `self-serve-cody` repo.)

- It adds the `BUILD.bazel` file, in order to "properly" (?) build this trivial Go app. But why? Wouldn't we only ever invoke this via `go run ...`. So why do we add the additional overhead? What benefit is this providing?
- You add the `README.md`, which I greatly appreciate in principle. But for something so trivial -- e.g. a bash script that literally sets some env vars and calls `curl` twice -- why not just expand on the comments there?
- You generated the Go code by asking Chat GPT to translate the script... and it absolutely shows. There is nothing "robust" or "reusable" about that Go app. It's strictly more work to maintain than a trivial Bash script.

All of this is to say:

1. I appreciate and commend your intention to check-in a reusable script for operating Cody Pro. That is a "good" thing, and something I want to encourage. (So thank you.)
2. Introducing a build script and README file for a trivial Bash script is very much overkill, and unnecessary.
3. The AI-generated Code is providing zero benefit beyond a trivial Bash script. And it would be an absolute waste of time for anybody to review the code in order to bring it up to any meaningful standard.

I would recommend that you close this repo, and just add the original Bash script -- with some subset of your comments from the README file -- into the `dev/` folder in the `self-serve-cody` repo.

vdavid:
@chrsmith, I appreciate your opinion on this one.
Don't worry, the script is mainly AI-generated, and I spent very little time with it.
That said, it was very useful to me, and I intend to keep on using it until we have a permanent solution.
It'd be more convenient if I could just merge it so that it's there for me, but of course, I can't do that without an approving review, so if you think there's harm in checking this in, I'll just keep it on a branch.

chrsmith (changes requested):
> Don't worry, the script is mainly AI-generated, and I spent very little time with it.

But you _will_. CI/CD is failing because the code isn't lint-clean, and for good reason.

Spending the time to fix this so that it passes the most basic static analysis is not time well spent.

> That said, it was very useful to me, and I intend to keep on using it until we have a permanent solution.

Absolutely. FWIW, I have the original script checked into a private repo on my machine. So that I can easily `source` and run it as needed.

> It'd be more convenient if I could just merge it so that it's there for me, but of course, I can't do that without an approving review, so if you think there's harm in checking this in, I'll just keep it on a branch.

Yes, I _literally_ believe that checking this code as-is into the branch would do harm. Keep this alive in a branch as long as you need.

There are some things you can cut corners on, or make tradeoffs for. But IMO, you must _never_, cut corners when it comes to security, reliability, or code quality.

Yes, practically speaking, this is just 100 lines of Go so #yolo #wcgw. But as an engineering organization, we must hold ourselves to a higher standard. And also, because we're better than that.

Because sometimes that 100 lines of ":shrug: it's just temporary hack" becomes load-bearing. And you find yourself with a codebase that is extremely difficult to safely modify.

vdavid:
@chrsmith I can accept the claim that this is not worth out time because it adds less value than what we could do with that time.
But I disagree with the claim that this is "cutting corners" or "not meeting some hifher standards". I do prefer this script to the bash script that preceded it, I think it is objectively better documented, and it decreases bus factor by a lot compared to the bash script in a private repo on your machine. I am happy that this exists in the repo now (even if just on a branch) because this way it can be shared with others.
And as far as I took a quick glance at it, there are no linter errors just a gofmt missing.
I don't want to spend more time with this just wanted to avoid confusing "not worth our time" with "subpar quality".

PR #62431, opened 2024-05-03
svelte: Improve page title updates

### LINKED ISSUE #62429: Use a different approach for setting the page title
The current one is broken, sometimes it's not updated properly. It's probably better to have each page return a title from the `load` function and use that value in the root layout.

fkling:
Closes #62429 

I originally thought we have to create the page title in data loaders and set it only in the root layout, but it appears that setting the title in the layout files is what's potentially problematic (sometimes the page title "reverts" to what's being set by the root layout when navigating to a new page).

Only setting the page title in `+page.svelte` files seems to solve the issue. At least for now I prefer that because it makes setting the page title very explicit and consistent.

I also changed the file preview logic to use the (newly refactored) component instead of the page component. Otherwise opening the preview updates the page title, which is not correct.

 Note: One might think that having the title set in loaders would allow
 us to "compose" the title hierarchically more easily. In theory that's
 true, but having to `await` the parent's data is annoying too.

## Test plan

Manual testing by navigating to different pages and verifying that the title updates.

### DIFF
--- a/client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/-/commits/[...path]/+page.svelte
+++ b/client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/-/commits/[...path]/+page.svelte
@@ -46,10 +46,18 @@
     $: if ($commitsQuery?.data?.repository?.commit?.ancestors) {
         commits = $commitsQuery.data.repository.commit.ancestors
     }
+    $: pageTitle = (() => {
+        const parts = ['Commits']
+        if (data.path) {
+            parts.push(data.path)
+        }
+        parts.push(data.displayRepoName, 'Sourcegraph')
+        return parts.join(' - ')
+    })()
 </script>
 
 <svelte:head>
-    <title>Commits - {data.displayRepoName} - Sourcegraph</title>
+    <title>{pageTitle}</title>
 </svelte:head>
 
 <section>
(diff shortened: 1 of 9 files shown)


client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/-/commits/[...path]/+page.svelte
```diff
@@ -46,10 +46,18 @@
     $: if ($commitsQuery?.data?.repository?.commit?.ancestors) {
         commits = $commitsQuery.data.repository.commit.ancestors
     }
+    $: pageTitle = (() => {
+        const parts = ['Commits']
+        if (data.path) {
+            parts.push(data.path)
+        }
+        parts.push(data.displayRepoName, 'Sourcegraph')
+        return parts.join(' - ')
```
Comment (bahrmichael) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/-/commits/[...path]/+page.svelte:55]: Would this be the same as `Commits - ${data.path} - Sourcegraph`? No need to change, just making sure I understand this right. Maybe this could use a unit test though.

bahrmichael (approved):
LGTM

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/-/commits/[...path]/+page.svelte
Comment (fkling) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/-/commits/[...path]/+page.svelte:55]: Only if `data.path` was always defined, but it's optional. I.e. `repo/-/commits` is a valid route.

PR #62450, opened 2024-05-06
svelte: Give last commit message as much space as possible

fkling:
I don't see a reason why we need to restrict the width to a specific pixel value. I find it annoying when the message is cut off even if enough space is available. There are already conventions about the size of the subject line so most of the time the message won't take up a lot of space anyway.

For smaller screens we should think of a different solution anyway IMO.

We could still restrict it to a relative value of the available width if we wanted to.

| Before | After |
|--------|--------|
| |  | 

## Test plan

Manual testing.


fkling:
Not relevant anymore

PR #62452, opened 2024-05-06
feat: popover to configure editor

### LINKED ISSUE #62292: Add config popover to "open in editor" button
Follow up to [https://github.com/sourcegraph/sourcegraph/issues/61629](https://github.com/sourcegraph/sourcegraph/issues/61629).

In [https://github.com/sourcegraph/sourcegraph/pull/61782](https://github.com/sourcegraph/sourcegraph/pull/61782) I added an OpenInEditor button for the new webapp. To reduce scope for the ticket I decided to not re-implement the popover that gives users a convenient way to configure an editor. Currently they see a tooltip that tells them to go to the user settings and fill out some JSON over there.

Ideally we would re-implement this popover and our new webapp. We can probably copy&paste a bunch of code from the `OpenInEditorAction.tsx` from the react app.

bahrmichael:
Closes #62292

This PR introduces a new popover to configure the editor. 

~I had to extend the initial settings to get the current settings, introduce new methods to update settings, and handle the two-step update (because each SettingsEdit can only modify one field). With this solution we get the same look-and-feel like the React app, and the editor icon updates without a page reload.~ 

There's a new `updateUserSetting` on the root layout, with which this change and other features can update user settings. It fetches the `lastID` before writing a change, which means it will force the given setting and won't recheck if something changed.

https://github.com/sourcegraph/sourcegraph/assets/1830132/4d53c2b3-ed76-4195-bfc4-8207ecad0427

## Test plan

Manual testing, see video above


bahrmichael:
The current setup works, but I'm not very happy with it. The main problem is that `editSettings` needs the latest settings ID (which comes from the subjects), and can only modify one field at a time.

So to get the lastID when starting out (to make sure that the user's settings haven't been modified in another tab or so), we need to get the lastID that was present when the user open the popover. The lastID changes when we make an edit call, so we have to request it again in between.

I'm hoping there's another way to make this work nicely, but couldn't find one yet. Maybe there's a better Svelte-style approach?

client/web-sveltekit/src/lib/repo/open-in-editor/OpenInEditor.svelte
```diff
@@ -1,32 +1,83 @@
...
+            keyPath: [{property: 'openInEditor'}, {property: 'editorIds'}],
+        });
+        lastId.set(newLastId2);
+        isSaving = false;
+
+        openInEditor = {
+            editorIds: [$selectedEditorId],
+            'projectPaths.default': $defaultProjectPath,
+        }
+    }
```
fkling, line 71: I think it would make sense to abstract some of details of updating settings. E.g. we could provide an API that actually takes an arbitrary number of fields so that the details of the lastId stuff is hidden.
For reference, here is the React version: https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@cb03009ca0fb1c504f3903bf48e7f5531b6d1cb4/-/blob/client/web/src/open-in-editor/OpenInEditorActionItem.tsx?L80-95

fkling (commented):
I left a couple of comments inline. This might be a tricky one because I assume you just want to get this "done", but at the same time its introducing new behavior (updating user settings) for which it makes sense to have a reasonable API, instead of just doing it "ad hoc". Let me know what you think!

bahrmichael:
> I left a couple of comments inline. This might be a tricky one because I assume you just want to get this "done", but at the same time its introducing new behavior (updating user settings) for which it makes sense to have a reasonable API, instead of just doing it "ad hoc". Let me know what you think!

I'm happy to take the hard route. This was my attempt to get something done, so that we can iterate together on making it nice :)

client/web-sveltekit/src/lib/repo/open-in-editor/OpenInEditor.svelte
```diff
@@ -1,32 +1,83 @@
...
     $: sourcegraphBaseURL = new URL($page.url).origin
 
-    $: ({ repoName, filePath, position, range } = parseBrowserRepoURL($page.url.toString()))
+    $: ({repoName, filePath, position, range} = parseBrowserRepoURL($page.url.toString()))
     $: start = position ?? range?.start
+
+    $: lastId = writable<number>(subjects.at(-1)?.latestSettings?.id);
+    $: subjectId = writable<string>(subjects.at(-1)?.id);
+    $: defaultProjectPath = writable<string>('');
+    $: selectedEditorId = writable<typeof editorIds[number] | undefined>();
```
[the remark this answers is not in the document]
bahrmichael: That makes sense! I was having fun with stores, and maybe a bit too much :D

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.ts
```diff
@@ -111,6 +114,29 @@ async function loadFileView({ parent, params, url }: PageLoadEvent) {
...
+            const settingsResult = await client.query(LatestSettingsQuery, {})
+            if (!settingsResult.data || settingsResult.error) {
+                throw new Error(`Failed to fetch latest settings during editor update: ${settingsResult.error}`)
+            }
+            const newLastId = settingsResult.data.viewerSettings.subjects.at(-1)?.latestSettings?.id
+            if (!newLastId) {
+                throw new Error('Failed to get new last ID from settings result')
+            }
+            return newLastId
+        },
```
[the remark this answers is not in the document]
Comment (bahrmichael) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.ts:139]: If I can just refresh the IDs it makes things a lot easier! I'll push a new version where I was able to remove a lot of code, and pull the lastID before sending a request.

client/web-sveltekit/src/lib/repo/open-in-editor/OpenInEditor.svelte
bahrmichael: I'm pushing a new version that hides the lastID, but with a different approach.

bahrmichael:
@fkling I've done another pass, and it looks much cleaner now. The main change was to drop the initial subjects fetch, assuming that we can fetch the lastID before submitting the SettingsEdit.

I'm happy to merge this if you agree.

client/web-sveltekit/src/lib/repo/open-in-editor/OpenInEditor.svelte
```diff
@@ -37,23 +79,66 @@
...
+            </span>
+        </Tooltip>
+        <div slot="content" class="open-in-editor-popover">
+            <form on:submit={handleEditorUpdate} novalidate>
+                <h3>Set your preferred editor</h3>
+                <p>
+                    Open this and other files directly in your editor. Set your path and editor to get started. Update
+                    any time in your user settings.
+                </p>
+                <p class="form-label">Default projects path</p>
```
fkling, line 103: I think this should be an actual `<label>` element, associated with the `input` element.

client/web-sveltekit/src/lib/repo/open-in-editor/OpenInEditor.svelte
```diff
@@ -37,23 +79,66 @@
...
+                    spellcheck={false}
+                    bind:value={defaultProjectPath}
+                    class="form-input"
+                />
+                <p class="small form-info">
+                    The directory that contains your repository checkouts. For example, if this repository is
+                    checked out to <code>/Users/username/projects/cody</code>, then set your default projects path
+                    to <code>/Users/username/projects</code>.
+                </p>
+                <p class="form-label editor-label">Editor</p>
```
fkling, line 121: I think this should be an actual `<label>` element, associated with the `select` element.

client/web-sveltekit/src/lib/repo/open-in-editor/OpenInEditor.svelte
```diff
@@ -37,23 +79,66 @@
...
 {:else if editorSettingsErrorMessage}
-    <Tooltip tooltip={editorSettingsErrorMessage}>
-        <a href="/help/integration/open_in_editor" target="_blank">
-            <DefaultEditorIcon />
-            <span data-action-label> Editor </span>
-        </a>
-    </Tooltip>
+    <Popover let:registerTrigger let:toggle placement="left-start">
+        <Tooltip tooltip="Set your preferred editor">
+            <span use:registerTrigger on:click={() => toggle()}>
```
fkling, line 91:
```suggestion
            <span use:registerTrigger on:click={toggle}>
```

client/web-sveltekit/src/lib/repo/open-in-editor/OpenInEditor.svelte
```diff
@@ -37,23 +79,66 @@
...
+                <p>
+                    Open this and other files directly in your editor. Set your path and editor to get started. Update
+                    any time in your user settings.
+                </p>
+                <p class="form-label">Default projects path</p>
+                <input
+                    id="OpenInEditorForm-projectPath"
+                    type="text"
+                    name="projectPath"
+                    placeholder="/Users/username/projects"
```
fkling, line 108: Future improvement: Use different placeholder values for different operation systems.

fkling (approved):
lgtm, just some nits inside.

client/web-sveltekit/src/lib/repo/open-in-editor/OpenInEditor.svelte
bahrmichael: Sveltecheck doesn't like this:

```
Svelte: Type (open?: boolean | undefined) => void is not assignable to type MouseEventHandler<HTMLSpanElement
```

fkling: Ah, because of the argument. OK!

PR #62454, opened 2024-05-06
[Svelte]: Bring resize panels to blob page

vovakulikov:
This PR simply adds new resizable panels to the blob UI page and makes history and reference panels resizable.

Small loom demo https://www.loom.com/share/1efe0ce1b2a54a0e885129089372dc2d?sid=aae37f4b-0ddd-4161-bd82-5a9795ff64e1

## Changes
- Resizable panel components got imperative API, this means that now you can call expand/collapse programmatically. 
- Panel remembers its size before it was collapsed, if other layout configuration hasn't been changed (same amount of panels in the group with same ordering)
- Panel component got new set of callbacks (resize, expand, collapse callbacks)
- You also can read isCollapsed state through slow let varialble
- Previosly we did some work around shadows ans setting z-indexes for proper shadow layout, this corrupted some of resize hanldels, this PR fixes them by fixing z-index layout. 

## Test plan
- Check the blob page and try to open/close the history or reference panel
- Also try to open it in different tabs and check that the bottom panel remembers the previous state


client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte
```diff
@@ -137,110 +124,166 @@
...
-        gap: 0.5rem;
-        justify-content: space-between;
-        align-items: center;
-        margin-bottom: 0.5rem;
+    // Forcing the left sidebar panel (file-tree) to be over
+    // right content panel to make sure that box-shadow of the
+    // sidebar is rendered over content panel.
+    :global([data-panel-id='sidebar-panel']) {
+        z-index: 1;
+        position: relative;
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte:263]: What do you we need `position: relative` for?

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte
```diff
@@ -265,21 +329,13 @@
...
-            height: 32vh;
-            // Disable flex layout so that tabs simply fill the available space
-            display: block;
 
-            .last-commit {
-                display: none;
-            }
+        :global([data-tabs]) {
+            width: 100%;
+            height: 100%;
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte:338]: Height is already set by the tabs component. Should we maybe also just set `width: 100%` there?

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte
```diff
@@ -137,110 +124,166 @@
...
+    // sidebar is rendered over content panel.
+    :global([data-panel-id='sidebar-panel']) {
+        z-index: 1;
+        position: relative;
+        box-shadow: var(--sidebar-shadow);
+    }
+
+    :global([data-panel-id='content-panel']) {
+        z-index: 0;
+        position: relative;
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte:269]: I understand that `position: relative` and `z-index` go together here, but can you explain what this is used for?

fkling (approved):
Could you summarize/explain the changes you made to the resizable panels logic?

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte
Comment (vovakulikov) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte:263]: If you're asking about position:relative reason is it because without new stacking context z-index does nothing and position:relative creates this context. 

If you're asking about reason for this z-index:1 is about bringing right side shadow which is set to the sidebar panel over the next (content) panel.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte
Comment (vovakulikov) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte:338]: True, removed this

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte
Comment (vovakulikov) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte:269]: So the content panel has a few elements with custom z-index within for proper shadows (like breadcrumbs and Code/Blame action row). To ensure that none of these elements float outside the content-panel and overlaps with other elements I create here another stacking context and set z-index to 0 to ensure that sidebar element which has z-index:1 will be rendered over the content-panel elements. 

This is needed to ensure that side-bar file tree panel shadows are rendered over the content panel. 

I left a comment about this a bit above in the code, let's me know if it's not clear

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte
Comment (fkling) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte:263]: Would using `isolation: isolate` have the same effect? Just wondering because `position: relative` does more than just create a new stack context, and it's not necessarily obvious (to others) what it does.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte
```diff
@@ -137,110 +124,166 @@
         } else if (selectedTab === TabPanels.References) {
             // Close references panel when navigating to a URL that doesn't have the 'references' view state
             selectedTab = null
+            bottomPanel.collapse()
```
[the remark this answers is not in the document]
Comment (vovakulikov) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte:127]: Nope, removed this

PR #62456, opened 2024-05-06
sg: cloud ephemeral check if deployment exists already

burmudar:
Before trying to create new deployment we check whether the deployment exists already. If it does
1. Notify user that deployment will not proceed
2. Give alternative methods for creating a deployment

Screenshot 2024-05-07 at 15 50 03

## Test plan
Tested locally


dev/sg/internal/cloud/deploy_command.go
```diff
@@ -37,6 +39,15 @@ var DeployEphemeralCommand = cli.Command{
 	},
 }
 
+func deployUpgradeSuggestion(name, version string) string {
+	var text = "You might want to try one of the following:\n" +
```
burmudar, line 43: We have to do this since escaping the back ticks doesn't work and it's annoying otherwise

dev/sg/internal/cloud/deploy_command.go
```diff
@@ -138,6 +149,45 @@ func checkVersionExistsInRegistry(ctx context.Context, version string) error {
 	return nil
 }
 
+func cancelBuild(ctx context.Context, build *buildkite.Build) error {
```
burmudar, line 152: can remove - it's not used anymore

dev/sg/internal/bk/bk.go
```diff
@@ -212,6 +212,10 @@ func (c *Client) GetJobAnnotationsByBuildNumber(ctx context.Context, pipeline st
 	return result, nil
 }
 
+func (c *Client) CancelBuild(ctx context.Context, org, pipeline, number string) (*buildkite.Build, error) {
```
burmudar, line 215: Going to keep this - just makes the buildkite client a little more complete

Strum355 (approved):

Issue #62428, opened 2024-05-03
Audit global styles


fkling:
Currently we import most (all?) of the existing global styles of the web app. That's probably not necessary and we should double check what we really need.

E.g. a common problem is using the `.container` class for the root element of a component, but that is also a global class which applies additional styles, which is surprising.

PR #62457, opened 2024-05-06
svelte: Reduce imported global CSS

fkling:
Part of [#62428](https://github.com/sourcegraph/sourcegraph/issues/62428)

Until now we've simply imported the global `base.scss` file, which in turn imports a bunch of other files and defines global CSS classes. Many (most?) of them are actually not used.

This commit reduces to the number of imported files to the ones that contain styles that are used. My methodology was to inspect every file that `base.scss` imports and find out whether it defines custom properties (variables) or global classes. In some cases I extracted variables into their own files.

I can't guarantee that the included files only declare styles that are used, but it's a first step.

## Test plan

Compared all pages "side-by-side", switching between S2 and local tabs to catch any "drift", including the revision selector, search results popovers and search input suggestions.

### DIFF
--- a/client/web-sveltekit/src/lib/TreeNode.svelte
+++ b/client/web-sveltekit/src/lib/TreeNode.svelte
@@ -35,13 +35,11 @@
 </script>
 
 <li
-    class="treeitem"
-    class:selectable
-    class:selected
     role="treeitem"
     aria-selected={selectable ? selected : undefined}
     aria-expanded={expandable ? expanded : undefined}
     {tabindex}
+    data-treeitem
     data-node-id={nodeID}
 >
     <span class="label" data-treeitem-label>
@@ -65,11 +63,11 @@
     </span>
     {#if expanded && children}
         {#await children}
-            <div class="ml-4">
+            <div class="loading">
                 <LoadingSpinner center={false} />
             </div>
         {:then treeProvider}
-            <ul role="group" class="ml-2">
+            <ul role="group">
                 {#each treeProvider.getEntries() as entry (treeProvider.getNodeID(entry))}
                     <svelte:self {entry} {treeProvider} let:entry let:toggle let:expanded>
                         <slot {entry} {toggle} {expanded} />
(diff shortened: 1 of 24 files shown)


client/web-sveltekit/src/lib/TreeNode.svelte
fkling, line 1: Made a couple of changes here and related files that removed the use of classes for "outer component" styling.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/-/branches/all/+page.svelte
```diff
@@ -110,12 +116,6 @@
         }
     }
 
-    form,
-    div,
-    :global([data-scroller]) {
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/-/branches/all/+page.svelte:115]: This was affecting all scroll instances, which is wrong.

client/web-sveltekit/src/lib/styles/dropdown.scss
```diff
@@ -0,0 +1,43 @@
+:root {
```
vovakulikov, line 1: Why do we have these variables on the global level? Should it be incorporated with a special Dropdown component instead of variables?

client/web-sveltekit/src/lib/styles/popover.scss
```diff
@@ -0,0 +1,9 @@
+:root {
```
vovakulikov, line 1: Same question as about dropdown variables: why do we need to have them on the global level instead of having them on the popover level? 

We still can have some global variable that can effect popover variables like different box-shadows levels for instance.

vovakulikov:
Havent' tested this manually yet, so just left a few questions, will review it properly by the end of the day

client/web-sveltekit/src/lib/styles/dropdown.scss
vovakulikov: Agree, maybe we should leave a comment here that these variables are used at the moment but they shouldn't be used in any new components freely

vovakulikov (approved):
Checked this manually, haven't noticed any UI regressions, thanks

PR #62463, opened 2024-05-06
release(upgradetest): confirm postrelease version is newer than the latest sourcegraph version

BolajiOlajide:
Last week, I ran into an obscure error where an internal release build had it's release tests failing. 

[Context](https://sourcegraph.slack.com/archives/C05DWT4ANHH/p1715011866175999)

This led to an hypothesis where we believe the version I was trying to release `5.3.666` was assumed by the tests to be behind the latest version `5.3.12303`. I added this check to ensure this doesn't repeat itself.

## Test plan

Will test during the release testing later.

### DIFF
--- a/testing/tools/upgradetest/tests-util.go
+++ b/testing/tools/upgradetest/tests-util.go
@@ -800,18 +800,20 @@ func handleVersions(cCtx *cli.Context, overrideStd, overrideMVU, overrideAuto []
 			continue
 		}
 		validTags = append(validTags, v)
-	}
 
-	// Get latest Version and latestMinorVersion
-	for _, tag := range validTags {
 		// Track latest full version
-		if latestFullVer == nil || tag.GreaterThan(latestFullVer) {
-			latestFullVer = tag
+		if latestFullVer == nil || v.GreaterThan(latestFullVer) {
+			latestFullVer = v
 		}
 		latestMinorVer, err = semver.NewVersion(fmt.Sprintf("%d.%d.0", latestFullVer.Major(), latestFullVer.Minor()))
 		if err != nil {
 			return nil, nil, nil, nil, nil, nil, err
 		}
+	}
+
+	semverPostRelease := semver.MustParse(postRelease)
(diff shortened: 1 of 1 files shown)


testing/tools/upgradetest/tests-util.go
```diff
@@ -800,20 +800,22 @@ func handleVersions(cCtx *cli.Context, overrideStd, overrideMVU, overrideAuto []
...
-		if latestFullVer == nil || tag.GreaterThan(latestFullVer) {
-			latestFullVer = tag
+		if latestFullVer == nil || v.GreaterThan(latestFullVer) {
+			latestFullVer = v
 		}
 		latestMinorVer, err = semver.NewVersion(fmt.Sprintf("%d.%d.0", latestFullVer.Major(), latestFullVer.Minor()))
 		if err != nil {
 			return nil, nil, nil, nil, nil, nil, err
 		}
 	}
```
BolajiOlajide, line 812: @DaedalusG I took the liberty of tweaking the logic here as I don't think we need the second loop here. Moving this logic after appending to `validTags` will achieve the same result albeit faster.

Let me know if you have concerns

testing/tools/upgradetest/tests-util.go
```diff
@@ -800,20 +800,22 @@ func handleVersions(cCtx *cli.Context, overrideStd, overrideMVU, overrideAuto []
...
 
+	semverPostRelease := semver.MustParse(postRelease)
```
DaedalusG, line 814: We should probably wrap this in a check to make sure the postRelease flag is set. There are ways to run this test without specifying postRelease targets

BolajiOlajide: Oh I see. Added an `if` check to only do this is the `postRelease` flag is set.

BolajiOlajide: @DaedalusG Can you re-review when you have some time?

DaedalusG (approved):

PR #62466, opened 2024-05-06
Executors: run script as script, not as command

camdencheek:
This fixes an issue that is generating failures on repos with spaces in their names.

Previously, we were passing the script name as a shell command with `-c`. However, that means it's subject to shell escaping. If we instead pass the file as a script for `sh` to run, it avoids the need to escape the file name and also reduces the surface area for injection attacks.

[Slack thread](https://sourcegraph.slack.com/archives/C02MR5PPMKJ/p1714772085831219)

## Test plan

Updated unit tests. I still want to test this end-to-end, but I'm struggling to get this all built and running locally, so I'm just going to test the image after-merge. It's a very small, high-confidence change, so I'm not terribly worried.

### DIFF
--- a/cmd/executor/internal/worker/runtime/kubernetes.go
+++ b/cmd/executor/internal/worker/runtime/kubernetes.go
@@ -67,8 +67,8 @@ func (r *kubernetesRuntime) NewRunnerSpecs(ws workspace.Workspace, job types.Job
 				Key:  key,
 				Name: strings.ReplaceAll(key, ".", "-"),
 				Command: []string{
-					"/bin/sh -c " +
-						filepath.Join(command.KubernetesJobMountPath, files.ScriptsPath, scriptName),
+					"/bin/sh",
+					filepath.Join(command.KubernetesJobMountPath, files.ScriptsPath, scriptName),
 				},
 				Dir:   step.Dir,
 				Env:   step.Env,
@@ -90,7 +90,6 @@ func (r *kubernetesRuntime) NewRunnerSpecs(ws workspace.Workspace, job types.Job
 						Name: strings.ReplaceAll(key, ".", "-"),
 						Command: []string{
 							"/bin/sh",
-							"-c",
 							filepath.Join(command.KubernetesJobMountPath, files.ScriptsPath, ws.ScriptFilenames()[i]),
 						},
 						Dir:       step.Dir,
--- a/cmd/executor/internal/worker/runtime/kubernetes_test.go
+++ b/cmd/executor/internal/worker/runtime/kubernetes_test.go
@@ -58,7 +58,7 @@ func TestKubernetesRuntime_NewRunnerSpecs(t *testing.T) {
 					{
 						Key:       "step.kubernetes.key-1",
 						Name:      "step-kubernetes-key-1",
-						Command:   []string{"/bin/sh", "-c", "/job/.sourcegraph-executor/script.sh"},
+						Command:   []string{"/bin/sh", "/job/.sourcegraph-executor/script.sh"},
 						Dir:       ".",
 						Env:       []string{"FOO=bar"},
 						Operation: operations.Exec,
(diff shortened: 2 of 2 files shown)


cmd/executor/internal/worker/runtime/kubernetes.go
```diff
@@ -67,8 +67,8 @@ func (r *kubernetesRuntime) NewRunnerSpecs(ws workspace.Workspace, job types.Job
 				Key:  key,
 				Name: strings.ReplaceAll(key, ".", "-"),
 				Command: []string{
-					"/bin/sh -c " +
-						filepath.Join(command.KubernetesJobMountPath, files.ScriptsPath, scriptName),
+					"/bin/sh",
+					filepath.Join(command.KubernetesJobMountPath, files.ScriptsPath, scriptName),
```
camdencheek, line 71: I stumbled across this in the tests. It's not obvious to me that this ever worked since, elsewhere, we use the first element of `Command` as the executable name.

eseliger: Good spot, this seems wrong to me as well.

eseliger (approved):

cmd/executor/internal/worker/runtime/kubernetes.go
peterguy: Concur: looks like it should alway fail with `/bin/sh -c ...: no such file or directory` (see [this PoC](https://go.dev/play/p/wUPUykGo-2H).

Based on what I'm seeing as I trace the execution path for this code, enabling `KUBERNETES_SINGLE_JOB_POD` should have resulted in this error all of the time. Am I seeing that wrong?

camdencheek: > enabling KUBERNETES_SINGLE_JOB_POD should have resulted in this error all of the time

That's also my read. Based on some other comments in the code, it looks like `KUBERNETES_SINGLE_JOB_POD` was never quite finished

eseliger: yeah that's what I alluded to in some slack thread recently, it doesn't seem like the k8s executors work has been fully finished before the team left :/ (@peterguy you asked about the TODOs I meant, this is them :D)

peterguy: Heh, This is the first `TODO` I saw as I was looking into them recently based on [your comment](https://sourcegraph.slack.com/archives/C02MR5PPMKJ/p1712710026677489?thread_ts=1712591330.697389&cid=C02MR5PPMKJ) about them, @eseliger :-).
I'm just a bit confused about `KUBERNETES_SINGLE_JOB_POD` and this code because toggling `KUBERNETES_SINGLE_JOB_POD` on apparently [solves some problems](https://sourcegraph.slack.com/archives/C065U3U5FTP/p1710361794977859?thread_ts=1710267405.261839&cid=C065U3U5FTP) with Batch Changes and bitbucket repos. Granted, there's really low signal about that environment variable in internal discussions, so maybe that example is a red herring.

peterguy: Followup on `KUBERNETES_SINGLE_JOB_POD`: it's used to address the issue of no `root` access on a K8s cluster, which is required for running [the "git safe directory" command](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@d48851738332de346fce7c4850c7f9c2b15a2591/-/blob/cmd/batcheshelper/run/post.go?L36-41). The recommended workaround for that in the [IE playbook](https://docs.google.com/document/d/1iSbI6pjEArzu_Spwterk3zC93kqbij6_RnAIVFdbGcc/edit#heading=h.xi8zfv917izg) is to toggle on `KUBERNETES_SINGLE_JOB_POD` so that the pod role can be updated to allow permissions. It could be that `EXECUTOR_ADD_SAFE` was toggled off along the way ([it defaults to `true`](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@d48851738332de346fce7c4850c7f9c2b15a2591/-/blob/cmd/batcheshelper/main.go?L165-176)), and that bypassed changing the `git` config.

Issue #61941, opened 2024-04-16
Svelte: Implement new symbol kind icons


taiyab:
We have a new set of symbol icons ready for first draft implementation.

Open to change these if they're still not doing a great job at representing each symbol kind.

<img width="1144" alt="image" src="https://github.com/sourcegraph/sourcegraph/assets/5337876/b9098c51-be77-443d-867f-3fdb8859150b">

Figma: https://www.figma.com/file/EPgl0qVbqGSumvf52brhda/Code-Search?type=design&node-id=3027%3A85653&mode=design&t=hKVca2TmuCChXvqd-1

taiyab:
There are two things we can improve on going forwarrd:
- Color system
- Increasing clarity of icons with a few alternative versions

The colors I've used above are [based on what family the symbol kind belongs to](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@42965898db22c0a743b1903e1509581ba218c976/-/blob/client/web-sveltekit/src/lib/search/SymbolKind.svelte?L2:14-2:24), however, this might differ from the existing color scheme, which we need to resolve.

PR #62473, opened 2024-05-06
Svelte: implement symbol redesign

camdencheek:
This implements the designs for the new symbol kind icons [here](https://www.figma.com/file/EPgl0qVbqGSumvf52brhda/Code-Search?type=design&node-id=3027%3A85653&mode=design&t=hKVca2TmuCChXvqd-1).

screenshot-2024-05-07_16-51-12@2x
screenshot-2024-05-07_16-50-41@2x
screenshot-2024-05-07_17-03-31@2x
screenshot-2024-05-07_17-04-00@2x

Closes https://github.com/sourcegraph/sourcegraph/issues/61941

## Test plan

Added storybook test. Manually tested the two places this was slotted in.


taiyab:
I checked out the branch here but didn't see any of the new symbol icons in search results, etc.

The icons visually look like they use actual text over being true outlined svgs? Apologies if this is wrong, I can't clearly see where the source for each individual icon is.

taiyab:
> Did you want us to just copy the paths straight from the designs? I was thinking we wanted to parameterize these so we could easily adjust the short names, colors, spacing, etc. Totally fine if that's what you want, just misunderstood 🙂

As long as it's strictly not adopting any styling from the stylesheets, then fine. To my eye I'm noticing some font-weight weirdness/faux bolding but could be wrong.

taiyab (commented):
LGTM!

camdencheek:
Converting to draft since we decided to just use the SVGs from Figma for now. We can't guarantee that our code font exists (it didn't on my system 🙃), so until we get fonts figured out, we're just going to use what we know looks good, which are the paths from Figma

camdencheek:
Alrighty, this is ready for review again

taiyab (approved):
**Much** better.

taiyab:
Excited about getting this in! :D

client/web-sveltekit/src/lib/search/SymbolKindIcon.svelte
```diff
@@ -0,0 +1,154 @@
...
+</Tooltip>
+
+<style lang="scss">
+    div {
+        display: contents;
+
+        // TODO(@taiyab): incorporate these colors into the semantic colors
+
+        :global(.theme-light) & {
+            --icon-color: var(--text-muted);
```
vovakulikov, line 117: Can we move --icon-color: var(--text-muted) on the top level, it seems we set it 
both times for dark and light theme

vovakulikov (approved):

client/web-sveltekit/src/lib/search/SymbolKindIcon.svelte
camdencheek: Done!

taiyab:
Quick note on this @camdencheek: The symbols weren't updated in places like the search suggestions:
CleanShot 2024-05-09 at 19 48 54@2x

There might've been other places missed out.

camdencheek:
Oh weird. Those must use a different source for the icons because I deleted the code that should be handling that 😄

Issue #62471, opened 2024-05-06
Unexpected panic when running `sg lint`


chrsmith:
Not going to lie. I feel really bad that my current branch is so messed up that the `sg` tool literally dies when trying to look at it. 😭 (And this does seem reproducible. I'll keep the commit locally in case you need it for a repro.)

The full output:

```
 ~/go/src/github.com/sourcegraph/sourcegraph % sg lint --fix client
👉 Fixing checks from target: client, format
✅ Running checks  ██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████  100%
│                                                                                                                                                                                              
├── 1. client  Running checks...                                                                                                                                                            34s
└── 2. format  Done!                                                                                                                                                                        34s
⚠️ Encountered panic - please open an issue with the command output:
        https://github.com/sourcegraph/sourcegraph/issues/new?template=sg_bug.md
❌ panic: panic: runtime error: invalid memory address or nil pointer dereference
stacktrace:
goroutine 96 [running]:
runtime/debug.Stack()
        runtime/debug/stack.go:24 +0x64
github.com/sourcegraph/conc/panics.NewRecovered(0x1, {0x109ec6d60, 0x10ced2250})
        github.com/sourcegraph/conc@v0.3.1-0.20240108182409-4afefce20f9b/panics/panics.go:59 +0x74
github.com/sourcegraph/conc/panics.(*Catcher).tryRecover(0x14002da3330)
        github.com/sourcegraph/conc@v0.3.1-0.20240108182409-4afefce20f9b/panics/panics.go:28 +0x58
panic({0x109ec6d60?, 0x10ced2250?})
        runtime/panic.go:770 +0x124
go.opentelemetry.io/otel/sdk/trace.(*recordingSpan).End.deferwrap1()
        go.opentelemetry.io/otel/sdk@v1.24.0/trace/span.go:405 +0x2c
go.opentelemetry.io/otel/sdk/trace.(*recordingSpan).End(0x14001600180, {0x0, 0x0, 0x14002f539b0?})
        go.opentelemetry.io/otel/sdk@v1.24.0/trace/span.go:443 +0x8ac
panic({0x109ec6d60?, 0x10ced2250?})
        runtime/panic.go:770 +0x124
[...]

PR #62476, opened 2024-05-07
sg: skip honey event duration if event is nil

burmudar:
`runScripSerialized` worked fine if you ran `pnpm dedupe` because it was 'instrumented' with `timeCheck` but if you didn't and executed `runScriptSerialized` without `timeCheck` the context won't be Honeycomb enriched and thus the event will be nil and then the whole method falls apart.

Previously this also just checked the locking duration of the mutex 🤔 and I assume it wanted to check the duration of the `pnpm dedupe`?

Either way, it now checks if the event is nil before using it.

I think this addresses #62471 and https://github.com/sourcegraph/devx-support/issues/933
## Test plan
Tested locally


dev/sg/linters/linters.go
```diff
@@ -133,6 +133,11 @@ func runScript(name string, script string) *linter {
 	return &linter{
 		Name: name,
 		Check: func(ctx context.Context, out *std.Output, args *repo.State) error {
+			defer func() {
```
Strum355, line 136: TODO: rm

Strum355:
Not sure if we even still need any of this 🤔 we could maybe delete this as we dont check honeycomb stats for sg lint anymore

dev/sg/linters/linters.go
burmudar: :wat: I thought I remove all of this 🤣

burmudar:
> Not sure if we even still need any of this 🤔 we could maybe delete this as we dont check honeycomb stats for sg lint anymore

yeah then I'll remove it

dev/sg/linters/linters.go
```diff
@@ -149,14 +149,19 @@ func runScriptSerialized(name string, script string) *linter {
...
-			event := honey.FromContext(ctx)
-
 			t1 := time.Now()
 			runScriptSerializedMu.Lock()
-			t2 := time.Since(t1)
-			event.AddField("pnpm_lock_duration", t2.Seconds())
-			event.AddField("pnpm_lock_duration_ms", t2.Milliseconds())
 			defer runScriptSerializedMu.Unlock()
+			// record the time it took to run the script, if it does have a honey event
+			defer func() {
```
burmudar, line 156: need to remove this

dev/sg/linters/linters.go
```diff
@@ -61,42 +59,42 @@ var Targets = []Target{
...
-			timeCheck(checkUnversionedDocsLinks()),
+			checkUnversionedDocsLinks(),
 		},
 	},
 	{
 		Name:        "pnpm",
 		Description: "Check pnpm lockfiles for optimality",
 		Checks: []*linter{
-			timeCheck(runScriptSerialized("pnpm dedupe", "dev/check/pnpm-deduplicate.sh")),
+			runScriptSerialized("pnm dedupe", "dev/check/pnpm-deduplicate.sh"),
```
Strum355, line 79:
```suggestion
			runScriptSerialized("pnpm dedupe", "dev/check/pnpm-deduplicate.sh"),
```

Strum355 (approved):

dev/sg/linters/linters.go
burmudar: Great catch

PR #62478, opened 2024-05-07
[Svelte]: Fix default blob page panel distribution

vovakulikov:
Prior to this PR, it was possible to see this incorrect state when the bottom panel collapsed but some of the action tabs were selected

Screenshot 2024-05-07 at 11 40 35

This happened because users' panels without saved layout in their local storage tried to calculate an initial layout automatically, and, despite the fact that the bottom panel has default size 1 (which is equal to its collapsed size 1), instead, it got a 1.09... value because the content panel had 90 as a default size and the rest space was distributed proportionally. 

So what we got at the end is 
- Content panel has 98.91... size
- Bottom panel has 1.09... size 
- Resize panels logic sees that panel has a bigger size than its collapsed size and fires onExpand callback 
- This callback set History tab selected
- As a result, a visually collapsed panel but with a selected tab in it. 

In this PR, I removed the default size setting from the content panel, which makes it take all available space (this is how we ensure that the bottom panel always gets 1 as default size). I also changed IDs to make sure that existing values in local storage won't have an effect there.

## Test plan
- Clear the local storage
- Check the blob UI page (directory and file)
- Check that you can expand and collapse panels and go between different files in one session

### DIFF
--- a/client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte
+++ b/client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte
@@ -196,15 +196,15 @@
         <PanelResizeHandle id="blob-page-panels-separator" />
     {/if}
 
-    <Panel id="content-panel" order={2}>
+    <Panel id="blob-content-panels" order={2}>
         <PanelGroup id="content-panels" direction="vertical">
-            <Panel id="main-content-panel" order={1} defaultSize={90}>
+            <Panel id="main-content-panel" order={1}>
                 <slot />
             </Panel>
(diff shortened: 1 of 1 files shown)


fkling (approved):
Do we need to make additional changes to ensure that we don't end up with values like `1.09`? It seems to be it's possible to hit this problem again in the future. Or maybe I just don't fully understand it 😅

vovakulikov:
@fkling it's actually an expected behaviour. Space distribution should work like this or we will break (change) standard panel behaviour. I think this is more like an edge case where you have to be careful with your panel default settings otherwise this happens. 

Users can't reach this incorrect panel layout state so I consider this as my mistake not a panels bug

PR #62485, opened 2024-05-07
Syntactic indexing: enqueuer and scheduler

keynmol:
This PR introduces three main components:

- Enqueuer – a thin layer responsible for actually inserting the syntactic indexing records into the database.
- Scheduler - a service that 
  - Identifies repositories that haven't been processed in a while
  - Identifies policies that match those repositories (policies that have syntactic indexing enabled)
  - Identifies commits that match any of the policies
  - And finally, enqueues the jobs to index the discovered repositories and commits
- Scheduler job – a periodic routine that triggers Scheduler on with specified interval. This job runs as part of the main Worker service, and only schedules jobs if the experimental syntactic indexing feature is enabled.

Refactoring:
- Making some methods public in policies.Service to make it easier to test logic that depends on glob matching of repository names (this matching requires a separate state to be updated)
- Extracting some test utilities into a separate package

TODO:

- ~[ ] Tests for policy iterator~ - we're currently investigating whether policy iterator is needed at all, as pagination of policies is in question.
- [x] Tests for Scheduler
- [x] Wire in experimental feature flag

## Test plan

- New tests for all components


internal/codeintel/syntactic_indexing/scheduler_test.go
```diff
@@ -0,0 +1,228 @@
...
+	if _, err := db.ExecContext(ctx, query); err != nil {
+		t.Fatalf("unexpected error while inserting configuration policies: %s", err)
+	}
+
+	for policyID, patterns := range map[int][]string{
+		1100: {"github.com/*"},
+	} {
+		if err := store.UpdateReposMatchingPatterns(ctx, patterns, policyID, nil); err != nil {
+			t.Fatalf("unexpected error while updating repositories matching patterns: %s", err)
+		}
```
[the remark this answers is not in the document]
varungandhi-src: Idiomatic way would likely be to only expose a public interface with the method (or handful of methods) of interest. Sometimes that can be tricky though.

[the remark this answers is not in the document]
keynmol: > Sometimes that can be tricky though

Yeah, I think exposing this particular method is not worth it.

I think I can expose a new service whose sole purpose is to be used in the periodic job that updates the `repo <--> policy` matching cache. 

So store will remain internal, and the service will delegate to it. If if works out, I'll extract it to a separate PR.

internal/codeintel/syntactic_indexing/enqueuer_test.go
```diff
@@ -0,0 +1,87 @@
...
+	require.True(t, len(scheduled) == 1)
+	require.Equal(t, scheduled[0].Commit, tacosCommit)
+	require.Equal(t, scheduled[0].RepositoryID, tacosRepoId)
+	require.Equal(t, scheduled[0].State, jobstore.Queued)
+	require.Equal(t, scheduled[0].RepositoryName, tacosRepoName)
+
+	// cannot schedule same repo+revision twice
+	result, err := enqueuer.QueueIndexes(ctx, tacosRepoId, tacosCommit, EnqueueOptions{})
+	require.Empty(t, result)
+	require.NoError(t, err)
```
varungandhi-src, line 77: You could use the `unwrap` function here if there's nothing to add as error context

internal/codeintel/syntactic_indexing/internal/policy_iterator.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/sourcegraph/sourcegraph/internal/codeintel/policies"
+	policiesshared "github.com/sourcegraph/sourcegraph/internal/codeintel/policies/shared"
+)
+
+type PolicyType int
+
+const (
+	SyntacticIndexing PolicyType = 0
+	PreciseIndexing   PolicyType = 1
+)
```
varungandhi-src, line 15: Unless this is perf sensitive, it'd be better to use a string enum to allow for easier print debugging and logging.

internal/codeintel/syntactic_indexing/scheduler_test.go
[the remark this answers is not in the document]
keynmol: Thanks to our call, I realised it was much easier than I thought initially: https://github.com/sourcegraph/sourcegraph/pull/62485/commits/42022f267e6aefc21b4f20089c14ce12ea27d73f

internal/codeintel/syntactic_indexing/internal/policy_iterator.go
```diff
@@ -0,0 +1,94 @@
...
+	forSyntacticIndexing := &f
+	forPreciseIndexing := &f
+
+	if p.PolicyType == SyntacticIndexing {
+		forSyntacticIndexing = &t
+		forPreciseIndexing = nil
+	} else {
+		forPreciseIndexing = &t
+		forSyntacticIndexing = nil
+	}
```
varungandhi-src, line 45:
```suggestion
	forSyntacticIndexing = pointers.Ptr(p.PolicyType == SyntacticIndexing)
	forPreciseIndexing = pointers.Ptr(p.PolicyType == PreciseIndexing)
```

internal/codeintel/syntactic_indexing/internal/policy_iterator.go
```diff
@@ -0,0 +1,94 @@
...
+		if handlerError != nil {
+			return handlerError // propagate error from the handler
+		}
+
+		offset = offset + len(policies)
+
+		if offset >= totalCount {
+			break
+		}
+	}
```
varungandhi-src, line 80:
```suggestion

	options := policiesshared.GetConfigurationPoliciesOptions{
		RepositoryID:         p.RepositoryID,
		ForSyntacticIndexing: forSyntacticIndexing,
		ForPreciseIndexing:   forPreciseIndexing,
		Limit:                p.BatchSize,
	}
	for offset := 0; ; {
	    options.Offset = offset
		policies, totalCount, err := p.Service.GetConfigurationPolicies(ctx, options)
		if err != nil {
			return err
		}
		if len(policies) == 0 {
			break
		}		
		if handlerError := handle(policies); handlerError != nil {
			return handlerError
		}
		if offset = offset + len(policies); offset >= totalCount {
			break
		}
	}
```

nit: Please don't add whitespace after every statement.

internal/codeintel/syntactic_indexing/jobstore/store.go
```diff
@@ -1,22 +1,32 @@
...
 	"github.com/sourcegraph/sourcegraph/internal/database/basestore"
+	"github.com/sourcegraph/sourcegraph/internal/database/dbutil"
 	"github.com/sourcegraph/sourcegraph/internal/observation"
 	dbworkerstore "github.com/sourcegraph/sourcegraph/internal/workerutil/dbworker/store"
 )
 
 type SyntacticIndexingJobStore interface {
 	DBWorkerStore() dbworkerstore.Store[*SyntacticIndexingJob]
+	InsertIndexes(ctx context.Context, indexes []SyntacticIndexingJob) ([]SyntacticIndexingJob, error)
+	IsQueued(ctx context.Context, repositoryID int, commit string) (bool, error)
```
varungandhi-src, line 22: Thoughts on pushing this logic into a single method? Otherwise, this API creates the risk of TOCTOU because something might get queued in-between the `IsQueued` check and the `Insert` check. AFAICT, that's impossible given that these are currently invoked from the same goroutine, and there's only going to be one such goroutine, and the worst case scenario is that there will be some redundant work, but it seems unnecessarily error-prone.

internal/codeintel/syntactic_indexing/jobstore/store.go
```diff
@@ -57,7 +67,140 @@ func NewStoreWithDB(observationCtx *observation.Context, db *sql.DB) (SyntacticI
...
+
+	indexes = []SyntacticIndexingJob{}
+
+	err = s.db.WithTransact(ctx, func(tx *basestore.Store) error {
+		ids, err := basestore.ScanInts(tx.Query(ctx, sqlf.Sprintf(insertIndexQuery, sqlf.Join(values, ","))))
+		if err != nil {
+			return err
+		}
+		// TODO: add operations and record this operation
+		// s.operations.indexesInserted.Add(float64(len(ids)))
```
varungandhi-src, line 109: Does this need fixing?

internal/codeintel/syntactic_indexing/jobstore/store.go
```diff
@@ -57,7 +67,140 @@ func NewStoreWithDB(observationCtx *observation.Context, db *sql.DB) (SyntacticI
...
+			return err
+		}
+
+		queries := make([]*sqlf.Query, 0, len(ids))
+		for _, id := range ids {
+			queries = append(queries, sqlf.Sprintf("%d", id))
+		}
+
+		indexes, err = scanIndexes(tx.Query(ctx, sqlf.Sprintf(getIndexesByIDsQuery, sqlf.Join(queries, ", "), authzConds)))
+		return err
```
varungandhi-src, line 122: I don't understand this code. It looks like we're performing insertions into `syntactic_scip_indexing_jobs` and returning the ids for the just-inserted rows. Then we're reading more data from those rows in a separate query (through a view). Can we expand the `RETURNING` clause to return more columns during insertion? I don't understand what role the extra `authzConds` is playing here.

internal/codeintel/syntactic_indexing/jobstore/store.go
```diff
@@ -57,7 +67,140 @@ func NewStoreWithDB(observationCtx *observation.Context, db *sql.DB) (SyntacticI
...
+const isQueuedQuery = `
+SELECT EXISTS(
+	SELECT queued_at
+	FROM syntactic_scip_indexing_jobs
+	WHERE
+		repository_id  = %s AND
+		commit = %s
+	ORDER BY queued_at DESC
+	LIMIT 1
+)
```
varungandhi-src, line 159: I think `LIMIT 1` inside `EXISTS` will be redundant.

varungandhi-src (commented):
Left some initial comments; will take a deeper look at more of the code shortly.

internal/codeintel/syntactic_indexing/jobstore/store.go
keynmol: I imagine this is more of a hint to Postgres query optimiser, to not plan for returning more rows.

keynmol: To rephrase - large chunks of this are adapted from the way auto-indexing code does it, and I've kept SQL queries mostly intact if they made sense.

varungandhi-src: Found this bug report from 2009 indicating that the optimizer sometimes doesn't figure it out by itself. 💩 

https://www.postgresql.org/message-id/b61e61d30910021046i37234b7fqc38a1376485ce234%40mail.gmail.com

It's fine to leave this in to be safe.

cmd/worker/shared/init/db/db.go
```diff
@@ -13,14 +13,23 @@ import (
...
 }
 
+func InitRawDB(observationCtx *observation.Context) (*sql.DB, error) {
+	rawDB, err := initDatabaseMemo.Init(observationCtx)
+	if err != nil {
+		return nil, err
+	}
+
+	return rawDB, err
+}
```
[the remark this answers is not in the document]
keynmol: For bootstrapping the scheduler, a raw (`sql.DB`) is required as a dependency to some services.
AFAIK there's no way to get a raw `sql.DB` handle out of our `database.DB` wrapper - and so I need to expose the raw memoised handle as well.

I will apply your suggestion, thanks

internal/codeintel/syntactic_indexing/enqueuer_test.go
```diff
@@ -0,0 +1,87 @@
...
+	repoSchedulingStore := reposcheduler.NewSyntacticStore(observationCtx, db)
+
+	repoStore := db.Repos()
+
+	gsClient := gitserver.NewMockClient()
+	enqueuer := NewIndexEnqueuer(observationCtx, jobStore, repoSchedulingStore, repoStore, gsClient)
+
+	tacosRepoId, tacosRepoName, tacosCommit := 1, "tangy/tacos", testutils.MakeCommit(1)
+	empanadasRepoId, empanadasRepoName, empanadasCommit := 2, "salty/empanadas", testutils.MakeCommit(2)
+	mangosRepoId, mangosRepoName, mangosCommit := 3, "juicy/mangos", testutils.MakeCommit(3)
```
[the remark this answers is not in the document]
keynmol: One thing I want to ensure is that the repo is set up - to avoid triggering a short circuit in the logic when the repo is not found.

To that end, I removed mangosCommit and am requesting a non-existent commit in the test: b1ca27b6e9c

internal/codeintel/syntactic_indexing/scheduler.go
```diff
@@ -0,0 +1,143 @@
...
+		Config:                      &config,
+	}, nil
+}
+
+func BootstrapSyntacticJobScheduler(observationCtx *observation.Context, rawDB *sql.DB) (SyntacticJobScheduler, error) {
+	database := database.NewDB(observationCtx.Logger, rawDB)
+
+	gitserverClient := gitserver.NewClient("codeintel-syntactic-indexing")
+
+	codeIntelDB := codeintelshared.NewCodeIntelDB(observationCtx.Logger, rawDB)
```
varungandhi-src, line 60: Something is off here. This function is getting passed in a `frontend` database. However, it is using that to call `NewCodeIntelDB`. In local dev, we only have one database. However, the `codeintel` and `frontend` databases are different in production.

internal/codeintel/syntactic_indexing/enqueuer.go
```diff
@@ -0,0 +1,146 @@
...
+	if !options.force {
+		values := []jobstore.SyntacticIndexingJob{
+			{
+				State:        jobstore.Queued,
+				Commit:       commit,
+				RepositoryID: repositoryID,
+			}}
+
+		return s.jobStore.InsertIndexes(ctx, values)
+	}
```
varungandhi-src, line 143: If `options.force` is true, then this code won't create any new indexing jobs. Is that intentional?

internal/codeintel/syntactic_indexing/internal/policy_iterator.go
```diff
@@ -0,0 +1,73 @@
...
+		options.Offset = 0
+		policiesBatch, totalCount, err := p.Service.GetConfigurationPolicies(ctx, options)
+		if err != nil {
+			return err
+		}
+		if len(policiesBatch) == 0 {
+			break
+		}
+		if handlerError := handle(policiesBatch); handlerError != nil {
+			return handlerError // propagate error from the handler
```
varungandhi-src, line 54:
```suggestion
			return handlerError
```

nit: Comment doesn't add anything over the surrounding code

internal/codeintel/syntactic_indexing/enqueuer_test.go
```diff
@@ -0,0 +1,87 @@
...
+	// but also doesn't insert a new job
+	result := unwrap(enqueuer.QueueIndexes(ctx, tacosRepoId, tacosCommit, EnqueueOptions{}))(t)
+	require.Empty(t, result)
+
+	// cannot schedule for non existent repositories
+	_, err = enqueuer.QueueIndexes(ctx, 250, tacosCommit, EnqueueOptions{})
+	require.Error(t, err)
+
+	// cannot schedule for non existent revisions
+	_, err = enqueuer.QueueIndexes(ctx, mangosRepoId, testutils.MakeCommit(100), EnqueueOptions{})
```
varungandhi-src, line 84: Please add a test or two for the case when `force` is set to true.

internal/codeintel/syntactic_indexing/jobstore/store.go
varungandhi-src: Thanks for making a serious go at it. I didn't expect this to be so complicated, guess I underestimated the weirdness of SQL. I think it's not a big deal to keep it as-is.

internal/codeintel/syntactic_indexing/scheduler.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/sourcegraph/sourcegraph/internal/codeintel/syntactic_indexing/internal"
+	"github.com/sourcegraph/sourcegraph/internal/codeintel/syntactic_indexing/jobstore"
+	"github.com/sourcegraph/sourcegraph/internal/database"
+	"github.com/sourcegraph/sourcegraph/internal/gitserver/gitdomain"
+	"github.com/sourcegraph/sourcegraph/internal/observation"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+)
+
+type SyntacticJobScheduler interface {
+	// TODO: make it return job that were queued successfully?
```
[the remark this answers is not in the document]
keynmol: I don't think it needs to be addressed – our current tests actually check the database state after invoking Schedule, which is more reliable. 

Will remove the TODO for now.

internal/codeintel/syntactic_indexing/jobstore/store.go
keynmol: While you are right that it's susceptible to TOCTOU (if we use those methods independently somewhere else, not in the current case), I think in this case all of this is best effort - the implication of scheduling something twice is fairly harmless (and hopefully infrequent), but to make this atomic we'd have to commit some sins with doing both in the same query, and historically those parts of the codebase have been the hardest to debug.

varungandhi-src: I don't think we need them to be in the same query but in the same transaction. Not sure if keeping these as separate queries but having them be part of the same transaction would help with debuggability (compared to a combined query). Leaving the final decision up to you.

internal/codeintel/syntactic_indexing/scheduler.go
keynmol: You are right - for now I just separated the raw DB arguments and added a dbtest helper with only the codeintel schema: https://github.com/sourcegraph/sourcegraph/pull/62485/commits/6b3685acab6cb109ddc7c953936bbb83d7e2d31c

internal/codeintel/syntactic_indexing/enqueuer.go
```diff
@@ -0,0 +1,146 @@
...
+		queueIndexes: op("QueueIndexes"),
+	}
+}
+
+// QueueIndexes enqueues a set of index jobs for the following repository and commit. If a non-empty
+// configuration is given, it will be used to determine the set of jobs to enqueue. Otherwise, it will
+// the configuration will be determined based on the regular index scheduling rules: first read any
+// in-repo configuration (e.g., sourcegraph.yaml), then look for any existing in-database configuration,
+// finally falling back to the automatically inferred configuration based on the repo contents at the
+// target commit.
```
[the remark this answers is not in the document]
keynmol: This is just a stray comment, I didn't remove it when removing the notion of configuration (which is purely a precise auto-indexing/inference concept).

Will rework the comment to capture what the method's intention is.

internal/codeintel/syntactic_indexing/enqueuer.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/sourcegraph/sourcegraph/internal/database"
+	"github.com/sourcegraph/sourcegraph/internal/gitserver"
+	"github.com/sourcegraph/sourcegraph/internal/metrics"
+	"github.com/sourcegraph/sourcegraph/internal/observation"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+	"go.opentelemetry.io/otel/attribute"
+)
+
+type IndexEnqueuer interface {
+	QueueIndexingJobs(ctx context.Context, repositoryId int, rev string, options EnqueueOptions) (_ []jobstore.SyntacticIndexingJob, err error)
```
varungandhi-src, line 19: At the moment, it seems like the only call-site of this is going to pass in commits for the 3rd argument (obtained from `CommitsDescribedByPolicy`), not arbitrary revisions like HEAD. Is that correct?

Is it possible that `CommitsDescribedByPolicy` can return non-existent commits? Or can we avoid the call to gitserver to resolve the revision (and take an `api.CommitID` as the argument instead of `rev string`)?

internal/codeintel/syntactic_indexing/enqueuer_test.go
```diff
@@ -0,0 +1,96 @@
...
+	require.Error(t, err)
+
+	// cannot schedule for non existent revisions
+	_, err = enqueuer.QueueIndexingJobs(ctx, mangosRepoId, testutils.MakeCommit(100), EnqueueOptions{})
+	require.Error(t, err)
+
+	// force: true in EnqueueOptions allows scheduling the same (repo, revision) twice
+	reinserted := unwrap(enqueuer.QueueIndexingJobs(ctx, tacosRepoId, tacosCommit, EnqueueOptions{force: true}))(t)
+	require.Equal(t, 1, len(reinserted))
+	require.NotEqual(t, reinserted[0].ID, scheduled[0].ID) // ensure it's actually a new job
```
varungandhi-src, line 90: Thanks for adding this test!

internal/codeintel/syntactic_indexing/enqueuer.go
keynmol: From a cursory look at the CommitsDescribedByPolicy I think we always return commits, and not arbitrary revisions.

ListRefs can probably return a ref that will be removed by the time ResolveRevision is called, which is our last verification that commit exists.

I think to change this we'd need confirmation from the source team that ListRefs will always populate CommitID field.

Then we can move reference resolution at the time of indexing.

varungandhi-src: OK. I don't mind merging this as-is, but I vaguely remember us hitting ResolveRevision way too often in some cases, so it would be nice if we could avoid that here if it's already being done implicitly as part of the calling code.

keynmol: Also if we go with `tar` streaming from gitserver directly, it will also do the existence of the ref for us. I will address it once we have the actual indexer in place.

My assumption is that this amount of checks exists in autoindexing code to avoid scheduling too many jobs in the first place - because launching executors is much much heavier than what we're doing here. So in our case scheduling an extra job for a commit that disappeared might not be such a big deal.

internal/codeintel/syntactic_indexing/jobstore/store.go
```diff
@@ -57,7 +67,139 @@ func NewStoreWithDB(observationCtx *observation.Context, db *sql.DB) (SyntacticI
...
+			index.State,
+			index.Commit,
+			index.RepositoryID,
+			actor.UID,
+		))
+	}
+
+	indexes = []SyntacticIndexingJob{}
+
+	err = s.db.WithTransact(ctx, func(tx *basestore.Store) error {
```
varungandhi-src, line 103: nit: I hate to keep playing "whitespace police" and point out unhelpful variable names, but please don't add blank lines before/after most statements like this, it doesn't help with readability. For example, there is virtually no other code in the codebase where we have a blank line between the function declaration and the function declaration and the first statement. Wrt variable names, we're again using `indexes` here for indexing _jobs_. Then there are generic variable names like `values` and `queries`. OK, I understand that's related to SQL's `VALUE`, but we can use a more descriptive name, say `indexJobValues`. Similarly, insert of `queries`, we can use something like `lookupIDs`.

Also, I understand a bunch of this is copy-pasted from the precise indexing code, but let's take this opportunity to improve things in both places (or at least in the new place) rather than copy-pasting code with bad formatting/naming.

internal/codeintel/syntactic_indexing/jobstore/store.go
```diff
@@ -57,7 +67,137 @@ func NewStoreWithDB(observationCtx *observation.Context, db *sql.DB) (SyntacticI
...
+		ids, err := basestore.ScanInts(tx.Query(ctx, sqlf.Sprintf(insertIndexQuery, sqlf.Join(values, ","))))
+		if err != nil {
+			return err
+		}
+		s.operations.indexingJobsInserted.Add(float64(len(ids)))
```
varungandhi-src: We have a counter here for jobs added. Thoughts on adding a similar counter for jobs that were skipped because they were already queued? That might potentially be useful in case we're running queuing too frequently compared to the queue processing speed.

internal/codeintel/syntactic_indexing/enqueuer.go
varungandhi-src: The same problem applies to repos. We're doing a `repoStore.Get(` outside the `for commit, policyMatches := range commitMap` loop. But we just pass in an int ID and then we'll call `repoStore.Get(` again inside `QueueIndexingJobs`, once for each (commit, policy) pair. That seems quite wasteful. Repos are much less likely to go missing compared to commits as well.

internal/codeintel/syntactic_indexing/scheduler_config.go
```diff
@@ -0,0 +1,28 @@
...
+	RepositoryProcessDelay time.Duration
+	RepositoryBatchSize    int
+	PolicyBatchSize        int
+}
+
+func (c *SchedulerConfig) Load() {
+	intervalName := env.ChooseFallbackVariableName("CODEINTEL_SYNTACTIC_INDEXING_SCHEDULER_INTERVAL", "PRECISE_CODE_INTEL_AUTO_INDEXING_TASK_INTERVAL")
+	repositoryProcessDelayName := env.ChooseFallbackVariableName("CODEINTEL_SYNTACTIC_INDEXING_SCHEDULER_REPOSITORY_PROCESS_DELAY", "PRECISE_CODE_INTEL_AUTO_INDEXING_REPOSITORY_PROCESS_DELAY")
+	repositoryBatchSizeName := env.ChooseFallbackVariableName("CODEINTEL_SYNTACTIC_INDEXING_SCHEDULER_REPOSITORY_BATCH_SIZE", "PRECISE_CODE_INTEL_AUTO_INDEXING_REPOSITORY_BATCH_SIZE")
+	policyBatchSizeName := env.ChooseFallbackVariableName("CODEINTEL_SYNTACTIC_INDEXING_SCHEDULER_POLICY_BATCH_SIZE", "PRECISE_CODE_INTEL_AUTO_INDEXING_POLICY_BATCH_SIZE")
```
varungandhi-src, line 22: IIUC, this was copy-pasted-tweaked from similar config variables for precise auto-indexing. I'm wondering if:
1. We should remove the fallback to the old precise variables (looks like those were renamed for consistency)
2. Perhaps we should fall back to the more generic values such as `CODEINTEL_INDEXING_SCHEDULER_INTERVAL` (should we introduce a separate `CODEINTEL_PRECISE_INDEXING_SCHEDULER_INTERVAL`, that falls back to `CODEINTEL_INDEXING_SCHEDULER_INTERVAL`).

I'm having trouble having coming up with a case where the current fallback behavior is actually desirable. If you applied some configuration settings for precise, it's plausible that the different perf characteristics mean that you _don't_ want those settings to apply for syntactic.

internal/codeintel/syntactic_indexing/scheduler.go
```diff
@@ -0,0 +1,134 @@
...
+				}
+			}
+
+			return nil
+
+		})
+
+		if err != nil {
+			return err
+		}
```
varungandhi-src, line 130: Doesn't this seem a bit too pessimistic? If there was a failure for one repo, why should we skip syntactic indexing for all the other repos?

internal/codeintel/syntactic_indexing/scheduler.go
```diff
@@ -0,0 +1,121 @@
...
+			commitMap, err := s.PolicyMatcher.CommitsDescribedByPolicy(ctx, int(repoToIndex.ID), repo.Name, policies, currentTime)
+			if err != nil {
+				return err
+			}
+			for commit, policyMatches := range commitMap {
+				if len(policyMatches) == 0 {
+					continue
+				}
+				if _, err := s.Enqueuer.QueueIndexingJobs(ctx, int(repoToIndex.ID), commit, EnqueueOptions{force: false}); err != nil {
+					if errors.HasType(err, &gitdomain.RevisionNotFoundError{}) {
```
varungandhi-src, line 108: I'm a bit concerned about this using `HasType` instead of `Is`, which I think is the more common function, but if we have a test for this, then it's OK. The difference between the two is not quite clear to me. https://sourcegraph.slack.com/archives/C02UC4WUX1Q/p1717067410264679

internal/codeintel/syntactic_indexing/scheduler_test.go
```diff
@@ -0,0 +1,197 @@
...
+			--                        							              |    |    |     |      |      |     ↙ index_intermediate_commits
+			(1000, 2,    'policy  1 abc', 'GIT_TREE', '', null,              false, 0, false, true,  false,  0, false),
+			-- Policy below specifically disables syntactic indexing for repo with ID=3
+			(1003, 3,    'policy  3 bcd', 'GIT_TREE', '', null,              false, 0, false, false, false, 0,  false),
+			-- Policy below enables syntactic indexing for all repositories starting with 'github.com'
+			(1100, NULL, 'policy 10 def', 'GIT_TREE', '', '{github.com/*}',  false, 0, false, true,  false, 0,  false)
+	`
+	unwrap(db.ExecContext(ctx, query))(t)
+
+	for _, policyID := range []int{1100} {
```
varungandhi-src, line 192: Should this array have 1000 and 1003 too? If this test was passing before this, does this mean we're not testing the logic that ought to be triggered by policies 1000 and 1003 (and should we be testing that here or is that handled by the policy matcher tests?)

varungandhi-src (approved):
Suggestions 
1. (Strong) Let's avoid doing the repo check again inside the doubly nested loop.
2. (Weak) Let's avoid doing the commit check multiple times
3. (Weak) Let's avoid revisiting the same commit multiple times per repo
4. (Weak) Let's simplify the env vars. We can make it more complicated later.
5. (Strong) Please clarify the policies array elements in `scheduler_test.go`

Would prefer Strong suggestions to be addressed pre-merge.

Thanks for your patience. It's been a bit hard for me to wrap my head around all the logic here, so this took longer to review than anticipated.

internal/codeintel/syntactic_indexing/scheduler_test.go
keynmol: This intentionally only includes policy 1100 - it's the only policy that has non-null repository patterns (and a null repository id), which triggers a lookup in the special separate table: https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/internal/codeintel/policies/internal/store/configurations.go?L59-65

So we are filling this lookup table only for this policy - the other two will match the repository by id directly.

internal/codeintel/syntactic_indexing/scheduler_config.go
keynmol: I agree at this point we probably don't need any fallback at all - we have no idea about the perf characteristics of syntactic, so it should steer clear of precise as much as possible.

internal/codeintel/syntactic_indexing/scheduler.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+	}
+
+	for repoId, commits := range commitsToSchedule {
+		for _, commitId := range commits.Values() {
+			if _, err := s.Enqueuer.QueueIndexingJobs(ctx, repoId, commitId, enqueueOptions); err != nil {
+				allErrors = errors.Append(allErrors, errors.Newf("Failed to schedule syntactic indexing of repo [ID=%s], commit [%s]: %v", repoId, commitId, err))
+			}
+		}
+	}
```
keynmol, line 141: @varungandhi-src I've re-done this part, main highlights:

1. Collecting all repos and commits once
2. Removing fail fast behavior in repo iterator
3. Removing fail fast behavioru in (repo, commit) iterator

Additionally the enqueuer no longer performs revision checks, we'll leave that to the syntactic worker itself.

This means the scheduler will always do best effort scheduling, while returning all accumulated errors.

I hope my understanding of `return allErrors` is right here and equivalent to `return nil` if no `allErrors = errors.Append` calls ever happened.

varungandhi-src: Thanks, this looks much better!

> I hope my understanding of return allErrors is right here and equivalent to return nil if no allErrors = errors.Append calls ever happened.

Yep, we're following this pattern elsewhere too. I think you don't even need to declare the type as `MultiError`, you can see other places where `errors.Append(` can take the first argument as just `error` instead of `MultiError`. (search for `var errs error`)

internal/codeintel/syntactic_indexing/scheduler.go
varungandhi-src: It looks like HasType is correct here, I see other places also using `HasType` rather than `Is` with `RevisionNotFoundError`, since we don't want to compare the payload.

internal/codeintel/syntactic_indexing/jobstore/store.go
keynmol: Just saw this - good idea, will add the counter

keynmol:
I will merge this PR and if there are any other serious issues to attend to, it can happen in subsequent PRs, given this component is disabled by default.

Issue #62426, opened 2024-05-03
Fuzzy finder


fkling:
... or something that resembles it.

My suggestion would be to try to use the same fetcher/caching logic as the search input, because that should be relatively reusable. If that doesn't work or doesn't provide the same results as the current fuzzy finder we can try to refactor the existing one to decouple it more from React.

```[tasklist]
### Tasks
- [ ] https://github.com/sourcegraph/sourcegraph/pull/62494
- [ ] Add support for toggle between repo and global (?)
- [ ] Add result count
- [ ] Add empty state
```

taiyab:
Follow up to this: https://github.com/sourcegraph/sourcegraph/issues/62673

PR #62494, opened 2024-05-07
svelte: First pass on fuzzy finder

fkling:
Part of #62426

This commit adds a first version of the fuzzy finder. To get things moving quickly I'm using the same mechanism to fetch data as the search input.

Supported features:

- Repos, symbols, files search
- Automatic scoping of files and symbol search to current repository
- Activting search modes via keyboard shortcuts
- Keyboard support

Not yet supported:

- "All" search
- Result counts
- Scope toggle (i.e. search all files even when inside repository)
- Full a11y support
- Take into account visited repos/files

Additional changes:

- Introduced separate tab headers component. In this case I wanted to reuse the same panel/UI and only change the data source.
- Added a function for formatting keyboard shortcuts that work with our hotkey implementation.

Follow up work:

- Cleanup and tweak sources (matching, ranking, etc)
- Design updates
- Work on "not yet supported" items

https://github.com/sourcegraph/sourcegraph/assets/179026/074455c0-eb1c-49c8-a206-3974bec0e494

## Test plan

Manual testing

### DIFF
--- a/client/web-sveltekit/src/lib/TabsHeader.svelte
+++ b/client/web-sveltekit/src/lib/TabsHeader.svelte
@@ -1,0 +1,109 @@
+<script lang="ts" context="module">
+    export interface Tab {
+        id: string
+        title: string
+        icon?: string
+    }
+</script>
+
+<script lang="ts">
+    import { createEventDispatcher } from 'svelte'
+
+    import Icon from '$lib/Icon.svelte'
+
+    export let id: string
+    export let tabs: Tab[]
+    export let selected: number | null = 0
+
+    const dispatch = createEventDispatcher<{ select: number }>()
+
+    function selectTab(event: MouseEvent) {
+        const index = (event.target as HTMLElement).closest('[role="tab"]')?.id.match(/\d+$/)?.[0]
+        if (index) {
+            dispatch('select', +index)
+        }
+    }
+</script>
+
+<div class="tabs-header" role="tablist" data-tab-header>
+    {#each tabs as tab, index (tab.id)}
+        <button
+            id="{id}--tab--{index}"
+            aria-controls={tab.id}
+            aria-selected={selected === index}
+            tabindex={selected === index ? 0 : -1}
+            role="tab"
+            on:click={selectTab}
+            data-tab
+            >{#if tab.icon}<Icon svgPath={tab.icon} aria-hidden inline /> {/if}<span data-tab-title={tab.title}
+                >{tab.title}</span
+            ><slot name="after-title" {tab} /></button
+        >
+    {/each}
+</div>
+
+<style lang="scss">
+    .tabs-header {
+        --icon-fill-color: var(--header-icon-color);
+
+        display: flex;
+        align-items: stretch;
+        justify-content: var(--align-tabs, center);
+        gap: var(--tabs-gap, 0);
+        border-bottom: 1px solid var(--border-color);
+    }
+
+    [role='tab'] {
+        all: unset;
+
+        cursor: pointer;
+        align-items: center;
+        min-height: 2rem;
+        padding: 0.25rem 0.75rem;
+        color: var(--text-body);
+        display: inline-flex;
+        flex-flow: row nowrap;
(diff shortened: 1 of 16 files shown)


camdencheek (approved):
Looks great! Leaving a few notes that are fine to leave for followup:

1) I expected control+n and control+p to work for next/previous as it does in the search input

2) The double border on the tabs looks a little funny when they are not selected

screenshot-2024-05-08_09-09-43@2x

3) It looks like the keyboard shortcuts get propagated and it's still activating menu items on mac?

https://github.com/sourcegraph/sourcegraph/assets/12631702/82131a7f-12d6-4704-a964-e94a2043fdd9

4) There doesn't appear to be a way to expand the scope outside of the active repo. Definitely skippable for now though.

taiyab:
@fkling Nice work getting this in Felix!

client/web-sveltekit/src/lib/TabsHeader.svelte
```diff
@@ -0,0 +1,109 @@
+<script lang="ts" context="module">
+    export interface Tab {
+        id: string
+        title: string
+        icon?: string
+    }
+</script>
+
+<script lang="ts">
```
vovakulikov, line 9: It looks like that most of our tabs should have keyboard navigation suffixes and shortcuts. 
Might be a good idea to have API over tabs that allows you to set shortcuts without writing 
any logic in consumer directly

client/web-sveltekit/src/lib/fuzzyfinder/FuzzyFinder.svelte
```diff
@@ -0,0 +1,362 @@
...
+    let query = ''
+
+    const client = getGraphQLClient()
+    const tabs: (Tab & { source: CompletionSource<FuzzyFinderResult> })[] = [
+        { id: 'repos', title: 'Repos', source: createRepositorySource(client) },
+        { id: 'symbols', title: 'Symbols', source: createSymbolSource(client, () => scope) },
+        { id: 'files', title: 'Files', source: createFileSource(client, () => scope) },
+    ]
+
+    function selectNext() {
```
vovakulikov, line 50: Something that we discussed in the pair session, this solution is fine for the migration part
since it's been in the React version but we probably should use our standard (melt-ui) logic 
around keyboard navigation like in this component, even if melt-ui doesn't provide a clean 
abstraction for only list navigation (but with Combobox logic). It would be easier for us to 
migrate it to our unified logic later.

client/web-sveltekit/src/lib/fuzzyfinder/FuzzyFinder.svelte
```diff
@@ -0,0 +1,362 @@
...
+    }
+    $: if (open) {
+        dialog?.showModal()
+        input?.select()
+    } else {
+        dialog?.close()
+    }
+</script>
+
+<dialog bind:this={dialog} on:close>
```
vovakulikov, line 175: Have been seeing need for the dialog component, so probably should have Dialog in the svelte 
wildcard package.

client/web-sveltekit/src/lib/fuzzyfinder/FuzzyFinder.svelte
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    let dialog: HTMLDialogElement | undefined
+    let listbox: HTMLElement | undefined
+    let input: HTMLInputElement | undefined
+    let query = ''
+
+    const client = getGraphQLClient()
+    const tabs: (Tab & { source: CompletionSource<FuzzyFinderResult> })[] = [
+        { id: 'repos', title: 'Repos', source: createRepositorySource(client) },
```
vovakulikov, line 45: It's not a big deal but referring to your own comment about having a consistent logic 
around data fetching (even if it's not always an optimal solution), we probably could have it
inside loader function (data sources calls)

vovakulikov (commented):
I know it's merged but left a few comments 
Just for any other further improvements, thanks for the finder!

PR #62498, opened 2024-05-07
Svelte: Implement and instantiate FilePopover

jasonhawkharris:
This PR Implements and instantiates the file popover on the SvelteKit web app. 

Dark Mode: 
https://github.com/sourcegraph/sourcegraph/assets/62355966/f1405cc2-54e0-49ba-aa26-7ddf2d258970

Light Mode: 
https://github.com/sourcegraph/sourcegraph/assets/62355966/a8eb53ed-2a38-4574-a76a-5f5559ded015

Questions: 
1. This is working very well, but is there a better way to implement it? 

## Test plan

Added a storybook test and a playwright test

### DIFF
--- a/client/web-sveltekit/src/lib/repo/filePopover/FilePopover.gql
+++ b/client/web-sveltekit/src/lib/repo/filePopover/FilePopover.gql
@@ -1,0 +1,63 @@
+query FileOrDirPopoverQuery($repoName: String!, $revision: String!, $filePath: String!) {
+    repository(name: $repoName) {
+        commit(rev: $revision) {
+            path(path: $filePath) {
+                ...on GitBlob {
+                    ...FilePopoverFragment
+                }
+                ...on GitTree {
+                    ...DirPopoverFragment
+                }
+            }
+        }
+    }
+}
+
+fragment FilePopoverFragment on GitBlob {
+    __typename
+    path
+    languages
+    byteSize
+    totalLines
+    history(first: 1) {
+        nodes {
+            commit {
+                ...FilePopoverLastCommitFragment
+            }
+        }
+    }
+    ...FileIcon_GitBlob
+}
+
+fragment DirPopoverFragment on GitTree {
+    __typename
+    path
+    # TODO(camdencheek): fix this to use a count, which currently does not exist in our API.
+    # This currently does not scale well with very large directories.
+    files {
+        name
+    }
+    directories {
+        name
+    }
+    history(first: 1) {
+        nodes {
+            commit {
+                ...FilePopoverLastCommitFragment
+            }
+        }
+    }
+}
+
+fragment FilePopoverLastCommitFragment  on GitCommit {
+    abbreviatedOID
+    oid
+    subject
+    canonicalURL
+    author {
+        date
+        person {
+            ...Avatar_Person
+        }
+    }
+}
--- a/client/web-sveltekit/src/lib/repo/filePopover/FilePopover.stories.svelte
+++ b/client/web-sveltekit/src/lib/repo/filePopover/FilePopover.stories.svelte
@@ -1,0 +1,58 @@
+<script lang="ts" context="module">
+    import { Story } from '@storybook/addon-svelte-csf'
+
+    import type { FilePopoverFragment, DirPopoverFragment, FilePopoverLastCommitFragment } from './FilePopover.gql'
+    import FilePopover from './FilePopover.svelte'
+
+    export const meta = {
+        component: FilePopover,
+    }
+
+    const lastCommit: FilePopoverLastCommitFragment = {
+        abbreviatedOID: '1234567',
+        oid: '1234567890123456789012345678901234567890',
+        subject: 'Test subject',
+        canonicalURL: 'https://sourcegraph.com/about',
+        author: {
+            date: '2021-01-01T00:00:00Z',
+            person: {
+                __typename: 'Person',
+                name: 'camdencheek',
+                avatarURL: 'https://github.com/camdencheek.png',
+                displayName: 'Camden Cheek',
+            },
+        },
+    }
+
+    const fileEntry: FilePopoverFragment = {
+        __typename: 'GitBlob',
+        name: 'results.go',
+        path: 'internal/search/results/results.go',
+        languages: ['Go'],
+        byteSize: 325,
+        totalLines: 12,
+        history: {
+            nodes: [
+                {
+                    commit: lastCommit,
+                },
+            ],
+        },
+    }
+
+    const dirEntry: DirPopoverFragment = {
+        __typename: 'GitTree',
+        path: 'internal/search/results',
+        files: [{ name: 'results.go' }, { name: 'results_test.go' }],
+        directories: [{ name: 'testdata' }],
+        history: {
+            nodes: [{ commit: lastCommit }],
+        },
+    }
+</script>
+
+<Story name="Default">
+    <FilePopover repoName={'github.com/sourcegraph/sourcegraph'} entry={fileEntry} />
+
+    <FilePopover repoName={'github.com/sourcegraph/sourcegraph'} entry={dirEntry} />
+</Story>
--- a/client/web-sveltekit/src/lib/repo/filePopover/FilePopover.svelte
+++ b/client/web-sveltekit/src/lib/repo/filePopover/FilePopover.svelte
@@ -38,16 +38,14 @@
     import NodeLine from './NodeLine.svelte'
 
     export let repoName: string
-    export let filePath: string
     export let entry: FilePopoverFragment | DirPopoverFragment
 
     function splitPath(filePath: string): [string, string] {
         let parts = filePath.split('/')
-        console.log({ parts })
         return [parts.slice(0, parts.length - 1).join('/'), parts[parts.length - 1]]
     }
 
-    $: [dirName, baseName] = splitPath(filePath)
+    $: [dirName, baseName] = splitPath(entry.path)
     $: lastCommit = entry.history.nodes[0].commit
 </script>
 
--- a/client/web-sveltekit/src/lib/repo/filePopover/FilePopover.svelte
+++ b/client/web-sveltekit/src/lib/repo/filePopover/FilePopover.svelte
@@ -1,0 +1,190 @@
+<script lang="ts" context="module">
+    import { type DirPopoverFragment, type FilePopoverFragment, FileOrDirPopoverQuery } from './FilePopover.gql'
+
+    export interface FilePopoverInputs {
+        repoName: string
+        revision: string
+        filePath: string
+    }
+
+    export async function fetchPopoverData(args: FilePopoverInputs): Promise<DirPopoverFragment | FilePopoverFragment> {
+        const client = getGraphQLClient()
+        const result = await client.query(FileOrDirPopoverQuery, args)
+        if (result.error) {
+            throw new Error('could not fetch file or dir popover', result.error)
+        }
+        const fragment = result.data?.repository?.commit?.path
+        if (!fragment) {
+            throw new Error('entry does not exist')
+        }
+        return fragment
+    }
+</script>
+
+<script lang="ts">
+    import { mdiFolder } from '@mdi/js'
+
+    import Avatar from '$lib/Avatar.svelte'
+    import { pluralize } from '$lib/common'
+    import { getGraphQLClient } from '$lib/graphql'
+    import Icon from '$lib/Icon.svelte'
+    import { displayRepoName } from '$lib/shared'
+    import Timestamp from '$lib/Timestamp.svelte'
+    import { formatBytes } from '$lib/utils'
+    import Badge from '$lib/wildcard/Badge.svelte'
+
+    import FileIcon from '../FileIcon.svelte'
+
+    import NodeLine from './NodeLine.svelte'
+
+    export let repoName: string
+    export let filePath: string
+    export let entry: FilePopoverFragment | DirPopoverFragment
+
+    function splitPath(filePath: string): [string, string] {
+        let parts = filePath.split('/')
+        console.log({ parts })
+        return [parts.slice(0, parts.length - 1).join('/'), parts[parts.length - 1]]
+    }
+
+    $: [dirName, baseName] = splitPath(filePath)
+    $: lastCommit = entry.history.nodes[0].commit
+</script>
+
+<div class="root section muted">
+    <div class="repo-and-path section mono">
+        <small>
+            {displayRepoName(repoName).replaceAll('/', ' / ')}
+            ·
+            {dirName ? `${dirName.replaceAll('/', ' / ')}` : '/'}
+        </small>
+    </div>
+
+    <div class="lang-and-file section">
+        {#if entry.__typename === 'GitBlob'}
+            <FileIcon file={entry} inline={false} size="1.5rem" />
+            <div class="file mono">
+                <div>{baseName}</div>
+                <small>
+                    {entry.languages[0] ? `${entry.languages[0]} ·` : ''}
+                    {entry.totalLines}
+                    {pluralize('Line', entry.totalLines)} ·
+                    {formatBytes(entry.byteSize)}
+                </small>
+            </div>
+        {:else if entry.__typename === 'GitTree'}
+            <Icon svgPath={mdiFolder} --icon-fill-color="var(--primary)" --icon-size="1.5rem" />
+            <div class="file mono">
+                <div class="title">{baseName}</div>
+                <small>
+                    Subdirectories {entry.directories.length}
+                    · Files {entry.files.length}
+                </small>
+            </div>
+        {/if}
+    </div>
+
+    <div class="last-changed section">Last Changed @</div>
+
+    <div class="commit">
+        <div class="node-line"><NodeLine /></div>
+        <div class="commit-info">
+            <Badge variant="link">
+                <a href={lastCommit.canonicalURL} target="_blank">
+                    {lastCommit.abbreviatedOID}
+                </a>
+            </Badge>
+            <div class="body">{lastCommit.subject}</div>
+            <div class="author">
+                <Avatar avatar={lastCommit.author.person} --avatar-size="1.0rem" />
+                {lastCommit.author.person.displayName}
+                ·
+                <Timestamp date={lastCommit.author.date} />
+            </div>
+        </div>
+    </div>
+</div>
+
+<style lang="scss">
+    .root {
+        width: 400px;
+        background: var(--body-bg);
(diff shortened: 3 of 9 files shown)


.tool-versions
```diff
@@ -10,7 +10,7 @@ trivy 0.47.0
 kustomize 4.5.7
 awscli 2.4.7
 python 3.11.3 system
-rust 1.73.0
+rust 1.68.0
```
camdencheek, line 13: This doesn't look right 🤔

client/web-sveltekit/src/lib/Icon.svelte
```diff
@@ -15,7 +15,7 @@
         inline?: boolean
     }
 
-    export let svgPath: string
+    export let svgPath: string | undefined
```
camdencheek, line 18: Why would we allow an icon with no path?

client/web-sveltekit/src/lib/Popover.svelte
```diff
@@ -110,5 +111,7 @@
         border-radius: var(--popover-border-radius);
         color: var(--body-color);
         box-shadow: var(--popover-shadow);
+        border: 0;
+        border-radius: var(--popover-border-radius);
```
camdencheek, line 115: Looks like these are duplicate styles

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte
```diff
@@ -103,19 +105,28 @@
...
+                            <Icon svgPath={getDirectoryIconPath(entry, expanded)} inline />
+                        {:else}
+                            <FileIcon inline file={entry.__typename === 'GitBlob' ? entry : null} />
+                        {/if}
+                        {isRoot ? '..' : entry.name}
+                    </a>
+                    <div slot="content">
+                        <!--
+                        TODO: should revision be the revision associated with the path, not the other way around?
+                        -->
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte:126]: Not quite sure what you mean here. Mind expanding?

client/web-sveltekit/src/lib/TreeNode.svelte
```diff
@@ -34,32 +32,19 @@
...
-            // by something other than user interaction. If we always 'center' then the tree
-            // will "jump" as the user selects an entry with the keyboard or mouse, which is
-            // disorienting.
-            // But if we never 'center' then going back and forth might position the selected
-            // entry at the top or bottom of the scroll container, which is not very visible.
-            // So we only 'center' if focus is not on the tree container, which likely means
-            // that the user is not interacting with the tree.
-            label.scrollIntoView({ block: container.contains(document.activeElement) ? 'nearest' : 'center' })
-        }
-    }
```
camdencheek, line 51: Why did we delete the `scrollIntoView` stuff?

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -255,4 +295,116 @@ For example:
...
+                color: var(--text-muted);
+                margin-right: 0.5rem;
+                margin-left: 0.5rem;
+            }
+
+            .separator {
+                margin-right: 0.5rem;
+            }
+        }
+    } */
```
camdencheek, line 400: Can we delete all the commented code?

camdencheek (commented):
A couple of general comments about the design we should follow up on.


The top line is almost always going to get cut off because we're trying to fit a full repo name and a large path into a small space:

screenshot-2024-05-13_14-23-38@2x



Long commit summaries should be wrapped, not truncated. And we should linkify the PR number.

screenshot-2024-05-13_14-26-19@2x

The radius of the popover background does not match the border radius of its border.
screenshot-2024-05-13_14-27-28@2x

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -164,6 +164,46 @@ For example:
         }
     }
 
+    .header {
```
camdencheek, line 167: This whole `.header` block seems like it's duplicate?

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -255,4 +295,116 @@ For example:
         align-items: center;
         padding: 0.75rem;
     }
+
+    .last-commit {
```
camdencheek, line 299: Similarly, this whole `.last-commit` block looks duplicated?

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -255,4 +295,116 @@ For example:
...
+                margin-left: 0.5rem;
+            }
+
+            .separator {
+                margin-right: 0.5rem;
+            }
+        }
+    } */
+
+    .footer {
```
camdencheek, line 402: `.footer` block also duplicated?

client/web-sveltekit/src/lib/repo/filePopover/FilePopover.gql
```diff
@@ -0,0 +1,65 @@
...
+        oid
+        subject
+        canonicalURL
+        author {
+            date
+            person {
+                __typename
+                displayName
+                avatarURL
+                name
```
camdencheek, line 51: This should probably use the `Avatar_Person` fragment, right?

camdencheek: Same for `FilePopoverFragment` above

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte
Comment (jasonhawkharris) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte:126]: oh. I ended up fixing this. It's written in my own tongue. Wasn't meant for others to read haha

client/web-sveltekit/src/lib/repo/filePopover/FilePopover.svelte
```diff
@@ -0,0 +1,269 @@
...
+                }
+            }
+        }
+
+        .last-commit {
+            display: flex;
+            flex-flow: column nowrap;
+            align-items: flex-start;
+            justify-content: center;
+            gap: 0.5rem 0.5rem 0rem;
```
camdencheek, line 213: I don't think `gap` has a 3-param form?

client/web-sveltekit/src/lib/TreeNode.svelte
jasonhawkharris: I think I may have accidentally deleted this when reviewing a merged conflict. That's the only explanation I can think of. I'll put it back.

client/web-sveltekit/src/lib/repo/filePopover/FilePopover.svelte
```diff
@@ -0,0 +1,269 @@
...
+                height: 90px;
+                gap: 0.5rem 1.25rem;
+
+                div {
+                    display: flex;
+                    flex-flow: column nowrap;
+                    align-items: flex-start;
+                    justify-content: center;
+                    gap: 0.25rem;
+                    width: 325px;
```
camdencheek, line 239: It's weird to me that we limit the size of the commit div here. I'd expect the node line to have a fixed size and the commit section to just use the rest of the space

client/web-sveltekit/src/lib/repo/filePopover/FilePopover.svelte
```diff
@@ -0,0 +1,269 @@
...
+                    a {
+                        font-family: var(--monospace-font-family);
+                        background-color: var(--color-bg-2);
+                        padding: 0.15rem 0.25rem;
+                        border-radius: 3px;
+                    }
+
+                    .msg {
+                        color: var(--text-body);
+                        text-overflow: ellipsis ellipsis;
```
camdencheek, line 250: Only one `ellipsis` is needed. The two-param form is for when you want different behavior for clipping the start and end of the text.

```suggestion
                        text-overflow: ellipsis;
```

camdencheek: Also, it looks like this might not be working? Maybe because the width is set explicitly

client/web-sveltekit/src/lib/repo/filePopover/FilePopover.stories.svelte
camdencheek, line 1: Viewing this in storybook fails for me with `can't access property "split", filePath2 is undefined`

jasonhawkharris:
> Long commit summaries should be wrapped, not truncated. And we should linkify the PR number.
Do we actually need the repository name? I feel like the file path should suffice, since by the time the user get's to a file tree view, they'll more than likely know what repo they're in. Right?

client/web-sveltekit/src/lib/repo/filePopover/FilePopover.gql
```diff
@@ -0,0 +1,65 @@
...
+
+fragment DirPopoverFragment on GitTree {
+    path
+    isDirectory
+    files {
+        name
+    }
+    directories {
+        name
+    }
```
camdencheek, line 39: What happens if there are a stupid number of files or directories?

camdencheek (commented):
Looks a little odd that there is no border here

screenshot-2024-05-13_15-05-19@2x

client/web-sveltekit/src/lib/repo/filePopover/FilePopover.gql
camdencheek: [This](https://sourcegraph.sourcegraph.com/github.com/sgtest/weird/-/tree/dirs/wide) might be useful for testing 🙂

jasonhawkharris: hmmm... Good question. I'll think on that. Thanks for bringing it up.

jasonhawkharris:
> Looks a little odd that there is no border here

This was part of the original design. I have to say, I like it better in dark mode. It does look a little awkward in light mode.

client/web-sveltekit/src/lib/Icon.svelte
jasonhawkharris: I wrote this in to test something and forgot to delete it.

client/web-sveltekit/src/lib/repo/filePopover/FilePopover.svelte
jasonhawkharris: I believe this will read: top, right and left, bottom.

camdencheek: `gap` is different than `margin` or `padding` because it's space between, not space around. It only supports at most 2 arguments, one for the gap between rows and one for the gap between columns

https://developer.mozilla.org/en-US/docs/Web/CSS/gap

client/web-sveltekit/src/lib/repo/filePopover/FilePopover.gql
jasonhawkharris: All good here. However, I did find another issue using the repo you sent over. Empty files cause a crash because they have no 'blame' object. I'm working around this for the time being by checking if blame is not an empty repo before assigning the data to filePopover, but this could probably use a follow-up issue.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte
```diff
@@ -105,7 +105,7 @@
                     We handle navigation via the TreeView's select event, to preserve the focus state.
                     Using a link here allows us to benefit from data preloading.
                 -->
-                <Popover let:registerTrigger placement="right-end" showOnHover>
+                <Popover let:registerTrigger placement="right-end" showOnHover delay={300}>
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte:108]: Oh, that's nice and easy 🙂

client/web-sveltekit/src/lib/Popover.svelte
```diff
@@ -123,6 +123,6 @@
         color: var(--body-color);
         box-shadow: var(--dropdown-shadow);
         border: 0;
-        border-radius: var(--popover-border-radius);
+        border-radius: var(--border-radius);
```
camdencheek, line 126: Why did this change?

client/web-sveltekit/src/lib/TreeNode.svelte
camdencheek, line 1: Why did anything in this file change?

client/web-sveltekit/src/lib/TreeNode.svelte
```diff
@@ -49,17 +37,29 @@
...
+    class:selected
     role="treeitem"
     aria-selected={selectable ? selected : undefined}
     aria-expanded={expandable ? expanded : undefined}
     {tabindex}
-    data-treeitem
     data-node-id={nodeID}
 >
-    <span bind:this={label} class="label" data-treeitem-label>
+    <span class="label" data-treeitem-label>
```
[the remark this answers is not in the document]
jasonhawkharris: What on earth!!!

client/web-sveltekit/src/lib/TreeNode.svelte
jasonhawkharris: 🤦🏼

client/web-sveltekit/src/lib/repo/filePopover/FilePopover.svelte
```diff
@@ -0,0 +1,271 @@
...
+                filePopover = result?.data?.repository?.commit?.path
+            }
+            return filePopover
+        } else if (result?.data?.repository?.commit?.path?.__typename === 'GitTree') {
+            dirPopover = result?.data?.repository?.commit?.path
+            return dirPopover
+        }
+
+        return null
+    }
```
camdencheek, line 50: Why did we depart from the pattern established for `RepoPopover` and do the data fetching in the component rather than in the `+page.ts`?

client/web-sveltekit/src/lib/repo/filePopover/FilePopover.svelte
```diff
@@ -0,0 +1,271 @@
...
+
+    const CENTER_DOT = '\u00B7' // interpunct
+
+    $: abbreviatedPath = abbreviatedFilePath(filePath)
+    $: repo = displayRepoName(repoName).replace('/', ' / ')
+    $: repoAndPath = `${repo} ${CENTER_DOT} ${abbreviatedPath}`
+    $: fileOrDirName = getFileName(filePath)
+    $: fileCommit = filePopover?.blame[0].commit
+    $: languages = filePopover?.languages[0]
```
camdencheek, line 71: What if languages is empty?

client/web-sveltekit/src/lib/Popover.svelte
jasonhawkharris: [There are two values.](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@e2814f5fdc2faed81d5026d4031b56a4b87d168b/-/blob/client/wildcard/src/global-styles/utilities/border-radius.scss?L6) --popover-border-radius (equal to 5px) and --border-radius (equal to 3px).

camdencheek: Right, why did you change the `Popover` component to use `--border-radius`?

client/web-sveltekit/src/lib/repo/filePopover/FilePopover.svelte
jasonhawkharris: @camdencheek Long story short, I tried to implement it around 3 or 4 times and it was a giant pain each time, always with many bugs to work through. Why does it matter if we do it here? It seems to be better here because it's contained directly in the component that calls it. It also only mounts on hover, so we're only calling it when we need it.

client/web-sveltekit/src/lib/repo/filePopover/FilePopover.svelte
jasonhawkharris:
```
<small
    >{languages ? filePopover.languages[0] : ''}
    {languages ? CENTER_DOT : ''}
    {filePopover.totalLines} Lines {CENTER_DOT} Bytes {formatBytes(filePopover.byteSize)}</small
>
```

client/web-sveltekit/src/lib/repo/filePopover/FilePopover.svelte
camdencheek: In general, we want to separate data fetching from rendering because the source of the data might be different when we use the component in different places. e.g., we may want to batch these calls, or cancel a call if you hover over something different, or preload the data for some locations. 

Another more concrete reason is that this is now _much_ harder to test because, rather than just passing in the data it needs to render, you now need to mock out the API call.

camdencheek: Another concrete example is, if you re-hover, we now re-fetch because the component is re-created.

client/web-sveltekit/src/lib/repo/filePopover/FilePopover.svelte
camdencheek: Ah, I see. Forgot that out-of-bounds array accesses in javascript just evaluate to undefined.

client/web-sveltekit/src/lib/Popover.svelte
jasonhawkharris: ahhh... I see what you're saying. I'll change it.

client/web-sveltekit/src/lib/repo/filePopover/FilePopover.svelte
jasonhawkharris: Okay. I see.

jasonhawkharris: In that case, with the little time I have left, I think I need to cry "uncle" here and pass this along to someone else. I don't see me getting this done today. I've tried a number of times and failed to reproduce what we did with RepoPopover.

taiyab:
With @jasonhawkharris on leave, any chance this gets merged anytime soon? @camdencheek

camdencheek:
Yep, this is first on my list tomorrow

camdencheek (approved):

taiyab:
@camdencheek @jasonhawkharris As a follow up to this:
- The path needs each individual dir name to be linked
- We need a solution to show the full path, always (so decently presented wrapping, slightly wider popover maybe too, condensing space between `/` and dir names a bit)
- link the commit message to the commit too

PR #62505, opened 2024-05-07
fix: deprecation in pkg call

jdpleiness:
Fix minor deprecation when calling the `intstr` pkg.

## Test plan
Previous tests apply


Chickensoupwithrice (approved):

Chickensoupwithrice (commented):
Deprecation notice: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/intstr#FromInt

PR #62508, opened 2024-05-07
msp: add infra and runtime support for job checkins

bobheadxi:
Closes CORE-21 - allows jobs to register check-ins using Sentry when they are configured as cron jobs: https://docs.sentry.io/product/crons/, for a nice view of "is my job running or nah" without using GCP's less-than-beautiful console views

1. Adds the configured schedule and deadline as environment variables for MSP jobs
2. Adds a contract mechanism for checking in, for example:
```go
	func work(ctx context.Context) (err error) {
		done, err := c.Diagnostics.JobExecutionCheckIn(ctx)
		if err != nil { /* failed to register check-in */ }
		defer done(err)

		// ... do work
	}
```

## Test plan

```sh
TestJobExecutionCheckIn_SENTRY_DSN='...' go test -v ./runtime/contract
```

image

In Slack:

image

It appears the message is not currently customizable: https://develop.sentry.dev/sdk/check-ins/


lib/managedservicesplatform/runtime/contract/diagnostics.go
```diff
@@ -25,6 +32,9 @@ type diagnosticsContract struct {
 	OpenTelemetry opentelemetry.Config
 	sentryDSN     *string
 
+	cronSchedule *string
```
rafax, line 35: Is schedule really optional? Is there a default?

rafax (approved):
Looks reasonable, but I'm not the expert on msp

lib/managedservicesplatform/runtime/contract/diagnostics.go
rafax: Ok fair, didn't notice this is shared between services and jobs

unknwon: ----- Split of topic -----

It's kinda surpising to see cron schedule is part of "diagnostics"? 🤔 or is it for **exposing** the cron schedule for diagnostics?

lib/managedservicesplatform/runtime/contract/diagnostics.go
```diff
@@ -47,6 +58,10 @@ func loadDiagnosticsContract(
 		},
 		sentryDSN: env.GetOptional("SENTRY_DSN", "Sentry error reporting DSN"),
 
+		// only MSP jobs with a schedule configured will have these values
```
unknwon, line 61:
```suggestion
		// Only MSP jobs with a schedule configured will have these values.
```

lib/managedservicesplatform/runtime/contract/diagnostics.go
```diff
@@ -200,3 +215,152 @@ func (c diagnosticsContract) ConfigureSentry(liblog *log.PostInitCallbacks) bool
...
+	}
+
+	monitor := fmt.Sprintf("%s-%s", c.internal.service.Name(), c.internal.environmentID)
+	monitorConfig := &sentry.MonitorConfig{
+		Schedule:   sentry.CrontabSchedule(*c.cronSchedule),
+		MaxRuntime: int64(math.Ceil(c.cronDeadline.Minutes())),
+		// Half the deadline is used as a margin to allow for a grace period
+		// for execution to report they are active
+		CheckInMargin: int64(math.Ceil(c.cronDeadline.Minutes() / 2)),
+		// Set in MSP
```
unknwon, line 292:
```suggestion
		// When running in MSP, the timezone is always set to UTC.
```

Assuming this is the actual intention.

"Set in MSP" sounds like an action to do, not a fact grabbed from a source.

unknwon (approved):
TIL!

lib/managedservicesplatform/runtime/contract/diagnostics.go
bobheadxi: Yeah it's only exposing the cron schedule for diagnostics, we don't (can't) actually apply the cron schedule from the runtime

PR #62510, opened 2024-05-07
Rockskip: simplify ruler function

jtibshirani:
Adds a test for the ruler function and simplifies the function to remove tail
recursion.

A tiny warm-up as we work on improving Rockskip test coverage.

## Test plan

Added new test case.

### DIFF
--- a/internal/rockskip/server.go
+++ b/internal/rockskip/server.go
@@ -266,11 +266,10 @@ func DeleteOldRepos(ctx context.Context, db *sql.DB, maxRepos int, threadStatus 
 //
 // https://oeis.org/A007814
 func ruler(n int) int {
-	if n == 0 {
-		return 0
+	height := 0
+	for n > 0 && n%2 == 0 {
+		height++
+		n = n / 2
 	}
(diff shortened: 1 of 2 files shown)


internal/rockskip/server.go
```diff
@@ -266,11 +266,10 @@ func DeleteOldRepos(ctx context.Context, db *sql.DB, maxRepos int, threadStatus
...
-		return 0
+	height := 0
+	for n > 0 && n%2 == 0 {
+		height++
+		n = n / 2
 	}
-	if n%2 != 0 {
-		return 0
-	}
-	return 1 + ruler(n/2)
```
mmanela, line 275: Out of curiosity, does go optimize away tail recursion like many functional programming languages do?

mmanela (approved):

internal/rockskip/server.go
jtibshirani: I believe tail-call optimization is *not* widely implemented (https://stackoverflow.com/questions/12102675/tail-call-optimization-in-go).

PR #62502, opened 2024-05-07
appliance: deploy redis services

craigfurman:
**appliance: delete redis-exporter config**

**appliance: remove unused extra env vars config fields**

**appliance/compare-helm: support multiple components**

Helps with redis services, which have helm templates with component
names including redis, redis-cache, redis-store.

**appliance: deploy redis services**

---

Closes https://github.com/sourcegraph/sourcegraph-operator/issues/98

## Test plan

Golden tests included

PR #62512, opened 2024-05-08
appliance: add pgsql service definition

jdpleiness:
Add `pgsql` service definition.

## Test plan

Golden files

### DIFF
--- a/internal/appliance/config/embed.go
+++ b/internal/appliance/config/embed.go
@@ -8,13 +8,13 @@ var (
 	//go:embed pgsql/postgresql.conf
 	fs embed.FS
 
-	postgresqlConfig []byte
+	pgsqlConfig []byte
 )
 
 func init() {
-	postgresqlConfig, _ = fs.ReadFile("pgsql/postgresql.conf")
+	pgsqlConfig, _ = fs.ReadFile("pgsql/postgresql.conf")
 }
 
-func PostgresqlConfig() string {
-	return string(postgresqlConfig)
+func DefaultPGSQLConfig() string {
+	return string(pgsqlConfig)
 }
--- a/internal/appliance/config/embed.go
+++ b/internal/appliance/config/embed.go
@@ -1,0 +1,20 @@
+package config
+
+import (
+	"embed"
+)
+
+var (
+	//go:embed pgsql/postgresql.conf
+	fs embed.FS
+
+	postgresqlConfig []byte
+)
+
+func init() {
+	postgresqlConfig, _ = fs.ReadFile("pgsql/postgresql.conf")
+}
+
+func PostgresqlConfig() string {
+	return string(postgresqlConfig)
+}
--- a/internal/appliance/defaults.go
+++ b/internal/appliance/defaults.go
@@ -48,7 +48,7 @@ func newDefaultConfig() Sourcegraph {
 			},
 			PGSQL: PGSQLSpec{
 				StandardConfig: config.StandardConfig{
-					PrometheusPort: pointers.Ptr(6060),
+					PrometheusPort: pointers.Ptr(9187),
 				},
 				StorageSize: "200Gi",
 				DatabaseConnection: &DatabaseConnectionSpec{
--- a/internal/appliance/defaults.go
+++ b/internal/appliance/defaults.go
@@ -3,10 +3,9 @@ package appliance
 import (
 	"fmt"
 
-	"k8s.io/utils/ptr"
-
 	"github.com/sourcegraph/sourcegraph/internal/appliance/config"
 	"github.com/sourcegraph/sourcegraph/lib/errors"
+	"github.com/sourcegraph/sourcegraph/lib/pointers"
 )
 
 // Default config.
@@ -27,7 +26,7 @@ func newDefaultConfig() Sourcegraph {
 			},
 			RepoUpdater: RepoUpdaterSpec{
 				StandardConfig: config.StandardConfig{
-					PrometheusPort: ptr.To(6060),
+					PrometheusPort: pointers.Ptr(6060),
 				},
 			},
 			StorageClass: StorageClassSpec{
@@ -35,27 +34,40 @@ func newDefaultConfig() Sourcegraph {
 			},
 			Symbols: SymbolsSpec{
 				StandardConfig: config.StandardConfig{
-					PrometheusPort: ptr.To(6060),
+					PrometheusPort: pointers.Ptr(6060),
 				},
 				Replicas:    1,
 				StorageSize: "12Gi",
 			},
 			GitServer: GitServerSpec{
 				StandardConfig: config.StandardConfig{
-					PrometheusPort: ptr.To(6060),
+					PrometheusPort: pointers.Ptr(6060),
 				},
 				Replicas:    1,
 				StorageSize: "200Gi",
 			},
+			PGSQL: PGSQLSpec{
+				StandardConfig: config.StandardConfig{
+					PrometheusPort: pointers.Ptr(6060),
+				},
+				StorageSize: "200Gi",
+				DatabaseConnection: &DatabaseConnectionSpec{
+					Host:     "pgsql",
+					Port:     "5432",
+					User:     "sg",
+					Password: "password",
+					Database: "sg",
+				},
+			},
 			RedisCache: RedisSpec{
 				StandardConfig: config.StandardConfig{
-					PrometheusPort: ptr.To(9121),
+					PrometheusPort: pointers.Ptr(9121),
 				},
 				StorageSize: "100Gi",
 			},
 			RedisStore: RedisSpec{
 				StandardConfig: config.StandardConfig{
-					PrometheusPort: ptr.To(9121),
+					PrometheusPort: pointers.Ptr(9121),
 				},
 				StorageSize: "100Gi",
 			},
@@ -71,8 +83,11 @@ var defaultImages = map[string]map[string]string{
 }
 
 var defaultImagesForVersion_5_3_9104 = map[string]string{
+	"alpine":         "alpine-3.14:5.3.2@sha256:982220e0fd8ce55a73798fa7e814a482c4807c412f054c8440c5970b610239b7",
 	"blobstore":      "blobstore:5.3.2@sha256:d625be1eefe61cc42f94498e3c588bf212c4159c8b20c519db84eae4ff715efa",
 	"gitserver":      "gitserver:5.3.2@sha256:6c6042cf3e5f3f16de9b82e3d4ab1647f8bb924cd315245bd7a3162f5489e8c4",
+	"pgsql":          "postgres-12-alpine:5.3.2@sha256:1e0e93661a65c832b9697048c797f9894dfb502e2e1da2b8209f0018a6632b79",
+	"pgsql-exporter": "postgres_exporter:5.3.2@sha256:b9fa66fbcb4cc2d466487259db4ae2deacd7651dac4a9e28c9c7fc36523699d0",
 	"redis-cache":    "redis-cache:5.3.2@sha256:ed79dada4d1a2bd85fb8450dffe227283ab6ae0e7ce56dc5056fbb8202d95624",
 	"redis-exporter": "redis_exporter:5.3.2@sha256:21a9dd9214483a42b11d58bf99e4f268f44257a4f67acd436d458797a31b7786",
 	"redis-store":    "redis-store:5.3.2@sha256:0e3270a5eb293c158093f41145810eb5a154f61a74c9a896690dfdecd1b98b39",
--- a/internal/appliance/pgsql.go
+++ b/internal/appliance/pgsql.go
@@ -67,11 +67,7 @@ func (r *Reconciler) reconcilePGSQLStatefulSet(ctx context.Context, sg *Sourcegr
 		},
 	})
 
-	databaseSecretName := cfg.DatabaseSecret
-	if databaseSecretName == "" {
-		databaseSecretName = "pgsql-auth"
-	}
-
+	databaseSecretName := "pgsql-auth"
 	ctr.Env = append(ctr.Env, container.EnvVarsPostgres(databaseSecretName)...)
 
 	ctr.Ports = []corev1.ContainerPort{{Name: name, ContainerPort: 5432}}
@@ -113,8 +109,7 @@ func (r *Reconciler) reconcilePGSQLStatefulSet(ctx context.Context, sg *Sourcegr
 		return err
 	}
 
-	// TODO figure out method to override container image
-	initCtr := container.NewContainer("correct-data-dir-permissions", nil, config.ContainerConfig{
+	initCtr := container.NewContainer("correct-data-dir-permissions", cfg, config.ContainerConfig{
 		Image: initCtrImage,
 		Resources: &corev1.ResourceRequirements{
 			Requests: corev1.ResourceList{
@@ -136,8 +131,7 @@ func (r *Reconciler) reconcilePGSQLStatefulSet(ctx context.Context, sg *Sourcegr
 		return err
 	}
 
-	// TODO figure out method to override container image
-	pgExpCtr := container.NewContainer("pgsql-exporter", nil, config.ContainerConfig{
+	pgExpCtr := container.NewContainer("pgsql-exporter", cfg, config.ContainerConfig{
 		Image: pgExpCtrImage,
 		Resources: &corev1.ResourceRequirements{
 			Requests: corev1.ResourceList{
@@ -177,6 +171,7 @@ func (r *Reconciler) reconcilePGSQLStatefulSet(ctx context.Context, sg *Sourcegr
 	}
 
 	podTemplate := pod.NewPodTemplate(name, cfg)
+	podTemplate.Template.Spec.TerminationGracePeriodSeconds = pointers.Ptr[int64](120)
 	podTemplate.Template.Spec.InitContainers = []corev1.Container{initCtr}
 	podTemplate.Template.Spec.Containers = []corev1.Container{ctr, pgExpCtr}
 	podTemplate.Template.Spec.ServiceAccountName = name
@@ -202,7 +197,7 @@ func (r *Reconciler) reconcilePGSQLPersistentVolumeClaim(ctx context.Context, sg
 
 func (r *Reconciler) reconcilePGSQLConfigMap(ctx context.Context, sg *Sourcegraph, owner client.Object) error {
 	cm := configmap.NewConfigMap("pgsql-conf", sg.Namespace)
-	cm.Data = map[string]string{"postgresql.conf": config.PostgresqlConfig()}
+	cm.Data = map[string]string{"postgresql.conf": config.DefaultPGSQLConfig()}
 
 	return reconcileObject(ctx, r, sg.Spec.PGSQL, &cm, &corev1.ConfigMap{}, sg, owner)
 }
--- a/internal/appliance/pgsql.go
+++ b/internal/appliance/pgsql.go
@@ -1,0 +1,239 @@
+package appliance
+
+import (
+	"context"
+
+	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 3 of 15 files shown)


internal/appliance/defaults.go
```diff
@@ -27,26 +26,39 @@ func newDefaultConfig() Sourcegraph {
 			},
 			RepoUpdater: RepoUpdaterSpec{
 				StandardConfig: config.StandardConfig{
-					PrometheusPort: ptr.To(6060),
+					PrometheusPort: pointers.Ptr(6060),
```
jdpleiness, line 29: Switched these over to use our own pointers pkg

internal/k8s/resource/container/container.go
```diff
@@ -115,3 +115,124 @@ func EnvVarsOtel() []corev1.EnvVar {
 		{Name: "OTEL_EXPORTER_OTLP_ENDPOINT", Value: "http://$(OTEL_AGENT_HOST):4317"},
 	}
 }
+
+func EnvVarsPostgres(secretName string) []corev1.EnvVar {
```
jdpleiness, line 119: These are used by `pgsql`, `codeintel`, and `codeinsights` so it makes sense to me to put them here.

internal/k8s/resource/container/container.go
```diff
@@ -115,3 +115,124 @@ func EnvVarsOtel() []corev1.EnvVar {
...
+			},
+		},
+		{
+			Name:  "POSTGRES_DB",
+			Value: "$(POSTGRES_DATABASE)",
+		},
+	}
+}
+
+func EnvVarsPostgresExporter(secretName string) []corev1.EnvVar {
```
jdpleiness, line 183: This is also used by all three database services

internal/appliance/defaults.go
```diff
@@ -27,26 +26,39 @@ func newDefaultConfig() Sourcegraph {
...
 				},
 				Replicas:    1,
 				StorageSize: "200Gi",
 			},
+			PGSQL: PGSQLSpec{
+				StandardConfig: config.StandardConfig{
+					PrometheusPort: pointers.Ptr(6060),
+				},
+				StorageSize: "200Gi",
+				DatabaseConnection: &DatabaseConnectionSpec{
```
jdpleiness, line 54: I added these in as defaults, as that is the behavior we want if no secret or overrides are specified, but I don't see how these defaults ever get used, or is that still WIP?

internal/appliance/pgsql.go
```diff
@@ -0,0 +1,239 @@
...
+		{Name: "dshm", MountPath: "/dev/shm"},
+		{Name: "lockdir", MountPath: "/var/run/postgresql"},
+	}
+
+	initCtrImage, err := getDefaultImage(sg, "alpine")
+	if err != nil {
+		return err
+	}
+
+	// TODO figure out method to override container image
```
jdpleiness, line 116: Like this says, atm there is no way for a user to set the container image override on this, but still WIP

craigfurman: Overriding container images is a standard-config feature, recently merged: https://github.com/sourcegraph/sourcegraph/blob/6cd5834127cdc88241b78c4dd684e004aa818aab/internal/appliance/config/config.go#L26

Looks like you already embedded StandardConfig in spec.go, so you if you pass the pgsql config element into NewContainer on the line below this, that override will work 🙌 

Overrides would look something like this in the spec, as the string keys of the containerConfig map are container names:

```yaml
pgsql:
  containerConfig:
    pgsql:
      image: ...
      resources: ...
    correct-data-dir-permissions:
      image: ...
      resources: ...
    pgsql-exporter:
      image: ...
      resources: ...
```

Holler if you want to chat/pair on this. If in doubt, write a golden test for the overrides!

internal/appliance/testdata/sg/pgsql/with-storage.yaml
```diff
@@ -0,0 +1,56 @@
...
+    disabled: true
+
+  indexedSearch:
+    disabled: true
+
+  indexedSearchIndexer:
+    disabled: true
+
+  pgsql:
+    storageSize: "500Gi"
```
craigfurman, line 26: This is something I was thinking of making a standard feature out of, but that can be done later in a refactor. If it introduces weird-looking if-statements, better to have a little duplication. No action needed, just a note.

internal/k8s/resource/container/container.go
craigfurman: 👍 

You can call NewEnvVarSecretKeyRef here btw, like in EnvVarsRedis() above

internal/appliance/spec.go
```diff
@@ -158,25 +158,20 @@ type IndexedSearchIndexerSpec struct {
...
 type PGSQLSpec struct {
-	// Disabled defines if pgsql is enabled or not.
-	// Default: false
-	Disabled bool `json:"disabled,omitempty"`
+	config.StandardConfig
 
-	// ExistingSecret is the name of an existing secret to use for Postgres credentials.
-	ExistingSecret string `json:"existingSecret,omitempty"`
+	// DatabaseSecret is the name of an existing secret to use for Postgres connection string.
+	DatabaseSecret string `json:"databaseSecret,omitempty"`
```
craigfurman, line 166: It might make things simpler to not make secret names configurable. Users with external DBs could just disable pgsql then create their secret in SG's namespace with the expected name. See this thread: https://github.com/sourcegraph/sourcegraph/pull/62502#discussion_r1592884183

I don't mind which way the decision goes, but it might make sense to make it and apply it to both PRs. WDYT?

internal/appliance/defaults.go
```diff
@@ -27,26 +26,39 @@ func newDefaultConfig() Sourcegraph {
...
 				StandardConfig: config.StandardConfig{
-					PrometheusPort: ptr.To(6060),
+					PrometheusPort: pointers.Ptr(6060),
 				},
 				Replicas:    1,
 				StorageSize: "200Gi",
 			},
+			PGSQL: PGSQLSpec{
+				StandardConfig: config.StandardConfig{
+					PrometheusPort: pointers.Ptr(6060),
```
craigfurman, line 51:
```suggestion
					PrometheusPort: pointers.Ptr(9187),
```

See compare-helm diff comment

internal/appliance/testdata/sg/pgsql/default.yaml
```diff
@@ -0,0 +1,55 @@
+spec:
```
craigfurman, line 1: Annotated output of compare helm:

```
  -component pgsql \
  -golden-file internal/appliance/testdata/golden-fixtures/pgsql/default.yaml

```

```diff
2c2
< # helm: Secret/pgsql-auth
---
> # golden: Secret/pgsql-auth
9a10
> immutable: false
11a13,15
>   annotations:
>     appliance.sourcegraph.com/configHash: 86fa1e5cbb702a914107796b2236964ac6042a4f399010458f077f8a854c0cc8
>   creationTimestamp: "2024-04-19T00:00:00Z"
13,14d16
<     app: pgsql
<     app.kubernetes.io/component: pgsql
16a19,28
>   namespace: NORMALIZED_FOR_TESTING
>   ownerReferences:
>   - apiVersion: v1
>     blockOwnerDeletion: true
>     controller: true
>     kind: ConfigMap
>     name: sg
>     uid: NORMALIZED_FOR_TESTING
>   resourceVersion: NORMALIZED_FOR_TESTING
>   uid: NORMALIZED_FOR_TESTING
19c31
< # helm: ConfigMap/pgsql-conf
---
> # golden: ConfigMap/pgsql-conf
78a91
> immutable: false
82c95,96
<     description: Configuration for PostgreSQL
---
>     appliance.sourcegraph.com/configHash: 86fa1e5cbb702a914107796b2236964ac6042a4f399010458f077f8a854c0cc8
>   creationTimestamp: "2024-04-19T00:00:00Z"
84d97
<     app.kubernetes.io/component: pgsql
86a100,109
>   namespace: NORMALIZED_FOR_TESTING
>   ownerReferences:
>   - apiVersion: v1
>     blockOwnerDeletion: true
>     controller: true
>     kind: ConfigMap
>     name: sg
>     uid: NORMALIZED_FOR_TESTING
>   resourceVersion: NORMALIZED_FOR_TESTING
>   uid: NORMALIZED_FOR_TESTING
88c111
< # helm: PersistentVolumeClaim/pgsql
---
> # golden: PersistentVolumeClaim/pgsql
91a115,119
>   annotations:
>     appliance.sourcegraph.com/configHash: 86fa1e5cbb702a914107796b2236964ac6042a4f399010458f077f8a854c0cc8
>   creationTimestamp: "2024-04-19T00:00:00Z"
>   finalizers:
>   - kubernetes.io/pvc-protection
93d120
<     app.kubernetes.io/component: pgsql
95a123,132
>   namespace: NORMALIZED_FOR_TESTING
>   ownerReferences:
>   - apiVersion: v1
>     blockOwnerDeletion: true
>     controller: true
>     kind: ConfigMap
>     name: sg
>     uid: NORMALIZED_FOR_TESTING
>   resourceVersion: NORMALIZED_FOR_TESTING
>   uid: NORMALIZED_FOR_TESTING
102a140,142
>   volumeMode: Filesystem
> status:
>   phase: Pending
104c144
< # helm: Service/pgsql
---
> # golden: Service/pgsql
109c149,150
<     prometheus.io/port: "9187"
---
>     appliance.sourcegraph.com/configHash: 86fa1e5cbb702a914107796b2236964ac6042a4f399010458f077f8a854c0cc8
>     prometheus.io/port: "6060"
```

Noted in defaults.go

```diff
110a152
>   creationTimestamp: "2024-04-19T00:00:00Z"
115a158,167
>   namespace: NORMALIZED_FOR_TESTING
>   ownerReferences:
>   - apiVersion: v1
>     blockOwnerDeletion: true
>     controller: true
>     kind: ConfigMap
>     name: sg
>     uid: NORMALIZED_FOR_TESTING
>   resourceVersion: NORMALIZED_FOR_TESTING
>   uid: NORMALIZED_FOR_TESTING
116a169,175
>   clusterIP: NORMALIZED_FOR_TESTING
>   clusterIPs:
>   - NORMALIZED_FOR_TESTING
>   internalTrafficPolicy: Cluster
>   ipFamilies:
>   - IPv4
>   ipFamilyPolicy: SingleStack
119a179
>     protocol: TCP
123,124c183
<     app.kubernetes.io/instance: release-name
<     app.kubernetes.io/name: sourcegraph
---
>   sessionAffinity: None
125a185,186
> status:
>   loadBalancer: {}
127c188
< # helm: StatefulSet/pgsql
---
> # golden: StatefulSet/pgsql
132c193,195
<     description: Postgres database for various data.
---
>     appliance.sourcegraph.com/configHash: 86fa1e5cbb702a914107796b2236964ac6042a4f399010458f077f8a854c0cc8
>   creationTimestamp: "2024-04-19T00:00:00Z"
>   generation: 1
135,136d197
<     app.kubernetes.io/instance: release-name
<     app.kubernetes.io/managed-by: Helm
138c199
<     app.kubernetes.io/version: 5.3.2
---
>     app.kubernetes.io/version: 5.3.9104
140d200
<     helm.sh/chart: sourcegraph-5.3.2
141a202,211
>   namespace: NORMALIZED_FOR_TESTING
>   ownerReferences:
>   - apiVersion: v1
>     blockOwnerDeletion: true
>     controller: true
>     kind: ConfigMap
>     name: sg
>     uid: NORMALIZED_FOR_TESTING
>   resourceVersion: NORMALIZED_FOR_TESTING
>   uid: NORMALIZED_FOR_TESTING
142a213,217
>   minReadySeconds: 10
>   persistentVolumeClaimRetentionPolicy:
>     whenDeleted: Retain
>     whenScaled: Retain
>   podManagementPolicy: OrderedReady
148,149d222
<       app.kubernetes.io/instance: release-name
<       app.kubernetes.io/name: sourcegraph
154d226
<         checksum/pgsql.secret: 180940fdb956526d197a8efaf15bc2f14a3db83e09610917f8b9040fa5232d39
155a228
>       creationTimestamp: null
158,159d230
<         app.kubernetes.io/instance: release-name
<         app.kubernetes.io/name: sourcegraph
161c232
<         group: backend
---
>       name: pgsql
163d233
<       affinity: null
198a269
>           failureThreshold: 3
199a271,273
>           periodSeconds: 10
>           successThreshold: 1
>           timeoutSeconds: 1
203a278
>           protocol: TCP
207a283,286
>           failureThreshold: 3
>           periodSeconds: 10
>           successThreshold: 1
>           timeoutSeconds: 1
218,219c297,298
<           runAsGroup: 999
<           runAsUser: 999
---
>           runAsGroup: 101
>           runAsUser: 100
```

This user ID diff might stop the process from running successfully, not 100%
sure though. Depends on what's in the rootFS. May as well squash the diff?
Similar diffs for other containers although I haven't commented on these.

```diff
225a305,307
>           successThreshold: 1
>           timeoutSeconds: 1
>         terminationMessagePath: /dev/termination-log
261a344
>         imagePullPolicy: IfNotPresent
266c349
<             memory: 50Mi
---
>             memory: 50M
```

Default resources are often differing by M vs Mi in this diff. Very minor note,
probably not a real problem, but may as well squash the diff.

```diff
269,270c352,358
<             memory: 50Mi
<         securityContext: null
---
>             memory: 50M
>         securityContext:
>           allowPrivilegeEscalation: false
>           readOnlyRootFilesystem: true
>           runAsGroup: 101
>           runAsUser: 100
>         terminationMessagePath: /dev/termination-log
271a360
>       dnsPolicy: ClusterFirst
283c372
<             memory: 50Mi
---
>             memory: 50M
286c375
<             memory: 50Mi
---
>             memory: 50M
290,291c379,382
<           runAsGroup: 999
<           runAsUser: 999
---
>           runAsGroup: 101
>           runAsUser: 100
>         terminationMessagePath: /dev/termination-log
>         terminationMessagePolicy: FallbackToLogsOnError
295c386,387
<       nodeSelector: null
---
>       restartPolicy: Always
>       schedulerName: default-scheduler
297c389
<         fsGroup: 999
---
>         fsGroup: 101
299,302c391,395
<         runAsGroup: 999
<         runAsUser: 999
<       terminationGracePeriodSeconds: 120
<       tolerations: null
---
>         runAsGroup: 101
>         runAsUser: 100
>       serviceAccount: pgsql
>       serviceAccountName: pgsql
>       terminationGracePeriodSeconds: 30
```

terminationGracePeriodSeconds differs here (it's longer in helm).
Looks like we also unconditionally use a ServiceAccount here but not in helm,
which is probably harmless! I probably should have passed an extra-arg to my
compare-helm invocation to account for this.

```diff
303a397,402
>       - emptyDir: {}
>         name: lockdir
>       - emptyDir:
>           medium: Memory
>           sizeLimit: 1Gi
>         name: dshm
311,316d409
<       - emptyDir:
<           medium: Memory
<           sizeLimit: 1G
<         name: dshm
<       - emptyDir: {}
<         name: lockdir
318a412,505
> status:
>   availableReplicas: 0
>   replicas: 0
> ---
> # golden: ConfigMap/sg
> apiVersion: v1
> data:
>   spec: |
>     spec:
>       requestedVersion: "5.3.9104"
> 
>       blobstore:
>         disabled: true
> 
>       codeInsights:
>         disabled: true
> 
>       codeIntel:
>         disabled: true
> 
>       frontend:
>         disabled: true
> 
>       gitServer:
>         disabled: true
> 
>       indexedSearch:
>         disabled: true
> 
>       indexedSearchIndexer:
>         disabled: true
> 
>       pgsql: {}
> 
>       postgresExporter:
>         disabled: true
> 
>       preciseCodeIntel:
>         disabled: true
> 
>       redisCache:
>         disabled: true
> 
>       redisExporter:
>         disabled: true
> 
>       redisStore:
>         disabled: true
> 
>       repoUpdater:
>         disabled: true
> 
>       searcher:
>         disabled: true
> 
>       symbols:
>         disabled: true
> 
>       syntectServer:
>         disabled: true
> 
>       worker:
>         disabled: true
> kind: ConfigMap
> metadata:
>   annotations:
>     appliance.sourcegraph.com/currentVersion: 5.3.9104
>     appliance.sourcegraph.com/managed: "true"
>   creationTimestamp: "2024-04-19T00:00:00Z"
>   name: sg
>   namespace: NORMALIZED_FOR_TESTING
>   resourceVersion: NORMALIZED_FOR_TESTING
>   uid: NORMALIZED_FOR_TESTING
> ---
> # golden: ServiceAccount/pgsql
> apiVersion: v1
> kind: ServiceAccount
> metadata:
>   annotations:
>     appliance.sourcegraph.com/configHash: 86fa1e5cbb702a914107796b2236964ac6042a4f399010458f077f8a854c0cc8
>   creationTimestamp: "2024-04-19T00:00:00Z"
>   labels:
>     deploy: sourcegraph
>   name: pgsql
>   namespace: NORMALIZED_FOR_TESTING
>   ownerReferences:
>   - apiVersion: v1
>     blockOwnerDeletion: true
>     controller: true
>     kind: ConfigMap
>     name: sg
>     uid: NORMALIZED_FOR_TESTING
>   resourceVersion: NORMALIZED_FOR_TESTING
>   uid: NORMALIZED_FOR_TESTING
```

craigfurman (commented):
A few issues with the compare-helm diff, some more notes and replies. A great start, this is definitely nearly ready!

internal/appliance/config/embed.go
```diff
@@ -0,0 +1,20 @@
...
+	postgresqlConfig []byte
+)
+
+func init() {
+	postgresqlConfig, _ = fs.ReadFile("pgsql/postgresql.conf")
+}
+
+func PostgresqlConfig() string {
+	return string(postgresqlConfig)
+}
```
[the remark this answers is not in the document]
jdpleiness: Yeah, even though it's not much, shoving it in the `config` package gives a little more protection against mutability which is something I was going for with this. These aren't something we generally tell people to change in our configurations. Instead we would point them to use their own Postgres instance at that point.

The configurations for codeintel and codeinsights are each different and will likely go in here as well so I'll rename to reflect that.

internal/appliance/defaults.go
jdpleiness: I noticed this, but I don't know if there is a good reason it should be different than the others. 🤔

internal/appliance/defaults.go
jdpleiness: Ahh okay, I missed that but it makes sense! 💯

internal/appliance/spec.go
jdpleiness: Ah got it. I like that idea so let's go with it.

internal/appliance/testdata/sg/pgsql/default.yaml
jdpleiness: >Default resources are often differing by M vs Mi in this diff. Very minor note,
probably not a real problem, but may as well squash the diff.

I've been doing this intentionally just to standardize on the units we use. Most these limits/requests are guesses at best. If we find otherwise, I'll revert. 

>This user ID diff might stop the process from running successfully, not 100%
sure though. Depends on what's in the rootFS. May as well squash the diff?
Similar diffs for other containers although I haven't commented on these.

This was another attempt to standardize as well, but I'm a bit more leery of this for the reasons you mentioned and I'm not 100% sure myself.

internal/appliance/pgsql.go
jdpleiness: Ahh I missed that this was a map of ContainerConfigs, my bad 🤡

craigfurman: No worries! All new code, and changing fast

internal/appliance/defaults.go
craigfurman: For sure, consistency would be good, and I'd be on board with opening issues to fix that, but in the mean time, we need to direct prometheus to scrape the port that's listening. The current config will just cause scrape errors.

craigfurman (approved):
🚀 

FYI: I added `checksum/pgsql.secret` to the list in https://github.com/sourcegraph/sourcegraph/issues/62227

PR #62518, opened 2024-05-08
gitserver: Implement CommitLog using Commits

eseliger:
This PR reimplements what CommitLog does using Commits. This reduces the surface area, but also removes an API that currently wasn't respecting sub repo permissions properly.
Another side-effect of this is that we have to make more gRPC requests for indexing, but in return we accumulate less data in memory as we don't need to store all paths in all diffs here.

Closes https://github.com/sourcegraph/sourcegraph/issues/62103

Test plan:

Existing tests still pass, adjusted test suite as well.

PR #62514, opened 2024-05-08
gitserver: Move CommitsUniqueToBranch to callsite

eseliger:
This has only one caller and is now implemented using existing gitserver stuff, so I'm moving this custom response mapping to where it's used. One less API we need to expose via gRPC.

Closes https://github.com/sourcegraph/sourcegraph/issues/62102

Test plan:

Adjusted tests, and existing E2E tests still pass.

PR #62519, opened 2024-05-08
gitserver: Always use Commits in HasCommitAfter

eseliger:
There used to be two implementations of this, one for with subrepo enabled, one for without. This PR aligns those two to both use the Commits method.
One code path less, and one API less that we need to port to gRPC.

Test plan:

Existing unit tests are still passing.

### DIFF
--- a/internal/gitserver/commands.go
+++ b/internal/gitserver/commands.go
@@ -1064,46 +1064,20 @@ func (c *clientImplementor) HasCommitAfter(ctx context.Context, repo api.RepoNam
 	})
 	defer endObservation(1, observation.Args{})
 
-	if authz.SubRepoEnabled(c.subRepoPermsChecker) {
-		return c.hasCommitAfterWithFiltering(ctx, repo, date, revspec)
-	}
-
 	if revspec == "" {
 		revspec = "HEAD"
 	}
 
-	commitid, err := c.ResolveRevision(ctx, repo, revspec, ResolveRevisionOptions{EnsureRevision: false})
+	// TODO: Because N: 1 currently has a special meaning because of `isRequestForSingleCommit`,
+	// we ask for two commits here, but the second one we never actually need.
+	// One we figure out why `isRequestForSingleCommit` exists in the first place,
+	// we should update this.
+	commits, err := c.Commits(ctx, repo, CommitsOptions{N: 2, After: date, Range: revspec})
 	if err != nil {
 		return false, err
 	}
 
-	args, err := commitLogArgs([]string{"rev-list", "--count"}, CommitsOptions{
-		N:     1,
-		After: date,
-		Range: string(commitid),
-	})
-	if err != nil {
-		return false, err
-	}
-
-	cmd := c.gitCommand(repo, args...)
-	out, err := cmd.Output(ctx)
-	if err != nil {
-		return false, errors.WithMessage(err, fmt.Sprintf("git command %v failed (output: %q)", cmd.Args(), out))
-	}
-
-	out = bytes.TrimSpace(out)
-	n, err := strconv.Atoi(string(out))
-	return n > 0, err
-}
-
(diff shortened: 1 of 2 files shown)


eseliger:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62519?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#62519** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62519?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#62518** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62518?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#62514** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62514?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#62513** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62513?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#62107** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62107?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

internal/gitserver/commands.go
```diff
@@ -1129,46 +1129,20 @@ func (c *clientImplementor) HasCommitAfter(ctx context.Context, repo api.RepoNam
...
-
 	if revspec == "" {
 		revspec = "HEAD"
 	}
 
-	commitid, err := c.ResolveRevision(ctx, repo, revspec, ResolveRevisionOptions{EnsureRevision: false})
+	// TODO: Because N: 1 currently has a special meaning because of `isRequestForSingleCommit`,
+	// we ask for two commits here, but the second one we never actually need.
+	// One we figure out why `isRequestForSingleCommit` exists in the first place,
+	// we should update this.
```
camdencheek, line 1139: I'd guess this is because when `path` is set, it can be much more expensive to fetch two commits than one commit because we potentially have to iterate over the entire history of the repo to find another modification to that file. And when we're just interested in the most recent commit to a file (which is a common operation in clients), knowing whether more commits exist is useless and potentially expensive.

eseliger: My understanding is that this should only matter for sub-repo permissions 🤔 So if a commit isn't visible, we will try harder to find more commits that satisfy the range. But not exactly sure, so I put that off for later 😬 Previously, this implementation would get ALL commits, so this is still better than no `N` at all I guess 🤷

camdencheek: Aah, yeah, that would make sense. Missed the subrepo checker right above. I misunderstood this as part of the pagination API, but it really seems to just hydrate a full page if some were entries were filtered out in the client

internal/gitserver/commands.go
```diff
@@ -1129,46 +1129,20 @@ func (c *clientImplementor) HasCommitAfter(ctx context.Context, repo api.RepoNam
...
+	commits, err := c.Commits(ctx, repo, CommitsOptions{N: 2, After: date, Range: revspec})
```
camdencheek, line 1140: Given how minimal this function is now, should we just inline it and slim down the `gitserver.Client` interface?

camdencheek (approved):

internal/gitserver/commands.go
eseliger: yes 😬 https://github.com/sourcegraph/sourcegraph/pull/62520

camdencheek: hahaah, perfect 😄

ggilmore (approved):

eseliger:
### Merge activity

* **May 10, 7:24 PM EDT**: [Graphite](https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62519) rebased this pull request as part of a merge.

PR #62521, opened 2024-05-08
dev/linearhooks: add support for project modifier

michaellzc:
https://sourcegraph.slack.com/archives/C06GAS7FUGL/p1715062692858249

implements a `modifier` spec so we can override the project when moving the issue to another team

## Test plan

added tests and manual testing.

### DIFF
--- a/dev/linearhooks/config.example.yaml
+++ b/dev/linearhooks/config.example.yaml
@@ -13,3 +13,8 @@ spec:
           # teamId is the identifier of the destination team. Issues that match the rule will be moved to this team.
           # Either identifier (i.e. unique team prefix) or UUID is accepted
           teamId: BAR
+          # modifier is an optional field that allows for additional configuration when moving the issue.
+          # Any errors encountered while applying the modifier will be logged, but the issue will still be moved regardless.
+          modifier:
+            # (optional) projectName is the name of the project that the issue will be moved to in the destination team
+            projectName: 'Test Project 1'
--- a/dev/linearhooks/internal/handlers/handlers.go
+++ b/dev/linearhooks/internal/handlers/handlers.go
@@ -3,6 +3,7 @@ package handlers
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"time"
 
@@ -107,15 +108,35 @@ func (h *Handler) HandleIssueMover(w http.ResponseWriter, r *http.Request) {
 func moveIssue(logger log.Logger, client graphql.Client, rules []RuleSpec, issue linearschema.IssueData) error {
 	for _, r := range rules {
 		if newTeamId := r.identifyTeamToMoveTo(issue); newTeamId != nil {
-			logger.Info("moving issue",
-				log.String("from.id", issue.Team.Key),
-				log.String("to", r.Dst.TeamID),
+			logger := logger.With(
+				log.String("issue.id", issue.Identifier),
+				log.String("from.key", issue.Team.Key),
+				log.String("to.id", r.Dst.TeamID),
 			)
+			logger.Info("moving issue")
+
 			teamUUID, err := getTeamUUID(context.Background(), client, r.Dst.TeamID)
 			if err != nil {
 				return errors.Wrapf(err, "get dst team %q UUID", r.Dst.TeamID)
 			}
-			if err := moveIssueToTeam(context.Background(), client, issue.Identifier, teamUUID); err != nil {
+
+			input := moveIssueToTeamInput{
+				issueId:  issue.Identifier,
+				teamUUID: teamUUID,
+			}
+			if r.Dst.Modifier != nil && r.Dst.Modifier.ProjectName != "" {
+				logger := logger.With(log.String("to.projectName", r.Dst.Modifier.ProjectName))
+				projectUUID, err := getProjectUUID(context.Background(), client, teamUUID, r.Dst.Modifier.ProjectName)
+				if err != nil {
+					logger.Error("unable to resolve project UUID, issue is moved without setting the desired project. you should inspect the error and consider updating the rule.",
+						log.Error(err),
+						log.String("to.uuid", teamUUID),
+					)
+				} else {
+					input.projectId = projectUUID
+				}
+			}
+			if err := moveIssueToTeam(context.Background(), client, input); err != nil {
 				return errors.Wrapf(err, "move issue %q to team %q", issue.Identifier, teamUUID)
 			}
 			return nil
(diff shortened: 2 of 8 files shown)


michaellzc:
* **#62521** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62521?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @michaellzc and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

dev/linearhooks/internal/handlers/handlers.go
```diff
@@ -107,15 +108,37 @@ func (h *Handler) HandleIssueMover(w http.ResponseWriter, r *http.Request) {
...
 				return errors.Wrapf(err, "get dst team %q UUID", r.Dst.TeamID)
 			}
-			if err := moveIssueToTeam(context.Background(), client, issue.Identifier, teamUUID); err != nil {
+
+			input := moveIssueToTeamInput{
+				issueId: issue.Identifier,
+				teamId:  teamUUID,
+			}
+			if r.Dst.Modifier != nil {
+				if r.Dst.Modifier.ProjectName != "" {
```
varungandhi-src, line 128:
```suggestion
			if r.Dst.Modifier != nil && r.Dst.Modifier.ProjectName != "" {
```

dev/linearhooks/internal/handlers/handlers.go
```diff
@@ -141,8 +173,32 @@ func getTeamUUID(ctx context.Context, c graphql.Client, idOrKey string) (string,
 	return resp.Team.Id, nil
 }
 
-func moveIssueToTeam(ctx context.Context, c graphql.Client, issueId, teamId string) error {
-	_, err := lineargql.MoveIssueToTeam(ctx, c, issueId, teamId)
+func getProjectUUID(ctx context.Context, c graphql.Client, teamID, projectName string) (string, error) {
```
varungandhi-src, line 176:
```suggestion
func getProjectUUID(ctx context.Context, c graphql.Client, teamUUID, projectName string) (string, error) {
```

It's a bit confusing to use 'ID' here and 'UUID' at the caller.

dev/linearhooks/internal/handlers/handlers.go
```diff
@@ -107,15 +108,37 @@ func (h *Handler) HandleIssueMover(w http.ResponseWriter, r *http.Request) {
...
+
 			teamUUID, err := getTeamUUID(context.Background(), client, r.Dst.TeamID)
 			if err != nil {
 				return errors.Wrapf(err, "get dst team %q UUID", r.Dst.TeamID)
 			}
-			if err := moveIssueToTeam(context.Background(), client, issue.Identifier, teamUUID); err != nil {
+
+			input := moveIssueToTeamInput{
+				issueId: issue.Identifier,
+				teamId:  teamUUID,
```
varungandhi-src, line 125:
```suggestion
				teamUUID:  teamUUID,
```

It's a bit confusing to use 'id' and 'UUID' interchangeably.

dev/linearhooks/internal/lineargql/operations.graphql
```diff
@@ -8,10 +8,24 @@ query GetTeamById($id: String!) {
...
+query GetProjectsByTeamId($teamId: String!) {
+    team(id: $teamId) {
+        projects {
+            nodes {
+                name
+                id
+            }
+        }
+    }
+}
```
varungandhi-src, line 31:
```suggestion
query GetProjectsByTeamId($teamId: String!, $projectName: String!) {
    team(id: $teamId) {
        projects(filter: { name: { eq: $projectName } }) {
            nodes {
                name
                id
            }
        }
    }
}
```

We already have the project name at the caller; this query should be cheaper rather than filtering after-the-fact. Many common fields in the Linear GraphQL API support similar filtering options.

dev/linearhooks/internal/handlers/handlers_test.go
```diff
@@ -63,11 +64,107 @@ func Test_moveIssue(t *testing.T) {
...
+					"projectId": "",
+					"teamId":    "another-team-uuid",
 				},
 			}),
 		},
+		{
+			name: "match with project modifier",
+			rs: []RuleSpec{
+				{
+					Src: SrcSpec{TeamID: "team-a-uuid", Labels: []string{"a"}},
```
varungandhi-src, line 77: It seems a bit weird that `teamId` above is `another-team-uuid` and below it is `team-a-uuid`. Are these deliberately different/why is this working if they're different? 🤔

dev/linearhooks/config.example.yaml
```diff
@@ -13,3 +13,8 @@ spec:
           # teamId is the identifier of the destination team. Issues that match the rule will be moved to this team.
           # Either identifier (i.e. unique team prefix) or UUID is accepted
           teamId: BAR
+          # (optional) modifier is a collection of additional actions that can be performed on the issue
+          modifier:
+            # (optional) projectName is the name of the project that the issue will be moved to in the destination team
+            # If unable to find the project, the issue will be moved regardless
+            projectName: 'Test Project 1'
```
varungandhi-src, line 20: Not entirely sure what the benefit is of an extra level of optionality here with 'modifier' (which is also a bit awkward name-wise); for example, GitHub Actions YAML mixes optional and non-optional keys together at the same 'level' in the tree. But it's not a big deal.

Is the point that problems with fields outside of `modifier:` will result in "hard errors" whereas problems with fields inside of `modifier:` will result in "soft errors"? If so, it would be useful to document that property on `modifier:` rather than on `projectName:`.

varungandhi-src (approved):
Left some minor suggestions for simplification.

dev/linearhooks/internal/handlers/handlers_test.go
michaellzc: > It seems a bit weird that `teamId` above is `another-team-uuid` and below it is `team-a-uuid`. Are these deliberately different/why is this working if they're different? 🤔

😅 it's a way to workaround for the shared global cache. otherwise, it's hard to assert the behavior.

I plan on refactoring the graphql client out of its standalone pkg. at least, during test, we can start from a clean state everytime

dev/linearhooks/config.example.yaml
michaellzc: yeah, that's the intention. `modifer` is a collection of best-effort modification we will make when moving the issue

> Not entirely sure what the benefit is of an extra level of optionality here with 'modifier' 

😂 I think I am too used to the craziness of k8s API. That said, I agree, we shouldn't have more too much nesting

PR #62531, opened 2024-05-08
feat: Add new GraphQL API for getting occurrences

varungandhi-src:
Fixes GRAPH-126

## Test plan

TODO:
- [ ] Add tests


cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -47,6 +47,139 @@ extend type GitBlob {
...
+    filter == { provenance: { equals: Precise } }.
+
+    Commit matching is done based on graph order.
+    For merge commits or their children, it is possible that the
+    same blob may have code graph data at different ancestors,
+    in which case this API will return multiple results.
+
+    EXPERIMENTAL: This API may change in the future.
+    """
+    codeGraphData(filter: CodeGraphDataFilter): [CodeGraphData!]
```
keynmol, line 66: I think we should tighten the result here even further - 

```suggestion
    codeGraphData(filter: CodeGraphDataFilter): [CodeGraphData!]!
```

Just to indicate that we will never return null here, and it will always be an object. Unless there's a situation where `null` here is an acceptable response?

keynmol: The exact meaning of `[...!]!` is covered in https://graphql.org/learn/schema/#object-types-and-fields (see appearsIn attribute)

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -47,6 +47,139 @@ extend type GitBlob {
...
+}
+
+"""
+EXPERIMENTAL: This type may change in a backwards-incompatible way.
+"""
+input CodeGraphDataFilter {
+    """
+    If this field is not set, then the codeGraphData API
+    will go through each provenance each provenance one by one
```
keynmol, line 75:
```suggestion
    will go through each provenance one by one
```

I wonder if we should delegate this priority fallback process to the client instead, and make the filter required. This will simplify the implementation and can give better guarantees on execution time of a single API call.

varungandhi-src: Delegating this to the client will likely cause redundant work, as the plan is to implement syntactic indexing internally using text search for a first-level filtering process.

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
varungandhi-src: Making this non-null will affect error propagation; if there is an error getting the code graph data, then you won't be able to read other fields for `GitBlob`, which seems like a net negative from a client's POV.

varungandhi-src: We can also strengthen this to become non-optional later, that would be backwards-compatible.

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -47,6 +47,167 @@ extend type GitBlob {
...
+"""
+enum CodeGraphDataProvenance {
+    """
+    Based on a compiler, a type-checker or a similar data source
+    which doesn't have false positives.
+    Generally, the results are specific to a particular build configuration,
+    such as for a specific OS or CPU, which can matter for
+    codebases having a large amount of platform-specific code.
+    """
+    Precise
```
eseliger, line 162: See https://spec.graphql.org/June2018/#sec-Enum-Value, specifically

> It is recommended that Enum values be “all caps”.

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -47,6 +47,167 @@ extend type GitBlob {
...
+    provenance: CodeGraphDataProvenance!
+
+    """
+    The commit associated with this code graph data.
+
+    In general, this will be an ancestor of the commit at which code
+    graph data was requested, as code graph data may not be available
+    at the exact commit for the blob.
+    """
+    commit: String!
```
eseliger, line 108: would there be benefits of making this a `GitCommit` type?

eseliger (commented):
GQL schema LGTM :)

internal/codeintel/codenav/transport/graphql/root_resolver_code_graph.go
```diff
@@ -0,0 +1,154 @@
...
+	// Index inside occurrences array in document
+	Index int
+}
+
+func (s scipOccurrence) Symbol() (*string, error) {
+	return pointers.Ptr(s.impl.Symbol), nil
+}
+
+func (s scipOccurrence) Range() (resolverstubs.RangeResolver, error) {
+	// FIXME(issue: GRAPH-571): Below code is correct iff the indexer uses UTF-16 offsets
```
mmanela, line 148: How big of an issue is this?  What is the impact?

internal/codeintel/resolvers/codenav.go
```diff
@@ -140,3 +148,132 @@ type DiagnosticResolver interface {
...
+		out.Syntactic = true
+		out.Precise = true
+	} else {
+		p := *a.Filter.Provenance.Equals
+		switch p {
+		case ProvenancePrecise:
+			out.Precise = true
+		case ProvenanceSyntactic:
+			out.Syntactic = true
+		case ProvenanceSearchBased:
```
mmanela, line 213: Should we set `out.SearchBased = true` here?

mmanela: Or is this a TBD since we are not implementing that yet?

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -47,6 +47,167 @@ extend type GitBlob {
     Experimental: This API is likely to change in the future.
     """
     symbolInfo(line: Int!, character: Int!): SymbolInfo
+
+    """
+    Return the code graph data associated with this blob.
+
+    If there are multiple tools (i.e. name and version pairs) which
```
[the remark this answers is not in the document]
varungandhi-src: It's up to the client. SCIP can be used in an "overlay" fashion. For example, say you have C++ code being generated from Protobufs, then a protobuf indexer can link the generated code to the schema, and the C++ indexer can link the generated code to usages. In such a situation, the same generated file would be indexed by two different indexers.

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -47,6 +47,167 @@ extend type GitBlob {
...
+    """
+    Information about the tool which generated this code graph data
+    """
+    toolInfo: CodeGraphToolInfo
+
+    """
+    Occurrences are guaranteed to be sorted by range. It is possible
+    for there to be multiple occurrences for the same exact source range.
+    """
+    occurrences(first: Int, after: String): SCIPOccurrenceConnection
```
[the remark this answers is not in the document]
varungandhi-src: For other APIs, we don't document these as it's following a well-known GraphQL pattern (so I think it's fair to assume as "background knowledge"?). https://graphql.org/learn/pagination/

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
varungandhi-src: Doing this directly creates an import cycle because `type GitCommitResolver struct` that we'd need to use is in the `frontend/graphqlbackend` package, however, the code intel logic is organized mostly inside `internal/codeintel/resolvers`, and there's only a light layer of logic in `frontend/graphqlbackend` calling out to that.

It's also a bit weird that the `GitCommitResolver` takes in a full `database.DB` and not a `RepoCommitsChangelistsStore` or similar.

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
varungandhi-src: Fixed to use SCREAMING_CASE

internal/codeintel/codenav/transport/graphql/root_resolver_code_graph.go
varungandhi-src: I've added this to the issue:

> code navigation in files with multi-byte codepoints may not work correctly.

internal/codeintel/codenav/transport/graphql/root_resolver.go
```diff
@@ -111,6 +117,68 @@ func (r *rootResolver) GitBlobLSIFData(ctx context.Context, args *resolverstubs.
 	), nil
 }
 
+func (r *rootResolver) CodeGraphData(ctx context.Context, opts *resolverstubs.CodeGraphDataOpts) (_ *[]resolverstubs.CodeGraphDataResolver, err error) {
+	ctx, _, endObservation := r.operations.codeGraphData.WithErrors(ctx, &err, observation.Args{Attrs: opts.Attrs()})
+	endObservation.OnCancel(ctx, 1, observation.Args{})
+
+	makeResolvers := func(prov resolverstubs.CodeGraphDataProvenance) ([]resolverstubs.CodeGraphDataResolver, error) {
+		indexer := ""
+		if prov == resolverstubs.ProvenanceSyntactic {
```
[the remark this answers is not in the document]
varungandhi-src: Not specifying an indexer will return all precise uploads. https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/internal/codeintel/uploads/shared/types.go?L356-362

internal/codeintel/codenav/transport/graphql/root_resolver.go
```diff
@@ -111,6 +117,68 @@ func (r *rootResolver) GitBlobLSIFData(ctx context.Context, args *resolverstubs.
...
+
+	provs := opts.Args.ProvenancesForSCIPData()
+	if provs.Precise {
+		preciseResolvers, err := makeResolvers(resolverstubs.ProvenancePrecise)
+		if len(preciseResolvers) != 0 || err != nil {
+			return &preciseResolvers, err
+		}
+	}
+
+	if provs.Syntactic {
```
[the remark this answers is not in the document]
varungandhi-src: No, there are currently no plans for generating SCIP data operating on pure text. I've added a comment describing what else we could do here for syntactic code nav.

internal/codeintel/resolvers/codenav.go
varungandhi-src: There is no implementation plan for search-based here. Search-based code nav will continue to operate using source ranges.

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
varungandhi-src: Oh wait, my comment here was wrong.

> Delegating this to the client will likely cause redundant work, as the plan is to implement syntactic indexing internally using text search for a first-level filtering process.

This point will apply to the `usagesForSymbol` API, not here. However, given that both of them will end up being:

```
    provenance: CodeGraphDataProvenanceComparator
```

It would be weird if the filter for `usagesForSymbol` followed `Precise -> Syntactic -> SearchBased` and the filter for `codeGraphData` didn't have fallback.

cmd/frontend/graphqlbackend/codeintel.codenav.graphql
```diff
@@ -47,6 +47,170 @@ extend type GitBlob {
...
+EXPERIMENTAL: This type may change in a backwards-compatible way.
+"""
+enum SymbolRole {
+    DEFINITION
+    REFERENCE
+    """
+    Applicable for forward declarations in languages with header files (C, C++ etc.)
+    as well as standalone signatures in languages with separate interface files (OCaml etc.).
+    """
+    FORWARD_DEFINITION
```
keynmol, line 213: > Applicable for forward declarations
> FORWARD_DEFINITION

Should this be called FORWARD_DECLARATION? Or just DECLARATION may be?

If memory serves, in things like Java `interface`s they're called `declarations`

varungandhi-src: The terminology here matches SCIP. https://github.com/sourcegraph/scip/blob/main/scip.proto#L516-L521

I don't think it's worth diverging from that here.

internal/codeintel/codenav/transport/graphql/root_resolver.go
```diff
@@ -111,6 +117,77 @@ func (r *rootResolver) GitBlobLSIFData(ctx context.Context, args *resolverstubs.
...
+		syntacticResolvers, err := makeResolvers(resolverstubs.ProvenanceSyntactic)
+		if len(syntacticResolvers) != 0 || err != nil {
+			return &syntacticResolvers, err
+		}
+
+		// Enhancement idea: if a syntactic SCIP index is unavailable,
```
keynmol, line 160: I think this is a good idea – I remember us discussing having a two-pronged approach, with relying on batch indexing for infrequently changing files, and doing online recalculation for missing data. Should be fast enough for a single file

varungandhi-src: I think we should do it post MVP. If we start taking lockfiles and similar into account, then there will be 2 sub-flavors of syntactic (with and without versioning information), so we will need some fuzzy matching support to take advantage of it.

internal/codeintel/codenav/transport/graphql/root_resolver.go
```diff
@@ -111,6 +117,77 @@ func (r *rootResolver) GitBlobLSIFData(ctx context.Context, args *resolverstubs.
...
+		// analyze the file.
+	}
+
+	// We do not currently have any way of generating SCIP data
+	// during purely textual means.
+
+	return &[]resolverstubs.CodeGraphDataResolver{}, nil
+}
+
+func preferUploadsWithLongestRoots(uploads []shared.CompletedUpload) []shared.CompletedUpload {
```
keynmol, line 172: Can you please add a comment about the assumptions you make on ther order of `uploads`? As that's important information for why `sortedMap.Oldest()` is used.

internal/codeintel/codenav/transport/graphql/root_resolver.go
```diff
@@ -189,3 +266,60 @@ func (r *gitBlobLSIFDataResolver) VisibleIndexes(ctx context.Context) (_ *[]reso
...
+	svc CodeNavService,
+	upload shared.CompletedUpload,
+	opts *resolverstubs.CodeGraphDataOpts,
+	provenance resolverstubs.CodeGraphDataProvenance,
+	operations *operations,
+) resolverstubs.CodeGraphDataResolver {
+	return &codeGraphDataResolver{
+		sync.Once{},
+		/*document*/ nil,
+		/*documentRetrievalError*/ nil,
```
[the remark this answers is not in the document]
varungandhi-src: To make sure we get an error when adding new fields and forgetting to add something here. This is a constructor function, so it's extra important that updating it doesn't get missed.

internal/codeintel/codenav/transport/graphql/root_resolver.go
varungandhi-src: There is no pre-condition related to the ordering of uploads in the input slice.

Here, the primary reason for using a `orderedmap` over `map` is to avoid introducing non-determinism. `Oldest()` doesn't mean "the one that was inserted first", it means "smallest" based on key sort order.

varungandhi-src: I've added a comment here that the purpose is non-determinism.

internal/codeintel/codenav/transport/graphql/root_resolver.go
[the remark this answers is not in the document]
keynmol: Huh. I had no idea about this behaviour.

```go
// You can edit this code!
// Click here and start typing.
package main

import "fmt"

type Hello struct {
	Name string
	Age  int
	Nick string
}

func main() {
	fmt.Println(Hello{"asd", 25}) // FAILS
	fmt.Println(Hello{Name: "asd", Age: 25}) // OK
}
```

This is an unpleasant surprise 😬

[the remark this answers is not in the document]
varungandhi-src: Can I interest you in some blog posts?

https://typesanitizer.com/blog/go-experience-report.html
https://fasterthanli.me/articles/i-want-off-mr-golangs-wild-ride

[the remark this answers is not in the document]
keynmol: ohh I'm always down to read some fasterthanlime

keynmol (approved):
Phew, went through it.

Looks good!

PR #62536, opened 2024-05-08
Svelte: Styling updates to search filters

taiyab:
## Before
CleanShot 2024-05-08 at 12 27 36@2x

## After
CleanShot 2024-05-08 at 12 27 06@2x

## Test plan
Tested locally for visual changes.


client/web-sveltekit/src/lib/search/dynamicFilters/Section.svelte
```diff
@@ -87,6 +87,10 @@
         display: flex;
         flex-direction: column;
         gap: 0.5rem;
+
+        h4 {
+            margin-bottom: 0.25rem;
+        }
```
camdencheek, line 93: The space between the header and the list is already handled by the `article` flex box. We can just expand the gap on line 89

camdencheek (approved):

PR #62538, opened 2024-05-08
svelte: Fix file tree behavior

fkling:
While working on #62494 I noticed that the file tree didn't work as expected.
In particular, when being at the repo root and navigating to a file via the fuzzy finder the file tree would  switch to the parent directory, just like as if the file was directly navigated to.

However, what should happen is that the tree expands all directories to reveal the file, because the repo root is the top level directory and we should always be showing the most top level directory that was visited.

The issue is that we are not properly detecting the repo root level (I think we switched from using `'.'` to `''` at some point). To avoid this problem I'm now using a constant instead.

I also changed how the tree entries are scrolled into view. The current solution doesn't work for deeply nested entries because at the time the scroll logic is invoked the entry hasn't been loaded yet. By moving the logic to the node it can scroll itself into view when it's available.

The waterfall loading nature of the file tree also became much more apparent. We can avoid this by using making use of the `ancestors` field but that needs to be done in a separate PR because that entails larger changes.

## Test plan

Manual testing.

### DIFF
--- a/client/web-sveltekit/src/lib/TreeNode.svelte
+++ b/client/web-sveltekit/src/lib/TreeNode.svelte
@@ -13,6 +13,8 @@
     export let entry: T
     export let treeProvider: TreeProvider<T>
 
+    let label: HTMLElement | undefined
+
     $: treeState = getTreeContext()
     $: nodeID = treeProvider.getNodeID(entry)
     $: expandable = treeProvider.isExpandable(entry)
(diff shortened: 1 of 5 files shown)


fkling:
@vovakulikov That should fix the issue you discovered when expanding a directory via the chevron and subsequently selecting a file in it.

camdencheek (approved):
LGTM! +1 to using `ancestors`. That should be _way_ cheaper for opening a file from root.

Issue #60411, opened 2024-02-12
gitserver: Move GetObject to new gRPC pattern


eseliger:
Note: This already is a gRPC call, but we should make sure it's aligned with how the other ones are implemented.

PR #62561, opened 2024-05-08
gitserver: grpc: fixup and add tests for GetCommit RPC

ggilmore:
Part of https://github.com/sourcegraph/sourcegraph/issues/60411

This PR fixes up the GetObject gRPC server method by:
- Adding unit tests for it since there weren't any before
- Adding a check to fail early if the repository isn't cloned yet (like all of our other grpc methods do). 

## Test plan

New unit tests

### DIFF
--- a/cmd/gitserver/internal/server_grpc.go
+++ b/cmd/gitserver/internal/server_grpc.go
@@ -320,18 +320,45 @@ func (gs *grpcServer) Archive(req *proto.ArchiveRequest, ss proto.GitserverServi
 }
 
 func (gs *grpcServer) GetObject(ctx context.Context, req *proto.GetObjectRequest) (*proto.GetObjectResponse, error) {
+	accesslog.Record(ctx,
+		req.GetRepo(),
+		log.String("objectname", req.GetObjectName()),
+	)
+
+	if req.GetRepo() == "" {
+		return nil, status.New(codes.InvalidArgument, "repo must be specified").Err()
+	}
+
+	if req.GetObjectName() == "" {
+		return nil, status.New(codes.InvalidArgument, "object name must be specified").Err()
+	}
+
 	repoName := api.RepoName(req.GetRepo())
 	repoDir := gs.fs.RepoDir(repoName)
 
-	// Log which actor is accessing the repo.
-	accesslog.Record(ctx, string(repoName), log.String("objectname", req.GetObjectName()))
+	if err := gs.checkRepoExists(ctx, repoName); err != nil {
+		return nil, err
+	}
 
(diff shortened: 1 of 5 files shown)


internal/gitserver/v1/gitserver.proto
```diff
@@ -47,6 +47,14 @@ service GitserverService {
     option idempotency_level = NO_SIDE_EFFECTS;
   }
   rpc Exec(ExecRequest) returns (stream ExecResponse) {}
+
+  // GetObject returns the object with the given OID in the given repository.
+  //
+  // If the object is not found, an error with a RevisionNotFoundPayload is
```
pjlast, line 53: Should we distinguish between RevisionNotFound and ObjectNotFound?

cmd/gitserver/internal/server_grpc.go
```diff
@@ -320,18 +320,45 @@ func (gs *grpcServer) Archive(req *proto.ArchiveRequest, ss proto.GitserverServi
...
+	}
 
 	backend := gs.getBackendFunc(repoDir, repoName)
 
 	obj, err := backend.GetObject(ctx, req.GetObjectName())
 	if err != nil {
-		gs.svc.LogIfCorrupt(ctx, repoName, err)
 		gs.logger.Error("getting object", log.Error(err))
+
+		var e *gitdomain.RevisionNotFoundError
```
eseliger, line 349: can we also document on iface.go for the backend that this is one of the expected errors that are returned?

internal/gitserver/v1/gitserver.proto
eseliger: I would be fine with keeping it simple for now, until some caller needs to distinguish them, but no strong opinion.

eseliger (approved):

cmd/gitserver/internal/server_grpc.go
ggilmore: This is already in the current PR, unless you mean something else?

```go
	// GetObject allows to read a git object from the git object database.
	//
	// If the object specified by objectName does not exist, a RevisionNotFoundError is returned.
	GetObject(ctx context.Context, objectName string) (*gitdomain.GitObject, error)
```

eseliger: Oh I must have looked at the wrong branch, all good then!

ggilmore:
### Merge activity

* **May 10, 4:52 PM EDT**: @ggilmore merged this pull request with [Graphite](https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62561).

PR #62566, opened 2024-05-09
sg+release: push releases to the cloud ephemeral registry too

burmudar:
This allows internal and promoted releases to also publish images to the Cloud Ephemeral registry
## Test plan
CI + the push all script was previously tested to handle multiple registries


filiphaftek (approved):
Nice!

PR #62567, opened 2024-05-09
rockskip: ruler function is just bits.TrailingZeros

keegancsmith:
I couldn't help but make this change after seeing the algorithm in https://github.com/sourcegraph/sourcegraph/pull/62510. Feel free to just close this if this makes it less clear, but this kinda feels clearer in my head but I am probably weird.

Test Plan: TestRuler passes

### DIFF
--- a/internal/rockskip/server.go
+++ b/internal/rockskip/server.go
@@ -267,5 +267,8 @@ func ruler(n int) int {
 	if n <= 0 {
 		return 0
 	}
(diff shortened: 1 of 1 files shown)


jtibshirani (approved):
I like this, it's clear to me at least :) Could you add a short comment because it's not 100% obvious this is equivalent?

keegancsmith:
@jtibshirani updated with a comment

Issue #62572, opened 2024-05-09
Svelte: Search dropdown/popover not disappearing after search submission


taiyab:
Loom: https://www.loom.com/share/1c54d47218e54c0884a5e4b94552b4fc

PR #62574, opened 2024-05-09
[Svelte]: Hide suggestion UI after query was submitted

vovakulikov:
Closes https://github.com/sourcegraph/sourcegraph/issues/62572

Not the ideal solution but cheap one, I'm curious what @fkling 
thinks about having this as standard behavior for suggestion logic.

https://github.com/sourcegraph/sourcegraph/assets/18492575/52247f4b-3dd9-40ee-b448-7a10ec7cb3f4

## Test plan
- Go to the search result page and try to submit a query 
- You should see no suggestions after the submit action 
- Suggestions should appear right after you interact with the input

### DIFF
--- a/client/web-sveltekit/src/lib/search/input/SearchInput.svelte
+++ b/client/web-sveltekit/src/lib/search/input/SearchInput.svelte
@@ -198,8 +198,8 @@
         await invalidate(`search:${url}`)
         void goto(url)
 
-        // Rest interaction state since after success submit we should visually hide
-        // any suggestions UI but still keep focus on input, after user interacts with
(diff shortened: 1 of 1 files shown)


vovakulikov:
This search box interaction makes me think: Should we autofocus search input on the search result page as well as on the home page, or is this too much? 

cc @taiyab

taiyab:
> This search box interaction makes me think: Should we autofocus search input on the search result page as well as on the home page, or is this too much?
> 
> cc @taiyab

I think so, yes.

camdencheek (approved):

PR #62548, opened 2024-05-08
Gitserver: add `StatusTypeChanged` to supported statuses in `ChangedFiles`

camdencheek:
This adds support for `T` statuses (type changed) to `ChangedFiles`. This was previously hard-erroring, which was causing problems for code insights and hybrid search.

Apologies for the large diff. Part of this required adding a "type changed" member to the `StatusAMD` enum, which would be confusing because `AMD` stands for "added, modified, deleted". I renamed `StatusAMD` to `Status`, renamed the members to a pattern like `StatusAdded`, and added `StatusTypeChanged`. I checked every consumer of this enum to ensure that it safely handled new values. 

Because of the noise from the rename, I added comments to all the important bits of this PR. 

[Related Slack discussion](https://sourcegraph.slack.com/archives/C05EMJM2SLR/p1715176143546019)

## Test plan

Added test both to the git CLI backend and to hybrid search, which is a consumer of this API.

PR #62576, opened 2024-05-09
PR feedback followup

camdencheek:
I accidentally pushed my PR feedback commit to a different branch before merging https://github.com/sourcegraph/sourcegraph/pull/62548 🤦 This is just followup on a couple of pieces of feedback from that PR

## Test plan

N/A, no change in behavior, just style.


ggilmore (approved):
Thanks!

Issue #62001, opened 2024-04-18
Svelte: Implement file tree first draft rework


taiyab:
Somewhat visually revamped version of @fkling's existing file tree implementation with a few added bits:

- Repository root link anchored to top of the file tree
- Nesting line added to better visually grok what items belong in what directories
- Hover popovers added for files and directories, as stated here: https://github.com/sourcegraph/sourcegraph/issues/61930

<img width="858" alt="image" src="https://github.com/sourcegraph/sourcegraph/assets/5337876/eb0ccb9b-7f0d-493e-a358-2e6925d8a3ad">

A few things to note:
- The "Go to file" input/button should open up the finder to do file search across repo
- We should keep as much of the existing behaviour of the file tree that makes sense in the current Svelte implementation
- We can ignore the "New file pane" and "Keyboard shortcuts" options until we have something set for those

Figma: https://www.figma.com/file/EPgl0qVbqGSumvf52brhda/Code-Search?type=design&node-id=3071%3A244168&mode=design&t=WsYpFEoSGqcV24My-1

PR #62585, opened 2024-05-10
Svelte: First run on file tree visual update

vovakulikov:
Part of https://github.com/sourcegraph/sourcegraph/issues/62001

This PR mostly updates the visual representation of the file tree side panel. It also adds a few action buttons to the file tree header. 

| Expanded | Collapsed |
| ---------  | ---------- |
| Screenshot 2024-05-13 at 20 31 35 | Screenshot 2024-05-13 at 20 31 57 |

Loom Demo https://www.loom.com/share/0e72844210934a86868dd722e9c2b3e9

## Todo 
- [x] Implement a collapsed layout for the file tree panel
- [x] Add go to the root button
- [x] Add search file shortcut button
- [x] Update file tree spacing and nesting layout
- [x] Update file tree colors for selected states
- [x] Add move scope to the directory button to the file item    
- [x] Add vertical lines for the nesting levels (Instead I adjusted the paddings for nested levels since vertical lines were too noise on some deep nested levels)

## Test plan
- Manual testing of side file tree panel


camdencheek:
@vovakulikov I see you requested review, but this is still in draft state. What kind of review are ya looking for?

vovakulikov:
@camdencheek you can ignore it for now. For some reason I thought that since this is in draft mode you won't be notified

vovakulikov:
@taiyab I fixed the file icon color for selected states, I haven't added vertical lines (I tried and it looked a bit crowded with additional elements) but I think our paddings for nesting levers were off, I think I fixed them now, (basically I did repeat nesting from JetBrains IDEs) 

Screenshot 2024-05-13 at 15 20 17

taiyab (commented):
This looks great! I've added some comments on the Loom for feedback.

(The drag to collapse menu is genius).

vovakulikov:
@taiyab echo your comments here, just to keep it in one place

> How come all the controls/buttons get crowded onto one line? Flexbox problem?

This actually was intentional, because on the big screens, there will be enough space for these controls, we could either 
- just have it in two rows always  
- set a bit bigger minimal sizes for the rev and search files buttons to switch to one row with sizes you're comfortable with

> I'm happy to hide the "collapse" button (maybe keep the expand one), so by default, at least for now, the way to collapse the panel is to drag it down until it pops away.

I remember we had a big conversation that having non-jumping controls for show/hide actions is very vital. If @fkling is happy here I can live with not having this but I would prefer to still have it

vovakulikov:
On a bit of technical note (echo it from Slack to keep convos in one place) 

Before this PR, I thought that file tree providers knew about each other as we go from folder to folder, but it seems they don't, and the only way it works smoothly at the moment is because of our cache on the rural level. This doesn't cause any big problems at the moment, but I think if we need to have more interaction about which folder we want to open and when we probably should have something like a graph rather than a plain list of data sources. 

I've seen parents linking in the codebase (I think these arguments are not used and are obsolete at the moment), so @fkling probably has already thought about this, but I am curious what problem you had with this.

taiyab:
> just have it in two rows always

This please.

> that having non-jumping controls for show/hide actions is very vital

I'm proposing this as a temporary solution, as I was intending to actually have the button to the left of the file header so it always maintained its position.

After some experimenting, we can slightly adjust what you've done:
CleanShot 2024-05-14 at 05 39 58@2x

- Change the icon. I've added one for left panel collapse and one for left panel expand.
- Group the revision selector and home button into one container with 4px extra padding to the left + right of it (as shown in the screenshot sbove)

**Collapse left panel**
CleanShot 2024-05-14 at 05 41 26@2x

**Expand left panel**
CleanShot 2024-05-14 at 05 42 00@2x

Thanks!

taiyab:
As a side ask from me too, can we always highlight the panel dividers/separators like this when you hover them? 

CleanShot 2024-05-14 at 05 52 15@2x

- Change to `primary-01`
- Add a glow effect (probably same as input focus)
- Transition the hover effects over 250ms (fade in/out)

Basically, like Slack.

client/web-sveltekit/src/lib/TreeNode.svelte
```diff
@@ -110,19 +136,67 @@
...
+        position: relative;
         display: flex;
+        gap: 0.25rem;
         align-items: center;
+        padding-right: 0.25rem;
+        padding-left: calc(var(--tree-node-nested-level) * 1.25rem + var(--tree-node-left-padding));
+
+        // Change icon color based on selected item state
+        --icon-fill-color: var(--tree-node-expand-icon-color);
+        color: var(--tree-node-lable-color, var(--text-body));
```
fkling, line 161:
```suggestion
        color: var(--tree-node-label-color, var(--text-body));
```

client/web-sveltekit/src/lib/TreeNode.svelte
```diff
@@ -110,19 +136,67 @@
...
+
+        li[data-treeitem][aria-selected='true'] > & {
+            color: var(--tree-node-lable-color, var(--body-bg));
```
fkling, line 164:
```suggestion
            color: var(--tree-node-label-color, var(--body-bg));
```

client/web-sveltekit/src/lib/TreeNode.svelte
```diff
@@ -110,19 +136,67 @@
...
+        }
```
fkling, line 165: Not sure if that's better, but we could also set the color/variable in `[role="treeitem"]` (and `&[aria-selected="true"]`) and just use `color: var(...)` here, and avoid using this selector.
No preferences though.

client/web-sveltekit/src/lib/TreeNode.svelte
```diff
@@ -110,19 +136,67 @@
...
 
     .hidden {
         visibility: hidden;
+
+        // If we're in the flat list we can omit expand icon rendering
+        // since none of items is expandable, hence there aren't any offsets
+        :global([data-tree-view-flat-list='true']) & {
+            width: 0;
+            margin-left: 0.5rem;
+        }
```
fkling, line 200: I didn't quite understand why this is necessary until I opened the page and changed the styles dynamically.
I agree that not using as much horizontal space when it's not needed looks good, but I also noticed that the whole tree "shifts" when going up to a non-flat parent, which feels a bit weird. Overall I wonder if this edge case is worth having the extra rules.

client/web-sveltekit/src/lib/fuzzyfinder/FuzzyFinderContainer.svelte
```diff
@@ -1,14 +1,25 @@
+<script lang="ts" context="module">
+    import { writable } from 'svelte/store'
+
+    const fuzzyFinderState = writable(false)
+    const scopeState = writable<FuzzyFinderTabId | ''>('')
```
fkling, line 5: The variable names here are not very descriptive IMO. Maybe it makes sense to combine them into a single value:

```js
const fuzzyFinderState = writable<{open: boolean, selectedTab: FuzzyFinderTabId}>()
```

(arguably still not very descriptive, but now we are storing all state)

client/web-sveltekit/src/lib/fuzzyfinder/FuzzyFinderContainer.svelte
```diff
@@ -38,12 +49,18 @@
...
+            $scopeState = 'files'
             return false
         },
     })
 
+    scopeState.subscribe(tab => {
+        if (tab) {
+            finder?.selectTab(tab)
+        }
+    })
```
fkling, line 62: Likely not a problem since this component currently only gets destroyed when the page closes, but still: If you subscribe manually you should also unsubscribe:

```suggestion
    onDestroy(scopeState.subscribe(tab => {
        if (tab) {
            finder?.selectTab(tab)
        }
    }))
```

Or you could use a reactive statement:

```suggestion
    $: if (tab) {
        finder?.selectTab(tab)
    }
```

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte
```diff
@@ -17,24 +27,25 @@
...
     import { fetchSidebarFileTree } from '$lib/repo/api/tree'
-    import { sidebarOpen } from '$lib/repo/stores'
     import type { LastCommitFragment } from '$testing/graphql-type-mocks'
 
+    import Icon from '$lib/Icon.svelte'
     import Tabs from '$lib/Tabs.svelte'
     import TabPanel from '$lib/TabPanel.svelte'
     import LastCommit from '$lib/repo/LastCommit.svelte'
     import LoadingSpinner from '$lib/LoadingSpinner.svelte'
-    import SidebarToggleButton from '$lib/repo/SidebarToggleButton.svelte'
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.svelte:36]: I think you can delete `SidebarToggleButton` and all the related code (global store, callsite in file header).

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/fileTreeStore.ts
```diff
@@ -29,6 +29,13 @@ interface FileTreeStore extends Readable<FileTreeProvider | Error | null> {
      * Sets the current repo, revision, and path for the file tree.
      */
     set(args: { repoName: string; revision: string; path: string }): void
+
+    /**
+     * Resets file tree top level path, in some cases like jump to scope
+     * we have to invalidate cache since top level path has been changed
+     * to a lower level.
+     */
+    resetTopPathCache(repoName: string, revision: string): void
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/fileTreeStore.ts:38]: This feels very "manual". Maybe there is a better way to do this at some point, but I can't think of anything else atm either.

fkling (approved):
lgtm 👍🏻

client/web-sveltekit/src/lib/TreeNode.svelte
vovakulikov: > but I also noticed that the whole tree "shifts" when going up to a non-flat parent, which feels a bit weird

I have no preferences here, but I think the tree jumps regardless of the padding here; the current level gets additional paddings anyway as we open the parent level, no?

client/web-sveltekit/src/lib/TreeNode.svelte
fkling: I though "resetting" it to the default color by a treeitem descendant would override that setting but maybe not.

camdencheek (commented):
This PR seems to sometimes introduce a horizontally scrolling container, which is confusing because the content isn't actually horizontally scrollable except for a pixels in either direction

https://github.com/sourcegraph/sourcegraph/assets/12631702/d8b878f2-7a30-4579-b6f6-8e3fcf307fdf

camdencheek:
Not quite sure what's going on, but when I run this branch locally, the fuzzy finder activates on every page load. This does not happen for me on `main`. Anyone else?

client/web-sveltekit/src/lib/TreeNode.svelte
vovakulikov: No, CSS cascade doesn't work with recursive like this (if I understood your comment correctly), interesting thing that I wanted to have left padding for nesting also based on CSS variables but it turned out this expression doesn't work

```scss
:root {
   --nesting-level: 0;
 }
 
 .item-children {
     --nesting-level: calc(var(--nesting-level) + 1);
  }
  
```

vovakulikov:
@camdencheek, this is my bad. One commit slipped when I was fixing it by Felix's comments. I just pushed it, it should be working now.

vovakulikov:
> because the content isn't actually horizontally scrollable except for a pixels in either direction

I'll take a look; it's actually an interesting question: should this block be scrollable by a horizontal axis? It probably should (the proper scroll, not just a few pixels) if the tree has that nested structure like this. Otherwise it won't be possible to work with this tree only if they jump the scope with new control that this PR adds but this thing should be optional I guess

camdencheek (approved):

taiyab:
Excited for this to be merged in! Great work @vovakulikov.

vovakulikov:
I'm going to merge this one since this doesn't introduce any regressions in the current layout behavior but again I think the problem still remains

PR #62586, opened 2024-05-10
v2t: add v2 telemetry to the client/shared folder

dadlerj:
This is a tough one, given how much abstraction there is here...

V2 telemetry is built on a foundation of using known/safe strings only for event logging, while the /shared folder is almost exclusively filled with extensible abstractions like "action items", "commands", and "extensions" that can be created by any code intelligence client (web, browser extensions).

I've run into this concretely in two places so far, but both of these still need to be updated before I can mark this as ready for review and mergeable:

* First, in ActionItem.tsx, I am currently simply logging _every_ event as `blob.action` / `executed`, with the action ID as a metadata parameter (whereas v1 telemetry logged the action ID as the event _name_ itself. This really doesn't work for us—it means that things as critical as 'go to definition' and 'find references' are all logged as the abstract `blob.action` / `executed`... I need to figure out how to enforce the `KnownString` type on the action ID (or perhaps to add a new action property called eventFeature or similar that could have that constraint), which will require updates to all clients, unfortunately...

* Second, in commands.ts, I got lazy, and didn't log anything just marked it as TBD. I'm not really sure what an example of a command is yet (given all the abstraction here), but I assume enforcing the same `KnownString` requirements will be necessary here.
* 
## Test plan

1. `sg start`
2. visit page
3. check if events appear in event_logs table locally

### DIFF
--- a/client/client-api/src/contribution.ts
+++ b/client/client-api/src/contribution.ts
@@ -1,6 +1,5 @@
 import type { Primitive } from 'utility-types'
 
-import type { KnownString } from '@sourcegraph/telemetry'
 import type { Expression, TemplateExpression } from '@sourcegraph/template-parser'
 
 /**
@@ -132,7 +131,7 @@ export interface ActionContribution {
      * Properties to enable event telemetry to be recorded when an action is executed.
      */
     telemetryProps: {
-        feature: KnownString<string>
+        feature: string
 
         // No `action` prop is provided, because action items only log telemetry when executed (and thus use an
         // 'executed' action.
--- a/client/client-api/src/contribution.ts
+++ b/client/client-api/src/contribution.ts
@@ -132,7 +132,7 @@ export interface ActionContribution {
      * Properties to enable event telemetry to be recorded when an action is executed.
      */
     telemetryProps: {
-        feature: KnownString
+        feature: KnownString<string>
 
         // No `action` prop is provided, because action items only log telemetry when executed (and thus use an
         // 'executed' action.
--- a/client/client-api/src/contribution.ts
+++ b/client/client-api/src/contribution.ts
@@ -133,7 +133,11 @@ export interface ActionContribution {
      */
     telemetryProps: {
         feature: KnownString
-        action: KnownString
+
+        // No `action` prop is provided, because action items only log telemetry when executed (and thus use an
+        // 'executed' action.
+
+        privateMetadata?: { [key: string]: any }
     }
 }
 
--- a/client/client-api/src/contribution.ts
+++ b/client/client-api/src/contribution.ts
@@ -131,6 +131,12 @@ export interface ActionContribution {
      * Properties to enable event telemetry to be recorded when an action is executed.
      */
     telemetryProps: {
+        /**
+         * feature must be camelCase and '.'-delimited, e.g. 'myFeature.subFeature'.
+         *
+         * Most ActionContribution features should be prefixed with 'blob.' to indicate that they are actions
+         * that occur on text blobs.
+         */
         feature: string
 
         // No `action` prop is provided, because action items only log telemetry when executed (and thus use an
--- a/client/client-api/src/contribution.ts
+++ b/client/client-api/src/contribution.ts
@@ -1,5 +1,6 @@
 import type { Primitive } from 'utility-types'
 
+import type { KnownString } from '@sourcegraph/telemetry'
 import type { Expression, TemplateExpression } from '@sourcegraph/template-parser'
 
 /**
@@ -126,6 +127,14 @@ export interface ActionContribution {
      * (e.g., because the client is not graphical), then the client may hide the item from the toolbar.
      */
     actionItem?: ActionItem
+
+    /**
+     * Properties to enable event telemetry to be recorded when an action is executed.
+     */
+    telemetryProps: {
+        feature: KnownString
+        action: KnownString
+    }
 }
 
 /**
--- a/client/shared/src/actions/ActionItem.tsx
+++ b/client/shared/src/actions/ActionItem.tsx
@@ -342,9 +342,7 @@ export class ActionItem extends React.PureComponent<ActionItemProps, State, type
         )
     }
 
-    public runAction = <Feature extends string>(
-        event: React.MouseEvent<HTMLElement> | React.KeyboardEvent<HTMLElement>
-    ): void => {
+    public runAction = (event: React.MouseEvent<HTMLElement> | React.KeyboardEvent<HTMLElement>): void => {
         const action = (isAltEvent(event) && this.props.altAction) || this.props.action
 
         if (!action.command) {
--- a/client/shared/src/actions/ActionItem.tsx
+++ b/client/shared/src/actions/ActionItem.tsx
@@ -362,7 +362,16 @@ export class ActionItem extends React.PureComponent<ActionItemProps, State, type
         this.props.telemetryService.log(action.id)
         if (action.telemetryProps) {
             this.props.telemetryRecorder.recordEvent(
-                action.telemetryProps.feature as KnownString<Feature>,
+                // 👷 HACK: We have no control over what gets sent over Comlink/
+                // web workers, so we depend on action contribution implementations
+                // to give type guidance to ensure that we don't accidentally share
+                // arbitrary, potentially sensitive string values. In this
+                // RPC handler, when passing the provided event to the
+                // TelemetryRecorder implementation, we forcibly cast all
+                // the inputs below (feature) into known types
+                // (the string 'feature') so that the recorder will accept
+                // it. DO NOT do this elsewhere!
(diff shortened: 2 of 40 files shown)


vovakulikov:
@dadlerj 

>This really doesn't work for us—it means that things as critical as 'go to definition' and 'find references' are all logged as the abstract blob.action / executed

I haven't checked the existing dashboards with these events, but why wouldn't it work for us if we had an event with the same Feature name but differences in action type in eventProperties as dynamic arguments? I think it still would be possible to make a difference in amplitude/looker dashboards. I thought that part of the reason why we migrated to the new telemetry mechanism was partially about having strict/fixed event/feature names to minimize the surface of telemetry names.

Same suggestions about `commands.ts`, let me know if it wouldn't work I recently did the same thing for svelte prototype on old telemetry rails (by same I mean I used the same events from react app but made difference for svelte by enforcing svelte parameter in even properties, in amplitude at lease there is a way to parse JSON event bodies and select one property and filter datasets by this)

bobheadxi:
> why wouldn't it work for us if we had an event with the same Feature name but differences in action type in eventProperties as dynamic arguments?

Safe metadata only accepts numeric fields and known keys, i.e. a similar problem will arise where we still need to have a known set of all the possible values here (either translated to int enum, or for inclusion in keys), which is by design (we shouldn't be able to record things we don't know about)

dadlerj:
I was able to sort of "solve" this in [these two commits](https://github.com/sourcegraph/sourcegraph/compare/6c46579ba3a3094354f52d833eccd85c88ca0d9f...570f7cd81b80ef87f104585205134f140f6b79ee) in this PR by requiring all features be `KnownString` types. 

(Thanks @bobheadxi for giving me the tip of upgrading the version of the `@sourcegraph/telemetry` library to make this work!)

bobheadxi:
@dadlerj I'm 80% sure that won't work - locally I see:

image

I'm going to poke around this this afternoon as well and see if I can come up with something for ya

dadlerj:
Hah, interesting, because I don't get that locally (and I DO get successful events logged). I wonder if the
image

I haven't dug into CI issues yet though and I wouldn't be surprised if my local configuration isn't enforcing the rules correctly due to weird package versioning issues that I'm drowning in.

bobheadxi:
This problem will only fail the typechecker, but by nature of TS you don't _need_ to fulfill the typechecker for things to run :)

bobheadxi:
I think I understand the problem a bit better now - https://github.com/GoogleChromeLabs/comlink has a type `Remote` that turns an interface into an "over the wire" interface, i.e. for communication between web workers, making changes like:

```diff
- interface { foo(): bar }
+ interface { foo(): Promise<Remote<bar>> }
```

In this process it strips out most typechecking except the most basic primitives. For example, if you remove `Remote` wrapping:

```diff
- extHostAPI: Promise<Remote<FlatExtensionHostAPI>>
+ extHostAPI: Promise<FlatExtensionHostAPI>
```

You get tons of compilation errors about `Expression` etc types where callsites omit the typically required adapters to turn a string into `Expression`. This kind of makes sense because over the wire it is impossible to enforce types, but makes for an unfortunate situation here.

In Cody agent protocol, we receive over events over the wire as well, and there we forcibly cast types into `KnowString<'known'>`, i.e. we pretend the string is a const known string. We can do that here as well, **but** we will completely lose the typechecking; in Cody agent, we still retain the typechecking at the callsite, but we can't do that here because of the use of `Remote<>`.

dadlerj:
@bobheadxi can you check the commit I just pushed? Obviously lots of issues, but curious if this is roughly the _structure_ of what you were proposing?

client/shared/src/actions/ActionItem.tsx
```diff
@@ -341,7 +343,9 @@ export class ActionItem extends React.PureComponent<ActionItemProps, State, type
         )
     }
 
-    public runAction = (event: React.MouseEvent<HTMLElement> | React.KeyboardEvent<HTMLElement>): void => {
+    public runAction = <Feature extends string>(
+        event: React.MouseEvent<HTMLElement> | React.KeyboardEvent<HTMLElement>
+    ): void => {
```
bobheadxi, line 348: I think we can remove `<Feature extends string>` from the signature, because it isn't really used anyway (see my other comment)

client/shared/src/actions/ActionItem.tsx
```diff
@@ -356,6 +360,19 @@ export class ActionItem extends React.PureComponent<ActionItemProps, State, type
 
         // Record action ID (but not args, which might leak sensitive data).
         this.props.telemetryService.log(action.id)
+        if (action.telemetryProps) {
+            this.props.telemetryRecorder.recordEvent(
+                action.telemetryProps.feature as KnownString<Feature>,
```
bobheadxi, line 365: Here is the example I'm thinking of: https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/agent/src/agent.ts?L684-695; this works because `'feature'` fulfills the type requirements of `KnownString` (sorry that my other comments on this were a bit confusing)

```suggestion
                action.telemetryProps.feature as 'feature',
```

Note the docstring I put above in particular that explains the usage:

```ts
// 👷 HACK: We have no control over what gets sent over JSON RPC,
                // so we depend on client implementations to give type guidance
                // to ensure that we don't accidentally share arbitrary,
                // potentially sensitive string values. In this RPC handler,
                // when passing the provided event to the TelemetryRecorder
                // implementation, we forcibly cast all the inputs below
                // (feature, action, parameters) into known types (strings
                // 'feature', 'action', 'key') so that the recorder will accept
                // it. DO NOT do this elsewhere!
```

I think we can adapt that docstring to this callsite as well

dadlerj: No need to reply to this, the words being used here are hard to parse. I'll work on a next draft and get your take on that.

dadlerj:
@vovakulikov @bobheadxi this PR is finally ready for review. I have it working locally and it passes CI! Let me know if you have any questions or comments, I'd love to get this in (and thus getting the pre-Svelte web codebase to ~100% v2 telemetry coverage) as soon as I can.

client/client-api/src/contribution.ts
```diff
@@ -126,6 +126,18 @@ export interface ActionContribution {
      * (e.g., because the client is not graphical), then the client may hide the item from the toolbar.
      */
     actionItem?: ActionItem
+
+    /**
+     * Properties to enable event telemetry to be recorded when an action is executed.
+     */
+    telemetryProps: {
+        feature: string
```
bobheadxi, line 134: We should document the expected format here (maybe copied over from the telemetry recorder docstrings)

client/shared/src/api/contract.ts
```diff
@@ -175,9 +176,16 @@ export interface MainThreadAPI {
 
     /**
      * Log an event (by sending it to the server).
+     *
+     * @deprecated use recordEvent instead
```
bobheadxi, line 180:
```suggestion
     * @deprecated use getTelemetryRecorder().recordEvent instead
```

client/shared/src/codeintel/legacy-extensions/telemetry.ts
```diff
@@ -33,31 +44,38 @@ export class TelemetryEmitter {
...
         try {
-            sourcegraph.logTelemetryEvent(`codeintel.${action}`, {
+            sourcegraph.logTelemetryEvent(`codeintel.${action + xrepo ? '.xrepo' : ''}`, {
                 ...args,
                 durationMs: this.elapsed(),
                 languageId: this.languageID,
                 repositoryId: this.repoID,
             })
+            const telemetryRecorder = sourcegraph.getTelemetryRecorder()
+            telemetryRecorder?.recordEvent(`blob.codeintel${xrepo ? '.xrepo' : ''}`, action, {
```
bobheadxi, line 73: Very elegant, nice!

bobheadxi (approved):
I'm not 100% an expert here, but if CI type checks pass, everything here overall LGTM! Unfortunate that there is so much stuff under `/legacy-extensions` that still need to be maintained, but thank you for powering through these 💪

client/shared/src/actions/ActionItem.tsx
```diff
@@ -356,6 +357,28 @@ export class ActionItem extends React.PureComponent<ActionItemProps, State, type
 
         // Record action ID (but not args, which might leak sensitive data).
         this.props.telemetryService.log(action.id)
+        if (action.telemetryProps) {
+            this.props.telemetryRecorder.recordEvent(
+                // 👷 HACK: We have no control over what gets sent over Comlink/
```
vovakulikov, line 362: Is this a Comlink issue though? It looks like that in this place we don't run anything in web-worker, no? 
ActionItem is a react component and it's used under WebHoverOverlay, or are you referring that we get 
actions through extensions API which is web-workers

vovakulikov (approved):

PR #62589, opened 2024-05-10
Adjust SCIM schema to be compliant

### LINKED ISSUE #62185: SCIM: emails has a display value which is incompatible with Entra ID
- **Sourcegraph version:** v5.3 (potentially older)
- **Platform information:** SCIM setup with Entra ID

#### Steps to reproduce:

1. Set up SCIM with Entra ID following instructions in the [docs](https://sourcegraph.com/docs/admin/scim#how-to-use)
2a. Click "Test Schema" 
2b. Begin provisioning

#### Expected behavior:
1. Clicking "Test Schema" will result in a valid configuration. image
2. Provisioning will create, update, and delete users as expected.

#### Actual behavior:
1. When clicking "Test Schema", the following error is displayed: `SCIM Validation failed. The attribute emails[type eq "work"].display for User is not supported by the SCIM protocol. Please refer to the SCIM RFC`
image (11)
**Note:** According to [SCIM schema](https://scim.cloud/), the `emails` schema can be found below, whereas Sourcegraph includes an additional sub-attribute [(code here)](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@1ff9bc70eb7f47acb494b31b78302a1d1ea342e0/-/blob/cmd/frontend/internal/auth/scim/user_schema.go?L65)`display` under email.
```
"emails":[
    {
      "value":"dschrute@example.com",
      "type":"work",
      "primary": true
    }
  ]
```
2. When provisioning users, the following errors arise for all user updates in the Entra ID provisioning logs. *Note:* user creation and deletion seems to be working as expected.
```
"246ff006-4f12-4fe3-96fc-f9f6223932b1","2024-04-17T11:06:38Z","EntryExportUpdate","export","failure","Failed to update User 'redacted@email.com' in customappsso"
```
```
"2024-04-17T11:06:20Z","2a35d8fd-574d-48e3-927c-8c398e225a01","scim.2a35d8fd574d48e3927c8c398e225a01.d3fe4109-8727-4d5b-bbce-880663f1d87a","246ff006-4f12-4fe3-96fc-f9f6223932b1","f7d15e8f-21c6-4d5f-8de5-d2f86c17e4d6","Update","10019","4c26daf3-d932-4604-b7fb-1c2320cadb52","SourceGraph","","Azure AD Provisioning Service","system","failure","SystemForCrossDomainIdentityManagementServiceIncompatible","StatusCode: BadRequest
Message: Processing of the HTTP request resulted in an exception. Please see the HTTP response returned by the 'Response' property of this exception for details.
Web Response: 
{""schemas"":[""urn:ietf:params:scim:api:messages:2.0:Error""],""scimType"":""invalidPath"",""detail"":""The \""path\"" attribute was invalid or malformed."",""status"":""400""}. This operation was retried 0 times. It will be retried again after this date: 2024-04-17T11:06:16.7694551Z UTC","","nonServiceFailure","","39397638-a5c3-4a4c-b3ea-6bf779ca2292","Microsoft Entra ID","eed31091-3a64-41a7-a6b0-34f58b75bf8e","customappsso","9e01e4f7-e451-4556-8faf-83562aaa2e20","Redacted_Name","User","1148","","urn:ietf:params:scim:schemas:extension:enterprise:2.0:User"
```
3. Provisioning is quarantined.
image (6) (1)

pjlast:
Closes #62185 

Fixes our SCIM Schema discovery to be compliant with the RFC, as tested with https://scimvalidator.microsoft.com/

## Test plan

Tested with https://scimvalidator.microsoft.com/

### DIFF
--- a/cmd/frontend/internal/auth/scim/user_schema.go
+++ b/cmd/frontend/internal/auth/scim/user_schema.go
@@ -57,22 +57,19 @@ func (u *UserSCIMService) Schema() schema.Schema {
 				Description: optional.NewString("Email addresses for the user. The value SHOULD be canonicalized by the service provider, e.g., 'bjensen@example.com' instead of 'bjensen@EXAMPLE.COM'. Canonical type values of 'work', 'home', and 'other'."),
 				MultiValued: true,
 				Name:        "emails",
+				Required:    true,
 				SubAttributes: []schema.SimpleParams{
 					schema.SimpleStringParams(schema.StringParams{
-						Description: optional.NewString("Email addresses for the user. The value SHOULD be canonicalized by the service provider, e.g., 'bjensen@example.com' instead of 'bjensen@EXAMPLE.COM'. Canonical type values of 'work', 'home', and 'other'."),
+						Description: optional.NewString("Email addresses for the user. The value SHOULD be canonicalized by the service provider, e.g., 'bjensen@example.com' instead of 'bjensen@EXAMPLE.COM'. Canonical type value is 'work'."),
 						Name:        "value",
 					}),
 					schema.SimpleStringParams(schema.StringParams{
-						Description: optional.NewString("A human-readable name, primarily used for display purposes. READ-ONLY."),
-						Name:        "display",
-					}),
-					schema.SimpleStringParams(schema.StringParams{
-						CanonicalValues: []string{"work", "home", "other"},
-						Description:     optional.NewString("A label indicating the attribute's function, e.g., 'work' or 'home'."),
+						CanonicalValues: []string{"work"},
+						Description:     optional.NewString("A label indicating the attribute's function."),
 						Name:            "type",
 					}),
 					schema.SimpleBooleanParams(schema.BooleanParams{
-						Description: optional.NewString("A Boolean value indicating the 'primary' or preferred attribute value for this attribute, e.g., the preferred mailing address or primary email address. The primary attribute value 'true' MUST appear no more than once."),
+						Description: optional.NewString("A Boolean value indicating the 'primary' or preferred attribute value for this attribute, e.g., the preferred mailing address or primary email address. The primary attribute value 'true' MUST NOT appear more than once."),
 						Name:        "primary",
 					}),
 				},
--- a/cmd/frontend/internal/licensing/init/init.go
+++ b/cmd/frontend/internal/licensing/init/init.go
@@ -29,7 +29,7 @@ func Init(
 ) error {
 	// Enforce the license's max user count by preventing the creation of new users when the max is
 	// reached.
-	database.BeforeCreateUser = nil // enforcement.NewBeforeCreateUserHook()
+	database.BeforeCreateUser = enforcement.NewBeforeCreateUserHook()
 
 	// Enforce non-site admin roles in Free tier.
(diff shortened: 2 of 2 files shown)


pjlast:
@vdavid I'm tagging you since you've worked on this before, just in case you have some insight.

vdavid (approved):
Approving to unblock because I trust @pjlast that he tested this properly with Microsoft's validator, but I think the PR doesn't do a good job at describing how it fixes https://github.com/sourcegraph/sourcegraph/issues/62185.
I suggest adding the description of the fix in the PR description before merging the fix, and updating the comment [here](https://github.com/sourcegraph/sourcegraph/pull/62589/files#diff-cd80da0bea66ec2ade8091318025864c966c8f2b59ca781a039bfd36fca7705fR30) to describe the commented-out code.
(Consider this: Imagine we merge the PR without change, then someone sees the code in six months when debugging some other issue; gets curious why the right side of `init.go:32` is commented out; they check the git blame; find the PR, but still don't find any context; they check the issue, so they understand what _triggered_ the change, but not how it provides the fix.)

pjlast:
@vdavid that `init.go` commenting was for testing 😅 sorry, should have removed it before marking the PR as ready for review.

I hear you on the description etc. I'll flesh it out, and check if we can perhaps validate the schema through tests and not only using an external tool so that it's more future proof. Thought I'd poke you first in case I invest in changing something that was perhaps done for a specific reason.

vdavid:
> @vdavid that init.go commenting was for testing 😅 sorry, should have removed it before marking the PR as ready for review.

Oh, is the `Required: true` part the fix, then?

> check if we can perhaps validate the schema through tests and not only using an external tool so that it's more future proof
Would definitely be nice, but implementing the in-house validation didn't seem like a worthwhile investment to me at the time, and still doesn't, TBH, as there seem to be quite a few aspects to check for. Based on our experience of the SCIM server generally working pretty reliably, I'd probably rather invest in documenting how to use the external tool and running the external test suite as part of the release process.

pjlast:
It seems like the validator tool thinks only "work" emails can be "primary", because all of the examples in the SCIM RFC has the "work" email as the primary email. But nowhere in the spec does it actually require that.

vdavid:
Might be useful info: it's a known problem that Microsoft's SCIM implementation is incorrect, and that's why we added the `scim.identityProvider` setting to the site config. We only use its [accessor](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/cmd/frontend/internal/auth/scim/init.go?L25) to adjust our multi-replace behavior ([here](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/cmd/frontend/internal/auth/scim/resource_patch.go?L365)) because Microsoft implemented it wrong (=not according to the standard), but apparently, we might need to rely on this setting for this "emails" stuff as well?

pjlast:
@vdavid ah fantastic! That's some good insight. I knew tagging you was the right thing to do 🥳 yeah perhaps using that setting to also change some behaviours regarding emails is the way to go here.

pjlast:
Closing this PR as I don't think satisfying a non-compliant tool is the answer here. Might look into treating the Microsoft validator as a special case in the code.

PR #62597, opened 2024-05-10
Batch changes: make window ranges exclusive to avoid infinite loop

camdencheek:
This fixes an issue where we can potentially make no progress in the main loop of `Estimate`. Commentary inline.

Related [Slack thread](https://sourcegraph.slack.com/archives/C05EA9KQUTA/p1713644445735079)

## Test plan

Added unit test for this case and fuzz test to search for other edge cases.

### DIFF
--- a/internal/batches/types/scheduler/window/config.go
+++ b/internal/batches/types/scheduler/window/config.go
@@ -38,7 +38,7 @@ func (cfg *Configuration) Estimate(now time.Time, n int) *time.Time {
 	// would include the given entry. If we hit a week in the future, we'll
 	// bail, because a lot can happen in a week.
 	rem := n
-	at := now
+	at := now.UTC() // we _should_ always be using UTC, but don't trust the caller
 	until := at.Add(7 * 24 * time.Hour)
 	for at.Before(until) {
 		schedule := cfg.scheduleAt(at)
@@ -54,10 +54,12 @@ func (cfg *Configuration) Estimate(now time.Time, n int) *time.Time {
 			nextAt := schedule.ValidUntil()
 			if nextAt == at {
 				// This loop has a complex end condition, and we've hit infinite loops here before.
-				// As a protective measure, if we make no progress, do not continue looping.
+				// As a protective measure, if we make no progress, do not continue looping
+				// and just return an unknown estimate.
 				log.Scoped("changeset scheduler").Error("infinite loop calculating estimate")
 				return nil
 			}
+			at = nextAt
 			continue
 		}
 
--- a/internal/batches/types/scheduler/window/config.go
+++ b/internal/batches/types/scheduler/window/config.go
@@ -4,6 +4,7 @@ import (
 	"math"
 	"time"
 
+	"github.com/sourcegraph/log"
 	"github.com/sourcegraph/sourcegraph/lib/errors"
 	"github.com/sourcegraph/sourcegraph/schema"
 )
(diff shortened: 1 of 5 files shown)


internal/batches/types/scheduler/window/config.go
```diff
@@ -50,7 +51,13 @@ func (cfg *Configuration) Estimate(now time.Time, n int) *time.Time {
...
 		total := schedule.total()
 		if total == 0 {
-			at = schedule.ValidUntil()
+			nextAt := schedule.ValidUntil()
+			if nextAt == at {
+				// This loop has a complex end condition, and we've hit infinite loops here before.
+				// As a protective measure, if we make no progress, do not continue looping.
+				log.Scoped("changeset scheduler").Error("infinite loop calculating estimate")
+				return nil
+			}
```
peterguy, line 61: Should `at` updated to `nextAt` if they are not equal?

camdencheek: Oh, I just pushed up some local changes which included that. Did you catch an older commit?

camdencheek: (see next line)

peterguy: oh, probably; I'll refresh

internal/batches/types/scheduler/window/config.go
```diff
@@ -50,7 +51,15 @@ func (cfg *Configuration) Estimate(now time.Time, n int) *time.Time {
...
 		if total == 0 {
-			at = schedule.ValidUntil()
+			nextAt := schedule.ValidUntil()
+			if nextAt == at {
+				// This loop has a complex end condition, and we've hit infinite loops here before.
+				// As a protective measure, if we make no progress, do not continue looping
+				// and just return an unknown estimate.
+				log.Scoped("changeset scheduler").Error("infinite loop calculating estimate")
+				return nil
+			}
```
camdencheek, line 61: Struggling to put this in words that make sense, but the gist of what happened is:
- `schedule.ValidUntil()` returns its end time
- that end time can match (exactly) the end time of the active window
- whether a time is considered "in" a window is based on _inclusive_ start and end times
- we detect that the end time of the window is "in" the window, even though there is zero time left in the window
- since there is zero time left, we make no progress that iteration (`schedule.total() == 0`)
- but since `schedule.ValidUntil()` returns the same end time, we do not make it past the window that we are at the end at. 

To solve this, I:
1) made the end time of a window exclusive so we can not end up in a situation where we make zero progress and also do not progress the interval
2) because this is a complex loop condition, I also added a failsafe which returns early if we make no progress

internal/batches/types/scheduler/window/config_test.go
```diff
@@ -154,6 +154,56 @@ func TestConfiguration_Estimate(t *testing.T) {
...
+		})
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		now := time.Date(2024, 3, 10, 15, 35, 0, 0, time.UTC)
+		estimate := cfg.Estimate(now, 1000)
+		if estimate == nil {
+			t.Fatal("expected non-nil estimate")
+		}
```
camdencheek, line 180: This previously looped forever. Now it does not loop forever, and the non-nil check validates that we are not hitting the failsafe

internal/batches/types/scheduler/window/config_test.go
```diff
@@ -154,6 +154,56 @@ func TestConfiguration_Estimate(t *testing.T) {
...
+	})
+
+}
+
+func FuzzEstimate(f *testing.F) {
```
camdencheek, line 185: Becuase I was still uncomfortable with this loop, I added a fuzz test which I ran for ~10 minutes. It found no issues, but was able to surface the issues this PR fixes.

peterguy (approved):
Ran some of my own tests using other time values. Ensured that each piece of the solution does what it's supposed to do. Looks good. It's a pretty complex algorithm, so there will probably be other issues with it.

BolajiOlajide (commented):
Nice. Thank you @camdencheek

PR #62636, opened 2024-05-13
sg/msp: upgrade sentry

jac:
Fixes issue in Sentry with conditions array not accepting an empty array even though it is an optional field in Sentry
https://github.com/sourcegraph/managed-services-platform-cdktf/pull/8

Also changes Sentry to send to slack an alert at most once every 60 mins instead of 15.
Sentry alerts were a bit spammy so this should reduce that somewhat
## Test plan

tested in msp-testbed-test


bobheadxi (approved):
thank you!

PR #62642, opened 2024-05-13
URL Path Escape Model Name

RXminuS:
URL Model names weren't path-escaped leading to issues with Bedrock provisioned throughput model names that include `:` and `/` characters. 

## Test plan

- Existing unit tests should pass


eseliger (approved):
Good find!

Issue #61745, opened 2024-04-09
Context: Improve how we assemble final context for LLM


jtibshirani:
When assembling context, it’s not enough to just return the right file, the snippet within it needs to be helpful and accurate. We should evaluate + improve how we assemble context for the LLM.

An evaluation idea: we can ask an LLM to rerank the files given the snippets. If its top file matches the correct one, we know that choice of snippet is good enough.

jtibshirani:
Initial eval results on CodeSearchNet:

```
Recall (files)	91/99
Recall (chunks)	70/99
Average chunk overlap	0.81
```

Although we returned correct files for 91 queries, we only found relevant overlapping chunks for 70 of them. So definitely some room for improvement here. Out of these correct chunks, we tended to overlap quite a bit with the correct range (~80%).

jtibshirani:
Experiment: I tried ordering the chunks by BM25 instead of the current ranking, which boosts matches on symbol definitions. It performed significantly worse on CodeSearchNet:

```
Recall (files)	91/99
Recall (chunks)	40/99
Average chunk overlap	0.49
```

The data consists of general programming questions like "set working directory" and "encrypt aes ctr mode". The annotators often considered symbol definitions (and their doc strings) to be the best answer to these questions.

So there's not good evidence for changing to BM25 chunk ranking. In the future, maybe we could explore a symbol match + BM25 hybrid, where you get a boost both for matching a symbol definition, and also for matching more query terms.

PR #62643, opened 2024-05-13
Context: more precise chunk sizing

jtibshirani:
Currently, when retrieving context chunks, we hardcode the number of lines to
20. Historically, we've limited chunks to 1024 characters, and we chose 20
lines to roughly mirror that.

In evals, I found that we're often returning fewer than 1024 characters. This
PR updates the context resolver to load an adaptive number of lines based on
the 1024 character limit. 

Addresses #61745
## Test plan

Added new test. Also manually tested using GraphQL console.


jtibshirani:
This shows an improvement on CodeSearchNet:
```
Recall (files)	91/99
Recall (chunks)	74/99
Average chunk overlap	0.89
```

In the near future, we should be able to increase the chunk size even more, given our increased context window limits.

cmd/frontend/internal/context/resolvers/context.go
```diff
@@ -68,6 +70,13 @@ func (r *Resolver) GetCodyContext(ctx context.Context, args graphqlbackend.GetCo
 	})
 }
 
+// The rough size of a file chunk in runes. The value 1024 is due to historical reasons -- Cody context was once based
+// on embeddings, and we chunked files into ~1024 characters (aiming for 256 tokens, assuming each token takes 4
+// characters on average).
+//
+// Ideally, the caller would pass a token 'budget' and we'd use a tokenizer and attempt to exactly match this budget.
```
jtibshirani, line 77: Calling attention to this -- I think an ideal context API would let callers pass in a "token budget" that the backend must respect in its responses. The backend is allowed to optimize within that budget in whatever way it wants. (Or maybe this would be "character budget" to keep it simple?)

cmd/frontend/internal/context/resolvers/context.go
```diff
@@ -83,6 +92,29 @@ func (r *Resolver) fileChunkToResolver(ctx context.Context, chunk *codycontext.F
...
+		return nil, err
+	}
+
+	numLines := countLines(content, chunkSizeRunes)
+	endLine := chunk.StartLine + numLines - 1 // subtract 1 because endLine is inclusive
+	return graphqlbackend.NewFileChunkContextResolver(gitTreeEntryResolver, chunk.StartLine, endLine), nil
+}
+
+// countLines finds the number of lines corresponding to the number of runes. We 'round up' to include a line even if
+// it pushes the chunk over the desired number of runes. This is okay since the chunk size limit is very conservative.
```
keegancsmith, line 108: you can get very long lines. Should we put in an upperbound just to ensure counting up doesn't get too big?

jtibshirani: hmm, how about I switch to rounding down. That way we stay within the "token budget" which seems desirable.

PR #62664, opened 2024-05-14
bazel: transcribe test ownership to bazel tags

Strum355:
Transcribes from https://docs.google.com/spreadsheets/d/1Vu-Nk0dSCDJb_1cffbd29sr3NQkhJAmVAU9x_M7TnTU/edit#gid=0 into Bazel tags as part of #60455. Includes a quick script dev/check-test-ownership.sh that reports the % of tests that have an owning team assigned to them. 
We exclude certain types of tests due to their irrelevance (mostly things in e.g. ./.aspect, ./docs and diff tests that just check to make sure generated files are as expected).
Once we have this data being fed into BigQuery, we can start including it in our Redash dashboards.
A follow-up PR will be required that checks in CI to make sure we don't fall below the 70% ownership threshold outlined by the OKR.

## Test plan

Not changing any logic about anything, just metadata about the build


eseliger:
I wonder if this could be implemented using owner/codeowner files or so. That way we would get Own data in Sourcegraph on tests as well 😬

burmudar (approved):

Strum355:
> I wonder if this could be implemented using owner/codeowner files or so. That way we would get Own data in Sourcegraph on tests as well 😬

I had originally looked into that, but it operates at a different level than we'd want and there doesn't appear to be the same level of query-ability vs `bazel query`. Having it as part of bazel tags gives us a 1:1 relationship with the data that we consume from bazel for test timings/cache hits/etc, and an easy way to find what test targets aren't covered. We also get to avoid checking out sg/sg in https://buildkite.com/sourcegraph/devx-build-metrics for every data upload we perform

PR #62667, opened 2024-05-14
Update HubSpot contacts integration

akalia25:
Based on new cookies added to the GTM script, we are updating our private app integration to capture these values on a signup/sigin flow. The Google Tag Manager script only updates properties on HubSpot when a HubSpot form is completed and not during a regular sign-in/sign-up flow. 

## Test plan
Test via CI and see if I can test with a local deployment


cmd/frontend/internal/app/ui/handlers.go
```diff
@@ -586,20 +586,40 @@ func servePingFromSelfHosted(w http.ResponseWriter, r *http.Request) error {
...
-		SessionUTMSource:       getCookie("sg_utm_source"),
-		SessionUTMMedium:       getCookie("sg_utm_medium"),
-		SessionUTMContent:      getCookie("sg_utm_content"),
-		SessionUTMTerm:         getCookie("sg_utm_term"),
-		GoogleClickID:          getCookie("gclid"),
-		MicrosoftClickID:       getCookie("msclkid"),
-		HasAgreedToToS:         tosAccepted == "true",
+		IsServerAdmin:              true,
+		AnonymousUserID:            anonymousUserId,
+		LastSourceURL:              getCookie("last_page_seen_url"),
```
dadlerj, line 591: Are `LastSourceURL` and `LastPageSeenURL` the same thing? If so, one of the two should be deleted from the `ContactProperties` struct up in contacts.go!

cmd/frontend/hubspot/contacts.go
```diff
@@ -93,15 +131,36 @@ func newAPIValues(h *ContactProperties) *apiProperties {
 	apiProps.set("anonymous_user_id", h.AnonymousUserID)
 	apiProps.set("database_id", h.DatabaseID)
 	apiProps.set("has_agreed_to_tos_and_pp", h.HasAgreedToToS)
-	apiProps.set("last_source_url", h.LastSourceURL)
+	apiProps.set("last_source_url", h.LastPageSeenUrl)
+	apiProps.set("last_page_seen_url_short", h.LastPageSeenShort)
```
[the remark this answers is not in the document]
akalia25: There's no `last_page_seen_url` HubSpot contact property, instead we're updating the `last_source_url` Contact Property(as its connected to custom workflows) with the new `last_page_seen_url` Cookie that gets created/set on GTM.

cmd/frontend/internal/app/ui/handlers.go
akalia25: ah yes, good catch! i'll remove `LastPageSeenUrl` from the `ContactProperties`. :)

cmd/frontend/hubspot/contacts.go
```diff
@@ -42,33 +42,68 @@ type ContactProperties struct {
 	HasAgreedToToS               bool   `json:"has_agreed_to_tos_and_pp"`
 	VSCodyInstalledEmailsEnabled bool   `json:"vs_cody_installed_emails_enabled"`
 
+	// The URL of the last page the user saw across multiple cookie duration sessions.
+	LastPageSeenShort string `json:"last_page_seen_short"`
+	LastPageSeenMid   string `json:"last_page_seen_mid"`
+	LastPageSeenLong  string `json:"last_page_seen_long"`
+
 	// The URL of the first page a user landed on their latest session on a Sourcegraph site.
 	LastSourceURL string `json:"last_source_url"`
```
dadlerj, line 51: Could you just add or update the comments on this field and the various `LastPageSeen` fields above to make it more clear that they are (essentially) the same (and just have a different name for legacy reasons).

dadlerj (approved):
LGTM thanks!

cmd/gitserver/internal/git/gitcli/refs.go
eseliger, line 1: this diff somehow seems messed up, this change should not be reverted :)

akalia25: Thanks for helping fix this branch!!

cmd/frontend/hubspot/contacts.go
akalia25: added additional comments explaining the use of the legacy HS contact property `last_source_url` with the updated cookie names :)

PR #62671, opened 2024-05-14
gitserver(ListRefs): Fix type detection for lightweight tags

eseliger:
In git, there exist two types of tags: Lightweight tags, those are boring old refs that live in refs/tags/ and point directly to a commit.
And then there are annotated tags, those live in refs/tags/ too, but they don't point to a commit directly, instead they point to a tag object, which then points to a commit object.
That is used to add a message to a tag.

However, we didn't detect lightweight tags correctly, so they would be reported as Type: Branch (as they point to a commit).

This PR fixes it.

Test plan:

Added tests and verified manually using gRPC UI.


eseliger:
* **#62671** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62671?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

mmanela (approved):

PR #62675, opened 2024-05-14
sg msp: add category flag for 'tfc sync'

bobheadxi:
Enables staged rollouts of workspace updates.

## Test plan

```
sg msp tfc sync -all -category=test
```


jac (approved):
nit: rather than saying "must be used with '-all'" it might be clearer to say "can only be used with -all" as, at least to me, the current wording could also imply that "-all" requires "-category" to be specified

bobheadxi:
Good call @jac , I updated it and all similar occurrences!

PR #62678, opened 2024-05-14
Web: add mermaid diagram rendering

camdencheek:
This adds rendering for mermaid diagrams to the web app (both React and Svelte). We do not do this on the backend because mermaid is a JS library. This was not planned work, but it's been requested enough and is easy enough that I figured it was worth a couple hours to slot it in.

[Related Slack thread](https://sourcegraph.slack.com/archives/C05MHAP318B/p1715076201061259)

## Test plan

Manual testing. See demo video.


internal/htmlutil/policy.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 var (
-	chromaOnly   = regexp.MustCompile(`^(?:chroma-[a-zA-Z0-9\-]+)|chroma$`)
+	chromaOnly   = regexp.MustCompile(`^(?:chroma-[a-zA-Z0-9\-]+)|chroma|language-[a-zA-Z]+$`)
```
camdencheek, line 17: `GoldMark` already adds a `class="language-mermaid"`, but we were stripping that during sanitization

peterguy:
Screenshot 2024-05-15 at 13 56 13
Screenshot 2024-05-15 at 13 55 57

Examples of light and dark mode in Chrome.

peterguy (approved):
The render in dark mode isn't great (see previous comment for screenshot), but this otherwise looks a-ok!

camdencheek:
@peterguy good catch on dark mode!  I updated the PR to also support dark mode (at least somewhat). It doesn't live-update when the theme changes because that would take a bit more work to manually specify the theme, but it does at least use the theme you have configured.


screenshot-2024-05-16_14-39-18@2x

PR #62679, opened 2024-05-14
Docs: update integration test instructions

jtibshirani:
This PR updates the README for GraphQL integration tests to encourage devs to
run them in CI instead of running locally. It also update the local
instructions to use the `backend_integration_test` target.

## Test plan

Tested these instructions manually

### DIFF
--- a/dev/gqltest/README.md
+++ b/dev/gqltest/README.md
@@ -16,7 +16,7 @@ instead.
 
 Steps:
 1. Request "CI Secrets Read Access" in Entitle
-2. Run `sg bazel test //testing:backend_integration_test`
+2. Run `sg test bazel-backend-integration`
 
 ## How to add new tests
 
--- a/dev/gqltest/README.md
+++ b/dev/gqltest/README.md
@@ -2,102 +2,20 @@
 
 This directory contains API-based integration tests in the form of standard Go tests. It is called gqltest since most of our API is GraphQL. However, the test suite has been extended to test other endpoints such as streaming search.
 
-## How to set up credentials
-
-Tests use environment variables to accept credentials of different external services involved, it is suggested to use [direnv](https://direnv.net/) to persist those credentials for your convenience. Here is a comprehensive example `.envrc` file (you're free to use any other means, e.g. `.profile` or `.bashrc`):
-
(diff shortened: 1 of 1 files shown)


dev/gqltest/README.md
```diff
@@ -2,102 +2,20 @@
...
-
-```sh
-→ go test -long -run TestSearch -base-url "https://sourcegraph.test:3443"
-2020/07/17 14:20:59 Site admin authenticated: gqltest-admin
-PASS
-ok  	github.com/sourcegraph/sourcegraph/dev/gqltest	3.073s
-```
+Steps:
+1. Request "CI Secrets Read Access" in Entitle
+2. Run `sg bazel test //testing:backend_integration_test`
```
jtibshirani, line 18: I'm not 100% sure this works, as I wasn't able to do this successfully (which was maybe because I'm on Mac OS??)

dev/gqltest/README.md
```diff
@@ -108,4 +26,3 @@ Adding new tests to this test suite is as easy as adding a Go test, here are som
   - Delete new users created during the test.
   - Delete external services created during the test.
   - Although, sometimes you would not want to delete an entity so you could login and inspect the failure state.
-- Prefix your branch name with `backend-integration/` will run integration tests in CI on your pull request.
```
jtibshirani, line 111: This didn't seem helpful, as we always run integration tests on pull requests 🤔

dev/gqltest/README.md
camdencheek: I think we actually want [`sg test bazel-backend-integration`](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@200ad688ed5331c81e49cc978d30b9f919eb4531/-/blob/sg.config.yaml?L1766-1782), which is the setup that actually uses the entitle permissions. Otherwise, the required env vars won't be set. 

I haven't tested this in a while though, so YMMV 🙂

camdencheek (approved):
Thank you!

dev/gqltest/README.md
jtibshirani: Thanks for the correction!

Issue #62390, opened 2024-05-02
Svelte: dotcom readiness


fkling:
This is an epic to collect tasks needed to be done to enable the Svelte app on dotcom. This list is not complete yet, a big part of working on this is completing the list.

vovakulikov:
This is a kind of random thought. In our react codebase, we almost everywhere use the `isSourcegraphDotCom` flag. I understand that this is a very convenient way to distinguish between commercial and marketing instances. 

But I would like to challenge us to think not in terms of in-place flag checks but in terms of extension points of the app to make necessary differences in UI/logic/etc. Why? I actually don't have a clear list of arguments here but it feels like a coupled code-base problem when we have to make differences not in one but in many places at once for some specific instance (Implementation leak problem), consumers know too much about environment, leads to complexity of testing and reusing it later.

What do I propose instead? So ideally I would like to have one place (file/store/etc) where we will define feature set specific to the Sourcegraph dot com instance and include it by one check via `isSourcegraphDotCom`.

fkling:
I agree, that would be ideal. My gut reaction is that this only works if all of the dotcom differences are data driven. E.g. I can totally see how this works for main navigation, but there might be other areas where it's not as obvious.
Of course we can always add more specific settings/props so that we do not couple it to `isSourcegraphDotCom` specifically. Maybe that's actually a good approach.

PR #62682, opened 2024-05-14
Svelte: update navigation bar for dotcom deployment

camdencheek:
This updates the navigation bar in svelte to show the dotcom-specific nav bar on dotcom (whenever svelte is enabled there).

Contributes to https://github.com/sourcegraph/sourcegraph/issues/62390

## Test plan

Tested that the links work, the "current active" is accurate, and that everything looks right both in the dotcom version and the non-dotcom version.

Used `SOURCEGRAPHDOTCOM_MODE=true sg start enterprise-sveltekit` for testing. 

One caveat is that the "About sourcegraph" link does not work in this mode locally because that's handled by Cloudflare.


client/web-sveltekit/src/lib/navigation/mainNavigation.ts
```diff
@@ -72,76 +67,5 @@ export interface NavigationMenu {
...
  */
 export function isCurrent(entry: NavigationEntry, page: Page): boolean {
-    return page.url.pathname.startsWith(entry.href)
+    return page.url.pathname === entry.href
 }
-
-/**
- * The main navigation of the application.
- */
-export const mainNavigation: (NavigationMenu | NavigationEntry)[] = [
```
camdencheek, line 81: Small opinionated change: I think it makes sense for components to live in `lib` and any static data (like route definitions) to go in `routes`. This keeps `lib` dotcom-agnostic.

client/web-sveltekit/src/lib/navigation/mainNavigation.ts
```diff
@@ -72,76 +67,5 @@ export interface NavigationMenu {
  * by means of comparing the entry's href with the current page's URL.
  */
 export function isCurrent(entry: NavigationEntry, page: Page): boolean {
-    return page.url.pathname.startsWith(entry.href)
+    return page.url.pathname === entry.href
```
camdencheek, line 70: This was necessary because the "About Sourcegraph" entry links to `/`, which prefixes _every_ path, so it was always being shown as active. I _think_ this is okay since we do not have sub-paths for these pages. The alternative was to add overridable "isCurrent" logic to every entry, which didn't feel ideal. 

screenshot-2024-05-14_19-17-21@2x

client/web-sveltekit/src/routes/+layout.svelte
```diff
@@ -1,27 +1,30 @@
...
     import { getGraphQLClient } from '$lib/graphql/apollo'
     import { isRouteEnabled } from '$lib/navigation'
 
     import type { LayoutData } from './$types'
-    import FuzzyFinderContainer from '$lib/fuzzyfinder/FuzzyFinderContainer.svelte'
+    import { mainNavigation, dotcomMainNavigation } from './navigation'
 
     export let data: LayoutData
 
+    const navigationEntries = window.context.sourcegraphDotComMode ? dotcomMainNavigation : mainNavigation
```
camdencheek, line 26: Should I be depending on `window.context` directly here?

fkling: Good question. I know @vovakulikov has some thoughts around keeping dotcom differences managable.

fkling: Maybe we could do it similar to data loading: Pages/components should not access `window.context` directly. Instead it should only be accessed in data loaders. So in this case the root layout data loader could return the navigation entries.

No idea whether that's good or not.

fkling (approved):
Accepting to unblock. We probably want to think about the "best" way for accessing this information (see inline comment for a suggestion).

It would also be nice if we could run dotcom mode standalone without a backend. I suspect we can add something like `pnpm dev:dotcom`, which sets some env variables which in turn could populate `window.context`. But this doesn't have to be done in this PR.

client/web-sveltekit/src/routes/+layout.svelte
vovakulikov: I was thinking about this recently: how `sourcegraphDotCom` checks consumed our codebase and how painful it is working with these conditions in UI when you have to think about different layouts in different environments. 

Ideally, I would have a different place, like a dotcom directory, where I keep all specific to dotcom logic/UI. 
Then, in the app, I would add some extension points with which we could extend the main app with dotcom logic. 

The problem is that this isn't usually trivial. Sometimes, UI and logic are combined and coupled very closely. 
In this particular example, it's simple enough because the dotcom difference doesn't have anything with UI, but with the data that we provide, this is the best-case scenario. If some page is very different on dotcom compared to the main app I would redo this page from scratch with some env-agnostic components rather than populate it with checks. 

Back to this specific change here, I don't have a clear thought on how exactly we should achieve this clear separation between different environments, I would try to not working with check in the UI (which this PR doesn't do, which is great) and see how other dotcom parts will go into the main app, and then come up with more specific idea

vovakulikov (approved):

client/web-sveltekit/src/routes/+layout.svelte
camdencheek: I like the idea of putting the `window.context` accesses in the data loaders. It makes sense to me. And separates concerns a little better. It's not obvious to me that `window.context` will always be the source of this information.

client/web-sveltekit/src/routes/+layout.ts
```diff
@@ -45,6 +45,7 @@ export const load: LayoutLoad = async ({ fetch }) => {
 
     return {
         user: result.data.currentUser,
+        sourcegraphDotComMode: window.context.sourcegraphDotComMode,
```
fkling, line 48: Ah, I thought we could return the actual navigation entries from here, making the page complete unaware of dotcom mode. 
But as Vova said, sometimes it might not be as easy to do (though maybe we should work towards this), so this is probably good enough for now.

camdencheek: Ah! I see. That makes sense. Went ahead and passed the entries here to make the page entirely unaware of dotcom mode.

PR #62685, opened 2024-05-15
msp/cloudrun: use GA launch stage

bobheadxi:
VPC direct egress is now GA: see example in https://registry.terraform.io/providers/hashicorp/google/5.29.0/docs/resources/cloud_run_v2_service#example-usage---cloudrunv2-service-directvpc and https://cloud.google.com/run/docs/configuring/vpc-direct-vpc

image

This also fixes the infinite `GA` -> `BETA` drift we have in TFC

## Test plan

n/a


jac (approved):
Nice!

PR #62690, opened 2024-05-15
Cody Ignore: set lower client version constraint to enable clients testing

taras-yemets:
Sets version constraints to lower pre-release versions for VS Code and JetBrains Cody extensions if `cody-context-filters-clients-test-mode` feature flag is set to `true` to enable testing clients Cody Ignore implementation in the prod environment.
More context in [this](https://sourcegraph.slack.com/archives/C06Q9R098N4/p1715724952334729?thread_ts=1715038896.786949&cid=C06Q9R098N4) thread.

## Test plan
- Updated unit tests
- CI


cmd/frontend/internal/httpapi/completions/handler_test.go
```diff
@@ -267,34 +267,44 @@ func TestCheckClientCodyIgnoreCompatibility(t *testing.T) {
...
+				"client-name":    []string{string(types.CodyClientJetbrains)},
+				"client-version": []string{"5.5.10"},
+			},
+			want:              nil,
+			isClientsTestMode: true,
+		},
 		{
 			// See https://pkg.go.dev/github.com/Masterminds/semver#readme-working-with-pre-release-versions
-			name: "jetbrains: pre-release version doesn't match constraint if \"cody-context-filters-allow-pre-release-client-versions\" feature flag is enabled",
+			name: "jetbrains: pre-release version doesn't match constraint if \"cody-context-filters-clients-test-mode\" feature flag is enabled",
```
kalanchan, line 282: can we add one more test here to show that the test will pass if we use a JetBrains version like `5.5.10-nightly` ?

kalanchan (approved):
I'm going to approve for now so that I can test this on SG02 and unblock testing. 

@dominiccooney please review afterwards and leave any comments/concerns with the approach. 

@taras-yemets if you could update the tests again that would be great. 


next steps: we plan on cutting JB 6.0.0 next week, but this PR will unblock us for now. We should think of a way to test VS Code 1.20.0 as well before next week

PR #62691, opened 2024-05-15
msp: Expose contract mock for tests

burmudar:
I needed to provide a dummy contract and saw that MSP already had this available but it was private.

This PR makes it public while trying to retain it's original intent.

## Test plan
* Tested in spirit here https://github.com/sourcegraph/repoferee/pull/14/commits/5e649af3f47c3300ae7bef6f619ee537523415cd
* `bazel test //lib/managedservicesplatform/runtime/...`
* CI

### DIFF
--- a/lib/managedservicesplatform/runtime/contract/mock/contract_mock.go
+++ b/lib/managedservicesplatform/runtime/contract/mock/contract_mock.go
@@ -1,0 +1,37 @@
+package mock
+
+import (
+	"cmp"
+	"testing"
+
+	"github.com/sourcegraph/log/logtest"
+
+	"github.com/sourcegraph/sourcegraph/lib/managedservicesplatform/runtime/contract"
+)
+
+type MockServiceMetadata struct {
+	name    string
+	version string
+}
+
+type MockError struct {
+	error
+}
+
+func (m MockServiceMetadata) Name() string    { return cmp.Or(m.name, "mock-name") }
+func (m MockServiceMetadata) Version() string { return cmp.Or(m.version, "mock-version") }
+
+// NewMockContract returns a new contract instance from the given env. If there is an error parsing the given environment
+// a MockError is returned that contains the error.
+//
+// Otherwise a new contract instance is returned as well as the environment validation result from `env.Validate()`
+func NewMockContract(t *testing.T, mockServiceMeta contract.ServiceMetadataProvider, env ...string) (*contract.Contract, error) {
+	t.Helper()
+	e, err := contract.ParseEnv(env)
+	if err != nil {
+		return nil, MockError{err}
+	}
+
+	c := contract.New(logtest.Scoped(t), mockServiceMeta, e)
+	return &c, e.Validate()
+}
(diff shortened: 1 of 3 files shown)


jac (approved):
Nice QOL improvement 👍

lib/managedservicesplatform/runtime/contract/mock/contract_mock.go
```diff
@@ -0,0 +1,37 @@
+package mock
```
bobheadxi, line 1: Can we name this `contracttest` per stdlib convention?

lib/managedservicesplatform/runtime/contract/mock/contract_mock.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+type ServiceMetadata struct {
+	name    string
+	version string
+}
+
+type MockError struct {
+	error
+}
```
bobheadxi, line 19: what is this type for?

lib/managedservicesplatform/runtime/contract/mock/contract_mock.go
```diff
@@ -0,0 +1,37 @@
...
+	"testing"
+
+	"github.com/sourcegraph/log/logtest"
+
+	"github.com/sourcegraph/sourcegraph/lib/managedservicesplatform/runtime/contract"
+)
+
+type ServiceMetadata struct {
+	name    string
+	version string
```
bobheadxi, line 14: should this be exported so that callers can set this values in testing?

lib/managedservicesplatform/runtime/contract/mock/contract_mock.go
burmudar: It's so that you can distinguish between between the mock failing because `contract.ParseEnv` failed and the error returned by `env.Validate()`

lib/managedservicesplatform/runtime/contract/mock/contract_mock.go
burmudar: Updated 👍🏼

lib/managedservicesplatform/runtime/contract/mock/contract_mock.go
bobheadxi: I suggest against this - callers should be able to not differentiate between mock errors vs not, unless they DIY it (provide a sentinel error to return)

lib/managedservicesplatform/runtime/contract/contractest/contract_mock.go
```diff
@@ -0,0 +1,38 @@
...
+
+// NewMockContract returns a new contract instance from the given env. If there is an error parsing the given environment
+// a MockError is returned that contains the error.
+//
+// Otherwise a new contract instance is returned as well as the environment validation result from `env.Validate()`
+func NewContract(t *testing.T, mockServiceMeta contract.ServiceMetadataProvider, env ...string) (*contract.Contract, error) {
+	t.Helper()
+	e, err := contract.ParseEnv(env)
+	if err != nil {
+		return nil, MockError{err}
```
bobheadxi, line 33: A naked error/simply wrapped error is sufficient IMO:

```suggestion
		return nil, errors.New(err, "contract.ParseEnv")
```

PR #62695, opened 2024-05-15
Fix bedrock URL encoding to mimic AWS CLI

RXminuS:
Follow up from #62642. Brings URL encoding in-line with AWS cli.

## Test plan
- added unit test
- Locally verified that Bedrock arn is properly encoded and completion requests succeed


loujar (approved):

internal/completions/client/awsbedrock/bedrock.go
```diff
@@ -282,6 +271,28 @@ func (c *awsBedrockAnthropicCompletionStreamClient) makeRequest(ctx context.Cont
 	return resp, nil
 }
 
+// Builds a bedrock api URL from the configured endpoint url.
+// If the endpoint isn't valid, falls back to the default endpoint for the specified fallbackRegion
+func buildApiUrl(endpoint string, model string, stream bool, fallbackRegion string) *url.URL {
+	apiURL, err := url.Parse(endpoint)
+	if err != nil || apiURL.Scheme == "" {
```
eseliger, line 278: Do we want to just swallow the error here? Could be hard to debug why requests are being made to the wrong region 🤔

eseliger (approved):

internal/completions/client/awsbedrock/bedrock.go
RXminuS: This is how it worked before, I've not changed any functionality

RXminuS: I think the idea is that it should handle blank config settings etc as well

PR #62698, opened 2024-05-15
SSC: Add seat count to SSC telemetry

vdavid:
This PR is built on top of https://github.com/sourcegraph/sourcegraph/pull/62694

Data team: does the name `seatCount` sound good to you? It's meant to represent the fact that the number isn't the actual number of team members but the number of seats the customer bought for their team. I suppose these two numbers will mostly be equal, but still logically different.

@chrsmith, the solution uses the `useCodySubscriptionData()` hook which uses the `/team/current/subscription` endpoint, rather than the `useCodySubscriptionSummaryData()` hook that uses `/team/current/subscription/summary`. This makes it expensive, and I'm unsure if we want to use that hook on all of these pages, especially considering that we don't have caching in the underlying `useSSCData()` hook yet.
Potential solutions:
a.) Let's add the seat count to the the /summary endpoint. Not sure if this is technically viable or this info is only with Stripe? I think the latter, unfortunately.
b.) Let's implement that front-end caching. But then, I'm not sure if having the `/summary` endpoint makes sense anymore. Maybe better merge it with `/subscription`?

Note: also removed several deprecated `EVENT_LOGGER.log()` calls to clean up a bit.

## Test plan

Haven't tested, plus might cause potential performance issues, hence the draft state.


kelsey-brown:
Ah yes, good call out. I didn't realize this two could theoretically be different. `seatCount` SGTM

chrsmith (changes requested):
_Update_: Removed my lengthy comment on GitHub, and will start a thread on Slack, just so we can better coordinate on the specific tradeoffs to concern and approaches available to us.

In short, I think the right way to do this is to update the "subscription summary" endpoint to include this information. Because that is something we were planning on making available to all of the UI pages anyways, and is also the most efficient way to retrieve this data on the backend.

dadlerj:
Thank you David! I left a comment in Slack on how we'd love for this value to be logged by all cody clients, not just the SSC service. More details here: https://sourcegraph.slack.com/archives/C0477DQU02J/p1715808987351859?thread_ts=1715127123.424479&cid=C0477DQU02J

PR #62701, opened 2024-05-15
fix/telemetry(auth): return authenticated context from session.SetActorFromUser

bobheadxi:
This change fixes telemetry for various sign-in, sign-up telemetry events - `session.SetActorFromUser` accepts a context and returns a new one, but never does anything to the returned context. It makes sense for this to authenticate the context because it is only ever used on a success case, and in turn this fixes telemetry by providing the telemetry SDK an actor to extract a user ID for.

For the events affected by `session.SetActorFromUser` I've incremented the event version to `2`.

## Test plan

1. Test on sign-in-success that a event with the user is recorded
2. Test on `session.SetActorFromUser` to verify context is authenticated

### DIFF
--- a/internal/telemetry/events.go
+++ b/internal/telemetry/events.go
@@ -16,12 +16,6 @@ type eventFeature string
 const (
 	// FeatureExample is a value for testing - do not use.
 	FeatureExample eventFeature = "exampleFeature"
-
-	// FeatureSignIn, FeatureSignOut, and FeatureSignUp are added here as telemetry
-	// examples - most callsites can directly provide the relevant feature.
-	FeatureSignIn  eventFeature = "signIn"
-	FeatureSignOut eventFeature = "signOut"
-	FeatureSignUp  eventFeature = "signUp"
 )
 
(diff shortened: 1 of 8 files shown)


internal/telemetry/events.go
```diff
@@ -16,12 +16,6 @@ type eventFeature string
 const (
 	// FeatureExample is a value for testing - do not use.
 	FeatureExample eventFeature = "exampleFeature"
-
-	// FeatureSignIn, FeatureSignOut, and FeatureSignUp are added here as telemetry
-	// examples - most callsites can directly provide the relevant feature.
-	FeatureSignIn  eventFeature = "signIn"
-	FeatureSignOut eventFeature = "signOut"
-	FeatureSignUp  eventFeature = "signUp"
 )
```
bobheadxi, line 19: This was back before I realised it was kind of dumb to define event features all in one place, we now define features directly at callsite. This should be a no-op change.

akalia25 (approved):
Thank you for the quick fix!

eseliger (approved):

PR #62702, opened 2024-05-15
sg/msp: add repo annotation to delivery pipeline

jac:
Adds the repository as an annotation to the delivery pipeline so that we can later query it e.g. from cloud-relay to include a link to the commit in the message we send to slack

I also tried adding annotations to the releases but unfortunately they don't get sent as part of the notification message

side note: the annotations support free-form text without the same limitations the labels have

part of CORE-36
## Test plan

locally tested

### DIFF
--- a/dev/managedservicesplatform/internal/resource/deliverypipeline/deliverypipeline.go
+++ b/dev/managedservicesplatform/internal/resource/deliverypipeline/deliverypipeline.go
@@ -36,6 +36,9 @@ type Config struct {
 	// TargetStages lists targets in the order they should appear in the pipeline.
 	TargetStages []Target
 
+	// Repository is the source repository for this pipeline.
+	Repository string
+
 	// Suspended prevents releases and rollouts from being created, rolled back,
 	// etc using this pipeline: https://cloud.google.com/deploy/docs/suspend-pipeline
 	Suspended bool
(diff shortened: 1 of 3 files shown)


dev/managedservicesplatform/internal/resource/deliverypipeline/deliverypipeline.go
```diff
@@ -36,6 +36,9 @@ type Config struct {
 	// TargetStages lists targets in the order they should appear in the pipeline.
 	TargetStages []Target
 
+	// Repository is the source repository for this pipeline.
```
bobheadxi, line 39:
```suggestion
	// Repository is the source code repository for the images delivered to this pipeline.
```

dev/managedservicesplatform/stacks/cloudrun/cloudrun.go
```diff
@@ -66,6 +66,7 @@ type Variables struct {
 	IAM iam.CrossStackOutput
 
 	Service     spec.ServiceSpec
+	Repository  string
```
bobheadxi, line 69:
```suggestion
    // Repository is the source code repository for the image to deploy
	Repository  string
```

Repository is ambiguous in the context of images, as many image hosts also call themselves "repositories" - a clarifying docstring here would be good I think, just to be safe

bobheadxi (approved):

bobheadxi:
Ah, didn't realize automerge was on - @jac FYI about my review comments, just docstring suggestions :)

dev/managedservicesplatform/internal/resource/deliverypipeline/deliverypipeline.go
```diff
@@ -89,6 +92,10 @@ func New(scope constructs.Construct, id resourceid.ID, config Config) (*Output,
 				Stages: pointers.Ptr(newStages(config)),
 			},
 
+			Annotations: &map[string]*string{
+				"repository": pointers.Ptr(config.Repository),
```
bobheadxi, line 96: I might also suggest calling this `source.repository`

PR #62263, opened 2024-04-30
lib/enterpriseportal: initial service API for RFC 885

bobheadxi:
See [RFC 885 Sourcegraph Enterprise Portal (go/enterprise-portal)](https://docs.google.com/document/d/1tiaW1IVKm_YSSYhH-z7Q8sv4HSO_YJ_Uu6eYDjX7uU4/edit#heading=h.tdaxc5h34u7q) - closes CORE-6. The only files requiring in-depth review are the `.proto` files, as everything else is generated:

- `lib/enterpriseportal/subscriptions/v1/subscriptions.proto`
- `lib/enterpriseportal/codyaccess/v1/codyaccess.proto`

This PR only introduces API definitions - implementation will come as subsequent PRs, tracked in the ["Launch Enterprise Portal" Linear project](https://linear.app/sourcegraph/project/launch-sourcegraph-enterprise-portal-ee5d9ea105c2).

Before reviewing the diffs, **please review this PR description in depth**.

### Design goals

This initial schema aims to help achieve CORE-97 by adding our initial "get subscription Cody access", as well our general Stage 1 goal of providing read-only access to our existing enterprise subscription mechanisms. In doing so, we can start to reshape the API in a way that accommodates future growth and addresses some debt we have accumulated over time, before the Stage 2 goal of having the new Enterprise Portal be the source-of-truth for all things subscriptions.

I am also aiming for a conservative approach with the Cody Gateway access related RPCs, to ease migration risks and allow for Cody teams to follow up quickly with more drastic changes in a V2 of the service after a Core-Services-driven migration to use the new service: https://github.com/sourcegraph/sourcegraph/pull/62263#issuecomment-2101874114

### Design overview

- **Multiple services**: Enterprise Portal aims to be the home of most Enterprise-related subscription and access management, but each component should be defined as a separate service to maintain clear boundaries between "core" capabilities and future extensions. One problem we see in the `dotcom { productSubscriptions }` is the embedding of additional concepts like Cody Gateway access makes the API surface unwieldy and brittle, and encourages an internal design that bundles everything together (the `product_subscriptions` table has 10 `cody_gateway_*` columns today). More concretely, this PR designs 2 services that Enterprise Portal will implement:
  - `EnterprisePortalSubscriptionsService` (`subscriptions.proto`): subscriptions and licenses CRUD
  - `EnterprisePortalCodyGatewayService` (`codygateway.proto`): Enterprise Cody Gateway access
[...]

PR #62706, opened 2024-05-15
feat/enterprise-portal: DB layer for {Get/List}CodyGatewayAccess 

bobheadxi:
Part of CORE-112. We need to implement the `CodyAccess` service proposed in https://github.com/sourcegraph/sourcegraph/pull/62263, so that Cody Gateway can depend on it as we start a transition over to Enterprise Portal as the source-or-truth for Cody Gateway access; see the [Linear project](https://linear.app/sourcegraph/project/kr-launch-enterprise-portal-for-cody-gateway-and-cody-analytics-ee5d9ea105c2/overview). This PR implements the data layer by reading directly from the Sourcegraph.com Cloud SQL database, and a subsequent PR https://github.com/sourcegraph/sourcegraph/pull/62771 will expose this via the API and also implement auth; nothing in this PR is used yet.

Most things in this PR will be undone by the end of a [follow-up project](https://linear.app/sourcegraph/project/kr-enterprise-portal-manages-all-enterprise-subscriptions-12f1d5047bd2/overview) tentatively slated for completion by end-of-August.

### Query

I've opted to write a new query specifically to fetch the data required to fulfill the proposed `CodyAccess` RPCs; the existing queries fetch a lot more than is strictly needed, and often make multiple round trips to the database. The new query fetches everything it needs for get/list in a single round trip.

`EXPLAIN ANALYZE` of the new list-all query against the Sourcegraph.com production database indicates this is likely performant enough for our internal-only use cases, especially as this will only be around for a few months.

```
                                                                                QUERY PLAN                                                                                 
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 HashAggregate  (cost=1610.56..1629.45 rows=1511 width=121) (actual time=23.358..24.921 rows=1512 loops=1)
   Group Key: ps.id
   ->  Hash Left Join  (cost=958.18..1585.58 rows=1999 width=1094) (actual time=8.258..12.255 rows=2748 loops=1)
         Hash Cond: (ps.id = active_license.product_subscription_id)
         ->  Hash Right Join  (cost=67.00..689.14 rows=1999 width=956) (actual time=1.098..3.970 rows=2748 loops=1)
               Hash Cond: (product_licenses.product_subscription_id = ps.id)
               ->  Seq Scan on product_licenses  (cost=0.00..616.88 rows=1999 width=919) (actual time=0.015..1.769 rows=2002 loops=1)
                     Filter: (access_token_enabled IS TRUE)
                     Rows Removed by Filter: 1789
               ->  Hash  (cost=48.11..48.11 rows=1511 width=53) (actual time=1.055..1.056 rows=1512 loops=1)
                     Buckets: 2048  Batches: 1  Memory Usage: 93kB
                     ->  Seq Scan on product_subscriptions ps  (cost=0.00..48.11 rows=1511 width=53) (actual time=0.016..0.552 rows=1512 loops=1)
         ->  Hash  (cost=874.39..874.39 rows=1343 width=154) (actual time=7.123..7.125 rows=1343 loops=1)
               Buckets: 2048  Batches: 1  Memory Usage: 248kB
               ->  Subquery Scan on active_license  (cost=842.02..874.39 rows=1343 width=154) (actual time=5.425..6.461 rows=1343 loops=1)
                     ->  Unique  (cost=842.02..860.96 rows=1343 width=162) (actual time=5.422..6.268 rows=1343 loops=1)
                           ->  Sort  (cost=842.02..851.49 rows=3788 width=162) (actual time=5.421..5.719 rows=3791 loops=1)
                                 Sort Key: product_licenses_1.product_subscription_id, product_licenses_1.created_at DESC
                                 Sort Method: quicksort  Memory: 1059kB
                                 ->  Seq Scan on product_licenses product_licenses_1  (cost=0.00..616.88 rows=3788 width=162) (actual time=0.003..1.872 rows=3791 loops=1)
 Planning Time: 2.266 ms
 Execution Time: 28.568 ms
```

We noted the lack of index on `product_livenses.subscription_id`, but it doesn't seem to be an issue at this scale, so I've left it as is.

### Pagination

After discussing with Erik, we decided there is no need to implement pagination for the list-all RPC yet; a rough upper bound of 1kb per subscription * 1511 rows (see `EXPLAIN ANALYZE` above) is 1.5MB, which is well below the per-message limits we have set for Sourcegraph-internal traffic (40MB), and below the [default 4MB limit](https://pkg.go.dev/google.golang.org/grpc#MaxRecvMsgSize) as well. In https://github.com/sourcegraph/sourcegraph/pull/62771 providing pagination parameters will result in a `CodeUnimplemented` error.

We can figure out how we want to implement pagination as part of the [follow-up project](https://linear.app/sourcegraph/project/kr-enterprise-portal-manages-all-enterprise-subscriptions-12f1d5047bd2/overview) to migrate the data to an Enterprise-Portal-owned database.

### Testing

A good chunk of this PR's changes are exposing a small set of `cmd/frontend` internals **for testing** via the new `cmd/frontend/dotcomproductsubscriptiontest`:

- seeding test databases with subscriptions and licenses
- for "regression testing" the new read queries by validating what the new read queries get, against what the existing GraphQL resolvers resolve to. This is important because the GraphQL resolvers has a lot of the override logic

See `TestGetCodyGatewayAccessAttributes` for how all this is used.

image

There is also some hackery involved in setting up a `pgx/v5` connection used in MSP from the `sql.DB` + `pgx/v4` stuff used by `dbtest`; see `newTestDotcomReader` docstrings for details.

## Test plan

```
go test -v ./cmd/enterprise-portal/internal/dotcomdb
```


bobheadxi:
* **#62771** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62771?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#62706** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62706?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

cmd/frontend/dotcomproductsubscriptiontest/dotcomproductsubscriptiontest.go
```diff
@@ -0,0 +1,81 @@
+// Package dotcomproductsubscriptiontest provides a shim for internals from
+// cmd/frontend/internal/dotcom/productsubscription, to be used in tests in
+// packages outside of cmd/frontend.
+//
+// 👷 This package is intended to be a short-lived mechanism, and should be
```
eseliger, line 5: let's hope the short lived will be true 😆

cmd/enterprise-portal/internal/dotcomdb/dotcomdb.go
```diff
@@ -1,32 +1,250 @@
+// Package dotcomdb provides a read-only shim over the Sourcegraph.com database
+// and aims to provide values as they behave in Sourcegraph.com API today for
+// Enterprise Portal to serve through its new API.
+//
+// 👷 This package is intended to be a short-lived mechanism, and should be
+// removed as part of https://linear.app/sourcegraph/project/12f1d5047bd2/overview.
```
eseliger, line 6: totally off topic:

I wonder what our stance on using linear links vs github issue links is, I assume these links aren't publicly accessible and will mean that no outside people can click on them?

eseliger (approved):
LGTM, seems reasonable enough to me to temporarily do this slightly hacky thing of reaching across cmds to ensure a smooth migration :) 

Thanks also for the query plan, always good to know people think about their load on a shared system 🚀 

Also - we do a migration. Finally. Yay!

cmd/enterprise-portal/internal/dotcomdb/dotcomdb.go
```diff
@@ -1,32 +1,250 @@
...
+
+func (r *Reader) GetCodyGatewayAccessAttributesBySubscription(ctx context.Context, subscriptionID string) (*CodyGatewayAccessAttributes, error) {
+	query := selectCodyGatewayAccessAttributes +
+		"WHERE ps.id = $1 GROUP BY ps.id"
+	row := r.conn.QueryRow(ctx, query, subscriptionID)
+	return r.scanOneCodyGatewayAccessAttributes(row)
+}
+
+func (r *Reader) GetCodyGatewayAccessAttributesByAccessToken(ctx context.Context, token string) (*CodyGatewayAccessAttributes, error) {
+	// Below is copy-pasta from 'func (t dbTokens) LookupProductSubscriptionIDByAccessToken'
```
chrsmith, line 192: Perhaps "is copy-pasta from" could be replaced with a slightly more formal/boring "This was copied from". But style nits aside, thank you for adding this comment!

cmd/enterprise-portal/internal/dotcomdb/dotcomdb.go
```diff
@@ -1,32 +1,250 @@
...
+	// in 'cmd/frontend/internal/dotcom/productsubscription'.
+	if !strings.HasPrefix(token, productsubscription.AccessTokenPrefix) &&
+		!strings.HasPrefix(token, license.LicenseKeyBasedAccessTokenPrefix) {
+		return nil, errors.WithSafeDetails(ErrCodyGatewayAccessNotFound, "invalid token with unknown prefix")
+	}
+	decoded, err := hex.DecodeString(token[len(license.LicenseKeyBasedAccessTokenPrefix):])
```
chrsmith, line 198: For readability, consider moving the non-trivial expression to a local variable.
```suggestion
	tokenSansPrefix := token[len(license.LicenseKeyBasedAccessTokenPrefix):]
	decoded, err := hex.DecodeString(tokenSansPrefix)
```

cmd/enterprise-portal/internal/dotcomdb/dotcomdb.go
```diff
@@ -1,32 +1,250 @@
...
+	}
+	// End copy-pasta
+
+	query := selectCodyGatewayAccessAttributes +
+		"GROUP BY ps.id HAVING $1 = ANY(array_agg(tokens.license_key_hash))"
+	row := r.conn.QueryRow(ctx, query, decoded)
+	return r.scanOneCodyGatewayAccessAttributes(row)
+}
+
+func (r *Reader) scanOneCodyGatewayAccessAttributes(row pgx.Row) (*CodyGatewayAccessAttributes, error) {
```
chrsmith, line 210: Since this doesn't depend on `r *Reader` and the name `scanOneCodyGatewayAccessAttributes` is a bit overkill[1], what do you think about just making this an independent function?

[1] I guess the only thing I can really fault is scan "one", which is kinda implied because it takes a `pgx.Row` and not `pgx.Rows`. But that name is... long.

```go
func scan(row pgx.Row) (*CodyGatewayAccessAttributes, error) {
    ...
}
```

cmd/enterprise-portal/internal/dotcomdb/dotcomdb.go
```diff
@@ -1,32 +1,250 @@
...
+	return &attrs, nil
+}
+
+func (r *Reader) GetAllCodyGatewayAccessAttributes(ctx context.Context) ([]*CodyGatewayAccessAttributes, error) {
+	query := selectCodyGatewayAccessAttributes + "GROUP BY ps.id"
+	rows, err := r.conn.Query(ctx, query)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get cody gateway access attributes")
+	}
+	defer rows.Close()
```
chrsmith, line 240: ::sigh:: I am honor bound to point out that this is incorrect because it silently ignores the error returned on `Close()`. And therefore, the "correct" way to write this would be to express this as:

```suggestion
	defer yoloHelperLibrary.CallCloseAndIgnoreTheErrorAnyways(rows)
```

cmd/enterprise-portal/internal/dotcomdb/dotcomdb_test.go
```diff
@@ -0,0 +1,277 @@
...
+	ctx := context.Background()
+
+	// Set up a Sourcegraph test database.
+	sgtestdb := dbtest.NewDB(t)
+
+	// HACK: Extract the underlying pgx connection from the Sourcegraph test
+	// database, so that we can tease out the connection string that was used
+	// to connect to it, since we want to use pgx/v5 directly as offered by MSP
+	// utilities. This is sneaky but easier than changing the extensive
+	// infrastructure around dbtest
```
chrsmith, line 37: Super minor nitpick: Long form comments like this should end with terminal punctuation.

```suggestion
	// infrastructure around dbtest.
```

chrsmith: Here and throughout the file. 👉 👈 Please.

cmd/frontend/internal/dotcom/productsubscription/codygateway_graphql.go
```diff
@@ -16,6 +19,18 @@ type codyGatewayAccessResolver struct {
 	sub *productSubscription
 }
 
+// NewCodyGatewayAccessResolver returns a new CodyGatewayAccessResolver for the
+// given product subscription. ONLY FOR TESTING, DO NOT USE - see package
```
chrsmith, line 23: Just to confirm my understanding, the GraphQL types/schema are already defined. We are just re-implementing the same resolver so that we can verify the implementation is correct?

sg.config.yaml
```diff
@@ -382,6 +382,24 @@ commands:
       - cmd/msp-example
       - lib/managedservicesplatform
 
+  enterprise-portal:
```
chrsmith, line 385: What are the ramifications of adding this to the `sg.config.yaml` file? Will all new local `sg` clusters include the `enterprise-portal` binary now? Should they?

It seems like we would want a one-off micro service to steer clear of the larger "Sourcegraph Enterprise Cluster Config" machinery?

chrsmith (approved):
Some minor nitpicks but overall this looks good to me.

That query does give me pause... I'm guessing that because it needed to scan 1k - 2k rows, it's doing a full table scan. (And it just so happens that we don't have a ton of enterprise licenses floating around.) That isn't great, but definitely not a deal breaker.

But it may be enough to consider adding a logging statement or a trace span to the various queries we run. Just so that if something goes break down, it will be easier to pinpoint _that_ as the problem.

(But as far as trying to "do more with one epic query" vs. a more chatty series of queries, PostgreSQL can probably handle both situations just fine.)

cmd/frontend/dotcomproductsubscriptiontest/dotcomproductsubscriptiontest.go
bobheadxi: We're committed, I pinky-promise 😁

cmd/enterprise-portal/internal/dotcomdb/dotcomdb.go
bobheadxi: Core Services doesn't maintain public versions of our issues so Linear links are the only thing we have, as our work is generally internal-facing, even though we do prefer to be source-available per Sourcegraph tradition 😄

cmd/frontend/internal/dotcom/productsubscription/codygateway_graphql.go
bobheadxi: Not quite re-implementing, this just exposes a constructor for the same _existing_ implementation and logic for us to verify the new implementation is correct (and remains correct for the time being)

sg.config.yaml
bobheadxi: This file is pretty bad for reading but this is created under a single "service", so that you can do `sg run <service>`, and is the closest equivalent we have of "microservice" for local development.

For running larger sets of services, i.e. local `sg` clusters, that's the kinda poorly named `sg start <runset>`, which runs a _group of_ services. This PR doesn't touch any of the existing groups, so anyone starting up a local `sg` cluster with `sg start` default will not get Enterprise Portal (yet; maybe we'll want to add it to the default set later)

Hope that makes sense? `sg start -h` and `sg run -h` has a bit more docs; we used to have dev docs but that site has left us for the time being, pending a soon(tm) migration to Notion

cmd/enterprise-portal/internal/dotcomdb/dotcomdb.go
```diff
@@ -1,32 +1,250 @@
...
+
+	// Chat
+	if c.ChatCompletionsRateLimit != nil {
+		limits.ChatSource = codyaccessv1.CodyGatewayRateLimitSource_CODY_GATEWAY_RATE_LIMIT_SOURCE_OVERRIDE
+		limits.Chat.Limit = *c.ChatCompletionsRateLimit
+	}
+	if c.ChatCompletionsRateSeconds != nil {
+		limits.ChatSource = codyaccessv1.CodyGatewayRateLimitSource_CODY_GATEWAY_RATE_LIMIT_SOURCE_OVERRIDE
+		limits.Chat.IntervalSeconds = *c.ChatCompletionsRateSeconds
+	}
```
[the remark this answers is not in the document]
bobheadxi: Updated with:

```go
func maybeApplyOverride[T ~int32 | ~int64](limit *T, override *T) codyaccessv1.CodyGatewayRateLimitSource {
	if override != nil {
		*limit = *override
		return codyaccessv1.CodyGatewayRateLimitSource_CODY_GATEWAY_RATE_LIMIT_SOURCE_OVERRIDE
	}
	// no override
	return codyaccessv1.CodyGatewayRateLimitSource_CODY_GATEWAY_RATE_LIMIT_SOURCE_PLAN
}
```

cmd/enterprise-portal/internal/dotcomdb/dotcomdb.go
```diff
@@ -1,32 +1,250 @@
...
+	ps.cody_gateway_chat_rate_interval_seconds,
+	-- CodeCompletionsRateLimit override
+	ps.cody_gateway_code_rate_limit,
+	ps.cody_gateway_code_rate_interval_seconds,
+	-- EmbeddingsRateLimit override
+	ps.cody_gateway_embeddings_api_rate_limit,
+	ps.cody_gateway_embeddings_api_rate_interval_seconds,
+	-- "Active license"
+	MAX(active_license.license_tags),
+	MAX(active_license.license_user_count),
```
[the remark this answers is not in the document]
bobheadxi: `SELECT DISTINCT ON` returns exactly 1 row: https://www.postgresql.org/docs/current/sql-select.html (not sure how to link to section)

> SELECT DISTINCT ON ( expression [, ...] ) keeps only the first row of each set of rows where the given expressions evaluate to equal. The DISTINCT ON expressions are interpreted using the same rules as for ORDER BY (see above)

But, when we use `array_agg` + `GROUP BY`, Postgres complains that `active_license` now requires an aggregation as well, even though there's only one row

cmd/enterprise-portal/internal/dotcomdb/dotcomdb.go
bobheadxi: TIL that the `pgx` library actually doesn't return errors on `Close`! 

```
// Rows is the result set returned from *Conn.Query. Rows must be closed before
// the *Conn can be used again. Rows are closed by explicitly calling Close(),
// calling Next() until it returns false, or when a fatal error occurs.
//
// Once a Rows is closed the only methods that may be called are Close(), Err(),
// and CommandTag().
//
// Rows is an interface instead of a struct to allow tests to mock Query. However,
// adding a method to an interface is technically a breaking change. Because of this
// the Rows interface is partially excluded from semantic version requirements.
// Methods will not be removed or changed, but new methods may be added.
type Rows interface {
	// Close closes the rows, making the connection ready for use again. It is safe
	// to call Close after rows is already closed.
	Close()
```

bobheadxi: In this case, I think this is correct 🤔

lib/enterpriseportal/codyaccess/v1/codyaccess.proto
```diff
@@ -75,7 +75,7 @@ message CodyGatewayAccess {
   // Rate limit for code completions access, or null if not enabled.
   optional CodyGatewayRateLimit code_completions_rate_limit = 4;
   // Rate limit for embedding text chunks, or null if not enabled.
-  optional CodyGatewayRateLimit embeddings_rate_limt = 5;
+  optional CodyGatewayRateLimit embeddings_rate_limit = 5;
```
unknwon, line 78: 😂

unknwon (approved):

cmd/enterprise-portal/internal/dotcomdb/dotcomdb.go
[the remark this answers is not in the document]
bobheadxi: Updated with a docstring:

```
	-- "Active license": we aggregate for tokens below, so we need to apply MAX
	-- here to make this look like an aggregated value. This is okay becuase
	-- active_license uses 'SELECT DISTINCT ON' which returns exactly 1 row.
	MAX(active_license.license_tags),
	MAX(active_license.license_user_count),
```

lib/enterpriseportal/codyaccess/v1/codyaccess.proto
bobheadxi: Yeah, 190 comments on https://github.com/sourcegraph/sourcegraph/pull/62263 and I still missed this 😆

cmd/enterprise-portal/internal/dotcomdb/dotcomdb.go
```diff
@@ -1,32 +1,250 @@
...
+	) tokens ON tokens.product_subscription_id = ps.id
+`
+
+type GetCodyGatewayAccessAttributesOpts struct {
+	BySubscription *string
+	ByAccessToken  *string
+}
+
+func (r *Reader) GetCodyGatewayAccessAttributesBySubscription(ctx context.Context, subscriptionID string) (*CodyGatewayAccessAttributes, error) {
+	query := selectCodyGatewayAccessAttributes +
```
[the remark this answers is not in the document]
bobheadxi: I added a helper to more safely build the various query variants:

```go
type queryConditions struct {
	whereClause  string
	havingClause string
}

func newCodyGatewayAccessQuery(conds queryConditions) string {
```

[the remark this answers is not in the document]
bobheadxi: I think it reads a lot nicer now, thank you!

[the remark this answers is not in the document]
bobheadxi: > Just to avoid some unintentional reformat suddenly breaking things?

The testing that exists at the moment is sufficient to catch this I think as well!

cmd/enterprise-portal/internal/dotcomdb/dotcomdb.go
chrsmith: #TIL as well! Sorry for spewing nonsense PR comments.

This reminds me of that proverb: "if an error log falls in the forest, and, uh... no deferred Goroutine is around to catch it, does it make a sound... when reported to Sentry?

cmd/enterprise-portal/internal/dotcomdb/dotcomdb.go
```diff
@@ -1,32 +1,250 @@
...
+	array_agg(tokens.license_key_hash) as license_key_hashes
+FROM product_subscriptions ps
+	LEFT JOIN (
+		SELECT DISTINCT ON (product_subscription_id)
+			product_subscription_id,
+			license_tags,
+			license_user_count
+		FROM product_licenses
+		-- Get most recently created license key as the "active license"
+		ORDER BY product_subscription_id, created_at DESC
```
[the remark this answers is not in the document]
bobheadxi: Added the alias, and also updated all aliases to be more descriptive (`subscription`, `licenses`)

bobheadxi:
> But it may be enough to consider adding a logging statement or a trace span to the various queries we run. Just so that if something goes break down, it will be easier to pinpoint that as the problem.

I've added query and connect tracing hooks to the underlying MSP -> Cloud SQL connector util (side effect: after I land this PR, if you update `lib/managedservicesplatform` in SSC you'll get it too!)

cmd/enterprise-portal/internal/dotcomdb/dotcomdb.go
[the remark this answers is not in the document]
chrsmith: Thanks. As far as "general software engineering principles", here's a line from the Google Golang style guide that captures this sentiment well:

https://google.github.io/styleguide/go/decisions.html#variable-names

> The general rule of thumb is that the length of a name should be proportional to the size of its scope and inversely proportional to the number of times that it is used within that scope. A variable created at file scope may require multiple words, whereas a variable scoped to a single inner block may be a single word or even just a character or two, to keep the code clear and avoid extraneous information.

While applying this to a SQL query doesn't really make sense, just referring to fields from the same table name, and the query spanning many lines is an indication that we may want to use a more verbose identifier.

cmd/enterprise-portal/internal/dotcomdb/dotcomdb.go
bobheadxi: Thanks, I ended up removing the cheeky wording with something more formal and direct :)

bobheadxi:
Thank you for the feedback everyone! Merging for now; this code is still unused, so happy to address any post-merge feedback 😁

PR #62709, opened 2024-05-15
gitserver: Use git-diff-tree to simplify validation and improve security

eseliger:
At some point in the past, we noticed that passing /etc/passwd and /dev/null to `git diff` allows to view the file contents, even though it's not part of the repo itself.
That allows to read arbitrary files on disk, so we had to add in-app validation for file paths, including relative paths.
That makes the validation much more complex and harder to reason about, and requires that we didn't make a mistake.

Instead, we switch to `git diff-tree` which validates paths are part of the tree properly.
By that, we can remove a bunch of complexity around git diff argument validation.

See here for the difference:

```
➜  sourcegraph git:(main) git diff-tree --find-renames --full-index --inter-hunk-context=3 --no-prefix -- /etc/hosts /dev/null
fatal: /etc/hosts: '/etc/hosts' is outside repository at '/Users/erik/Code/sourcegraph/sourcegraph'
➜  sourcegraph git:(main) git --no-pager diff --find-renames --full-index --inter-hunk-context=3 --no-prefix -- /etc/hosts /dev/null
diff --git etc/hosts etc/hosts
deleted file mode 100644
index deadbeef..0000000000000000000000000000000000000000
--- etc/hosts
+++ /dev/null
@@ -1,16 +0,0 @@
-##
-# Host Database
-#
-# localhost is used to configure the loopback interface
-# when the system is booting.  Do not change this entry.
-##
-127.0.0.1	localhost
```

This PR also skips over arguments past the `--` boundary which in git means "everything after this will not be interpreted as an argument".
This previously caused `git cat-file HEAD -- -file.txt` to fail argument validation, although it's perfectly valid and not a risk.

Test plan:

Adjusted the tests for argument validation and existing tests for diff and integration and E2E tests are still passing.

### DIFF
--- a/cmd/gitserver/internal/git/gitcli/diff.go
+++ b/cmd/gitserver/internal/git/gitcli/diff.go
@@ -23,10 +23,17 @@ func (g *gitCLIBackend) RawDiff(ctx context.Context, base string, head string, t
 		return nil, err
 	}
 
-	return g.NewCommand(ctx, WithArguments(buildRawDiffArgs(baseOID, headOID, typ, paths)...))
+	// Since we pass down baseOID which is guaranteed to be an api.CommitID, this
+	// should not fail but lets guard against any changes in the future.
+	args, err := buildRawDiffArgs(baseOID, headOID, typ, paths)
+	if err != nil {
+		return nil, err
+	}
+
+	return g.NewCommand(ctx, WithArguments(args...))
 }
 
-func buildRawDiffArgs(base, head api.CommitID, typ git.GitDiffComparisonType, paths []string) []string {
+func buildRawDiffArgs(base, head api.CommitID, typ git.GitDiffComparisonType, paths []string) ([]string, error) {
 	var rangeType string
 	switch typ {
 	case git.GitDiffComparisonTypeIntersection:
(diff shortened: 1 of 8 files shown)


eseliger:
* **#62709** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62709?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

cmd/gitserver/internal/git/gitcli/diff.go
```diff
@@ -36,15 +43,23 @@ func buildRawDiffArgs(base, head api.CommitID, typ git.GitDiffComparisonType, pa
...
 
+	if err := checkSpecArgSafety(rangeSpec); err != nil {
+		return nil, err
+	}
+
 	return append([]string{
-		"diff",
+		// Note: We use git diff-tree instead of git diff because git diff lets
+		// you diff any arbitrary files on disk, which is a security risk, diffing
+		// /dev/passwd to /dev/null is crazy.
```
evict, line 53:
```suggestion
		// /etc/passwd to /dev/null is crazy.
```

cmd/gitserver/internal/git/gitcli/exec.go
```diff
@@ -26,7 +20,6 @@ var (
 		"log":       append([]string{}, gitCommonAllowlist...),
 		"show":      append([]string{}, gitCommonAllowlist...),
 		"remote":    {"-v"},
-		"diff":      append([]string{}, gitCommonAllowlist...),
```
evict, line 29: :fire:

evict (approved):
Very nice! :+1:

PR #62715, opened 2024-05-16
Add Cody Pro REST API client library

chrsmith:
Introduces an API client library for interacting with the Cody Pro-specific REST API.

For the Cody Pro-specific experience, there are a lot of imperative APIs that we need to make available to the frontend. (e.g. adding/removing team members, sending team invites, updating billing details, etc.)

The overhead of adding and maintaining the wiring for all of these APIs can easily get out of hand, very quickly. And yet, if we get any of the minor details wrong (such as the URL suffix for the API, or have a type definition that doesn't perfectly aline with the backend) things break down.

So in order for us to keep our sanity, and have a clear, process for hooking up new REST API endpoints as we add them to the backend, I've added some abstraction layers to cleanly separate the "API type definitions", the "API endpoints", and the "React-specific calling pattern".

All of which so that we can make API calls with a clear pattern:

```ts
const client = useApiClient()
const { loading, error, data } = useApiCaller(client.createStripeCheckoutSession(req))
```

And, relying on strong typing to provide intelligence and catch errors.

image

image

The specific implementation was inspired by the Apollo GraphQL React library, and its `useQuery` hook.

## Design goals

To help frame the implementation, let me first call out the design goals:

**Provide a clear place to put the API type definitions**

e.g. `api/types.ts` and related files like `api/teamSubscriptions.ts`. One of the most error prone parts of wiring up a new REST API is typos in the API type definition. (e.g. the Golang `json:"foo"` annotations not matching line up with the TypeScript `interface` fields.

If we wanted to invest a bunch of time here, we could look into a tool like OpenAPI to generate client libraries automatically. (But I've never heard a single good thing about using OpenAPI in production, and honestly don't think we'd need to do it.)

Just having an approximate 1:1 mapping between the TypeScript type definitions and the Golang counterparts will make it much easier to review that things are correct.

**Provide an easy way to encode endpoint metadata**

In order to call the right API endpoint, there are a few things that you need to supply: the HTTP method, the URL suffix, and the request body. Having the specific `('GET', '/team/subscription')` data sprinkled throughout the codebase isn't great. Since the caller of these API endpoints shouldn't need to care about that.

So a goal was to try to put all of that endpoint metadata in a single place (`api/client.ts`) So that you can define new methods with as little boilerplate as possible. e.g.

```ts
    createStripeCheckoutSession(
        requestBody: types.CreateCheckoutSessionRequest
    ): Call<types.CreateCheckoutSessionResponse> {
        return { method: 'POST', urlSuffix: '/checkout/session', requestBody }
    }
```

**Use React idioms**

Rather than replicating the same `useState()` calls for `{ loading, error, data }` on every page, and duplicate the same error handling logic, I wanted to put that into a single place so that we only need to write that code once.

## Implementation

So putting that all together, we have the following:

1. `class Client`

This is the thing that we update whenever we need to wire up a new API endpoint. Its methods provide the types for the request and response bodies as applicable. (Or may just be left as undefined, or `void` where applicable.)

```ts
getCurrentSubscription(): Call<types.Subscription> {
    return { method: 'GET', urlSuffix: '/team/current/subscription' }
}
```

The return type for these methods is `Call<ResponseBody>`. `Call<>` is just a description of a REST API call to make, but isn't meaningful on its own.

2. `export class CodyProApiCaller implements Caller`

This is what will actually make the API calls. It is just a wrapper around `fetch`:

```
async call<Resp>(call: Call<Resp>): Promise<{ data?: Resp; response: Response }>
```

3. The `useApiClient` and `useApiCaller` hooks

The final piece is how the React component can get ahold of the `Client` and pass any returned `Call<>` objects to a `Caller`.

```ts
// client defines the API surface area, and how API calls can get made.
const client = useApiClient()

// Create a Call<CreateCheckoutSessionResponse>.
const createCheckoutSessionCallObj = client.createStripeCheckoutSession(req)

// Create the React state and return the loading/error status of the API call,
// or a resolved `CreateCheckoutSessionResponse` object.
const { loading, error, data } = useApiCaller(createCheckoutSessionCallObj)
```

## Example 

To illustrate the steps involved with registering some new API endpoints to this client library, the second commit shows what is involved in hooking up our Subscription-related APIs.

1. Defining the new data types (`api/teamSubscriptions.ts`) and exporting them from `types.ts` for convenience.
2. Adding the new methods to the `Client` class in `client.ts`.

From there, we can call the new method using the `useApiClient` and `useApiCaller` hooks for a consistent pattern to make API calls with a consistent `{ loading, error, data }` signature.

See https://github.com/sourcegraph/sourcegraph/pull/62715/commits/78d79d6ad26a69f369c078f1019504fa87859f6f.

## Test plan

Manually tested. Going through the "New Cody Pro Subscription" flow worked like normal.


client/web/src/cody/management/api/hooks/useApiClient.tsx
```diff
@@ -0,0 +1,71 @@
...
+                }
+                setLoading(false)
+            } catch (err) {
+                setData(undefined)
+                setError(err)
+                setResponse(undefined)
+                setLoading(false)
+            }
+        })()
+    }, [caller])
```
taras-yemets, line 68: We must declare `call` as a `useEffect` hook dependency ([docs](https://react.dev/reference/react/useEffect#specifying-reactive-dependencies)). We also get a lint error on line 31: "This hook does not specify all of its dependencies: callbiomelint/correctness/useExhaustiveDependencies".

`call` value is re-calculated on each render and thus `Object.is(prevCall, call)` will always return `false`. For example, [`client.createStripeCheckoutSession(req)`](https://github.com/sourcegraph/sourcegraph/blob/78d79d6ad26a69f369c078f1019504fa87859f6f/client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx#L42) creates a new object on each render.
Including `call` as a hook dependency will have an undesired side-effect of `useEffect` being called on each render. 
To avoid it, we can move the `call` definition outside of the component (it's value is stable anyway).

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
```diff
@@ -14,99 +17,48 @@ export const CodyProCheckoutForm: React.FunctionComponent<{
...
+        // BUG: Due to the race conditions between Stripe, the SSC backend,
+        // and Sourcegraph.com, immediately loading the Dashboard page isn't
+        // going to show the right data reliably. We will need to instead show
+        // some intersitular or welcome prompt, to give various things to sync.
+        returnUrl: `${origin}/cody/manage?session_id={CHECKOUT_SESSION_ID}`,
     }
+
+    // Make the API call to create the Stripe Checkout session.
+    const client = useApiClient()
+    const { loading, error, data } = useApiCaller(client.createStripeCheckoutSession(req))
```
taras-yemets, line 42: `useApiCaller` argument doesn't depend either on component's state or props. As this argument should be included as a hook dependency (https://github.com/sourcegraph/sourcegraph/pull/62715#discussion_r1603095355), to avoid unnecessary hook calls, we should probably define argument value outside the component.

client/web/src/cody/management/api/hooks/useApiClient.tsx
```diff
@@ -0,0 +1,71 @@
...
+// useApiCaller will issue a REST API call to the backend, returning the
+// loading status and the response JSON object and/or error object as React
+// state.
+export function useApiCaller<Resp>(call: Call<Resp>): ReactFriendlyApiResponse<Resp> {
+    const { caller } = useContext(CodyProApiClientContext)
+
+    const [loading, setLoading] = useState(true)
+    const [error, setError] = useState<any>(undefined)
+    const [data, setData] = useState<any>(undefined)
+    const [response, setResponse] = useState<any>(undefined)
```
taras-yemets, line 29: The current approach work totally fine, I just want to mention the alternative one.

`loading`, `error`, `data` and `response` are not independent values IMO. E.g., we shouldn't have `data` and `error` defined and `isLoading` set to `true` at the same time (optimistic UI updates seem to be outside of the scope given the PR description). 

The way these values are updated looks to me like a good use case for `useReducer` hook. 
For example, we can maintain state as a single object:
```
interface State<T> {
    loading: boolean
    error: Error // or whatever type of an error we want
    data: T
    response: unknown // TODO: replace with actual type
}
```
and 4 actions that may change it (let's say "call_init", "call_error", "call_success"). [This](https://www.robinwieruch.de/react-hooks-fetch-data/) article has an example of a similar approach.

client/web/src/cody/management/api/hooks/useApiClient.tsx
```diff
@@ -0,0 +1,71 @@
...
+                    }
+                }
+                setLoading(false)
+            } catch (err) {
+                setData(undefined)
+                setError(err)
+                setResponse(undefined)
+                setLoading(false)
+            }
+        })()
```
taras-yemets, line 67: We should consider returning a cleanup function not to suffer from race conditions (note on `ignore` variable in the [docs](https://react.dev/reference/react/useEffect#fetching-data-with-effects)).

client/web/src/cody/management/api/hooks/useApiClient.tsx
```diff
@@ -0,0 +1,71 @@
...
+export function useApiClient(): Client {
+    const { client } = useContext(CodyProApiClientContext)
+    return client
+}
+
+export interface ReactFriendlyApiResponse<T> {
+    loading: boolean
+    error?: Error
+    data?: T
+    response?: Response
```
taras-yemets, line 17: Do we plan to use `response` in React components?

client/web/src/cody/management/api/client.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    getCurrentSubscription(): Call<types.Subscription> {
+        return { method: 'GET', urlSuffix: '/team/current/subscription' }
+    }
+
+    getCurrentSubscriptionSummary(): Call<types.SubscriptionSummary> {
+        return { method: 'GET', urlSuffix: '/team/current/subscription/summary' }
+    }
+
+    updateCurretSubscription(requestBody: types.UpdateSubscriptionRequest): Call<types.Subscription> {
```
taras-yemets, line 31:
```suggestion
    updateCurrentSubscription(requestBody: types.UpdateSubscriptionRequest): Call<types.Subscription> {
```

client/web/src/cody/management/api/components/CodyProApiClient.ts
```diff
@@ -0,0 +1,17 @@
...
+}
+
+// Context for supplying a Cody Pro API client to a React component tree.
+//
+// The default value will be a functional API client that makes HTTP requests
+// to the current Sourcegraph instance's backend.
+export const CodyProApiClientContext = createContext<CodyProApiClient>({
+    client: new Client(),
+    caller: new CodyProApiCaller(),
+})
```
taras-yemets, line 17: To be able to use context value we need to wrap a component tree with the context provider ([docs](https://react.dev/reference/react/createContext#returns)). I couldn't find this wrapper so I expect `useContext` calls will error out.

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
```diff
@@ -14,99 +17,48 @@ export const CodyProCheckoutForm: React.FunctionComponent<{
...
+        //
+        // BUG: Due to the race conditions between Stripe, the SSC backend,
+        // and Sourcegraph.com, immediately loading the Dashboard page isn't
+        // going to show the right data reliably. We will need to instead show
+        // some intersitular or welcome prompt, to give various things to sync.
+        returnUrl: `${origin}/cody/manage?session_id={CHECKOUT_SESSION_ID}`,
     }
+
+    // Make the API call to create the Stripe Checkout session.
+    const client = useApiClient()
```
taras-yemets, line 41: We need to call `useApiClient` to be able to use some [`Client`](https://github.com/sourcegraph/sourcegraph/blob/78d79d6ad26a69f369c078f1019504fa87859f6f/client/web/src/cody/management/api/client.ts#L20) method which result is used as `useApiCaller` argument. Given that `Client` is stateless and its methods are pure functions, I wonder what is the advantage of providing `Client` instance as a context value compared to making its methids `static` and using them directly, e.g.,
```
const { loading, error, data } = useApiCaller(Client.createStripeCheckoutSession(req))
```

client/web/src/cody/management/api/client.ts
```diff
@@ -0,0 +1,100 @@
...
+// the current Sourcegraph instance's SSC proxy API endpoint.
+export class CodyProApiCaller implements Caller {
+    // e.g. "https://sourcegraph.com"
+    private origin: string
+
+    constructor() {
+        this.origin = window.location.origin
+    }
+
+    async call<Resp>(call: Call<Resp>): Promise<{ data?: Resp; response: Response }> {
```
taras-yemets, line 70: nit: `Resp` sounds to me like a shorthand for `Response`. As we use it to type the response body and later we return `data` of type `Resp`, maybe `Data` or `Result` will be more descriptive, WDYT?

client/web/src/cody/management/api/client.ts
```diff
@@ -0,0 +1,100 @@
...
+        })
+
+        const rawBody = await fetchResponse.text()
+        if (fetchResponse.status >= 200 && fetchResponse.status <= 299) {
+            const typedResp = JSON.parse(rawBody) as Resp
+            return {
+                data: typedResp,
+                response: fetchResponse,
+            }
+        }
```
taras-yemets, line 90: We want need response text if the status code is not 2xx. Thus we can move reading response text inside the `if` block.

We are not sure that response text is a valid JSON string. We may need to wrap response text parsing with try/catch block:
```
if (fetchResponse.status >= 200 && fetchResponse.status <= 299) {
    try {
        // read response as JSON
        const typedResp = await fetchResponse.json() as Resp 
        return {
            data: typedResp,
            response: fetchResponse,
        }
    } catch (e) {
        // handle error
    }
}
```

client/web/src/cody/management/api/client.ts
```diff
@@ -0,0 +1,100 @@
+import * as types from './types'
+
+// Call is the bundle of data necessary for making an API request.
+// This is a sort of "meta request" in the same veign as the `gql`
+// template tag, see: https://github.com/apollographql/graphql-tag
+export interface Call<Resp> {
```
taras-yemets, line 6: nit: `Call` type definition and usage resemble a subset of `RequestInit` + resource path. I don't know if "Call" is a descriptive enough name. However, I don't have a better alternative.

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
```diff
@@ -14,99 +17,48 @@ export const CodyProCheckoutForm: React.FunctionComponent<{
...
 
-    const options /* unexported EmbeddedCheckoutProviderProps.options */ = {
-        clientSecret,
+        // URL the user is redirected to when the checkout process is complete.
+        //
+        // BUG: Due to the race conditions between Stripe, the SSC backend,
+        // and Sourcegraph.com, immediately loading the Dashboard page isn't
+        // going to show the right data reliably. We will need to instead show
+        // some intersitular or welcome prompt, to give various things to sync.
+        returnUrl: `${origin}/cody/manage?session_id={CHECKOUT_SESSION_ID}`,
```
taras-yemets, line 37: I couldn't find `CHECKOUT_SESSION_ID` definition? Where do we get it from?

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
taras-yemets: Ah, sorry, actually it depends on `customerEmail` prop and `showPromoCodeField` value. 
In this case, to avoid re-creating `req` on each render we should consider wrapping it with `useMemo()`.

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
```diff
@@ -14,99 +17,48 @@ export const CodyProCheckoutForm: React.FunctionComponent<{
...
                 </>
             )}
 
-            {clientSecret && (
-                <EmbeddedCheckoutProvider stripe={stripePromise} options={options}>
+            {data && data.clientSecret && (
+                <EmbeddedCheckoutProvider stripe={stripePromise} options={{ clientSecret: data.clientSecret }}>
                     <EmbeddedCheckout />
                 </EmbeddedCheckoutProvider>
             )}
```
taras-yemets, line 61: Can there be a case when we have both `error` and `data.clientSecret`?

taras-yemets (commented):
I agree with the motivation and overall approach! Great work, Chris!
Left a few comments about implementation details inline.

client/web/src/cody/management/api/client.ts
```diff
@@ -0,0 +1,100 @@
+import * as types from './types'
```
taras-yemets, line 1: We [use](https://sourcegraph.com/search?q=context:global+r:%5Egithub%5C.com/sourcegraph/sourcegraph$+%22import+type%22&patternType=keyword&sm=0) `import type` to allow compiler safely drop imports and not load these modules at runtime.

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
chrsmith: Stripe will automatically find/replace that value when the checkout is complete.
https://docs.stripe.com/payments/checkout/custom-success-page#modify-the-success-url

So the flow is:

1. We render the Stripe iframe, with the `clientSecret` of XXX which tells Stripe what the configuration settings are.
2. When the checkout is complete, Stripe will redirect the user to "returnUrl" but with the _actual_ CHECKOUT_SESSION_ID
3. On the page that _receives_ that page, we then use that `session_id` URL query parameter, to load the Checkout Session results from Stripe and see if the checkout was complete or not.

That third step isn't in the Sourcegraph.com code yet. But if you look in the SSC UI, that's what the `WelcomePageComponent` does.

client/web/src/cody/management/api/hooks/useApiClient.tsx
taras-yemets: > it's value is stable anyway

correcting myself here: not always stable, see https://github.com/sourcegraph/sourcegraph/pull/62715#discussion_r1603288864

chrsmith (commented):
Thanks for the detailed review @taras-yemets ! You definitely caught a lot of subtle issues that I was unaware of. Please take another look.

- Remove the need to putting a `Client` instance in the `CodyProApiClientContext`, and instead just use static methods.
- Updated the pattern to require `useMemo()` to avoid unintentionally resending the same API request.
- Returned a function object the `useEffect` hook within `useApiCaller`, so (hopefully) it will handle data races and things... I think?

So now the pattern looks something like:

```ts
// Make the API call to create the Stripe Checkout session.
const call = useMemo(() => client.createStripeCheckoutSession(req), [])
const { loading, error, data } = useApiCaller(call)
```

Things seem to be working OK when running locally, but a couple of things stand out:

**Is there some way we can avoid the need to memoize the `Call<T>`?**

I understand why it's important to have a stable reference to the `Call<T>`. I'm just wondering if we can somehow remove the use of `useMemo` and push the "memoize the return values" to the `Client` class and its static methods. (But doing so in some way that makes sense.)

I was thinking about something along the lines of:

```ts
// Lookup of known calls we have memoized. Keyed by:
// `${call.method}:${call.urlSuffix}`
let memoizedCalls = new Map<string, Call>()
function memoize(c: Call): Call {
    ...
}

    ...
    static updateCurrentSubscription(requestBody: types.UpdateSubscriptionRequest): Call<types.Subscription> {
        return { method: 'PATCH', urlSuffix: '/team/current/subscription', requestBody }
    }
```

However, when I thought more about it, we would need to maybe only want to memoize `GET` HTTP calls, since anything else wouldn't be idempotent? But this isn't about idempotency, but rather "state change management"...

And then I realized I should just ask you if you had any ideas :D 

**How can I test this thing?**

Given the importance of making sure we have clarity around when API calls get made, and that we are properly returning HTTP status codes and things, do you have any suggestions for how we can test this client library?

Should I just try to create a component specific for verifying lifecycle hooks, and a Storybook entry for that? Or is there an example in the repo where we have "unit tests" for React components, that deal with async data loading, verifying DOM elements, etc.?

vdavid:
@chrsmith, I mentioned in yesterday's call that we prefer modules to classes in our TypeScript codebase, but I didn't find our reasoning for that at the time. I did now: it's [here](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@5022b91a4343f4daf2674e9a5df03ecdeb8e4afb/-/blob/doc/dev/background-information/languages/typescript.md#avoiding-classes).
I don't see a good reason why this logic couldn't be implemented with TypeScript modules, so I suggest refactoring the classes into modules. Cody/Copilot could probably do 90% of the job. Let me know if you need help—or I can do it if needed.

client/web/src/cody/management/api/client.ts
vdavid: I agree that it's not great and that it's good enough. It's a pretty easy rename later, so let's just leave it for now.

client/web/src/cody/management/api/hooks/useApiClient.tsx
vdavid: I think what Taras means is not related to what the hook returns (the interface would remain the same), I think his suggestion was related to how the hook could store its state _internally_. It's no big deal, though, as the hook is currently pretty simple. I agree it might scale better later with `useReducer`—that said, I don't think I ever actually used `useReducer`, and we have two uses of it in the monorepo, so `useState` is usually sufficient.

client/web/src/cody/management/api/hooks/useApiClient.tsx
```diff
@@ -0,0 +1,85 @@
...
+// you MUST ensure that the provided call is the same between repains of the
+// calling React component. i.e. you pretty much always need to create it via
+// `useMemo()`.
+export function useApiCaller<Resp>(call: Call<Resp>): ReactFriendlyApiResponse<Resp> {
+    const { caller } = useContext(CodyProApiClientContext)
+
+    const [loading, setLoading] = useState(true)
+    const [error, setError] = useState<any>(undefined)
+    const [data, setData] = useState<any>(undefined)
+    const [response, setResponse] = useState<any>(undefined)
```
vdavid, line 27: You can definitely type these more strongly:

```suggestion
    const [error, setError] = useState<Error | undefined>(undefined)
    const [data, setData] = useState<Resp | undefined>(undefined)
    const [response, setResponse] = useState<Response | undefined>(undefined)
```

client/web/src/cody/management/api/hooks/useApiClient.tsx
vdavid: Ahh, this is neat, TIL, @taras-yemets. I think the fix is correct.

client/web/src/cody/management/api/teamSubscriptions.ts
```diff
@@ -0,0 +1,118 @@
...
+
+export interface UpdateSubscriptionRequest {
+    customerUpdate?: CustomerUpdateOptions
+    subscriptionUpdate?: SubscriptionUpdateOptions
+}
+
+export interface GetSubscriptionInvoicesResponse {
+    invoices: Invoice[]
+    continuationToken?: string
+}
```
vdavid, line 118: Thanks for creating all these types, @chrsmith. What's your plan to ensure that any engineer touching the back end will sync these types?

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
```diff
@@ -14,99 +18,47 @@ export const CodyProCheckoutForm: React.FunctionComponent<{
     stripePromise: Promise<Stripe | null>
     customerEmail: string | undefined
 }> = ({ stripePromise, customerEmail }) => {
-    const [clientSecret, setClientSecret] = useState('')
-    const [errorDetails, setErrorDetails] = useState('')
-    const [urlSearchParams] = useSearchParams()
-
-    // Optionally support the "showCouponCodeAtCheckout" URL query parameter, which is present
+    // Optionally support the "showCouponCodeAtCheckout" URL query parameter, which if present
```
vdavid, line 21: Idea to further avoid confusion

```suggestion
    // Optionally support the "showCouponCodeAtCheckout" URL query parameter, which, if present,
```

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
```diff
@@ -14,99 +18,47 @@ export const CodyProCheckoutForm: React.FunctionComponent<{
...
+        showPromoCodeField,
 
-    const options /* unexported EmbeddedCheckoutProviderProps.options */ = {
-        clientSecret,
+        // URL the user is redirected to when the checkout process is complete.
+        //
+        // BUG: Due to the race conditions between Stripe, the SSC backend,
+        // and Sourcegraph.com, immediately loading the Dashboard page isn't
+        // going to show the right data reliably. We will need to instead show
+        // some intersitular or welcome prompt, to give various things to sync.
```
vdavid, line 37: interstitial?

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
```diff
@@ -14,99 +18,47 @@ export const CodyProCheckoutForm: React.FunctionComponent<{
...
+        returnUrl: `${origin}/cody/manage?session_id={CHECKOUT_SESSION_ID}`,
```
vdavid, line 38:
```suggestion
        //
        // "{CHECKOUT_SESSION_ID}" is a magic string that Stripe will replace with the actual checkout session ID.
        returnUrl: `${origin}/cody/manage?session_id={CHECKOUT_SESSION_ID}`,
```

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
vdavid: I think it's good as-is.

client/web/src/cody/management/api/hooks/useApiClient.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+// useApiCaller will issue a REST API call to the backend, returning the
+// loading status and the response JSON object and/or error object as React
+// state.
+//
+// IMPORTANT: In order to avoid the same API request being made multiple times,
+// you MUST ensure that the provided call is the same between repains of the
+// calling React component. i.e. you pretty much always need to create it via
+// `useMemo()`.
+export function useApiCaller<Resp>(call: Call<Resp>): ReactFriendlyApiResponse<Resp> {
```
vdavid, line 21: Is `Resp` short for `ResponseType`? Could it be `CallType`, `ResponseType`, or just `T` as we usually do with TypeScript generics? I'd probably go with `T`

vdavid (approved):
Approving to unblock, but I suggest refactoring the classes to modules before merging. ([See](https://github.com/sourcegraph/sourcegraph/pull/62715#issuecomment-2117125135))

client/web/src/cody/management/api/hooks/useApiClient.tsx
taras-yemets: _Cleanup_ function should be returned by the hook `setup` function. 
```
useEffect(() => {
  let ignore = false  

  ;(() => {
    // fetch data and set state if `ignore` is still `false`, etc.
  })()

  // cleanup
  return () => {
    ignore = true
  }
}, [...dependencies])  
```

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
```diff
@@ -14,99 +18,47 @@ export const CodyProCheckoutForm: React.FunctionComponent<{
...
     }
+
+    // Make the API call to create the Stripe Checkout session.
+    const call = useMemo(() => Client.createStripeCheckoutSession(req), [])
```
taras-yemets, line 42:
```suggestion
    const call = useMemo(() => Client.createStripeCheckoutSession(req), [req])
```

Same as https://github.com/sourcegraph/sourcegraph/pull/62715#discussion_r1603095355: we must include `req` as a hook dependency.
`req` object is re-cretated on each render. So either we memoize it too, e.g.
```
const req = useMemo(() => ({
        // ... other fields
        customerEmail,
        showPromoCodeField,
    }), 
    [customerEmail, showPromoCodeField] // reactive dependencies - when they change we want `req` value to be recalculated
)
```
or create `req` inside a memoized `call`, e.g.,
```
const call = useMemo(() =>
    Client.createStripeCheckoutSession({
        // ... other fields
        customerEmail,
        showPromoCodeField,
    }),
    [customerEmail, showPromoCodeField]
) 
```

I prefer the latter as `req` is only used once and there's no additional value in definitin it as a separate `const`.

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
taras-yemets: My suggestion was to return early as we do for `loading`, e.g.
```
if (loading) {
    return <LoadingSpinner />
}

if (error) {
    return (
        <div>
            <H3>Awe snap!</H3>
            <Text>There was an error creating the checkout session: {error.message}</Text>
        </div>
    )
}

if (data?.clientSecret) {
    return (
        <div>
            <EmbeddedCheckoutProvider stripe={stripePromise} options={{ clientSecret: data.clientSecret }}>
                <EmbeddedCheckout />
            </EmbeddedCheckoutProvider>
        </div>
    )
}

return null
```

But if we are sure that we won't get truthy values for both `error` and `data.clientSecret`, it's good as-is.

client/web/src/cody/management/api/hooks/useApiClient.tsx
vdavid: Ah, true, I overlooked that the return was inside the async function. 🙈

client/web/src/cody/management/api/hooks/useApiClient.tsx
```diff
@@ -0,0 +1,85 @@
...
+                    } else {
+                        // For a 4xx or 5xx response this is where we provide any standardized logic for
+                        // error handling. For example:
+                        //
+                        // - On a 401 response, we need to force-logout the user so they can refresh their
+                        //   SAMS access token.
+                        // - On a 500 response, perhaps replace the current UI with a full-page error. e.g.
+                        //   http://github.com/500 or http://github.com/501
+                        setData(undefined)
+                        setError(Error(`unexpected status code: ${callerResponse.response.status}`))
```
taras-yemets, line 58: nit: `string` may be the sufficient type for the `error` field. An empty string means no error; non-empty may be presented to the user as-is.

client/web/src/cody/management/api/hooks/useApiClient.tsx
```diff
@@ -0,0 +1,85 @@
...
+            // the React runtime may invoke (setting ignore = true) outside of our view.
+            // https://react.dev/reference/react/useEffect#fetching-data-with-effects
+            // https://maxrozen.com/race-conditions-fetching-data-react-with-useeffect
+            let ignore = false
+
+            try {
+                console.debug(`calling API '${call.method} ${call.urlSuffix}'`)
+                const callerResponse = await caller.call(call)
+
+                if (!ignore) {
```
taras-yemets, line 42: nit: we can return early in case of `ingore` (this line) and after processing 2xx result (line 48).

client/web/src/cody/management/api/client.ts
```diff
@@ -0,0 +1,102 @@
...
+// Client provides the metadata for the methods exposed from the Cody Pro API client.
+//
+// This doesn't _do_ anything, it just returns the metadata for what needs to be done.
+// It is used in conjunction with a Caller implementation for actually fetching data.
+export class Client {
+    // Subscriptions
+
+    static getCurrentSubscription(): Call<types.Subscription> {
+        return { method: 'GET', urlSuffix: '/team/current/subscription' }
+    }
```
taras-yemets, line 25: > Re: I understand why it's important to have a stable reference to the Call<T>. I'm just wondering if we can somehow remove the use of useMemo and push the "memoize the return values" to the Client class and its static methods. (But doing so in some way that makes sense.) - https://github.com/sourcegraph/sourcegraph/pull/62715#pullrequestreview-2061728371


```
const currentSubscriptionArgs = {
  method: 'GET',
  urlSuffix: '/team/current/subscription'
}

class Client {
    static getCurrentSubscription(): Call<types.Subscription> {
        return currentSubscriptionArgs
    }
}
```

Now the reference to the `currentSubscriptionArgs` is stable and `args` inside `MyComponent` don't need a `useMemo` wrapper.
```
const MyComponent = () => {
  const args = Client.getCurrentSubscription()
}
```

taras-yemets: > However, when I thought more about it, we would need to maybe only want to memoize GET HTTP calls, since anything else wouldn't be idempotent? 

You're right, the above approach won't work for cases when `args` depend on some reactive value. In these case we would benefit from `useMemo` hook, e.g.:
```
const MyComponent = (props) => {
  const args = useMemo(() => Client.getSubscriptionById(id), [props.id])
}

**Comment by taras-yemets** [comment]
> Re: How can I test this thing?

@chrsmith, I think unit tests should be sufficient. We [use](https://sourcegraph.com/search?q=context:global+r:%5Egithub%5C.com/sourcegraph/sourcegraph$+renderhook&patternType=keyword&sm=0) the `renderHook` utility across the codebase ([docs](https://testing-library.com/docs/react-testing-library/api/#renderhook)).

File: client/web/src/cody/management/api/components/CodyProApiClient.ts (hunk 7 above)
Comment (taras-yemets) [reply client/web/src/cody/management/api/components/CodyProApiClient.ts]: If it works without a provider and we need only the default context value, maybe we can use them as create a class instance next to its declaration and export it? They are stable anyway.
```
export const client = new Client()
export const caller = new CodyProApiCaller()
```
Then, we can just import them into React components. What do you think?

Comment (taras-yemets) [reply client/web/src/cody/management/api/components/CodyProApiClient.ts]: Also, I wonder if we really need classes here 🤔 
Do they add any additional value compared to:
```
export const getSomeCodyProApiCallArgs = (param): Call<SomeType> => {
    // Not stable, will need to memorize in component.
    // Or, if `param` is not reactive, `getSomeCodyProApiCallArgs` can be called outside of component and be stable.
    return {
    // ... some fields
    param
  }
} 

export function callCodyProApi<T, R> = (call: Call<T>): Promise<{ data?: R; response: Response }> {
  // do stuff
}
```
?

Comment (chrsmith) [reply client/web/src/cody/management/api/components/CodyProApiClient.ts]: > Then, we can just import them into React components. What do you think?

I'm inclined to check what we have now in as-is, but I would be more than happy to refactor it as we find a better or clearer way to represent the data.

e.g. for the GET API calls, we could side-step the memoization requirement by just exporting a constant. (Since the is no request body.)

```ts
export const listCurrentTeamMembersCall = { ... }

export const getMemoizedUpdateSubscriptionSettingsCall(req: ReqBody) = {
     return useMemo(() => ..., [req])
}
```

That might be an improvement? I would want to wire up more API calls to get a better sense for what impact it would have on the ergonomics.

> Also, I wonder if we really need classes here 🤔

I'm not wed to using a class, but one thing that we would be more difficult to do, is having some additional piece of data associated with each call. (e.g. wiring in a "request ID" or some custom header.) Having all of the "get `Call<T>` objects" be routed through the same class instance makes that a lot easier. (But perhaps that's not a requirement we'd ever have.)

client/web/src/cody/management/api/hooks/useApiClient.tsx
chrsmith: Done.

client/web/src/cody/management/api/teamSubscriptions.ts
chrsmith: Great question. I think this is a situation where we will "add process as needed", rather than trying to formalize something up-front.

It shouldn't be too difficult to remember "if you add APIs to the backend, you probably want to add them to the frontend too". Since ultimately, we won't be able to _use_ those backend APIs unless we export them to the frontend.

So I don't believe this will be a major problem. But if we do find ourselves unintentionally making breaking changes, or not doing a good job keeping these in-sync, we will need to add some sort of unit/integration tests to ensure that we are intentional and know what we are doing when touching these files. (But to start with, let us try erring on the side of trust. #WCGW.)

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
chrsmith: Done.

client/web/src/cody/management/api/components/CodyProApiClient.ts
taras-yemets: This is a non-blocking comment. Feel free to leave as it is.

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
chrsmith: Added

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
chrsmith: Good catch! Fixed. And yeah, I agree the second format where we construct the request object too as part of the memoization process is probably the easier to understand.

client/web/src/cody/management/api/hooks/useApiClient.tsx
chrsmith: I changed the type to be `Error | undefined`, since otherwise we cannot say for certain what is being caught when we have the exception handler, right?

TBH I don't have a good sense for how to "properly" think about `Error` and runtime exceptions in JavaScript. So I'm happy to do whatever you think we should be doing 😅 .

client/web/src/cody/management/api/hooks/useApiClient.tsx
chrsmith: Done

taras-yemets (approved):
Great work! Thank you, Chris!

We can address the rest of comments (if we find them reasonable), add tests, etc. in the follow-up PRs.

chrsmith:
Thanks @vdavid, @taras-yemets  for the detailed code review! Y'all caught a lot of important defects in the code, and the end result seems a lot better.

- I addressed the feedback you had in the is last round, e.g. converting `class Client` into `module Client`
- Fixing the totally incorrect usage of that `loading` field in the hook set up function.
- Added a couple of unit tests to sanity check things.

I'll do some more manual testing on my machine, but then check this in. I'll then update the recently added Team management pages to use this to (🤞) simplify/clean up the code there. And also validate that this is a net improvement.

Issue #62713, opened 2024-05-16
Svelte: preserve repo revision in search bar


peterguy:
In the Svelte web app prototype, the revision of the repository is not preserved in the search bar. See attached screen recording for details.

To replicate this behavior:

1. search for `repo:github.com/sourcegraph/sourcegraph rev:at.time(1 month ago)` as a convenient way to get right to a particular revision.
2. click on the link to the repo at the revision.
3. See that the revision is in the address bar.
4. Type slash (`/`) (or click on the search bar) and see that the revision is missing from the search bar.

https://github.com/sourcegraph/sourcegraph/assets/129280/9b492afe-54e4-42aa-8faf-8bc279fbf873

taiyab:
@peterguy So we're only fixing this here and not in React? Meaning, until Svelte is shipped, no one gets a fix for this (if there is one)?

fkling:
@taiyab The React app includes the revision: 
![2024-05-16_13-22](https://github.com/sourcegraph/sourcegraph/assets/179026/404af422-63e5-43ff-bdad-f08b2a43e79d)

PR #62726, opened 2024-05-16
svelte: Fix various z-index/layering issues

fkling:
While looking into #62713 I stumbled upon two z-index issues:

I'm worried that layering already gets out of hand and is difficult to understand. So this commit removes a bunch of z-index settings to simplifying this.

Most importantly:

- Instead of setting a z-index on the header to make the sidebar navigation work, we can "portal" the sidebar to the end of the document.
- Likewise the repo search input is portalled to ensure that it renders above all other content.

## Test plan

Opened popovers/tooltips/etc on various pages:

- Main search input/suggestions
- Global sidebar navigation
- User menu
- `experimental` popover
- Code intel hovercards
- Repo search input
- Tooltips
- Revision picker

### DIFF
--- a/client/web-sveltekit/src/lib/navigation/GlobalHeader.svelte
+++ b/client/web-sveltekit/src/lib/navigation/GlobalHeader.svelte
@@ -150,11 +150,6 @@
         align-self: stretch;
         min-width: 0;
 
-        // Ensure that any content inside navigation portal block
-        // can't overlap any static content like sidebar navigation
-        // in the global header layout.
-        isolation: isolate;
-
         &-list {
             display: flex;
             gap: 1rem;
--- a/client/web-sveltekit/src/routes/[...repo=reporev]/RepoSearchInput.svelte
+++ b/client/web-sveltekit/src/routes/[...repo=reporev]/RepoSearchInput.svelte
@@ -14,7 +14,7 @@
     export let repoName: string
 
     const {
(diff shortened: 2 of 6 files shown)


client/web-sveltekit/src/lib/navigation/GlobalHeader.svelte
```diff
@@ -150,11 +150,6 @@
         align-self: stretch;
         min-width: 0;
 
-        // Ensure that any content inside navigation portal block
-        // can't overlap any static content like sidebar navigation
-        // in the global header layout.
-        isolation: isolate;
```
fkling, line 156: If this is set, the search input in the header is not above the main page content.

client/web-sveltekit/src/routes/[...repo=reporev]/RepoSearchInput.svelte
```diff
@@ -43,29 +43,25 @@
...
     .wrapper {
         flex: 1;
         position: absolute;
+        top: 1rem;
         left: 1rem;
         right: 1rem;
-        // This seems needed to prevent the file headers (which are position: sticky) from overlaying
-        // the search input. Alternatively we could portal the search input with melt, but then
-        // it would be more difficult to position it over the repo header.
-        z-index: 2;
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/RepoSearchInput.svelte:68]: This isn't relevant anymore since the header is now merged into the global navbar.

PR #62728, opened 2024-05-16
docs: Add link to Entitle permission

varungandhi-src:
For the reviewer: can you check if the link opens up the Form in the correct state for you?

(I used the 'Copy Form link' button.)

CleanShot 2024-05-16 at 21 01 03@2x

## Test plan

n/a


jtibshirani (approved):
I followed the link and it brought me to the correct permission.

PR #62732, opened 2024-05-16
appliance: fix test flakiness when container env vars are specified

craigfurman:
Since introducing the standard feature for container env vars in https://github.com/sourcegraph/sourcegraph/pull/62624, some test flakiness has been introduced. This is because container.go was iterating through an unsorted map, which made golden file assertions non-deterministic. This is fixed by sorting the map keys when appending user-provided env vars.

## Test plan

Ran `go test -count=10 -failfast ./internal/appliance` and it passed. Before this commit, 10 runs would reliably tease out the flake.

### DIFF
--- a/internal/k8s/resource/container/container.go
+++ b/internal/k8s/resource/container/container.go
@@ -1,6 +1,8 @@
 package container
 
 import (
+	"sort"
+
 	"github.com/grafana/regexp"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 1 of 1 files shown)


internal/k8s/resource/container/container.go
```diff
@@ -29,9 +31,7 @@ func NewContainer(name string, cfg config.StandardComponent, defaults config.Con
 
 	if cfg != nil {
 		if ctrConfig, ok := cfg.GetContainerConfig()[name]; ok {
-			for k, v := range ctrConfig.EnvVars {
-				ctr.Env = append(ctr.Env, corev1.EnvVar{Name: k, Value: v})
-			}
+			ctr.Env = append(ctr.Env, newSortedEnvVars(ctrConfig.EnvVars)...)
```
craigfurman, line 34: I thought about replacing this with something that generically returns sorted map entries from the maps package, but erred on the side of quick-and-simple here. We'd have to write a generic `sort.Interface` implementation to accomplish that, and since we're only ever operating on string keys in this package, `sort.Strings()` was sitting right there...

jdpleiness: I ran into something similar to this in the POC as well but handled it in the resource level: https://github.com/sourcegraph/sourcegraph-operator/blob/main/internal/controller/resource/container/container.go#L87

jdpleiness (approved):
Looks good, just comments 😄

PR #62739, opened 2024-05-16
Svelte: Remove underline from search results footer hover

taiyab:
Removes the hover underline from the help footer in the search results filter panel.

## Test plan
Tested locally for visual change.


client/web-sveltekit/src/lib/search/dynamicFilters/HelpFooter.svelte
```diff
@@ -63,6 +63,9 @@
             }
 
             padding: 1rem;
+            &:hover {
+                text-decoration: none;
+            }
```
camdencheek, line 68: Looks like we set `test-decoration: none` for both `:hover` and `:not(:hover)`. Can we just move it out a level to the `a` block?

camdencheek (approved):

client/web-sveltekit/src/lib/search/dynamicFilters/HelpFooter.svelte
taiyab: Done.

PR #62741, opened 2024-05-16
Svelte: Fix font-smoothing problem

taiyab:
Should fix font rendering issue.

## Test plan
Tested locally for visual change.

### DIFF
--- a/client/web-sveltekit/src/routes/svelte-overrides.scss
+++ b/client/web-sveltekit/src/routes/svelte-overrides.scss
@@ -17,6 +17,9 @@ body {
     --code-font-size: 13px;
     --border-radius: 4px;
 
+    -webkit-font-smoothing: antialiased;
+    -moz-osx-font-smoothing: grayscale;
+
     input::placeholder {
         color: var(--text-muted);
     }


camdencheek (commented):
Didn't we determine that the actual issue was that the font we're using by default (and which exists in Figma) is not installed by default on MacOS?

Overriding font smoothing seems to be [somewhat](https://ui.dev/rwd/articles/font-smoothing) [controversial](https://usabilitypost.com/2012/11/05/stop-fixing-font-smoothing/), so I'm hesitant to stamp this

taiyab:
> Didn't we determine that the actual issue was that the font we're using by default (and which exists in Figma) is not installed by default on MacOS?
> 
> Overriding font smoothing seems to be [somewhat](https://ui.dev/rwd/articles/font-smoothing) [controversial](https://usabilitypost.com/2012/11/05/stop-fixing-font-smoothing/), so I'm hesitant to stamp this

Sure, let's fix that then and reassess the font weights, etc.

camdencheek:
> let's fix that

Have we made any decisions on fonts?

PR #62744, opened 2024-05-16
Svelte: unrevert `RepoPopover`

camdencheek:
This re-applies https://github.com/sourcegraph/sourcegraph/pull/61989 after it was reverted. 

More detail inline, but in short:
- It reverts the changes to `Popover.svelte` that removed the border.
- We only start loading data on hover, not on mount
- Various fixes in text overflow conditions
- Removes the language from the popover data because it can be very expensive to calculate (another reason to pre-calculate language, but that's for another day)
- Moves the data loading out of the page loader. Exports the data loading function from the component so data loading is still orchestrated by the caller. (I know this will be controversial, reasoning inline)
- Adds a delay to the popover so it doesn't get in the way as your mouse moves over the page.
- Uses the display name instead of the author name
- Linkifies the commit message

# Test plan

Storybook and lots of manual testing. Now that I'm writing this, we should probably add a playwright test too.


client/web-sveltekit/src/lib/Popover.svelte
```diff
@@ -109,5 +109,6 @@
         border-radius: var(--popover-border-radius);
         color: var(--body-color);
         box-shadow: var(--popover-shadow);
+        overflow: hidden;
```
camdencheek, line 112: Small addition that ensures that the contents of the popover do not overflow the popover. Notably, this ensures that the border radius gets applied to any children as well.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.gql
```diff
@@ -1,31 +1,33 @@
...
+        ...RepoPopoverFragment
+    }
+}
+
+fragment RepoPopoverFragment on Repository {
     name
     description
     stars
     isPrivate
-    language
```
camdencheek, line 6: I removed language because this is calculated with inventory, which is stupidly expensive and flaky on large repos.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -1,184 +1,152 @@
...
-For example:
+    import { RepoPopoverQuery } from './RepoPopover.gql'
+
+    export async function fetchRepoPopoverData(client: GraphQLClient, repoName: string): Promise<RepoPopoverFragment> {
+        const response = await client.query(RepoPopoverQuery, { repoName })
+        if (!response.data?.repository || response.error) {
+            throw new Error(`Failed to fetch repo info: ${response.error}`)
+        }
+        return response.data.repository
+    }
```
camdencheek, line 12: I know this is controversial, but it was really pretty rough to thread this from page data, through every intermediate component, and finally into this component. 

I think this strikes a good balance by colocating the definition of the data fetcher with the component that uses it, but not fetching the data on mount, which allows the caller to handle data fetching however it wants. In this case, we do not want to fetch the data on page load, and we also want to add a delay before rendering. That would not be possible if we fetched the data on mount inside the component.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -1,184 +1,152 @@
...
-    $: avatar = repo.commit?.author.person
-    $: codeHostKind = repo.externalServices.nodes[0].kind
-    $: codeHostIcon = getIconPathForCodeHost(codeHostKind)
+    $: commit = data.commit
+    $: author = commit?.author
+    $: avatar = author?.person
+    $: codeHostKind = data.externalServices.nodes[0].kind
 </script>
 
 <div class="root">
```
camdencheek, line 39: In the component itself, I made a number of DOM and CSS simplifications, most of which did not affect rendering, a few of which fixed rendering for weird edge cases like empty description or long commit message.

client/web-sveltekit/src/lib/utils/promises.ts
```diff
@@ -97,3 +97,11 @@ export function toReadable<D, E = Error>(promise: PromiseLike<D>): Readable<Load
...
     return { subscribe }
 }
+
+// Returns a promise that is guaranteed to take at least `delayMillis` milliseconds to resolve.
+// If the wrapped promise resolves before then, the returned promise will wait until `delayMillis`
+// has elapsed before resolving.
+export async function delay<T>(promise: Promise<T>, delayMillis: number): Promise<T> {
+    const [awaited] = await Promise.all([promise, new Promise(resolve => setTimeout(resolve, delayMillis))])
+    return awaited
+}
```
camdencheek, line 107: Small utility to delay resolution of a promise. We may want to add this as a general option on the Popover component, but I wanted to see how it feels first.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.gql
```diff
@@ -1,31 +1,32 @@
...
-                ...Avatar_Person
+            oid
+            subject
+            canonicalURL
+            author {
+                date
+                    person {
+                        ...Avatar_Person
+                    }
             }
```
fkling, line 30: This seems to be weirdly indented.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
fkling: > but it was really pretty rough to thread this from page data, through every intermediate component, and finally into this component.

I suspected that this would happen eventually, and was hoping that when that time comes we know how we want to approach this.
I actually thought about this recently. I think there is kind of a forth type of data: on demand data that is unrelated (or only loosely related) to the current URL. For this kind of data it doesn't make sense to use the data loader. We will never preload the data and it's likely that this data (components) is needed (used) on multiple pages.
I arrived at a similar solution: These components would define a "loader" that handles the data loading for them and they would get passed an instance of the loader. This way we can still easily test the component and we are separating concerns. This was with more complex data loading in mind, e.g. infinity scroll.
In a case like this it seems fine to just have a function returning the results and pass those to the component.

So yeah, I think this approach is perfectly fine and we should use similar patterns for similar components.

Side note: The whole "only fetch data in data loader" principle was never meant to be absolute. It should help us understand what kind of data we are working with and determine appropriate exceptions to this rule.

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -1,184 +1,152 @@
...
-    $: url = repo.commit?.canonicalURL
-    $: commitSHA = repo.commit?.oid
-    $: author = repo.commit?.author.person.name
-    $: commitDate = repo.commit?.author.date
-    $: avatar = repo.commit?.author.person
-    $: codeHostKind = repo.externalServices.nodes[0].kind
-    $: codeHostIcon = getIconPathForCodeHost(codeHostKind)
+    $: commit = data.commit
+    $: author = commit?.author
+    $: avatar = author?.person
```
fkling, line 35: I don't think you need to extract this (see below).
```suggestion
```

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
```diff
@@ -1,184 +1,152 @@
...
-                    <div class="commit-number">
-                        <small class="commit-number"
-                            ><a href={url} target="_blank">#{truncateCommitNumber(commitSHA, 6)}</a></small
-                        >
+            <div class="commit-info">
+                <small class="subject"><a href={commit.canonicalURL}>{commit.subject}</a></small>
+                {#if author && avatar}
+                    <div class="author">
+                        <Avatar {avatar} --avatar-size="1.0rem" />
+                        <small>{avatar.displayName} · <Timestamp date={author?.date} /></small>
```
fkling, line 80: I think this should work (i.e. the `author?.avatar` check will narrow down the type to `author` and `avatar` being defined). I don't see any advantage of extracting `avatar` here.
```suggestion
                {#if author?.avatar}
                    <div class="author">
                        <Avatar avatar={author.avatar} --avatar-size="1.0rem" />
                        <small>{author.avatar.displayName} · <Timestamp date={author.date} /></small>
```

fkling (approved):
Thank you!

client/web-sveltekit/src/lib/repo/RepoPopover/RepoPopover.svelte
camdencheek: This all makes sense to me! Thanks for adding some color 🙂

Issue #62676, opened 2024-05-14
Username normalization breaks repo permissions


gabtorre:
- **Sourcegraph version:** 5.3.12303 
- **Platform information:** 

We encountered an issue where usernames with suffixes are breaking permissions syncs with Bitbucket, as Bitbucket requires exact username matches. Sourcegraph normalizes usernames by adding a [suffix](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@058f831423994c6544605114505b1fd21d7796c3/-/blob/internal/auth/userpasswd/handlers.go?L553-555) if the username is in email format. This discrepancy affects setups where SCIM and SAML are involved and the username has to be set as an email.

#### Steps to reproduce:

1. Set up SCIM and SAML with Okta.
2. Create a user and sync permissions with Bitbucket.

#### Expected behavior:
Usernames should match exactly between Sourcegraph and Bitbucket without normalization issues.

#### Actual behavior:
Usernames in Sourcegraph get a suffix added during normalization, causing mismatches with Bitbucket usernames.

If you would like immediate help on this, please email support@sourcegraph.com (you can still create the issue, but there are no [SLAs](https://sourcegraph.com/support/) on issues like there are for support requests).

noah-berman:
This is affecting Citadel after a recent upgrade (to 5.3.11625), as well

PR #62747, opened 2024-05-16
auth: Make username from email deduplication dotcom only

eseliger:
https://github.com/sourcegraph/sourcegraph/pull/60339 introduced a new behavior that would always add a random suffix to usernames that are derived from email addresses.
The reason behind it seems to be that on dotcom, emails are from various domains and there are a lot of duplicates usually, and a lot of names that trigger our suspiciousness detection, like me@sourcegraph.com or me@sgdev.org.

For customer instances, this is not a real risk, and because of historic reasons, we depend on the username to be exactly what was given by the auth provider, so that bitbucket server permission syncing can work.
So this change for customer instances caused https://github.com/sourcegraph/sourcegraph/issues/62676.

To simplify the fix for this and make sure it can land in the next patch release, I opted for making this dotcom specific for now, as that's the least invasive change we can make.

Closes https://github.com/sourcegraph/sourcegraph/issues/62676

Test plan:

Created a new SAML application in Okta dev and lined that to my local instance.
A new user for `erik@sourcegraph.com` on main was called `erik-acbdef` after signup.
On this branch, with everything reset to fresh, a new user for `erik@sourcegraph.com` is called `erik`.
This should fix the bitbucket perms syncing issue.

### DIFF
--- a/internal/auth/userpasswd/handlers.go
+++ b/internal/auth/userpasswd/handlers.go
@@ -13,6 +13,7 @@ import (
 	"github.com/gorilla/mux"
 	"github.com/sourcegraph/log"
 
+	"github.com/sourcegraph/sourcegraph/internal/dotcom"
 	"github.com/sourcegraph/sourcegraph/internal/lazyregexp"
 	"github.com/sourcegraph/sourcegraph/internal/telemetry"
 	"github.com/sourcegraph/sourcegraph/internal/telemetry/teestore"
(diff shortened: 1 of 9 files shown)


eseliger:
* **#62747** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62747?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

internal/auth/userpasswd/handlers.go
```diff
@@ -557,13 +558,17 @@ func NormalizeUsername(name string) (string, error) {
...
+
 		// NOTE: When we derive the username from the email address, it is high chance
-		// that the username is not unique, so we always append a random suffix to the
-		// username.
-		var err error
-		name, err = AddRandomSuffix(name[:i])
-		if err != nil {
-			return "", errors.Wrap(err, "add random suffix")
+		// that the username is not unique on dotcom, because many emails, are of formats
+		// like me@XX.com. So we always append a random suffix to the username.
```
bobheadxi, line 565:
```suggestion
		// like me@XX.com. So we always append a random suffix to the username in dotcom.
```

bobheadxi (approved):
Test cases LGTM

bobheadxi:
> For customer instances, this is not a real risk, and because of historic reasons, we depend on the username to be exactly what was given by the auth provider, so that bitbucket server permission syncing can work.

Is this the case for any of the other permissions sources? Should we document the dependency in the docstring of `NormalizeUsername`? Should this even be a `// 🚨 SECURITY: `-worthy docstring, as it has implications on permissions enforcement?

eseliger:
> Is this the case for any of the other permissions sources? Should we document the dependency in the docstring of NormalizeUsername? Should this even be a // 🚨 SECURITY: -worthy docstring, as it has implications on permissions enforcement?

I think so, bitbucket server is the only one I'm aware of that has this restriction. They didn't have oauth support in the past, but they added it 1.5y ago - it's somewhere on our roadmap to finally switch to that and get rid of this hack with matching usernames.

If this is security worthy: Maybe? I can add a quick sentence about bitbucket but not sure if this should trigger an alert for the security team.

eseliger:
Added that comment

mmanela (approved):

eseliger:
I will merge this to make sure we get it in for Mondays patch release in time, but happy to follow up on further comments.

unknwon (commented):
Thank you!

PR #62754, opened 2024-05-17
PLG: migrate subscription management page

taras-yemets:
This PR migrates the [Cody subscription management page](https://github.com/sourcegraph/self-serve-cody/blob/main/cody/web/src/subscription/SubscriptionPage.tsx) from SSC to the `/cody/subscription/manage` path on dotcom. The page is available only if the [embedded Cody Pro UI is enabled](https://github.com/sourcegraph/sourcegraph/blob/be5fc7e6df70707db24d11ac81b0ed0a274c8e0e/client/web/src/cody/util.ts#L16).

Fixes https://github.com/sourcegraph/self-serve-cody/issues/513

[2-min Loom](https://www.loom.com/share/de07686cb64541cc81f34327f93dd2cb?sid=a10bdae0-1cc4-4702-8b71-39ca03b5ceba)

### Implementation Details:
1. Added a `refetch` function to the `useApiCaller` hook to allow resource refetching after mutations.
2. Changed `Date` types to `string` in the `Invoice` and `Subscription` interfaces, partially addressing [this discussion](https://github.com/sourcegraph/sourcegraph/pull/62797#discussion_r1607997956).

### UI Updates:
The UI was implemented based on [Figma designs](https://www.figma.com/design/FMSdn1oKccJRHQPgf7053o/Cody-PLG-GA?node-id=3496-5821&t=oqVSaYQbV9cO3a0u-0) with some adjustments (cc: @rrhyne):
  1. **Button Styling**: "Cancel subscription" and other secondary buttons now use `@sourcegraph/wilcard` components, resulting in some style differences (e.g., background color) from the Figma design.
  2. **Stripe Form Inputs**: Styled to match the `@sourcegraph/wilcard` form fields.
  3. **Async Action Buttons**: Buttons triggering asynchronous actions (cancel subscription, renew subscription, and save changes) now include icons (cancel, refresh, and check, respectively). During async operations, these icons are replaced by loading spinners. Since Figma doesn't specify loading states for these buttons/forms, submit buttons are disabled and display loading indicators when actions are in progress.
  
  <details>
  
  <summary><b>Screenshots</b></summary>
  
  
| Description | Light | Dark |
|--|--|--|
| Active subscription |Screenshot 2024-05-24 at 17 02 52Screenshot 2024-05-24 at 17 03 05Screenshot 2024-05-24 at 17 03 12Screenshot 2024-05-24 at 17 03 16Screenshot 2024-05-24 at 17 03 26|Screenshot 2024-05-24 at 17 04 41Screenshot 2024-05-24 at 17 04 58|
Canceled subscription |Screenshot 2024-05-24 at 17 03 48|Screenshot 2024-05-24 at 17 04 33|

</details>

    

## Test plan
- CI
- Tested manually
  - Run Sourcegraph in the enterprise mode (`sg start`)
    - navigate to `/cody/subscription/manage` - page should not be rendered
  - Run Sourcegraph in dotcom mode (`sg start dotcom`)  
    - Navigate to `/cody/manage` page
      - user is a subscription admin
        - "Manage subscription" link is rendered
        - Link click navigates to `/cody/subscription/manage` route
        - Ensure Subscription management page is rendered
        - Ensure user can change payment method
        - Ensure user can change billing address
        - Ensure user can cancel and renew subscription
        - Ensure invoices are listed and can be downloaded
      - user is not admin
        - "Manage subscription" link is not rendered
        - navigate to `/cody/subscription/manage` by changing the URL in browser
        - Ensure Subscription management page is not rendered and user is redirected to the New subscription page


client/web/src/cody/management/CodyManagementPage.tsx
```diff
@@ -133,11 +132,9 @@ export const CodyManagementPage: React.FunctionComponent<CodyManagementPageProps
                                 <ButtonLink
                                     variant="primary"
                                     size="sm"
-                                    href={manageSubscriptionRedirectURL}
-                                    onClick={event => {
-                                        event.preventDefault()
+                                    to="/cody/subscription/manage"
+                                    onClick={() => {
```
vdavid, line 136: That is more concise indeed, thanks. 👍

client/web/src/cody/management/api/hooks/useApiClient.tsx
```diff
@@ -26,62 +27,47 @@ export function useApiCaller<Resp>(call: Call<Resp>): ReactFriendlyApiResponse<R
...
-                    // - On a 401 response, we need to force-logout the user so they can refresh their
-                    //   SAMS access token.
-                    // - On a 500 response, perhaps replace the current UI with a full-page error. e.g.
-                    //   http://github.com/500 or http://github.com/501
-                    setData(undefined)
-                    setError(new Error(`unexpected status code: ${callerResponse.response.status}`))
-                    setResponse(callerResponse.response)
+            // If we received a 200 response, all is well. We can just return
+            // the unmarshalled JSON response object as-is.
+            if (callerResponse.response.status >= 200 && callerResponse.response.status <= 299) {
```
vdavid, line 37: I guess this would work here as well:
```suggestion
            if (callerResponse.response.ok) {
```

client/web/src/cody/management/subscription/manage/CodySubscriptionManagePage.tsx
```diff
@@ -0,0 +1,250 @@
...
+        return null
+    }
+
+    // This page only applies to users who have a Cody Pro subscription to manage.
+    // Otherwise, direct them to the ./new page to sign up.
+    if (subscriptionData.plan !== CodySubscriptionPlan.PRO) {
+        return <Navigate to="/cody/manage/subscription/new" replace={true} />
+    }
+
+    return <PageContent />
```
vdavid, line 69: I like how the edge-cases are handled here before fetching the subscription. Maybe leave a comment here or on `PageContent` that it's separated to avoid a costly API call if we don't need it? Otherwise, someone might be inclined to merge the two because none of the two components are that large.

client/web/src/cody/management/subscription/manage/CodySubscriptionManagePage.tsx
```diff
@@ -0,0 +1,250 @@
...
+        <title>Back icon</title>
+        <path
+            d="M6.49967 12.8667L7.43967 11.9267L4.38634 8.86666H15.1663V7.53333H4.38634L7.44634 4.47333L6.49967 3.53333L1.83301 8.19999L6.49967 12.8667Z"
+            fill="#0B70DB"
+        />
+    </svg>
+)
+
+/* eslint-disable react/forbid-dom-props */
+const CodyIcon: React.FC<{ className?: string }> = ({ className }) => (
```
vdavid, line 151: Would [this](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/client/web/src/cody/components/CodyIcon.tsx?L9) work instead of inlining the svg in this component?

If so, maybe also extract BackIcon to its own file next to CodyIcon.tsx for better discoverability?

client/web/src/cody/management/api/hooks/useApiClient.tsx
```diff
@@ -26,62 +27,47 @@ export function useApiCaller<Resp>(call: Call<Resp>): ReactFriendlyApiResponse<R
...
-        // the React runtime may invoke (setting ignore = true) outside of our view.
-        // https://react.dev/reference/react/useEffect#fetching-data-with-effects
-        // https://maxrozen.com/race-conditions-fetching-data-react-with-useeffect
-        let ignore = false
-
-        ;(async () => {
-            try {
-                const callerResponse = await caller.call(call)
+    const callApi = useCallback(async () => {
+        setLoading(false)
```
vdavid, line 31: Do you think the `ignore` thing was unnecessary in the first place? If it was necessary, what change made it unnecessary? (This is a genuine question on my side, I don't have the answer. I think I hadn't seen this `ignore` pattern in React before and when I saw it here, I thought "there must be a more concise solution than this" but I couldn't find a better one. I'm curious if you thought about it and what the solution is as you see it.)

client/web/src/cody/management/subscription/manage/InvoiceHistory.module.scss
```diff
@@ -0,0 +1,34 @@
...
+}
+
+.invoice {
+    margin-top: 1rem;
+
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+
+    &-col {
```
vdavid, line 13: TIL this [works](https://css-tricks.com/the-sass-ampersand/#aa-modifying-the-ampersand). I thought only `&.something` was allowed, never tried it without the dot. Thanks for teaching me this :)

vdavid (approved):
TBH, I didn't do a deep-review on the large new components because they didn't have a diff with the old ones (as they are in another repo), but I did skim through them and nothing stood out as wrong.

Great job overall, this must've been a lot of work with many small details to get right. Thanks for the small improvements everywhere!

client/web/src/cody/management/subscription/manage/CodySubscriptionManagePage.tsx
taras-yemets: > Would [this](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/client/web/src/cody/components/CodyIcon.tsx?L9) work instead of inlining the svg in this component?

No, it appeared to be a different one.

> maybe also extract BackIcon to its own file next to CodyIcon.tsx for better discoverability?

It's used in one place so I colocated definition and usage. If we are going to reuse it, we should move it to a separate location.

client/web/src/cody/management/api/hooks/useApiClient.tsx
taras-yemets: I thought about the current approach (custom data fetching hook) as I was migrating the Manage Subscription page.

It works without `ignore`, but is not reliable for this and a bunch of other reasons:
1. Re-fetching After Mutations: Multiple components may mutate a resource and then trigger a re-fetch. The current approach might not handle all edge cases and side effects effectively.
2. Resource Sharing Across Components: When multiple components need the same resource, we either:
  - Call useApiCaller() in each component, resulting in multiple API calls. Although caching can mitigate this, implementing it reliably can be complex.
  - Prop-drill the data (as done in this PR), which is not scalable either.

I recommend using a third-party library to manage server state. I've previously used [React Query](https://tanstack.com/query/latest/docs/framework/react/overview), which addresses these issues and many others out of the box. React Query's useQuery and useMutation hooks are similar to Apollo's, making them familiar to our team. 

I plan to create a PR demonstrating the difference in approaches using React Query.

taras-yemets: @vdavid, @chrsmith, please let me know what you think on https://github.com/sourcegraph/sourcegraph/pull/62939

client/web/src/cody/management/api/hooks/useApiClient.test.tsx
```diff
@@ -1,138 +1,116 @@
...
 
 import { useApiCaller } from './useApiClient'
 
-// FakeCaller is a testing fake for the Caller interface, for simulating
-// making API calls. Only supports one call being made at a time, otherwise
-// will fail.
-//
-// It's hard to do async hook testing correctly. This might be helpful:
-// https://react-hooks-testing-library.com/usage/advanced-hooks#async
-class FakeCaller implements Caller {
```
taras-yemets, line 17: Refactored to use `vitest` mocking utilities instead of custom ones. 
cc: @chrsmith

rrhyne:
Thank you @taras-yemets! I've created a QA page with some small changes here:

[QA notes in Figma](https://www.figma.com/design/FMSdn1oKccJRHQPgf7053o/Cody-PLG-GA?node-id=5187-12468)

I haven't addressed a few things in the old design light address display to keep the scope low. 

One QQ though, how difficult would it be to add a remove credit card button? If it adds all kinds of hidden scope, forget it, but if it's a small API call and the system handles the ramifications, it would help establish more trust with users.

vdavid:
@taras-yemets, I think the back end (our DB, Go code, and Stripe) can handle it, but the front end as a whole must be prepared for the related use case of re-adding a payment method.

Issue #62578, opened 2024-05-09
Code insights: Extend GraphQL schema to expose repositories that cause incomplete datapoints


bahrmichael:
https://github.com/sourcegraph/sourcegraph/issues/62295 surfaced problems with debugging timeout and generic problems that lead to incomplete data points. The documentation says that the user should reduce the scope of the query, but we see that customers prefer to run queries over all of their repositories. I'm not sure how we would bisect repositories to exclude the problematic ones.

Our GraphQL schema for `TimeoutDatapointAlert` and `GenericIncompleteDatapointAlert` contains the time and maybe a generic reason that something went wrong. 

We could help users that don't have access to logs by exposing the repository/repositories that caused problems.

Example:

```
type GenericIncompleteDatapointAlert implements IncompleteDatapointAlert {
    """
    The data point that is incomplete.
    """
    time: DateTime!

    """
    A message describing why the datapoint was marked incomplete.
    """
    reason: String!

    """
    The repositories that this data point is incomplete for.
    """
    repositories: [String!]!
}
```

As a next step we could surface this information in our UI, but including this info in our GraphQL should be a good start.

camdencheek:
Seems useful! 

One question: can't we have multiple reasons? In particular, wouldn't there potentially be a different reason per incomplete repository?

bahrmichael:
> wouldn't there potentially be a different reason per incomplete repository?

In theory yes. Currently we have only timeout and generic as reasons. From what I can see the code encourages adding new types when there are errors that we want to distinguish from each other.

https://github.com/sourcegraph/sourcegraph/blob/d7dc2b9d9a1e7c44c5de59617b0dae4c7d969cee/cmd/frontend/internal/insights/resolvers/insight_series_resolver.go#L554-L559

https://github.com/sourcegraph/sourcegraph/blob/d7dc2b9d9a1e7c44c5de59617b0dae4c7d969cee/internal/insights/store/store.go#L875-L879

I feel like changing the shape of the incomplete datapoints and its sub-fields can become a rabbit-hole and would avoid going this route unless necessary.

An alternative to satisfy the need to to isolate out repos could be an additional field next to the incomplete datapoints that just has a list of repos with problems.

```gql
"""
Status indicators for a specific series of insight data.
"""
type InsightSeriesStatus {

    ...

    """
    Data points that are flagged terminally incomplete for this series.
    """
    incompleteDatapoints: [IncompleteDatapointAlert!]!

[...]

Issue #62295, opened 2024-04-30
Sentry: Code Insights results may be incomplete


bahrmichael:
This ticket originated from a [Slack message](https://sourcegraph.slack.com/archives/C05EA9KQUTA/p1713821754294589), reported by @mike-r-mclaughlin.

> We’ve been struggling to get Code Insights to work at Sentry.  the screenshot below is what we see in the UI.  From the screenshot it seems like we hit a timeout.  Which timeout is it referring to and can that be configured?

![image](https://github.com/sourcegraph/sourcegraph/assets/1830132/a8896547-1ff6-463d-9b60-2b192e1c724b)

> It’s not entirely clear which log to check. we checked searcher (no warnings/errors during the time period this was being generated) and gitserver (also no seemingly relevant warning/error messages).  we did find some relevant errors/warnings in frontend 

```
{
  "SeverityText": "ERROR",
  "Timestamp": 1712170603579550779,
  "InstrumentationScope": "Resolver.live_preview_resolver",
  "Caller": "query/streaming_query_executor.go:154",
  "Function": "github.com/sourcegraph/sourcegraph/internal/insights/query.(*StreamingRepoQueryExecutor).ExecuteRepoList",
  "Body": "repo search encountered skipped events",
  "Resource": {
    "service.name": "frontend",
    "service.version": "5.2.7",
    "service.instance.id": "da92ab9c4de7"
  },
  "Attributes": {
    "reasons": "[shard-match-limit: Not all results have been returned due to hitting a match limit. Sourcegraph has limits for the number of results returned from a line, document and repository.]",
    "query": "fork:yes archived:yes repo:.* count:21"
  }
}
```

Logs: https://sourcegraph.slack.com/files/U01H4LDHVMF/F0703Q1EU2X/sourcegraph-frontend-0_--tail_100.txt

bahrmichael:
This comment is WIP.

---

Given that we're struggling to set it up, I assume that the datapoints come from the backfill process.

I'll document my findings here, and will later pull them into some docs.

---

[...]

PR #45282, opened 2022-12-06
insights: surface error information from global queries that are terminally failed

coury-clark:
Closes #44371

Write an incomplete row for global queries that terminally fail.

## Test plan

Tested a query with `timeout:2s`

Here is the queue row after it moved to the failed state:
| state | finished\_at | num\_failures |
| :--- | :--- | :--- |
| failed | 2022-12-06 18:26:29.840695 +00:00 | 10 |

Here is the incomplete row that was written
| id | series\_id | reason | time | repo\_id |
| :--- | :--- | :--- | :--- | :--- |
| 162 | 46 | timeout | 2022-12-06 18:26:27.808281 | null |

PR #62756, opened 2024-05-17
feat: incomplete datapoints can now resolve the affected repositories

bahrmichael:
Closes https://github.com/sourcegraph/sourcegraph/issues/62578

For https://github.com/sourcegraph/sourcegraph/issues/62295

Previously with our GraphQL api, you couldn't figure out which repositories caused incomplete datapoints. With this change you can now provide an argument to the `incompleteDatapoints` to not aggregate points for repositories, and then resolve the repositories for each datapoint.

This PR is needed to help debug incomplete datapoints in Code Insights. When customers create Code Insights for a large number of repositories, it's hard to understand how big the impact of incomplete datapoints is, and which repositories those issues are coming from. If you don't have access to the logs it's basically impossible to isolate problematic repositories. 

Queries work as before, when you don't add the `aggregateRepositories=false` parameter or resolve the repository.
Screenshot 2024-05-24 at 12 22 58

When you add the `aggregateRepositories=false` parameter and resolve the repository, you get individual datapoints for each repository that had a problem.

Screenshot 2024-05-24 at 12 22 34

If you set `aggregateRepositories=true` and attempt to resolve the repository, it will be null.
Screenshot 2024-05-24 at 12 22 43

## Test plan

- Existing code paths are covered by CI
- I will add more tests if this approach is accepted


bahrmichael:
The license check is a known issue: https://sourcegraph.slack.com/archives/C04MYFW01NV/p1715937672950199

cmd/frontend/graphqlbackend/insights.graphql
```diff
@@ -141,7 +141,13 @@ type InsightSeriesStatus {
...
     Data points that are flagged terminally incomplete for this series.
     """
-    incompleteDatapoints: [IncompleteDatapointAlert!]!
+    incompleteDatapoints(
+        """
+        By default, incomplete datapoints are aggregated across all repositories.
+        Setting this to false will allow resolving the repository.
+        """
+        aggregateRepositories: Boolean = true
+    ): [IncompleteDatapointAlert!]!
```
camdencheek, line 150: If I understand this correctly, before we would return N `IncompleteDatapointAlert`s per N data points. With `aggregateRepositories = false`, if there are M repositories, then we can return up to `M * N` items. 

I think it might make more sense to change the signature of `TimeoutDatapointAlert.repository` to return a _set_ of repositories that were incomplete so we can maintain the 1:1 relationship of datapoint to `IncompleteDatapoint`.

internal/insights/store/store.go
```diff
@@ -852,6 +853,31 @@ func (s *Store) LoadAggregatedIncompleteDatapoints(ctx context.Context, seriesID
...
 }
 
+// This is a variant of LoadAggregatedIncompleteDatapoints, but it retains the repoId so that
+// the graphql resolvers can resolve the repository.l
+func (s *Store) LoadIncompleteDatapoints(ctx context.Context, seriesID int) (results []IncompleteDatapoint, err error) {
+	if seriesID == 0 {
+		return nil, errors.New("invalid seriesID")
+	}
+
+	q := "select reason, time, repo_id from insight_series_incomplete_points where series_id = %s group by reason, time, repo_id;"
```
camdencheek, line 863: If returning a set of repositories that failed, we could instead do something like:

```suggestion
	q := "select reason, time, ARRAY_AGG(repo_id) from insight_series_incomplete_points where series_id = %s group by reason, time;"
```

cmd/frontend/graphqlbackend/insights.graphql
bahrmichael: I've updated it, so that we don't have `N * M` items, but `N` items with a list of repositories (`M`).

cmd/frontend/graphqlbackend/insights.graphql
```diff
@@ -141,7 +141,13 @@ type InsightSeriesStatus {
     """
     Data points that are flagged terminally incomplete for this series.
     """
-    incompleteDatapoints: [IncompleteDatapointAlert!]!
+    incompleteDatapoints(
+        """
+        By default, incomplete datapoints are aggregated across all repositories.
+        Setting this to false will allow resolving the repository.
+        """
+        aggregateRepositories: Boolean = true
```
camdencheek, line 149: Q: now that `repositories` is an array, do we still need this parameter? If a client doesn't care about the repository list, they can just exclude that from the list of fields in their query. Excluding this also removes the (documented but still maybe surprising) dependency between the `repositories` field and this argument

internal/insights/store/store.go
```diff
@@ -852,6 +854,42 @@ func (s *Store) LoadAggregatedIncompleteDatapoints(ctx context.Context, seriesID
...
+			&tmp.Reason,
+			&tmp.Time,
+			pq.Array(&repoIds)); err != nil {
+			return err
+		}
+		mappedRepoIds := make([]int, len(repoIds))
+		for i, repoId := range repoIds {
+			if repoId.Valid {
+				mappedRepoIds[i] = int(repoId.Int64)
+			}
```
camdencheek, line 882: Q: the DB schema says `repo_id` is nullable, but that's kinda surprising to me. Do you understand why that is?

camdencheek (approved):

cmd/frontend/graphqlbackend/insights.graphql
bahrmichael: Yes! Thank you for the reminder. I was able to clean it up, and things seem to work as expected. Since I can't find any problematic to the store method, it should be good as long as CI passes.

internal/insights/store/store.go
bahrmichael: I found https://github.com/sourcegraph/sourcegraph/pull/45282 which inserts null here and mentions global queries. The repoId and repoName should be available though, based on the types that this incomplete insert runs on. I haven't found any places where the repoId and repoName on RecordSeriesPointArgs are not set. Maybe it's to reduce the number of inserts for global queries?

In the backend documentation it sounds like there should also be repo information no matter if it's global or not.

https://github.com/sourcegraph/sourcegraph/blob/d4a6b274037c67e3b76250b2c67f7c80df34da51/doc/dev/background-information/insights/backend.md?plain=1#L144-L145

camdencheek: Good find! 

Not blocking, just thinking out loud to try to understand this better. What is a global code insight? It kinda makes sense that a global job wouldn't have a repo ID because it's running against everything, but when would we do that? Maybe there's a special case for an insight that only runs against public repositories, so we know that all users can view all the data, and don't need to keep track of which repo the points are for?

PR #62768, opened 2024-05-17
chore: Factor out logic for modifying site config

varungandhi-src:
Multiple integration tests have essentially the same code for
modifying the site config, and then resetting it to it original
state. Refactor that out into a common method.

## Test plan

Covered by existing tests

### DIFF
--- a/dev/gqltest/search_test.go
+++ b/dev/gqltest/search_test.go
@@ -14,7 +14,6 @@ import (
 
 	"github.com/sourcegraph/sourcegraph/internal/extsvc"
 	"github.com/sourcegraph/sourcegraph/internal/gqltestutil"
-	"github.com/sourcegraph/sourcegraph/schema"
 )
 
 func TestSearch(t *testing.T) {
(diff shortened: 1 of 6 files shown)


dev/gqltest/search_test.go
```diff
@@ -739,30 +735,7 @@ func testSearchClient(t *testing.T, client searchClient) {
...
-				t.Fatal(err)
-			}
-		}()
-
-		siteConfig.ExperimentalFeatures = &schema.ExperimentalFeatures{StructuralSearch: "enabled"}
-		err = client.UpdateSiteConfiguration(siteConfig, lastID)
-		if err != nil {
-			t.Fatal(err)
-		}
+		enableStructuralSearch(t)
```
varungandhi-src, line 738: This part is the main 'logic' change, where we're using the global client instead of the client passed in as the argument. I can make the logic more complex to accommodate the locally passed client, but strictly speaking, all these tests run serially, so I think this is OK?

varungandhi-src:
cc @stefanhengl @pjlast could you review/approve?

eseliger (approved):
thank you!

PR #62770, opened 2024-05-17
feat(search): Add Syntax Highlighting for Hack Language

mmanela:
As part of GRAPH-205 for support Hack language, this PR adds Syntax highlighting.  
Fixes GRAPH-617

Key notes
1. Added a reference to the Hack tree-sitter package
2. Add a highlights query for identifying the items to highlight (Note: I built our version off of the solid foundation found in the [Neovim](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/hack/highlights.scm)  tree sitter repo. But iterated a decent amount for how we would like it highlighted. 
3. Added updated test cases trying to cover many interesting languages features. Hack is a large language and  I am not covering everything possible but many key features.
4. It [seems](https://github.com/slackhq/tree-sitter-hack/blob/main/grammar.js) there are some gaps in the upstream tree-sitter grammar such as doesn't support `module` statement and doesn't support string interpolation (parses just as string) 

Upstream bugs filed
- https://github.com/slackhq/tree-sitter-hack/issues/68
- https://github.com/slackhq/tree-sitter-hack/issues/69
- https://github.com/slackhq/tree-sitter-hack/issues/70
- https://github.com/facebook/hhvm/issues/9447
 
# Screenshots
 
image

## Test plan
 - Unit tests
 - Validate UI and ensure it looks appropriate


docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__hack.hack.snap
```diff
@@ -0,0 +1,871 @@
...
+  use namespace HH\Lib\{C, Vec};
+//^^^ Keyword
+//    ^^^^^^^^^ Keyword
+//              ^^ IdentifierNamespace
+//                ^ PunctuationDelimiter
+//                 ^^^ IdentifierNamespace
+//                      ^ IdentifierNamespace
+//                       ^ PunctuationDelimiter
+//                         ^^^ IdentifierNamespace
+//                             ^ PunctuationDelimiter
```
varungandhi-src, line 16: Marking `C` and `Vec` as `IdentifierNamespace` is not correct.

https://docs.hhvm.com/hack/source-code-fundamentals/namespaces

>  ```
>  use namespace NS1\{C, I, T}; // instead of `NS1\C, NS1\I, NS1\T`
>  ```
>
>  Imported names can designate a namespace, a sub-namespace, a class or interface or trait, a function, or any built-in type.

varungandhi-src:
> It [seems](https://github.com/slackhq/tree-sitter-hack/blob/main/grammar.js) there are some gaps in the upstream tree-sitter grammar such as doesn't support module statement and doesn't support string interpolation (parses just as string)

Please file 1-2 issues upstream so that we can link them here.

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__hack.hack.snap
varungandhi-src: Hmm, these docs indicate that it can only be namespaces?

https://docs.hhvm.com/hack/statements/use

🤔

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/files/hack.hack
```diff
@@ -0,0 +1,173 @@
+// imports
+use namespace HH\Lib\{C, Vec};
+use const Space\Const\C;
+use type Space\Type\T;
```
varungandhi-src, line 4: Please add test cases for `use`:
- without `namespace`/`const`/`type`.
- with `function`

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__hack.hack.snap
```diff
@@ -0,0 +1,871 @@
...
+//^^^^^^^^ Keyword
+//         ^ IdentifierFunction
+//           ^^^^^^^^ IdentifierBuiltinType
+//                    ^^ Identifier
+//                       ^ PunctuationDelimiter
+//                         ^^^^ IdentifierBuiltinType
+  
+  // Shape constructor
+//^^^^^^^^^^^^^^^^^^^^ Comment
+  function foo(int $arg): shape(...){
```
varungandhi-src, line 66: Is highlighting deliberately missing for `...` ?

muglug:
Thanks for doing this!

I know you have the tree-sitter grammar already, but we also have a slightly-more-up-to-date TextMate grammar here that supports `module` and more: https://github.com/slackhq/vscode-hack/blob/master/syntaxes/hack.json

mmanela:
@muglug  Thanks. Our highlighting is based on tree-sitter, so would be great to get the missing parts added to the hack tree-sitter lib. I plan to file issues there on what I found

muglug:
Sounds good — at Slack we’re not yet using modules, so that’s no sort of blocker. Once we start considering that we’ll push necessary changes to the tree-sitter grammar.

mmanela:
Yea I think the bigger annoyance will be the lack of string interpolation support

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__hack.hack.snap
mmanela: Good catch

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__hack.hack.snap
mmanela: From what I read seems like namespace in this case, so I plan to leave as that

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/files/hack.hack
```diff
@@ -0,0 +1,179 @@
...
+
+namespace Hack\Example\namespace {
+
+    // Functions
+    function f(dynamic $d): void {}
+    function g(arraykey $a): void {}
+
+    // Shape constructor
+    function foo(int $arg): shape(...){
+    return shape();
```
varungandhi-src, line 19: It would be good to have test cases for more complex shape types.

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/files/hack.hack
```diff
@@ -0,0 +1,179 @@
...
+    case Position::Bottom:
+      break;
+  }
+
+  // Expressions
+  $d = dict[];
+  $xhp = <tt>Hello <strong>{$user_name}</strong>
+    Text in the markup
+    <!-- this is a comment -->
+</tt>;
```
varungandhi-src, line 83: There is a `->:` operator marked as [Access XHP Attributes](https://docs.hhvm.com/hack/expressions-and-operators/XHP-attribute-selection) here, let's add a test for that.

More generally, it would be useful to go through the various operators on that page and make sure each "group" at least has one test case (e.g. you don't need to test each bitwise operator)

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/files/hack.hack
```diff
@@ -0,0 +1,179 @@
+// imports
```
varungandhi-src, line 1: Q: Did you find any online compiler/playground where this code can be pasted and checked if it parses properly?

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/files/hack.hack
```diff
@@ -0,0 +1,179 @@
...
+use namespace HH\Lib\{C, Vec};
+use const Space\Const\C;
+use type Space\Type\T;
+use function UseNS\f;
+use T;
+
+// NOTE: Tree-sitter grammar seems to not support module
```
varungandhi-src, line 8: Please link the bug reports here and below as well.

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/files/hack.hack
```diff
@@ -0,0 +1,179 @@
...
+  $s = shape('name' => 'db-01', 'age' => 365);
+  $x = true;
+  $y = false;
+  $x = True;
+  $y = FALSE;
+
+// heredoc
+$s = <<<EOT
+	$('a') abc $(function{return;})
+EOT;
```
varungandhi-src, line 100: There are 4 types of string literals, looks like we're missing coverage for the nowdoc style of literals. https://docs.hhvm.com/hack/source-code-fundamentals/literals#string-literals

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/files/hack.hack
```diff
@@ -0,0 +1,179 @@
...
+    yield $keyPrefix.$i => $i * $i; // specify a key/value pair
+  }
+}
+
+<<__EntryPoint>>
+function main(): void {
+  $v1 = -10;
+  $v2 = "red";
+  $a1 = "foo"."bar";
+  echo "\$v1 = ".$v1.", \$v2 = ".$v2."\n";
```
varungandhi-src, line 53: `echo` is one built-in function. Please add tests for various built-in functions such as:
- exit: https://docs.hhvm.com/hack/expressions-and-operators/exit
- invariant: https://docs.hhvm.com/hack/expressions-and-operators/invariant
- `idx()`: https://docs.hhvm.com/hack/expressions-and-operators/coalesce#-and-idx

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/files/hack.hack
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  using ($new = new Object(), $file = new File('using', '+using')) {}
+
+  switch ($pos) {
+    case Position::Bottom:
+      break;
+  }
+
+  // Expressions
```
varungandhi-src, line 78: Please add tests for:
1. Tuple unpacking using `list(...)`. https://docs.hhvm.com/hack/expressions-and-operators/list, including with nesting
2. `@` operator: https://docs.hhvm.com/hack/expressions-and-operators/error-control
3. type assertions using `is`, `as` and `?as`, as well as the legacy type predicate functions (these could be marked as built-in functions): https://docs.hhvm.com/hack/expressions-and-operators/type-assertions
4. Ternary operator `? :` https://docs.hhvm.com/hack/expressions-and-operators/ternary
5. Coalescing using `??` and `??=`: https://docs.hhvm.com/hack/expressions-and-operators/coalesce

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/files/hack.hack
```diff
@@ -0,0 +1,179 @@
...
+    ++$i;
+  } while ($i <= 10);
+
+  foreach (($array as vec[]) as $item) {}
+
+  for (; $i <= 5; ) {
+    ++$i;
+  }
+
+  using ($new = new Object(), $file = new File('using', '+using')) {}
```
varungandhi-src, line 71: Please add tests for different kinds of `new` calls: (https://docs.hhvm.com/hack/expressions-and-operators/new)
- `new $clsname()`
- Different keywords between `new` and the constructor call

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/files/hack.hack
```diff
@@ -0,0 +1,179 @@
...
+  //       string interpolation
+  $y = "hello $x[0]";
+  $y = "hello $x->foo";
+  // Control Flow
+  $i = 1;
+  do {
+    ++$i;
+  } while ($i <= 10);
+
+  foreach (($array as vec[]) as $item) {}
```
varungandhi-src, line 65: Please add tests for `continue`, `try`, `catch`, `finally`. https://docs.hhvm.com/hack/statements/introduction

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/files/hack.hack
```diff
@@ -0,0 +1,179 @@
...
+
+  // Arithmetic
+  $a1 = -10 + 100;
+  $a1 = 100 + -3.4e2;
+  $a1 = 9.5 + 23.444;
+  $a1 = (1 << 63) >> 63;
+  $a1 = 1 > 2;
+}
+
+// Types
```
varungandhi-src, line 115: Please add tests for:
- `noreturn`: https://docs.hhvm.com/hack/built-in-types/noreturn

docker-images/syntax-highlighter/crates/tree-sitter-all-languages/Cargo.toml
```diff
@@ -35,3 +35,4 @@ tree-sitter-zig = { git = "https://github.com/maxxnino/tree-sitter-zig", rev = "
 # and it relies to tree-sitter 0.19.x creating a type mismatch. So depend
 # on our own fork.
 tree-sitter-sql = { git = "https://github.com/sourcegraph/tree-sitter-sql", rev = "6ea62b8d33c9e39f6de8a398a6885d1873ed7e91" }
+tree-sitter-hack = { git = "https://github.com/slackhq/tree-sitter-hack", rev = "fca1e294f6dce8ec5659233a6a21f5bd0ed5b4f2", version = "0.0.4" }
```
varungandhi-src, line 38:
```suggestion
tree-sitter-hack = { git = "https://github.com/slackhq/tree-sitter-hack", rev = "fca1e294f6dce8ec5659233a6a21f5bd0ed5b4f2" }
```

For `git` dependencies, specifying the `rev` is enough, it's unclear why `version` is being specified here.

varungandhi-src (changes requested):
Please go through https://docs.hhvm.com and identify the different syntax forms which aren't being tested and add tests for those. I've identified some which are missing, I haven't tried to be exhaustive.

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/files/hack.hack
mmanela: Yes we can run it via docker. This file does not parse properly, will take some work to change it to that. I can work on that next.

mmanela: This is the simplest way to run the compiler. 
```
docker pull hhvm/hhvm 
docker run  --tty --interactive hhvm/hhvm:latest /bin/bash -l 
hhvm --version
```

mmanela: Ok did a pass and its much closer to compiling but some of the features together (especially modules) make it hard to compile. But was able to validate most of the code.

docker-images/syntax-highlighter/crates/syntax-analysis/queries/hack/highlights.scm
```diff
@@ -0,0 +1,445 @@
...
+((variable) @variable.builtin
+  (#eq? @variable.builtin "$this"))
+
+[
+  (comment)
+  (xhp_comment)
+] @comment
+
+((comment) @comment.documentation
+  (#match? @comment.documentation "^[*][*][^*].*[*]$"))
```
varungandhi-src, line 12: We do not have anything such as `@comment.documentation` in `MATCHES_TO_SYNTAX_KINDS`, please remove this.

docker-images/syntax-highlighter/crates/syntax-analysis/queries/hack/highlights.scm
```diff
@@ -0,0 +1,445 @@
...
+
+"function" @keyword.function
```
varungandhi-src, line 14: Please merge this with the other keywords; see my comment below for the various `@keyword` suffixes.

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/files/hack.hack
```diff
@@ -0,0 +1,334 @@
...
+    }
+
+    // Magic Constants (some of them)
+    $a = __CLASS__;
+    $a = __DIR__;
+    $a = __FILE__;
+
+    // Expressions
+    $tuple = tuple('one', 'two', 'three');
+    list($one, $two, $three) = $tuple;
```
varungandhi-src, line 123: Please add a test for function pointer syntax `f<>`. https://docs.hhvm.com/hack/modules/closures-and-function-pointers

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/files/hack.hack
```diff
@@ -0,0 +1,334 @@
...
+    echo "\$v1 = ".$v1.", \$v2 = ".$v2."\n";
+    exit ("Closing down\n");
+
+    // NOTE: Tree-sitter grammar does not support
+    //       string interpolation
+    //       https://github.com/slackhq/tree-sitter-hack/issues/69
+    $y = "hello $x[0]";
+    $y = "hello $x->foo";
+
+
```
varungandhi-src, line 87:
```suggestion
```

Stray whitespace

docker-images/syntax-highlighter/crates/syntax-analysis/queries/hack/highlights.scm
```diff
@@ -0,0 +1,445 @@
...
+(new_expression
+  . (_) @type)
+
+(new_expression
+  (qualified_identifier
+    (identifier) @type))
+
+(scoped_identifier
+  (qualified_identifier
+    (identifier) @type))
```
varungandhi-src, line 80: This seems unjustified. Looking at https://github.com/slackhq/tree-sitter-hack/blob/main/grammar.js#L81-L94, it seems plausible that this might refer to a [static method](https://docs.hhvm.com/hack/classes/methods).

docker-images/syntax-highlighter/crates/syntax-analysis/queries/hack/highlights.scm
```diff
@@ -0,0 +1,445 @@
...
+
+
+
```
varungandhi-src, line 83:
```suggestion
```

nit: Stray whitespace

docker-images/syntax-highlighter/crates/syntax-analysis/queries/hack/highlights.scm
```diff
@@ -0,0 +1,445 @@
...
+  "nothing"
+  "num"
+] @type.builtin
+
+
+(shape_type_specifier (open_modifier)) @type.builtin
+
+(field_specifier (optional_modifier) @identifier.operator)
+
+(null) @constant.builtin
```
varungandhi-src, line 122:
```suggestion
(null) @constant.null
```

docker-images/syntax-highlighter/crates/syntax-analysis/queries/hack/highlights.scm
```diff
@@ -0,0 +1,445 @@
...
+
+(alias_declaration
+  "newtype"
+  .
+  (_) @type)
+
+(alias_declaration
+  "type"
+  .
+  (_) @type)
```
varungandhi-src, line 144: I think you might be able to combine these with `["newtype" "type"]` perhaps? Seems a bit redundant otherwise.

docker-images/syntax-highlighter/crates/syntax-analysis/queries/hack/highlights.scm
```diff
@@ -0,0 +1,445 @@
...
+  "@"
+  "?"
+  "~"
+  "?->"
+  "->"
+] @operator
+
+(integer) @number
+
+(float) @number.float
```
varungandhi-src, line 236: We don't have `number.float` in `MATCHES_TO_SYNTAX_KINDS`, this can simply be `@number`.

docker-images/syntax-highlighter/crates/syntax-analysis/queries/hack/highlights.scm
```diff
@@ -0,0 +1,445 @@
...
+
+; Eplxicitly handle NAN and INF since they are
+; not mentioned in grammar
+(qualified_identifier
+  . (identifier) @variable.builtin  .
+  (#eq? @variable.builtin "NAN"))
+
+(qualified_identifier
+  . (identifier) @variable.builtin  .
+  (#eq? @variable.builtin "INF"))
```
varungandhi-src, line 271: These should be `@constant.builtin`, not `@variable.builtin`

docker-images/syntax-highlighter/crates/syntax-analysis/queries/hack/highlights.scm
```diff
@@ -0,0 +1,445 @@
...
+    (identifier) @function.call .))
+
+(call_expression
+  function: (selection_expression
+    (qualified_identifier
+      (identifier) @function.call .)))
+
+(selection_expression
+    (_)
+    (xhp_class_identifier) @variable)
```
varungandhi-src, line 256: Why are we specifically checking for `xhp_class_identifier` in this sub-tree? Should we instead mark `xhp_class_identifier` as `@identifier` everywhere?

docker-images/syntax-highlighter/crates/syntax-analysis/queries/hack/highlights.scm
```diff
@@ -0,0 +1,445 @@
...
+; not mentioned in grammar
+(qualified_identifier
+  . (identifier) @keyword  .
+  (#eq? @keyword "internal"))
+  (qualified_identifier
+  . (identifier) @keyword  .
+  (#eq? @keyword "module"))
+
+(namespace_declaration
+    name:  (qualified_identifier (identifier) @variable.module))
```
varungandhi-src, line 284: Would prefer `@identifier.module` to be used here (we should probably rename that wholesale to `identifier.namespace` everywhere since we're using `IdentifierNamespace` in SCIP). Using `variable` in the name conveys some kind of local bindings, but very few languages like OCaml support first-class modules.

docker-images/syntax-highlighter/crates/syntax-analysis/queries/hack/highlights.scm
```diff
@@ -0,0 +1,445 @@
...
+  . (identifier) @variable.builtin  .
+  (#eq? @variable.builtin "INF"))
+
+
+; Explicitly handle internal amd module since they are not
+; not mentioned in grammar
+(qualified_identifier
+  . (identifier) @keyword  .
+  (#eq? @keyword "internal"))
+  (qualified_identifier
```
varungandhi-src, line 279:
```suggestion

(qualified_identifier
```

Please use consistent formatting and spacing.

docker-images/syntax-highlighter/crates/syntax-analysis/queries/hack/highlights.scm
```diff
@@ -0,0 +1,445 @@
...
+    (qualified_identifier
+      (identifier) @type .)
+    alias: (identifier)? @type))
+
+(use_clause
+  (use_type
+    "namespace")
+  (qualified_identifier
+    (_) @variable.module .)
+  alias: (identifier)? @variable.module)
```
varungandhi-src, line 328: 1. When will this be triggered instead of the similar `use_statement` query earlier?
2. Why does this use `(_) @variable.module .` whereas the sub-fragment of the `use_statement` query is `(identifier) @variable.module .`?

varungandhi-src (changes requested):
I've gone through about 700/1650 lines in the snapshot output and 325/450 lines in the queries. Please go through the feedback and re-request review after addressing existing comments.

docker-images/syntax-highlighter/crates/syntax-analysis/queries/hack/highlights.scm
```diff
@@ -0,0 +1,445 @@
...
+  "print"
+  "echo"
+  "newtype"
+  "clone"
+  "as"
+  "concurrent"
+  ; "nameof" This is missing in the grammar
+] @keyword.operator
+
+"return" @keyword.return
```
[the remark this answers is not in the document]
mmanela: I will update for Hack and we can discuss the broader change after. But curious on if it would it be useful to keep these though, since having the distinction in the highlights can help us in the future if we want to decide to identifier these separately later? But I guess since we don't test them today we don't trust it anyway. So for now I do agree.

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__hack.hack.snap
```diff
@@ -0,0 +1,1662 @@
...
+//    ^^^^^^^^^^^^ Comment
+      $keyedcontainer = dict[];
+//    ^^^^^^^^^^^^^^^ Identifier
+//                    ^ IdentifierOperator
+//                      ^^^^ IdentifierBuiltinType
+//                            ^ PunctuationDelimiter
+      $r = idx($keyedcontainer, 'key', 23);
+//    ^^ Identifier
+//       ^ IdentifierOperator
+//         ^^^ IdentifierFunction
```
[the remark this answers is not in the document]
mmanela: They call it built-in in their docs so I will go with that.

docker-images/syntax-highlighter/crates/syntax-analysis/queries/hack/highlights.scm
mmanela: This matches the left hand side of the scoped identifier.  So for `Person::typicalGreeting()` it would match `Person` as a type. Is that not right and its assuming too much

docker-images/syntax-highlighter/crates/syntax-analysis/queries/hack/highlights.scm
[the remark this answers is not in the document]
varungandhi-src: > since having the distinction in the highlights can help us in the future if we want to decide to identifier these separately later

In practice, no mainstream IDEs today distinguish between various kinds of keywords via highlighting.

> But I guess since we don't test them today we don't trust it anyway

Yes, exactly. When we want to use one of these, we want to try to be consistent across languages.

docker-images/syntax-highlighter/crates/syntax-analysis/queries/hack/highlights.scm
varungandhi-src: You're right, I misread the query.

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__hack.hack.snap
```diff
@@ -0,0 +1,1662 @@
...
+expression: "snapshot_treesitter_syntax_kinds(&document, &contents)"
+---
+  new module foo.bar.baz{};
+//^^^ Keyword
+//    ^^^^^^ Keyword
+//           ^^^ Identifier
+//              ^ IdentifierOperator
+//               ^^^ Identifier
+//                  ^ IdentifierOperator
+//                   ^^^ IdentifierType
```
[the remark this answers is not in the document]
varungandhi-src: Can we avoid marking the last one as `IdentifierType` and just fall back to `Identifier`?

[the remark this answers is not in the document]
mmanela: I believe it's this query 

```
(collection
  (qualified_identifier
    (identifier) @type .)) 
```

Since it sees the last segment as collection definition since it ends with the {}. I will see if I can condition this more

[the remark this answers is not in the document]
varungandhi-src: Ah. I suppose that's OK, that doesn't seem easily fixable. 😐

docker-images/syntax-highlighter/crates/syntax-analysis/queries/hack/highlights.scm
mmanela: 1. `use_statement` is used when you have selecting specific sub-items  like `use namespace HH\Lib\{C, Vec}` and `use_clause` is used when you don't `use namespace HH\Lib`.
2. Ah, I used the `(_)` before you had the suggestion to add a identifier node into the hack grammar

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/files/hack.hack
mmanela: Done

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__hack.hack.snap
```diff
@@ -0,0 +1,1662 @@
...
+//                                                 ^ IdentifierOperator
+//                                                   StringLiteral 143:50..145:6
+        Text in the markup
+        <!-- this is a comment -->
+//      ^^^^^^^^^^^^^^^^^^^^^^^^^^ Comment
+//                                 StringLiteral 145:32..146:2
+    </tt>;
+//  ^^ Tag
+//    ^^ Tag
+//      ^ IdentifierOperator
```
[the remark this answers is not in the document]
mmanela: Done with one annoying caveat. In the Jsx grammar it wouldnt let me match in `"</"` and I had to match on the characters individually. I am not sure why, that is why for the jsx snap it shows up like this.

image

docker-images/syntax-highlighter/crates/syntax-analysis/queries/javascript/highlights-jsx.scm
```diff
@@ -1,7 +1,18 @@
...
 ;; This file should be kept in sync with tsx/highlights.scm
 (jsx_opening_element
-  name: (identifier) @tag)
+  "<" @tag.delimiter
+  name: (identifier) @tag
+  ">" @tag.delimiter)
+
 (jsx_closing_element
-  name: (identifier) @tag)
+  ["<" "/"] @tag.delimiter
```
mmanela, line 9: I want to do `"</" @tag.delimiter` but this errors and not sure why.

varungandhi-src: Please split out the JSX change into a separate PR. This PR is already quite big.

internal/highlight/language_test.go
```diff
@@ -296,6 +296,17 @@ func TestGetLanguageFromContent(t *testing.T) {
 %setupPythonIfNeeded Check if python is installed and configured.  If it's`,
 			Expected: "matlab",
 		},
+		{
```
varungandhi-src, line 299: Please avoid adding unrelated changes in the same PR. This change is related to a different issue.

docker-images/syntax-highlighter/crates/syntax-analysis/queries/javascript/highlights-jsx.scm
varungandhi-src: Couple of minor points:

> I want to do ...

If something is sub-optimal in the code, then prefer adding a code comment first and then flagging it in review, instead of just having a PR review comment. That way someone who reads the code next has the context available; coming to this PR and going through the comments is much more time-consuming (and ergo, less likely).

> but this errors and not sure why.

It would be helpful to state what error you got, what steps you took to investigate it, etc.

docker-images/syntax-highlighter/crates/syntax-analysis/queries/hack/highlights.scm
```diff
@@ -0,0 +1,472 @@
...
+
+(call_expression
+  function: (qualified_identifier
+    (identifier) @function.builtin .)
+    (#eq? @function.builtin "is_bool"))
+
+(call_expression
+  function: (qualified_identifier
+    (identifier) @function.builtin .)
+    (#eq? @function.builtin "is_string"))
```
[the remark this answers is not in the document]
varungandhi-src: It seems like `#any-of?` is broken, I changed this to use `#match?` instead.

docker-images/syntax-highlighter/crates/syntax-analysis/queries/hack/highlights.scm
```diff
@@ -0,0 +1,421 @@
...
+(xhp_open_close (xhp_identifier) @tag  (xhp_attribute)? @tag.attribute)
+
+(type_specifier (qualified_identifier (identifier) @type))
+(type_specifier) @type
+
+(variable) @variable
+(identifier) @variable
+(pipe_variable) @variable
+
+(braced_expression) @none
```
varungandhi-src, line 421: There is no `none` in `MATCHES_TO_SYNTAX_KINDS`. Also, applying any kind of highlighting information to braced expressions as a whole doesn't make much sense -- the punctuation is handled separately, and the sub-expression inside will be handled separately.

docker-images/syntax-highlighter/crates/syntax-analysis/queries/javascript/highlights-jsx.scm
varungandhi-src: We're using v0.20.0 of the tree-sitter-javascript grammar, and that has this definition:

```
    jsx_closing_element: $ => seq(
      '<',
      '/',
      field('name', $._jsx_element_name),
      '>'
    ),
```

https://github.com/tree-sitter/tree-sitter-javascript/blob/rust-0.20.0/grammar.js

docker-images/syntax-highlighter/crates/syntax-analysis/queries/hack/highlights.scm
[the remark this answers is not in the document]
mmanela: Yea I tried any-of multiple times to condense this and it didn't work. I think any-of matches multiple captures

docker-images/syntax-highlighter/crates/syntax-analysis/queries/javascript/highlights-jsx.scm
mmanela: Ah! I didn't even think that the grammar I was looking at was newer than the one we were using. Thanks for seeing that

varungandhi-src: For mainstream grammars, you can use the playground to check a query against the latest version. https://tree-sitter.github.io/tree-sitter/playground

docker-images/syntax-highlighter/crates/syntax-analysis/queries/hack/highlights.scm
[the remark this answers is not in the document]
mmanela: Good to see that `match` works for it. Thanks

varungandhi-src (approved):
LGTM except for the outstanding comments on splitting JSX and language detection changes into different PRs.

PR #62775, opened 2024-05-18
gitserver: Switch corruption logs to use separate file

eseliger:
We still incur a ton of git config calls, those add pressure to the process spawning, so we instead do a much cheaper fs.Stat now to check for corruption.

Test plan:

Adjusted tests, CI and integration tests still pass.

PR #62776, opened 2024-05-18
gitserver: Made git GC recloning less dependent on time

eseliger:
On small instances, we gave git GC a lot of time to recover after a failure. However, on large instances, the time based approach could mean that after only a single failed attempt we will already reclone a repo.

This PR changes that to use a counter instead, and moves away from git config for that, fs.ReadFile on a file is much cheaper than spawning a whole process.

Test plan:

Tests still pass, added some new tests.


eseliger:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62776?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#62776** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62776?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#62775** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62775?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

cmd/gitserver/internal/cleanup.go
```diff
@@ -1259,3 +1227,48 @@ func setLastSizeCalculation(dir common.GitDir, when time.Time) error {
 	// We use modtime to track the last time fetched.
 	return os.Chtimes(path, time.Time{}, when)
 }
+
+const gcFailedCounterFilepath = ".sourcegraph-gc-fail-counter"
```
mmanela, line 1231: >  fs.ReadFile on a file is much cheaper than spawning a whole process.

What is our logic then for when it warrants putting stuff in git config and when not to?

cmd/gitserver/internal/cleanup.go
```diff
@@ -113,10 +111,9 @@ var (
 var sgMaintenanceScript string
 
 const (
-	day = 24 * time.Hour
-	// repoTTLGC is how often we should re-clone a repository once it is
-	// reporting git gc issues.
-	repoTTLGC = 2 * day
+	// gcFailureRecloneThreshold is the amount of times git gc has to fail before
+	// a repo is considered for recloning.
+	gcFailureRecloneThreshold = 5
```
mmanela, line 116: Why 5?

mmanela (approved):

cmd/gitserver/internal/cleanup.go
eseliger: Random number to get started - 1 seems too risky, 5 is not 10 😆 I want to tune this as we learn more about the likelihood of this happening

cmd/gitserver/internal/cleanup.go
eseliger: Given the issues we've seen from storing sg specific things in git config, I think we should only store config in there that git itself cares about - everything else we can do with lower IOPS / process spawning overhead ourselves.

ggilmore: Note that we could also use go-git to modify the git config in a structured way (if we want to keep the property of storing things in the git config without spawning the process). I don't feel strongly about this though.

ggilmore (approved):

cmd/gitserver/internal/cleanup.go
eseliger: I thought about switching (had a very old branch actually that switched those to go-git), but when I looked at the implementation it didn't seem to use the same lockfile mechanism that git itself uses so it would probably not be thread safe and come with its own challenges :(

eseliger:
### Merge activity

* **May 21, 2:06 PM EDT**: [Graphite](https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62776) rebased this pull request after merging its parent, because this pull request is set to merge when ready.

PR #62778, opened 2024-05-18
repoupdater: Add retry policy to repositoryservice

eseliger:
This new client didn't have any retries set, so we would always fast-fail when gitserver is unavailable. That makes the operation more noisy than it has to be, and potentially all repos back in the fetch queue, because it thinks the fetch itself failed.
All these methods are safe to be called a second time, so we can safely do that.

This is an easy fix to ensure repo updater waits a bit for gitserver to be back online.

Test plan:

Verified locally that fetches don't immediately fail when gitserver is down.

### DIFF
--- a/cmd/repo-updater/internal/gitserver/BUILD.bazel
+++ b/cmd/repo-updater/internal/gitserver/BUILD.bazel
@@ -14,6 +14,7 @@ go_library(
         "//internal/api",
         "//internal/gitserver/connection",
         "//internal/gitserver/v1:gitserver",
+        "//internal/grpc/defaults",
     ],
 )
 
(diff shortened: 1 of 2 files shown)


eseliger:
* **#62778** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62778?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

ggilmore (approved):
@ggilmore approved this PR from Slack with [Graphite](https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62778)

PR #62779, opened 2024-05-18
fix: Show full info for highlighter errors

varungandhi-src:
Previously, if you made a malformed request, the actual error
would not be shown because the default anyhow to_string impl
would only show the top-most context attached to the error
(see: https://docs.rs/anyhow/latest/anyhow/struct.Error.html#display-representations),
which would be something like `/scip endpoint`.

## Test plan

Added snapshot test

### DIFF
--- a/docker-images/syntax-highlighter/src/snapshots/syntect_server__tests__check_error.snap
+++ b/docker-images/syntax-highlighter/src/snapshots/syntect_server__tests__check_error.snap
@@ -1,0 +1,5 @@
+---
+source: src/lib.rs
+expression: result.unwrap_err()
+---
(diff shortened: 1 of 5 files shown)


docker-images/syntax-highlighter/src/snapshots/syntect_server__tests__check_error.snap
```diff
@@ -0,0 +1,5 @@
+---
+source: src/lib.rs
+expression: result.unwrap_err()
+---
+{"error":"/scip endpoint: Tree-sitter backend requires a language to be specified"}
```
varungandhi-src, line 5: Previously, this was:

```
{"error": "/scip endpoint"}
```

keynmol (approved):
Nice

PR #62788, opened 2024-05-20
fix: backporting

craigfurman:
Revert line that tries to fetch a specific remote branch that has never actually been pushed, back to fetching from the default remote.

I'm not 100% sure this fixes the underlying problem in https://linear.app/sourcegraph/issue/REL-38/having-trouble-back-porting-panw-pr-to-540-need-help, so let's execute the test plan pre-merge.

## Test plan

There don't appear to be many automated tests for backporting, which is fair enough as this is quite fiddly. We can manually test this by running a backport command pre-merge, and seeing if it fixes https://linear.app/sourcegraph/issue/REL-38/having-trouble-back-porting-panw-pr-to-540-need-help:

```
go run ./dev/sg backport <args>
```

### DIFF
--- a/dev/sg/internal/backport/backport.go
+++ b/dev/sg/internal/backport/backport.go
@@ -8,7 +8,6 @@ import (
 	"github.com/urfave/cli/v2"
 
 	"github.com/sourcegraph/sourcegraph/dev/sg/internal/execute"
-	"github.com/sourcegraph/sourcegraph/dev/sg/internal/repo"
 	"github.com/sourcegraph/sourcegraph/dev/sg/internal/std"
 	"github.com/sourcegraph/sourcegraph/lib/errors"
 	"github.com/sourcegraph/sourcegraph/lib/output"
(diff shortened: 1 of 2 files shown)


dev/sg/internal/backport/backport.go
```diff
@@ -68,7 +66,7 @@ func Run(cmd *cli.Context, prNumber int64, version string) error {
 
 	// Fetch latest change from remote
 	p = std.Out.Pending(output.Styled(output.StylePending, "Fetching latest changes from remote..."))
-	if _, err := backportBranchRepo.FetchOrigin(cmd.Context); err != nil {
+	if _, err := execute.Git(cmd.Context, "fetch"); err != nil {
```
craigfurman, line 69: The old code ran `git fetch -a`, but I am not sure if the `-a` was load-bearing. From `man git-fetch`:

```
       -a, --append
           Append ref names and object names of fetched refs to the existing contents of .git/FETCH_HEAD.
           Without this option old data in .git/FETCH_HEAD will be overwritten.
```

I can't think of _why_ this would matter, so I left it out. I wonder if the original code was trying to run `git fetch --all`. I thought `-a` was short for `--all` until reading that man page today.

BolajiOlajide: The intent here is to update the latest changes so we can have access to the commit created by the PR. We could just run a `fetch origin <PR branch>` to fix.

craigfurman: Then this line works, I think. `git fetch` fetches all branches from the default remote (conventionally "origin")

craigfurman: Cool, holding off while we wait for @RXminuS's approval to use his PR as a test case, on the slack thread.

craigfurman: This change works! We do still get an error though when running `go run ./dev/sg backport -p 62720 -r 5.4.0`:

```
Unable to cherry-pick merge commit: "6a7666c42cf349a6b702b8c72c9cc2c4e8421ad7". This might be the result of a merge conflict. Manually run `git cherry-pick 6a7666c42cf349a6b702b8c72c9cc2c4e8421ad7` and fix on your machine.: exit status 1
```

But that's not an issue with the tool, @RXminuS will need to resolve conflicts. Let's merge this 🚀

BolajiOlajide (approved):

PR #62789, opened 2024-05-20
chore: Remove client construction from SignUp/In funcs

varungandhi-src:
These functions are used in test code, which makes it tricker to
pass in loggers for checking requests and responses. Instead, make
the API method-based. This introduces slightly more boilerplate
since Client construction is fallible (so you can't chain 
`NewClient(...).SignUp(...)`), but it allows calling code
to pass in loggers more easily for debugging test failures.

## Test plan

Covered by existing tests


eseliger (approved):
Nice, looks good 👍

PR #62790, opened 2024-05-20
Make SSC base url dynamic

vdavid:
For a long time, it was annoying that we couldn't properly test some SSC features because `https://accounts.sourcegraph.com/cody/subscription` (production URL) was hard-coded for the "Manage subscription" links.
This PR makes that setting dynamic.

Also:
- It wires up the "Create a Cody Pro team" button to pass an extra `?team=1` argument which we'll handle on the checkout page.
- It contains an unrelated event logging cleanup because the deprecation warnings made it difficult for me to focus on the linter warnings related to my PR.

**Note:** If `sscBaseUrl` is not configured in the site config, it falls back to the previously hard-coded URL `https://accounts.sourcegraph.com/cody`.

The PR consists of four individual commits to make the PR review easier.

## Test plan

Tested it manually: it works nicely.

### DIFF
--- a/client/web/src/cody/subscription/CodySubscriptionPage.tsx
+++ b/client/web/src/cody/subscription/CodySubscriptionPage.tsx
@@ -29,7 +29,7 @@ import type { UserCodyPlanResult, UserCodyPlanVariables } from '../../graphql-op
 import { EventName } from '../../util/constants'
 import { CodyColorIcon } from '../chat/CodyPageIcon'
 import { isCodyEnabled } from '../isCodyEnabled'
-import { manageSubscriptionRedirectURL, isEmbeddedCodyProUIEnabled } from '../util'
+import { isEmbeddedCodyProUIEnabled } from '../util'
 
 import { USER_CODY_PLAN } from './queries'
 
@@ -39,6 +39,8 @@ interface CodySubscriptionPageProps extends TelemetryV2Props {
     isSourcegraphDotCom: boolean
     authenticatedUser?: AuthenticatedUser | null
 }
+
+const manageSubscriptionRedirectURL = window.context.frontendCodyProConfig?.sscBaseUrl + '/subscription'
 
 export const CodySubscriptionPage: React.FunctionComponent<CodySubscriptionPageProps> = ({
     isSourcegraphDotCom,
--- a/client/web/src/cody/subscription/CodySubscriptionPage.tsx
+++ b/client/web/src/cody/subscription/CodySubscriptionPage.tsx
@@ -27,7 +27,7 @@ import { CodySubscriptionPlan } from '../../graphql-operations'
 import type { UserCodyPlanResult, UserCodyPlanVariables } from '../../graphql-operations'
 import { CodyColorIcon } from '../chat/CodyPageIcon'
 import { isCodyEnabled } from '../isCodyEnabled'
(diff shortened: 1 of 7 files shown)


client/web/src/cody/subscription/CodySubscriptionPage.tsx
```diff
@@ -227,7 +227,9 @@ export const CodySubscriptionPage: React.FunctionComponent<CodySubscriptionPageP
                                                 telemetryRecorder.recordEvent('cody.planSelection', 'click', {
                                                     metadata: { tier: 1, team: 1 },
                                                 })
-                                                window.location.href = manageSubscriptionRedirectURL // TODO: Use team link or argument
+                                                const url = new URL(manageSubscriptionRedirectURL)
+                                                url.searchParams.append('team', '1')
```
chrsmith, line 231: What is the purpose of `?team=1`? It is probably worth calling out in a code comment, since it isn't entirely obvious why (or even if) that is required?

client/web/src/cody/util.ts
```diff
@@ -1,8 +1,6 @@
 // The URL to direct users in order to manage their Cody Pro subscription.
 import { useState, useEffect } from 'react'
 
-export const manageSubscriptionRedirectURL = 'https://accounts.sourcegraph.com/cody/subscription'
```
chrsmith, line 4: ⚠️  Rather than defining this as a constant in two places, can we instead just make this a function instead? And that way we only have the logic for what the subscription management URL should be, in only one place?

e.g.

```ts
// getManageSubscriptionUrl returns the URL that a user should navigate to in order
// to manage their Cody Pro subscription.
export function getManageSubscriptionUrl(): string {
    ...
}
```

Having a "single source of truth" for something important like that seems like it would be in our best interest. Since otherwise, it's super easy to forget to find all places where we have code to generate the URL. (And therefore, could use a different name or do it in such a way we'd miss it when we need to modify the URL when we go live with the UI migrated to the Sourcegraph.com.)

chrsmith (approved):
This LGTM, and thanks for taking care of it. But IMHO we should probably keep the specific URL we need the user to navigate to in the `utils.ts` file.

Just so that we have a single location where that exists in the code, and therefore can make any changes without unintentionally forgetting to update "all places" where we generate the URL and introduce bugs down the road.

client/web/src/cody/util.ts
vdavid: Addressed in https://github.com/sourcegraph/sourcegraph/pull/62790/commits/d9c85597355147298fdcb32cd74953d75c51832b

client/web/src/cody/subscription/CodySubscriptionPage.tsx
vdavid: Addressed it in https://github.com/sourcegraph/sourcegraph/pull/62790/commits/8ae664cab7eb704d8d38011f4a79b2ffebe16149

chrsmith:
> Note: If sscBaseUrl is not configured in the site config, it falls back to the previously hard-coded URL https://accounts.sourcegraph.com/cody.

For posterity, the "default" value in the `site.schema.json` file absolutely does not do what you think it does. And if you do want to set a default value, you need to actually hard-code it in all of the locations where we would read the value.

(Which I guess is the purpose of `internal/conf/computed.go`.) 😭

PR #62797, opened 2024-05-20
Add missing API types

chrsmith:
Adds the REST API types for the Team Invites and Team Members sections. This isn't used by anything yet, but will be part of updating the Mange `AuthenticatedCodyManageTeamPage` page to use the API client hook.

NOTE: I didn't add this in this PR because I'm trying to tease out my local branch into separate PRs. But here's the relevant update to the `Client` type, so that you can call the endpoints that read/write these data types.

```ts
...
export module Client {
    // Invites

    export function getTeamInvites(): Call<types.ListTeamInvitesResponse> {
        return { method: 'GET', urlSuffix: '/team/current/invites' }
    }

    export function createTeamInvite(requestBody: types.CreateTeamInviteRequest): Call<types.TeamInvite> {
        return { method: 'POST', urlSuffix: '/team/current/invites', requestBody }
    }

    // This endpoint is to be called by a team admin, for convenience since the
    // current team ID may not be known.
    export function cancelCurrentTeamInvite(inviteId: string): Call<void> {
        return { method: 'POST', urlSuffix: `/team/current/invites/${inviteId}/cancel` }
    }

    export function getInvite(teamId: string, inviteId: string): Call<types.TeamInvite> {
        return { method: 'GET', urlSuffix: `/team/${teamId}/invites/${inviteId}` }
    }

    export function acceptInvite(teamId: string, inviteId: string): Call<void> {
        return { method: 'GET', urlSuffix: `/team/${teamId}/invites/${inviteId}/accept` }
    }

    export function cancelTeamInvite(teamId: string, inviteId: string): Call<void> {
        return { method: 'GET', urlSuffix: `/team/${teamId}/invites/${inviteId}/cancel` }
    }

    // Members

    export function getCurrentTeamMembers(): Call<types.ListTeamMembersResponse> {
        return { method: 'GET', urlSuffix: '/team/current/members' }
    }

    export function updateTeamMembers(requestBody: types.UpdateTeamMembersRequest): Call<void> {
        return { method: 'PATCH', urlSuffix: '/team/current/members', requestBody }
    }

    ...
```

## Test Plan

CI/CD

### DIFF
--- a/client/web/src/cody/management/api/teamInvites.ts
+++ b/client/web/src/cody/management/api/teamInvites.ts
@@ -1,0 +1,26 @@
+import { TeamRole } from './teamMembers'
+
+export type TeamInviteStatus = 'sent' | 'errored' | 'accepted' | 'canceled'
+
+export interface TeamInvite {
+    id: string
+
+    email: string
+    role: TeamRole
+
+    status: TeamInviteStatus
+    error?: string
+
+    sentAt: Date
+    acceptedAt?: Date
+}
+
+export interface CreateTeamInviteRequest {
+    email: string
+    role: TeamRole
+}
+
+export interface ListTeamInvitesResponse {
+    invites: TeamInvite[]
+    continuationToken?: string
+}
--- a/client/web/BUILD.bazel
+++ b/client/web/BUILD.bazel
@@ -233,6 +233,8 @@ ts_project(
         "src/cody/management/api/components/CodyProApiClient.ts",
         "src/cody/management/api/hooks/useApiClient.tsx",
         "src/cody/management/api/stripeCheckout.ts",
+        "src/cody/management/api/teamInvites.ts",
+        "src/cody/management/api/teamMembers.ts",
         "src/cody/management/api/teamSubscriptions.ts",
         "src/cody/management/api/types.ts",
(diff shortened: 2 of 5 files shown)


vdavid (approved):
👍

client/web/src/cody/management/api/teamInvites.ts
```diff
@@ -0,0 +1,26 @@
...
+    id: string
+
+    email: string
+    role: TeamRole
+
+    status: TeamInviteStatus
+    error?: string
+
+    sentAt: Date
+    acceptedAt?: Date
```
taras-yemets, line 15: If we going parse API response as JSON to `TeamInvite` (or structure including it), should this and other similar fields be of type `string`, WDYT?

chrsmith: Yes, these should be `string`. This was a mistake on my part.

On the Golang side these are `time.Time`, which end up getting marshaled as an ISO-8601 string. So the types defined here are just wrong.

When using the GraphQL / Apollo library, IIRC it converts times to Dates automatically. (Or maybe it was an extension or custom configuration step or something?)

Anyways, it seems like we would want these data types to match 100% what is sent across the wire. (And not introduce any shim or wrapper later that will automatically convert `string` to `Date`.)

PR #62801, opened 2024-05-20
gitserver: Framework to support integration testing against gitserver

eseliger:
This PR tinkers a bit with building a test helper to run integration tests that are still ~lightweight against a real gitserver.
The caller can either clone a real repo to disk / embed it in the git repo, or can create a small repo on the fly, and then get a running gitserver gRPC server that returns all the data required.

These tests should only exist outside of cmd/ and internal/, as there is a big potential to do cross-cmd imports from here, which can cause bad coupling. But for just these tests, that should be fine.

The most trivial rockskip indexing job that I put in here to POC this runs in 6.3s, including all setup and teardown. That seems very reasonable to me.

Test plan:

The POC test passes.

### DIFF
--- a/dev/rockskipintegration/main_test.go
+++ b/dev/rockskipintegration/main_test.go
@@ -1,0 +1,101 @@
+package main
+
+import (
+	"context"
+	"os"
+	"os/exec"
+	"testing"
+
+	"github.com/bazelbuild/rules_go/go/runfiles"
+	"github.com/sourcegraph/go-ctags"
+	"github.com/sourcegraph/log"
+	"github.com/sourcegraph/log/logtest"
+	"github.com/stretchr/testify/require"
+
+	"github.com/sourcegraph/sourcegraph/cmd/symbols/fetcher"
+	symbolsgitserver "github.com/sourcegraph/sourcegraph/cmd/symbols/gitserver"
+	symbolsParser "github.com/sourcegraph/sourcegraph/cmd/symbols/parser"
+	symbolstypes "github.com/sourcegraph/sourcegraph/cmd/symbols/types"
+	"github.com/sourcegraph/sourcegraph/dev/gitserverintegration"
+	"github.com/sourcegraph/sourcegraph/internal/api"
+	"github.com/sourcegraph/sourcegraph/internal/ctags_config"
+	"github.com/sourcegraph/sourcegraph/internal/database"
+	"github.com/sourcegraph/sourcegraph/internal/database/dbtest"
+	"github.com/sourcegraph/sourcegraph/internal/env"
+	"github.com/sourcegraph/sourcegraph/internal/observation"
+	"github.com/sourcegraph/sourcegraph/internal/rockskip"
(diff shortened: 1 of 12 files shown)


dev/rockskipintegration/main_test.go
```diff
@@ -0,0 +1,175 @@
...
+	require.NoError(t, database.NewDB(logtest.Scoped(t), db).Repos().Create(ctx, &types.Repo{Name: "github.com/sourcegraph/rockskiptest"}))
+	_, err = db.ExecContext(ctx, "INSERT INTO rockskip_repos (repo, last_accessed_at) VALUES ($1, NOW())", "github.com/sourcegraph/rockskiptest")
+	require.NoError(t, err)
+
+	sgs := symbolsgitserver.NewClient(observationCtx, gs)
+	ctagsConfig := symbolstypes.LoadCtagsConfig(env.BaseConfig{})
+	// universal-ctags installed with brew is called ctags, this probably needs work
+	// in CI.
+	ctagsConfig.UniversalCommand = "ctags"
+	svc, err := rockskip.NewService(
```
eseliger, line 62: I used the internal code for this, but it would be nice if the integration test could also spawn a symbols server like we do for gitserver here, and solely test via the external facing APIs.

dev/rockskipintegration/main_test.go
```diff
@@ -0,0 +1,191 @@
...
+			Kind:      "chapter",
+		},
+	}, res)
+}
+
+// newTestGitserverWithRepos spawns a new gitserver with the given repos cloned,
+// the map holds the repo name to path on disk mappings. The repos will be cloned
+// from the location on disk into gitserver, for a most realistic setup.
+// Two clients will be returned to interact with the gitserver.
+func newTestGitserverWithRepos(t *testing.T, repos map[api.RepoName]string) (gitserver.Client, v1.GitserverRepositoryServiceClient) {
```
mmanela, line 114: Should this leave in a more common place for others to re-use?

eseliger: yes - I mostly just wanted to POC that this can work :)

keegancsmith (approved):
Rad!

```
❯ go test -count=1 ./dev/rockskipintegration/ 
ok  	github.com/sourcegraph/sourcegraph/dev/rockskipintegration	0.865s
```

My desktop linux machine likely has a much better FS without our soc2 monitoring :P

Strum355:
> Rad!
> 
> ```
> ❯ go test -count=1 ./dev/rockskipintegration/ 
> ok  	github.com/sourcegraph/sourcegraph/dev/rockskipintegration	0.865s
> ```
> 
> My desktop linux machine likely has a much better FS without our soc2 monitoring :P

0.8s vs 6s....thanks jamf...

eseliger:
```
ok      github.com/sourcegraph/sourcegraph/dev/rockskipintegration      5.029s
```

still quite good for an integration test, but man 😢

PR #62803, opened 2024-05-20
feat/internal/memcmd: add internal/memcmd package to allow for memory tracking of exec.Cmd  processes

ggilmore:
This PR adds a new package memcmd, that adds a new abstraction called "Observer" that allows you to track the memory that a command (and all of its children) is using. (This package uses a polling approach with procfs, since [maxRSS on Linux is otherwise unreliable](https://jkz.wtf/random-linux-oddity-1-ru_maxrss) for our purposes).

Example usage

```go

import (
	"context"
	"fmt"
	"os/exec"
	"time"

	"github.com/sourcegraph/sourcegraph/internal/memcmd"
)

func Example() {
	const template = `
#!/usr/bin/env bash
set -euo pipefail

word=$(head -c "$((10 * 1024 * 1024))" </dev/zero | tr '\0' '\141') # 10MB worth of 'a's
sleep 1
echo ${#word}
`

	cmd := exec.Command("bash", "-c", template)
	err := cmd.Start()
	if err != nil {
		panic(err)
	}

	observer, err := memcmd.NewLinuxObserver(context.Background(), cmd, 1*time.Millisecond)
	if err != nil {
		panic(err)
	}

	observer.Start()
	defer observer.Stop()

	err = cmd.Wait()
	if err != nil {
		panic(err)
	}

	memoryUsage, err := observer.MaxMemoryUsage()
	if err != nil {
		panic(err)
	}

	fmt.Println((0 < memoryUsage && memoryUsage < 50*1024*1024)) // Output should be between 0 and 50MB

	// Output:
	// true
}

```

## Test plan

Unit tests

Note that some tests only work on darwin, so you'll have to run those locally. 

## Changelog 

This feature adds a package that allows us to track the memory usage of commands invoked via exec.Cmd.

### DIFF
--- a/internal/memcmd/BUILD.bazel
+++ b/internal/memcmd/BUILD.bazel
@@ -1,0 +1,50 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library")
+load("//dev:go_defs.bzl", "go_test")
+
+go_library(
+    name = "memcmd",
+    srcs = [
+        "observer.go",
+        "observer_darwin.go",
+        "observer_linux.go",
+    ],
+    importpath = "github.com/sourcegraph/sourcegraph/internal/memcmd",
+    visibility = ["//:__subpackages__"],
+    deps = [
+        "//lib/errors",
+    ] + select({
+        "@io_bazel_rules_go//go/platform:android": [
+            "@com_github_prometheus_procfs//:procfs",
+        ],
+        "@io_bazel_rules_go//go/platform:linux": [
+            "@com_github_prometheus_procfs//:procfs",
+        ],
+        "//conditions:default": [],
+    }),
+)
+
+go_test(
+    name = "memcmd_test",
+    srcs = [
+        "observer_darwin_test.go",
+        "observer_example_test.go",
+        "observer_linux_test.go",
+        "observer_test.go",
+    ],
+    embed = [":memcmd"],
+    deps = select({
+        "@io_bazel_rules_go//go/platform:android": [
+            "//lib/errors",
+            "@com_github_dustin_go_humanize//:go-humanize",
+            "@com_github_google_go_cmp//cmp",
+            "@com_github_sourcegraph_conc//pool",
+        ],
+        "@io_bazel_rules_go//go/platform:linux": [
+            "//lib/errors",
+            "@com_github_dustin_go_humanize//:go-humanize",
+            "@com_github_google_go_cmp//cmp",
+            "@com_github_sourcegraph_conc//pool",
+        ],
+        "//conditions:default": [],
+    }),
+)
--- a/internal/memcmd/BUILD.bazel
+++ b/internal/memcmd/BUILD.bazel
@@ -11,12 +11,15 @@ go_library(
     importpath = "github.com/sourcegraph/sourcegraph/internal/memcmd",
     visibility = ["//:__subpackages__"],
     deps = [
+        "//internal/bytesize",
         "//lib/errors",
     ] + select({
         "@io_bazel_rules_go//go/platform:android": [
+            "//internal/env",
             "@com_github_prometheus_procfs//:procfs",
         ],
         "@io_bazel_rules_go//go/platform:linux": [
+            "//internal/env",
             "@com_github_prometheus_procfs//:procfs",
         ],
         "//conditions:default": [],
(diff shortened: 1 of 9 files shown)


ggilmore:
* **#62803** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62803?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

internal/memcmd/BUILD.bazel
```diff
@@ -0,0 +1,50 @@
...
+    name = "memcmd_test",
+    srcs = [
+        "observer_darwin_test.go",
+        "observer_example_test.go",
+        "observer_linux_test.go",
+        "observer_test.go",
+    ],
+    embed = [":memcmd"],
+    deps = select({
+        "@io_bazel_rules_go//go/platform:android": [
```
Strum355, line 36: android? :thinking:

ggilmore: @Strum355 

I generated this via sg bazel configure on a linux machine - I didn't tweak these rules manually. Is there a way to fix this?

internal/memcmd/observer.go
```diff
@@ -0,0 +1,87 @@
...
+type Observer interface {
+	// Start starts the observer. It should be called before any other method.
+	//
+	// After Start is called, callers must call Stop when they are done with the
+	// observer to release any resources.
+	//
+	// Calling Start() multiple times is safe and has no effect after the first invocation.
+	Start()
+
+	// Stop stops the observer and releases any associated resources.
```
eseliger, line 24: Maybe to clarify that is how it's meant to be used? 

```suggestion
	// Stop stops the observer and releases any associated resources. For accurate measurement,
	// Stop must be called _after_ Wait has been called on the *exec.Cmd.
```

internal/memcmd/observer_darwin.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	usage, ok := state.SysUsage().(*syscall.Rusage)
+	if !ok {
+		return 0, errors.New("failed to get rusage")
+	}
+
+	// On macOS, MAXRSS is the maximum resident set size used (in bytes, not kilobytes).
+	// See getrusage(2) for more information.
+	return uint64(usage.Maxrss), nil
```
eseliger, line 65: this is only correct if the process has been spawned in it's own process group. Should we denote that somewhere in a comment?

internal/memcmd/observer_darwin_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"os/exec"
+	"testing"
+
+	"github.com/dustin/go-humanize"
+
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+)
+
+func TestNewMacObserverIntegration(t *testing.T) {
+	cmd := allocatingGoProgram(t, 500*1024*1024) // 500 MB
```
eseliger, line 17: do you think 500M is required to prove this can work, or could we go lower? That will make the test quite resource intensive.

internal/memcmd/observer_darwin_test.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	memoryUsage, err := observer.MaxMemoryUsage()
+	if err != nil {
+		t.Fatalf("failed to get memory usage: %v", err)
+	}
+
+	t.Logf("memory usage: %s", humanize.Bytes(memoryUsage))
+	memoryLow := uint64(400 * 1024 * 1024)  // 400MB
+	memoryHigh := uint64(650 * 1024 * 1024) // 650MB
```
eseliger, line 46: curious why this is not a symmetric interval :D

internal/memcmd/observer_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+		if !hasHome {
+			if home, err := os.UserHomeDir(); err == nil {
+				cmd.Env = append(cmd.Env, fmt.Sprintf("HOME=%s", home))
+			}
+		}
+
+	}
+
+	return cmd
```
eseliger, line 101: should we run the script once before we return it, so that the GOCACHE is already populated, and the impact of recompiling is minimal on the actual test setup?

eseliger (approved):
LGTM pending the bazel question from Noah and an approving review from the dev-infra team who own bazel

internal/memcmd/observer.go
```diff
@@ -0,0 +1,87 @@
...
+	//
+	// Calling this method will also stop the observer
+	//
+	// It is only valid to call this method after:
+	// 1) Start() has been called and
+	// 2) the underlying process has stopped.
+	//
+	// See the individual observer implementations for more details on how memory
+	// usage is calculated.
+	MaxMemoryUsage() (memoryBytes uint64, err error)
```
eseliger, line 40: idea: to signify what the return unit is, we could use the `bytesize.B` type here

internal/memcmd/observer.go
```diff
@@ -0,0 +1,87 @@
...
+	Stop()
+
+	// MaxMemoryUsage returns the maximum memory usage in bytes of the process since
+	// the observer was started.
+	//
+	// Calling this method will also stop the observer
+	//
+	// It is only valid to call this method after:
+	// 1) Start() has been called and
+	// 2) the underlying process has stopped.
```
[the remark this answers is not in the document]
ggilmore: On certain platforms like macos, the memory usage information fields we read are only collected after we `Wait()` for the process to terminate: https://pkg.go.dev/os#ProcessState.SysUsage

internal/memcmd/observer_darwin.go
ggilmore: I fixed this by making it explicitly error.

internal/memcmd/observer_darwin_test.go
ggilmore: My thought process:

There is some variance due to sampling, so I want to ensure that we're relatively close to the lower bound (under sampling). However, I decided to add more padding to the upper end of the range due to the memory overhead of compilation, the bash script itself, and the go runtime.

internal/memcmd/observer_darwin_test.go
ggilmore: The go compilation / runtime / etc. already adds 60MB of overhead. I wanted to make sure the measurement sample was large enough so that were definitely capturing the memory usage of the program itself (and not the memory usage of the process used to build it) - so it needs to be quite a bit larger than 60 MB. I've halved this to 250 megabytes though given this comment.

Strum355 (approved):

ggilmore:
### Merge activity

* **Jun 10, 5:20 PM EDT**: @ggilmore merged this pull request with [Graphite](https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62803).

PR #62818, opened 2024-05-21
SSC: Three small fixes for Teams & invites

vdavid:
Finally, I could start testing the experience end-to-end.
I found the first few mistakes during that.

Three changes:
1. Fixed a few copy-paste mistakes in the team member list action logging
2. The URL in the case of single users was wrong—now it works correctly. Plus the logging was likely suppressed by the "to" part because `event.preventDefault()` was not called.
3. Made the number of invites dynamic in the notification, pluralized:
CleanShot 2024-05-21 at 14 19 54@2x

The three changes are three commits in this PR to make the review easier.

## Test plan

Tested manually.

### DIFF
--- a/client/web/src/cody/subscription/CodySubscriptionPage.tsx
+++ b/client/web/src/cody/subscription/CodySubscriptionPage.tsx
@@ -238,10 +238,11 @@ export const CodySubscriptionPage: React.FunctionComponent<CodySubscriptionPageP
                                         </Button>
                                         <Link
                                             className="text-center"
-                                            to="https://sourcegraph.com/contact/request-info?utm_source=cody_subscription_page"
+                                            to={manageSubscriptionRedirectURL}
                                             target="_blank"
                                             rel="noreferrer noopener"
-                                            onClick={() => {
+                                            onClick={event => {
+                                                event.preventDefault()
                                                 telemetryRecorder.recordEvent('cody.planSelection', 'click', {
                                                     metadata: { tier: 1, team: 0 },
                                                 })
--- a/client/web/src/cody/team/InviteUsers.tsx
+++ b/client/web/src/cody/team/InviteUsers.tsx
@@ -78,7 +78,9 @@ export const InviteUsers: React.FunctionComponent<InviteUsersProps> = ({
         <>
             {invitesSendingStatus === 'success' && (
                 <div className={classNames('mb-4', styles.alert, styles.blueSuccessAlert)}>
-                    <H3>{invitesSentCount} {pluralize('invite', invitesSentCount)} sent!</H3>
+                    <H3>
+                        {invitesSentCount} {pluralize('invite', invitesSentCount)} sent!
+                    </H3>
                     <Text size="small" className="mb-0">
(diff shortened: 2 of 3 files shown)


client/web/src/cody/subscription/CodySubscriptionPage.tsx
```diff
@@ -238,10 +238,11 @@ export const CodySubscriptionPage: React.FunctionComponent<CodySubscriptionPageP
                                         </Button>
                                         <Link
                                             className="text-center"
-                                            to="https://sourcegraph.com/contact/request-info?utm_source=cody_subscription_page"
+                                            to={manageSubscriptionRedirectURL}
```
chrsmith, line 241: ⚠️ I'm not sure this is right, and just want to double check my understanding:

```ts
export const manageSubscriptionRedirectURL = `${window.context.frontendCodyProConfig?.sscBaseUrl}/subscription`
```

So the current way this is configured, this will redirect users to https://accounts.sourcegraph.com/cody/subscription. ( ✅ )

But down the road, do we plan on setting `sscBaseUrl` to https://sourcegraph.com/cody/manage? And then expect the `/cody/manage/subscription` page to redirect to `/cody/manage/subscription/new` if the user doesn't have an active Cody Pro subscription? Since (IIRC) we don't have a `/cody/manage/subscription` page wired up yet, that seems like maybe a dangerous assumption to make?

So for clarity, would we want to export a new `createNewCodyProSubscriptionRedirectUrl` just to remove any ambiguity? Since in this case, we expect the user to _create_ a new subscription/Cody Pro team, not manage an existing one.

chrsmith (approved):
Just one thing that may be A-OK and I'm just confused about, but otherwise all good things!

client/web/src/cody/subscription/CodySubscriptionPage.tsx
chrsmith: JFYI, and perhaps this will fix it, it looks like this is broken. (At least in our dev environment.) It's very likely I have misconfigured something.

https://cody-services-dev.sgdev.dev/subscription shows this link:
image

And the "Upgrade yourself to Pro" points to: https://cody-services-dev.sgdev.dev/subscription which 404s. (Which is surprising, since I didn't set the `sscBackendUrl` value in the site config yet. So it should have defaulted to an entirely different domain, accounts.sourcegraph.com...)

image

vdavid: > But down the road, do we plan on setting sscBaseUrl to https://sourcegraph.com/cody/manage?

I mean, if we want to migrate this to some other URL, we can change the definition of `manageSubscriptionRedirectURL`, right? So I'd not be too forward-looking with this.

> Update: After setting the sscBackendUrl the link now works, but it redirects to the exact same page (with the behavior that is checked-in now).

Did you figure out what was the problem? TBH I didn't fully get what you meant was wrong here.
Your PR https://github.com/sourcegraph/sourcegraph/pull/62835 fixes the behavior when `sscBackedUrl` is _not_ defined, but when it is defined, the link worked in my tests.

PR #62822, opened 2024-05-21
svelte: Preserve current revision in repo search input

### LINKED ISSUE #62713: Svelte: preserve repo revision in search bar

fkling:
Closes #62713 

## Test plan

Manually tested with default revision, selected branch and selected revision.

### DIFF
--- a/client/web-sveltekit/src/routes/[...repo=reporev]/RepoSearchInput.svelte
+++ b/client/web-sveltekit/src/routes/[...repo=reporev]/RepoSearchInput.svelte
@@ -15,7 +15,7 @@
     /**
      * The revision to search in. If not provided, the default branch is used.
      */
-    export let revision: { commitID: string; revision: string } | undefined
+    export let revision: string
 
     const {
         elements: { trigger, overlay, content, portalled },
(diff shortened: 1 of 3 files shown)


client/web-sveltekit/src/routes/[...repo=reporev]/RepoSearchInput.svelte
```diff
@@ -34,6 +37,25 @@
         )
     }
 
+    function formatRevision(revision: string, commitID: string): string {
+        // Do not append revison for the default branch
+        if (!revision) {
+            return ''
+        }
+
+        // If the (URL) revision is or starts with the commit ID, use the abbreviated commit ID
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/RepoSearchInput.svelte:46]: Oh, that's a smart way to do it.

camdencheek (approved):
Tried it out locally, LGTM!

client/web-sveltekit/src/routes/[...repo=reporev]/RepoSearchInput.svelte
Comment (fkling) [reply client/web-sveltekit/src/routes/[...repo=reporev]/RepoSearchInput.svelte:46]: I in the React app we only check whether the revision is 40 chars, assume its an commit SHA and abbreviate it. This produces incorrect behavior when there is a branch name with exactly 40 chars. We should fix that... at the same time I wish there was a way to abstract all these details from components. Maybe we should compute the human readable revision in the data loader and simply pass that around.

Comment (fkling) [reply client/web-sveltekit/src/routes/[...repo=reporev]/RepoSearchInput.svelte]: I moved the computation to the data loader. We are already doing something similar for repo name. https://github.com/sourcegraph/sourcegraph/pull/62822/commits/7497104c39bd238e3160c2c8b1b7a6f57db5a4ca

taiyab:
@fkling I think this broke showing suggestions right away as soon as you reveal search via `/`.

PR #62830, opened 2024-05-21
Update sensitive metadata allowlist to filter on keys

akalia25:
This pull request introduces a mechanism to selectively include specific `privateMetadata` keys to be allowlisted and sent for all users. Previously, allowlisting only worked to allowlist all keys on the event to be sent. This gives us greater flexibility/safety on what fields we'll be sending on events that have an allowList.

**Benefits**
- Enables selective logging of `privateMetadata` keys that are deemed safe and valuable for analysis.
- Maintains privacy and security by excluding potentially sensitive metadata fields not present in the allowed list, but are in the event.

co-authored with: @bobheadxi 

## Test plan
- Added unit tests to verify expected behavior


internal/telemetry/sensitivemetadataallowlist/redact.go
```diff
@@ -10,23 +14,36 @@ type redactMode int
...
+		// Do private metadata redaction in this if case, as the next case will strip
+		// everything
+		for key, value := range event.Parameters.PrivateMetadata.Fields {
+			if !slices.Contains(allowedPrivateMetadataKeys, key) {
+				// Strip all keys that are NOT in the allowlist
+				delete(event.Parameters.PrivateMetadata.Fields, key)
+			} else if _, isString := value.Kind.(*structpb.Value_StringValue); !isString {
+				// Strip all values that are not strings, even if they ARE in the allowlist
+				event.Parameters.PrivateMetadata.Fields[key] =
+					structpb.NewStringValue("ERROR: value of allowlisted key was not a string")
```
bobheadxi, line 44: This will emit the actual type we got in the error value:

```suggestion
					structpb.NewStringValue(fmt.Sprintf("ERROR: value of allowlisted key was not a string, got: %T", value.Kind))
```

internal/telemetry/sensitivemetadataallowlist/redact.go
```diff
@@ -10,23 +14,36 @@ type redactMode int
...
-	if mode >= redactMarketing {
+	if mode >= redactMarketingAndUnallowedPrivateMetadataKeys {
+		// Do private metadata redaction in this if case, as the next case will strip
+		// everything
+		for key, value := range event.Parameters.PrivateMetadata.Fields {
+			if !slices.Contains(allowedPrivateMetadataKeys, key) {
+				// Strip all keys that are NOT in the allowlist
+				delete(event.Parameters.PrivateMetadata.Fields, key)
+			} else if _, isString := value.Kind.(*structpb.Value_StringValue); !isString {
+				// Strip all values that are not strings, even if they ARE in the allowlist
```
bobheadxi, line 42: Can you add some more context here based on what we discussed, i.e. to prevent arbitrarily deep values and also because we only expect string values to be allowlisted?

internal/telemetry/sensitivemetadataallowlist/sensitivemetadataallowlist.go
```diff
@@ -81,14 +99,17 @@ func (e EventTypes) validate() error {
 type EventType struct {
 	Feature string
 	Action  string
-
-	// Future: maybe restrict to specific, known private metadata fields as well
+	// AllowedPrivateMetadataKeys is the list of field names permitted to be exported from the `privateMetadata` object.
+	AllowedPrivateMetadataKeys []string
```
bobheadxi, line 103: I think this needs more documentation, e.g. this needs to be a top-level key, the value of allowed keys must be strings, etc.

bobheadxi:
For context, Aditya and I paired on this over a call; I'm aligned with overall direction, but let's make sure everything is sufficiently documented :)

internal/telemetry/sensitivemetadataallowlist/redact.go
akalia25: Great suggestion, added!

internal/telemetry/sensitivemetadataallowlist/sensitivemetadataallowlist.go
akalia25: Added more documentation here, lmk if its sufficient :)

internal/telemetry/sensitivemetadataallowlist/redact.go
akalia25: Added more documentation here, lmk if its sufficient :)

internal/telemetry/sensitivemetadataallowlist/redact_test.go
```diff
@@ -27,10 +29,11 @@ func TestRedactEvent(t *testing.T) {
...
-		name   string
-		mode   redactMode
-		event  *telemetrygatewayv1.Event
-		assert func(t *testing.T, got *telemetrygatewayv1.Event)
+		name        string
+		mode        redactMode
+		event       *telemetrygatewayv1.Event
+		allowedKeys []string
+		assert      func(t *testing.T, got *telemetrygatewayv1.Event)
 	}{
```
bobheadxi, line 37: Can you add a test case here for what happens to alowlisted, but not string values? In `assert`, you can provide `autogold.Expect` and use that to assert that the field is replaced with an error message in the format we expect

bobheadxi: The autogold is similar to what we did in https://github.com/sourcegraph/sourcegraph/pull/62830/files#diff-b61c40a8ec14aafd5f7584a33f4fb07acf6067d60b16c0219fc674fd43747a7aR77 ; you can update autogold assertions with `go test $pkg -update`

internal/telemetry/sensitivemetadataallowlist/sensitivemetadataallowlist.go
```diff
@@ -30,20 +30,38 @@ func AllowedEventTypes() EventTypes {
 		EventType{
 			Feature: string(telemetry.FeatureExample),
 			Action:  string(telemetry.ActionExample),
+			AllowedPrivateMetadataKeys: []string{
+				"testField",
+			},
+		},
+		// reason for allowlisting
```
bobheadxi, line 37: I think this was a TODO, can you update this docstring with more context on these events?

internal/telemetry/sensitivemetadataallowlist/redact.go
bobheadxi: LGTM!

internal/telemetry/sensitivemetadataallowlist/sensitivemetadataallowlist.go
bobheadxi: LGTM, thank you!

internal/telemetry/sensitivemetadataallowlist/sensitivemetadataallowlist.go
akalia25: ah yes, added details on why we are allowing `languageId` on autocomplete. thank you!

internal/telemetry/teestore/teestore_test.go
```diff
@@ -318,7 +319,8 @@ func TestToEventLogs(t *testing.T) {
     "AnonymousUserID": "anonymous",
     "Argument": {
       "private": "sensitive-data",
-      "telemetry.privateMetadata.exportable": false
+      "telemetry.privateMetadata.exportable": false,
+	  "telemetry.privateMetadata.exportableFields": null
```
bobheadxi, line 323: Is this spacing correct? This uses autogold so you can update with `-update` as well

bobheadxi (approved):
Nice work!

internal/telemetry/teestore/teestore_test.go
akalia25: > you can update with -update as well

Ah, yes!! My editor auto-saved with the weird spacing, which was also causing the tests to fail! I even tried `sg lint—-fix go`, but your `go test—-update` actually corrected it! Thank you!

PR #62833, opened 2024-05-21
Svelte: fixup popover placement and root path rendering

camdencheek:
This offsets the file popover so the file names line up. Additionally, it hides the directory name when the file is located in the root directory because it looks awkward.

screenshot-2024-05-21_17-07-40@2x

# Test plan

Manual visual testing.

### DIFF
--- a/client/web-sveltekit/src/lib/Popover.svelte
+++ b/client/web-sveltekit/src/lib/Popover.svelte
@@ -1,14 +1,15 @@
 <script lang="ts">
-    import type { Placement } from '@floating-ui/dom'
+    import type { Placement, OffsetOptions } from '@floating-ui/dom'
     import type { Action } from 'svelte/action'
 
     import { popover, onClickOutside, portal } from './dom'
 
-    export let placement: Placement = 'bottom'
     /**
      * Show the popover when hovering over the trigger.
      */
     export let showOnHover = false
+    export let placement: Placement = 'bottom'
+    export let offset: OffsetOptions = showOnHover ? 0 : 3
 
     let isOpen = false
     let trigger: HTMLElement | null
@@ -87,7 +88,7 @@
             reference: target ?? trigger,
             options: {
                 placement,
-                offset: showOnHover ? 0 : 3,
+                offset,
                 shift: { padding: 4 },
             },
         }}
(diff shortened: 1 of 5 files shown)


client/web-sveltekit/src/lib/Popover.svelte
```diff
@@ -1,14 +1,15 @@
...
 
     import { popover, onClickOutside, portal } from './dom'
 
-    export let placement: Placement = 'bottom'
     /**
      * Show the popover when hovering over the trigger.
      */
     export let showOnHover = false
+    export let placement: Placement = 'bottom'
+    export let offset: OffsetOptions = showOnHover ? 0 : 3
```
camdencheek, line 12: This exposes the offset options, which can be used in conjunction with `Placement`. The default is the same default as before

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte
```diff
@@ -124,7 +124,15 @@
                     We handle navigation via the TreeView's select event, to preserve the focus state.
                     Using a link here allows us to benefit from data preloading.
                 -->
-                <Popover let:registerTrigger placement="right-end" showOnHover>
+                <Popover
+                    let:registerTrigger
+                    placement="right-start"
+                    offset={{
+                        // A number which visually offsets popover header so the file names line up
+                        crossAxis: -37,
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte:132]: This is a little awkward, but it doesn't need to be perfect so it's okay if it gets out of sync. I think an ideal solution here would probably look like "this point in my popover component defines the ideal anchor point"

taiyab (commented):
Visually: LGTM!

vovakulikov:
@taiyab @camdencheek  not a blocker for this PR but since we're changing file tooltip here, 5c of my feeback. 

When I saw it for the first time I thought that the commit badge was in the incorrect place, I would expect it right after the "Last changed @" line, like this 

Screenshot 2024-05-22 at 11 18 35

Also, I would try to add an arrow UI to the tooltip that points to the target (file entry in the file tree) similar to what we have in the tooltip (also not a blocker to this PR)

camdencheek:
@taiyab any thoughts on an arrow for the popover component specifically? Our tooltips have an arrow, but it was never added for the popovers

taiyab:
> When I saw it for the first time I thought that the commit badge was in the incorrect place, I would expect it right after the "Last changed @" line, like this

I'd rather try and keep this as consistent as possible to how we (ought to) render the commits in other places too (history list, etc). So, instead of adding the SHA on the "Last changed" line, we just remove the "@" which is causing the confusion.

> any thoughts on an arrow for the popover component specifically? Our tooltips have an arrow, but it was never added for the popovers

No need for this. Thanks.

camdencheek:
Closing since we are removing file popovers from the file tree

PR #62837, opened 2024-05-22
syncer: Fix issues causing excessive fetches

eseliger:
This PR is the result of quite a journey. You'll see a bunch of meaningless one-liners, and yes - this was a bit frustrating :D

tl;dr I noticed that S2 was fetching repos a ton of times, but the numbers didn't check out. Why would we need 9 fetches a second to keep 1100 repos fresh, of which most are very inactive?
Turns out, we trigger fetches for repos that are marked as modified during code host syncing.
That in itself seems mostly unnecessary (so I reduced the conditions under which that would trigger an immediate update),
and the condition for detecting when a repo was modified was also not correctly computed in a bunch of cases.
These are fixed by one-liners here as well. Most likely, this is not a complete list, but it's what I found locally.

Here's some more writeup:
https://www.notion.so/sourcegraph/May-22-2024-S2-fetches-repos-a-lot-e19c282bb3564140b09086ebce172450?pvs=4

Test plan:

See the notion journey as well, but I verified that the repos are no longer marked as modified erroneously in my local instance.


eseliger:
* **#62837** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62837?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

internal/repos/azuredevops.go
```diff
@@ -180,7 +180,7 @@ func (s *AzureDevOpsSource) makeRepo(p azuredevops.Repository) (*types.Repo, err
 				CloneURL: cloneURL,
 			},
 		},
-		Metadata: p,
+		Metadata: &p,
```
eseliger, line 183: Metadata is always a pointer when unmarshalled from the DB, so it should also be a pointer on this end, otherwise the changeset syncer always detects it as "Metadata changed" because it switched from non-pointer to pointer.

internal/extsvc/github/v4.go
```diff
@@ -598,6 +598,7 @@ fragment RepositoryFields on Repository {
 	isDisabled
 	viewerPermission
 	stargazerCount
+	visibility
```
eseliger, line 601: There's a condition below that adds it to the conditional fields, but those are never added here, so instead we add it here explicitly.
For dotcom, we would not have fetched visibility previously.

internal/repos/github.go
```diff
@@ -409,8 +409,9 @@ func (s *GitHubSource) makeRepo(r *github.Repository) *types.Repo {
 	// so we don't want to store it.
 	metadata.ViewerPermission = ""
 	metadata.Description = sanitizeToUTF8(metadata.Description)
+	metadata.Visibility = github.Visibility(strings.ToLower(string(r.Visibility)))
```
eseliger, line 412: Depending on if the value comes from GQL or REST it can be uppercase or lowercase. This unifies it, although I don't love that this exists in the first place.

internal/extsvc/github/common.go
```diff
@@ -1805,9 +1805,6 @@ type Repository struct {
 	// https://developer.github.com/changes/2019-12-03-internal-visibility-changes/#repository-visibility-fields
 	Visibility Visibility `json:"visibility,omitempty"`
 
-	// Parent is non-nil for forks and contains details of the parent repository.
-	Parent *ParentRepository `json:",omitempty"`
```
eseliger, line 1809: the parent wasn't persisted in the DB, so it would always trigger a "metadata changed" - this was added for supporting deduplication of objects but we stopped that project, so it's safe to remove it for now.

cmd/repo-updater/internal/scheduler/scheduler.go
```diff
@@ -281,8 +281,14 @@ func (s *UpdateScheduler) UpdateFromDiff(diff types.RepoSyncDiff) {
...
-	for _, r := range diff.Modified.Repos() {
-		s.upsert(r, true)
+	for _, r := range diff.Modified {
+		// Modified repos only need to be updated immediately if their name changed,
+		// otherwise we just make sure they're part of the scheduler, but don't
+		// trigger a repo update.
+		// TODO: Should we consider these too?
+		// RepoModifiedExternalRepo
+		// RepoModifiedSources
+		s.upsert(r.Repo, r.Modified&types.RepoModifiedName == types.RepoModifiedName)
```
eseliger, line 291: I don't believe there's strong reasons to believe that any of those changes will make a difference to the git repo itself, thus fetching is most likely not needed.
Thoughts on the two fields in the TODO note? I wasn't 1000000000% convinced there, but would be inclined to leave them out for now.
We still fetch repos regularly, this will only affect cadence.

ggilmore: > RepoModifiedExternalRepo

https://github.com/sourcegraph/sourcegraph/blob/17a9768c3abbe0f406e7eefc869e48539024a9f5/internal/api/api.go#L68-L91

I don't see how this should affect contents on disk. The only one of these that looks slightly suspicious is the `ServiceID` field - but I believe that's already covered since the service ID (e.g. github.examplecompany.com) would show up any way in the repository name.

> RepoModified sources
https://github.com/sourcegraph/sourcegraph/blob/207b712f0a3dfdd02ecbe532a8dd02e52a899185/internal/types/types.go#L78-L80

This is just the set of all external services that define the repo, right? I don't see how a change here could ever change the git content on disk:

https://github.com/sourcegraph/sourcegraph/blob/207b712f0a3dfdd02ecbe532a8dd02e52a899185/internal/types/types.go#L33-L37

1) Having a new external service define the same repo as an old external service shouldn't affect the contents on disk.
2) Changing ID shouldn't affect anything
3) Changing the clone URL I don't believe would change anything on disk. (The only thing I could see is many a rename due to how we address repositories, but we have other mechanisms for that).

internal/repos/azuredevops.go
ggilmore: [facepalm](https://app.graphite.dev//settings/meme-library?org=sourcegraph)

cmd/repo-updater/internal/scheduler/scheduler_test.go
```diff
@@ -613,7 +613,6 @@ func Test_updateScheduler_UpdateFromDiff(t *testing.T) {
 			},
 			finalQueue: []*repoUpdate{
 				{Repo: a, Seq: 1, Updating: false},
-				{Repo: b, Seq: 2, Updating: false},
```
ggilmore, line 616: Why did this line need to be deleted?

internal/extsvc/github/testdata/golden/TestV3Client_Fork_success_user
```diff
@@ -14,9 +14,5 @@
    "Nodes": []
   },
   "visibility": "public",
-  "Parent": {
```
ggilmore, line 17: These were deleted because we don't use the parent field any more (and don't request it - right)?

internal/repos/testdata/sources/GITLAB/TestGitlabSource_ListRepos
```diff
@@ -34,7 +34,10 @@
     "forks_count": 0,
     "empty_repo": false,
     "default_branch": "main",
-    "topics": ["topic1", "topic2"]
+    "topics": [
+     "topic1",
```
ggilmore, line 38: Doesn't change your PR, but it would be great if we can use prettier or something on these files so that formatting changes like this don't show up in the diff.

ggilmore (approved):

internal/extsvc/github/testdata/golden/TestV3Client_Fork_success_user
eseliger: Correct - it's not part of the stored JSON anymore

cmd/repo-updater/internal/scheduler/scheduler_test.go
eseliger: Because the repo is modified and we no longer insert modified into the update queue

internal/repos/testdata/sources/GITLAB/TestGitlabSource_ListRepos
eseliger: This is actually what the generated output looks like, I think they were erroneously prettier formatted before :/

cmd/repo-updater/internal/scheduler/scheduler.go
eseliger: > This is just the set of all external services that define the repo, right? I don't see how a change here could ever change the git content on disk:


Yes that is correct to my understanding 

Cool so I can drop the TODO here I think. Thanks for checking my thoughts!

cmd/repo-updater/internal/scheduler/scheduler_test.go
ggilmore: Can you add a comment to this effect explaining why it's omitted?

eseliger: yep done 👍

PR #62839, opened 2024-05-22
Add default values for CG config so it starts

rafax:
This PR fixes local dev configuration of Cody Gateway, providing a default (invalid) value for Sourcegraph Embeddings API and pointing to actual API tokens. Without this, Cody Gateway will refuse to start.

## Test plan

- CG starts locally


taras-yemets (approved):
Thank you, Rafal!

PR #62842, opened 2024-05-22
svelte: Improve commit page loading state

fkling:
It was reported ([Slack thread](https://sourcegraph.slack.com/archives/C05MHAP318B/p1716347339416449)) that we show the old commit data when navigating to a different commit page.
The page was explicitly implemented to show the previously loaded data in case there is an error when fetching more data. But this also affects normal page navigation.

In this commit the implementation was changed to return the previously loaded data when an error occurs, which simplifies loading state and data rendering on the page itself.

## Test plan

Manual testing + new integration test.

### DIFF
--- a/client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/-/commit/[...revspec]/+page.svelte
+++ b/client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/-/commit/[...revspec]/+page.svelte
@@ -15,7 +15,6 @@
     import CopyButton from '$lib/wildcard/CopyButton.svelte'
 
     import type { PageData, Snapshot } from './$types'
-    import type { CommitPage_DiffConnection } from './page.gql'
 
     interface Capture {
         scroll: ScrollerCapture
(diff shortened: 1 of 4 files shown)


client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/-/commit/[...revspec]/+page.svelte
```diff
@@ -157,11 +151,17 @@
...
         padding: 1rem;
+    }
+
+    ul.diffs {
+        // Removes globally set margin
+        margin: 0;
+        list-style: none;
 
-        li {
+        li:not(:last-child) {
```
Comment (vovakulikov) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/-/commit/[...revspec]/+page.svelte:164]: Found not that easy to read selectors with `not` operator, maybe `& + &` would be simpler here

vovakulikov (commented):
Tested manually, not a blocker but with big commits (I usually test on `DefinitelyTyped` repository) we still show old content as we fetch the new one, so with big commits when it takes senseable time to fetch new commit data it looks like a freeze. 

My suggestion here would be to add some inline loader state (maybe against the link which cases transition, or some global loading indicator somewhere at the top of the page.

vovakulikov (approved):
Approve since code LGTM

fkling:
> we still show old content as we fetch the new one

@vovakulikov do you mean the commit information (title, message, ...)?

Showing some kind of navigation indicator is a general problem. We have to find a solution that works across the app. Other places have that issue too (e.g when using fuzzy finder). That requires some design input (cc @taiyab)

PR #62852, opened 2024-05-22
Make the URL dynamic, now with a fallback

vdavid:
I made this URL dynamic before, [here](https://github.com/sourcegraph/sourcegraph/pull/62790), but that didn't work [because](https://github.com/sourcegraph/sourcegraph/pull/62790#issuecomment-2123533378) the "default" value in the `site.schema.json` file doesn't do what I thought it did. So @chrsmith did a quick fix in https://github.com/sourcegraph/sourcegraph/pull/62835, but then we lost the ability to make this dynamic.

So this PR makes it dynamic again, this time with a working fallback.

### Alternatives considered:

I tried to do the same thing on the back end, but it didn't feel right.

1. I tried a transformation in `computed.go`:

```go
// GetCodyProConfig returns the CodyProConfig for the current site config,
// with some handy defaults.
func GetCodyProConfig(siteConfig schema.SiteConfiguration) (c conftypes.CodyProConfig) {
	c.SamsBackendOrigin = "https://accounts.sourcegraph.com"
	c.SscBackendOrigin = "https://accounts.sourcegraph.com"
	c.SscBaseUrl = "https://accounts.sourcegraph.com/cody"
	c.StripePublishableKey = nil

	if siteConfig.Dotcom == nil || siteConfig.Dotcom.CodyProConfig == nil {
		return
	}

	codyProConfig := siteConfig.Dotcom.CodyProConfig
	if codyProConfig != nil {
		c.SamsBackendOrigin = codyProConfig.SamsBackendOrigin
		c.SscBackendOrigin = codyProConfig.SscBackendOrigin
		c.SscBaseUrl = codyProConfig.SscBaseUrl
		c.StripePublishableKey = &codyProConfig.StripePublishableKey
	}
	return
}
```

But having the defaults there was meh, plus I had to introduce the new type `conftypes.CodyProConfig` but nothing prevents another eng to just use the config directly, which might become confusing.

2. I thought about setting the defaults in `makeFrontendCodyProConfig`, which might be better, but then, on the front end, we already have a single location where we access the SSC base URL, and we already had the default there, so I just went with that.

## Test plan

Here is a [1-min Loom](https://www.loom.com/share/a86ed717a19344c5bb14a5d4e45a1f14) where I test it with different settings and demo that it works as intended.

### DIFF
--- a/client/web/src/cody/util.ts
+++ b/client/web/src/cody/util.ts
@@ -1,7 +1,7 @@
 // The URL to direct users in order to manage their Cody Pro subscription.
 import { useState, useEffect } from 'react'
 
-// URL the user needs to navigate to in4 order to modify their Cody Pro subscription.
+// URL the user needs to navigate to in order to modify their Cody Pro subscription.
 export const manageSubscriptionRedirectURL = `${
     window.context.frontendCodyProConfig?.sscBaseUrl || 'https://accounts.sourcegraph.com/cody'
 }/subscription`
--- a/client/web/src/cody/util.ts
+++ b/client/web/src/cody/util.ts
@@ -1,14 +1,10 @@
 // The URL to direct users in order to manage their Cody Pro subscription.
 import { useState, useEffect } from 'react'
 
-// URL the user needs to navigate to in order to modify their Cody Pro subscription.
-//
-// BUG: This should be configurable via the `window.context.frontendCodyProConfig`,
-// but because the backend doesn't supply a meaningful default, and this is used
-// as a value instead of a function to compute the value, we just fix things by
-// going back to hard-coding it.
-// export const manageSubscriptionRedirectURL = `${window.context.frontendCodyProConfig?.sscBaseUrl}/subscription`
-export const manageSubscriptionRedirectURL = 'https://accounts.sourcegraph.com/cody/subscription'
+// URL the user needs to navigate to in4 order to modify their Cody Pro subscription.
+export const manageSubscriptionRedirectURL = `${
+    window.context.frontendCodyProConfig?.sscBaseUrl || 'https://accounts.sourcegraph.com/cody'
+}/subscription`
 
 /**
  * useEmbeddedCodyProUi returns if we expect the Cody Pro UI to be served from sourcegraph.com. Meaning


client/web/src/cody/util.ts
```diff
@@ -1,14 +1,10 @@
...
 
-// URL the user needs to navigate to in order to modify their Cody Pro subscription.
-//
-// BUG: This should be configurable via the `window.context.frontendCodyProConfig`,
-// but because the backend doesn't supply a meaningful default, and this is used
-// as a value instead of a function to compute the value, we just fix things by
-// going back to hard-coding it.
-// export const manageSubscriptionRedirectURL = `${window.context.frontendCodyProConfig?.sscBaseUrl}/subscription`
-export const manageSubscriptionRedirectURL = 'https://accounts.sourcegraph.com/cody/subscription'
+// URL the user needs to navigate to in4 order to modify their Cody Pro subscription.
```
chrsmith, line 4:
```suggestion
// URL the user needs to navigate to in order to modify their Cody Pro subscription.
```

client/web/src/cody/util.ts
```diff
@@ -1,14 +1,10 @@
...
+export const manageSubscriptionRedirectURL = `${
+    window.context.frontendCodyProConfig?.sscBaseUrl || 'https://accounts.sourcegraph.com/cody'
```
chrsmith, line 6: I would suggest you make this a function. Even if it isn't strictly required.

By making this a constant expression, the value is set at some "random but still deterministic time", by virtue of whatever module happens to import whatever module happens to import this `util` module. Right?

That makes me a little nervous, because depending on when the `window.context` value is populated, it could be the case that this is set incorrectly for the lifetime of the "React browser app session" or whatever we call it.

It could be the case that we _happen_ to load the `window.context` in the page `<head>` or ensure that it is before the React app loads... but I'd personally suggest we err on the side of being robust here. And not implicitly couple our behavior with the initialization order for the `window.context` data.

chrsmith (approved):
This looks good to me, although with a small suggestion.

I agree with your thoughts on other ways to set the "defaults" for the `codyProConfig` data, and just updating the logic here in the code is probably the right call for now. (And after we ship the embedded UI, we can remove some of the logic here entirely, since it will just be a static route to some other page in the same React app.)

client/web/src/cody/util.ts
vdavid: This is a non-issue. `window` is created prior to any execution context. See http://www.ecma-international.org/ecma-262/5.1/#sec-15.1

chrsmith: Sorry, I should have been clearer. I'm specifically wondering about the `context` field. That's the thing that we use to wire up configuration data from the backend, and "smuggle" it into the frontend JavaScript code. Right?

So it's possible that `window.context.frontendCodyProConfig` _would_ be set, but because whatever loading process hadn't completed yet, `window.context` is still left as `undefined`.

vdavid: Ah, sorry, it was quite late when I answered and apparently I was more tired than I felt.

It's still a non-issue, though. We load the site config [synchronously](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/client/web/dev/utils/get-site-config.ts?L15) and [insert](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/client/web/dev/utils/get-index-html.ts?L36) it directly into the index.html served. For direct proof, see `view-source:https://sourcegraph.com/search`, it's line 42, the entire context is there, set synchronously, _before_ the React app is served.

PR #62858, opened 2024-05-22
SSC: Fix invite date formatting

vdavid:
This is a step toward https://github.com/sourcegraph/self-serve-cody/issues/725

The list of invitations used to look like this:

CleanShot 2024-05-22 at 17 16 15@2x

Now it looks like this:

CleanShot 2024-05-22 at 17 17 19@2x

which intends to match [this design](https://www.figma.com/design/FMSdn1oKccJRHQPgf7053o/Cody-PLG-GA?node-id=4658-18897&t=m1L9vFMX0uBbl2Y0-0).

Some other examples with the current solution are:
- Invite sent 5 seconds ago
- Invite sent 1 hour ago
- Invite sent yesterday
- Invite sent last week
@rrhyne, is this format fine?

## Test plan

- Added tests
- Tested it manually, it renders well.

### DIFF
--- a/client/web/src/cody/team/TeamMemberList.test.ts
+++ b/client/web/src/cody/team/TeamMemberList.test.ts
@@ -1,0 +1,29 @@
+import { describe, expect, it } from 'vitest'
+import { formatInviteDate } from './TeamMemberList'
+
+// This is a bit weird test, but I wanted to give the design some examples for how the formatting looks,
+// and this was the easiest way to do that. Rather than throwing away the test, I'm leaving it here.
+describe('getFilterFnsFromCodyContextFilters', () => {
+    it('looks good for typical input', async () => {
+        // These are in the format of the normal output of our Go back end
+        const inputDates = [
+            '2024-05-22T15:59:55.000000+00:00',
+            '2024-05-22T14:17:55.000000+00:00',
+            '2024-05-21T14:17:55.000000+00:00',
+            '2024-05-15T14:17:55.000000+00:00',
+            ]
+
+        const now = new Date('2024-05-22T16:00:00.000000+00:00')
+
+        const expectedOutput = [
+            '5 seconds ago',
+            '1 hour ago',
+            'yesterday',
+            'last week',
+            ]
+
+        const outputDates = inputDates.map(date => formatInviteDate(date, now))
+
+        expect(outputDates).toEqual(expectedOutput)
+    })
+})
--- a/client/web/src/cody/team/TeamMemberList.test.ts
+++ b/client/web/src/cody/team/TeamMemberList.test.ts
@@ -1,49 +1,33 @@
 import { describe, expect, it } from 'vitest'
+
 import { formatInviteDate } from './TeamMemberList'
 
-// This is a bit weird test, but I wanted to give the design some examples for how the formatting looks,
-// and this was the easiest way to do that. Rather than throwing away the test, I'm leaving it here.
-describe('getFilterFnsFromCodyContextFilters', () => {
-    it('looks good for typical input', async () => {
-        // These are in the format of the normal output of our Go back end
+describe('formatInviteDate', () => {
+    it('shows relative descriptions of time in the desired format', async () => {
+        // ISO-8601 (RFC3339) strings, just the way the backend returns them
         const inputDates = [
             '2024-05-22T15:59:55.000000+00:00',
             '2024-05-22T14:17:55.000000+00:00',
             '2024-05-21T14:17:55.000000+00:00',
             '2024-05-15T14:17:55.000000+00:00',
-            ]
+        ]
 
         const now = new Date('2024-05-22T16:00:00.000000+00:00')
 
-        const expectedOutput = [
-            '5 seconds ago',
-            '1 hour ago',
-            'yesterday',
-            'last week',
-            ]
+        const expectedOutput = ['5 seconds ago', '1 hour ago', 'yesterday', 'last week']
 
         const outputDates = inputDates.map(date => formatInviteDate(date, now))
 
         expect(outputDates).toEqual(expectedOutput)
     })
 
-    it('does not fail too badly with atypical input', async () => {
+    it('handles malformed input', async () => {
         // These are in the format of the normal output of our Go back end
-        const inputDates = [
-            null,
-            '',
-            '1T14:17:55.000000+00:00',
-            '2024-05-15T14:17:55',
-            ]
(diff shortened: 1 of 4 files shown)


client/web/src/cody/team/TeamMemberList.test.ts
```diff
@@ -0,0 +1,52 @@
+import { describe, expect, it } from 'vitest'
+import { formatInviteDate } from './TeamMemberList'
+
+// This is a bit weird test, but I wanted to give the design some examples for how the formatting looks,
+// and this was the easiest way to do that. Rather than throwing away the test, I'm leaving it here.
+describe('getFilterFnsFromCodyContextFilters', () => {
+    it('looks good for typical input', async () => {
+        // These are in the format of the normal output of our Go back end
```
chrsmith, line 8: "in the format of the normal output of our Go back end" is a bit too wishy washy. There is a precise format that we expect, and so we should call that out.

Read: broadly speaking, our REST API types from the backend represent all dates as ISO-8601 strings, specifically using the RFC 3339 profile. (Which you are parsing in the `formatInviteDate` function.)

- https://stackoverflow.com/questions/522251/whats-the-difference-between-iso-8601-and-rfc-3339-date-formats
- https://go.dev/src/time/format.go
- https://en.wikipedia.org/wiki/ISO_8601

```suggestion
        // The backend returns times as ISO-8601 (RFC3339) strings.
```

client/web/src/cody/team/TeamMemberList.test.ts
```diff
@@ -0,0 +1,52 @@
...
+        const expectedOutput = [
+            '5 seconds ago',
+            '1 hour ago',
+            'yesterday',
+            'last week',
+            ]
+
+        const outputDates = inputDates.map(date => formatInviteDate(date, now))
+
+        expect(outputDates).toEqual(expectedOutput)
```
chrsmith, line 27: Nit: The vertical whitespace here isn't really adding anything, and just making it harder to scroll through the code. Same with the function below.

```suggestion
            ]
        const outputDates = inputDates.map(date => formatInviteDate(date, now))
        expect(outputDates).toEqual(expectedOutput)
```

client/web/src/cody/team/TeamMemberList.test.ts
```diff
@@ -0,0 +1,52 @@
...
+        const inputDates = [
+            null,
+            '',
+            '1T14:17:55.000000+00:00',
+            '2024-05-15T14:17:55',
+            ]
+
+        const now = new Date('2024-05-22T16:00:00.000000+00:00')
+
+        const expectedOutput = [
```
chrsmith, line 41: This is fine as-is, but what do you think about declaring the "input" and "expected output" side-by side? That would make it much easier to read, match up the expected values.

Also, since there is no functional difference between the "positive" test case and this "negative" one, maybe we would just want to put them into a single `it` block?

```ts

const tests: { input: string | null, wantOutput: string }[] = {
    // Negative tests.
    { input: null, wantOutput: '' },
    { input: '', wantOutput: '' },
    { input: '1T14:17:55.000000+00:00', wantOutput: '' },
    ...
}

for (const test of tests) {
    ...
}

File: client/web/src/cody/team/TeamMemberList.tsx [hunk 4]
```diff
@@ -34,6 +35,16 @@ export interface TeamInvite {
     acceptedAt: string | null
 }
 
+// This little function is extracted to make it testable.
+// Same for the "now" parameter.
+export const formatInviteDate = (sentAt: string | null, now?: Date): string => {
+    try {
+        return sentAt ? intlFormatDistance(sentAt || '', now ?? new Date()) : ''
```
chrsmith, line 42: I find code written in this "point-free"-like style to be extremely difficult to read.

Can you please break this expression up into multiple steps? e.g. using an `if` statement instead of the ternary operator? And creating a local variable named `now` rather than using the nullish coalescing operator?

It is absolutely OK to have a function be comprised of more than one line of code. (Or even have mutable local variables!) And it will make a huge difference in how easy the code is to understand and maintain!

chrsmith (changes requested):
This PR looks great, and I am happy to see you added unit tests for the new functionality.

I'm just blocking the approval because of that excessively complicated one-liner. If we need to have a larger conversation about how many language features in a single line of code is too many, we can have that. But in short, we should optimize for making our code easier to understand and maintain. Getting fancy with syntax or language features is counter to that goal. And even if there were efficiency gains to be made in that style, I'd still maintain that the increased complexity and cost of maintenance is not worth it.

client/web/src/cody/team/TeamMemberList.test.ts
vdavid: My mental model is Arrange-Act-Assert, and my intention with the blank line is to clearly separate Act from Assert. I imagine it'd help me a lot when reading the test.

client/web/src/cody/team/TeamMemberList.test.ts
vdavid: Yeah, that approach seems equally good.

I like the ease of `expect(outputDates).toEqual(expectedOutput)` in the end, so leaving it as-is.

I think it's easier to understand with the two test suites being separate so I'm keeping it as-is, too.

client/web/src/cody/team/TeamMemberList.tsx
vdavid: I disagree but to move forward, converted it to an if-else.

taras-yemets (approved):

rrhyne (approved):
LGTM

client/web/src/cody/team/TeamMemberList.tsx
```diff
@@ -34,6 +35,18 @@ export interface TeamInvite {
     acceptedAt: string | null
 }
 
+// This tiny function is extracted to make it testable. Same for the "now" parameter.
+export const formatInviteDate = (sentAt: string | null, now?: Date): string => {
+    try {
+        if (sentAt) {
+            return intlFormatDistance(sentAt || '', now ?? new Date())
```
chrsmith, line 42: Since we already have the `if (sentAt)` the TypeScript compiler knows that the value is truth, and won't give us grief that the type could be `null` right?

So can't we just write this as:

```suggestion
            return intlFormatDistance(sentAt, now ?? new Date())
```

chrsmith (approved):
Thank you for making the suggested changes. 🙏

PR #62860, opened 2024-05-22
SSC: Fix notification colors

vdavid:
Our notifications had no dark theme style, and they weren't design-approved.
This PR makes them match the [figma](https://www.figma.com/design/FMSdn1oKccJRHQPgf7053o/Cody-PLG-GA?node-id=4600-18103&t=0WJfuMjhfhwqZokw-0).

Caveats:
- I had no design for the error notification style, so I came up with my best reds that matched the other colors. I also still have no icon for that style.
- In the figma, there were illustrations on the left side of the notifications, but I assumed those were only for specific messages, so I left the checkmark and the crappy X I made.

Light theme looks:

CleanShot 2024-05-22 at 17 36 01@2x

Dark theme looks:

CleanShot 2024-05-22 at 17 34 52@2x

## Test plan

Tested with a temporary state to demo. See the screenshots.

### DIFF
--- a/client/web/src/cody/team/CodyManageTeamPage.module.scss
+++ b/client/web/src/cody/team/CodyManageTeamPage.module.scss
@@ -8,7 +8,7 @@
 .alert {
     position: relative;
     padding: 1.5rem 1.5rem 1.5rem 5rem;
-    color: #181b26;
+    color: var(--gray-11);
     &::after {
         content: '';
         position: absolute;
@@ -22,9 +22,11 @@
 }
 
 .purple-success-alert {
-    background-color: #e8d1ff;
+    background-color: var(--violet-02);
+    color: var(--violet-06);
     :global(.theme-dark) & {
-        background-color: #d68cf3;
+        background-color: var(--pink);
+        color: var(--gray-08);
     }
     &::after {
         // Purple checkmark SVG
@@ -57,7 +59,7 @@
 .user-badges {
     position: relative;
     left: -31px;
-    border: 1px solid #e6ebf2;
+    border: 1px solid var(--gray-03);
     border-radius: 50%;
 }
 
--- a/client/web/src/cody/team/CodyManageTeamPage.module.scss
+++ b/client/web/src/cody/team/CodyManageTeamPage.module.scss
@@ -22,15 +22,21 @@
 }
 
 .purple-success-alert {
-    background-color: #f2e7fb;
+    background-color: #e8d1ff;
+    :global(.theme-dark) & {
+        background-color: #d68cf3;
+    }
     &::after {
         // Purple checkmark SVG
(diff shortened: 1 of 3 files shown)


chrsmith (commented):
I'll defer to Rob for approval. But this all looks good to me.

QQ: For common things like `purple-success-alert`, is `CodyManageTeamPage.module.scss` the right place to define these things?

Would we be better off if we just imported these styles from another CSS file?

e.g. maybe we keep `CodyManageTeamPage.module.scss` but just have an `@import '../alerts';` statement so we don't need to duplicate these if we need to pop up alerts on another page?

vdavid:
My thinking is this:
- Until we actually use these on other pages, it makes more sense to keep these with the page they belong to.
- Once we want to use them on other pages then it'd be very awkward to keep them with this page, so we'll naturally extract them.

So rather than future-proofing now, I want to organically transform the structure. It's very cheap, and the benefit is that each state is a valid one, plus, if we change our plans, we're still in a good place.

vdavid:
I've updated the text colors:

CleanShot 2024-05-23 at 13 40 17@2x

CleanShot 2024-05-23 at 13 40 38@2x

But then, with an eyedropper Chrome plugin, I couldn't confirm the _real_ colors that get rendered because it seems to be broken even for the figma 🤯:

CleanShot 2024-05-23 at 13 43 00@2x

But then, in the CSS, I use the correct dark theme background colors: `#d68cf3` and `#51cf66`. So they should be correct even if they render differently by the time the screenshot reaches the figma.

I've also replaced some literals with the matching CSS variables.

vdavid:
Okay, just realized I made a mistake with not updating the checkmark icon and some text colors. Now it's like this:

CleanShot 2024-05-23 at 14 28 06@2x

CleanShot 2024-05-23 at 14 26 23@2x

The reds are a best-effort.

rrhyne:
Thanks. I did not specify a red version anywhere, I think we should remove that for now.

vdavid:
...or, broader question because I might be off-track with the whole notification direction: how should I communicate it when an action (like sending the invites) fails?

rrhyne:
Roger. I put together these warning states:

https://www.figma.com/design/FMSdn1oKccJRHQPgf7053o/Cody-PLG-GA?node-id=5082-11931&t=35JUr8zVHPGQAO13-4

We shouldn't go all red there as there isn't anything critical happening.

vdavid:
@rrhyne the updated designs, now with the new icons:

CleanShot 2024-05-23 at 23 39 33@2x

CleanShot 2024-05-23 at 23 40 05@2x

If it looks roughly right, could you approve? I can handle further tweaks in a follow-up, but I want to use these notifications on other pages, and for that, I need to do a refactor which I'd want to do from these new designs.

taras-yemets (approved):

PR #62862, opened 2024-05-22
Svelte: linkify and wrap dir entries

camdencheek:
This makes each dir entry in the file popover header a link, and allows wrapping so the full path is always visible. Uses the same wrapping logic as the file header so the slashes work right. Additionally, it linkifies the commit message to point to the same page as the OID link

Fixes SRCH-125, SRCH-126, SRCH-127, and SRCH-98

# Test plan

https://github.com/sourcegraph/sourcegraph/assets/12631702/2e773a3c-10c2-41d4-8ada-c2c6b43d5a9c

### DIFF
--- a/client/web-sveltekit/src/lib/repo/filePopover/FilePopover.svelte
+++ b/client/web-sveltekit/src/lib/repo/filePopover/FilePopover.svelte
@@ -67,17 +67,17 @@
             Ideally we'd be able to use `break-after: avoid;`, but that's not widely supported.
         -->
         {#each displayRepoName(repoName).split('/') as repoFragment, i}
-            <div>
-                {#if i > 0}<div>/</div>{/if}
-                <div>{repoFragment}</div>
-            </div>
+            <span>
+                {#if i > 0}<span>/</span>{/if}
+                <span>{repoFragment}</span>
+            </span>
         {/each}
-        {#if dirNameBreadcrumbs.length}<div>·</div>{/if}
+        {#if dirNameBreadcrumbs.length}<span>·</span>{/if}
         {#each dirNameBreadcrumbs as [name, href], i}
-            <div>
-                {#if i > 0}<div>/</div>{/if}
-                <div><a {href}>{name}</a></div>
-            </div>
+            <span>
+                {#if i > 0}<span>/</span>{/if}
+                <span><a {href}>{name}</a></span>
+            </span>
         {/each}
     </div>
 
@@ -105,7 +105,7 @@
         {/if}
     </div>
 
-    <div class="last-changed section">Last Changed @</div>
+    <div class="last-changed section">Last Changed</div>
 
     <div class="commit">
         <div class="node-line"><NodeLine /></div>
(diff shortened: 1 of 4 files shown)


taiyab (commented):
Sweet, we can just condense the gap between the slashes and dir names slightly and we're good to go!

client/web-sveltekit/src/lib/repo/filePopover/FilePopover.svelte
```diff
@@ -38,24 +39,46 @@
...
+                {#if i > 0}<div>/</div>{/if}
+                <div>{repoFragment}</div>
+            </div>
+        {/each}
+        {#if dirNameBreadcrumbs.length}<div>·</div>{/if}
+        {#each dirNameBreadcrumbs as [name, href], i}
+            <div>
+                {#if i > 0}<div>/</div>{/if}
+                <div><a {href}>{name}</a></div>
+            </div>
```
fkling, line 80: I know it doesn't matter since `display` is set to `flex`, but I feel conceptually these should be `<span>`s, not `<div>`s (but maybe my preferences are outdated 😅 )

client/web-sveltekit/src/lib/repo/filePopover/FilePopover.svelte
```diff
@@ -38,24 +39,46 @@
...
-        return [parts.slice(0, parts.length - 1).join('/'), parts[parts.length - 1]]
+        return [parts.slice(0, parts.length - 1), parts[parts.length - 1]]
     }
 
-    $: [dirName, baseName] = splitPath(entry.path)
+    $: [dirNameEntries, baseName] = splitPath(entry.path)
+    $: dirNameBreadcrumbs = dirNameEntries.map((part, index, all): [string, string] => [
+        part,
+        resolveRoute(TREE_ROUTE_ID, {
+            repo: revision ? `${repoName}@${revision}` : repoName,
```
fkling, line 56: For future improvements: I wonder if we need something like `repoNameRevision`, passed from the data loader, so that we don't have to remember to check `revision` downstream.

fkling (approved):

client/web-sveltekit/src/lib/repo/filePopover/FilePopover.svelte
camdencheek: Makes sense to me. TIL that `span` can be `display: flex;` 😅

client/web-sveltekit/src/lib/repo/filePopover/FilePopover.svelte
camdencheek: Sounds nice!

client/web-sveltekit/src/lib/repo/filePopover/FilePopover.svelte
fkling: yes, this can be applied to any element. I don't know whether it's still relevant to think of inline and black elements these days, but I still do. Basically whenever something code be displayed alongside some other text I think of it as inline, and use `span` and if necessary `inline-flex` to reflect that.

PR #62864, opened 2024-05-22
worker: Reduce frequency of very frequently run jobs

eseliger:
I looked at background routines and noticed that many routines are running _very_ frequently, which can cause higher than necessary load, and also stress on redis to store all these runs.

I tried to put in some more sensible numbers (aligning it with what most other routines seem to use). For many instances, this will mean a 30x reduction in job invocations.

Test plan:

Code review from owners, and will watch S2 and dotcom for potential issues.

### DIFF
--- a/cmd/worker/internal/encryption/config.go
+++ b/cmd/worker/internal/encryption/config.go
@@ -17,7 +17,7 @@ type config struct {
 var ConfigInst = &config{}
 
 func (c *config) Load() {
-	c.EncryptionInterval = c.GetInterval("RECORD_ENCRYPTER_INTERVAL", "1s", "How frequently to encrypt/decrypt a batch of records in the database.")
-	c.MetricsInterval = c.GetInterval("RECORD_ENCRYPTER_METRICS_INTERVAL", "10s", "How frequently to update progress metrics related to encryption/decryption.")
+	c.EncryptionInterval = c.GetInterval("RECORD_ENCRYPTER_INTERVAL", "10s", "How frequently to encrypt/decrypt a batch of records in the database.")
+	c.MetricsInterval = c.GetInterval("RECORD_ENCRYPTER_METRICS_INTERVAL", "30s", "How frequently to update progress metrics related to encryption/decryption.")
 	c.Decrypt = c.GetBool("ALLOW_DECRYPTION", "false", "If true, encrypted records will be decrypted and stored in plaintext.")
 }
--- a/internal/codeintel/ranking/internal/background/coordinator/config.go
+++ b/internal/codeintel/ranking/internal/background/coordinator/config.go
@@ -13,5 +13,5 @@ type Config struct {
 }
 
 func (c *Config) Load() {
-	c.Interval = c.GetInterval("CODEINTEL_RANKING_COORDINATOR_INTERVAL", "1s", "How frequently to run the ranking coordinator.")
+	c.Interval = c.GetInterval("CODEINTEL_RANKING_COORDINATOR_INTERVAL", "30s", "How frequently to run the ranking coordinator.")
 }
--- a/internal/codeintel/uploads/internal/background/backfiller/config.go
+++ b/internal/codeintel/uploads/internal/background/backfiller/config.go
@@ -14,6 +14,6 @@ type Config struct {
 }
 
 func (c *Config) Load() {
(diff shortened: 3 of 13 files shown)


eseliger:
* **#62864** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62864?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

cmd/worker/internal/encryption/config.go
```diff
@@ -17,7 +17,7 @@ type config struct {
 var ConfigInst = &config{}
 
 func (c *config) Load() {
-	c.EncryptionInterval = c.GetInterval("RECORD_ENCRYPTER_INTERVAL", "1s", "How frequently to encrypt/decrypt a batch of records in the database.")
-	c.MetricsInterval = c.GetInterval("RECORD_ENCRYPTER_METRICS_INTERVAL", "10s", "How frequently to update progress metrics related to encryption/decryption.")
+	c.EncryptionInterval = c.GetInterval("RECORD_ENCRYPTER_INTERVAL", "10s", "How frequently to encrypt/decrypt a batch of records in the database.")
```
eseliger, line 20: This is only relevant for when encryption is initially enabled, or getting disabled. Records written while encryption is on will be encrypted right away. This should be more than sensible.

cmd/worker/internal/encryption/config.go
```diff
@@ -17,7 +17,7 @@ type config struct {
...
+	c.MetricsInterval = c.GetInterval("RECORD_ENCRYPTER_METRICS_INTERVAL", "30s", "How frequently to update progress metrics related to encryption/decryption.")
```
eseliger, line 21: Just metrics, should be fine to lag a bit more behind.

internal/codeintel/ranking/internal/background/coordinator/config.go
```diff
@@ -13,5 +13,5 @@ type Config struct {
 }
 
 func (c *Config) Load() {
-	c.Interval = c.GetInterval("CODEINTEL_RANKING_COORDINATOR_INTERVAL", "1s", "How frequently to run the ranking coordinator.")
+	c.Interval = c.GetInterval("CODEINTEL_RANKING_COORDINATOR_INTERVAL", "30s", "How frequently to run the ranking coordinator.")
```
eseliger, line 16: This and the following ranking configs are not doing any work on instances as the feature is disabled everywhere, but we still run the background routines, so this will reduce frequency but not affect any functionality.

internal/codeintel/uploads/internal/background/backfiller/config.go
```diff
@@ -14,6 +14,6 @@ type Config struct {
 }
 
 func (c *Config) Load() {
-	c.Interval = c.GetInterval("CODEINTEL_UPLOAD_BACKFILLER_INTERVAL", "10s", "The frequency with which to run periodic codeintel backfiller tasks.")
+	c.Interval = c.GetInterval("CODEINTEL_UPLOAD_BACKFILLER_INTERVAL", "30s", "The frequency with which to run periodic codeintel backfiller tasks.")
```
eseliger, line 17: Does this seem sensible to Graph?

internal/codeintel/uploads/internal/background/expirer/config.go
```diff
@@ -27,7 +27,7 @@ func (c *Config) Load() {
 	uploadProcessDelay := env.ChooseFallbackVariableName("CODEINTEL_UPLOAD_EXPIRER_UPLOAD_PROCESS_DELAY", "PRECISE_CODE_INTEL_RETENTION_UPLOAD_PROCESS_DELAY")
 
 	c.CommitBatchSize = c.GetInt(commitBatchSize, "100", "The number of commits to process per upload at a time.")
-	c.ExpirerInterval = c.GetInterval("CODEINTEL_UPLOAD_EXPIRER_INTERVAL", "1s", "How frequently to run the upload expirer routine.")
+	c.ExpirerInterval = c.GetInterval("CODEINTEL_UPLOAD_EXPIRER_INTERVAL", "30s", "How frequently to run the upload expirer routine.")
```
eseliger, line 30: Does this seem sensible to Graph?

keynmol (approved):
I think all of these are fine - my guess is that all those `1s` are remnants from the time when the feature was being developed, where local feedback loop needed to be tight.

Don't a see a reason for any of those frequencies to be 1s.

PR #62865, opened 2024-05-22
Svelte: add repo popovers to repo name in header and in dynamic filters

camdencheek:
This adds the `RepoPopover` hover info to the dynamic filters in the search sidebar (replacing the full repo name tooltip) and to the global header.

# Test plan

https://github.com/sourcegraph/sourcegraph/assets/12631702/0c16a1bc-bd0c-4589-af5f-61815956f36f

### DIFF
--- a/client/web-sveltekit/src/lib/search/dynamicFilters/Section.svelte
+++ b/client/web-sveltekit/src/lib/search/dynamicFilters/Section.svelte
@@ -30,7 +30,13 @@
         {/if}
         <ul>
             {#each limitedItems as item}
-                <li><slot name="item" {item}><SectionItem {item} {onFilterSelect} /></slot></li>
+                <li>
+                    <slot name="item" {item}>
+                        <SectionItem {item} {onFilterSelect}>
+                            <slot name="label" slot="label" let:label let:value {label} {value} />
+                        </SectionItem>
+                    </slot>
+                </li>
             {/each}
         </ul>
         {#if filteredItems.length === 0}
--- a/client/web-sveltekit/src/lib/search/dynamicFilters/Section.svelte
+++ b/client/web-sveltekit/src/lib/search/dynamicFilters/Section.svelte
@@ -32,9 +32,15 @@
             {#each limitedItems as item}
                 <li>
                     <slot name="item" {item}>
-                        <SectionItem {item} {onFilterSelect}>
-                            <slot name="label" slot="label" let:label let:value {label} {value} />
-                        </SectionItem>
+                        {#if $$slots.label}
+                            <SectionItem {item} {onFilterSelect}>
+                                <svelte:fragment slot="label" let:label let:value>
+                                    <slot name="label" {label} {value} />
+                                </svelte:fragment>
+                            </SectionItem>
+                        {:else}
+                            <SectionItem {item} {onFilterSelect} />
+                        {/if}
                     </slot>
                 </li>
             {/each}
--- a/client/web-sveltekit/src/lib/search/dynamicFilters/Section.svelte
+++ b/client/web-sveltekit/src/lib/search/dynamicFilters/Section.svelte
@@ -1,14 +1,10 @@
 <script lang="ts">
-    import { mdiClose } from '@mdi/js'
-
-    import { page } from '$app/stores'
-    import Icon from '$lib/Icon.svelte'
     import { Button } from '$lib/wildcard'
 
-    import CountBadge from './CountBadge.svelte'
-    import { updateFilterInURL, type SectionItem } from './index'
+    import type { SectionItemData } from './index.ts'
+    import SectionItem from './SectionItem.svelte'
 
-    export let items: SectionItem[]
+    export let items: SectionItemData[]
     export let title: string
     export let filterPlaceholder: string = ''
     export let showAll: boolean = false
@@ -34,25 +30,7 @@
         {/if}
         <ul>
             {#each limitedItems as item}
-                <li>
-                    <a
-                        href={updateFilterInURL($page.url, item, item.selected).toString()}
-                        class:selected={item.selected}
-                        on:click={() => onFilterSelect(item.kind)}
-                    >
-                        <span class="label">
-                            <slot name="label" label={item.label} value={item.value}>
-                                {item.label}
-                            </slot>
-                        </span>
-                        <CountBadge count={item.count} exhaustive={item.exhaustive} />
-                        {#if item.selected}
-                            <span class="close">
-                                <Icon svgPath={mdiClose} inline />
-                            </span>
-                        {/if}
-                    </a>
-                </li>
+                <li><slot name="item" {item}><SectionItem {item} {onFilterSelect} /></slot></li>
             {/each}
         </ul>
         {#if filteredItems.length === 0}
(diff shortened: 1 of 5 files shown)


client/web-sveltekit/src/lib/search/dynamicFilters/Section.svelte
```diff
@@ -35,23 +31,11 @@
...
-                            <span class="close">
-                                <Icon svgPath={mdiClose} inline />
-                            </span>
-                        {/if}
-                    </a>
+                    <slot name="item" {item}>
+                        <SectionItem {item} {onFilterSelect}>
+                            <slot name="label" slot="label" let:label let:value {label} {value} />
+                        </SectionItem>
+                    </slot>
```
camdencheek, line 38: Extracted this into a new component to make it easier to create a slot with a default, but overridable, value.

client/web-sveltekit/src/lib/search/dynamicFilters/Sidebar.svelte
```diff
@@ -32,20 +32,24 @@
...
-    import Tooltip from '$lib/Tooltip.svelte'
+    import { delay } from '$lib/utils'
+    import { Alert } from '$lib/wildcard'
     import Button from '$lib/wildcard/Button.svelte'
 
     import HelpFooter from './HelpFooter.svelte'
     import {
         type URLQueryFilter,
-        type SectionItem,
+        type SectionItemData,
```
camdencheek, line 52: Renamed because I wanted a component named `SectionItem` now, so it makes sense to call this `SectionItemData` instead.

client/web-sveltekit/src/lib/search/dynamicFilters/Sidebar.svelte
```diff
@@ -119,13 +124,24 @@
...
+                    </div>
+                    <svelte:fragment slot="content">
+                        {#await delay(fetchRepoPopoverData(getGraphQLClient(), item.label), 200) then data}
+                            <RepoPopover {data} withHeader />
+                        {:catch error}
+                            <Alert size="slim" variant="danger">{error}</Alert>
+                        {/await}
+                    </svelte:fragment>
+                </Popover>
             </svelte:fragment>
```
camdencheek, line 145: We want to wrap the whole item in the popover, not just the repo name, because otherwise the popover placement varies depending on the length of the repo name.

taiyab:
The one on the repo header on the repo pages needs to be offset a little on Y-axis (down a little).

client/web-sveltekit/src/lib/search/dynamicFilters/Section.svelte
```diff
@@ -35,23 +31,11 @@
         <ul>
             {#each limitedItems as item}
```
vovakulikov, line 32: By the way, do we need to specify a key properly in this loop here?

client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte
```diff
@@ -60,7 +64,16 @@
 
 <GlobalHeaderPortal>
     <nav aria-label="repository">
-        <h1><a href="/{repoName}">{displayRepoName}</a></h1>
+        <Popover showOnHover placement="bottom-start" let:registerTrigger>
+            <h1 use:registerTrigger><a href="/{repoName}">{displayRepoName}</a></h1>
```
Comment (vovakulikov) [inline client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte:68]: I would add a hover state here, it's been a case that people don't understand that this H1 element is clickable, since we are going to have a popover element to this hover state also makes sense here IMO

client/web-sveltekit/src/lib/search/dynamicFilters/Section.svelte
```diff
@@ -35,23 +31,11 @@
...
-                        <CountBadge count={item.count} exhaustive={item.exhaustive} />
-                        {#if item.selected}
-                            <span class="close">
-                                <Icon svgPath={mdiClose} inline />
-                            </span>
-                        {/if}
-                    </a>
+                    <slot name="item" {item}>
+                        <SectionItem {item} {onFilterSelect}>
+                            <slot name="label" slot="label" let:label let:value {label} {value} />
```
vovakulikov, line 36: @camdencheek I think there will be a problem, try to check sections where you don't pass override slot for label, I belive there won't be any labels since you don't pass any implementation from consumer and this slot within slot statement doesn't allow to enable default slot layout in the SectionItem

vovakulikov: I think this should solve this

```svelte
<SectionItem {item} {onFilterSelect}>
   <svelte:fragment let:label let:value >
    <slot name="label" {label} {value} />
   </svelte:fragment>
</SectionItem>
```

vovakulikov: Nope, I think, in this case, a low-level `label` slot wouldn't work. 🤔

camdencheek: Yeha, this one was tricky. I want a way to just pass through a slot, but that's kinda a pain (though it gets much better in Svelte 5). Do you have an example of where this was causing a problem?

vovakulikov: @camdencheek In my case I don't see labels for by file and utility section items, I think other sections without slot passing from the sidebar component level also have this problem (like the authors section)
Screenshot 2024-05-29 at 22 11 43

camdencheek: Oh woah. Good catch, I really thought I checked that. Will convert to draft while I figure this out.

client/web-sveltekit/src/lib/search/dynamicFilters/Section.svelte
camdencheek: I don't think so since the array is immutable. But we should probably be explicit about it

client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte
Comment (camdencheek) [reply client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte:88]: Agree, but I think we should save this for the planned header redesign

client/web-sveltekit/src/lib/search/dynamicFilters/Section.svelte
camdencheek: Okay, fixed in the latest commit by inspecting `$$slots`

client/web-sveltekit/src/lib/search/dynamicFilters/Section.svelte
```diff
@@ -32,9 +32,15 @@
             {#each limitedItems as item}
                 <li>
                     <slot name="item" {item}>
-                        <SectionItem {item} {onFilterSelect}>
-                            <slot name="label" slot="label" let:label let:value {label} {value} />
-                        </SectionItem>
+                        {#if $$slots.label}
```
vovakulikov, line 35: Can this if statement live within SectionItems tags here? Or this would break the default slot as well?

vovakulikov (approved):

client/web-sveltekit/src/lib/search/dynamicFilters/Section.svelte
camdencheek: Unfortunately, it seems there's an edge case where slot rendering cannot be conditional within a component. So I had to copy the fully component in the conditional. Another thing that gets much nicer in svelte 5

PR #62866, opened 2024-05-22
svelte: Fix failing/skipped playwright tests

fkling:
Follow up for #62820 to fix failing/skipped tests that are easy to fix (which is all but one).

Re-enabling the syntax highlighting error test should be done in a separate PR.

## Test plan

`pnpm test`


fkling:
@camdencheek 

> Any takeaways other than "we should run playwright in CI"?

It's easy to break GraphQL operation overwrites by renaming the query. TypeScript won't help atm because the overwrite API also accepts arbitrary strings, so that we can overwrite requests made by shared code (i.e. graphql queries that are not defined inside `web-sveltekit`). Running the tests in CI would have caught that of course, but the error still wouldn't have been obvious.

OTOH, changing the name of queries might not happen so often that it warrants any other solution then finally running the tests in CI.

PR #62867, opened 2024-05-22
Svelte: add copy button and reduce spacing

camdencheek:
This adds a copy button to the file path in the file view and reduces spacing between path elements.

Related to SRCH-140

### DIFF
--- a/client/web-sveltekit/src/lib/repo/FileHeader.svelte
+++ b/client/web-sveltekit/src/lib/repo/FileHeader.svelte
@@ -6,6 +6,7 @@
     import Icon from '$lib/Icon.svelte'
     import { DropdownMenu } from '$lib/wildcard'
     import { getButtonClassName } from '$lib/wildcard/Button'
+    import CopyButton from '$lib/wildcard/CopyButton.svelte'
 
     const TREE_ROUTE_ID = '/[...repo=reporev]/(validrev)/(code)/-/tree/[...path]'
     const BLOB_ROUTE_ID = '/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]'
@@ -63,6 +64,7 @@
                 {/if}
             </span>
         {/each}
+        <span class="copy-button"><CopyButton value={path} label="Copy path to clipboard" /></span>
     </h2>
     <div class="actions" use:overflow={{ class: 'compact', measureClass: 'measure' }}>
         <slot name="actions" />
@@ -95,6 +97,14 @@
     }
 
     h2 {
+        display: flex;
+        flex-wrap: wrap;
+        gap: 0.375em;
+        span {
+            display: flex;
+            gap: inherit;
+        }
+
         font-weight: 400;
         font-size: var(--code-font-size);
         font-family: var(--code-font-family);
@@ -118,6 +128,14 @@
 
         .last {
             color: var(--text-title);
+        }
+
+        .copy-button {
+            visibility: hidden;
+            margin-left: 0.5em;
+        }
+        &:hover .copy-button {
+            visibility: visible;
         }
     }
 


client/web-sveltekit/src/lib/repo/FileHeader.svelte
```diff
@@ -92,9 +94,25 @@
...
     }
 
     h2 {
+        display: flex;
+        flex-wrap: wrap;
+        gap: 0.375em;
+        span {
+            display: flex;
+            gap: inherit;
+        }
```
camdencheek, line 114: Changed this to `display: flex;` so we can control the spacing between elements.

camdencheek:
> does this *feel* good?

The existence of a copy button feels good. It's way easier to use than select+copy. The fact that it shows up when you hover the actions feels a little weird, but is not really noticeable or problematic. And I think there's a good chance that will be (much more easily) fixable with the new designs for the header.

camdencheek:
I like the reduced spacing as well. The increased density is helpful for that space

fkling:
I'll take a look tomorrow.

> For the life of me, I could not figure out how to remove the spaces.

Have you tried removing the space in line 52? Although I'd think that it would also be ignored with `display: flex`, but not sure. I can try some things out tomorrow too. Copy button aside I still think selecting and copying text should work as expected.

camdencheek:
> Have you tried removing the space in line 52?

I have. Even with no whitespace in the final, rendered DOM, copying still adds spaces. I think it might actually be `display: flex` that's the issue. I also tried `display: inline-flex`, but with no luck. Would love to see what you find 🙂

vovakulikov:
@camdencheek Copied value is affected by display rule of the elements from which we're getting values for copy buffer. (sometimes, parents can change these display rules implicitly, like the `display: flex` changes display on the child elements on flex-item. Also HTML is important as well, see examples below.

### Inline elements with space in HTML
So for example this 
```html
<div>
   <span>Hello</span>
   <span>World</span>
</div>
```
will produce "Hello World" with space between words, because there is a new line in HTML between inline elements,

### Inline elements with no space
However this 
```html
<div>
   <span>Hello</span><span>World</span>
</div>
```
will produce "HelloWorld" since there is no separator in HTML between inline elements. 

### Non inline elements

```html
<div>
   <div>Hello</div>
   <div>World</div>
</div>
```
This will produce 
```
Hello 
Worlds
```
regardless of separators in HTML, so this
```html
<div>
   <div>Hello</div><div>World</div>
</div>
```
also produces just two lines of text
```
Hello 
Worlds
```
since text elements are non inline elements

### Non-inline elements because of the parent layout
```html
<div styles="display:flex">
   <span>Hello</span><span>World</span>
</div>
```
will produce "Hello world" since text elements again are non inline cause of their parent.

camdencheek:
Thanks for the info @vovakulikov! It seems like I was running into a combination of complicated CSS behavior and also weird svelte behavior. It's really ugly to remove the spaces from this component's markup, and I still can't figure out how to ignore the inline icon, which still adds a space. 

The best I could come up with was to add a copy event listener that strips the newlines. It feels hacky, but it works more consistently than I expect any DOM massaging will.

camdencheek:
Okay, also added playwright tests that cover this since it would be a super easy thing to regress on

client/web-sveltekit/src/lib/repo/FileHeader.svelte
```diff
@@ -26,10 +27,18 @@
             }
         ),
     ])
+
+    // HACK: we use a flexbox for the path and an inline icon, but we still want the copied path to be usable.
+    // This event handler removes the newlines surrounding slashes from the copied text.
+    function stripSpaces(event: ClipboardEvent) {
+        const selection = document.getSelection() ?? ''
+        event.clipboardData?.setData('text/plain', selection?.toString()?.replaceAll(/\n?\/\n?/g, '/'))
```
fkling, line 35: `selection` should always be defined, since you do `?? ''` above. So I think you can drop the `?` here:

```suggestion
        event.clipboardData?.setData('text/plain', selection.toString().replaceAll(/\n?\/\n?/g, '/'))
```

fkling (approved):
Thank you! Indeed it looks like setting `display` affects the value that is copied, which I find surprising but I guess we can't do anything about it.

PR #39840, opened 2022-08-02
OAuth: support automatic token refresh and retry logic for GitLab

miveronese:
#36148

This PR is result from working under the guidance and in collaboration with @unknwon (many thanks to his help!). 

## Some context:
- The main goal of the first PR is to have a foundation for the new  "refresh  token + retry request" mechanism and also implement it for the GiLab auth provider.
- a next PR will be opened to use this new flow for GitHub.

## Test plan
- Unit tests were added.
- Manual ("pseudo e2e") tests were done to confirm that _the request to create a new token was triggered as expected._ 
Note: for manual tests, the access token stored in the DB was made invalid. But a new token wasn’t actually retrieved to replace the invalid one. Why?  Because the broken token used to request a new token was just invalid, but not expired. The manual tests went as far as  the request  to create a new token was triggered.

The rest of the flow was covered by unit tests -- on those, the API response expected when a token is expired and a new one is requested was mocked.

PR #62869, opened 2024-05-22
Refresh SAMS access tokens as needed

chrsmith:
The `ssc.APIProxyHandler` is an HTTP handler which proxies HTTP requests from Sourcegraph.com to the SSC backend, but replacing "Sourcegraph.com" credentials from the request with the user's "SAMS access token".

However, it's possible (extremely likely, actually) that the SAMS access token associated with the Sourcegraph user account has expired. Because a user's browser session will be valid for XX days, while the SAMS access token generated on login is only valid for YY minutes.

This PR updates the `ssc.APIProxyHandler` so that if it sees that a user's SAMS access token has expired, it will try to generate a new OAuth access token for the user via their SAMS refresh token.

> ℹ️ The SAMS refresh token may also expire, but we don't actually know _when_. So it's possible that refreshing the user's access token will fail, and the SSC API Proxy will still return a 401 status code.
> 
> Meaning that even with this PR, the SSC UI will need to have some sort of "your credentials have expired, please logout and login again" error experience. Although with these changes, that should be far less frequent.

This PR just uses the existing support we have for refreshing OAuth access tokens (`oauthtoken.GetAccountRefreshAndStoreOAuthTokenFunc`), that is used for how the Sourcegraph instance authenticates with 3rd party identity providers (e.g. `internal/authz/providers/github`). So 🤞 this is everything we need?

## Results

Testing locally it seemed to work as expected. (As if there were any doubt!) But looking at the logs, there definitely is (an unavoidable?) data race in refreshing the token.

The frontend send 3x HTTP requests in rapid success, all 3 of which went through the SAMS OAuth refresh process. (So while this is "a little slow", it doesn't appear to be a problem.)

However, this _would_ be a problem if SAMS were to implement more aggressive logic around the reuse of OAuth refresh tokens. e.g. it _could_ choose to only honor an OAuth refresh token once, and return a 400 if a user tries to use the same refresh token a second time.

[The implementation](https://github.com/sourcegraph/sourcegraph/blob/a376a02bf55ef9e81506a7d6005fd434e790a43b/internal/oauthtoken/oauth_token_helper.go#L48) has some logic to account for this, but there may be some hypothetical scenario where it would surface as a 500 error. (But given that this isn't the case for SAMS, I'm inclined to not worry about it.)

```
[       frontend] INFO Handler.SSC Proxy ssc/ssc_proxy.go:235 proxying SSC API request {"url": "/.api/ssc/proxy/team/current/subscription"}
[       frontend] INFO Handler.SSC Proxy ssc/ssc_proxy.go:235 proxying SSC API request {"url": "/.api/ssc/proxy/team/current/subscription/summary"}
[       frontend] INFO Handler.SSC Proxy ssc/ssc_proxy.go:235 proxying SSC API request {"url": "/.api/ssc/proxy/team/current/members"}
[       frontend] WARN Handler.SSC Proxy ssc/ssc_proxy.go:265 the user's SAMS token has expired {"expiry": 1715989719041120000}
[       frontend] WARN Handler.SSC Proxy ssc/ssc_proxy.go:265 the user's SAMS token has expired {"expiry": 1715989719041120000}
[       frontend] INFO Handler.SSC Proxy ssc/ssc_proxy.go:235 proxying SSC API request {"url": "/.api/ssc/proxy/team/current/invites"}
[       frontend] WARN Handler.SSC Proxy ssc/ssc_proxy.go:265 the user's SAMS token has expired {"expiry": 1715989719041120000}
[       frontend] WARN Handler.SSC Proxy ssc/ssc_proxy.go:265 the user's SAMS token has expired {"expiry": 1715989719041120000}
[       frontend] INFO Handler.SSC Proxy ssc/ssc_proxy.go:229 refresh user's SAMS token {"new expiration": 1716413928649510000}
[       frontend] INFO Handler.SSC Proxy ssc/ssc_proxy.go:101 Building proxy request {"method": "GET", "url": "http://localhost:19992/cody/api/v1/team/current/subscription"}
[       frontend] INFO Handler.SSC Proxy ssc/ssc_proxy.go:229 refresh user's SAMS token {"new expiration": 1716413928649711000}
[       frontend] INFO Handler.SSC Proxy ssc/ssc_proxy.go:101 Building proxy request {"method": "GET", "url": "http://localhost:19992/cody/api/v1/team/current/invites"}
[       frontend] INFO Handler.SSC Proxy ssc/ssc_proxy.go:229 refresh user's SAMS token {"new expiration": 1716413928651633000}
[       frontend] INFO Handler.SSC Proxy ssc/ssc_proxy.go:101 Building proxy request {"method": "GET", "url": "http://localhost:19992/cody/api/v1/team/current/members"}
```

## Test plan

Tested manually.

I didn't add unit tests for the new functionality, since it would amount to a whole lot of mocking for not much benefit.


cmd/frontend/internal/httpapi/httpapi.go
```diff
@@ -227,11 +227,22 @@ func NewHandler(
...
 			// and its use of `CookieMiddlewareWithCSRFSafety`.)
+			samsOAuthConfig, err := ssc.GetSAMSOAuthContext(db)
+			if err != nil {
+				// This situation is pretty bad, as it means no Cody Pro-related functionality
+				// can work properly. So while the site can continue to load as expected,
+				// we will supply a zero-value OAuth config that will only serve 503s.
+				//
+				// This makes the failure a lot more obvious than not registering the routes
+				// at all, and trying to figure out why we are seeing 404s or 405s.
+				logger.Error("error loading SAMS config, unable to register SSC API proxy", sglog.Error(err))
```
bobheadxi, line 238: FYI we get Sentry for MSP but I don't think we have the same Sentry enabled for dotcom, as it looks very quiet: https://sourcegraph.sentry.io/projects/dot-com-backend/?project=5596888

If you want to be notified on this, you have to monitor logs directly, or we can look into re-enabling Sentry for dotcom. If we re-enable Sentry I'd like to be explicit about allowlisting only errors from certain logger scopes, as IIRC it was disabled because it was super noisy and nobody every bothered to look - in which case, maybe you can use `samsAuthenticator.logger` which has a scope so that we can leave this door open?

unknwon:
> This PR just uses the existing support we have for refreshing OAuth access tokens (`oauthtoken.GetAccountRefreshAndStoreOAuthTokenFunc`

If you care at all 😃 I wrote initial version of this helper hahaha https://github.com/sourcegraph/sourcegraph/pull/39840

cmd/frontend/internal/httpapi/httpapi.go
```diff
@@ -227,11 +227,22 @@ func NewHandler(
...
+			}
 			sscBackendProxy := ssc.APIProxyHandler{
-				CodyProConfig: conf.Get().Dotcom.CodyProConfig,
-				DB:            db,
-				Logger:        logger.Scoped("SSC Proxy"),
-				URLPrefix:     "/.api/ssc/proxy",
+				CodyProConfig:    conf.Get().Dotcom.CodyProConfig,
+				DB:               db,
+				Logger:           logger.Scoped("SSC Proxy"),
```
bobheadxi, line 243: Scope is meant to be machine-readable for filtering: 
```suggestion
				Logger:           logger.Scoped("sscProxy"),
```

cmd/frontend/internal/ssc/ssc_proxy.go
```diff
@@ -140,29 +178,65 @@ func (p *APIProxyHandler) getSAMSCredentialsForUser(ctx context.Context, userID
...
+		p.DB.UserExternalAccounts(), externalAccountID, p.SAMSOAuthContext)
+
+	// Perform the refresh.
+	userBearerToken := auth.OAuthBearerToken{
+		Token:        currentToken.AccessToken,
+		RefreshToken: currentToken.RefreshToken,
+		Expiry:       currentToken.Expiry,
+	}
+	newToken, _ /* newRefreshToken */, newExpiry, err := refreshFn(
+		ctx, http.DefaultClient, &userBearerToken)
```
bobheadxi, line 213: Should we be using one of the preconfigured clients that we have, e.g. `httpcli.UncachedExternalDoer`?

cmd/frontend/internal/ssc/ssc_proxy.go
```diff
@@ -40,10 +46,40 @@ type APIProxyHandler struct {
...
+	// access token.
+	SAMSOAuthContext *oauthutil.OAuthContext
 }
 
 var _ http.Handler = (*APIProxyHandler)(nil)
 
+// GetSAMSOAuthContext returns the OAuthContext object to describe the SAMS
+// IdP registered to the current Sourcegraph instance. (As identified by
+// `GETSAMSServiceID()`)
+func GetSAMSOAuthContext(db database.DB) (*oauthutil.OAuthContext, error) {
```
unknwon, line 61: `db` not used?

unknwon (approved):

cmd/frontend/internal/ssc/ssc_proxy.go
chrsmith: Oh, yeah. I cribbed that code from ssc/ssc.go and just assumed it pulled that data from the database. But yes, I'll remove that unused parameter. (Which 🤞 was caught by the linter, too.)

vdavid: Done in https://github.com/sourcegraph/sourcegraph/pull/62869/commits/ae6e9ea59eec1ebcf6c7c32755d4a10929b4a904

cmd/frontend/internal/ssc/ssc_proxy.go
vdavid: Thanks for the review and comment, @bobheadxi!
I need a bit of clarification, as Chris is out till Wednesday, and I don't understand the severity of the comment. Is it
a.) a functional issue (probably not as otherwise Joe probably wouldn't have approved the PR),
b.) a performance issue (then it might be worth it to fix it before merging), or
c.) more like a tech debt / DRY / elegance thing?

Based on the answer, what do you reckon is the best call here?
a.) We (likely Chris) do this as a follow-up. I can make sure we have a ticket so it's not forgotten.
b.) If you have time, you or Joe pair with me tomorrow to update this or at least give me some pointers, and then I test it again. It would probably take me an unreasonable amount of time to figure this out on my own.
c.) Something else?

Thanks :)

bobheadxi: If you look up the symbol I mentioned in my comment, you'll see it's just a shared library for standardized HTTP clients with the some preconfigured o11y, timeouts, retry handling, etc. The suggestion is to just a variable replacement to replace the `http.DefaultClient` here with one of those :)

vdavid: Ah, thanks, I can definitely do that :)

PR #62875, opened 2024-05-23
sg: cloud eph - api now requires env during list

burmudar:
Add `dev` environment to cloud ephemeral list instance call

## Test plan
Tested locally


filiphaftek (approved):
thx

Strum355 (approved):

PR #62870, opened 2024-05-22
codeintel-service-def

jdpleiness:
Service definition for code intel.

## Test plan

Golden files

PR #62876, opened 2024-05-23
appliance: deploy prometheus

craigfurman:
Closes https://linear.app/sourcegraph/issue/REL-15/service-definition-prometheus

This is a very large PR (sorry!). I highly recommend reviewing one commit at a time, reading the commit messages. I presented it as one, rather than a stack, because I suspect the history won't be _that_ useful after review, and can probably be squashed. Holler if this is just too much, I'm happy to split it up into a stack if need be.

Please see self-review notes too.

Changelog:

**New YAML package to hold yaml-related utilities**

First function: ConvertYAMLStringsToMultilineLiterals.

**appliance: remove getter method for pgsql config**

In favor of just accessing the package-level variable.

**appliance: add Prometheus config element**

Disable it in all current golden tests.

**appliance: deploy Prometheus**

**appliance: preserve multiline literals in golden fixtures**

**appliance: compare-helm: normalize yaml**

Standardize indentation and represent multiline strings as literals.
This makes diffs of large nested documents easier to read (e.g. the
prometheus configmap).

**appliance: compare-helm: allow cleanup to run when diff exits non-zero**

**appliance: prometheus privileged config**

Optionally provision RBAC with cluster-level privileges, and
corresponding scrape config.

Due to kubernetes rules around namespaced objects not owning namespaced
ones, our ConfigMap cannot own the ClusterRole(Binding)s provisioned by
this config. As such, they will not be garbage-collected when the
ConfigMap is deleted. These cluster-scoped resources are given a
qualified metadata.name, in order to minimise the risk of clashing with
existing non-namespaced resources.

**appliance: prometheus can optionally reference an existing configmap**

Rather than creating one.

## Test plan

Golden tests included.


internal/appliance/reconciler/kubernetes.go
```diff
@@ -75,8 +76,15 @@ func createOrUpdateObject[R client.Object](
...
-		return errors.Newf("setting controller reference: %w", err)
+	// Namespaced objects can't own non-namespaced objects. Trying to
+	// SetControllerReference on cluster-scoped resources gives the following
+	// error: "cluster-scoped resource must not have a namespace-scoped owner".
+	// non-namespaced resources will therefore not be garbage-collected when the
+	// ConfigMap is deleted.
+	if !isNamespaced(obj) {
+		if err := ctrl.SetControllerReference(owner, obj, r.Scheme); err != nil {
+			return errors.Newf("setting controller reference: %w", err)
+		}
```
craigfurman, line 87: Note

internal/appliance/config/embed.go
```diff
@@ -8,13 +8,9 @@ var (
...
 
-	pgsqlConfig []byte
+	PgsqlConfig []byte
 )
 
 func init() {
-	pgsqlConfig, _ = fs.ReadFile("pgsql/postgresql.conf")
-}
-
-func DefaultPGSQLConfig() string {
```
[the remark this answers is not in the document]
jdpleiness: I changed this to drop the getters in https://github.com/sourcegraph/sourcegraph/pull/62870 like discussed.

internal/appliance/reconciler/kubernetes.go
jdpleiness: Could we use a finalizer to clean these up? It might be difficult to manually keep "ownership" of it though 🤔

internal/k8s/resource/pod/pod.go
```diff
@@ -54,6 +54,20 @@ func NewVolumeFromPVC(name, claimName string) corev1.Volume {
...
 
+func NewVolumeFromConfigMap(name, configMapName string) corev1.Volume {
+	return corev1.Volume{
+		Name: name,
+		VolumeSource: corev1.VolumeSource{
+			ConfigMap: &corev1.ConfigMapVolumeSource{
+				LocalObjectReference: corev1.LocalObjectReference{
+					Name: configMapName,
+				},
+				DefaultMode: pointers.Ptr(int32(0777)),
```
jdpleiness, line 65:
```suggestion
				DefaultMode: pointers.Ptr[int32](0777),
```

internal/slices/sliceutils.go
```diff
@@ -0,0 +1,9 @@
+package slices
+
+func Map[S, T any](list []S, f func(S) T) []T {
```
jdpleiness, line 3: I'm kind of surprised we don't have this somewhere in the monorepo already 🤷

internal/yaml/yaml.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"gopkg.in/yaml.v3"
+)
+
+// ConvertYAMLStringsToMultilineLiterals is intended to make large nested yaml
+// strings more human-readable, and diffable (so mainly useful for tests). Note
+// that in order to do this reliably, it removed trailing whitespace from all
+// lines in multiline string fields.
+//
+// Do not use this function in contexts where that is problematic!
```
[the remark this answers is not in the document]
jdpleiness: My other thought would be to move it under `internal/appliance` if that's the only place we are using it

jdpleiness (approved):
😅 that was a long one, but not too bad. I made a few comments, but nothing blocking. I did skim more lightly than normal over the yaml testing pkg as well.

internal/yaml/yaml.go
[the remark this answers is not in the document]
craigfurman: Yeah, I think I will do that for now. It can always be moved to a more generic place later if wanted.

internal/slices/sliceutils.go
craigfurman: Yeah I was surprised too. I also keep thinking this was added to the new-ish slices and maps stdlib packages, but it hasn't been.

PR #62880, opened 2024-05-23
feat(search/svelte): Add context specific suggestions to repo search input

fkling:
This PR adds a new suggestions source to the repository search input. It currently offers the following suggestions:

- Repository group/org (on every repo page)
- Current file and/or directory (on blob and tree pages)
- Language of current file (on blob pages)

When on a commit page it adds the revision of the commit to the repo filter.

A new context store is introduced to propagate those values to the search input. I also had to update our query processing logic to preserve pattern and filter token ranges (first commit) so that we can also replace filters, not just add the.

NOTE: I'm not convinced the way the information is propagated via context is the best way to do, but it's the simplest I could think of now. I can imagine in the future that individual pages create/register suggestions directly.

https://github.com/sourcegraph/sourcegraph/assets/179026/7c1b78e3-061b-4dc8-a1ee-fb3273e633ba

## Test plan

Manual testing.


taiyab:
This is amazing! _(And you wanna hide the search bar?)_

camdencheek:
Design feedback: I find it pretty difficult to see that the text describing the suggestion is not part of the suggestion. My eye scans right past that text because it's not very high contrast compared to the suggestion text. 

The new designs also don't really fix this, and this scenario (a query snippet that is not just a filter, which is highlighted blue) is not covered in the new designs. @taiyab could we add some examples of "query snippet with a description" to the design?

screenshot-2024-06-07_07-51-40@2x

client/shared/src/search/query/analyze.test.ts
```diff
@@ -1,70 +1,122 @@
...
+            expect(parse('abc content:"with  space" def /regex literal/ ghi |')).toMatchInlineSnapshot(`
+              abc content:"with  space" def /regex literal/ ghi
+              ^^^ (pattern)
+                  ^^^^^^^ (filter.field: literal)
+                          ^^^^^^^^^^^^^ (filter.value: literal)
+                  ^^^^^^^^^^^^^^^^^^^^^ (filter)
+                                        ^^^ (pattern)
+                                            ^^^^^^^^^^^^^^^ (pattern)
+                                                            ^^^ (pattern)
+            `)
```
camdencheek, line 119: Oooh, pretty 🙂

client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte
```diff
@@ -59,6 +61,20 @@
         { path: '/-/batch-changes', label: 'Batch changes', visibility: 'admin' },
         { path: '/-/settings', icon: mdiCog, label: 'Settings', visibility: 'admin' },
     ]
+    const repositoryContext = writable<RepositoryPageContext>({})
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte:64]: Q: rather than all the `beforeNavigate`/`afterNavigate` hooks, why not populate `repositoryContext` with the information we already have from `data`? We should already have the current repo and current revision, and we can add current file in the child layout, right?

client/web-sveltekit/src/routes/[...repo=reporev]/context.ts
```diff
@@ -0,0 +1,23 @@
+import { getContext, setContext } from 'svelte'
+import type { Writable } from 'svelte/store'
+
+/**
+ * This context allows repository pages to propagte relevant information to other components.
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/context.ts:5]:
```suggestion
 * This context allows repository pages to propagate relevant information to other components.
```

taiyab:
@fkling Note: The search suggestions that are queries should probably have the same styling treatment as keyword search queries, if it was ran as a keyword search query (and not if it wasn't).

camdencheek (approved):
Played around with it locally -- works great, thank you!

client/web-sveltekit/src/lib/search/codemirror/suggestions.ts
```diff
@@ -0,0 +1,181 @@
...
+        // Example: github.com/sourcegraph/sourcegraph
+        //       -> repo:^github\.com/sourcegraph
+        //        repo:^github\.com/sourcegraph/sourcegraph$
+        // Example: sourcegraph/sourcegraph
+        //       -> repo:^sourcegraph
+        //          repo:^sourcegraph/sourcegraph$
+        (tokens: Token[], position: number, { repoName, revision }: ScopeInformation): Option[] => {
+            const options: Option[] = []
+
+            {
```
vovakulikov, line 76: I don't think we need these blocks. Since both have no overlap scoped-variables this should be fine, no?

client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte
Comment (vovakulikov) [reply client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte:64]: I don't know; I agree with Camden that this feels a bit off when the parent level should know about lower lever child data. This repository context is about to create yet another horizontal stream of data in a world where we already have not trivial data passing through data-loaders.  

I think I'm not against this idea about global context but maybe we should make it a bit more global in this case
rather than a specific repositoryContext thing.

vovakulikov (approved):
Left a few minor comments, nothing bit. I think though we should address our data flows as soon as possible while we have not that many data sources and fields. Complexity around contexts and different data types in data loaders look a bit scary every time I have to touch it. (maybe we can simplify it)

client/shared/src/search/query/analyze.ts
```diff
@@ -100,32 +100,62 @@ const operatorKindToKeywordKind: Record<OperatorKind, KeywordKind> = {
 }
 
 /**
- * Converts a parse node into a sequence of Token's
+ * Converts a parse node into a sequence of Token's. This function generates
+ * new tokens as needed to represent the parse tree in a flat list. Those
+ * tokens won't have useful ranges. Range information is only preserved for
+ * pattern and parameter nodes.
```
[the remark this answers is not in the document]
fkling: This is all part of `getRelevantTokens` which takes in a a tree and basically prunes nodes that are irrelevant. E.g. for the query `foo OR file:bar lang:JS baz`, if we only want to keep all `file:` and `lang:` filters, we would convert

````
    OR
   / \
 foo   AND
    /       \
 file:bar   AND
         /       \
       lang:file  baz
```
 
to just

```
      AND
   /       \
file:bar  lang:JS
```

Later we "flatten" the tree into a list of tokens again (which is what `tokenize` is doing).

So far we only had the need to read the list of tokens but now we also want to use them to change them in the original search query. But some of the tokens returned here are basically synthetic, due to the transformation.

Or maybe it makes more sense to think about it this way: We have an input query and transform it into another query. Some tokens exist in both queries and we want to make changes based on the original query based on the new query.
So maybe what should be returned is a list of tokens and some kind of source map.

client/web-sveltekit/src/lib/search/codemirror/suggestions.ts
fkling: I added them to not "pollute" the function scope. The variables used here are only relevant for this specific computation.

client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte
Comment (fkling) [reply client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte:64]: > why not populate `repositoryContext` with the information we already have from `data`

`data` will only return what the loader (and any parent loader) returns, not "child" loaders. We can access all page data via `$page.data` but that is not strongly typed. We can add optional fields to the `PageData` type, but that would then apply to all routes.
Would you find it less confusing if we used `$page.data` and `data` instead of `repositoryContext` and `data`?

The core issue IMO is that we want to propagate data up from pages to the layout. Context solves that but I understand that it's "yet another thing".

Comment (camdencheek) [reply client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte:64]: Because I think of this more as page data than component data (which I understand is kinda odd since pages are components), this feels like it belongs in page data loaders. We can make one of the pieces of page data writable so child pages can add their own information. In particular, the before/after hooks feel pretty weird since we're effectively manually implementing page data loading. 

I'm having difficulty describing what I mean, so I think it might be easier to just show it. [This draft](https://github.com/sourcegraph/sourcegraph/pull/63207/files) moves all the propagation into the page data loaders so the page itself doesn't need to worry about it. Personally, I think that feels cleaner since it reduces the amount that the pages themselves need to know about the data loading lifecycle.

Comment (camdencheek) [reply client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte:64]: (Also, feel free to just ignore me. I do not feel strongly about any of this, but wanted to at least bring it up in case I was missing something about how this needs to work)

Comment (fkling) [reply client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte:64]: @camdencheek I don't think this would work due to data preloading. It would cause the context to updated if the user hovers over a file without clicking it. Using the page lifecycle ensures that the context is only changed when the page is actually navigated to/away from.

As I said in the PR, it doesn't feel ideal to me either, but I can't think of anything else right now apart from extending the global `PageData` object.

Comment (camdencheek) [reply client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte:64]: Ah, interesting. I was thinking that preloading would create a _new_ data object with a _new_ `writable` that would be updated by its child data fetchers.

client/shared/src/search/query/analyze.ts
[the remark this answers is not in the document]
fkling: @vovakulikov, @camdencheek I changed the implementation of this. Please have another and let me know if it makes more sense now.

camdencheek (approved):
The new approach makes more sense to me! Thanks for the refactor

client/shared/src/search/query/analyze.ts
```diff
@@ -138,30 +211,64 @@ function tokenize(node: Node | null): Token[] {
...
     }
 }
+
+/**
+ * Mutates the range of the tokens in place to align them with the given offset.
+ * Returns the same tokens for convenience.
+ *
+ * @param tokens The tokens to shift
+ * @param offset The offset to shift the tokens by
+ * @returns The shifted tokens
```
fkling, line 248:
```suggestion
 * Mutates the range of the tokens in place to align them relative to each other.
 * Returns the same tokens for convenience.
 *
 * @param tokens The tokens to shift
 * @returns The passed in tokens
```

PR #62882, opened 2024-05-23
Update blobstore package to remove unused port + ssl properties

shivasurya:
- How was it PREVIOUSLY?
  The blobstore container exposed port with test ssl keys configured
- How it will be from NOW on.
   The blobstore will have just the http port (probably 9000) and we disabled the unused port(ssl)

Blobstore container had unused port open configured with test private keys. After [investigating](https://sourcegraph.slack.com/archives/C02UC4WUX1Q/p1716478617084139) about the unused port, we decided to remove config in the image from upstream `sourcegraph/s3proxy`

## Test plan

- CI 🟢 

Additional testing

```bash
docker exec -u 0 -it blobstore sh
curl -i -X PUT http://127.0.0.1:9000/create-bucket-test-2022-01-09-attempt0
```

### DIFF
--- a/wolfi-packages/s3proxy.yaml
+++ b/wolfi-packages/s3proxy.yaml
@@ -1,13 +1,13 @@
 package:
   name: s3proxy
   version: 2.0.0
-  epoch: 3 # Independent from repo epoch
-  description: "Access other storage backends via the S3 API"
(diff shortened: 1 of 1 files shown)


willdollman:
I'll merge this now as there's a bit of time pressure to hit the release on monday

PR #62883, opened 2024-05-23
feat/sg: add 'sg sams' commands 'create-client-token' and 'introspect-token'

bobheadxi:
Right now, developing SAMS clients involves raw cURL commands (see [operator cheat sheet](https://sourcegraph.notion.site/Sourcegraph-Accounts-infrastructure-operations-b90a571da30443a8b1e7c31ade3594fb)) (which is fine), but other steps like "testing auth" require using [accounts-clients-example](https://github.com/sourcegraph/sourcegraph-accounts/tree/main/cmd/accounts-client-example), which isn't very well documented and requires a bit of hand-wringing to get to and start using.

This migrates the next steps after using cURL to set up clients (`create-client-token` and `introspect-token`) to a new `sg sams` toolchain for better DX (docs, completions, flags)

## Test plan

```sh
export SG_SAMS_CLIENT_ID="..."
export SG_SAMS_CLIENT_SECRET="..."
sg sams create-client-token -s 'enterprise_portal::codyaccess::read'
```

### DIFF
--- a/dev/sg/sams/sg_sams.go
+++ b/dev/sg/sams/sg_sams.go
@@ -1,0 +1,122 @@
+// Package msp exports 'sg sams' commands for the Sourcegraph Accounts Management System.
+package sams
+
+import (
+	"encoding/json"
+
+	"github.com/urfave/cli/v2"
+	"golang.org/x/oauth2/clientcredentials"
+
+	sams "github.com/sourcegraph/sourcegraph-accounts-sdk-go"
+
+	"github.com/sourcegraph/sourcegraph/dev/sg/internal/category"
+	"github.com/sourcegraph/sourcegraph/dev/sg/internal/std"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+)
+
+var clientCredentialsFlags = []cli.Flag{
+	&cli.StringFlag{
+		Name:    "sams-server",
+		Aliases: []string{"sams"},
+		EnvVars: []string{"SG_SAMS_SERVER_URL"},
+		Value:   "https://accounts.sgdev.org",
+		Usage:   "URL of the Sourcegraph Accounts Management System (SAMS) server",
+	},
+	&cli.StringFlag{
+		Name:     "client-id",
+		EnvVars:  []string{"SG_SAMS_CLIENT_ID"},
+		Usage:    "Client ID of the Sourcegraph Accounts Management System (SAMS) client",
+		Required: true,
+	},
+	&cli.StringFlag{
+		Name:     "client-secret",
+		EnvVars:  []string{"SG_SAMS_CLIENT_SECRET"},
+		Usage:    "Client secret for the Sourcegraph Accounts Management System (SAMS) client",
+		Required: true,
+	},
+	&cli.StringSliceFlag{
+		Name:    "scopes",
+		Aliases: []string{"s"},
+		EnvVars: []string{"SAMS_SCOPES"},
+		Value:   cli.NewStringSlice("openid", "profile", "email"),
+		Usage:   "OAuth scopes ('$SERVICE::$PERM::$ACTION') to request from the Sourcegraph Accounts Management System (SAMS) server",
+	},
+}
+
+// newClientCredentialsFromFlags returns a new client credentials config from
+// clientCredentialsFlags.
+func newClientCredentialsFromFlags(c *cli.Context) *clientcredentials.Config {
+	return &clientcredentials.Config{
+		ClientID:     c.String("client-id"),
+		ClientSecret: c.String("client-secret"),
+		TokenURL:     c.String("sams-server") + "/oauth/token",
+		Scopes:       c.StringSlice("scopes"),
+	}
+}
+
+// Command is the 'sg sams' toolchain for the Sourcegraph Accounts Management System (SAMS).
+var Command = &cli.Command{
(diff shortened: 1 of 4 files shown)


dev/sg/sams/sg_sams.go
```diff
@@ -0,0 +1,122 @@
...
+		Flags: clientCredentialsFlags,
+		Action: func(c *cli.Context) error {
+			tokenSource := newClientCredentialsFromFlags(c).
+				TokenSource(c.Context)
+			token, err := tokenSource.Token()
+			if err != nil {
+				return errors.Wrap(err, "generate token")
+			}
+
+			data, err := json.MarshalIndent(token, "", "  ")
```
chrsmith, line 114: Not to over-think it, but it would be nice to separate concerns from "CLI tooling stuff" from "SAMS client interactions"?

e.g. within this file, maybe have a:

```go
// samsClientAdmin is an API client for interacting with
// the SAMS APIs on behalf of a SAMS client, e.g. an OAuth
// app that _uses_ SAMS.
type samsClientAdmin struct {
   TokenSource ouath.TokenSource
}

// GetClientToken returns the current OAuth access token generated from
// the underlying token source.
func (sac *samsAdminClient) GetClientToken() string { ... }

// And later...
func (sac *samsAdminClient) ListScopes() ([]sdk.Scope, error) { ... }
func (sac *samsAdminClient) AddScopeToApp(newScope sdk.Scope) error { ... }
```

chrsmith (approved):
> We previously talked about making a SAMS-specific CLI, but IMO that's a pretty big pain point if we want SAMS integration adoption when everything else lives in sg, and all the nice tooling lives here as well.

Directionally, this sounds like the right call. (Though I still liked the idea of a ~Sam Adams 🍺~ `sans-admin` tool 😢.)

In case it helps, these are the things that I've needed to do as an admin of a SAMS integration:

CRUD on the scopes that a SAMS client can send in a request. (e.g. authorize "my app" to send tokens with "your app"'s permissions.) It seems like we'd need to be careful about how we expose that type of functionality though. (e.g. requiring that only an admin of "your app" can update "my app" to issue scopes for "your app", etc.)

bobheadxi:
> In case it helps, these are the things that I've needed to do as an admin of a SAMS integration:

It does help; I'd love to expand on the tooling here for all your SAMS-integration-development needs 😁 once @unknwon gets the chance to review, I'll open some tickets tracking some follow-up ideas

dev/sg/sams/sg_sams.go
bobheadxi: Hmm, I'm inclined to keep this as is for now - I don't think this makes the callsite here much simpler (e.g. I don't think this client type should handle JSON marshalling, so the callsite still needs to handle the get-token error and then the JSON marshalling), and in between the other client types (`oauth2.TokenSource` is arguably a client type, as the `.Token()` gives us a structured type and whatnot with the credentials, as well as the existing `sams.ClientV1`) we already have a lot of logic that is separated from the CLI tooling stuff.

bobheadxi: In other words, what we have here is basically already all CLI concerns (constructing type from flag, doing the thing, checking the error, dumping the output appropriately)

dev/sg/sams/sg_sams.go
```diff
@@ -0,0 +1,122 @@
+// Package msp exports 'sg sams' commands for the Sourcegraph Accounts Management System.
```
unknwon, line 1:
```suggestion
// Package sams exports 'sg sams' commands for the Sourcegraph Accounts Management System.
```

unknwon:
> > In case it helps, these are the things that I've needed to do as an admin of a SAMS integration:
> 
> It does help; I'd love to expand on the tooling here for all your SAMS-integration-development needs 😁 once @unknwon gets the chance to review, I'll open some tickets tracking some follow-up ideas

The direction sounds good, tho I'd want SAMS to first stop using shared-static management secret, e.g. somehow supporting/take advantage of (validating) credentials produced by `gcloud auth application-default`.

dev/sg/sams/sg_sams.go
```diff
@@ -0,0 +1,122 @@
...
+	&cli.StringFlag{
+		Name:     "client-secret",
+		EnvVars:  []string{"SG_SAMS_CLIENT_SECRET"},
+		Usage:    "Client secret for the Sourcegraph Accounts Management System (SAMS) client",
+		Required: true,
+	},
+	&cli.StringSliceFlag{
+		Name:    "scopes",
+		Aliases: []string{"s"},
+		EnvVars: []string{"SAMS_SCOPES"},
```
unknwon, line 40:
```suggestion
		EnvVars: []string{"SG_SAMS_SCOPES"},
```

Unless there is a reason not to be consistent? (e.g. a package is expecting `SAMS_SCOPES `?)

unknwon (approved):
🎉

dev/sg/sams/sg_sams.go
bobheadxi: I think it's better to have this be listed explicitly actually (i.e. not allowed to be set via env), as copy-pasting it into a PR test plan with `-s` is clearer, and this isn't sensitive

bobheadxi:
```suggestion
```

PR #62888, opened 2024-05-23
SSC: Handle "Accept invite" calls

vdavid:
Before this PR, we had no "Accept invite" functionality.
A few days ago, @chrsmith added the email-sending feature in https://github.com/sourcegraph/self-serve-cody/pull/777, and I tweaked it a bit in https://github.com/sourcegraph/self-serve-cody/pull/781. The current PR wires up the rest of the flow, and lets the user actually use the invite.

**Three changes happen in this PR:**

1. I extracted `CodyAlert` and `CodyContainer` because I wanted to reuse them on pages other than `CodyManageTeamPage`.
2. I split the remaining styles of `CodyManageTeamPage` into `InviteUsers` and `TeamMemberList`.
3. I added the new page at `/cody/invites/accept`. It makes a back-end call to `POST /team/${teamId}/invites/${inviteId}/accept`, and if it succeeds, redirects the user to `/cody/team/manage`, otherwise displays an error message.

For easier review, no. 1 and no. 2 are in one commit, and no. 3 is in a different commit.

There are a few problematic parts, but first, to have a common understanding of the process, I'll describe the user flow and how the user interacts with our system when using invites.

**This is a common happy path:**

1. An admin goes to `https://cody-services-dev.sgdev.dev/cody/subscription`, clicks the  "Create a Cody Pro team" button, and buys two seats.
2. Same admin goes to `https://cody-services-dev.sgdev.dev/cody/team/manage` and invites a colleague. This action sends an email to the colleague
3. The colleague receives the "You've been invited to join a Cody Pro team!" email and clicks the big "View invite" button.
4. The colleague arrives at the `https://cody-services-dev.sgdev.dev/cody/invites/accept` page, through a link that looks like this: `https://cody-services-dev.sgdev.dev/cody/invites/accept?inviteID=018f9b13-ddd1-7f1f-b030-29b6f8fc77dd&signin=OpenIDConnect&teamID=018f9b13-5bf7-77cc-b192-c2fcac17ad94` (has an `inviteID`, and a `teamID` argument).
5. The page checks the validity of the `inviteId`+`teamId` combo.
6. It's valid, so the user gets redirected to `https://cody-services-dev.sgdev.dev/cody/team/manage` and sees a list of the admin and themself.
7. 💰 

**Problems I'm seeing:**

1. (larger problem) With this architecture, anyone can accept the invite. This is problematic in at least two ways:
    - If the user forwards the email to someone (maybe even through an autoforward), that someone can redeem the invite for them.
    - If the user is logged in with a different user on sourcegraph.com, their invite will be redeemed for the wrong user. This can be as easy as just clicking the wrong SSO icon when presented. To solve this, the admin can kick out and re-invite the colleague, but it's still awkward.
    - **Suggestion 1: back end:** In step no. 5, compare the logged-in SAMS user's email address to the email address of the invite ([here](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/self-serve-cody@main/-/blob/cody/backend/internal/database/team_invites.go?L265)) and fail if they don't match.
    - **Suggestion 2: front end:** If suggestion no. 1 doesn't work for some reason, we can add the invited email address to the link, like `https://cody-services-dev.sgdev.dev/cody/invites/accept?inviteID=018f9b13-ddd1-7f1f-b030-29b6f8fc77dd&signin=OpenIDConnect&teamID=018f9b13-5bf7-77cc-b192-c2fcac17ad94&email=invited.person@company.com` (notice the last part), and then check it against the logged in user's email address on the front end. Fail if they don't match.
2. (smaller problem) The "**View** invite" button actually links `https://cody-services-dev.sgdev.dev/cody/invites/accept`, which straight away **accepts** the invite. The button text is misleading.
CleanShot 2024-05-23 at 23 56 58@2x
    - **Suggestion:** Let's use the copy "Accept invite".
3. (smaller problem) The name is missing from the email:
    - **Suggestion:** Let's just say "Hey, friend," or something. We won't know the name of most users.
    - **Suggestion 2:** I'd treat this as a nice-to-have because we don't even have a DB field for it, but in case the invitee already has a Sourcegraph account, use their first name.

That said, the current PR could be fine as-is; we just need to be mindful of these two caveats and fix them (or consciously ignore them) as a follow-up.

## Test plan

I've tested it manually. The flow is not perfect (A "Bad gateway" response displays on the "Accept" page for a split-second AND a "Forbidden" response is returned on the "Manage" page), but it accepts the invite correctly and the user shows up in the list:

CleanShot 2024-05-23 at 23 51 29@2x

Same on the admin's side:

CleanShot 2024-05-23 at 23 52 21@2x

I'll fix these problems in a follow-up PR.


chenkc805:
Thanks for writing this up so detailed and clearly @vdavid . 

- What's the pros/cons between the front-end vs back-end suggestions? They end up giving the same functionality but I'm not sure if there are trade-offs. I guess the front-end solution is susceptible to tampering, but is that the only downside?
- For "Suggestion: Let's use the copy "Accept invite".", let's make this change. Good call
- "Suggestion: Let's just say "Hey, friend," or something. We won't know the name of most users." yeah let's do this. Getting the email or name is a nice to have.

client/web/src/cody/invites/AcceptInvitePage.tsx
```diff
@@ -0,0 +1,87 @@
...
+    const teamId = parameters.get('teamID')
+    const inviteId = parameters.get('inviteID')
+    const [loading, setLoading] = React.useState(true)
+    const [errorMessage, setErrorMessage] = React.useState<string | null>(null)
+
+    useEffect(() => {
+        async function postAcceptInvite(): Promise<void> {
+            const response = await requestSSC(`/team/${teamId}/invites/${inviteId}/accept`, 'POST')
+            setLoading(false)
+            if (response.status.toString().startsWith('2')) {
```
taras-yemets, line 41:
```suggestion
            if (response.ok) {
```

client/web/src/cody/invites/AcceptInvitePage.tsx
```diff
@@ -0,0 +1,87 @@
...
+                await new Promise(resolve => setTimeout(resolve, 1000))
```
taras-yemets, line 42: Why do we need this timeout?

client/web/src/cody/invites/AcceptInvitePage.tsx
```diff
@@ -0,0 +1,87 @@
...
+    // Process query params
+    const parameters = useSearchParameters()
+    const teamId = parameters.get('teamID')
+    const inviteId = parameters.get('inviteID')
+    const [loading, setLoading] = React.useState(true)
+    const [errorMessage, setErrorMessage] = React.useState<string | null>(null)
+
+    useEffect(() => {
+        async function postAcceptInvite(): Promise<void> {
+            const response = await requestSSC(`/team/${teamId}/invites/${inviteId}/accept`, 'POST')
```
taras-yemets, line 39: Are we sure that `teamID` or `inviteID` is not `null`?

client/web/src/cody/invites/AcceptInvitePage.tsx
```diff
@@ -0,0 +1,87 @@
...
+            } else {
+                setErrorMessage(await response.text())
+            }
+        }
+
+        void postAcceptInvite()
+    }, [inviteId, navigate, teamId])
+
+    return (
+        <>
```
taras-yemets, line 53: nit: we probably don't need to wrap with `React.Fragment` here

client/web/src/cody/team/InviteUsers.module.scss
```diff
@@ -0,0 +1,3 @@
+.invite-users-header {
+    font-weight: normal;
+}
```
taras-yemets, line 3: nit: isn't it a default? I don't see `font-weight` being changed anywhere in `H2` parents.

client/web/src/cody/team/TeamMemberList.module.scss
```diff
@@ -0,0 +1,17 @@
...
+
+.avatar-placeholder {
+    border: 1px solid var(--border-color);
+}
+
+.team-member-list {
+    display: grid;
+    grid-template-columns: 1fr fit-content(10rem) 1fr fit-content(10rem) fit-content(10rem);
+    gap: 1rem;
+}
```
taras-yemets, line 17: nit: we can use the [widcard `Grid`](https://github.com/sourcegraph/sourcegraph/blob/3339660a77abaafe999d3692cdd611eede71d661/client/wildcard/src/components/Grid/Grid.tsx#L20) component to achieve similar effect too.

client/web/src/routes.constants.ts
```diff
@@ -48,6 +48,8 @@ export enum PageRoutes {
     CodySubscription = '/cody/subscription',
 
     CodyManageTeam = '/cody/team/manage',
+    // Accepts an invite to join a Cody team, then redirects to the Cody team page.
+    CodyAcceptInvite = '/cody/invites/accept',
```
taras-yemets, line 52:
```suggestion
    CodyAcceptInvite = '/cody/invite/accept',
```
though, not sure about it

client/web/src/routes.tsx
```diff
@@ -5,6 +5,7 @@ import { Navigate, useNavigate, type RouteObject } from 'react-router-dom'
 import { useExperimentalFeatures } from '@sourcegraph/shared/src/settings/settings'
 import { lazyComponent } from '@sourcegraph/shared/src/util/lazyComponent'
 
+import { CodyAcceptInvitePage } from './cody/invites/AcceptInvitePage'
```
taras-yemets, line 8: This import should probably be wrapped with `lazyComponent` (similar to others).

taras-yemets (approved):
Overall the code LGTM. Left a few comments inline. Approving to unblock.

client/web/src/cody/invites/AcceptInvitePage.tsx
vdavid: Oh, that's definitely more elegant. TIL it's the same. Fixed in https://github.com/sourcegraph/sourcegraph/pull/62888/commits/79f173077d7854e1cd7bb971507a2f33ae8e93eb

client/web/src/cody/invites/AcceptInvitePage.tsx
vdavid: I thought it'd be nice for UX to wait a second before navigating to the manage team page so that the user sees the success alert. I know it's a sacrifice on speed, but 1 sec seemed justified for this. Added this in a comment [here](https://github.com/sourcegraph/sourcegraph/pull/62888/commits/b8ef25c527fd12542fcbfe03a3d9aaa93470e3b4). @rrhyne for 👀 , let me know if this is silly and I'll just remove it.

client/web/src/cody/invites/AcceptInvitePage.tsx
vdavid: Well, I mean, if they are, it'll result in a broken URL, but then:
a.) either we already had some major error on the back end that resulted in this, in which case we already have that error logged somewhere and we'll notice and fix it with priority
b.) or the user tampered with the URL, in which case, well, it's expected that we fail.

client/web/src/cody/invites/AcceptInvitePage.tsx
vdavid: Oh, thanks, that was silly. Fixed in https://github.com/sourcegraph/sourcegraph/pull/62888/commits/a805af7d00e240c06bdc1c5029cbddca440988ff

client/web/src/routes.constants.ts
vdavid: It's /cody/invite**s**/accept. I probably would've named it `invite`, but @chrsmith named it this way and I found it difficult to argue for `invite` so just left it as-is. It's definitely a bit confusing that the other path is `/cody/team/manage` with singular "team", which was my naming. We can easily harmonize these later if we want.

client/web/src/routes.tsx
vdavid: Ah, nice catch, thank you! https://github.com/sourcegraph/sourcegraph/pull/62888/commits/4f924a5ca50ce9f7f5b8597f97bfdc6220f58962 fixes this. I relied on my IDE to do the importing, so I never noticed. I should've done a better self-review, noted.

client/web/src/cody/team/TeamMemberList.module.scss
vdavid: I think I'll just let this one go for now. I agree that it'd be nice to centralize CSS grid use, but I see no current value-add from that, and we have 100+ occasions in the monorepo where we don't do that. If there's a clear gain, we can later do a run-through the whole codebase and convert the custom uses to `Grid`. For now, I'll just add +1 to the custom ones.

client/web/src/cody/invites/AcceptInvitePage.tsx
taras-yemets: I mean that if any of them is `null,` we don't need to call the API as there's no chance the backend will handle it.

vdavid:
@chenkc805, I've added the next actions to new ticket https://github.com/sourcegraph/self-serve-cody/issues/785. For the back-end/front-end question, the back-end solution is a clear winner so marked that one as the next action.

PR #62890, opened 2024-05-23
Pings: Update hardcoded latest release strings

DaedalusG:
This is a quick fix for `SiteUpdateReadinessCheck` which lets Sourcegraph instances know when a new version of Sourcegraph is released via Pings. Accompanied by change in `deploy-docker` repo to remove `qa-` prefix from those repo tags 

## Test plan
Slight change not tested yet


Chickensoupwithrice:
Won't this be immediately out of date once merged? The new version will be cut with a new version number and then it'll be wrong again.

DaedalusG:
> Won't this be immediately out of date once merged? The new version will be cut with a new version number and then it'll be wrong again.

Yea we'll need to fix it again, but it'll rollout the banners correctly for now, and also verify that the that the tag fixes are working

PR #62891, opened 2024-05-24
repo-updater: Hydrate schedule on startup

eseliger:
Currently, when repo-updater restarts it loses all intel it collected over time. That causes a large flood of git fetch requests after it restarts. Every repo will be enqueued for an immediate update.

This PR fixes that by populating the scheduler with an initial delay per repo that is calculated with the same heuristic that the scheduler uses when it's fully warmed up.

This should avoid fetching git repos that are very stale (most likely the majority on instances with many repos).

Test plan:

Ran it locally, verified the scheduler state using the instrumentation tool for it, the schedule looks as expected and most repos aren't scheduled for the next 8h.

### DIFF
--- a/cmd/repo-updater/internal/scheduler/scheduler.go
+++ b/cmd/repo-updater/internal/scheduler/scheduler.go
@@ -18,6 +18,7 @@ import (
 	"github.com/sourcegraph/sourcegraph/internal/api"
 	"github.com/sourcegraph/sourcegraph/internal/conf"
 	"github.com/sourcegraph/sourcegraph/internal/database"
+	"github.com/sourcegraph/sourcegraph/internal/dotcom"
 	"github.com/sourcegraph/sourcegraph/internal/limiter"
 	"github.com/sourcegraph/sourcegraph/internal/ratelimit"
 	"github.com/sourcegraph/sourcegraph/internal/repoupdater/protocol"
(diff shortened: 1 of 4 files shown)


eseliger:
* **#62891** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62891?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

cmd/repo-updater/shared/main.go
```diff
@@ -362,6 +362,10 @@ func newUnclonedReposManager(ctx context.Context, logger log.Logger, isSourcegra
 			// of the queue.
 			managed := sched.ListRepoIDs()
 
+			if len(managed) == 0 {
+				return nil
+			}
```
eseliger, line 367: fixes an edge-case where the schedule is empty and all repos would be loaded into the scheduler accidentally.

cmd/repo-updater/internal/scheduler/scheduler.go
```diff
@@ -99,6 +100,65 @@ func (s *UpdateScheduler) Start() {
 	ctx, cancel := context.WithCancel(actor.WithInternalActor(context.Background()))
 	s.cancelCtx = cancel
 
+	if !dotcom.SourcegraphDotComMode() {
```
eseliger, line 103: This cannot work as easily for dotcom, it's too complex right now with the indexed repos and friends are indexed and fetched regularly only stuff.

I didn't want to spend too many brain cycles on this, but we can get back to it.

mmanela (approved):

PR #62893, opened 2024-05-24
bazel: Add container test for s3proxy

willdollman:
This test checks that the s3proxy service starts as expected, and that we're able to create an example bucket using a curl request. Previously we were running this manually whenever s3proxy was updated.

Also updated the bazel docs to mention a caveat of the setup steps.

## Test plan

- CI
- Running new test locally


shivasurya (approved):
Thank you 🏅

PR #62895, opened 2024-05-24
svelte: Enable 'extended matching' in fuzzy finder

fkling:
This commit enables extended matching in fuzzy finder which allows the use of multiple search terms and using special symbols such as ^ and $ to anchor terms.

I also refactored the code quite a bit to reduce code duplication.

Closes SRCH-133 SRCH-134

## Test plan

Manual testing.


fkling:
I realized that we would have to pre-process the input value to remove fzf specific characters such as `^`, `'` and `$`, so this is not ready to merge yet but can still be reviewed.

fkling:
We'll try a different approach that works more closely with our existing search query.

PR #62898, opened 2024-05-24
sg/msp: fix init prompts breaking when encountering whitespace

jac:
Closes CORE-133

Fixes reading input for `sg msp init` prompts if the input has whitespace

## Test plan

locally tested
image


unknwon (approved):
Assuming it works as advertised :D

PR #62903, opened 2024-05-24
appliance: move storage elements to standard feature

craigfurman:
Please review one commit at a time, in order to give another pair of eyes to this being a pure refactor (the only golden file changes are config hashes and the actual spec structure, with the exception of redis, in which I changed the quantities to make sure we weren't accidentally feeding cache config to store or vice-versa).

---

**appliance: move storage elements to standard feature**

Add a new standard block, persistentVolumeConfig, containing a storage
size and storage class name. pvc.NewPersistentVolumeClaim() can fetch
these values when a standard config is (optionally) passed to it,
similar to other standard features.

Remove service-specific code that parses storage quantities out of
config.

Regarding storage class, this makes the reconciler "dumber" and more
verbosely-configured: every PVC must be explicitly assigned a storage
class (or none), rather than taking from the global StorageClassSpec.
Note that this doesn't have direct implications for admin UX: the
appliance web app can still have features to configure a single, global,
storage class name - it can simply copy this field everywhere needed
into the ConfigMap it crafts. This relates to
https://linear.app/sourcegraph/issue/REL-16/appliance-can-create-recommended-storageclasses-for-various-k8s
also - the StorageClassSpec is unused until that feature is implemented.

**appliance: consolidate storage tests**

Now that we have a standard feature, we can replace most instances of
with-storage tests with one standard fixture. Symbols is the exception,
since there is some logic done on the storage size (to determine a cache
size env var).

## Test plan

Golden tests included.


jdpleiness (approved):
I checked out the PR, ran the golden tests, and didn't see any divergence on my end either 🎉 

>Regarding storage class, this makes the reconciler "dumber" and more
verbosely-configured: every PVC must be explicitly assigned a storage
class (or none), rather than taking from the global StorageClassSpec.

While this makes the rec "dumber", I think this is actually a more desirable behavior in the end as it allows more granular control over storage classes if needed.

PR #62906, opened 2024-05-24
SSC: Handle `team=1` input on checkout form

vdavid:
We already have this on the old SSC front end since my earlier PR https://github.com/sourcegraph/self-serve-cody/pull/779. Now bringing the same feature to this page because we'll migrate our front end to this one, so we'll need it.

**Note:** This PR contains a bunch of random small fixes. Sorry for polluting the PR with so many little changes. I promise they are trivial, and I carefully checked the changes.

## Test plan

Tested that with the `?team=1` parameter, the page displays the seat number setting UI, and without it, it doesn't.
For the refactors, CI.

### DIFF
--- a/client/web/src/cody/management/api/client.ts
+++ b/client/web/src/cody/management/api/client.ts
@@ -45,7 +45,7 @@ export module Client {
 export interface Call<Resp> {
     method: 'GET' | 'POST' | 'PATCH' | 'DELETE'
     urlSuffix: string
-    requestBody?: any
+    requestBody?: unknown
 
     // Unused. This will never be set, it is only to
     // pass along the expected response type.
--- a/client/web/src/cody/management/api/client.ts
+++ b/client/web/src/cody/management/api/client.ts
@@ -64,7 +64,7 @@ export interface Caller {
 // the current Sourcegraph instance's SSC proxy API endpoint.
 export class CodyProApiCaller implements Caller {
     // e.g. "https://sourcegraph.com"
-    private origin: string
+    private readonly origin: string
 
     constructor() {
         this.origin = window.location.origin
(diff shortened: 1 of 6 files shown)


client/web/src/cody/management/api/client.ts
```diff
@@ -45,7 +45,7 @@ export module Client {
 export interface Call<Resp> {
     method: 'GET' | 'POST' | 'PATCH' | 'DELETE'
     urlSuffix: string
-    requestBody?: any
```
vdavid, line 48: ESLint complaint about the explicit `any` and suggested either `never` or `unknown`. I went with `unknown` because `never` seemed like a non-solution.

client/web/src/cody/management/api/hooks/useApiClient.tsx
```diff
@@ -34,7 +34,7 @@ export function useApiCaller<Resp>(call: Call<Resp>): ReactFriendlyApiResponse<R
         // https://maxrozen.com/race-conditions-fetching-data-react-with-useeffect
         let ignore = false
 
-        ;(async () => {
```
vdavid, line 37: I thought this was fine with the linter but it wasn't because it discarded the returned promise. I refactored it to this longer format because this one may be obscure to some devs, and we use it very rarely in the codebase.

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
```diff
@@ -18,16 +18,18 @@ export const CodyProCheckoutForm: React.FunctionComponent<{
...
     const showPromoCodeField = urlSearchParams.get('showCouponCodeAtCheckout') !== null
 
     // Make the API call to create the Stripe Checkout session.
     const call = useMemo(() => {
-        const req: CreateCheckoutSessionRequest = {
+        const requestBody: CreateCheckoutSessionRequest = {
             interval: 'monthly',
-            seats: 1,
+            // If creating a team, we set seatCount=0, which means the user can adjust the seat count.
+            seats: creatingTeam ? 0 : 1,
```
vdavid, line 32: This is the meat of the PR, really. It's a copy of the functionality on the old SSC site, and also is very simple.

taras-yemets (approved):

PR #62908, opened 2024-05-24
svelte: Add lucide icons and new icon API

fkling:
See also #62904.

This PR adds a new icon component (`Icon2`), which takes an svg component imported by `unplugin-icons`. To make things easier the icons can be auto-imported via `unplugin-auto-import`.

Usage example:

```js
import Icon2 from '$lib/Icon2.svelte'

<Icon2 icon={ILucideX} inline />
```

`unplugin-icons` allows us to:

- use other icon sets besides lucide, while using the same API. Other icon sets are necessary for file icons. `react-icons`, which we use in the React app, also pulls from multiple icon sets.
- define our own icon sets

See https://github.com/unplugin/unplugin-icons for more information.

If it turns out that auto-importing (which involves generating TypeScrip types) doesn't work with our Bazel pipeline we can fall back to manual importing.

## Test plan

Manual testing.


camdencheek (approved):
LGTM! Stamping to unblock, but did not run it locally

client/web-sveltekit/src/lib/Icon2.svelte
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    export let icon: ComponentType<SvelteComponent<SvelteHTMLElements['svg']>>
+    export let inline: boolean = false
+</script>
+
+<span><svelte:component this={icon} class="icon {inline ? 'icon-inline' : ''}" /></span>
+
+<style lang="scss">
+    span {
```
vovakulikov, line 17: is there a way to ensure CSS incapsulation without this span root element?

client/web-sveltekit/src/lib/Icon2.svelte
```diff
@@ -0,0 +1,35 @@
...
+    type $$Props = SvelteHTMLElements['svg'] & {
+        icon: ComponentType<SvelteComponent<SvelteHTMLElements['svg']>>
+        inline?: boolean
+    }
+
+    export let icon: ComponentType<SvelteComponent<SvelteHTMLElements['svg']>>
+    export let inline: boolean = false
+</script>
+
+<span><svelte:component this={icon} class="icon {inline ? 'icon-inline' : ''}" /></span>
```
vovakulikov, line 14: Do we need `class="icon"`and default `class: "icon"` in Icons `'unplugin-icons/vite'` configuration?

vovakulikov (approved):

client/web-sveltekit/src/lib/Icon2.svelte
fkling: We could generate a random string on our own (e.g. `md5("icon")`) and use that as additional class.

client/web-sveltekit/src/lib/Icon2.svelte
fkling: No. The plugin configuration was a leftover from #62904.

client/web-sveltekit/src/lib/Icon2.svelte
fkling: Or CSS modules of course.

PR #62909, opened 2024-05-24
feat/cody-gateway: support wildcard models

bobheadxi:
In https://sourcegraph.slack.com/archives/C05SZB829D0/p1715638980052279 we shared a decision we landed on as part of #62263:

> Ignoring (then removing) per-subscription model allowlists: As part of the API discussions, we've also surfaced some opportunities for improvements - to make it easier to roll out new models to Enterprise, we're not including per-subscription model allowlists in the new API, and as part of the Cody Gateway migration (by end-of-June), we will update Cody Gateway to stop enforcing per-subscription model allowlists. Cody Gateway will still retain a Cody-Gateway-wide model allowlist. [@chrsmith](https://sourcegraph.slack.com/team/U061QHKUBJ8) is working on a broader design here and will have more to share on this later.

To support this, we first need to extend Cody Gateway's model allowlist enforcement to respect a notion of "allow all models that are allowed in Cody Gateway". To ensure models are explicitly provided today, an empty `AllowedModels` is considered invalid, so we add a special single-element-slice-`*` configuration that can be used to indicate an actor's rate limit allows all models (`prefixedMasterAllowlist`).

This change also unifies somewhat the way we enforce allowed models in various places by introducing `(*RateLimit).EvaluateAllowedModels(...)` as the unified way to construct the final allowlist for a given rate limit.

I'm planning to roll this out before rolling out actual functionality changes (https://github.com/sourcegraph/sourcegraph/pull/62911) to ensure changes in cached rate limits don't end up confusing an older revision of Cody Gateway that doesn't yet support wildcard models. With #62911, rolling out new models to Enterprise customers no longer require additional code/override changes.

Part of https://linear.app/sourcegraph/issue/CORE-135

## Test plan

Unit tests, and E2E test of this in https://github.com/sourcegraph/sourcegraph/pull/62911

### DIFF
--- a/cmd/cody-gateway/internal/httpapi/completions/upstream.go
+++ b/cmd/cody-gateway/internal/httpapi/completions/upstream.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"slices"
 	"strconv"
 	"strings"
 	"time"
@@ -137,7 +136,7 @@ func makeUpstreamHandler[ReqT UpstreamRequest](
 	// upstreamName is the name of the upstream provider. It MUST match the
 	// provider names defined clientside, i.e. "anthropic" or "openai".
 	upstreamName string,
-
+	// unprefixed upstream model names
 	allowedModels []string,
 
 	methods upstreamHandlerMethods[ReqT],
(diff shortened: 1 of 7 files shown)


cmd/cody-gateway/internal/httpapi/completions/upstream.go
```diff
@@ -193,6 +192,9 @@ func makeUpstreamHandler[ReqT UpstreamRequest](
 		// This isn't very robust, but should tide us through a brief transition
 		// period until everything deploys and our caches refresh.
 		for i := range rateLimit.AllowedModels {
+			if rateLimit.AllowedModels[i] == "*" {
```
rafax, line 195: Seeing that this is `upstream.go`, which already does many things and has a lot of dependencies, I think it would be good to hire the `"*"` as an implementation detail in limits.go (say, add limits.IsWildCard() method

rafax: Or, move the whole "is this prefixed or not" logic to a single place, handle wildcard and prefixing there?

rafax (approved):

cmd/cody-gateway/internal/httpapi/completions/upstream.go
bobheadxi: The docstring says this is `TEMPORARY`, so I think this was only meant to tide us over until the cache state migrates... ideally, this whole block of code should be removed. But, it doesn't feel appropriate to do in this PR 😅 

In light of this being a one-off thing, would you be okay with keeping this here? I can add this as a docstring as well. My intent is to discourage this type of checking because the actor model list should always be prefixed, and we should never manipulate the actor models list ourselves, instead depending only on `EvaluateAllowedModels` to do the check

bobheadxi:
Merging this first to start working my way through landing this large stack of PRs, as I shared in Slack: https://sourcegraph.slack.com/archives/C05SZB829D0/p1717173273215949?thread_ts=1715638980.052279&cid=C05SZB829D0

This is low-risk as it doesn't include any real functionality changes yet, just prepares them, as outlined in PR description.

Happy to address any post-merge reviews :)

PR #62919, opened 2024-05-24
feat(search): Add Syntax Highlighting for Magik language

mmanela:
Fixes GRAPH-621

Key notes
1. Added a reference to a fork of the Magik tree-sitter package since the upstream didn't support rust and was using a newer version of tree-sitter which we are not yet supporting.
2. Add a highlights query for identifying the items to highlight  
3. Added test cases based on language description on wikipedia, example in tree sitter package and other examples online. 
4. I do not have a way to run Magik programs and I could not find a full language specification, so I am not confident this covers everything.
5. The tree-sitter grammar is missing support for a couple things in their own examples. Upstream issues filed and added in comments

# Links
Following links I used to find about the language
- https://en.wikipedia.org/wiki/Magik_(programming_language)
- http://sw-gis.wikidot.com/magik-lang

# Screenshots
 
image

## Test plan
 - Unit tests
 - Validate UI and ensure it looks appropriate

## Changelog
- Added syntax highlighting for the Magik programming language.


mmanela:
@sebastiaanspeck Thanks for the quick work there. I pulled the changes in and they work great!

docker-images/syntax-highlighter/crates/syntax-analysis/queries/magik/highlights.scm
```diff
@@ -0,0 +1,195 @@
...
+(logical_operator
+    operator: _ @operator)
+
+(arithmetic_operator
+    operator: _ @operator)
+
+(unary_operator
+    operator: _ @operator)
+
+(class _ @keyword) @type
```
kritzcreek, line 66: IIUC this is trying to target `_class` as keyword here. It would be less brittle to just add `"_class"` to the list of keywords. I don't know if you want to highlight the entire class definition as `@type` because it also targets the surrounding pipe symbols.

I have no clue why I can't target the `java_classname` field though, seems like a weird tree-sitter bug.

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__magik.magik.snap
```diff
@@ -0,0 +1,744 @@
...
+//                ^ IdentifierOperator
+//                 ^^^^ IdentifierFunction
+//                      ^^^^ Identifier
+//                            ^^^ Identifier
+  _endmethod
+//^^^^^^^^^^ Keyword
+  
+  # Initialise method.
+//^^^^^^^^^^^^^^^^^^^^ Comment
+  _private _method person.init(name, age)
```
kritzcreek, line 454: Why aren't we marking the dot in here as `IdentifierOperator`? We seem to do so in other places

Do we need to add a `(method "." @operator)` query?

docker-images/syntax-highlighter/crates/syntax-analysis/queries/magik/highlights.scm
```diff
@@ -0,0 +1,195 @@
...
+
+(invoke
+  receiver: (variable) @function.builtin
+  (#eq? @function.builtin "def_mixin"))
+
+(invoke
+  receiver: (_) @function)
+
+(call
+  receiver: (variable) @variable )
```
kritzcreek, line 80:
```suggestion
```
`variable`'s already get marked as `@variable` universally.

docker-images/syntax-highlighter/crates/syntax-analysis/queries/magik/highlights.scm
```diff
@@ -0,0 +1,195 @@
...
+(call
+  operator: "." @operator)
+(call
+  message: (identifier) @function
+  "(")
+(call
+  message: (identifier) @variable)
```
kritzcreek, line 87:
```suggestion
```

same as above

docker-images/syntax-highlighter/crates/syntax-analysis/queries/magik/highlights.scm
```diff
@@ -0,0 +1,195 @@
...
+
+; NOTE(issues: https://github.com/krn-robin/tree-sitter-magik/issues/49): Grammar does not support traversing to the children in the pragma
+(pragma) @identifier.attribute
+
+(argument) @identifier.parameter
+
+; Literals
+[
+  (number)
+] @number
```
kritzcreek, line 160:
```suggestion
(number) @number
```

docker-images/syntax-highlighter/crates/syntax-analysis/queries/magik/highlights.scm
```diff
@@ -0,0 +1,195 @@
...
+
+[
+  (string_literal)
+] @string
```
kritzcreek, line 164:
```suggestion
(string_literal) @string
```

docker-images/syntax-highlighter/crates/syntax-analysis/queries/magik/highlights.scm
```diff
@@ -0,0 +1,195 @@
...
+  "_endproc"
+  "_cf"
+  "_scatter"
+  "_import"
+  "_optional"
+] @keyword
+
+[
+ "_package"
+] @include
```
kritzcreek, line 148:
```suggestion
"_package" @include
```

kritzcreek (changes requested):
We're currently depending on a `rust-bindings` branch for our fork of the upstream grammar, could we instead have our `main` be ahead of the upstream so it's easier to see at a glance how they compare when browsing via GH?

docker-images/syntax-highlighter/crates/syntax-analysis/queries/magik/highlights.scm
mmanela: Yea I did this just to make that whole part as @type just because I couldnt target it any other way. But open to just leave as not-highlighted

kritzcreek:
Codel looks good from my side WDYT about the branch management?

> We're currently depending on a rust-bindings branch for our fork of the upstream grammar, could we instead have our main be ahead of the upstream so it's easier to see at a glance how they compare when browsing via GH?

Does this change need an entry in the changelog as well?

mmanela:
@kritzcreek  Thanks for the feedback. I updated change log and our forked repo as you suggested

kritzcreek (approved):

docker-images/syntax-highlighter/crates/syntax-analysis/queries/magik/highlights.scm
```diff
@@ -0,0 +1,182 @@
...
+] @constant
+
+[
+ (variable)
+ (dynamic_variable)
+ (global_variable)
+ (global_reference)
+ (identifier)
+ (slot_accessor)
+ (label)
```
varungandhi-src, line 181: This one is interesting. Labels are not really variables. However, we don't really have a SyntaxKind in SCIP to mark labels. GoLand does highlight labels differently, using bold in the theme that I have.



We should probably add a `ControlFlowLabel` syntax kind to SCIP and use that here. I've filed a follow-up issue: https://linear.app/sourcegraph/issue/GRAPH-661/add-syntaxkindcontrolflowlabel-to-scip

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__magik.magik.snap
```diff
@@ -0,0 +1,749 @@
...
+//                        ^^^^^ IdentifierFunction
+//                              ^ NumericLiteral
+//                                 ^ IdentifierOperator
+//                                  ^ NumericLiteral
+//                                     ^ IdentifierOperator
+//                                      ^ NumericLiteral
+//                                         ^^ NumericLiteral
+  _global !current_view! << _unset
+//^^^^^^^ Keyword
+//        ^^^^^^^^^^^^^^ Identifier
```
varungandhi-src, line 135: The current view must be quite amazing to have two exclamation marks `!current_view!` 😆

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__magik.magik.snap
```diff
@@ -0,0 +1,749 @@
...
+//            ^^^^^ Keyword
+//                  ^^ IdentifierOperator
+//                     ^^^^^^ IdentifierBuiltin
+//                            ^^^^^^ Keyword
+  _endblock
+//^^^^^^^^^ Keyword
+  
+  _block @ label1?
+//^^^^^^ Keyword
+//       ^^^^^^^^^ Identifier
```
varungandhi-src, line 339: You'll notice that the space between `@` and `l` is being marked as an identifier too. It seems like the more idiomatic syntax is to not have a space there. See the code example below which shows other places where `@label` can be used, that would be useful to add a test case for to make sure the loop label is being picked up correctly in other places.

https://sourcegraph.com/github.com/krn-robin/tree-sitter-magik@main/-/blob/test/corpus/loop.txt?L85-94

docker-images/syntax-highlighter/crates/syntax-analysis/queries/magik/highlights.scm
```diff
@@ -0,0 +1,182 @@
...
+  "_when"
+  "_while"
+  "_with"
+] @keyword
+
+"_package" @include
+
+(regex_literal) @string.special
+
+; NOTE(issues: https://github.com/krn-robin/tree-sitter-magik/issues/49): Grammar does not support traversing to the children in the pragma
```
varungandhi-src, line 143: I see there is an open PR here: https://github.com/krn-robin/tree-sitter-magik/pull/50

Should we file a follow-up issue on our side to update the queries once that PR is merged/how would you like to handle this?

varungandhi-src:
@sebastiaanspeck does Magik have a standard nil/null value? I don't see it in the grammar, but I think other Smalltalk languages typically have `nil` (Smalltalk-80, Squeak, Pharo).

docker-images/syntax-highlighter/crates/syntax-analysis/queries/magik/highlights.scm
```diff
@@ -0,0 +1,182 @@
...
+[
+ (self)
+ (super)
+ (clone)
+] @variable.builtin
+
+[
+ (symbol)
+ (character_literal)
+] @constant
```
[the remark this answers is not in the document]
mmanela: Then is `@identifier` the best best for `(symbol)`?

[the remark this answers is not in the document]
varungandhi-src: Sorry I should've mentioned that. We should mark them as `@string`, since in most languages, symbols are simply interned strings. Other editors do the same. For example: https://github.com/github-linguist/linguist/blob/master/script/fast-submodule-update#L19-L28

docker-images/syntax-highlighter/crates/syntax-analysis/queries/magik/highlights.scm
```diff
@@ -0,0 +1,182 @@
...
+(string_literal) @string
+
+[
+  (true)
+  (false)
+] @boolean
+
+[
+  (maybe)
+  (unset)
```
varungandhi-src, line 160: Let's mark `unset` as `@constant.null` based on: https://github.com/sourcegraph/sourcegraph/pull/62919#issuecomment-2154183110

docker-images/syntax-highlighter/crates/syntax-analysis/src/highlighting/snapshots/syntax_analysis__highlighting__tree_sitter__test__magik.magik.snap
varungandhi-src: I meant this as feedback for @mmanela to add more tests to make sure the highlighting is working properly in cases such as `_continue@outer_loop`.

docker-images/syntax-highlighter/crates/syntax-analysis/queries/magik/highlights.scm
mmanela: https://linear.app/sourcegraph/issue/GRAPH-662/handle-pragma-for-magik-language-when-grammar-is-updated

docker-images/syntax-highlighter/crates/syntax-analysis/queries/magik/highlights.scm
```diff
@@ -0,0 +1,182 @@
...
+(method
+  exemplarname: (identifier) @type)
+(method
+  name: (identifier) @function)
+
+(method "." @operator)
+
+(procedure (label) @function)
+
+(documentation) @comment
```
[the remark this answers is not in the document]
mmanela: This doesn't seem right to me @sebastiaanspeck . Why would this be an attribute?

mmanela:
Addressed new comments in this [PR](https://github.com/sourcegraph/sourcegraph/pull/63155)

PR #62927, opened 2024-05-27
Support Google Gemini through AI Studio

rafax:
This PR adds support for Google Gemini models as exposed by AI Studio API.
Slack [context](https://sourcegraph.slack.com/archives/C072XJ7TT5G/p1716393829253889).

Closes AI-50.

## Test plan

- tested locally
- e2e tests


cmd/cody-gateway/shared/config/config.go
```diff
@@ -309,6 +317,16 @@ func (c *Config) Load() {
...
 
 	c.Environment = c.Get("CODY_GATEWAY_ENVIRONMENT", "dev", "Environment name.")
+
+	c.Google.AccessToken = c.GetOptional("CODY_GATEWAY_GOOGLE_ACCESS_TOKEN", "The Google AI Studio access token to be used.")
+	c.Google.AllowedModels = splitMaybe(c.Get("CODY_GATEWAY_GOOGLE_ALLOWED_MODELS",
+		strings.Join([]string{
+			"gemini-1.5-pro-latest",
+			"gemini-1.5-flash-latest",
+		}, ","),
+		"OpenAI models that can to be used."),
```
janhartman, line 327:
```suggestion
		"Google models that can be used."),
```

janhartman (approved):

cmd/cody-gateway/internal/httpapi/completions/upstream.go
```diff
@@ -64,7 +64,7 @@ var hopHeaders = map[string]struct{}{
 // Methods do not need to be concurrency-safe, as they are only called sequentially.
 type upstreamHandlerMethods[ReqT UpstreamRequest] interface {
 	// getAPIURLByFeature returns the upstream API endpoint to call for the given feature.
-	getAPIURLByFeature(codygateway.Feature) string
+	getAPIURL(codygateway.Feature, ReqT) string
```
vdavid, line 67: I see that the implementations of this method always ignore the second parameter, except the new `GoogleHandler`, but that only uses `req.ShouldStream()` (a bool) and req.Model (a string).
It feels meh. But the best alternative I could come up with it to use the `feature` argument and concat the two values as a string with some separator, but that's even uglier. So 👍 , just wanted to note that I thought about it and ack that this is the best solution we found without major changes.

vdavid (approved):
Looks 👌

cmd/cody-gateway/internal/httpapi/completions/upstream.go
rafax: UpstreamRequest only has 3 properties (`GetModel()` `ShouldStream()` and `BuildPrompt()`, so I decided that Google using 2/3 is close enough to pass the whole request.

cmd/cody-gateway/shared/config/config.go
rafax: Applied manually

PR #62911, opened 2024-05-24
feat/cody-gateway: use wildcard for enterprise allowlists

bobheadxi:
This change makes Cody Gateway always apply a wildcard model allowlist, irrespective of what the configured model allowlist is for an Enterprise subscription is in dotcom (see #62909).

The next PR in the stack, https://github.com/sourcegraph/sourcegraph/pull/62912, makes the GraphQL queries return similar results, and removes model allowlists from the subscription management UI.

Closes https://linear.app/sourcegraph/issue/CORE-135

### Context

In https://sourcegraph.slack.com/archives/C05SZB829D0/p1715638980052279 we shared a decision we landed on as part of #62263:

> Ignoring (then removing) per-subscription model allowlists: As part of the API discussions, we've also surfaced some opportunities for improvements - to make it easier to roll out new models to Enterprise, we're not including per-subscription model allowlists in the new API, and as part of the Cody Gateway migration (by end-of-June), we will update Cody Gateway to stop enforcing per-subscription model allowlists. Cody Gateway will still retain a Cody-Gateway-wide model allowlist. [@chrsmith](https://sourcegraph.slack.com/team/U061QHKUBJ8) is working on a broader design here and will have more to share on this later.

This means there is one less thing for us to migrate as part of https://github.com/sourcegraph/sourcegraph/pull/62934, and avoids the need to add an API field that will be removed shortly post-migration.

As part of this, rolling out new models to Enterprise customers no longer require additional code/override changes.

## Test plan

Set up Cody Gateway locally as documented, then `sg start dotcom`. Set up an enterprise subscription + license with a high seat count (for a high quota), and force a Cody Gateway sync:

```
curl -v -H 'Authorization: bearer sekret' http://localhost:9992/-/actor/sync-all-sources
```

Verify we are using wildcard allowlist:

```sh
$ redis-cli -p 6379 get 'v2:product-subscriptions:v2:slk_...'
[...]

PR #62934, opened 2024-05-27
feat/cody-gateway: use Enterprise Portal for actor/productsubscriptions

bobheadxi:
Migrates Cody Gateway to use the new Enterprise Portal's "read-only" APIs. For the most part, this is an in-place replacement - a lot of the diff is in testing and minor changes. Some changes, such as the removal of model allowlists, were made down the PR stack in https://github.com/sourcegraph/sourcegraph/pull/62911.

At a high level, we replace the data requested by `cmd/cody-gateway/internal/dotcom/operations.graphql` and replace it with Enterprise Portal RPCs:

- `codyaccessv1.GetCodyGatewayAccess`
- `codyaccessv1.ListCodyGatewayAccesses`

Use cases that previously required retrieving the active license tags now:

1. Use the display name provided by the Cody Access API https://github.com/sourcegraph/sourcegraph/pull/62968
2. Depend on the connected Enterprise Portal dev instance to only return dev subscriptions https://github.com/sourcegraph/sourcegraph/pull/62966

Closes https://linear.app/sourcegraph/issue/CORE-98
Related to https://linear.app/sourcegraph/issue/CORE-135 (https://github.com/sourcegraph/sourcegraph/pull/62909, https://github.com/sourcegraph/sourcegraph/pull/62911)
Related to https://linear.app/sourcegraph/issue/CORE-97

## Local development

This change also adds Enterprise Portal to `sg start dotcom`. For local development, we set up Cody Gateway to connect to Enterprise Portal such that zero configuration is needed - all the required secrets are sourced from the `sourcegrah-local-dev` GCP project automatically when you run `sg start dotcom`, and local Cody Gateway will talk to local Enterprise Portal to do the Enterprise subscriptions sync.

This is actually an upgrade from the current experience where you need to provide Cody Gateway a Sourcegraph user access token to test Enterprise locally, though the Sourcegraph user access token is still required for the PLG actor source.

The credential is configured in https://console.cloud.google.com/security/secret-manager/secret/SG_LOCAL_DEV_SAMS_CLIENT_SECRET/overview?project=sourcegraph-local-dev, and I've included documentation in the secret annotation about what it is for and what to do with it:

image

## Rollout plan

I will open PRs to set up the necessary configuration for Cody Gateway dev and prod. Once reviews taper down I'll cut an image from this branch and deploy it to Cody Gateway dev, and monitor it closely + do some manual testing. Once verified, I'll land this change and monitor a rollout to production.

Cody Gateway dev SAMS client: https://github.com/sourcegraph/infrastructure/pull/6108
Cody Gateway prod SAMS client update (this one already exists):

```
accounts=> UPDATE idp_clients
SET scopes = scopes || '["enterprise_portal::subscription::read", "enterprise_portal::codyaccess::read"]'::jsonb
WHERE id = 'sams_cid_018ea062-479e-7342-9473-66645e616cbf';
UPDATE 1
accounts=> select name, scopes from idp_clients WHERE name = 'Cody Gateway (prod)';
        name         |                                                              scopes                                                              
---------------------+----------------------------------------------------------------------------------------------------------------------------------
 Cody Gateway (prod) | ["openid", "profile", "email", "offline_access", "enterprise_portal::subscription::read", "enterprise_portal::codyaccess::read"]
(1 row)
```

Configuring the target Enterprise Portal instances: https://github.com/sourcegraph/infrastructure/pull/6127

## Test plan

Start the new `dotcom` runset, now including Enterprise Portal, and observe logs from both `enterprise-portal` and `cody-gateway`:

```
sg start dotcom
```

I reused the test plan from https://github.com/sourcegraph/sourcegraph/pull/62911: set up Cody Gateway external dependency secrets, then set up an enterprise subscription + license with a high seat count (for a high quota), and force a Cody Gateway sync:

```
curl -v -H 'Authorization: bearer sekret' http://localhost:9992/-/actor/sync-all-sources
```

This should indicate the new sync against "local dotcom" fetches the correct number of actors and whatnot.

Using the local enterprise subscription's access token, we run the QA test suite:

```sh
$ bazel test --runs_per_test=2 --test_output=all //cmd/cody-gateway/qa:qa_test --test_env=E2E_GATEWAY_ENDPOINT=http://localhost:9992 --test_env=E2E_GATEWAY_TOKEN=$TOKEN
INFO: Analyzed target //cmd/cody-gateway/qa:qa_test (0 packages loaded, 0 targets configured).
INFO: From Testing //cmd/cody-gateway/qa:qa_test (run 1 of 2):
==================== Test output for //cmd/cody-gateway/qa:qa_test (run 1 of 2):
PASS
================================================================================
INFO: From Testing //cmd/cody-gateway/qa:qa_test (run 2 of 2):
==================== Test output for //cmd/cody-gateway/qa:qa_test (run 2 of 2):
PASS
================================================================================
INFO: Found 1 test target...
Target //cmd/cody-gateway/qa:qa_test up-to-date:
  bazel-bin/cmd/cody-gateway/qa/qa_test_/qa_test
Aspect @@rules_rust//rust/private:clippy.bzl%rust_clippy_aspect of //cmd/cody-gateway/qa:qa_test up-to-date (nothing to build)
Aspect @@rules_rust//rust/private:rustfmt.bzl%rustfmt_aspect of //cmd/cody-gateway/qa:qa_test up-to-date (nothing to build)
INFO: Elapsed time: 13.653s, Critical Path: 13.38s
INFO: 7 processes: 1 internal, 6 darwin-sandbox.
INFO: Build completed successfully, 7 total actions
//cmd/cody-gateway/qa:qa_test                                            PASSED in 11.7s
  Stats over 2 runs: max = 11.7s, min = 11.7s, avg = 11.7s, dev = 0.0s

Executed 1 out of 1 test: 1 test passes.
```


bobheadxi:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62934?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#62934** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62934?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#62911** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62911?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#62912](https://github.com/sourcegraph/sourcegraph/pull/62912) <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62912?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#62909** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/62909?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

cmd/cody-gateway/internal/actor/productsubscription/productsubscription.go
```diff
@@ -252,71 +336,73 @@ func (s *Source) fetchAndCache(ctx context.Context, token string) (*actor.Actor,
...
+		if strings.HasPrefix(tag, "customer:") {
+			return strings.TrimPrefix(tag, "customer:")
 		}
 	}
-
-	// 2. Use the username of the account
-	if s.Account != nil && s.Account.Username != "" {
-		return s.Account.Username
-	}
 	return ""
```
bobheadxi, line 346: Discussed with @unknwon who added this, and agreed on the acceptable regression here - the customer name is inferred for Slack notifications and is only a QOL thing, so relying only on license tags and having a followup to switch this to the "get subscriptions" RPC when we implement that is acceptable: https://linear.app/sourcegraph/issue/CORE-136/enterprise-portal-implement-getlistenterprisesubscription

bobheadxi: No longer applies as I'm removing the licensing request in favour of https://github.com/sourcegraph/sourcegraph/pull/62968 and https://github.com/sourcegraph/sourcegraph/pull/62966

unknwon (approved):

cmd/cody-gateway/internal/actor/productsubscription/productsubscriptiontest/mocks.go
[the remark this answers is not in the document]
bobheadxi: Yeah, this specifically imported by another package (hence the other convention, `package foobartest`)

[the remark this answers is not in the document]
bobheadxi: In this PR it's used by `cmd/cody-gateway/internal/auth/auth_test.go`

cmd/cody-gateway/internal/actor/productsubscription/productsubscription_test.go
```diff
@@ -19,142 +20,59 @@ func TestNewActor(t *testing.T) {
...
-		s               dotcom.ProductSubscriptionState
-		devLicensesOnly bool
+		access *codyaccessv1.CodyGatewayAccess
 	}
 	tests := []struct {
-		name        string
-		args        args
+		name string
+		args args
+		// TODO add more coverage on created actor
```
chrsmith, line 28: Perhaps:

- `// TODO(#xxxx): Add more tests for scenario X.`
- `// TODO({gitHubAlias}): Add more tests for scenario X.`

Or just removing the comment entirely. e.g. if we are honest with ourselves, and you don't really think you'll get to adding those new tests... then maybe not add the TODO comment that will sit there for years?

cmd/cody-gateway/shared/main.go
```diff
@@ -348,3 +407,30 @@ func (p *dotcomPromptRecorder) Record(ctx context.Context, prompt string) error
 	key := fmt.Sprintf("prompt:%s:%s:%s", traceID, feature, reqActor.ID)
 	return p.redis.SetEx(key, p.ttlSeconds, prompt)
 }
+
+func dialEnterprisePortal(ctx context.Context, logger log.Logger, addr *url.URL, ts oauth2.TokenSource) (*grpc.ClientConn, error) {
```
chrsmith, line 411: Seems like this function and other functionality might be a bit too specific to include in `main.go`. What do you think about moving this into a separate `shared/enterprise_portal.go` and exposing a function like `func initializeEnterprisePortalActorSource(...)` or something?

i.e. is there a clean way we can isolate the "Enterprise Portal specific goo" to its own file, just to keep things better organized?

internal/grpc/grpcoauth/grpcoauth.go
```diff
@@ -0,0 +1,42 @@
...
+	"google.golang.org/grpc/credentials"
+
+	"github.com/sourcegraph/sourcegraph/internal/env"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+)
+
+// TokenSource supplies PerRPCCredentials from an oauth2.TokenSource.
+// It is a fork of the implementation in "google.golang.org/grpc/credentials/oauth",
+// but checks "internal/env" to toggle the "must require secure transport"
+// behaviour. Without these changes, the upstream token source cannot be used
```
chrsmith, line 16: What would we need to do in order to make this work? I am waaaay outside of my depth here, but it's kinda slick that running Sourcegraph locally uses the `sourcegraph.test` domain, and various infrastructure thingers ensure that the cert is registered locally.

So -- outside of this PR, and in some alternate reality where we have time to work on things like this -- is it just a matter of having the "local Enterprise Portal" be configured to use some well-known HTTPS cert?

internal/grpc/grpcoauth/grpcoauth.go
```diff
@@ -0,0 +1,42 @@
...
+// GetRequestMetadata gets the request metadata as a map from a TokenSource.
+func (ts TokenSource) GetRequestMetadata(ctx context.Context, uri ...string) (map[string]string, error) {
+	token, err := ts.Token()
+	if err != nil {
+		return nil, err
+	}
+	if !env.InsecureDev {
+		ri, _ := credentials.RequestInfoFromContext(ctx)
+		if err = credentials.CheckSecurityLevel(ri.AuthInfo, credentials.PrivacyAndIntegrity); err != nil {
+			return nil, errors.Newf("unable to transfer TokenSource PerRPCCredentials: %v", err)
```
chrsmith, line 31: I think the "correct" way to write this would be to use `%w` instead of `%v`. Since that way the caller can inspect the lineage of the error object.

See https://go.dev/play/p/scfNKZxJjDO.

internal/grpc/grpcoauth/grpcoauth.go
```diff
@@ -0,0 +1,42 @@
...
+		}
+	}
+	return map[string]string{
+		"authorization": token.Type() + " " + token.AccessToken,
+	}, nil
+}
+
+// RequireTransportSecurity indicates whether the credentials requires transport security.
+func (ts TokenSource) RequireTransportSecurity() bool {
```
chrsmith, line 40: Nit: For clarity, adding an "obvious" statement might be helpful. e.g.

```go
// RequireTransportSecurity ...
// For this implementation, we disable it if the INSECURE_DEV environment variable is set.
```

sg.config.yaml
```diff
@@ -1049,6 +1059,18 @@ bazelCommands:
       CODY_GATEWAY_FIREWORKS_ACCESS_TOKEN: sekret
       CODY_GATEWAY_SOURCEGRAPH_EMBEDDINGS_API_TOKEN: sekret
       CODY_GATEWAY_GOOGLE_ACCESS_TOKEN: sekret
+      # Connect to services that require SAMS M2M http://go/sams-m2m
+      SAMS_URL: https://accounts.sgdev.org
+      # Connect to Enterprise Portal running locally
+      CODY_GATEWAY_ENTERPRISE_PORTAL_URL: http://localhost:6081
+    externalSecrets:
+      SAMS_CLIENT_ID:
```
chrsmith, line 1067: What is this doing exactly? I thought I had _already_ wired up SAMS to Cody Gateway, so it can authenticate requests to a diagnostic set of APIs. We wouldn't want to have Cody Gateway require _two different_ SAMS OAuth client IDs, right?

In other words, I have no idea what I'm talking about, but just want to make sure you were aware that we have SAMS crews wired up to Cody Gateway "somehow" here:
https://github.com/sourcegraph/sourcegraph/blob/main/cmd/cody-gateway/internal/httpapi/maintenance.go#L32

chrsmith (approved):
Lots of nitpicks and some suggestions to clean up the code. But nothing large enough to block on.

As far as I understand what you are trying to accomplish, and how this aspect of Cody Gateway works these changes all LGTM. However, given the potential for _something_ going awry, do we want to intermix this new code changes and the necessary config changes together? Or instead, go with a slightly safer route where we control the entire "new codepath" with a single env var.

That way, if something does explode, it's just a matter of changing one env var to enable/disable this functionality. Instead of needing to back out a non-trivial PR under duress.

This doesn't seem particularly risky in any way, but given that we are adding a relatively new service as a dependency to the lynchpin of the entire Cody product offering, maybe we want to err on the side of caution? (Not trying to persuade you one way or another, just calling it out as something to consider.)

internal/grpc/grpcoauth/grpcoauth.go
bobheadxi: Yeah, that's the correct way, good catch :) This is retained from copy-paste (the change is just nesting this in `!env.InsecureDev`

cmd/cody-gateway/internal/actor/productsubscription/productsubscription.go
```diff
@@ -198,26 +212,24 @@ func removeUnseenTokens(seen collections.Set[string], cache ListingCache, syncLo
...
-			}
+	// Inspect the error to see if it's not-found error.
+	if statusErr, ok := status.FromError(err); ok && statusErr.Code() == codes.NotFound {
+		return nil, actor.ErrAccessTokenDenied{
+			Source: s.Name(),
+			Reason: "associated product subscription not found",
 		}
 	}
+
 	return nil, err
```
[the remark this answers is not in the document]
bobheadxi: Updated!

cmd/cody-gateway/internal/actor/productsubscription/productsubscription.go
```diff
@@ -252,71 +263,68 @@ func (s *Source) fetchAndCache(ctx context.Context, token string) (*actor.Actor,
...
 // subscription. It returns an empty string if no account name is available.
-func getSubscriptionAccountName(s dotcom.ProductSubscriptionState) string {
-	// 1. Check if the special "customer:" tag is present
-	if s.ActiveLicense != nil && s.ActiveLicense.Info != nil {
-		for _, tag := range s.ActiveLicense.Info.Tags {
-			if strings.HasPrefix(tag, "customer:") {
-				return strings.TrimPrefix(tag, "customer:")
-			}
+func getSubscriptionAccountName(activeLicenseTags []string) string {
+	// Check if the special "customer:" tag is present
```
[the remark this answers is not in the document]
bobheadxi: These tag values just work as they always have before, yeah (we're still reading from the dotcom database at this point)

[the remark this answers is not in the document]
bobheadxi: also, nit on the nit 😛  I do find that the "add period" docstrings are not _super_ productive and dilute the extra-good feedback, for not a _ton_ of gain. Correct spelling mistakes and grammar errors for sure as that can seriously hurt readability, but I'm not sure the missing periods hurt the understandability of this statement 😁 

But for now I'll address all the missing periods you already pointed out!

cmd/cody-gateway/internal/actor/productsubscription/productsubscription.go
```diff
@@ -252,71 +263,68 @@ func (s *Source) fetchAndCache(ctx context.Context, token string) (*actor.Actor,
...
+func newActor(
+	source *Source,
+	token string,
+	s *codyaccessv1.CodyGatewayAccess,
+	concurrencyConfig codygateway.ActorConcurrencyLimitConfig,
+) *actor.Actor {
+	name := s.GetSubscriptionDisplayName()
 	if name == "" {
-		name = s.Uuid
+		name = s.GetSubscriptionId()
```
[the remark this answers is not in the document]
bobheadxi: Yes, the subscription ID is the UUID (+ prefix) as we know today. Because we expose all sorts of IDs all over the place... this standardizes on the SAMS convention of exposing explicit, external, prefixed UUIDs

The prefix difference here is not important because this is just for diagnostics

cmd/cody-gateway/internal/actor/productsubscription/productsubscription.go
```diff
@@ -252,71 +263,68 @@ func (s *Source) fetchAndCache(ctx context.Context, token string) (*actor.Actor,
...
-		(s.ActiveLicense == nil || s.ActiveLicense.Info == nil ||
-			!containsOneOf(s.ActiveLicense.Info.Tags, licensing.DevTag, licensing.InternalTag))
-
 	now := time.Now()
 	a := &actor.Actor{
-		Key:           token,
-		ID:            s.Uuid,
+		Key: token,
+
+		// Maintain consistency with existing non-prefixed IDs
```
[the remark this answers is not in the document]
bobheadxi: Addressed, but https://github.com/sourcegraph/sourcegraph/pull/62934/files/ca862247ea1309b590b32febf2c38f42b4a80355#r1625263965 😁

cmd/cody-gateway/internal/actor/productsubscription/productsubscription.go
```diff
@@ -252,71 +263,68 @@ func (s *Source) fetchAndCache(ctx context.Context, token string) (*actor.Actor,
...
+		ID: strings.TrimPrefix(s.GetSubscriptionId(), subscriptionsv1.EnterpriseSubscriptionIDPrefix),
+
 		Name:          name,
-		AccessEnabled: !disallowedLicense && !s.IsArchived && s.CodyGatewayAccess.Enabled,
+		AccessEnabled: s.GetEnabled(),
```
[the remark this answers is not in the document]
bobheadxi: Yes:

1. Disallowed == internalMode && license.tags.containsOneOf("dev", "internal") which is now simply not served by Enterprise Portal dev. In the future, dev licenses can just be created in the dev instance, and we can worry less about abuse on our dev services
2. Archived: this is excluded from the new Enterprise Portal API, because it's now "get active Cody Access" and if an entity does not have access, we shouldn't provide it and shouldn't have to worry extra about it

cmd/cody-gateway/internal/actor/productsubscription/productsubscription.go
```diff
@@ -252,71 +263,68 @@ func (s *Source) fetchAndCache(ctx context.Context, token string) (*actor.Actor,
...
 		EndpointAccess: map[string]bool{
-			"/v1/attribution": !disallowedLicense && !s.IsArchived,
+			// always enabled even if !s.GetEnabled(), to allow BYOK customers
```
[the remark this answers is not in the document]
bobheadxi: I'm not super sure, I think it's a feature. Made the docstring based on the diff that introduced it, https://github.com/sourcegraph/sourcegraph/pull/60186 - cc @cbart . My understanding is that this is a Cody feature that is available via Cody Gateway, and available for anyone with any sort of "cody access" record (i.e. non-archived subscriptions)

cmd/cody-gateway/internal/actor/productsubscription/productsubscription_test.go
bobheadxi: Aha, in this case this _was_ a TODO for myself, my bad! Pushing up some test cases that generate golden assertions against the JSON representation of actors in various scenarios

internal/grpc/grpcoauth/grpcoauth.go
bobheadxi: I think this would be a question for SAMS local DX, which @unknwon had some ideas around - to be honest I'm not 100% sure what the plan might be in this scenario long-term. I imagine we could extend the tooling that sets up `sourcegraph.test` at some point - @unknwon any thoughts?

sg.config.yaml
bobheadxi: Yes, but it's not wired up in local dev, i.e. someone working on Cody Gateway must handcraft `sg.config.overwrite.yaml` or set their own env to get started. This makes it so you have to do absolutely nothing, and everything will "just work" - see PR description `# Local development` section :)

cmd/cody-gateway/shared/main.go
bobheadxi: Thanks, put it into `cmd/cody-gateway/internal/actor/productsubscription/enterpriseportal` with a single, tidy `Dial` entrypoint with this code :)

I think the other functionality setting up configuration _does_ belong here, as they just glue all the bits together into the various domain-specific constructors (connecting, getting SAMS credentials, scopes, and the like)

cmd/cody-gateway/internal/actor/productsubscription/productsubscription.go
```diff
@@ -198,26 +212,24 @@ func removeUnseenTokens(seen collections.Set[string], cache ListingCache, syncLo
 	}
 }
 
-func (s *Source) checkAccessToken(ctx context.Context, token string) (*dotcom.CheckAccessTokenResponse, error) {
-	resp, err := dotcom.CheckAccessToken(ctx, s.dotcom, token)
+func (s *Source) checkAccessToken(ctx context.Context, token string) (*codyaccessv1.CodyGatewayAccess, error) {
+	resp, err := s.enterprisePortal.GetCodyGatewayAccess(ctx, &codyaccessv1.GetCodyGatewayAccessRequest{
```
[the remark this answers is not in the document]
bobheadxi: > We are calling this pretty frequently since it is on the hot path for verifying user-supplied credentials. So are we confident that the Enterprise Portal will be able to serve this particular API endpoint quickly and reliably?

This code path is only reached on cache miss (see `auth_test.go`), so it's not as frequent as you might think. By default, the "full sync" implementation will populate the cache - this is a fallback mechanism.

That said, we have good reasons to believe this will work at least equivalently to dotcom:

1. Networking is similar (CF -> LB -> thing)
2. Backend is similar (request -> same Cloud SQL instance -> response). If anything this will probably be better, because we use a single roundtrip to DB (as opposed to N round trips for N subscriptions today) and a bundled Cloud SQL proxy (as opposed to standalone Cloud SQL proxy service in dotcom)

We do have all the standard MSP alerting, and we have metrics on the various RPCs through middleware. I will set up our own alerting on response codes, latencies, and the like as well before we roll out the new dependency to prod

bobheadxi:
> However, given the potential for something going awry, do we want to intermix this new code changes and the necessary config changes together? Or instead, go with a slightly safer route where we control the entire "new codepath" with a single env var.
>
> That way, if something does explode, it's just a matter of changing one env var to enable/disable this functionality. Instead of needing to back out a non-trivial PR under duress.
>
> This doesn't seem particularly risky in any way, but given that we are adding a relatively new service as a dependency to the lynchpin of the entire Cody product offering, maybe we want to err on the side of caution? (Not trying to persuade you one way or another, just calling it out as something to consider.)

All good points @chrsmith :) My thoughts:

1. Intermixing code and config changes: the config change surface is fairly minimal and has already been prepared out-of-band (see "Rollout plan") - I haven't added the Enterprise Portal address yet, but I'll get to that before doing anything. The config is harmless if we must roll back this change.
2. Having separate code paths for new and old: I have some concerns about this, as that would add some extra complexity to a pretty deeply involved codebase already. I also thought that this might make for an easier review, to see what existing mechanisms are removed and replaced with, as opposed to 100% positive diff 😬 Which brings us to:
3. Rolling back quickly: we can roll this back _quickly_ by reverting to an older revision of the image in infrastructure, and then handling the (possibly painful, as you say) code revert separately.

But, I'd like to point out that we have some safeguards to ensure the reliability of Enterprise Portal and this feature as a whole:

1. The "Rollout plan" section in PR also talks about how, once reviewed, I plan to cut a build to Cody Gateway's Dev instance and work with that for a little while (before merging anything).
2. Enterprise Portal has dev -> prod rollouts set up, and nothing will hit prod without verification in dev.
3. Existing alerting in Cody Gateway

There's a bit of polish I need to do before rolling this PR out to production as well (custom alerting on the metrics that we've added, verifying our opsgenie setup again, etc) we have robust visibility into things going wrong.

internal/grpc/grpcoauth/grpcoauth.go
unknwon: 😂 Why am I feeling like you two talking two different things, but anyway...


I think the change here is at least not "out of the way", setting up certificates for local dev is just pain in the ass as the state of the art. This package provides an option to skip the certificate part and focus on building the business logic (which IMO, is a very common practice) from a gRPC client to server, when both running locally (and need to be ran locally because cascading changes). Ideally, the "google.golang.org/grpc/credentials/oauth" should be possible to achieve this goal, but if it doesn't yet, this is a very thin wrapper.

bobheadxi:
Cutting candidate images: https://buildkite.com/sourcegraph/sourcegraph/builds/277206

cmd/cody-gateway/internal/actor/productsubscription/productsubscription.go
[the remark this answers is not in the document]
bobheadxi: Found a bit of a gotcha in our assumptions because of this feature actually: https://github.com/sourcegraph/sourcegraph/pull/63159

bobheadxi:
Made some final testing and tweaks to validate things - see comment thread here:

https://linear.app/sourcegraph/issue/CORE-98/enterprise-portal-use-portal-from-cody-gateway#comment-80991a40

PR #62939, opened 2024-05-28
Ty/plg-subscription-management-page-with-react-query

taras-yemets:
## Test plan
- CI
- Tested manually (see test plan from https://github.com/sourcegraph/sourcegraph/pull/62754

### DIFF
--- a/client/web/src/cody/management/api/hooks/useApiClient.tsx
+++ b/client/web/src/cody/management/api/hooks/useApiClient.tsx
@@ -1,6 +1,6 @@
-import { useEffect, useState, useContext, useCallback } from 'react'
+import { useEffect, useState, useContext } from 'react'
 
-import type { Call } from '../client'
+import { Call } from '../client'
 import { CodyProApiClientContext } from '../components/CodyProApiClient'
 
 export interface ReactFriendlyApiResponse<T> {
@@ -8,7 +8,6 @@ export interface ReactFriendlyApiResponse<T> {
     error?: Error
     data?: T
     response?: Response
-    refetch: () => Promise<void>
 }
 
 // useApiCaller will issue a REST API call to the backend, returning the
@@ -27,47 +26,62 @@ export function useApiCaller<Resp>(call: Call<Resp>): ReactFriendlyApiResponse<R
     const [data, setData] = useState<Resp | undefined>(undefined)
     const [response, setResponse] = useState<Response | undefined>(undefined)
 
-    const callApi = useCallback(async () => {
-        setLoading(false)
-        try {
-            const callerResponse = await caller.call(call)
+    useEffect(() => {
+        // `ignore` tracks if we should discard any results, because of any underlying race condition
+        // in the sequence of API calls. We return a handle to this in the function callback, which
+        // the React runtime may invoke (setting ignore = true) outside of our view.
+        // https://react.dev/reference/react/useEffect#fetching-data-with-effects
+        // https://maxrozen.com/race-conditions-fetching-data-react-with-useeffect
+        let ignore = false
 
-            // If we received a 200 response, all is well. We can just return
-            // the unmarshalled JSON response object as-is.
-            if (callerResponse.response.ok) {
-                setData(callerResponse.data)
-                setError(undefined)
-                setResponse(callerResponse.response)
-            } else {
-                // For a 4xx or 5xx response this is where we provide any standardized logic for
-                // error handling. For example:
-                //
-                // - On a 401 response, we need to force-logout the user so they can refresh their
-                //   SAMS access token.
-                // - On a 500 response, perhaps replace the current UI with a full-page error. e.g.
-                //   http://github.com/500 or http://github.com/501
-                setData(undefined)
-                setError(new Error(`unexpected status code: ${callerResponse.response.status}`))
-                setResponse(callerResponse.response)
+        ;(async () => {
+            try {
+                const callerResponse = await caller.call(call)
 
-                // Provide a clearer message. A 401 typically comes from the user's SAMS credentials
-                // having expired on the backend.
-                if (callerResponse.response.status === 401) {
-                    setError(new Error('Please log out and log back in.'))
+                if (ignore) {
+                    return
                 }
+
+                // If we received a 200 response, all is well. We can just return
+                // the unmarshalled JSON response object as-is.
+                setLoading(false)
+                if (callerResponse.response.status >= 200 && callerResponse.response.status <= 299) {
+                    setData(callerResponse.data)
+                    setError(undefined)
+                    setResponse(callerResponse.response)
+                } else {
+                    // For a 4xx or 5xx response this is where we provide any standardized logic for
+                    // error handling. For example:
+                    //
+                    // - On a 401 response, we need to force-logout the user so they can refresh their
+                    //   SAMS access token.
+                    // - On a 500 response, perhaps replace the current UI with a full-page error. e.g.
+                    //   http://github.com/500 or http://github.com/501
+                    setData(undefined)
+                    setError(new Error(`unexpected status code: ${callerResponse.response.status}`))
+                    setResponse(callerResponse.response)
+
+                    // Provide a clearer message. A 401 typically comes from the user's SAMS credentials
+                    // having expired on the backend.
+                    if (callerResponse.response.status === 401) {
+                        setError(new Error('Please log out and log back in.'))
+                    }
+                }
+            } catch (error) {
+                if (ignore) {
+                    return
+                }
+                setData(undefined)
+                setError(error)
+                setResponse(undefined)
+                setLoading(false)
             }
-        } catch (error) {
-            setData(undefined)
-            setError(error)
-            setResponse(undefined)
-        } finally {
-            setLoading(false)
+        })()
+
+        return () => {
+            ignore = true
         }
     }, [call, caller])
 
-    useEffect(() => {
-        void callApi()
-    }, [callApi])
-
-    return { loading, error, data, response, refetch: callApi }
+    return { loading, error, data, response }
 }
--- a/client/web/src/cody/management/api/react-query/QueryClientProvider.tsx
+++ b/client/web/src/cody/management/api/react-query/QueryClientProvider.tsx
@@ -1,0 +1,20 @@
+import type { ReactNode } from 'react'
+
+import { QueryClient, QueryClientProvider as ReactQueryClientProvider } from '@tanstack/react-query'
+
+const queryClient = new QueryClient({
+    defaultOptions: {
+        queries: {
+            // If query failed, it's not likely that refetching it will succeed, so don't retry.
+            retry: false,
+
+            // Once loaded, we consider query result to be stable. No need to refetch it again.
+            // See: https://tanstack.com/query/latest/docs/framework/react/guides/important-defaults
+            staleTime: Infinity,
+        },
+    },
+})
+
+export const QueryClientProvider: React.FC<{ children?: ReactNode | undefined }> = ({ children }) => (
+    <ReactQueryClientProvider client={queryClient}>{children}</ReactQueryClientProvider>
+)
--- a/client/web/src/cody/management/api/react-query/subscriptions.ts
+++ b/client/web/src/cody/management/api/react-query/subscriptions.ts
@@ -1,0 +1,60 @@
+import {
+    useQuery,
+    useMutation,
+    useQueryClient,
+    type UseMutationResult,
+    type UseQueryResult,
+} from '@tanstack/react-query'
+
+import { type Call, Client, CodyProApiCaller } from '../client'
+import type { UpdateSubscriptionRequest, Subscription } from '../teamSubscriptions'
+
+const apiCaller = new CodyProApiCaller()
+
+// Wrapper around `apiCaller.call` which throws an error in case of a non-2xx response.
+// Motivation taken from here: https://tanstack.com/query/latest/docs/framework/react/guides/query-functions#usage-with-fetch-and-other-clients-that-do-not-throw-by-default
+const callCodyProApi = async <Data>(call: Call<Data>): Promise<{ data?: Data; response: Response }> => {
+    const result = await apiCaller.call(call)
+
+    if (!result.response.ok) {
+        throw new Error(`Cody Pro API call failed with status ${result.response.status}`)
+    }
+
+    // TODO: handle 401 error (e.g., navigate to "/-/sign-out")
+
+    return result
+}
+
+// Motivation to use query factories taken from here: https://tkdodo.eu/blog/effective-react-query-keys#use-query-key-factories
+const queryKeys = {
+    all: ['subscription'] as const,
+    subscription: () => [...queryKeys.all, 'current-subscription'] as const,
+    subscriptionSummary: () => [...queryKeys.all, 'current-subscription-summary'] as const,
+}
+
+export const useCurrentSubscription = (): UseQueryResult<{ data?: Subscription; response: Response }> =>
+    useQuery({
+        queryKey: queryKeys.subscription(),
+        queryFn: async () => callCodyProApi(Client.getCurrentSubscription()),
+    })
+
+export const useUpdateCurrentSubscription = (): UseMutationResult<
+    { data?: Subscription; response: Response },
+    Error,
+    UpdateSubscriptionRequest
+> => {
+    const queryClient = useQueryClient()
+    return useMutation({
+        mutationFn: async requestBody => callCodyProApi(Client.updateCurrentSubscription(requestBody)),
+        onSuccess: data => {
+            // We get updated subscription data in response - no need to refetch subscription.
+            // All the `queryKeys.subscription()` subscribers (`useCurrentSubscription` callers) will get the updated value automatically.
+            queryClient.setQueryData(queryKeys.subscription(), data)
+
+            // Invalidate `queryKeys.subscriptionSummary()` queries. If the subscription summary is a subset of subscription, we can
+            // derive the updated subscription summary from the subscription response eliminating the need in subscription summary query invalidation
+            // causing data refetching.
+            return queryClient.invalidateQueries({ queryKey: queryKeys.subscriptionSummary() })
+        },
+    })
+}
--- a/client/web/src/cody/management/subscription/manage/CodySubscriptionManagePage.tsx
+++ b/client/web/src/cody/management/subscription/manage/CodySubscriptionManagePage.tsx
@@ -18,8 +18,8 @@ import {
 } from '../../../../graphql-operations'
 import type { LegacyLayoutRouteContext } from '../../../../LegacyRouteContext'
 import { USER_CODY_PLAN } from '../../../subscription/queries'
-import { Client } from '../../api/client'
-import { useApiCaller } from '../../api/hooks/useApiClient'
+import { QueryClientProvider } from '../../api/react-query/QueryClientProvider'
+import { useCurrentSubscription } from '../../api/react-query/subscriptions'
 
 import { InvoiceHistory } from './InvoiceHistory'
 import { PaymentDetails } from './PaymentDetails'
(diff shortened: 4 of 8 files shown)


client/web/src/cody/management/subscription/manage/PaymentDetails.tsx
```diff
@@ -142,56 +141,51 @@ const PaymentMethodForm: React.FC<PaymentMethodFormProps> = props => {
...
+    const [isStripeLoading, setIsStripeLoading] = useState(false)
+    const [stripeErrorMessage, setStripeErrorMessage] = useState('')
+
+    const [isErrorVisible, setIsErrorVisible] = useState(true)
 
-    const [isLoading, setIsLoading] = useState(false)
-    const [errorMessage, setErrorMessage] = useState('')
+    const isLoading = isStripeLoading || updateCurrentSubscriptionMutation.isPending
+    const errorMessage =
+        stripeErrorMessage || updateCurrentSubscriptionMutation.isError ? updateSubscriptionMutationErrorText : ''
```
taras-yemets, line 153: The form submit handler has two types of async operations: loading and validating Stripe elements and making an SSC request. 
We handle error and loading states for these two types separately and then use a derived values (`isLoading`, `errorMessage`) to define the component return value.

This logic is repeated in `BillingAddressForm` component and we may want to extract the shared part in future, but IMO it's out of scope of this PR.

client/web/src/cody/management/subscription/manage/SubscriptionDetails.tsx
```diff
@@ -101,79 +75,84 @@ export const SubscriptionDetails: React.FC<SubscriptionDetailsProps> = props =>
...
-                            {props.isLoading ? (
-                                <LoadingSpinner className="mr-1" />
-                            ) : (
-                                <Icon aria-hidden={true} className="mr-1" svgPath={mdiCheck} />
-                            )}
-                            <Text as="span">Yes, cancel</Text>
-                        </Button>
-                    </div>
-                </Modal>
-            )}
```
taras-yemets, line 176: I moved the logic for the `RenewSubscriptionButton` and `CancelSubscriptionButton` components back to `SubscriptionDetails`. The error message is rendered within `SubscriptionDetails`, and passing the `mutation` as a prop to the button components made the logic less readable. The readability of `CancelSubscriptionButton` would be even worse, as it would need to pass an `onSettled` callback to close the modal after the mutation completes.

IMO, this change is out of the scope of this PR. This is more of an FYI comment explaining why this kind of change is in a PR advertising the third-party state management library.

vdavid (commented):
@taras-yemets, I like the proposed solution, I think it's nice and clean. [The article you linked](https://tkdodo.eu/blog/why-you-want-react-query) is also very clear and convincing, and their docs seem clean and helpful. The [repo](https://github.com/TanStack/query) is popular and active. So yeah, I'm absolutely for using this. 👍

client/web/src/cody/management/api/hooks/useApiClient.tsx
```diff
@@ -1,14 +1,13 @@
-import { useEffect, useState, useContext, useCallback } from 'react'
+import { useEffect, useState, useContext } from 'react'
```
taras-yemets, line 1: Ignore changes to `useApiClient.tsx` from this PR. It's just reverted to the `main` rev.

client/web/src/cody/management/api/react-query/subscriptions.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    if (!result.response.ok) {
+        if (result.response.status === 401) {
+            throw new CodyProApiAuthError('Unauthorized. Please log out and log back in.')
+        }
+        throw new Error(`Cody Pro API call failed with status ${result.response.status}`)
+    }
+
+    return result
+}
```
taras-yemets, line 59: Custom query hooks (e.g., `useCurrentSubscription`) and mutation hooks (e.g., `useUpdateCurrentSubscription`) may need to show custom error messages depending on the context. For example, both the card number form and the cancel subscription button use the `useUpdateCurrentSubscription` hook, but they may need to display different error messages in case of an error (e.g., "Failed to save credit card" for the former and "Failed to change subscription status" for the latter). Therefore, I leave it to the callers to define the error message.

However, there is a specific case when a request to the Cody Pro backend results in a 401 error due to expired SAMS credentials (we may have more such "special" cases in the future). In this situation, regardless of the caller context, we want to show a generic message such as "You are not authorized. Sign out and sign back in." To handle this, we need callers to distinguish errors returned from the query/mutation. For "special" cases (as identified by the `getCodyProApiErrorMessage` method), the specific error message should be shown. For any other errors, the custom message should depend on the caller's context.

I am not entirely satisfied with the current approach, where callers must call `getCodyProApiErrorMessage` to get the message text. However, I haven't found a better solution. Do you have better ideas, @vdavid, @chrsmith?

vdavid: I'd probably get rid of `getCodyProApiErrorMessage` and just do
```ts
isCodyProApiAuthError(err) ? err.message : 'Custom error message'
``` 
whenever I want this logic.

If you don't think that's the right call here, then I suggest renaming the `message` parameter to `defaultMessage`, to better communicate its use, and maybe also mention it in `getCodyProApiErrorMessage`'s doc that we intend to make this function more generic once we have more special cases.

taras-yemets: Thank you, @vdavid! 

```
isCodyProApiAuthError(err) ? err.message : 'Custom error message'
```
With this approach to show custom (component-specific) errors in components, we need to throw empty errors in react-query hooks, and it doesn't feel right.

But the `defaultErrorMessage` approach would work, I think.

taras-yemets: Ah, now I see. Thank you!

chrsmith: I am 100,000% in support of adding more formality / types to describe the errors we receive. But let me suggest something slightly different...

Currently the backend serves 400 errors as a raw string, e.g. `"Not Found: No invite with that ID was found."`

The problem however, is that without any structure the frontend isn't able to _understand_ what the error is talking about. (There could be different causes for the 400 error, and we may want to show different error messages to match.) And this "401 / auth error" is another instance of that.

So what if we did the following:

1. Update the SSC REST API to no longer return errors (4xx and 5xx responses) as strings, and instead return a well-known error response type. e.g. `{ message: string, cause: string }`.
2. Update the `ssc_proxy.go` to surface 401/403 errors in this format as well.

That way, we can use a `CodyProApiError` type consistently, and not need to special case `isCodyProApiAuthError`. e.g.

```ts
if (subscriptionQueryResult.error) {
    if (subscriptionQueryResult.error.cause === 'expired_credentials') {
        return <Alert variant="danger">Could you do me a solid and login again?</Alert>
    }
    return <Alert variant="danger">{ subscriptionQueryResult.error.message }</Alert>
}
```

Ultimately it would be great if we could enforce that if `subscriptionQueryResult.error` is truthy, then `subscriptionQueryResult.error.message` will _always_ be safe to show to end users. And if the client wants to handle specific error cases, they can inspect `subscriptionQueryResult.error.cause`.

What do you think? (It would be more of a pain to make this type of breaking API change _after_ we launch Teams and Invites, so if we think this is something we would like, we should just go ahead and make the change. (See `backend/internal/rest/response.go` in the SSC codebase, as that's how we render the HTTP response bodies on the backend.)

client/web/src/cody/management/api/react-query/subscriptions.ts
```diff
@@ -0,0 +1,93 @@
...
+
+// Motivation to use query factories taken from here: https://tkdodo.eu/blog/effective-react-query-keys#use-query-key-factories
```
chrsmith, line 61: 💯 to adding a link to get some more background on potentially confusing or highly-sophisticated code snippets, such as this.

But it would be even better to explain what this actually _is_. e.g. it's nice I can go and read up on a query-key-factory in the linked blog. But can we at just put a one-sentence description here?

e.g.

```ts
// queryKeys provides the way for the dinglehop to reuse the grumbo.
```

client/web/src/cody/management/subscription/manage/InvoiceHistory.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+    if (error) {
+        logger.error('Error fetching current subscription invoices', error)
+        return null
+    }
+
+    if (response && !response.ok) {
+        if (response.status === 401) {
+            return <Navigate to="/-/sign-out" replace={true} />
+        }
```
chrsmith, line 33: I would love for us to not need to write this same three lines of code in a dozen different places. And instead provide some sort of infrastructure for handling this automatically. e.g. something like:

```ts
// buildCodyProComponent will take care of handling common states like:
// while the API call is loading, rendering a loading bar,
// or if we receive a 401 error redirecting the user to log-out.
return buildCodyProComponent(
    () => useApiCaller(invoicesCall),
    (data) => {
        return <>
            <H2 className="mb-4">Invoice history</H2>

            <hr className={classNames('w-100', styles.divider)} />

            {data.invoices.length ? (
                <ul className="mb-0 list-unstyled">
                    {data.invoices.map(invoice => (
                        <InvoiceItem key={invoice.periodStart} invoice={invoice} />
                    ))}
                </ul>
            ) : (
                <Text>You have no invoices.</Text>
            )}
        </>
    })
```

It's probably impossible to have a "one size fits all" solution for every conceivable case. But if there is some "common thing we know we will do pretty often", just doing that "common thing in one place" is preferable.

chrsmith (commented):
There's a whole lot of good stuff in this PR to discuss, and work through. I'm generally onboard with the idea but with some caveats.

## re: Approval, I'll defer to @vdavid 

I have ... _opinions_ on whether or not we should rely on a library like `react-query` or similar.

Personally, I am very nervous when a library/framework does things like "caching" or "resource sharing" because these are things that are critically important to how the product works. And therefore are things that I would want to have _direct_ control over. Even if it means needing to find, fix, debug, and spend months iterating on obscure data races and interactions between async data requests and user navigation to different components.

For example, I -- for better or for worse -- want to know _exactly_ when and under what conditions we make an API call to the server. And any 3rd party layers in-between the user's action and calls to `fetch()` get in the way of that.

However, given that I am probably in the minority among frontend or full-stack developers in that belief I'm totally happy going along with what y'all think is best. So if you can convince @vdavid this is the way to go, let's do it!

## re: this new usage pattern

> 1. useSomeResource hook: Fetches the resource. Regardless of how many components call it, the data would be fetched once, and all instances will get the latest data.
> 2. useUpdateResource hook: Returns methods to mutate the resource and provides the mutation state.

This sounds great. The things you are adding here are easy to understand, and appear straight-forward to use. 👍 

However, I think we need to be really careful about caching. It is definitely going to be the case that somebody will modify the Cody Pro team or its subscription outside of the current browser session. And so I'd suggest we err on the side of "refetching the same data" than showing stale-and-incorrect results to users. (Since it's super-trusting and annoying to troubleshoot bugs that suddenly disappear if you press F5.)

## meta goal

Again, I'll gladly let you and David figure out what React idioms we want to use for loading data on the frontend. If I were to just explicitly say what I would personally shoot for in terms of requirements:

- Avoid duplicating "dumb" code. e.g. even in this PR we have both `<LoadingSpinner className="mx-auto" />` and `<LoadingSpinner />`... if we can just have a single place where we define the "loading spinner" that would be great. Same thing for the "if 401 error, then redirect to `/sign-out`. Because inevitably we'll wind up with a half dozen slightly different variations on that same functionality.

- Eliminate invalid states entirely. This was one of the design goals when adding the `userApiCaller` and that `Call<T>` abstraction. It's OK to have a single file or module with a bunch of gnarly complexity, but we want to make it hard to screw things up. (e.g. if I call `userQuery` to make a POST request... I'm going to have a bad time, aren't I? Can we somehow lean into the type system to eliminate that from happening? See [Falling into the Pit of Success](https://blog.codinghorror.com/falling-into-the-pit-of-success/).

client/web/src/cody/management/api/react-query/QueryClientProvider.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+const queryClient = new QueryClient({
+    defaultOptions: {
+        queries: {
+            // If query failed, it's not likely that refetching it will succeed, so don't retry.
+            retry: false,
+
+            // Once loaded, we consider query result to be stable. No need to refetch it again.
+            // See: https://tanstack.com/query/latest/docs/framework/react/guides/important-defaults
+            staleTime: Infinity,
```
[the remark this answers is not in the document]
taras-yemets: Makes sense.

> Query instances via useQuery or useInfiniteQuery by default consider cached data as stale.

Let's stick to ☝🏻 [defaults](https://tanstack.com/query/latest/docs/framework/react/guides/important-defaults) until we need something special here.

client/web/src/cody/management/api/react-query/subscriptions.ts
taras-yemets: > So what if we did the following:
Update the SSC REST API to no longer return errors (4xx and 5xx responses) as strings, and instead return a well-known error response type. e.g. { message: string, cause: string }.
Update the ssc_proxy.go to surface 401/403 errors in this format as well.

Sounds good to me! I'd like it implement it in the follow-up PR. 



For this PR, I suggest we show custom error messages for each component ([example](https://github.com/sourcegraph/sourcegraph/blob/af3abe11682c4f3d0e41954d7c9ea877f7eb5259/client/web/src/cody/management/subscription/manage/SubscriptionDetails.tsx#L39-L41) - it what we basically had in the SSC codebase) and [handle 401 in a call API utility function](https://github.com/sourcegraph/sourcegraph/blob/af3abe11682c4f3d0e41954d7c9ea877f7eb5259/client/web/src/cody/management/api/react-query/callCodyProApi.ts#L22-L49) (as `useApiCaller` does). In this way we eliminate weird `isCodyProApiAuthError` checks. WDYT?

client/web/src/cody/management/api/react-query/QueryClientProvider.tsx
```diff
@@ -0,0 +1,20 @@
+import type { ReactNode } from 'react'
+
+import { QueryClient, QueryClientProvider as ReactQueryClientProvider } from '@tanstack/react-query'
+
+const queryClient = new QueryClient({
```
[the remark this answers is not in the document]
taras-yemets: Addressed in https://github.com/sourcegraph/sourcegraph/pull/62939/commits/d8e2da4cabf46769bcff9b9b9fae92535263aa16

client/web/src/cody/management/CodyManagementPage.tsx
```diff
@@ -133,11 +132,9 @@ export const CodyManagementPage: React.FunctionComponent<CodyManagementPageProps
                                 <ButtonLink
                                     variant="primary"
                                     size="sm"
-                                    href={manageSubscriptionRedirectURL}
```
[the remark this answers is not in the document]
taras-yemets: > using a <ButtonLink> is essentially a "foot gun" if you want to record click events using the telemetry recorder

Can you elaborate on why it's dangerous, please?

> would a "proper" fix for this to be making this a first-class thing? e.g. Adding a new property to <ButtonLink> like

`ButtonLink` is a wildcard component and IMO it's hard to justify introducing such a specific prop. Possible counter arguments: "do we need a record select/hover/focus/blur event handler too?"

taras-yemets:
@chrsmith, thanks for your review!

> However, I think we need to be really careful about caching. It is definitely going to be the case that somebody will modify the Cody Pro team or its subscription outside of the current browser session. And so I'd suggest we err on the side of "refetching the same data" than showing stale-and-incorrect results to users.

Valid concern. Rolled back to the defaults https://github.com/sourcegraph/sourcegraph/pull/62939#discussion_r1624421063. 

> Avoid duplicating "dumb" code. e.g. even in this PR we have both <LoadingSpinner className="mx-auto" /> and <LoadingSpinner />... if we can just have a single place where we define the "loading spinner" that would be great. Same thing for the "if 401 error, then redirect to /sign-out. Because inevitably we'll wind up with a half dozen slightly different variations on that same functionality.

I agree with the 401 handling  as mentioned in (https://github.com/sourcegraph/sourcegraph/pull/62939#discussion_r1624754215). It’s an app-level (I mean Cody Pro app) issue since we can’t make calls to the Cody Pro backend without a valid token. 
However, handling loading states requires a more nuanced approach. For instance, when loading subscription data for the first time, the entire page should be in a loading state as there’s nothing to display yet. On the other hand, when re-fetching subscription data due to refocusing, data becoming stale, or mutations, we might want to manage the loading state differently or even ignore it altogether.

> Eliminate invalid states entirely. This was one of the design goals when adding the userApiCaller and that Call<T> abstraction. It's OK to have a single file or module with a bunch of gnarly complexity, but we want to make it hard to screw things up. (e.g. if I call userQuery to make a POST request... I'm going to have a bad time, aren't I? Can we somehow lean into the type system to eliminate that from happening?

The custom query hooks still use `Call<T>` under the hood. 

Instead of using a custom hook like `useCurrentSubscription`, one can import `useQuery` directly and potentially use it incorrectly (e.g., with the wrong method, URL, etc.). It’s up to the caller to ensure it’s used properly. From our side, we can document the current approach, such as recommending using custom hooks for fetching and mutating resources via the Cody Pro API, and add tests to hooks and components. This way, we can ensure 🤞🏻 that the `useCurrentSubscription` hook (for example) uses the correct method and URL, and returns data of the appropriate shape.
However, implementing a custom lint rule to forbid direct useQuery usage would be overkill, in my opinion, as it’s a completely valid method to use.

Additionally, useQuery can be used with POST requests. In fact, most (if not all) usages of useQuery (Apollo one) in our codebase are POST requests to our GraphQL API. But with the current subscription endpoint, POST of course will result in an error. So it's up to developer to ensure it's used properly.

client/web/src/cody/management/api/react-query/callCodyProApi.ts
```diff
@@ -0,0 +1,48 @@
+import type { Call } from '../client'
+
+const buildRequest = (url: string, { headers = {}, ...init }: RequestInit = {}): Promise<Response> =>
```
vdavid, line 3: Not a blocker (easy to refactor later) but I find calling this `build*` misleading as it doesn't just build a request but actually sends it.
I'd ideally call this `fetch`, but of course, that's not possible, so I'd maybe go with `customFetch`, `fetchUrl`, `sourcegraphFetch`, or similar. Naming is hard.

client/web/src/cody/management/api/react-query/subscriptions.ts
```diff
@@ -0,0 +1,47 @@
...
+    type UseQueryResult,
+} from '@tanstack/react-query'
+
+import { Client } from '../client'
+import type { UpdateSubscriptionRequest, Subscription } from '../teamSubscriptions'
+
+import { callCodyProApi } from './callCodyProApi'
+
+// Use query key factories to re-use produced query keys in queries and mutations.
+// Motivation taken from here: https://tkdodo.eu/blog/effective-react-query-keys#use-query-key-factories
```
vdavid, line 15: Thanks for this comment; without it, the code would be a bit 🙃

client/web/src/cody/management/subscription/manage/CodySubscriptionManagePage.tsx
```diff
@@ -67,49 +66,31 @@ const AuthenticatedCodySubscriptionManagePage: React.FC<Props> = ({ telemetryRec
         return <Navigate to="/cody/manage/subscription/new" replace={true} />
     }
 
-    return <PageContent />
+    return (
+        <Page className="d-flex flex-column">
+            <PageContent />
+        </Page>
```
vdavid, line 72: Nit: Seems unrelated? Would be nice to call this out as a PR side-effect in the PR description

vdavid (approved):
Very nice! I like the simplicity, and all the extra goodies this comes with. Left a few nits.

client/web/src/cody/management/subscription/manage/CodySubscriptionManagePage.tsx
taras-yemets: It's a side-fix related to this PR. Previously we returned `<LoadingSpinner />` for the loading state and `null` for the error state. Now we return `<Alert>` for the error state. To ensure that either spinner, or alert, or actual page content have proper layout I moved the common wrapper one level higher (we use early returns for the data fetching states in the `PageContent` component - so that we don't need to wrap each return value with `<Page className="d-flex flex-column">` wrapper).

client/web/src/cody/management/api/react-query/callCodyProApi.ts
taras-yemets: 💯 makes sense!
Addressed in https://github.com/sourcegraph/sourcegraph/pull/62939/commits/a7e8133fec1c2a1e4db8f1f2f1fdab221f202c60

PR #62946, opened 2024-05-28
feat(code insights): language stats speed improvements by using archive loading

bahrmichael:
We previously improved the performance of Language Stats Insights by introducing parallel requests to gitserver: https://github.com/sourcegraph/sourcegraph/pull/62011

This PR replaces the previous approach where we would iterate through and request each file from gitserver with an approach where we request just one archive. This eliminates a lot of network traffic, and gives us an additional(!) performance improvement of 70-90%.

Even repositories like chromium (42GB) can now be processed (on my machine in just one minute).

---

Caching: We dropped most of the caching, and kept only the top-level caching (repo@commit). This means that we only need to compute the language stats once per commit, and subsequent users/requests can see the cached data. We dropped the file/directory level caching, because (1) the code to do that got very complex and (2) we can assume that most repositories are able to compute within the 5 minutes timeout (which can be increase via the environment variable `GET_INVENTORY_TIMEOUT`). The timeout is not bound to the user's request anymore. Before, the frontend would request the stats up to three times to let the computation move forward and pick up where the last request aborted. While we still have this frontend retry mechanism, we don't have to worry about an abort-and-continue mechanism in the backend.

---

Credits for the code to @eseliger: https://github.com/sourcegraph/sourcegraph/issues/62019#issuecomment-2119278481

I've taken the diff, and updated the caching methods to allow for more advanced use cases should we decide to introduce more caching. We can take that out again if the current caching is sufficient.

Todos:

- [x] Check if CI passes, manual testing seems to be fine
- [x] Verify that insights are cached at the top level

---

Test data:

- sourcegraph/sourcegraph: 9.07s (main) -> 1.44s (current): 74% better
- facebook/react: 17.52s (main) -> 0.87s (current): 95% better
- godotengine/godot: 28.92s (main) -> 1.98s (current): 93% better
- chromium/chromium: ~1 minute: 100% better, because it didn't compute before

## Changelog

- Language stats queries now request one archive from gitserver instead of individual file requests. This leads to a huge performance improvement. Even extra large repositories like chromium are now able to compute within one minute. Previously they timed out.

## Test plan

- New unit tests
- Plenty of manual testing


bahrmichael:
Currently two problems, the `pax_global_header` is probably from my changes in https://github.com/sourcegraph/sourcegraph/pull/62946/commits/ec96924a738785a1316e83d27bac8d64797f9f08.
Screenshot 2024-05-29 at 16 30 00
Screenshot 2024-05-29 at 16 29 49

eseliger:
for the pax_global_header you _might_ be able to get around that issue by adding a `fd.IsRegular()` check. It's a magical file that is unfortunately added to tar archives :|

bahrmichael:
> for the pax_global_header you _might_ be able to get around that issue by adding a `fd.IsRegular()` check. It's a magical file that is unfortunately added to tar archives :|

Thank you for the idea! I found that my changes to the reader have seemed to cause this, which I reverted in https://github.com/sourcegraph/sourcegraph/pull/62946/commits/fb2536ec3cb74a5c4ab6be7e55c08a0c18bf1066.

I'll double check the impact of pax_global_header so we don't report false numbers.

bahrmichael:
The pax_global_header is not counting towards the language stats, because it's skipped at https://github.com/sourcegraph/sourcegraph/blob/0e2a668954f2f8169da51224902865686fa8e472/cmd/frontend/internal/inventory/inventory.go#L89.

eseliger:
great!

cmd/frontend/internal/inventory/inventory.go
```diff
@@ -66,14 +61,8 @@ func getLang(ctx context.Context, file fs.FileInfo, getFileReader func(ctx conte
...
-		attribute.Bool("Safe", safe),
-		attribute.String("MatchedLang", matchedLang),
-	)
-
 	// No content
-	if rc == nil {
+	if rc == nil || skipEnhancedLanguageDetection {
 		lang.Name = matchedLang
 		lang.TotalBytes = uint64(file.Size())
 		return lang, nil
```
eseliger, line 68: could we save the call to getFileReader here at all, if `skipEnhancedLanguageDetection`?

cmd/frontend/internal/inventory/entries.go
```diff
@@ -1,17 +1,70 @@
...
 	"io/fs"
 	"sort"
+
+	"github.com/sourcegraph/sourcegraph/internal/gitserver"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
 )
 
+func (c *Context) All(ctx context.Context, gs gitserver.Client) (inv Inventory, err error) {
+	if c.CacheGet != nil {
+		if inv, ok := c.CacheGet(ctx, string(c.Repo)); ok {
```
eseliger, line 17: does the cache key need to account for what HEAD resolves to as well, or is that intentional to invalidate the cache less frequently?

cmd/frontend/internal/inventory/entries.go
```diff
@@ -1,17 +1,70 @@
...
+		defer func() {
+			if err == nil {
+				c.CacheSet(ctx, string(c.Repo), inv)
+			}
+		}()
+	}
+
+	r, err := gs.ArchiveReader(ctx, c.Repo, gitserver.ArchiveOptions{Treeish: "HEAD", Format: gitserver.ArchiveFormatTar})
+	if err != nil {
+		return Inventory{}, err
```
eseliger, line 31: I wonder if a RevisionNotFoundError here should mean empty inventory but no error? 

Use-case: An empty repo should IIRC throw this error, but that feels expected.

cmd/frontend/internal/inventory/entries.go
```diff
@@ -1,17 +1,70 @@
...
+	}
+
+	invs := make([]Inventory, 0)
+	tr := c.NewTarReader(r)
+	for {
+		th, err := tr.Next()
+		if err != nil {
+			if errors.Is(err, io.EOF) {
+				return Sum(invs), nil
```
eseliger, line 40: not sure how expensive that is but to save on memory here, we could build the sum for every or every N `th` we visit, not just keep 600k inventories in memory

cmd/frontend/internal/inventory/entries.go
```diff
@@ -1,17 +1,70 @@
...
+		case entry.Mode().IsRegular():
+			inv, err := c.fileTar(ctx, th, tr)
+			if err != nil {
+				return Inventory{}, err
+			}
+			invs = append(invs, inv)
+		case entry.Mode().IsDir():
+			// If we want to we could try to optimize cache invalidation at the tree level
+			// here. For now, we only iterate over all the files in the archive.
+			continue
```
eseliger, line 56: I'm not sure if we had that before, but this is the code path that would allow for incremental indexing. We can leave that out for now, but should monitor the latency and load this causes for large repos (i.e., let's check on large cloud customers after the release).

eseliger (approved):

cmd/frontend/internal/inventory/entries.go
eseliger: in that case, we should probably reimplement the caching at the dir level first

camdencheek: Okay, finally got the chance to take another look at this, and I think we should go ahead and move forward with this close to as-is, but using the commit hash as part of the cache key. We do some theoretical extra work by not caching at the tree level, but this also drastically reduces the load on the cache and massively improves the "cold start" time. Since very large repos previously would fail to index at all, this is a big improvement there, and 1.4s for sourcegraph/sourcegraph (a medium-sized repo) on a cache miss is acceptable IMO.

That said, I think we should follow this up by moving the language stats insight from an on-demand thing to a pre-computed thing, slotting it into the existing framework for computing updated values for code insights. That way, no user should see a cold start, and we will also have historical data for this so we could show language stats over time.

eseliger: that sounds great!

cmd/frontend/internal/inventory/entries.go
bahrmichael: At the end of May I gave it a try to cache based on the directory, and found that it's a bit harder than before. This is mainly because now we get a flat list of records, files and directories in them. 

I don't recall exactly if the directory came before the files or after, and am not sure if the order of files and directories is guaranteed. Can you confirm that @eseliger? 

I had two ideas: (1) Remember the directory, process files, and when a new directory appears, cache the above. Discarded that again because it felt too hacky. (2) Load everything into a tree structure and process like before, but then we'll need to load everything and actually lose the benefit directory level caching would bring us.

cmd/frontend/internal/inventory/inventory.go
bahrmichael: I think we can drop the nil check and move the getFileReader call downwards. Previously nil indicated that the language check should be skipped, but we now have that as a bool.

Previous nil init (see deleted code in this PR):
```
	if !useEnhancedLanguageDetection && !forceEnhancedLanguageDetection {
		// If USE_ENHANCED_LANGUAGE_DETECTION is disabled, do not read file contents to determine
		// the language. This means we won't calculate the number of lines per language.
		invCtx.NewFileReader = func(ctx context.Context, path string) (io.ReadCloser, error) {
			return nil, nil
		}
	}
```

New init of according bool:
```
ShouldSkipEnhancedLanguageDetection: !useEnhancedLanguageDetection && !forceEnhancedLanguageDetection,
```

cmd/frontend/internal/inventory/entries.go
bahrmichael: I've pushed an update in https://github.com/sourcegraph/sourcegraph/pull/62946/commits/7e0aeb3a542bf8e81adfa3c762b7e71d5c35c873 where I just took 1000, let me know if you think that a different value is better.

cmd/frontend/internal/inventory/entries.go
bahrmichael: I pushed a change that includes the commitID, and will spin up a ticket for moving this to a pre-computed style.

cmd/frontend/internal/inventory/entries.go
eseliger: > I don't recall exactly if the directory came before the files or after, and am not sure if the order of files and directories is guaranteed. Can you confirm that @eseliger?

I think the order should be guaranteed _within one commit_, but might differ between commits. No 100% guarantee - I'd need to check in the git code to be sure. How important is that? I don't think that's documented behavior of the API today, so would be something implicit we rely on.

bahrmichael: Within one commit is fine. I tried to use the folder as an open or close marker to its files. To determine when all files in a folder have been read and caching that before the whole process completes, the guaranteed order is essential.

Here's what I recall seeing:

```
...
folder1
folder1/a
folder2/a
folder2/b
folder2/c
folder2
folder3/a
...
```

If the enclosing folder always comes before or after the files, we can use the leading or trailing folder to determine when all files have been read. That's all the guarantees I'm looking for here. Either before or after, but always right before or right after all of its files.

cmd/frontend/internal/inventory/entries.go
bahrmichael: Tested it, and this is the error we get on a fresh repo without any commits:

Screenshot 2024-06-26 at 12 06 24

Once there's a commit, we can see this behavior (introduced originally by https://github.com/sourcegraph/sourcegraph/pull/20856).

Screenshot 2024-06-26 at 12 07 46

I agree that showing a styled error message is much more helpful than the error message we currently show.

bahrmichael: I found https://github.com/sourcegraph/sourcegraph/pull/61938 which handles the RevisionNotFoundError and turns it into a "" commit, which then is turned in the error message we see above here:

https://github.com/sourcegraph/sourcegraph/blob/bc49218e6ca9a5d9ab867b9a688a06a2c14ff699/cmd/frontend/backend/inventory.go#L51-L53

If there is no revision, the code execution doesn't get to the point where you recommended handling the error.

Changing the error message will still be helpful.

bahrmichael:
@camdencheek @eseliger I went through your comments on the code, and think this is ready for another review :) Let me know what you think!

cmd/frontend/internal/inventory/inventory.go
```diff
@@ -45,42 +41,33 @@ type Lang struct {
...
-		attribute.String("FileName", file.Name()),
-		attribute.Bool("Safe", safe),
-		attribute.String("MatchedLang", matchedLang),
-	)
-
-	// No content
-	if rc == nil {
+	if skipEnhancedLanguageDetection {
 		lang.Name = matchedLang
 		lang.TotalBytes = uint64(file.Size())
```
camdencheek, line 59: This seems to be missing `lang.TotalLines` when enhanced language detection is disabled. Is that right?

cmd/frontend/internal/inventory/entries.go
```diff
@@ -108,6 +172,33 @@ func (c *Context) file(ctx context.Context, file fs.FileInfo) (inv Inventory, er
...
+
+	lang, err := getLang(ctx, file.FileInfo(), func(ctx context.Context, path string) (io.ReadCloser, error) {
+		return io.NopCloser(r), nil
+	}, c.ShouldSkipEnhancedLanguageDetection)
+	if err != nil {
+		return Inventory{}, errors.Wrapf(err, "inventory file %q", file.FileInfo().Name())
+	}
+	if lang == (Lang{}) {
+		return Inventory{}, nil
+	}
```
camdencheek, line 198: Could you create a followup Linear issue for dealing with "unknown" or "ambiguous" results? Out of scope for this PR, but we should probably improve our handling there.

cmd/frontend/internal/inventory/entries.go
camdencheek: 1000 seems fine to me 🙂

cmd/frontend/internal/inventory/entries.go
camdencheek: This is starting to sound like a leetcode problem 🙂

camdencheek (changes requested):

cmd/frontend/internal/inventory/entries.go
```diff
@@ -1,22 +1,86 @@
...
+	"io"
 	"io/fs"
 	"sort"
+	"strconv"
+
+	"github.com/sourcegraph/sourcegraph/internal/gitserver"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
 )
 
+var maxInvsLength, _ = strconv.Atoi(env.Get("GET_INVENTORY_MAX_INV_IN_MEMORY", "1000", "When computing the language stats, every nth iteration all loaded files are aggregated into the inventory to reduce the memory footprint. Increasing this value may make the computation run faster, but will require more memory."))
```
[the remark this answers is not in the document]
bahrmichael: Will do, thank you!

cmd/frontend/backend/repos.go
```diff
@@ -130,16 +133,11 @@ func (s *repos) GetInventory(ctx context.Context, repo api.RepoName, commitID ap
...
-	if err != nil {
-		return nil, err
-	}
-
 	// In computing the inventory, sub-tree inventories are cached based on the OID of the Git
 	// tree. Compared to per-blob caching, this creates many fewer cache entries, which means fewer
 	// stores, fewer lookups, and less cache storage overhead. Compared to per-commit caching, this
 	// yields a higher cache hit rate because most trees are unchanged across commits.
-	inv, err := invCtx.Entries(ctx, root)
+	inv, err := invCtx.All(ctx, s.gitserverClient, commitID)
```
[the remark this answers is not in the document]
bahrmichael: Yeah, the current approach looks not well rounded. I'm inclined towards Option 1.

cmd/frontend/backend/inventory.go
```diff
@@ -109,54 +103,55 @@ func InventoryContext(logger log.Logger, repo api.RepoName, gsClient gitserver.C
 				gitServerSemaphore.Release(1)
 			}}, nil
 		},
-		CacheGet: func(ctx context.Context, e fs.FileInfo) (inventory.Inventory, bool) {
-			cacheKey := cacheKey(e)
+		CacheKey: func(e fs.FileInfo) string {
+			info, ok := e.Sys().(gitdomain.ObjectInfo)
+			if !ok {
+				return "" // not cacheable
+			}
```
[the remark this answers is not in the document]
eseliger: Yea I'm not sure the tar entries would satisfy this check

eseliger (approved):
A few last comments and Camden should also review, but from a Source perspective, this seems fine

bahrmichael:
FYI: I'm currently focussing on the batch changes work with Bolaji, and will continue to work on this afterwards. Probably by the end of the week.

cmd/frontend/backend/inventory.go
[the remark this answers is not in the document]
bahrmichael: 👍 I stopped using this for the new approach.

cmd/frontend/internal/inventory/inventory.go
bahrmichael: Yes, we can't guess the number of lines without knowing the content.

cmd/frontend/internal/inventory/entries.go
bahrmichael: Looking at the `getLang` function, here are the non-error cases when we return `Lang{}`:

```
if file == nil {
		return Lang{}, nil
	}
```

```
if !file.Mode().IsRegular() || enry.IsVendor(file.Name()) {
		return Lang{}, nil
	}
```

```
if err == io.EOF {
			// No bytes read, indicating an empty file
			return Lang{}, nil
		}
```

None of them look like unknown or ambiguous results to me. (2) deliberately skips files that we don't consider worth checking the language for.

cmd/frontend/internal/inventory/entries.go
```diff
@@ -1,22 +1,86 @@
...
+		default:
+			// Skip symlinks, submodules, etc.
+			continue
+		}
+
+		if len(invs) > maxInvsLength {
+			sum := Sum(invs)
+			invs = make([]Inventory, maxInvsLength)
+			invs = append(invs, sum)
+		}
```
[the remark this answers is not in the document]
bahrmichael: I will keep that in mind! Thank you for the patient and detailed explanation :)

With the rewrite of the tar file processing, I opted to compute the Sum every time we complete an inventory. That's the point where we want to cache, and waiting for too many records doesn't seem worth it. Also the benchmarking for Sum shows that it's not an issue if we call it a bit more often:

```
	// n=1000: ~8500 ns/op
	// n=100: ~1100 ns/op
	// n=10: ~360 ns/op
	// n=5: ~300 ns/op
	// n=2: ~280 ns/op
```

cmd/frontend/internal/inventory/entries.go
bahrmichael: That makes sense to me. Updated this to return an empty inventory.

cmd/frontend/backend/repos.go
```diff
@@ -119,6 +118,10 @@ func (s *repos) GetInventory(ctx context.Context, repo api.RepoName, commitID ap
 		return Mocks.Repos.GetInventory(ctx, repo, commitID)
 	}
 
+	if commitID == "" {
+		return nil, errors.Errorf("Could not find a commit in the repository. Please add a file first.")
+	}
```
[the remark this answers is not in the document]
bahrmichael: I've updated this to return an error that the commitID must not be empty, and updated the four usages of this function to handle the empty commitID. It feels a bit awkward to do that handling in all the places when it could default to an empty Inventory here, but I'm happy to roll with it.

bahrmichael:
@eseliger @camdencheek I completed another pass, and the PR is ready for you to review again. The major change is that I reworked the inventory code to be simpler to test, then added some complicated caching logic, and with @camdencheek decided to drop that caching again. This means the code is now just a bit simpler to test, and doesn't cache except at the repo-level. Let me know what you think!

cmd/frontend/internal/inventory/context.go
```diff
@@ -16,9 +20,15 @@ type Context struct {
 	// NewFileReader is called to get an io.ReadCloser from the file at path.
 	NewFileReader func(ctx context.Context, path string) (io.ReadCloser, error)
 
+	CacheKey func(e fs.FileInfo) string
+
 	// CacheGet, if set, returns the cached inventory and true for the given tree, or false for a cache miss.
-	CacheGet func(context.Context, fs.FileInfo) (Inventory, bool)
+	CacheGet func(context.Context, string, api.CommitID) (Inventory, bool)
```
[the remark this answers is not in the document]
bahrmichael: I used `string` so that it would support the repoId (now with a commitID suffix) as well as the file name (via the CacheKey function) from the other code paths that still use the old approach.

cmd/frontend/graphqlbackend/search_results_stats_languages.go
```diff
@@ -157,7 +157,7 @@ func searchResultsStatsLanguages(ctx context.Context, logger log.Logger, db data
 			}
 
 			// Inventory all full-entry (files and trees) matches together.
-			inv, err := invCtx.Entries(ctx, work.fullEntries...)
+			inv, err := invCtx.Entries(ctx, key.commitID, work.fullEntries...)
```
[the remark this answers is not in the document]
bahrmichael: Reworked this.

cmd/frontend/backend/repos.go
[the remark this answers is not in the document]
bahrmichael: I updated it to align with Option 1.

cmd/frontend/internal/inventory/entries.go
camdencheek: `GetLanguageByFilename` can return an empty string for the language:

```
func main() {
	language, _ := enry.GetLanguageByFilename("hslgha.hsjahsg9")
	fmt.Printf("Language: %q\n", language)
}
```

bahrmichael: Right! I created SRCH-737

cmd/frontend/internal/inventory/entries.go
```diff
@@ -1,17 +1,111 @@
...
+			invs = invs[:0]
+			invs = append(invs, sum)
+		}
+
+		entry := n.Header.FileInfo()
+
+		switch {
+		case entry.Mode().IsRegular():
+			lang, err := getLang(ctx, entry, func(ctx context.Context, path string) (io.ReadCloser, error) {
+				return n.FileReader, nil
```
camdencheek, line 92: When is `FileReader` closed?

cmd/frontend/internal/inventory/entries.go
```diff
@@ -1,17 +1,111 @@
...
+// Based on the benchmarking in entries_test.go, 1_000 is a good number after which we see diminishing returns.
+var maxInvsLength = env.MustGetInt("GET_INVENTORY_MAX_INV_IN_MEMORY", 1_000, "When computing the language stats, every nth iteration all loaded files are aggregated into the inventory to reduce the memory footprint. Increasing this value may make the computation run faster, but will require more memory.")
+var getInventoryTimeout = env.MustGetInt("GET_INVENTORY_TIMEOUT", 5, "Time in minutes before cancelling getInventory requests. Raise this if your repositories are large and need a long time to process.")
+
+func (c *Context) All() (inv Inventory, err error) {
+	// Top-level caching is what we need for users, directory-level caching is to speed up retries if the repo doesn't compute within the timeout.
+	// To help with the latter we detach the context above, instead of introducing rather complicated caching logic (it turned out to be a leetcode problem).
+	// Since we're planning to move this to a pre-computed style anyway, the upsides of introducing complicated code
+	// for retryable caching become less relevant.
+	ctx := context.Background()
```
camdencheek, line 26: Creating a new `context.Background()` is usually not what we want. Instead, we should still pass in a context and use `context.WithoutCancel` to remove the cancellation signal. That way we still retain all the other stuff contexts carry, like tracing information, authentication information, etc.

camdencheek (changes requested):
Almost there! Just a couple of quick things before I think this is ready to go

cmd/frontend/internal/inventory/entries.go
bahrmichael: This happens in `inventory.go` right after the function to get the file reader is called.

```
	rc, err := getFileReader(ctx, file.Name())
	if err != nil {
		return Lang{}, errors.Wrap(err, "Failed to create a file reader.")
	}
	if rc != nil {
		defer rc.Close()
	}
```

I'm adding a unit test to make sure this behavior doesn't break.

cmd/frontend/internal/inventory/entries.go
bahrmichael: That's better indeed! I updated it in https://github.com/sourcegraph/sourcegraph/pull/62946/commits/719c13dc1e35d4d2c444fd542c1770a69e816d5b to use `WithoutCancel` and brought back the trace one level up. I think tracing in the for loop that iterates over files wouldn't bring much value compared to the overhead of creating traces.

cmd/frontend/internal/inventory/entries.go
bahrmichael: You can see the test here: https://github.com/sourcegraph/sourcegraph/pull/62946/commits/f8c5be275b9ededc64df577dadfa874628052fb3

cmd/frontend/backend/inventory.go
```diff
@@ -109,54 +104,57 @@ func InventoryContext(logger log.Logger, repo api.RepoName, gsClient gitserver.C
...
-
-	if !useEnhancedLanguageDetection && !forceEnhancedLanguageDetection {
-		// If USE_ENHANCED_LANGUAGE_DETECTION is disabled, do not read file contents to determine
-		// the language. This means we won't calculate the number of lines per language.
-		invCtx.NewFileReader = func(ctx context.Context, path string) (io.ReadCloser, error) {
-			return nil, nil
-		}
+		NewTarReader: func(reader io.ReadCloser) *tar.Reader {
+			return tar.NewReader(reader)
+		},
```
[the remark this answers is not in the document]
bahrmichael: Nope, we don't need that anymore. Dropped it with https://github.com/sourcegraph/sourcegraph/pull/62946/commits/9c254829b5b23fce60f10dc5ac8602d71a9fa654.

cmd/frontend/backend/inventory.go
```diff
@@ -109,54 +104,57 @@ func InventoryContext(logger log.Logger, repo api.RepoName, gsClient gitserver.C
...
+		ShouldSkipEnhancedLanguageDetection: !useEnhancedLanguageDetection && !forceEnhancedLanguageDetection,
+		GitServerClient:                     gsClient,
+		CommitID:                            commitID,
```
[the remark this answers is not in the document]
bahrmichael: I like that, and also moved the language detection upwards. It felt odd to have the GitServerClient as a field after all the other functions, so I moved that up, too.

Here's the commit for that: https://github.com/sourcegraph/sourcegraph/pull/62946/commits/d0228c7e814a326bcb955b3c10f3b9a5462db153

cmd/frontend/graphqlbackend/git_commit.go
```diff
@@ -348,6 +348,9 @@ func (r *GitCommitResolver) FileNames(ctx context.Context) ([]string, error) {
 }
 
 func (r *GitCommitResolver) Languages(ctx context.Context) ([]string, error) {
+	if r.oid == "" {
+		return []string{}, nil
+	}
```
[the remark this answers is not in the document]
bahrmichael: This came from the conversations https://github.com/sourcegraph/sourcegraph/pull/62946#discussion_r1661566403 and https://github.com/sourcegraph/sourcegraph/pull/62946#discussion_r1661567202.

I wasn't aware that an empty string could mean HEAD and will check the code again. Do you have any objection to making sure we explicitly map empty string to HEAD when that is acceptable, and reject requests otherwise? I would prefer to remove the implicit meaning here.

cmd/frontend/internal/inventory/entries.go
bahrmichael: As far as I can see, it's just the general io problems with leaks when you don't close readers.

In the `All` code we have an ArchiveReader (which now gets closed with https://github.com/sourcegraph/sourcegraph/pull/62946/commits/2d90cef9cacbc557b32c57d6e32da2ac8fc489b7) that gets handed to various `io.NopClosers`.  I think with the ArchiveReader's closing and the other calls to close, we should be safe.

cmd/frontend/graphqlbackend/git_commit.go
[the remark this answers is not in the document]
bahrmichael: I reverted the changes to the handling of non-absolute commits in https://github.com/sourcegraph/sourcegraph/pull/62946/commits/83e5ee30d3a0300d187a46d96fd3d56b924295fb.

It all started with trying to improve the error message, and that lead to a couple bad changes, that we can probably better leave untouched for now.

cmd/frontend/backend/repos.go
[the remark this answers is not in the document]
bahrmichael: I reverted my changes to the error-handling, since there were effects on other parts of the code that I'll better leave untouched for now. More context in https://github.com/sourcegraph/sourcegraph/pull/62946#discussion_r1679570588.

cmd/frontend/graphqlbackend/git_commit.go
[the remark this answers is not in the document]
camdencheek: Thanks!

camdencheek (approved):
Looks great -- thank you!

PR #62950, opened 2024-05-28
 Adding the User param to the site config so that it can be supported by Azure as an extra param

arafatkatze:
[See what the field Does
](https://learn.microsoft.com/en-us/answers/questions/1607830/how-does-the-user-parameter-work-in-azure-openai-s)

## Test plan
Tested locally by setting up extra variable in siteadmin with azure config and it worked perfectly.

### DIFF
--- a/internal/completions/client/azureopenai/openai.go
+++ b/internal/completions/client/azureopenai/openai.go
@@ -3,7 +3,6 @@ package azureopenai
 import (
 	"context"
 	"crypto/tls"
-	"fmt"
 	"io"
 	"net"
 	"net/http"
@@ -278,7 +277,6 @@ func streamChat(
 	// Azure sends incremental deltas for each message in a chat stream
 	// build up the full message content over multiple responses
 	var content string
-	var promptTokens int32
 
 	for {
 		entry, err := resp.ChatCompletionsStream.Read()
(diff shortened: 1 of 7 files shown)


internal/completions/client/azureopenai/openai.go
```diff
@@ -288,6 +291,11 @@ func streamChat(
 			}
 			return nil
 		}
+		// These are only included in the last message, so we're not worried about overwriting
+		if entry.Usage != nil {
+			promptTokens = *entry.Usage.PromptTokens
+			fmt.Println("Prompt Tokens: ", promptTokens)
```
jamesmcnamara, line 297: Is this meant to be left in?

arafatkatze: cleaned up my testing mistakes

schema/site.schema.json
```diff
@@ -2838,6 +2838,10 @@
           "description": "DEPRECATED. Use chatModel instead.",
           "type": "string"
         },
+        "user": {
+          "description": "Its the user field for OpenAI config for both AzureOpenAI and OpenAI",
```
dominiccooney, line 2842: Just "The ...", we don't need "Its ..."

(Its is possessive anyway and definitely wrong.)

dominiccooney (approved):
LGTM, I assume the ,omitempty is all we need to make this a no-op for models who do not use it and sites which do not specify it.

schema/site.schema.json
arafatkatze: Fixed

PR #62954, opened 2024-05-28
chore/msp-example: refactor to align with service structure best practices

bobheadxi:
## Test plan

Basic example builds and runs: `sg run msp-example`


cmd/msp-example/main.go
```diff
@@ -3,9 +3,9 @@ package main
 import (
 	"github.com/sourcegraph/sourcegraph/lib/managedservicesplatform/runtime"
 
-	"github.com/sourcegraph/sourcegraph/cmd/msp-example/internal/example"
+	"github.com/sourcegraph/sourcegraph/cmd/msp-example/service"
```
unknwon, line 6: I was kinda expecting this?

```suggestion
	"github.com/sourcegraph/sourcegraph/cmd/msp-example/internal/service"
```

unknwon (approved):

cmd/msp-example/main.go
bobheadxi: Yeah I thought about that but I think it's a top-level construct and I want to avoid nesting too much, perhaps. That way `internal/foobar` still has meaning and it's clear it's not on the same level as `internal/service`

unknwon: > That way `internal/foobar` still has meaning and it's clear it's not on the same level as `internal/service`

I think that's not a useful "clear" signal than "service" is internal to the MSP service, should not be used by any others 🤔

bobheadxi: The way I think about it, "service" is effectively to entrypoint to all the stuff in `internal/`, i.e.

Command -> Service -> Internals

So the exported Service isn't useful in practice, but in terms of hierarchy I think it has an important place as "not just another internal package"

unknwon: OK I think it just a bit wired in a monorepo setup, where `cmd/` is the top-level directory that includes everything.

Looking at on SAMS repo, for example.

`cmd/accounts-servver` imports `backend/service` seems making sense, and actually required because `cmd` can't import `backend/internal/service`.

bobheadxi: Yeah, for a standalone repo case it would be:

```
cmd/my-service
service/...
internal/...
```

unknwon: All good, let's go!

PR #62955, opened 2024-05-28
feature(plg): Use custom Stripe checkout

vdavid:
**Context:** In our previous Stripe checkout implementation, we didn't have the freedom to design our seat selection UI and other parts of our checkout experience.

**Problem:** The current state causes three difficulties:
1. We can't implement Rob's [nice seat selector UI](https://www.figma.com/design/FMSdn1oKccJRHQPgf7053o/Cody-PLG-GA?node-id=4274-3934&t=nkf1bvC4RsxAQTTa-0)
2. We need a custom solution for adding and removing seats after a subscription already exists. It's an unhelpful temporary side-track, plus we don't even have designs for this, and we need to ship soon.
3. We can't add an "Explore an enterprise plan" notification (tracked by https://github.com/sourcegraph/self-serve-cody/issues/784 and https://github.com/sourcegraph/self-serve-cody/issues/711)

**Solution:**

I found Stripe's "Custom Checkout with Elements" solution ([here](https://docs.stripe.com/checkout/custom-checkout/overview)), which seemed to solve all our problems without major changes or completely custom implementations of complex controls like the payment form or the address form. But then I realized that this solution was in a "Private Beta" state. I contacted Stripe, who quickly got back to me and gave us access to the API, assuring us that regardless of the Beta state, it's safe to use in production.

So I've modified the back end in https://github.com/sourcegraph/self-serve-cody/pull/790 to let us use the new UI and created this PR to add the front end.

- This PR, together with https://github.com/sourcegraph/self-serve-cody/pull/790 pretty much completes https://github.com/sourcegraph/self-serve-cody/issues/632
- It's also a major step toward https://github.com/sourcegraph/self-serve-cody/issues/725
- if I make sure to _not_ display an editable email field, this PR may close https://github.com/sourcegraph/self-serve-cody/issues/573

**Current caveats:**
- Final design missing
- Dark theme looks especially bad
- "Pay" button not working
- Lack of tests

## Changelog

- Added Stripe Custom Checkout

## Test plan

No thorough QA yet


chrsmith (commented):
I'll follow up with more specific questions in Slack, but I don't see what the actual benefit to this PR is.

Does the "Custom Stripe Checkout" just give us more flexibility in how we render the Stripe Checkout page? If so, then is the point of all of this just to restyle the "+/- seats" buttons when a new subscription is created? But we would still have the same limitations that we have today with Stripe Checkout, in that it can only be used for _creating new_ subscriptions, but we would still need to roll our own UX for updating existing subscriptions?

Ultimately, this doesn't seem like it is something we would want to merge.

vdavid:
@chrsmith 

> I'll follow up with more specific questions in Slack, but I don't see what the actual benefit to this PR is.
> 
> Does the "Custom Stripe Checkout" just give us more flexibility in how we render the Stripe Checkout page? If so, then is the point of all of this just to restyle the "+/- seats" buttons when a new subscription is created? But we would still have the same limitations that we have today with Stripe Checkout, in that it can only be used for creating new subscriptions, but we would still need to roll our own UX for updating existing subscriptions?
> 
> Ultimately, this doesn't seem like it is something we would want to merge.

We want to:
- Render the "+/- seats" UI the way we designed it, which we can't do with the `EmbeddedCheckoutProvider` as it doesn't allow for the level of customization we want.
- Show an "Upgrade to Enterprise" notification in case of 30+ seats, which we also can't do with `EmbeddedCheckoutProvider`.
- Use the same UI (with the credit card and address parts disabled) for the "Add seats" functionality. For this, we had to develop the custom UI we have on the left side of the page, so that's not extra work to use it here.
- Avoid having to recreate the payment method and address logic because we lack the time for it.

I've investigated the alternatives and haven't found a good one, so I currently intend to merge this PR once it's done. I'm open to hearing the alternatives I might've missed, though.

client/web/src/cody/management/api/stripeCheckout.ts
```diff
@@ -3,11 +3,14 @@ import type { BillingInterval } from './teamSubscriptions'
 export interface CreateCheckoutSessionRequest {
     interval: BillingInterval
     seats: number
+    adjustableSeatCount?: boolean
```
chrsmith, line 6: Just calling this out, in https://github.com/sourcegraph/self-serve-cody/pull/797 I suggested some alternative names for this field. If you agree with and make those changes, just be sure to update the code here, too before merging.

chrsmith (commented):
I didn't look super-close at the React code, but this all looks good to me. :shipit: 

> We want to:
> ...
> I've investigated the alternatives and haven't found a good one, so I currently intend to merge this PR once it's done. I'm open to hearing the alternatives I might've missed, though.

All of the things you listed are great, and I am super-duper excited to have a better looking UI for the new subscription experience!

The "alternative" I'm referring to is just that, once we have all of the UI components working the way we want. (And aren't relying on Stripe Checkout to power the specific HTML/DOM.) Then we should be able to cut Stripe Checkout out of the equation entirely.

And just call our own API to create the new Stripe Customer / Subscription object. Since if we have the user-supplied seat count, name and billing address, and payment method, we have all the things we need. And being able to remove Stripe Checkout and rely on _just_ our code for the entire experience, seems like it would simplify our overall codebase. (So perhaps not in this PR, but maybe we are within striking distance of doing that in 1-2 follow up PRs.)

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
```diff
@@ -1,75 +1,175 @@
...
-            returnUrl: `${origin}/cody/manage?session_id={CHECKOUT_SESSION_ID}&welcome=1`,
+}> = ({ creatingTeam, customerEmail }) => {
+    const [updatingSeatCount, setUpdatingSeatCount] = React.useState(false)
+
+    const navigate = useNavigate()
+
+    const { total, lineItems, updateLineItemQuantity, email, updateEmail, status } = useCustomCheckout()
+
+    const [errorMessage, setErrorMessage] = React.useState<string | null>(null)
+    const [displayErrorMessage, setDisplayErrorMessage] = React.useState(false)
```
taras-yemets, line 42: Is there a case when we don't show a non-null error message?

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
```diff
@@ -1,75 +1,175 @@
...
+        },
+        [setErrorMessage, setDisplayErrorMessage]
+    )
+
+    const debouncedUpdateSeatCount = useDebouncedCallback(async newSeatCount => {
+        if (lineItems.length === 1) {
+            await updateLineItemQuantity({
+                lineItem: lineItems[0].id,
+                quantity: newSeatCount,
+            })
```
taras-yemets, line 58: Can it throw an error here?

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
vdavid: Yes, when there is a missing payment method or billing address. That is, right away when the page loaded, we displayed an error message, which felt wrong. Now we're only displaying the error message after clicking the "Subscribe" button if the data is still missing.

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
```diff
@@ -1,75 +1,175 @@
...
+}> = ({ creatingTeam, customerEmail }) => {
+    const [updatingSeatCount, setUpdatingSeatCount] = React.useState(false)
+
+    const navigate = useNavigate()
+
+    const { total, lineItems, updateLineItemQuantity, email, updateEmail, status } = useCustomCheckout()
+
+    const [errorMessage, setErrorMessage] = React.useState<string | null>(null)
+    const [displayErrorMessage, setDisplayErrorMessage] = React.useState(false)
+    const [seatCount, setSeatCount] = React.useState(lineItems[0]?.quantity)
```
taras-yemets, line 43: I think we can simplify the state management a bit.
```
const [updatingSeatCount, setUpdatingSeatCount] = React.useState(false)
const [seatCount, setSeatCount] = React.useState(lineItems[0]?.quantity)
const [debouncedSeatCount] = useDebounce(seatCount, 800);

useEffect(() => {
  setUpdatingSeatCount(true)
  await updateLineItemQuantity({
    lineItem: lineItems[0].id,
    quantity: newSeatCount,
  })
  // TODO: catch error if needed
  setUpdatingSeatCount(false)
}, [lineItems[0].id, debouncedSeatCount])

// whether to show loading icon on price label
const isLoading = seatCount !== debouncedSeatCount || updatingSeatCount

// rest logic

return (
  <>
    {/* existing content... */}
    <Button disabled={seatCount === MIN_SEAT_COUNT} onClick={() => setSeatCount(c => c > MIN_SEAT_COUNT ? c - 1 : c)}>
      <Icon aria-hidden={true} svgPath={mdiMinusThick} />
    </Button>
    <div className={styles.seatCountSelectorValue}>{seatCount}</div>
    <Button disabled={seatCount === MAX_SEAT_COUNT} onClick={() => setSeatCount(c => c < MAX_SEAT_COUNT ? c + 1 : c)}>
      <Icon aria-hidden={true} svgPath={mdiPlusThick} />
    </Button>
                            
  </>
)
```

With this approach we:
- handle updating the debounced value (component’s state) and updating Stripe for input (third-party hook state) separately.
- eliminate two `useCallback` functions.
- ensure that invalid seat counts are not set, as in the original implementation, and disable the button to notify the user when they can’t increase or decrease further.

This is not a blocking comment, and I am fine with the current approach as long as it works. I just wanted to mention this alternative.

taras-yemets (approved):
LGTM. Thank you!
Left a few comments inline.

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
taras-yemets: >Yes, when there is a missing payment method or billing address. 

`PayButton` component can set a payment method or billing address error and it receives a `setAndDisplayErrorMessage` (sets both `errorMessage` and `displayErrorMessage` to truth values)  callback as a prop. The only case I spotted where `errorMessage` is set separately (without changing `displayErrorMessage`) is when the session is expired - is it possible to get this on the page load?

My point here is to use a single variable for the error state if possible.

client/web/src/cody/management/subscription/new/NewCodyProSubscriptionPage.tsx
```diff
@@ -42,6 +41,13 @@ const AuthenticatedNewCodyProSubscriptionPage: FunctionComponent<NewCodyProSubsc
     authenticatedUser,
     telemetryRecorder,
 }) => {
+    const [urlSearchParams] = useSearchParams()
+    const initialSeatCount = useMemo(() => {
+        const defaultSeatCount = urlSearchParams.get('team') === '1' ? 2 : 1
+        const seatCountString = urlSearchParams.get('seats') || defaultSeatCount.toString()
+        return parseInt(seatCountString, 10) || defaultSeatCount
+    }, [urlSearchParams])
```
taras-yemets, line 49: A comment about the expected `team` (what does `1` mean and what are other possible values) and `seats` URL search params values would be helpful. 

We can also be more explicit when handling `NaN` for `parseInt` (it's not clear at glance why do we have two `|| defaultSeatCount`), e.g., return early in this case:
```
const initialSeatCount = useMemo(() => {
  const seatCountFromURL = parseInt(urlSearchParams.get('seats') || "", 10)
  if (Number.isNaN(seatCountFromURL)) {
    // default seat count
    return urlSearchParams.get('team') === '1' ? 2 : 1
  }
  return seatCountFromURL
}, [urlSearchParams])
```

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
vdavid: I'm not sure but [this try-catch](https://github.com/sourcegraph/sourcegraph/pull/62955/commits/723a72ad2e9bfe22c1ea3ee40653d133fde2f712) won't hurt.

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
vdavid: I've switched to this approach in https://github.com/sourcegraph/sourcegraph/pull/62955/commits/4037774b35dcc4b2435e2a4398de26a3efcc2aca

It sounded like a good idea at first but then once implemented, I think it's on par with the original solution, and slightly more code lines. But then, if it seems clearer for you, it might be clearer overall, so I kept it.

client/web/src/cody/management/subscription/new/NewCodyProSubscriptionPage.tsx
vdavid: Added 👍 
https://github.com/sourcegraph/sourcegraph/pull/62955/commits/3d7184a711c5cb0fa70ebe8b25efffcb019a25d3

client/web/src/cody/management/subscription/new/CodyProCheckoutForm.tsx
vdavid: Thanks! I think at some point of the development, the second variable was necessary, but then in the final state it wasn't. Thanks, I've removed it in https://github.com/sourcegraph/sourcegraph/pull/62955/commits/929ac6c5dccf570484a100bf7d99d988a4212e2c

PR #62956, opened 2024-05-28
[DO NOT Merge] Adds Endpoint URL to retrieve token keys

arafatkatze:
## Test plan
Tested locally by setting up a URL to retreive tokens from and then making repeated calls to it.


internal/completions/client/azureopenai/openai.go
```diff
@@ -52,9 +57,86 @@ type CompletionsClient interface {
...
+	if err != nil {
+		return "", err
+	}
+
+	return tokenResponse.AccessToken, nil
+}
+
+func startTokenRefresher(endpoint, tokenRetrievalEndpoint, clientID, clientSecret string) {
+	startTokenRefresherOnce.Do(func() {
+		ticker := time.NewTicker(1 * time.Minute)
```
arafatkatze, line 104: Will change this to 20 mins to update a few times before needed.

The actual requirement is to update token every hour.

internal/completions/client/azureopenai/openai.go
```diff
@@ -52,9 +56,85 @@ type CompletionsClient interface {
...
+		}()
+	})
+}
+
+func refreshAccessToken(endpoint, tokenRetrievalEndpoint, clientID, clientSecret string) {
+	apiClient.mu.Lock()
+	defer apiClient.mu.Unlock()
+
+	accessToken, err := getAccessToken(tokenRetrievalEndpoint, clientID, clientSecret)
+	if err != nil {
```
dominiccooney, line 118: Let's log more noisily here. If it is not working you will want to know why.

internal/completions/client/azureopenai/openai_test.go
```diff
@@ -44,8 +44,8 @@ func (c *mockAzureClient) GetChatCompletionsStream(ctx context.Context, body azo
 	return c.getChatCompletionsStream(ctx, body, options)
 }
 
-func getNewMockAzureAPIClient(mock *mockAzureClient) func(accessToken, endpoint string) (CompletionsClient, error) {
-	return func(accessToken, endpoint string) (CompletionsClient, error) {
+func getNewMockAzureAPIClient(mock *mockAzureClient) func(accessToken, endpoint, tokenRetreivalEndpoint, clientId, clientSecret string) (CompletionsClient, error) {
+	return func(accessToken, endpoint, tokenRetreivalEndpoint, clientId, clientSecret string) (CompletionsClient, error) {
```
dominiccooney, line 48: Spelling, retrieval

Other occurrences of this in this PR so check everywhere

schema/site.schema.json
```diff
@@ -2842,6 +2842,18 @@
           "description": "The user field for OpenAI config for both AzureOpenAI and OpenAI",
           "type": "string"
         },
+        "tokenRetreivalEndpoint": {
+          "description": "Its the azure field for the endpoint to retreive token from",
```
dominiccooney, line 2846: Spelling again, this one customer visible 

Feedback again about no "Its the" just start with "The"

dominiccooney (approved):
Approving to unblock

Some nits inline

GetClient taking so many strings in a row makes me nervous because it would be so easy to mix them up at callsites and see no obvious error

I am not familiar with sync.Once but I assume it will handle not repeatedly starting the periodic fetch when multiple calls to GetClient arrive before the first 1/20/60 minutes

Note the time period needs to be updated to something longer, presumably

internal/completions/client/azureopenai/openai_test.go
arafatkatze: Apologies for the spelling mistakes I fixed all of them

arafatkatze:
> GetClient taking so many strings in a row makes me nervous because it would be so easy to mix them up at callsites and see no obvious error

Yep, I am aware of it and it got me confused too so I tested it a few times to make sure but I hear you and I can see how errors at callsite might make this hard to debug.

> I am not familiar with sync.Once but I assume it will handle not repeatedly starting the periodic fetch when multiple calls to GetClient arrive before the first 1/20/60 minutes

That was the intent 

> Note the time period needs to be updated to something longer, presumably

I will do that right before making the image.

internal/completions/client/azureopenai/openai.go
arafatkatze: @dominiccooney I added logging and tested it and that works fine. 

Thanks for the recommendation

kalanchan:
@arafatkatze nit: we should aim to have better descriptive branch names, it will help with PR reviews at first glance

https://dev.to/shinjithdev/git-good-best-practices-for-branch-naming-and-commit-messages-oj4

internal/completions/client/azureopenai/openai.go
```diff
@@ -52,9 +56,87 @@ type CompletionsClient interface {
...
+	var tokenResponse TokenResponse
+	err = json.Unmarshal(body, &tokenResponse)
+	if err != nil {
+		return "", err
+	}
+
+	return tokenResponse.AccessToken, nil
+}
+
+func startTokenRefresher(endpoint, tokenRetrievalEndpoint, clientID, clientSecret string, logger log.Logger) {
```
Strum355, line 101: What deployment ultimately runs this? If its frontend, be aware that deployments may have >1 frontend instance running, which would mean this is being run >x1

internal/completions/client/azureopenai/openai.go
```diff
@@ -52,9 +56,87 @@ type CompletionsClient interface {
...
+	startTokenRefresherOnce.Do(func() {
+		ticker := time.NewTicker(1 * time.Minute)
+		go func() {
```
Strum355, line 104: Might be better to use existing patterns for timers like this, see: https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/internal/goroutine/periodic.go?L116:6-116:26
@vdavid may be able to talk more about this, he wrote part of the UI that gives us great insight into background routines

internal/completions/client/azureopenai/openai.go
arafatkatze: We have confirmed its a single instance.

internal/completions/client/azureopenai/openai.go
arafatkatze: @Strum355 Its always better to use a consistent pattern and I would definitely want to do that for the long term but right now there is a clock ticking for the trial deploy and I would really want to get this out soon. 

In my testing for this I tried failure modes such the server being down etc and those worked fine with this setup. I would definitely look deeper into the suggested solution when we get the proper fix for multiple cases out.

arafatkatze:
@kalanchan Thanks for the pointers and I will try to be mindful of this moving forward.

internal/completions/client/azureopenai/openai.go
vdavid: @arafatkatze, makes sense, but happy to chat if you need an intro to the more standard solution. That said, it's also been a while for me since I was in that code, but can definitely help sit down and figure it out together.

arafatkatze: @vdavid Thanks a ton for the offer, we made an image release from this branch. Actually I will reach out to you again once their team confirms that this solution actually works for them as we made out a new image release for them. 

We wanna do this in the right way for the long term(And also for multiple stakeholders instead of just this once) so I will ask you at the right time.

PR #62961, opened 2024-05-29
feat(search): Add inline file search to Svelte app

fkling:
This PR adds the inline file search to the file view.

## Changelog

- Refactored the existing CodeMirror extension to allow UI customization
- Moved utility functions around to avoid loading all of wildcard into the prototype (which lead to build errors)
- Tweaked the new icon component to better align within buttons (I missed that in #62908 because I didn't test it within buttons)
- Noticed that button group styles didn't apply correctly and the reason seemed to be that we were mixing sveltekit and react styles. So I copied the button styles into the sveltekit app.
- Added a switch component, following https://web.dev/articles/building/a-switch-component .

## Test plan

Manually tested the SvelteKit and React versions of the search panel.


client/web-sveltekit/src/lib/wildcard/Button.ts
fkling, line 1: This was copied from `@sourcegraph/wildcard` to use the right generated CSS classes.

client/web-sveltekit/src/lib/wildcard/Switch.svelte
fkling, line 1: The new switch component.

client/web-sveltekit/src/routes/svelte-overrides.scss
```diff
@@ -22,6 +22,10 @@ body {
     }
 }
 
+label {
+    margin-bottom: 0;
+}
```
fkling, line 27: Don't know why we set a margin by default, but it's rather annoying.

client/web/src/repo/blob/codemirror/search.ts
fkling, line 1: The actual UI was moved into a separate file/component.

client/web/src/repo/blob/codemirror/search/SearchPanel.tsx
fkling, line 1: UI that was previously in `search.tsx`.

client/web/src/util/dom.ts
```diff
@@ -99,15 +99,3 @@ export function useBreakpoint(size: keyof typeof breakpoints, debounceMs = 50):
...
-    tagName: K,
-    properties: Partial<HTMLElementTagNameMap[K]> | null = null,
-    ...children: (Node | string)[]
-): HTMLElementTagNameMap[K] {
-    const element = Object.assign(document.createElement(tagName), properties)
-    for (const child of children) {
-        element.append(typeof child === 'string' ? document.createTextNode(child) : child)
-    }
-    return element
-}
```
fkling, line 113: This was moved to remove the transitive dependency on wildcard which causes build issues.

camdencheek:
I can't seem to run this locally. Do I need to do anything special to generate the icons?

fkling:
@camdencheek 

> I can't seem to run this locally. Do I need to do anything special to generate the icons?

`pnpm install` and `pnpm dev` should work. What did you try?

camdencheek:
> What did you try?

I tried that 😄

client/web-sveltekit/src/lib/wildcard/Button.ts
fkling: The imported `getButtonClassName` function would used the CSS module defined in `@sourcegraph/wildcard`. But the `ButtonGroup.svelte` component actually uses the `Button.module.scss` version in `web-sveltekit`. It seems that the class names are generated differently somehow and bringing in `getButtonClassName` to use the same CSS module as `ButtonGroup.svelte` fixed the problems I was seeing.

Specifically:

| Before | After |
|--------|--------|
|  |  | 

We use the button group for the forward/backward buttons.

client/web-sveltekit/src/lib/Icon2.module.scss
```diff
@@ -7,9 +7,18 @@ svg.icon {
 
     color: var(--icon-fill-color, var(--color, inherit));
     fill: currentColor;
+    stroke-width: var(--stroke-width, 2);
+
+    g,
+    path {
+        // Overwrites Lucide's/unplugin-icons' default stroke-width setting
+        stroke-width: inherit;
+    }
```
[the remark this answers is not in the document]
fkling: It seemed quite heavy for smaller icons. Lucide's own API actually allows you make stroke widths absolute or relative to the size: https://lucide.dev/guide/basics/stroke-width.

We'll have to do another pass over icons anyway to tweak sizes and stroke width. Related issue: https://linear.app/sourcegraph/issue/SRCH-464/revamp-icon-sizes

fkling:
Which error/problem do you get? You can also message me in Slack and I'll take a look when I have time.

client/web-sveltekit/src/lib/wildcard/Button.ts
camdencheek: Ah, gotcha. Thanks!

client/web-sveltekit/src/lib/Icon2.module.scss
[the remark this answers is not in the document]
camdencheek: Got it! cool, will save further comments for that issue

camdencheek:
Not sure what changed, but it's fixed now. 

Seems this broke the layout of search results headers:

screenshot-2024-05-29_10-10-40@2x

camdencheek:
Should the keybind be scoped more tightly? Seems a little weird that `Cmd+F` opens the panel when, e.g., I have the top-level search input focused

camdencheek (approved):

camdencheek:
nit: I don't love the lucide icons for case sensitivity and regex 😕 They seem imbalanced (specifically, the case sensitive icon looks too small), and the regex icon doesn't really look like a regex icon to me.

screenshot-2024-05-29_21-14-55@2x

taiyab:
> (Also, should the text input be taller than the buttons?)

No. They should all be a standard size (probably resolving to 32px height in this case).

PR #62964, opened 2024-05-29
feat(appliance): deploy worker

craigfurman:
Closes https://linear.app/sourcegraph/issue/REL-83/service-definition-worker

## Test plan

Golden tests included.

## Changelog

- Deploys the Worker service.

### DIFF
--- a/internal/appliance/reconciler/worker.go
+++ b/internal/appliance/reconciler/worker.go
@@ -2,7 +2,6 @@ package reconciler
 
 import (
 	"context"
-	"fmt"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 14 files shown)


internal/appliance/reconciler/worker.go
```diff
@@ -0,0 +1,153 @@
...
+
+	ctr.Env = append(ctr.Env, container.EnvVarsRedis()...)
+	ctr.Env = addPreciseCodeIntelBlobstoreVars(ctr.Env, sg)
+	if !sg.Spec.Embeddings.Disabled && !sg.Spec.Blobstore.Disabled {
+		ctr.Env = append(
+			ctr.Env,
+			corev1.EnvVar{Name: "EMBEDDINGS_UPLOAD_BACKEND", Value: "blobstore"},
+			corev1.EnvVar{Name: "EMBEDDINGS_UPLOAD_AWS_ENDPOINT", Value: "http://blobstore:9000"},
+		)
+	}
```
craigfurman, line 68: Note that this logic is a little different to what we do in the helm chart: https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/76581d9e2d4c6cfe4e92b2b58c8f5dac87a97b07/charts/sourcegraph/templates/worker/worker.Deployment.yaml#L63-L70

AFAICT, we want to set these vars only when we are using the bundled blobstore - otherwise that hostname makes no sense! Users of external blobstores will have to configure these vars using the containerConfig (see config subpackage, and standard_config_test.go).

internal/appliance/reconciler/precise_code_intel.go
```diff
@@ -144,3 +136,16 @@ func (r *Reconciler) reconcilePreciseCodeIntelServiceAccount(ctx context.Context
 	sa := serviceaccount.NewServiceAccount("precise-code-intel-worker", sg.Namespace, cfg)
 	return reconcileObject(ctx, r, cfg, &sa, &corev1.ServiceAccount{}, sg, owner)
 }
+
+func addPreciseCodeIntelBlobstoreVars(env []corev1.EnvVar, sg *config.Sourcegraph) []corev1.EnvVar {
```
DaedalusG, line 140: This isn't a huge conceptual change -- here is the `*config.Sourcegraph` type mapped to the configMap the customer is using similar to a `values.yaml` file in helm? Looks like the API 👀 - _chat more later_

internal/appliance/reconciler/worker.go
DaedalusG: Agree with the logic here. Although I'm too familiar with the external blobstore stuff.

DaedalusG (approved):
Looks good to me, very similar to how we were doing things with CRDs

internal/appliance/reconciler/worker.go
craigfurman: There's actually an issue to rip out embeddings 😅 https://linear.app/sourcegraph/issue/REL-125/remove-references-to-embeddings

Let's just merge this for now and defer making sure we really do want to remove embeddings for that issue

internal/appliance/reconciler/precise_code_intel.go
craigfurman: Sorry, I'm not quite understanding this comment. This is a refactor in this file, extracting a function to add some env vars that are useful in 2 places, and calling it from the new worker.go. We can see that it's a refactor because the precise-code-intel golden fixtures haven't changed and the tests are still passing.

> is the *config.Sourcegraph type mapped to the configMap the customer is using similar to a values.yaml file in helm?

Yep! 👍 

> Looks like the API 👀

I'm not 100% sure what you mean, but indeed the ConfigMap schema is the API for deploying sourcegraph here. The appliance frontend will craft that configmap in response to user input, but users can ignore the web app and craft a configmap directly if they like 🙂

PR #62965, opened 2024-05-29
feat(codeintel): If SignatureDocumentation is present in SCIP, use it for hover text

keynmol:
Fixes GRAPH-610

1. We prepend a markdown-formatted signature to the hover text if SignatureDocumentation field if present
2. If the field is absent, we fall back to old behaviour of assuming that signature documentation is already present in the hover text

SignatureDocumentation was added to SCIP format a while back but it's not yet being utilised by our indexers. One of the first to do so will be scip-java (https://github.com/sourcegraph/scip-java/pull/707) – with the plan for rollout being:

1. Merge this PR and wait for a Sourcegraph release
2. Merge scip-java PR and tag a release
3. (optionally) update the default scip-java image in SG

This change allows handling both new SCIP indexes (those that emit SignatureDocumentation), and the old ones, which have signature prepended to Documentation field.

## Test plan

- New tests for the signature and documentation

## Changelog

- The backend now respects the SCIP [signature_documentation](https://github.com/sourcegraph/scip/blob/main/scip.proto#L409-L415) field, and if available, uses that to create the initial part of the hover text. SCIP indexers should not repeat the text in signature_documentation inside the documentation field.

### DIFF
--- a/internal/codeintel/codenav/internal/lsifstore/locations_by_position.go
+++ b/internal/codeintel/codenav/internal/lsifstore/locations_by_position.go
@@ -283,7 +283,7 @@ func extractOccurrenceData(document *scip.Document, occurrence *scip.Occurrence)
 	// we should include in reference and implementation searches.
 
 	if symbol := scip.FindSymbol(document, occurrence.Symbol); symbol != nil {
-		hoverText = symbol.Documentation
+		hoverText = symbolHoverText(symbol)
 
 		for _, rel := range symbol.Relationships {
 			if rel.IsDefinition {
@@ -366,8 +366,15 @@ func monikersToString(vs []precise.MonikerData) string {
 	return strings.Join(strs, ", ")
 }
 
-//
-//
+func symbolHoverText(symbol *scip.SymbolInformation) []string {
+	if symbol.SignatureDocumentation != nil {
+		signature := []string{fmt.Sprintf("```%s\n%s\n```", symbol.SignatureDocumentation.Language, symbol.SignatureDocumentation.Text)}
+		return append(signature, symbol.Documentation...)
+	} else {
+		return symbol.Documentation
+	}
+
+}
 
 func (s *store) ExtractDefinitionLocationsFromPosition(ctx context.Context, locationKey LocationKey) (_ []shared.Location, _ []string, err error) {
 	return s.extractLocationsFromPosition(ctx, extractDefinitionRanges, symbolExtractDefault, s.operations.getDefinitionLocations, locationKey)
--- a/internal/codeintel/codenav/internal/lsifstore/locations_by_position.go
+++ b/internal/codeintel/codenav/internal/lsifstore/locations_by_position.go
@@ -367,18 +367,11 @@ func monikersToString(vs []precise.MonikerData) string {
 }
 
 func symbolHoverText(symbol *scip.SymbolInformation) []string {
-	if symbol.SignatureDocumentation != nil {
-		signature := symbol.SignatureDocumentation
-		if signature.Text != "" && signature.Language != "" {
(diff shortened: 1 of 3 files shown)


internal/codeintel/codenav/internal/lsifstore/locations_by_position.go
```diff
@@ -366,8 +366,15 @@ func monikersToString(vs []precise.MonikerData) string {
 	return strings.Join(strs, ", ")
 }
 
-//
-//
+func symbolHoverText(symbol *scip.SymbolInformation) []string {
+	if symbol.SignatureDocumentation != nil {
+		signature := []string{fmt.Sprintf("```%s\n%s\n```", symbol.SignatureDocumentation.Language, symbol.SignatureDocumentation.Text)}
```
keynmol, line 371: I wonder if we should be defensive here about Language and Text being possibly nil?
Not sure if the required fields are enforced at serialisation time by the SCIP proto bindings

internal/codeintel/codenav/internal/lsifstore/metadata_by_position.go
```diff
@@ -94,7 +94,7 @@ func (s *store) GetHover(ctx context.Context, bundleID int, path string, line, c
 				}
 
 				// Return first match
-				return strings.Join(symbol.Documentation, "\n"), rangeBySymbol[symbolName], true, nil
+				return strings.Join(symbolHoverText(symbol), "\n"), rangeBySymbol[symbolName], true, nil
```
keynmol, line 97: I took a look at testing this particular logic, but unfortunately it seems to rely on the hardcoded SQL datadump: https://github.com/sourcegraph/sourcegraph/blob/main/internal/codeintel/codenav/internal/lsifstore/testdata/code-intel-extensions%407802976b.sql

So adding tests here is problematic as I'd need to either modify existing indexes in that dump or add a new one, both manually.

With existing tests passing, we're at least testing that fallback logic works fine for old indexes. To test this in a better fashion we'd need to modify our testing infrastructure, which is IMO long overdue but not fit for this particular PR.

internal/codeintel/codenav/internal/lsifstore/locations_by_position.go
varungandhi-src: They can't be `nil`, the fields have type `string`, so they will be default initialized to `""` if they're absent. But yes, we should be defensive.

internal/codeintel/codenav/internal/lsifstore/locations_by_position_test.go
```diff
@@ -556,6 +556,69 @@ func TestExtractOccurrenceData(t *testing.T) {
 		}
 	})
 
+	t.Run("documentation and signature documentation", func(t *testing.T) {
+		testCases := []struct {
+			explanation string
+			document    *scip.Document
+			occurrence  *scip.Occurrence
+			hoverText   []string
+		}{
```
varungandhi-src, line 565: Can you use autogold here? It's a bit annoying to have to edit test cases by hand when making changes.

varungandhi-src (approved):

internal/codeintel/codenav/internal/lsifstore/locations_by_position.go
keynmol: Sorry can you clarify – should we be defensive against empty text/language, given it's indistinguishable from missing fields? If so, does this make sense? https://github.com/sourcegraph/sourcegraph/pull/62965/commits/68340f7236cbb0ad6d6982cb12f76bd38869093a

varungandhi-src: Yes, I meant the patch that you made. I compressed it a bit. https://github.com/sourcegraph/sourcegraph/pull/62965/commits/6f9d5c9dc3b9205d98c0e5eede2aa8549a833547

PR #62966, opened 2024-05-29
feat/enterprise-portal: explicit opt-in to serve non-dev/non-internal data

bobheadxi:
Since we are reading from the dotcom database for the time being, the dev and prod Enterprise Portal instances serve the same data. This change adds a configuration so that we can set up the dev instance to only serve data from licenses with the dev or internal tag, toggled by `DOTCOM_INCLUDE_PRODUCTION_LICENSES`

This will become a non-issue for Enterprise Portal when we have our own database in https://linear.app/sourcegraph/issue/CORE-100, but for now this helps make sure that the dev instance only serves dev data. This is fairly important since we grant Enterprise Portal scopes more liberally in accounts.sgdev.org.

## Test plan

Updated integration tests to use the dev-only mode.

Also tested locally: with `DOTCOM_INCLUDE_PRODUCTION_LICENSES=false`, endpoints return nothing, while with `DOTCOM_INCLUDE_PRODUCTION_LICENSES=true`, endpoints return the expected data.

### DIFF
--- a/cmd/enterprise-portal/internal/dotcomdb/dotcomdb.go
+++ b/cmd/enterprise-portal/internal/dotcomdb/dotcomdb.go
@@ -207,7 +207,8 @@ FROM product_subscriptions subscription
 		clauses = append(clauses, "HAVING "+conds.havingClause)
 	}
 	if opts.DevOnly {
-		c := fmt.Sprintf("('%s' = ANY(MAX(active_license.license_tags)) OR '%s' = ANY(MAX(active_license.license_tags)))",
+		// '&&' operator: overlap (have elements in common)
+		c := fmt.Sprintf("ARRAY['%s','%s'] && MAX(active_license.license_tags)",
 			licensing.DevTag, licensing.InternalTag)
 		if conds.havingClause != "" {
 			clauses = append(clauses, "AND "+c)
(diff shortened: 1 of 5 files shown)


cmd/enterprise-portal/internal/dotcomdb/dotcomdb.go
```diff
@@ -199,6 +206,15 @@ FROM product_subscriptions subscription
 	if conds.havingClause != "" {
 		clauses = append(clauses, "HAVING "+conds.havingClause)
 	}
+	if opts.DevOnly {
+		c := fmt.Sprintf("('%s' = ANY(MAX(active_license.license_tags)) OR '%s' = ANY(MAX(active_license.license_tags)))",
+			licensing.DevTag, licensing.InternalTag)
```
unknwon, line 211: Wondering if you could use `@>` for "array contains (another array)" https://www.postgresql.org/docs/9.1/functions-array.html

unknwon (approved):

cmd/enterprise-portal/internal/dotcomdb/dotcomdb.go
bobheadxi: TIL! I think I want `&&` "overlap (have elements in common)" as either `dev` of `internal` should match

PR #62970, opened 2024-05-29
feat(sg): add version=auto for `sg release cut`

Chickensoupwithrice:
Previously, the version number had to be specified explicitly during the `sg release cut` command. Now we can optionally pass `--version=auto` to query release-registry for the next version. 
We need this in both places, since on main cut releases, this is the first step that happens, but during patch releases, the first step is `sg release create` which also has `--version=auto`.

## Test plan

Tested manually

## Changelog

- Add `--version=auto` to `sg release cut`

### DIFF
--- a/dev/sg/internal/release/cut.go
+++ b/dev/sg/internal/release/cut.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/Masterminds/semver"
 	"github.com/urfave/cli/v2"
+	"strings"
 
 	"github.com/sourcegraph/sourcegraph/dev/sg/internal/execute"
 	"github.com/sourcegraph/sourcegraph/dev/sg/internal/repo"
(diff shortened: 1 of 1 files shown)


BolajiOlajide (approved):
This looks good to me.
Thank you @Chickensoupwithrice .

~However, it seems like we have this logic duplicated now. It might be helpful to have this in a function and call it in the "sg release cal" and "sg release create" command which accepts the "--version=auto" flag.~

We don't have to do this now, I just thought to share here.

BolajiOlajide:
Ignore me

Chickensoupwithrice:
It seems like it might still be a good idea to have a function to parse version across all the different subcommands

Chickensoupwithrice:
Testing: 
image

PR #62974, opened 2024-05-29
Svelte: QoL fixes for hover popovers

camdencheek:
Fixes SRCH-449

Each commit is granular and has a useful message describing the change.

In short:
- Adds a `hoverDelay` parameter to `Popover` so each consumer doesn't need to do this manually
- Makes hovers only show when the mouse comes to rest
- Closes the popover when the window loses focus so we don't miss a mouseleave event and have popovers get stuck open
- Adds a delay to popover closing when the mouse leaves one of the target elements so there is room for error when moving the mouse from the trigger to the popover
- Adds preloading back to the file tree because the popover now does not create the element until the delay has passed
- Makes it so clicking the target closes the popover when hover is enabled.

## Test plan

Manual testing for feel. Would love if reviewers would pull this down and try it out on the file tree.

[Video walkthrough](https://share.cleanshot.com/HNpLjjv7)

### DIFF
--- a/client/web-sveltekit/src/lib/Popover.svelte
+++ b/client/web-sveltekit/src/lib/Popover.svelte
@@ -49,9 +49,17 @@
             }
         }
 
+        function handleMouseMoveTrigger(): void {
+            clearTimeout(delayTimer)
+            delayTimer = setTimeout(() => {
+                isOpen = true
+            }, hoverDelay)
+        }
+
         if (showOnHover) {
             node.addEventListener('mouseenter', handleMouseEnterTrigger)
             node.addEventListener('mouseleave', handleMouseLeaveTrigger)
+            node.addEventListener('mousemove', handleMouseMoveTrigger)
         }
 
         return {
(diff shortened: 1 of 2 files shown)


client/web-sveltekit/src/lib/Popover.svelte
```diff
@@ -68,6 +74,7 @@
                 node.removeEventListener('mouseenter', handleMouseEnterTrigger)
                 node.removeEventListener('mouseleave', handleMouseLeaveTrigger)
                 node.removeEventListener('mousemove', handleMouseMoveTrigger)
+                window.removeEventListener('blur', handleWindowLoseFocus)
```
camdencheek, line 74: There is one more situation that can cause this. If I re-tile my window so the window moves behind my mouse, a mouseleave event is not triggered and the popover does not disappear. The only way I could think of to get around this would be to add an event on `mousemove` and check if the mouse ever appears outside of the popover, but that sounded expensive

vovakulikov (approved):

client/web-sveltekit/src/lib/Popover.svelte
```diff
@@ -33,26 +36,49 @@
...
+
+        function handleWindowLoseFocus(): void {
+            clearTimeout(delayTimer)
+            toggle(false)
+        }
+
+        function handleClick(): void {
+            clearTimeout(delayTimer)
+            toggle(false)
+        }
```
fkling, line 64: I'd have a single function, maybe `close`, and use that in both cases.

PR #62976, opened 2024-05-29
Context: detect symbol names in query

jtibshirani:
This PR updates Cody context search to pull symbol names out of the query and
perform an additional symbol search. The symbol search results are deduplicated
with the code search results, so we always prefer the symbol definition over
another file match.

Benefits:
* When a query references a symbol, we always include its definition in context
(unless it is quite ambiguous). This is a basic thing users would expect, and
improves our "Find symbol" evals.
* We're more likely to return high quality chunks, since symbol results are
always anchored on the line with the definition.

Addresses SPLF-88

## Test plan

Adapted unit tests. Tested locally.


jtibshirani:
Eval results on golden queries:

Before
```
Breakdown by class:
Find symbol	7/10
Find string	2/2
Explain file	2/2
Explain concept	4/5
Check dependency	1/2
Find logic	31/43
Gather information	12/16
Changelog	0/2
Ownership	2/2
How-to	1/1
Foreign language	0/2
Long request	0/2

Combined recall	62/89
```

After
```
Breakdown by class:
Find symbol	9/10
Find string	2/2
Explain file	2/2
Explain concept	4/5
Check dependency	1/2
Find logic	31/43
Gather information	12/16
Changelog	0/2
Ownership	2/2
How-to	1/1
Foreign language	0/2
Long request	0/2

Combined recall	64/89
```

internal/search/codycontext/job.go
```diff
@@ -129,23 +154,48 @@ func (j *searchJob) Run(ctx context.Context, clients job.RuntimeClients, stream
...
 		if r.err != nil {
 			err = errors.Append(err, r.err)
 		}
 	}
-
 	return alert, err
 }
 
+// limitCodeResponse deduplicates code results that are already included in the symbol results,
+// and applies the result limit. Symbol results count towards the total code results limit.
```
keegancsmith, line 178: why would symbol results be included in the code results? Don't they have a different match.key?

internal/search/codycontext/query_transformer.go
```diff
@@ -96,50 +162,32 @@ func transformPatterns(patterns []string) []string {
...
-	}
-
-	if len(patternNodes) > 0 {
-		nodes = append(nodes, query.NewOperator(patternNodes, query.Or)...)
+// findSymbols extracts patterns that look like symbols. It uses the following heuristics:
+//   - If the pattern contains an underscore or camelCase, it's probably a symbol
+//   - If the pattern contains a namespace marker, each component is probably a symbol
+//   - If the pattern contains a dot, split but still check if the components look like symbols.
+//     This avoids overmatching in cases where the dot just represents a period.
+func findSymbols(patterns []string) []string {
```
keegancsmith, line 170: nice

internal/search/codycontext/query_transformer.go
```diff
@@ -96,50 +162,32 @@ func transformPatterns(patterns []string) []string {
...
+	var symbols []string
+	for _, pattern := range patterns {
+		if strings.Contains(pattern, "::") {
```
keegancsmith, line 173: Can you add a case for `->`?

keegancsmith (approved):
very nice.

One general question based on me not fully understanding stuff. Do we rely at all at NumContextLines to ensure we give enough context in the chunks? If so, do our symbol results respect that?

internal/search/codycontext/query_transformer.go
jtibshirani: Good point!

internal/search/codycontext/job.go
jtibshirani: They are both considered "file matches", so all that matters from the match key is repo, commit, and file. I added a test case exercising this deduplicating logic to make sure it works.

PR #62978, opened 2024-05-30
Svelte: simplify and grow last commit

camdencheek:
This makes various tweaks and simplifications to the `LastCommit` component shown on the file page. 

- Uses the `Timestamp` component
- Removes a bunch of unnecessary DOM elements and styles in favor of just using flexbox well
- Removes the max width on the last commit component, allowing it to take up as much space as is available (but leaving a healthy gap from the tabs). This was the primary reason I tackled this PR. Most commits I see are truncated to the "barely useful" level

## Test plan

https://github.com/sourcegraph/sourcegraph/assets/12631702/94ba7ef3-e4fd-444b-a300-08cdd50982c8

### DIFF
--- a/client/web-sveltekit/src/lib/Avatar.svelte
+++ b/client/web-sveltekit/src/lib/Avatar.svelte
@@ -63,6 +63,7 @@
         justify-content: center;
         position: relative;
         background: var(--secondary);
+        flex: none;
     }
 
     div::after {
--- a/client/web-sveltekit/src/lib/repo/LastCommit.svelte
+++ b/client/web-sveltekit/src/lib/repo/LastCommit.svelte
@@ -1,7 +1,6 @@
 <script lang="ts">
-    import { formatDistanceToNow } from 'date-fns'
-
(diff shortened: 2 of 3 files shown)


client/web-sveltekit/src/lib/Avatar.svelte
```diff
@@ -63,6 +63,7 @@
         justify-content: center;
         position: relative;
         background: var(--secondary);
+        flex: none;
```
camdencheek, line 66: Avatars, like icons, should never shrink

fkling (approved):
This has been bothering me as well. Great simplification too!

taiyab:
Thanks for this @camdencheek.

Once the new symbol tree goes in, we can consider removing the references tab from that bottom bar, and converting that entire bottom part to be a simple 4/1 split with the last commit + open all history action on the left (and left aligned), and ownership info on the right.

Having a long and right-aligned commit message like that doesn't really work, but we'll have to deal with it for now.

PR #62981, opened 2024-05-30
Svelte: add lines to directory groups

camdencheek:
This adds guidelines for open directories in the file tree. I recommend reviewing with whitespace changes hidden.

Fixes SRCH-443

Dark:
screenshot-2024-05-29_22-50-29@2x

Light:
screenshot-2024-05-29_22-50-54@2x

## Test plan

Manual testing. This is one where a screenshot test would be really nice 🙂

### DIFF
--- a/client/web-sveltekit/src/lib/TreeNode.svelte
+++ b/client/web-sveltekit/src/lib/TreeNode.svelte
@@ -112,15 +112,13 @@
                 <LoadingSpinner center={false} />
             </div>
         {:then treeProvider}
-            <div class="group-container">
-                <ul role="group">
-                    {#each treeProvider.getEntries() as entry (treeProvider.getNodeID(entry))}
-                        <svelte:self {entry} {treeProvider} let:entry let:toggle let:expanded on:scope-change>
-                            <slot {entry} {toggle} {expanded} />
-                        </svelte:self>
-                    {/each}
-                </ul>
-            </div>
+            <ul role="group">
+                {#each treeProvider.getEntries() as entry (treeProvider.getNodeID(entry))}
+                    <svelte:self {entry} {treeProvider} let:entry let:toggle let:expanded on:scope-change>
+                        <slot {entry} {toggle} {expanded} />
+                    </svelte:self>
+                {/each}
+            </ul>
         {:catch error}
             <slot name="error" {error} />
         {/await}
@@ -193,7 +191,7 @@
         display: flex;
     }
 
-    .group-container {
+    ul {
         position: relative;
         isolation: isolate;
         &::before {
--- a/client/web-sveltekit/src/lib/TreeNode.svelte
+++ b/client/web-sveltekit/src/lib/TreeNode.svelte
@@ -7,8 +7,8 @@
 </script>
 
 <script lang="ts" generics="T">
-    import { createEventDispatcher, getContext, setContext } from 'svelte'
     import { mdiChevronDown, mdiChevronRight, mdiImageFilterCenterFocusStrong } from '@mdi/js'
+    import { createEventDispatcher, getContext, setContext } from 'svelte'
 
     import Icon from '$lib/Icon.svelte'
     import { Button } from '$lib/wildcard'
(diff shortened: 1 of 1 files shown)


client/web-sveltekit/src/lib/TreeNode.svelte
```diff
@@ -81,27 +81,29 @@
...
         <!-- hide the open/close button to preserve alignment with expandable entries -->
-        <span class="expandable-icon-container" class:hidden={!expandable}>
-            <span class="scope-container">
-                <Button variant="icon" on:click={handleScopeChange}>
-                    <Icon svgPath={mdiImageFilterCenterFocusStrong} inline />
-                </Button>
-            </span>
-
-            <!-- We have to stop even propagation because the tree root listens for click events for
+        {#if expandable}
```
camdencheek, line 84: Slight change in behavior: instead of reserving space for the chevron icon so the file icon lines up with the folder icon, this lines up the file icon with the chevron of a sibling directory. This was necessary for the lines to make sense because we want the guideline to terminate at the file icon, not at an empty space in front of the file icon. This is visible in the screenshots.

client/web-sveltekit/src/lib/TreeNode.svelte
```diff
@@ -189,14 +193,21 @@
...
+            position: absolute;
+            content: '';
+            border-left: 1px solid var(--border-color);
+            height: 100%;
+            transform: translateX(
+                calc(
+                    var(--tree-node-nested-level) * 1.25rem + var(--tree-node-left-padding) + var(--icon-inline-size) /
+                        2 - 1px
+                )
+            );
```
camdencheek, line 209: Kinda a gnarly calculation, but it's using the same logic used to offset the labels within the list item.

client/web-sveltekit/src/lib/TreeNode.svelte
```diff
@@ -189,14 +191,21 @@
...
     }
 
-    .hidden {
-        visibility: hidden;
-
-        // If we're in the flat list we can omit expand icon rendering
-        // since none of items is expandable, hence there aren't any offsets
-        :global([data-tree-view-flat-list='true']) & {
-            width: 0;
-            margin-left: 0.5rem;
```
vovakulikov, line 199: This logic previously was giving additional left spacing for the flat list case 
See my comment with screenshots above. I think we should add a special value
for the flat list selector
```scss
[role="treeitem"] {
   /* ... styles .... */
   
      :global([data-tree-view-flat-list='true']) & {
            --tree-node-left-padding: 2.75rem;
       }
}
```

vovakulikov (commented):
Look good, I think there is only one small regression for the flat list items paddings
Compare 

| Before | After |
| ------- | ------- |
| Screenshot 2024-05-30 at 11 28 52 | Screenshot 2024-05-30 at 11 28 43 |

I left a comment on how this might be fixed below.

client/web-sveltekit/src/lib/TreeNode.svelte
camdencheek: Ah, I see now! I was struggling to figure out how `data-tree-view-flat-list` was actulaly triggered. Will fixup 👍

camdencheek: Okay, fixed it. Now that we're not specially indenting files (compared to dirs), the calculation actually got quite a bit simpler. I think there are some followup simplifications we could make to remove the absolute positioning and padding calculations.

fkling: Looks like we don't use `data-tree-view-flat-list` anywhere else, so we can remove the logic that sets it in `TreeView`.

fkling (approved):

client/web-sveltekit/src/lib/TreeNode.svelte
camdencheek: Yes! Planning a followup PR with simplifications. I started doing it in this PR, but it got pretty noisy

PR #62982, opened 2024-05-30
sg/msp: fix nil `domain` and `EnvironmentDomainTypeNone` in diagram gen

jac:
Not specifying a `domain` config or using `EnvironmentDomainTypeNone` was causing a nil panic during diagram generation. This PR fixes that

Closes CORE-140
## Test plan

nil Domain or EnvironmentDomainTypeNone generates a diagram like so:
image

### DIFF
--- a/dev/managedservicesplatform/operationdocs/diagram/diagram.go
+++ b/dev/managedservicesplatform/operationdocs/diagram/diagram.go
@@ -199,7 +199,7 @@ func (d *diagram) Generate(s *spec.Spec, e string) error {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


bobheadxi:
Without a domain, do we actually provision an IP address and ALB stack? IIRC we don't, and only provide the Cloud Run URL

jac:
> Without a domain, do we actually provision an IP address and ALB stack? IIRC we don't, and only provide the Cloud Run URL

good catch

bobheadxi (approved):
Thank you!

PR #62983, opened 2024-05-30
Remove external accounts modal feature flag

### LINKED ISSUE #62805: Remove feature flag for external accounts modal

pjlast:
Closes #62805 

#61721 introduced a modal reminding users to connect external accounts. It was put behind a feature flag in case of malfunction. But since then, it seems to work as expected based on customer feedback.

This PR removes the feature flag, and turns the feature on by default, since we think it's an improvement to the user experience over all.

## Test plan

Feature flag removed and verified to work by default.


eseliger (approved):
I assume there was no additional feedback from the customer who trialed this initially and this is working and good to ship to all?

eseliger:
Makes sense, we can ship that separately.

PR #62991, opened 2024-05-30
Svelte: make search header primary

camdencheek:
This implements the change to move the search bar back into the header, and the repo header down a level. Comments inline.

screenshot-2024-05-30_20-27-06@2x

Completes SRCH-456

## Test plan

Updated relevant playwright tests.
[Video overview](https://share.cleanshot.com/RVK1cVv5)


client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte
```diff
@@ -78,133 +76,125 @@
...
+            visibleNavEntryCount = Math.max(visibleNavEntryCount - 1, 0)
+        },
+    }}
+>
+    <button on:click={() => goto(data.repoURL)}>
+        <CodeHostIcon repository={repoName} codeHost={resolvedRevision?.repo?.externalRepository?.serviceType} />
+        <h1>{displayRepoName}</h1>
+    </button>
+
+    <TabsHeader id="repoheader" {tabs} selected={selectedTab} on:select={event => goto(tabs[event.detail].href)} />
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte:121]: One downside of using `TabsHeader` as it exists is that the buttons aren't links, which means we don't get preloading. So, while they are visually the same, I'm kinda wondering whether I should actually be using that component here.

client/web-sveltekit/src/lib/Popover.svelte
```diff
@@ -136,6 +136,7 @@
         background-color: var(--dropdown-bg);
         color: var(--body-color);
         box-shadow: var(--popover-shadow);
+        z-index: 1;
```
camdencheek, line 139: This is to avoid conflicting with the z-index of the search input.

client/web-sveltekit/src/lib/Tabs.svelte
```diff
@@ -68,5 +68,9 @@
         display: flex;
         flex-direction: column;
         height: 100%;
+
+        header {
+            border-bottom: 1px solid var(--border-color);
+        }
```
camdencheek, line 74: Moved the bottom border to the `Tabs` component so the `TabsHeader` component can be used separately.

client/web-sveltekit/src/lib/dom.ts
```diff
@@ -410,6 +411,55 @@ export const overflow: Action<HTMLElement, { class: string; measureClass?: strin
...
 }
 
+/**
+ * An action that resizes an element with the provided grow and shrink callbacks until the target element no longer overflows.
+ *
+ * @param grow A callback to increase the size of the contained contents
+ * @param shrink A callback to reduce the size of the contained contents
+ * @returns An action that updates the overflow state of the element.
+ */
+export const sizeToFit: Action<HTMLElement, { grow: () => void; shrink: () => void }> = (target, { grow, shrink }) => {
```
camdencheek, line 421: I added this function because I couldn't get the existing ones to do what I wanted. `overflow` isn't enough because I need more than a "small" and "large" state, and `computeFit` doesn't work in this case because the list of elements is not statically known. I think this is strictly more general than both of those with the tradeoff that changes need to be committed to the DOM with `tick` before we can re-measure, which I _think_ means there is a possibility of flashing. I was unable to observe any flashing though, so maybe there is another layer of batching happening somewhere. 

Anyways, this is probably the piece that could use the closest review

client/web-sveltekit/src/lib/dom.ts
```diff
@@ -410,6 +411,55 @@ export const overflow: Action<HTMLElement, { class: string; measureClass?: strin
...
+                if (target.scrollWidth == lastScrollWidth && target.clientWidth == lastClientWidth) {
+                    break
+                }
+                lastScrollWidth = target.scrollWidth
+                lastClientWidth = target.clientWidth
+            }
+            if (target.scrollWidth > target.clientWidth) {
+                shrink()
+                await tick()
+            }
```
fkling, line 450: This doesn't seem to work as intended. See the video:


https://github.com/sourcegraph/sourcegraph/assets/179026/31abe2be-f884-4570-9312-4780752580ea

I think what happens here is that this logic will simply add one more entry if there is space, it doesn't add `n` entries as long as there is space.
It's not a problem when you slowly resize the window because `update` will be called multiple times.
The sudden size change could be considered to be an edge case but I assume there a couple of ways to manage windows that can cause such size changes.

client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte
```diff
@@ -78,133 +76,120 @@
...
             }
         }
 
-        :global([data-icon]) {
-            --color: var(--icon-color);
+        :global([data-dropdown-trigger]) {
+            height: 100%;
+            align-self: stretch;
+            padding: 0.5rem;
+            --icon-fill-color: var(--text-muted);
```
Comment (fkling) [inline client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte:192]: We could add `margin-left: auto` here to make the button right align. Not sure if we want that.

client/web-sveltekit/src/lib/search/CodeHostIcon.svelte
```diff
@@ -1,13 +1,14 @@
...
     import { getIconPathForCodeHost } from '$lib/repo/shared/codehost'
+    import Tooltip from '$lib/Tooltip.svelte'
 
     export let repository: string
+    export let codeHost: string | undefined = undefined
     export let disableTooltip: boolean = false
 
     $: hostName = repository.split('/')[0]
-    $: svgPath = getIconPathForCodeHost(hostName)
+    $: svgPath = getIconPathForCodeHost(codeHost ?? hostName)
```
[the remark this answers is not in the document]
camdencheek: I tried that initially, but we also use the repo name for the tooltip. We should be making `codeHost` required, but that would mean we need to go find everywhere it's used and make sure we have a codehost available. Notably, we `don't` easily have that available on search results right now (though I should probably fix that).

client/web-sveltekit/src/lib/dom.ts
camdencheek: "statically known" was a lazy reason. In reality, it was mostly an issue with `TabsHeader` having its own wrapper, so we would need to add functionality to `TabsHeader` to apply the action. I got this mostly working, but it would not work more generally, such as if we have two different elements that can shrink (like in the file header where both the file name and the actions will be shrinkable). So it's a bit forward-looking as well as not wanting to expand the `TabsHeader` interface.

client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte
Comment (camdencheek) [reply client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte:192]: Was just following the designs, which do not have the button on the right. No personal preference

client/web-sveltekit/src/lib/dom.ts
camdencheek: Good catch. I pushed a fix for this. In short: I was relying on the scroll/client width to change on every shrink/grow event, but that's not necessarily true depending on styling (which is why the measureClass exists). Instead, I made shrink and grow return a boolean indicating whether or not we can keep shrinking/growing

client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte
```diff
@@ -78,133 +76,120 @@
...
-                            <span>{entry.label}</span>
-                        </span>
-                    </MenuLink>
-                {/if}
-            {/each}
-        </DropdownMenu>
-        <RepoSearchInput repoName={data.repoName} revision={data.displayRevision} />
-    </nav>
+    <div class="search-header">
+        <SearchInput {queryState} size="compat" onSubmit={handleSearchSubmit} />
```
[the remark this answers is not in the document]
Comment (camdencheek) [reply client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte:101]: Yep 😢 It also exists on the search page, so I didn't think it was appropriate to fix in this PR

client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte
```diff
@@ -78,133 +76,120 @@
...
+        },
+        shrink() {
+            visibleNavEntryCount = Math.max(visibleNavEntryCount - 1, 0)
+        },
+    }}
+>
+    <button on:click={() => goto(data.repoURL)}>
+        <CodeHostIcon repository={repoName} codeHost={resolvedRevision?.repo?.externalRepository?.serviceType} />
+        <h1>{displayRepoName}</h1>
+    </button>
```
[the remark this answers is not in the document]
Comment (camdencheek) [reply client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte:121]: Oh, because it's just a placeholder for the repo actions menu (SRCH-22). I was planning on knocking that out next, but I guess there's no reason for this to be a button right now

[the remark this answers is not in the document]
Comment (camdencheek) [reply client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte]: Pushed a change to make it an `<a>` instead

client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte
Comment (camdencheek) [reply client/web-sveltekit/src/routes/[...repo=reporev]/+layout.svelte]: Done 👍 It means there is a tiny bit of duplication, but I think it's probably best that tabs can be links.

camdencheek:
Fixed this up to incorporate @vovakulikov's recent changes, and this should be ready for review again

fkling (approved):

client/web-sveltekit/src/lib/TabsHeader.svelte
```diff
@@ -28,22 +36,48 @@
...
+                    <Icon svgPath={tab.icon} aria-hidden inline />
+                {/if}
+                <span data-tab-title={tab.title}>
+                    {tab.title}
+                </span>
+                {#if tab.shortcut}
+                    <KeyboardShortcut shorcut={tab.shortcut} />
+                {/if}
+            </button>
+        {/if}
```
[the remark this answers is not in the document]
camdencheek: aha! I didn't know that existed. Or if I did, I forgot 😄 

Anyways, much cleaner. Thanks!

PR #62993, opened 2024-05-31
Cody PLG: add support for Google Gemini models

abeatrix:
This change adds support for the Google Gemini 1.5 Flash and Pro models in the ModelBadges component and the isAllowedCustomChatModel function. This allows these models to be displayed and used in the Sourcegraph product subscriptions UI and chat functionality.

## Changelog

In your sg local instant, look for the gemini models in the license page.

Verify you are not seeing the following error when running the curl command listed in https://sourcegraph.slack.com/archives/C05AGQYD528/p1717105626504719?thread_ts=1717011825.198899&cid=C05AGQYD528: `{"error":"model \"google/gemini-1.5-flash-latest\" is not allowed, allowed: []"}`


olafurpg (approved):
LGTM 👍🏻

I'm not 100% familiar wit this part of the code. My understanding: this adds Gemini Pro/Flash to the same list of models where we already declare GPT-4o support.

PR #62994, opened 2024-05-31
feat/msp: use pgxpool instead of pgx.Conn

bobheadxi:
During manual testing for https://github.com/sourcegraph/sourcegraph/pull/62934, I started realizing that I would run into database errors:

```
failed to deallocate cached statement(s): conn busy
```

Turns out, `pgx.Conn` is meant for non-concurrent use. What we really want is a connection pool, with an off-the-shelf offering from `pgxpool`.

## Test plan

Integration tests pass, now with more cases using `t.Parallel()`. Also ran a quick sanity check by hand:

```
sg start
```

```
for i in {1..10}; do curl --header "Content-Type: application/json" --header 'authorization: bearer $SAMS_TOKEN' --data '{"filters":[{"filter":{"is_archived":false}}]}' \
    http://localhost:6081/enterpriseportal.subscriptions.v1.SubscriptionsService/ListEnterpriseSubscriptionLicenses & ; done
```

## Changelog

- The MSP runtime `lib/managedservicesplatform/contract.Contract`'s `ConnectToDatabase(...)` has been renamed to `GetConnectionPool(...)`, and now returns a `*pgxpool.Pool` instead of a `*pgx.Conn`
- The MSP runtime `lib/managedservicesplatform/cloudsql` helper library's `Connect(...)` has been renamed to `GetConnectionPool(...)`, and now returns a `*pgxpool.Pool` instead of a `*pgx.Conn`

### DIFF
--- a/cmd/enterprise-portal/internal/dotcomdb/BUILD.bazel
+++ b/cmd/enterprise-portal/internal/dotcomdb/BUILD.bazel
@@ -14,6 +14,7 @@ go_library(
         "//lib/enterpriseportal/subscriptions/v1:subscriptions",
         "//lib/errors",
         "@com_github_jackc_pgx_v5//:pgx",
+        "@com_github_jackc_pgx_v5//pgxpool",
     ],
 )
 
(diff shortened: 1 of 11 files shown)


unknwon (approved):
FYI asks for a [stdlib interface from pgx](https://pkg.go.dev/github.com/jackc/pgx/stdlib) would give you the pool (because of the stdlib semantics).

bobheadxi:
This is true, but at least release registry has expressed interest in using pgx directly, and as we're a postgres-only shop I wonder if it might be better to use pgx directly

PR #63000, opened 2024-05-31
Cody Gateway embeddings: powering with generated metadata

janhartman:
Connected to https://github.com/sourcegraph/bfg-private/pull/189 and https://github.com/sourcegraph/cody/pull/4414.

We're introducing a hacky MVP to enable embeddings being powered by metadata that's generated from code. This PR is the bare minimum to make this work on CG. We plan to trigger metadata generation only if we're using a new (fake) model (this comes in via a feature flag) and if the request isn't a real-time query, but is a background indexing request.
The implementation is really hacky, but is also really minimal.

## Test plan
Testing locally through a feature flag.


cmd/cody-gateway/internal/httpapi/embeddings/metadata.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/sourcegraph/sourcegraph/internal/completions/client/fireworks"
+	"github.com/sourcegraph/sourcegraph/internal/completions/types"
+)
+
+type metadataClient struct {
+	logger            log.Logger
+	completionsClient types.CompletionsClient
+}
+
+func GenerateMetadata(req codygateway.EmbeddingsRequest, logger log.Logger, completionsClient types.CompletionsClient) ([]string, error) {
```
rafax, line 18: Do you need this to be public (uppercase)?

cmd/cody-gateway/internal/httpapi/embeddings/metadata.go
```diff
@@ -0,0 +1,52 @@
...
+func (c *metadataClient) generateMetadataForChunk(input *string) (string, error) {
+	promptText := "Here is a section of code. " +
+		"Please write a paragraph of documentation for each high-level class, struct, function or similar. " +
+		"Be concise, write no more than a few sentences for each entry. " +
+		"Return your response in text format. Each entry name should be followed by a newline, then its documentation. " +
+		"Respond with nothing else, only the entry names and the documentation. Code: ```" +
+		*input +
+		"```"
+
+	resp, err := c.completionsClient.Complete(context.TODO(), types.CompletionsFeatureChat, types.CompletionsVersionLegacy, types.CompletionRequestParameters{
```
rafax, line 36: Seeing that we're making 3rd party LLM requests here, it'd be great to correctly propagate context (to save LLM calls and associated $$$ / LLM quota)

cmd/cody-gateway/internal/httpapi/embeddings/metadata.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func (c *metadataClient) generateMetadataForChunk(input *string) (string, error) {
+	promptText := "Here is a section of code. " +
+		"Please write a paragraph of documentation for each high-level class, struct, function or similar. " +
+		"Be concise, write no more than a few sentences for each entry. " +
+		"Return your response in text format. Each entry name should be followed by a newline, then its documentation. " +
+		"Respond with nothing else, only the entry names and the documentation. Code: ```" +
+		*input +
+		"```"
```
rafax, line 34: Nit: use multiline string?

cmd/cody-gateway/internal/httpapi/embeddings/metadata.go
```diff
@@ -0,0 +1,52 @@
...
+	completionsClient types.CompletionsClient
+}
+
+func GenerateMetadata(req codygateway.EmbeddingsRequest, logger log.Logger, completionsClient types.CompletionsClient) ([]string, error) {
+	c := metadataClient{
+		logger:            logger.Scoped("metadata_gen"),
+		completionsClient: completionsClient,
+	}
+
+	return iter.MapErr(req.Input, c.generateMetadataForChunk)
```
rafax, line 24: By default, this will make `runtime.GOMAXPROCS` concurrent calls to Fireworks - should we limit concurrency to something more reasonable (data on batch size https://redash.sgdev.org/queries/729#1486)? Do you know if Fireworks has concurrent requests rate-limits?

cmd/cody-gateway/internal/httpapi/embeddings/metadata.go
janhartman: No, leftover from when I thought I'd be creating a separate package. Fixed

cmd/cody-gateway/internal/httpapi/embeddings/metadata.go
janhartman: Wow, I literally left the TODO context as a reminder to myself and then forgot. fixed

cmd/cody-gateway/internal/httpapi/embeddings/metadata.go
janhartman: Good point. I've limited it to 15, that seems reasonable based on your graph. 
I'm not sure about Fireworks rate limits.

janhartman:
@rafax One more point I thought about: should we make the fireworks model we're calling configurable? I'd have to run an eval, but maybe if llama 8B can get us good results as well, the whole thing would be a lot faster.

rafax:
> @rafax One more point I thought about: should we make the fireworks model we're calling configurable? I'd have to run an eval, but maybe if llama 8B can get us good results as well, the whole thing would be a lot faster.

Deploying Cody Gateway is fast and easy, so up to you - for an experimental feature like this, I think it's simpler to change this here and redeploy than propagate the config variable through all the layers (and need to remove all that if we change our mind / change the architecture).

cmd/cody-gateway/internal/httpapi/embeddings/handler.go
```diff
@@ -126,6 +128,25 @@ func NewHandler(
 				}
 			}()
 
+			// Hacky experiment: Replace embedding model input with generated metadata text when indexing.
```
rafax, line 131: Link to a Linear issue / any artifact that provides context on the experiment?

rafax (approved):

cmd/cody-gateway/internal/httpapi/embeddings/handler.go
janhartman: https://linear.app/sourcegraph/issue/AI-40/llm-generated-metadata-eg-documentation-for-code

PR #63003, opened 2024-05-31
PLG: extract Cody Pro routes to a separate file

taras-yemets:
## Test plan
- CI
- Tested manually: migrated routes load corresponding pages

### DIFF
--- a/client/web/src/cody/CodyProPage.tsx
+++ b/client/web/src/cody/CodyProPage.tsx
@@ -1,0 +1,35 @@
+import { lazyComponent } from '@sourcegraph/shared/src/util/lazyComponent'
+
+import type { LegacyLayoutRouteContext } from '../LegacyRouteContext'
+
+import { CodyProRoutes } from './codyProRoutes'
+
+const routeComponents = {
+    [CodyProRoutes.NewProSubscription]: lazyComponent(
+        () => import('./management/subscription/new/NewCodyProSubscriptionPage'),
+        'NewCodyProSubscriptionPage'
+    ),
+    [CodyProRoutes.Manage]: lazyComponent(() => import('./management/CodyManagementPage'), 'CodyManagementPage'),
+    [CodyProRoutes.Subscription]: lazyComponent(
+        () => import('./subscription/CodySubscriptionPage'),
+        'CodySubscriptionPage'
+    ),
+    [CodyProRoutes.SubscriptionManage]: lazyComponent(
+        () => import('./management/subscription/manage/CodySubscriptionManagePage'),
+        'CodySubscriptionManagePage'
+    ),
+    [CodyProRoutes.ManageTeam]: lazyComponent(() => import('./team/CodyManageTeamPage'), 'CodyManageTeamPage'),
+    [CodyProRoutes.AcceptInvite]: lazyComponent(() => import('./invites/AcceptInvitePage'), 'CodyAcceptInvitePage'),
+}
+
+interface CodyProPageProps extends Pick<LegacyLayoutRouteContext, 'authenticatedUser' | 'telemetryRecorder'> {
+    path: CodyProRoutes
+}
+
+/**
+ * Renders the appropriate Cody Pro page component for the given route path.
+ */
+export const CodyProPage: React.FC<CodyProPageProps> = props => {
+    const Component = routeComponents[props.path]
+    return <Component authenticatedUser={props.authenticatedUser} telemetryRecorder={props.telemetryRecorder} />
+}
--- a/client/web/src/cody/codyProRoutes.tsx
+++ b/client/web/src/cody/codyProRoutes.tsx
@@ -1,0 +1,48 @@
+import type { RouteObject } from 'react-router-dom'
+
+import { lazyComponent } from '@sourcegraph/shared/src/util/lazyComponent'
+
+import { LegacyRoute } from '../LegacyRouteContext'
+
+import { isEmbeddedCodyProUIEnabled } from './util'
+
+export enum CodyProRoutes {
+    // The checkout form for a new Cody Pro subscription.
+    NewProSubscription = '/cody/manage/subscription/new',
+    // The Manage page is labeled as the "Dashboard" page.
+    Manage = '/cody/manage',
+    // The Subscriptions page is a comparison of different Cody product tiers.
+    Subscription = '/cody/subscription',
+    SubscriptionManage = '/cody/subscription/manage',
+
+    ManageTeam = '/cody/team/manage',
+    // Accepts an invite to join a Cody team, then redirects to the Cody team page.
+    AcceptInvite = '/cody/invites/accept',
+}
+
+/**
+ * Defines the routes to be rendered in the Cody Pro UI.
+ * If the embedded Cody Pro UI is enabled, all available routes are rendered.
+ * Otherwise, only the Manage and Subscription routes are rendered.
+ */
+const routes = isEmbeddedCodyProUIEnabled()
+    ? Object.values(CodyProRoutes)
+    : [CodyProRoutes.Manage, CodyProRoutes.Subscription]
+
+const CodyProPage = lazyComponent(() => import('./CodyProPage'), 'CodyProPage')
+
+export const codyProRoutes: RouteObject[] = routes.map(path => ({
+    path,
+    element: (
+        <LegacyRoute
+            render={props => (
+                <CodyProPage
+                    path={path}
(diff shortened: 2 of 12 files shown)


client/web/src/cody/codyProRoutes.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <LegacyRoute
+            render={props => (
+                <CodyProPage
+                    path={path}
+                    authenticatedUser={props.authenticatedUser}
+                    telemetryRecorder={props.platformContext.telemetryRecorder}
+                />
+            )}
+            condition={({ isSourcegraphDotCom, licenseFeatures }) =>
+                isSourcegraphDotCom && licenseFeatures.isCodyEnabled
```
taras-yemets, line 46: However, Cody Pro routes are available only on dotcom, and Cody should always be enabled there. I’m leaving the license check in place “just in case” since I couldn’t find a hardcoded `true` for the Cody feature when in dotcom mode in our backend code.

client/web/src/cody/CodyProPage.tsx
```diff
@@ -0,0 +1,35 @@
...
+    [CodyProRoutes.AcceptInvite]: lazyComponent(() => import('./invites/AcceptInvitePage'), 'CodyAcceptInvitePage'),
+}
+
+interface CodyProPageProps extends Pick<LegacyLayoutRouteContext, 'authenticatedUser' | 'telemetryRecorder'> {
+    path: CodyProRoutes
+}
+
+/**
+ * Renders the appropriate Cody Pro page component for the given route path.
+ */
```
chrsmith, line 31: I don't think I've ever seen a `/** */` style comment in the code base yet. So would you mind doing a quick sanity check too see how common this is? If we already have enough usages that either would be considered common, then we can leave this as-is.

But if most doc comments like this are using single-line (`//`) comments, we should try to be consistent with the larger codebase.

client/web/src/cody/CodyProPage.tsx
```diff
@@ -0,0 +1,35 @@
...
+    [CodyProRoutes.ManageTeam]: lazyComponent(() => import('./team/CodyManageTeamPage'), 'CodyManageTeamPage'),
+    [CodyProRoutes.AcceptInvite]: lazyComponent(() => import('./invites/AcceptInvitePage'), 'CodyAcceptInvitePage'),
+}
+
+interface CodyProPageProps extends Pick<LegacyLayoutRouteContext, 'authenticatedUser' | 'telemetryRecorder'> {
+    path: CodyProRoutes
+}
+
+/**
+ * Renders the appropriate Cody Pro page component for the given route path.
```
chrsmith, line 30: Would you mind expanding on _why_ we are doing this? It might be confusing enough to call out why we are treating Cody Pro as somehow "different" from the rest of the site. e.g. something like:

```
// Renders the appropriate Cody Pro page component for the given route path.
//
// This is to more easily isolate the Cody Pro-specific functionality (which
// only applies to non-Enterprise users) from the rest of the Sourcegraph UI.

File: client/web/src/cody/codyProRoutes.tsx [hunk 4]
```diff
@@ -0,0 +1,50 @@
...
+    element: (
+        <LegacyRoute
+            render={props => (
+                <CodyProPage
+                    path={path}
+                    authenticatedUser={props.authenticatedUser}
+                    telemetryRecorder={props.platformContext.telemetryRecorder}
+                />
+            )}
+            condition={({ isSourcegraphDotCom, licenseFeatures }) =>
```
chrsmith, line 45: 🔴 How are we accounting for the Cody Pro pages that aren't ready to be released? (e.g. the new Teams and Invites experience.)

Previously we were using `isEmbeddedCodyProUIEnabled()` to ensure that we only displayed those new pages IFF certain flags were set. However, with this change, you could navigate to `/cody/teams/manage` or `/cody/manage/subscription/new` which can easily lead to confusion.

So I think we just need to NOT include the `Manage = '/cody/manage',` route here, and keep that in the "older" routing set up. (i.e. keep that as it was.)

And then for these new pages, add the `isEmbeddedCodyProUIEnabled()` check to this condition. So that we only allow these new pages once we are ready to go live with the new UI in sourcegraph.com.

Does that all make sense?

chrsmith (commented):
All of this looks right, and does take a step towards a better overall architecture. (🙌 👏)

But I think we need to double check that users can still use `/cody/manage` (like they do today), and serve a 404 error on sourcegraph.com for our newer pages (at least until we set certain site config or user flags.)

Also, if possible, I think it would be cleaner to just have any "Cody Pro" pages serve a 404 if you aren't running on dotcom mode. Since "error: this page is only available on sourcegraph.com" seems less pleasant than just a generic 404. (Since the end result is the same: that page doesn't exist [for Enterprise users].)

client/web/src/cody/codyProRoutes.tsx
vdavid: It's not that important and not my decision, but publishing these pages sounds okay to me because we don't link them anywhere.

vdavid (approved):
Yeah, I like it. Please either wait for @chrsmith's response for how comfortable he is with publishing these pages, or ask Security. That's the only thing, otherwise this is great! Thanks! 🙂

client/web/src/cody/CodyProPage.tsx
chrsmith: Thanks @vdavid , perhaps a better question I should have asked is: do we have an "official" style guide for various languages at Sourcegraph? Even if it is just "do whatever the {Microsoft, Google} style guide for {language} says"?

vdavid: We don't have much docs guidance for these things, what we have is linter rules. Apparently none for this one, probably it was so low value that no one has uniformized it.

client/web/src/cody/codyProRoutes.tsx
chrsmith: I would prefer that we don't expose these pages yet, and keep them hidden under a feature flag. Even if we don't link to them. However, I don't feel strongly enough about this to cause you any grief. So feel free to merge this as-is, or after making the change @taras-yemets , I'll leave that up to you.

For reference, here's my thinking:

- Even if we don't link to them, they are still there. And so I think we need to act as if a user were able to access them (either intentionally or unintentionally.)
- There is no security concern, and the functionality that is available should just work. So it isn't like we believe something _bad_ will happen.
- But ultimately there is a _liability_ with these pages, because we haven't done a lot of testing and validation. So, for example, if someone were to (somehow) go through the signup flow and trips over some bug we haven't gotten to you -- that would just be noise and/or a distraction for the team. (And so, just doing a little work to not make that even possible, IMHO, seems like the right call.)

So take that and do what you will :)

taras-yemets: @chrsmith, we still have `isEnabledEmbeddedCodyProUIEnabled` check on line 28. No logic change here.

client/web/src/cody/CodyProPage.tsx
taras-yemets: I used the JSDoc `/**` comment because it's shown in the hover tooltips, as opposed to the `/*` and `//` comments.  

Screenshot 2024-06-03 at 10 01 14

client/web/src/cody/CodyProPage.tsx
taras-yemets: Thank you! Addressed in https://github.com/sourcegraph/sourcegraph/pull/63003/commits/5686aec05047045538b1fb6e1cc145835acd8537

client/web/src/cody/codyProRoutes.tsx
taras-yemets: When `isEmbeddedCodyProUIEnabled()` is `false`, only '/cody/manage' and '/cody/subscription' pages are available. 
All the `CodyProRoutes` are available when `isEmbeddedCodyProUIEnabled()` is `true`.

taras-yemets:
@chrsmith, @vdavid, thanks for your reviews!

https://github.com/sourcegraph/sourcegraph/blob/0189c78bf46a698bcc85a589dd4a35bc2d1e71c2/client/web/src/cody/codyProRoutes.tsx#L23-L34

This PR doesn't make changes to the conditional routes rendering depending on the `isEmbeddedCodyProUIEnabled` experimental thingy (https://github.com/sourcegraph/sourcegraph/pull/63003#discussion_r1623881578).

PR #63004, opened 2024-05-31
Remove echo test alert

pjlast:
This removes the critical alert from the 1s echo test, since it has a tendency to wake someone up in the middle of the night and turns our alerts into noise.

The warning still remains.

## Test plan

Alert removed and docs updated.

### DIFF
--- a/monitoring/definitions/git_server.go
+++ b/monitoring/definitions/git_server.go
@@ -266,28 +266,6 @@ func GitServer() *monitoring.Dashboard {
 					},
 					{
 						{
-							Name:        "echo_command_duration_test",
-							Description: "echo test command duration",
-							Query:       "max(src_gitserver_echo_duration_seconds)",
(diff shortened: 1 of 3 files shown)


monitoring/definitions/git_server.go
```diff
@@ -265,28 +265,6 @@ func GitServer() *monitoring.Dashboard {
 						},
 					},
 					{
-						{
```
eseliger, line 268: I don't think we want to remove this metric at all. It's been a useful indicator on instances that had slow requests.

I also believe that the warning is justified. This is not a metric that should spike on a healthy instance. (Dotcom is not a healthy instance IMO).

That said, we don't seem to care about the alert on dotcom, and the critical warning might be too trigger friendly.

I recommend that we drop the critical alert for now, but keep the rest. WDYT?

pjlast: Alright sounds good!

eseliger (approved):

PR #63005, opened 2024-05-31
build(bazel): bump to rules_esbuild 0.16.0

gregmagolan:
1.6.0 is the minimum rules_esbuild version that support bazel-lib 2.x. After this lands, I'll put up a PR to get onto bazel-lib 2.x and after that lands that unblocks upgrading to rules_oci 2 which resolves at least some caching issues with large image artifacts.

Also pins the esbuild toolchain to esbuild 0.19.2 which should be a no-op since LATEST_ESBUILD_VERSION for rules_esbuild 0.15.3 (the version currently on main) is 0.19.2.

Both @Strum355 and I tried bumping to rules_esbuild 0.17.0+ and hit build issues (with and without the sandbox plugin enabled that was added in that version) so there is work to do there to track down what changed in the ruleset that breaks this build.

## Test plan

CI

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -284,6 +284,7 @@ load("@aspect_rules_esbuild//esbuild:repositories.bzl", "esbuild_register_toolch
 
 esbuild_register_toolchains(
     name = "esbuild",
+    # Note, this differs from the version noted in package.json, however we've been inadvertently building with this version for some time now so we'll stick with it and revisit.
     esbuild_version = "0.19.2",
 )
 
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -270,9 +270,9 @@ swc_register_toolchains(
 # rules_esbuild setup ===========================
 http_archive(
     name = "aspect_rules_esbuild",
-    sha256 = "84419868e43c714c0d909dca73039e2f25427fc04f352d2f4f7343ca33f60deb",
-    strip_prefix = "rules_esbuild-0.15.3",
-    url = "https://github.com/aspect-build/rules_esbuild/releases/download/v0.15.3/rules_esbuild-v0.15.3.tar.gz",
+    sha256 = "46aab76044f040c1c0bd97672d56324619af4913cb9e96606ec37ddd4605831d",
+    strip_prefix = "rules_esbuild-0.16.0",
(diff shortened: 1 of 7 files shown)


WORKSPACE
```diff
@@ -270,21 +270,21 @@ swc_register_toolchains(
...
 rules_esbuild_dependencies()
 
 # Register a toolchain containing esbuild npm package and native bindings
-load("@aspect_rules_esbuild//esbuild:repositories.bzl", "LATEST_ESBUILD_VERSION", "esbuild_register_toolchains")
+load("@aspect_rules_esbuild//esbuild:repositories.bzl", "esbuild_register_toolchains")
 
 esbuild_register_toolchains(
     name = "esbuild",
-    esbuild_version = LATEST_ESBUILD_VERSION,
+    esbuild_version = "0.19.2",
```
Strum355, line 287:
```suggestion
    # Note, this differs from the version noted in package.json, however we've been inadvertently building with this version for some time now so we'll stick with it and revisit.
    esbuild_version = "0.19.2",
```

dev/esbuild.bzl
```diff
@@ -1,10 +1,16 @@
 load("@aspect_bazel_lib//lib:copy_to_directory.bzl", "copy_to_directory")
-load("@aspect_rules_esbuild//esbuild:defs.bzl", "esbuild")
+load("@aspect_rules_esbuild//esbuild:defs.bzl", _esbuild = "esbuild")
+
+def esbuild(name, **kwargs):
+    _esbuild(
+        name,
+        **kwargs
+    )
```
Strum355, line 8: Do you need this change for something in a later PR?

gregmagolan: On the bump to 0.17.0 it gives you a common place to disable the sandbox plugin if needed. In general it is a nice pattern to have a pass-through for rules so you can add global changes like in `oci_defs.bzl`:

```
def oci_tarball(name, **kwargs):
    _oci_tarball(
        name = name,
        # Don't build this by default with bazel build //... since most oci_tarball
        # targets do not need to be built on CI. This prevents the remote cache from
        # being overwhelmed in the event that oci_tarballs are cache busted en masse.
        tags = kwargs.pop("tags", []) + ["manual"],
        **kwargs
    )
```

gregmagolan: Not strictly necessary here but a good pattern to have in-place either way

Strum355 (approved):

PR #63006, opened 2024-05-31
build(bazel): pin bazel fetched esbuild version to 0.19.2

gregmagolan:
Should be a no-op since `LATEST_ESBUILD_VERSION` for rules_esbuild 0.15.3 (the version currently on main) is 0.19.2.

## Test plan

CI


gregmagolan:
Rolling this into https://github.com/sourcegraph/sourcegraph/pull/63005

PR #63009, opened 2024-05-31
Svelte: clean up file tree

camdencheek:
After the changes to the file tree layout in #62981, we can clean up the styles and DOM a bit. These changes are partially in prep for experimenting with alternative "current path" representations. Comments inline.

## Test plan

Manual testing. There should be no change in behavior, just DOM/style simplification.

### DIFF
--- a/client/web-sveltekit/src/lib/TreeNode.svelte
+++ b/client/web-sveltekit/src/lib/TreeNode.svelte
@@ -138,7 +138,7 @@
 
     .loading {
         // Indent with two rem since loading represents next nested level
-        margin-left: calc(var(--tree-node-nested-level) * #{$shiftWidth} + 2 * var(--icon-inline-size) + 2 *#{$gap});
+        margin-left: calc(var(--tree-node-nested-level) * #{$shiftWidth} + 2 * var(--icon-inline-size) + 2 * #{$gap});
         margin-top: 0.25rem;
     }
 
--- a/client/web-sveltekit/src/lib/TreeNode.svelte
+++ b/client/web-sveltekit/src/lib/TreeNode.svelte
@@ -126,9 +126,8 @@
 </li>
 
 <style lang="scss">
+    $leftPadding: 1.25rem;
     [role='treeitem'] {
-        --tree-node-left-padding: 1.25rem;
-
         border-radius: var(--border-radius);
 
         &[tabindex='0']:focus {
@@ -142,7 +141,7 @@
 
     .loading {
         // Indent with two rem since loading represents next nested level
-        margin-left: calc(var(--tree-node-nested-level) * 1.25rem + 1.15rem + var(--tree-node-left-padding));
+        margin-left: calc(var(--tree-node-nested-level) * 1.25rem + 1.15rem + #{$leftPadding});
         margin-top: 0.25rem;
     }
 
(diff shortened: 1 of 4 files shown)


client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte
```diff
@@ -178,13 +178,12 @@
     }
 
     a {
-        flex: 1;
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte:181]: This shouldn't need to know it's in a flex box, so moved that to the `TreeNode` container

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte
```diff
@@ -178,13 +178,12 @@
...
         text-overflow: ellipsis;
         overflow: hidden;
         white-space: nowrap;
-        padding: 0.2rem 0.25rem 0.2rem 0;
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte:185]: This has weird padding because it adds padding to half of the label. Instead, we just add the padding on the label container so we don't need to keep it consistent across components.

client/web-sveltekit/src/lib/TreeView.svelte
```diff
@@ -235,21 +235,14 @@
...
         $treeState = { ...$treeState, focused: treeProvider.getNodeID(entries[0]) }
     }
 </script>
 
-<ul
-    role="tree"
-    bind:this={treeRoot}
-    on:keydown={handleKeydown}
-    on:click={handleClick}
-    data-tree-view-flat-list={isFlatList}
```
camdencheek, line 251: `data-tree-view-flat-list` is no longer used because we no longer indent files more than directories

client/web-sveltekit/src/lib/TreeNode.svelte
```diff
@@ -126,9 +121,10 @@
 </li>
 
 <style lang="scss">
-    [role='treeitem'] {
-        --tree-node-left-padding: 1.25rem;
+    $shiftWidth: 1.25rem;
```
camdencheek, line 124: `$shiftWidth` is a constant, so using an SCSS variable rather than a CSS variable.

client/web-sveltekit/src/lib/TreeNode.svelte
```diff
@@ -79,30 +79,25 @@
     data-node-id={nodeID}
     style="--tree-node-nested-level: {level}"
 >
-    <span bind:this={label} class="label" data-treeitem-label>
+    <span bind:this={label} class="label" data-treeitem-label class:expandable>
+        <Button variant="icon" on:click={handleScopeChange} data-scope-button>
+            <Icon svgPath={mdiImageFilterCenterFocusStrong} inline />
+        </Button>
```
camdencheek, line 85: Removed the wrapping div, allowing the button to sit directly in the DOM. Added `data-scope-button` so it's still able to be referred to in the CSS.

client/web-sveltekit/src/lib/TreeNode.svelte
camdencheek: Also, with the guide lines, I think we could experiment with a more condensed display because the relationships are more obvious.

client/web-sveltekit/src/lib/wildcard/Button.svelte
```diff
@@ -15,6 +15,7 @@
         // but for some reason it's not recognized when using `svelte-check` and an error
         // is thrown instead.
         'data-testid'?: string
+        'data-scope-button'?: boolean
```
camdencheek, line 18: I'm hopeful this will be fixed with native typescript support in Svelte 5

fkling (approved):

PR #63013, opened 2024-05-31
feat(cody-gateway): Add FLAGGED_MODEL_NAMES check

chrsmith:
Cody Gateway may "flag" a request as a way to either identify clearly abusive traffic, and reject it before it reaches an LLM, or to allow for further analysis.

This PR adds the ability to specify a set of model names, and Cody Gateway will in-turn flag all requests for that specific model. This adds another knob for being able to identify abusive behavior and/or confirm that any sudden spikes in a particular LLM model are for legitimate reasons.

With this PR no "default" is provided. And a follow-up PR will need to be made in the `infrastructure` repo to add model names as applicable.

- Part of https://linear.app/sourcegraph/issue/PRIME-263/flag-all-opus-all-gpt4-models-4o-4-turbo-etc.
- https://github.com/sourcegraph/abuse-ban-bot/pull/19, which updates the logic so that flagging a request for the `model_used` reason won't erroneously ban the user. (Since in this case, the "flag" isn't a sign of anything nefarious.)

## Test plan

Updated existing unit tests.

## Changelog

Add a FLAGGED_MODEL_NAMES env var for Cody Gateway. Any requests using a supplied model will be flagged.

### DIFF
--- a/cmd/cody-gateway/internal/httpapi/completions/flagging.go
+++ b/cmd/cody-gateway/internal/httpapi/completions/flagging.go
@@ -25,7 +25,7 @@ type flaggingConfig struct {
 	MaxTokensToSampleFlaggingLimit int
 	ResponseTokenBlockingLimit     int
 
-	// FlaggedModelNames is a slice of LLM model names, e.g. "gtp-3.5-turbo",
+	// FlaggedModelNames is a slice of LLM model names, e.g. "gpt-3.5-turbo",
 	// that will lead to the request getting flagged.
 	FlaggedModelNames []string
 
--- a/cmd/cody-gateway/internal/httpapi/completions/flagging.go
+++ b/cmd/cody-gateway/internal/httpapi/completions/flagging.go
@@ -2,6 +2,7 @@ package completions
 
 import (
 	"context"
+	"slices"
 	"strings"
 
 	"github.com/sourcegraph/sourcegraph/cmd/cody-gateway/shared/config"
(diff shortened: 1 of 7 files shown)


cmd/cody-gateway/internal/httpapi/completions/flagging.go
```diff
@@ -24,6 +25,10 @@ type flaggingConfig struct {
 	MaxTokensToSampleFlaggingLimit int
 	ResponseTokenBlockingLimit     int
 
+	// FlaggedModelNames is a slice of LLM model names, e.g. "gtp-3.5-turbo",
```
sqs, line 28:
```suggestion
	// FlaggedModelNames is a slice of LLM model names, e.g. "gpt-3.5-turbo",
```

gpt?

sqs (approved):

cmd/cody-gateway/internal/httpapi/completions/flagging.go
chrsmith: Perhaps "Google Premium TensorFlow" hasn't been publicly announced yet...

cmd/cody-gateway/internal/httpapi/completions/flagging_test.go
```diff
@@ -40,12 +40,14 @@ func TestMakeFlaggingConfig(t *testing.T) {
 
 func TestIsFlaggedRequest(t *testing.T) {
 	validPreamble := "You are cody-gateway."
+	flaggedModelNames := []string{"dangerous-llm-model"}
```
chenkc805, line 43: maybe "expensive-llm-model" instead?

chenkc805: actually this is just a test file so ignore

chenkc805 (approved):

PR #63014, opened 2024-05-31
Embeddings: hide site admin references

jtibshirani:
This PR fixes a bug in the site admin UI where we showed references to
embeddings even when they were disabled in site config.

In follow-up work, we plan to remove the embeddings logic completely.

Closes https://github.com/sourcegraph/cody-issues/issues/50

## Test plan

Manually tested!

### DIFF
--- a/client/web/src/cody/isCodyEnabled.tsx
+++ b/client/web/src/cody/isCodyEnabled.tsx
@@ -10,3 +10,5 @@ export const isCodyEnabled = (): boolean => {
 }
 
 export const isSignInRequiredForCody = (): boolean => !window.context.isAuthenticatedUser
+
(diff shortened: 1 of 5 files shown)


client/web/src/cody/isCodyEnabled.tsx
```diff
@@ -10,3 +10,5 @@ export const isCodyEnabled = (): boolean => {
 }
 
 export const isSignInRequiredForCody = (): boolean => !window.context.isAuthenticatedUser
+
+export const isEmbeddingsEnabled = (): boolean => window.context?.codyEnabled && window.context?.embeddingsEnabled
```
jtibshirani, line 14: I couldn't decide if 'embeddings' was singular or plural 😊

camdencheek (approved):
LGTM, thanks!

PR #63015, opened 2024-05-31
build(bazel): upgrade to rules_oci 2

gregmagolan:
Stacked on top of https://github.com/sourcegraph/sourcegraph/pull/63012 which should land first before this comes out of Draft.

rules_oci 2 has significant performance improvements for build times and for cacheing performance.

## Test plan

CI


gregmagolan:
Replaced by https://github.com/sourcegraph/sourcegraph/pull/63829

PR #63016, opened 2024-05-31
msp/runtime: add Job runtime

jac:
Currently we have a Service runtime entrypoint which parses env, creates a logger, initialises telemetry, creates the MSP contract, etc. according to MSP conventions.
However we don't have the equivalent for Jobs which requires operators to set up everything manually. 
This PR remedies the situation by adding a Job entrypoint which performs similar setup to the Service entrypoint.

Closes CORE-117
## Test plan
Successfully executed a mock job in GCP using this as an entrypoint

### DIFF
--- a/lib/managedservicesplatform/runtime/job.go
+++ b/lib/managedservicesplatform/runtime/job.go
@@ -1,0 +1,130 @@
+package runtime
+
+import (
+	"context"
+	"flag"
+	"os"
+
+	"cloud.google.com/go/profiler"
+	"github.com/sourcegraph/log"
+
+	"github.com/sourcegraph/sourcegraph/lib/managedservicesplatform/runtime/contract"
+	"github.com/sourcegraph/sourcegraph/lib/managedservicesplatform/runtime/internal/opentelemetry"
+)
+
+type Job[ConfigT any] interface {
+	contract.ServiceMetadataProvider
+	// Execute should use given configuration to perform the desired
+	// job execution.
+	Execute(
+		ctx context.Context,
+		logger log.Logger,
+		contract Contract,
+		config ConfigT,
+	) error
+}
+
+// ExecuteJob handles the entire lifecycle of the program performing a
+// job execution, and should be the only thing called in a MSP job
+// program's main package, for example:
+//
+//	runtime.ExecuteJob[example.Config](example.Job{})
+//
+// Where example.Config is your runtime.ConfigLoader implementation, and
+// example.Job is your runtime.Job implementation.
+func ExecuteJob[
+	ConfigT any,
+	LoaderT ConfigLoader[ConfigT],
+](job Job[ConfigT]) {
+	// To get around os.Exit not honoring defer we
+	// wrap the body in another function which we can safely defer inside
+	err := func() (err error) {
+		flag.Parse()
+		passSanityCheck(job)
+
+		// Resource representing the job
+		res := log.Resource{
+			Name:       job.Name(),
+			Version:    job.Version(),
+			Namespace:  "",
+			InstanceID: "",
+		}
+
+		liblog := log.Init(res, log.NewSentrySink())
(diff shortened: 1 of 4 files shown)


lib/managedservicesplatform/runtime/job.go
```diff
@@ -0,0 +1,130 @@
...
+		defer done(&err)
+
+		startLogger.Info("starting job execution",
+			log.String("job", job.Name()),
+			log.String("executionID", executionID),
+			log.Bool("msp", ctr.MSP),
+			log.Bool("sentry", sentryEnabled))
+
+		err = job.Execute(ctx, log.Scoped("job"), ctr, *config)
+		return
```
bobheadxi, line 123: I think this will work as well for capturing the error

```suggestion
		return job.Execute(ctx, log.Scoped("job"), ctr, *config)
```

lib/managedservicesplatform/runtime/job.go
```diff
@@ -0,0 +1,130 @@
...
+//
+// Where example.Config is your runtime.ConfigLoader implementation, and
+// example.Job is your runtime.Job implementation.
+func ExecuteJob[
+	ConfigT any,
+	LoaderT ConfigLoader[ConfigT],
+](job Job[ConfigT]) {
+	// To get around os.Exit not honoring defer we
+	// wrap the body in another function which we can safely defer inside
+	err := func() (err error) {
```
bobheadxi, line 41: The control flow here is a little bit confusing; all error paths except the actual execution use `logger.Fatal` to exit. Maybe we can do:

```suggestion
	// To get around os.Exit not honoring defer we
	// wrap the body in another function which we can safely defer inside
	failed := func() (failed bool) {
```

Instead?

jac: oh 🤦 
ofc logger.Fatal calls os.Exit

bobheadxi: I think the structure here still makes sense, as the error logging is handled by job execution check-in - we just want to let defers handle properly before using os.Exit ourselves. logger.Fatal isn't as bad because it will at least flush logs correctly first

bobheadxi: (but, maybe we should refactor both entrypoints to avoid logger.Fatal so that full cleanup can happen 🤔 )

jac: Given that `defer done(&err)` wasn't adding any benefit I remove the `defer` and moved the call to `done` after `job.Execute`. As such we don't need a nested function anymore.

The `log.Fatal` calls are still a concern as we might never reach the point to check in. I think that might be okay though as like you said it will flush the logs first

lib/managedservicesplatform/runtime/job.go
```diff
@@ -0,0 +1,124 @@
...
+		log.Bool("msp", ctr.MSP),
+		log.Bool("sentry", sentryEnabled))
+
+	err = job.Execute(ctx, log.Scoped("job"), ctr, *config)
+	done(err)
+
+	if err != nil {
+		os.Exit(1)
+	}
+	os.Exit(0)
```
bobheadxi, line 123:
```suggestion
	// Logging the error is handled by `done(err)` above, so here we just exit
	// with an appropriate exit code.
	if err != nil {
		os.Exit(1)
	}
	os.Exit(0)
```

bobheadxi (approved):
Nice!

bobheadxi:
@jac should we file a follow-up issue to migrate https://github.com/sourcegraph/abuse-ban-bot to this new entrypoint?

jac:
> @jac should we file a follow-up issue to migrate https://github.com/sourcegraph/abuse-ban-bot to this new entrypoint?

Yeah, ideally we should migrate over all msp jobs if possible

PR #63019, opened 2024-05-31
build(bazel): bump to rules_esbuild 0.20.1 with sandbox plugin disabled

gregmagolan:
Bump to the latest non-rules_js-2 rules_esbuild release but leaves the sandbox plugin disabled since it causes build failures. A follow-up will be needed to enable the plugin so bundling is hermetic to fix potentional non-hermetic flaky build failures seen recently.

## Test plan

CI


Strum355:
Superseded by https://github.com/sourcegraph/sourcegraph/pull/63022

Strum355:
Not superseded actually, I misread 0.20.1 in the other PR as 0.21.0 :sweat_smile:

PR #63022, opened 2024-06-01
build(bazel): upgrade to rules_js 2.0 RC

gregmagolan:
Bumps to rules_js (and friends) to 2.0 RCs.

This brings in performance improvements for analysis phase since npm package depsets and now much smaller. It also adds support for pnpm v9 and allows for linking js_library targets as 1p deps instead of npm_package targets. See https://github.com/aspect-build/rules_js/issues/1671 for more details.

## Test plan

CI


Strum355 (approved):
full send

PR #63024, opened 2024-06-02
chore: Change errors.HasType to respect multi-errors

varungandhi-src:
With this patch, the `errors.HasType` API behaves similar to `Is` and `As`,
where it checks the full error tree instead of just checking a linearized version
of it, as cockroachdb/errors's `HasType` implementation does not respect
multi-errors (https://github.com/cockroachdb/errors/issues/145).

As a consequence, a bunch of relationships between HasType and Is/As that
you'd intuitively expect to hold are now true; see changes to `invariants_test.go`.

Q: Why not make the change upstream and submit a version bump PR here?
A: This PR also tweaks the API to use a type argument rather than a value
argument, since the call-site is less clearer when you pass a _value_ to check
the _type_. However, that would be a breaking change to make upstream,
so we just make that here.

## Test plan

If we have tests for the existing error paths, then this should be good enough.

## Changelog

### DIFF
--- a/lib/errors/cockroach.go
+++ b/lib/errors/cockroach.go
@@ -109,6 +109,11 @@ func AsInterface[I any](err error, target *I) bool {
 	return errors.As(err, target)
 }
 
+func HasTypeGeneric[T error](err error) bool {
+	var zero T
+	return errors.As(err, &zero)
+}
+
 // Extend multiError to work with cockroachdb errors. Implement here to keep imports in
 // one place.
 
--- a/lib/errors/cockroach.go
+++ b/lib/errors/cockroach.go
@@ -62,7 +62,6 @@ var (
 	// concrete type but with different data.
 	Is        = errors.Is
 	IsAny     = errors.IsAny
-	HasType   = errors.HasType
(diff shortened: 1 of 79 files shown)


lib/errors/cockroach.go
```diff
@@ -78,14 +78,24 @@ var (
...
 	As        = errors.As
-	HasType   = errors.HasType
 	Cause     = errors.Cause
 	Unwrap    = errors.Unwrap
 	UnwrapAll = errors.UnwrapAll
 
 	BuildSentryReport = errors.BuildSentryReport
 )
 
+// HasType checks if the error tree err has a node of type T.
```
camdencheek, line 88: I think it might be better to specifically mention that this is just a convenience wrapper around `As` that discards the assigned variable since this no longer specifically checks the type, and also responds to implementations of `interface{ As(...) }`, which is allowed to completely ignore the actual type.

varungandhi-src: I added a CAVEAT bit to the docs here mentioning `interface { As(...) }`.

PR #63029, opened 2024-06-03
chore: Use new cmp package to simplify comparisons

varungandhi-src:
There is also a [`cmp.Or` function](https://pkg.go.dev/cmp#Or) but since that's a function and
not a macro, it would cause all comparisons to always happen,
so we don't use that for chaining results.

## Test plan

Covered by existing tests

## Changelog

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -2,6 +2,7 @@
 package api
 
 import (
+	"cmp"
 	"fmt"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


kritzcreek (approved):
LGTM, I was confused about the "macro" comment at first... but after googling for "go macros" for a while I think you wanted to convey that there's no way to lazily evaluate the arguments to `cmp.Or`.

varungandhi-src:
> but after googling for "go macros" for a while I think you wanted to convey that there's no way to lazily evaluate the arguments to cmp.Or

I did mean macro, because macros are much more common in languages like Rust, C etc. whereas lazy evaluation is not that common. In most languages, the usual or operator `||` is macro-like, it doesn't do lazy evaluation.

PR #29652, opened 2022-01-12
honeycomb: slice-wrap values for multiple insert of same key

Strum355:
Currently, if `AddField` n co are called multiple times for the same key, only the last value remains. This can happen using the `observation` package with the traceLogger. 

With this PR, if a key is added multiple times, it will be wrapped in a slice after the second insert and that slice appended to thereafter. 

We apply a custom JSON marshalling scheme for (subjectively) optimal representation in Honeycomb. Image 1 shows what it would look like normally, Image 2 shows the custom marshalling:

image
image

PR #63040, opened 2024-06-03
chore: Simplify RequestState/PositionalRequestArgs o11y code

varungandhi-src:
Moving the common observability attributes code into helper methods makes
reading the observation related code less painful.

TODO:
- [x] Add a function to de-duplicate the attributes, because the current implementation
  will end up creating a slice on seeing duplicate keys for the fields common between
  RequestState and PositionalRequestArgs.

@Strum355, do you know why we keep multiple values for the same key instead
of having overwriting semantics? You added this behavior here:
https://github.com/sourcegraph/sourcegraph/pull/29652

## Test plan

Covered by existing tests

## Changelog

### DIFF
--- a/internal/observation/fields.go
+++ b/internal/observation/fields.go
@@ -1,6 +1,7 @@
 package observation
 
 import (
+	orderedmap "github.com/wk8/go-ordered-map/v2"
 	"go.opentelemetry.io/otel/attribute"
 	"go.uber.org/zap" //nolint:logging // This is an expected usage
 
@@ -21,4 +22,24 @@ func attributesToLogFields(attributes []attribute.KeyValue) []log.Field {
 		}
 	}
 	return fields
+}
+
+// MergeAttributes merges a list of attributes into a single list of
+// attributes, with last-preferred semantics.
+func MergeAttributes(attributes []attribute.KeyValue, more ...attribute.KeyValue) []attribute.KeyValue {
+	m := orderedmap.New[string, attribute.Value]()
+	for _, attr := range attributes {
+		m.Set(string(attr.Key), attr.Value)
+	}
+	for _, attr := range more {
+		m.Set(string(attr.Key), attr.Value)
+	}
+	var out []attribute.KeyValue
+	for p := m.Oldest(); p != nil; p = p.Next() {
+		out = append(out, attribute.KeyValue{
+			Key:   attribute.Key(p.Key),
+			Value: p.Value,
+		})
+	}
+	return out
 }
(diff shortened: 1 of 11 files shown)


internal/observation/fields.go
```diff
@@ -1,6 +1,7 @@
 package observation
 
 import (
+	orderedmap "github.com/wk8/go-ordered-map/v2"
```
kritzcreek, line 4: Are we already depending on this somewhere else? I was surprised to see that a library for a datastructure also includes yaml and json parsers https://github.com/wk8/go-ordered-map/blob/85ca4a2b29d3241fa4513f82be3d38fe2392a791/go.mod#L5-L11

kritzcreek (commented):
LGTM in general (thanks for doing this, I assume I "nerdsniped" you with my comment during our call)

Just have one question regarding the ordered-map dependency.

internal/observation/fields.go
varungandhi-src: It's for implementing deserialization from JSON and YAML, since the stdlib implementation for JSON parsing is slower. https://sourcegraph.com/github.com/wk8/go-ordered-map/-/blob/json.go?L109:9-109:19

varungandhi-src: > Are we already depending on this somewhere else?

Yes, but not in many places

varungandhi-src:
> LGTM in general (thanks for doing this, I assume I "nerdsniped" you with my comment during our call)

No, I was working on this locally before our call.

internal/observation/fields.go
kritzcreek: Well if we're already pulling it in I guess that's fine. I understand what it's used for, it just screams bad library at me. Probably just my intuitions being wrong in the context of Go again.

kritzcreek (approved):

internal/observation/fields.go
varungandhi-src: It's similar in Rust, many libraries depend on serde via a crate feature. https://github.com/rust-random/rand/blob/master/Cargo.toml#L33

Go typically doesn't do features.

kritzcreek: Right, if it was feature flagged or in a "ordered-map-serialisation" library I wouldn't have flinched at all.

varungandhi-src: It can't be in a separate library because you can't add methods to a type in a separate library (somewhat similar to the orphan rule), and the composable way for a type to be (de)serializable is by adding methods with specific signatures.

Strum355:
> @Strum355, do you know why we keep multiple values for the same key instead of having overwriting semantics? You added this behavior here: github.com/sourcegraph/sourcegraph/pull/29652
The idea was for use in per-loop-iteration data point collection. Im not particularly happy with the implementation, Im also not sure if its actively being used anywhere. Bit of a hack that I wouldn't be against removing

PR #63041, opened 2024-06-03
Svelte: Add telemetry v2 to svelte client

vovakulikov:
Completes [SRCH-68](https://linear.app/sourcegraph/issue/SRCH-68/svelte-migrate-telemetry-service-to-v2-events)

This PR simply adds yet another telemetry v2 usage to the svelte FE client. 

## Test plan
Manual testing of newly added events

### DIFF
--- a/client/web-sveltekit/src/lib/telemetry2.ts
+++ b/client/web-sveltekit/src/lib/telemetry2.ts
@@ -1,4 +1,8 @@
 import { gql } from '@urql/core'
+
+import type { BillingCategory, BillingProduct } from '@sourcegraph/shared/src/telemetry'
+import { sessionTracker } from '@sourcegraph/shared/src/telemetry/web/sessionTracker'
+import { userTracker } from '@sourcegraph/shared/src/telemetry/web/userTracker'
 import {
     type MarketingTrackingProvider,
     type TelemetryEventMarketingTrackingInput,
@@ -8,13 +12,9 @@ import {
     TelemetryRecorderProvider as BaseTelemetryRecorderProvider,
 } from '@sourcegraph/telemetry'
 
-import type { BillingCategory, BillingProduct } from '@sourcegraph/shared/src/telemetry'
 import type { GraphQLClient } from '$lib/graphql'
-import type { ExportTelemetryEventsResult } from '$lib/graphql-operations'
-
 import { getGraphQLClient } from '$lib/graphql'
-import { userTracker } from '@sourcegraph/shared/src/telemetry/web/userTracker'
-import { sessionTracker } from '@sourcegraph/shared/src/telemetry/web/sessionTracker'
+import type { ExportTelemetryEventsResult } from '$lib/graphql-operations'
 
 /**
  * TelemetryRecorderProvider is the default provider implementation for the
--- a/client/web-sveltekit/src/lib/telemetry2.ts
+++ b/client/web-sveltekit/src/lib/telemetry2.ts
@@ -1,0 +1,109 @@
+import { gql } from '@urql/core'
+import {
+    type MarketingTrackingProvider,
+    type TelemetryEventMarketingTrackingInput,
+    type TelemetryEventInput,
+    type TelemetryExporter,
+    MarketingTrackingTelemetryProcessor,
+    TelemetryRecorderProvider as BaseTelemetryRecorderProvider,
+} from '@sourcegraph/telemetry'
+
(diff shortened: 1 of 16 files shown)


client/web-sveltekit/src/lib/telemetry2.ts
```diff
@@ -0,0 +1,109 @@
...
+            }
+        )
+    }
+}
+
+function getTelemetrySourceClient(): string {
+    if (window.context?.sourcegraphDotComMode) {
+        return 'dotcom.svelte-web'
+    }
+    return 'server.svelte-web'
```
vovakulikov, line 60: @dadlerj I changed this to `'server.svelte-web'` to be able select/filter out only svelte 
events later on our dashboards

camdencheek: Why is this part of the client name? Aren't all of our events attached to an instance anyways?

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+page.svelte
```diff
@@ -16,6 +17,7 @@
 
     onMount(() => {
         SVELTE_LOGGER.logViewEvent(SVELTE_TELEMETRY_EVENTS.ViewRepositoryPage)
+        TELEMETRY_V2_RECORDER.recordEvent('repo', 'view')
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+page.svelte:20]: Q: why are we still maintaining the old event logs? Are we able to delete those after this is merged?

client/web-sveltekit/src/routes/search/SearchResults.svelte
```diff
@@ -181,15 +193,21 @@
...
+                    <ol on:copy={handleResultCopy}>
                         {#each resultsToShow as result, i}
                             {@const component = getSearchResultComponent(result)}
                             {#if i === resultsToShow.length - 1}
-                                <li use:observeIntersection on:intersecting={loadMore}>
+                                <li
+                                    use:observeIntersection
+                                    on:intersecting={loadMore}
+                                    on:click={() => handleSearchResultClick(i)}
+                                >
```
camdencheek, line 204: Do we want to be tracking every click on a search result? Or just clicks for the result contents that navigate? e.g., there are also things like the preview button or "show more matches" buttons that are not the same kind of actions.

camdencheek (approved):

client/web-sveltekit/src/routes/search/SearchResults.svelte
vovakulikov: Good point, the current behavior in the react version works with this flaw as well, so I didn't try to fix it here just to avoid any differences between different tracking systems. 

I'll review our telemetry anyway in the near future when we will be removing the old telemetry so I don't take this as a big problem for now.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+page.svelte
Comment (vovakulikov) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+page.svelte:20]: I hesitate to delete the old system just because their dashboards are still very convenient. As soon as we completely switch to the new version, I'll remove the old telemetry altogether.

client/web-sveltekit/src/routes/search/SearchResults.svelte
camdencheek: Makes sense to me 👍

client/web-sveltekit/src/lib/telemetry2.ts
vovakulikov: @camdencheek are you asking about dotcom and `server.<instance-name>` naming?

camdencheek: I'm asking why our client name has to include whether or not we're on dotcom. I think our client name should just be `svelte-web`, because that's what is sourcing these events. Whether or not the instance is dotcom should be determined by the instance ID attached to the event logs

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+page.svelte
Comment (camdencheek) [reply client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+page.svelte:20]: Sounds good 👍

client/web-sveltekit/src/lib/telemetry2.ts
camdencheek: (also, if we can get rid of one more thing that is special-cased for dotcom, let's do it)

PR #63043, opened 2024-06-03
gitserver: Increase default GitLongCommandTimeout

eseliger:
This has historically been set to 1 hour.
We've seen several reports of users running into the limit for clones of very large repositories, but we have seen no complaints of processes hanging for very long and clogging any queues.
So it feels sensible to me to increase the default for this value to 2h.

We might come back here later and decide that we don't really need a deadline here at all and instead hard-code a day or so to prevent infinite clogging, but let's see how far 2x gets us for now.

Test plan:

CI still passes.

### DIFF
--- a/internal/conf/computed.go
+++ b/internal/conf/computed.go
@@ -618,7 +618,7 @@ func AuthLockout() *schema.AuthLockout {
 	return val
 }
 
-const defaultGitLongCommandTimeout = time.Hour
+const defaultGitLongCommandTimeout = 2 * time.Hour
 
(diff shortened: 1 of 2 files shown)


eseliger:
* **#63043** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/63043?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

pjlast (approved):

PR #63044, opened 2024-06-03
Use anyhow context, bail, and ensure instead of panicking

keynmol:
I was indexing some projects and occasionally it'd fail with 

```
thread 'main' panicked at crates/syntax-analysis/src/globals.rs:191:17:
declare only one scope per match
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

With no extra information, circumventing the error handling in the levels above.

This PR uses anyhow's helpers to ensure that the errors are propagated

## Test plan

- N/A


kritzcreek (approved):
Nice, thanks!

PR #63047, opened 2024-06-03
chore: Replace errors.As with generic As and AsInterface

varungandhi-src:
Splits the signature of `errors.As` into two more specialized functions
which catch more errors at compile-time using generics.

## Test plan

Covered by existing tests, added some new tests.

## Changelog

### DIFF
--- a/lib/errors/cockroach.go
+++ b/lib/errors/cockroach.go
@@ -60,24 +60,8 @@ var (
 	// returns true for a value other than the one returned by As,
 	// since an error tree can contain multiple errors of the same
 	// concrete type but with different data.
-	Is    = errors.Is
-	IsAny = errors.IsAny
-	// As checks if the error tree err is of type target, and if so,
-	// sets target to the value of the error. This can be used in two ways:
-	//
-	// 1. If looking for an error of concrete type T, then the second
-	//    argument must be a non-nil pointer of type *T. This implies that
-	//    if the error interface is implemented with a pointer receiver,
-	//    then target must be of type **MyConcreteType.
-	// 2. If looking for an error satisfying an interface I (with a value
-	//    or pointer receiver), then the second argument must be of type I.
-	//
-	// For error types which do not contain any data, As is equivalent to Is.
-	//
-	// For error types which contain data, As will return an arbitrary
-	// error of the target type, in case there are multiple errors of the
-	// same concrete type in the error tree.
-	As        = errors.As
+	Is        = errors.Is
+	IsAny     = errors.IsAny
 	HasType   = errors.HasType
 	Cause     = errors.Cause
 	Unwrap    = errors.Unwrap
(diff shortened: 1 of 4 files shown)


lib/errors/invariants_test.go
```diff
@@ -120,20 +120,12 @@ func TestInvariants(t *testing.T) {
 			if Is(err, &payloadLessPtrError{}) {
 				// This can be false, see Counter-example 1
 				//require.True(t, HasType(err, &payloadLessPtrError{}))
-				require.Panics(t, func() {
-					var check payloadLessPtrError
-					require.True(t, As(err, &check))
-				})
```
varungandhi-src, line 126: These panics become type errors, so we can just remove them instead of checking for panics.

camdencheek (approved):
Very nice, thank you!

cc @bobheadxi just because I think you probably have the most context around `lib/errors`

lib/errors/multi_error.go
```diff
@@ -2,6 +2,8 @@ package errors
 
 import (
 	"fmt"
+
+	"github.com/cockroachdb/errors" //nolint:depguard // needed for implementation of multiError.As
```
bobheadxi, line 6: Why do we need the upstream `As`, instead of your new `As`?

camdencheek: We can constrain _our_ usage of `errors.As` with the new interface, but we shouldn't change the implementation of `interface{ As() }` to be specific to this library because multierrors sourced to this library can be  passed to code that doesn't use this library. Otherwise, 3rd-party code that calls, e.g., `errors.As(err, netErr)` on an error from this library will no longer work correctly.

bobheadxi: > we shouldn't change the implementation of interface{ As() } to be specific to this library

Ah, this clicks for me 😁 Could we add a brief docstring to this effect where we use As?

camdencheek: We could probably also use the stdlib `errors.As()` here. Cockroach shouldn't be adding any special behavior to that one

varungandhi-src: I've tried to make this change semantically a no-op as much as possible. It's potentially a bit riskier to move between libraries (ideally they should have the same semantics but...), we can make that change in a follow-up PR.

lib/errors/cockroach.go
```diff
@@ -86,6 +70,39 @@ var (
...
+// error of the target type, in case there are multiple errors of the
+// same concrete type in the error tree.
+//
+// Compared to errors.As, this method uses a generic argument to prevent
+// a runtime panic when target is not a pointer to an error type.
+func As[T error](err error, target *T) bool {
+	return errors.As(err, target)
+}
+
+// AsInterface checks if the error tree err is of type target, and if so,
```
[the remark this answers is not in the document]
varungandhi-src: Good catch, fixed the doc comments

lib/errors/multi_error.go
varungandhi-src: > Could we add a brief docstring to this effect where we use As?

Added here: https://github.com/sourcegraph/sourcegraph/pull/63047/commits/bbf7a9f38d86a061b97eb95f4cd5de59f706fe59

PR #63048, opened 2024-06-03
GRAPH-646: scip syntax java package namespace fix

keynmol:
Fixes GRAPH-646

The tree in reproducer is:

```
(program ; [0, 0] - [4, 0]
  (package_declaration ; [0, 0] - [1, 16]
    (marker_annotation ; [0, 0] - [0, 11]
      name: (identifier)) ; [0, 1] - [0, 11]
    (scoped_identifier ; [1, 8] - [1, 15]
      scope: (identifier) ; [1, 8] - [1, 11]
      name: (identifier))) ; [1, 12] - [1, 15]
  (class_declaration ; [3, 0] - [3, 12]
    name: (identifier) ; [3, 6] - [3, 9]
    body: (class_body))) ; [3, 10] - [3, 12]
```

I _think_ this was an error introduced in https://github.com/sourcegraph/sourcegraph/pull/53305/files#diff-0403a6bac5a125ef13ce5966a6c668eb3d5ca9c7fbb382d098f178c5d2c59f66R2-R6 - annotation implicitly being marked as opening of a scope?

## Test plan

- Reproducer + snapshot test


docker-images/syntax-highlighter/crates/syntax-analysis/src/snapshots/syntax_analysis__test__scip_snapshot_globals.java.snap.new
kritzcreek, line 1: I think `.snap.new` files are for files you still need to review with `insta`. We don't want to commit these and instead accept their changes into the `.snap` file.

kritzcreek (changes requested):

docker-images/syntax-highlighter/crates/syntax-analysis/src/snapshots/syntax_analysis__test__scip_snapshot_globals.java.snap.new
varungandhi-src: Let's add `.snap.new` to `.gitignore` so that they don't accidentally get committed.

docker-images/syntax-highlighter/crates/scip-syntax/testdata/java/index.scip
```diff
@@ -1 +1 @@
-../../../../../../bazel-bin/docker-images/syntax-highlighter/crates/scip-treesitter-cli/index-java.scip
\ No newline at end of file
+../../../../../../bazel-bin/docker-images/syntax-highlighter/crates/scip-syntax/index-java.scip
```
keynmol, line 1: Looks like this wasn't updated during crate reshuffle - only affects local `cargo test`, Bazel tests work fine.

kritzcreek (approved):
:+1:

PR #63049, opened 2024-06-03
chore/msp/runtime: disable telemetry on Google Monitoring/Trace APIs

bobheadxi:
Right now these emit a bunch of metrics and traces _for exporting metrics and traces_ that aren't very interesting. In particular they make the GCP console trace view very annoying to use because it's just filled with a bunch of "export trace" and "export metric" traces.

This disables telemetry for both. We only care about actual failures, which we do log.

https://sourcegraph.com/github.com/googleapis/google-api-go-client/-/blob/transport/grpc/dial.go?L362-367

## Test plan

n/a


unknwon (approved):
_no comment_

PR #63051, opened 2024-06-03
worker: add SAMS notifications subscriber

unknwon:
Part of CORE-92

This PR add a new worker for subscribing to [SAMS notifications](https://www.notion.so/sourcegraph/SAMS-notifications-distribution-system-0d174480e0044b05b545d37d24263d5a). The current use case is to automatically (hard-)delete users on Sourcegraph.com when the corresponding user is deleted from SAMS. 

This worker is only started when running in the Sourcegraph.com mode and the credentials file (`service_account.json`) is provided, which has been configured since https://github.com/sourcegraph/deploy-sourcegraph-cloud/pull/18591.

## Test plan

Tested locally end-to-end using my test GCP project.

1. Create test service account and download the credentials file
2. Create test Pub/Sub topic and subscription
3. Assign "Pub/Sub Subscriber" role to the service account on the subscription
1. Boot up SAMS locally
	```zsh
	export NOTIFICATION_GCP_PUBSUB_PROJECT="joes-playground-425319"
	export NOTIFICATION_GCP_PUBSUB_TOPIC="sams-notifications"
	overmind start
	```
4. Boot up SG in dotcom mode locally
	```yaml
	# sg.config.overwrite.yaml
	commandsets:
	  dotcom:
	    env:
	      SOURCEGRAPHDOTCOM_MODE: true
	      SOURCEGRAPH_ACCOUNTS_CREDENTIALS_FILE: './accounts-credentials.json'
	      SOURCEGRAPH_ACCOUNTS_NOTIFICATIONS_PROJECT: 'joes-playground-425319'
	```
	```json
	# dev-private/.../site-config.json
    "auth.providers": [
	  {
	    "allowSignup": true,
	    "clientID": "sams_cid_018ea03c-55d2-765c-9074-98acfe0bc520",
	    "clientSecret": "REDACTED",
	    "displayName": "Sourcegraph Accounts",
	    "configID": "sams",
	    "issuer": "http://localhost:9991",
	    "type": "openidconnect"
	  }
	]
	```
	```zsh
	sg start dotcom
	```
1. Delete a SAMS user
	```zsh
	curl -X DELETE -H "Authorization: Bearer foolmeifyoucan"  http://localhost:9991/api/management/v1/users/018d21f2-10ba-7f83-84b9-615678d6b383
	```
1. Wait for SG to pick it up (temporarily changed from DEBUG to INFO logging)
	CleanShot 2024-06-03 at 15 43 14@2x


cmd/worker/internal/sourcegraphaccounts/notifications_subscriber.go
```diff
@@ -0,0 +1,235 @@
...
+func (s *notificationsSubscriber) Config() []env.Config {
+	return []env.Config{s.config}
+}
+
+func (s *notificationsSubscriber) Routines(ctx context.Context, observationCtx *observation.Context) ([]goroutine.BackgroundRoutine, error) {
+	logger := observationCtx.Logger.Scoped("sourcegraphAccounts.notificationsSubscriber")
+	if !dotcom.SourcegraphDotComMode() || s.config.GCP.CredentialsFile == "" {
+		logger.Debug("worker disabled because it is not running on Sourcegraph.com or missing GCP credentials file")
+		return nil, nil
+	}
```
bobheadxi, line 56: I'm not sure if we have a convention or not, but this is a log that will show up in customer instances as well. IMO we should silently exit outside dotcom, and fail/succeed with more visible logs in dotcom:

```suggestion
	if !dotcom.SourcegraphDotComMode() {
		return nil, nil // not relevant
	}
	if s.config.GCP.CredentialsFile == "" {
		logger.Info("worker disabled because SOURCEGRAPH_ACCOUNTS_CREDENTIALS_FILE GCP is not set")
		return nil, nil
	}
	logger.Debug("worker enabled")
```

cmd/worker/internal/sourcegraphaccounts/notifications_subscriber.go
```diff
@@ -0,0 +1,235 @@
...
+func (s *notificationsSubscriber) Description() string {
+	return "Receives notifications from Sourcegraph Accounts."
+}
+
+func (s *notificationsSubscriber) Config() []env.Config {
+	return []env.Config{s.config}
+}
+
+func (s *notificationsSubscriber) Routines(ctx context.Context, observationCtx *observation.Context) ([]goroutine.BackgroundRoutine, error) {
+	logger := observationCtx.Logger.Scoped("sourcegraphAccounts.notificationsSubscriber")
```
bobheadxi, line 52: The logger is already scoped (see your screenshot)

```suggestion
```

IMO in frameworks that pass loggers to subcomponents, the framework sets the scope and fields about the subcomponent

cmd/worker/internal/sourcegraphaccounts/notifications_subscriber.go
```diff
@@ -0,0 +1,235 @@
...
+	var samsProvider *schema.OpenIDConnectAuthProvider
+	authProviders := conf.Get().SiteConfig().AuthProviders
+	for _, p := range authProviders {
+		if p.Openidconnect != nil && strings.HasPrefix(p.Openidconnect.ClientID, "sams_cid_") {
+			samsProvider = p.Openidconnect
+			break
+		}
+	}
+	if samsProvider == nil {
+		logger.Debug("worker disabled because SAMS provider is not configured")
```
bobheadxi, line 70: The default state in dotcom is this is enabled, so IMO disabled state should be more visible

```suggestion
		logger.Info("worker disabled because SAMS provider is not configured")
```

cmd/worker/internal/sourcegraphaccounts/notifications_subscriber.go
```diff
@@ -0,0 +1,235 @@
...
+		return nil, nil
+	}
+
```
bobheadxi, line 73: Maybe add info log `worker enabled` with fields indicating the SAMS provider details? In case we ever do run into an accident where SAMS connections are configured

cmd/worker/internal/sourcegraphaccounts/notifications_subscriber.go
```diff
@@ -0,0 +1,235 @@
...
+	credentialsJSON, err := os.ReadFile(s.config.GCP.CredentialsFile)
+	if err != nil {
+		return nil, errors.Wrap(err, "read GCP credentials file")
+	}
+	credentials, err := google.CredentialsFromJSON(ctx, credentialsJSON, pubsub.ScopePubSub)
+	if err != nil {
+		return nil, errors.Wrap(err, "parse GCP credentials JSON")
+	}
```
bobheadxi, line 111: Should this happen up-front in config load instead?

cmd/worker/internal/sourcegraphaccounts/notifications_subscriber.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+func (c *notificationsSubscriberConfig) Load() {
+	c.GCP.CredentialsFile = c.Get("SOURCEGRAPH_ACCOUNTS_CREDENTIALS_FILE", "", "Path to the Google Cloud credentials file")
+	c.GCP.ProjectID = c.Get("SOURCEGRAPH_ACCOUNTS_NOTIFICATIONS_PROJECT", "sourcegraph-dev", "The GCP project that the service is running in")
+	c.GCP.SubscriptionID = c.Get("SOURCEGRAPH_ACCOUNTS_NOTIFICATIONS_SUBSCRIPTION", "sams-notifications", "GCP Pub/Sub subscription ID to receive SAMS notifications from")
+}
+
+func handleOnUserDeleted(
```
bobheadxi, line 148: IMO the handlers deserve their own file for readability and expansion - e.g. in the future, we might have:

```
user_deletion.go
user_rename.go
```

etc.

mockgen.test.yaml
```diff
@@ -347,6 +347,10 @@
   path: github.com/sourcegraph/sourcegraph/cmd/worker/internal/telemetry
   interfaces:
     - bookmarkStore
+- filename: cmd/worker/internal/sourcegraphaccounts/mocks_test.go
+  path: github.com/sourcegraph/sourcegraph/cmd/worker/internal/sourcegraphaccounts
+  interfaces:
+    - notificationsSubscriberStore
```
bobheadxi, line 353: :yes:

cmd/worker/internal/sourcegraphaccounts/notifications_subscriber.go
unknwon: Good call, updated in https://github.com/sourcegraph/sourcegraph/pull/63051/commits/42cf67cadd042b9fb62eb27e01b707401c5e0f5f

cmd/worker/internal/sourcegraphaccounts/notifications_subscriber.go
unknwon: Updated in https://github.com/sourcegraph/sourcegraph/pull/63051/commits/42cf67cadd042b9fb62eb27e01b707401c5e0f5f

cmd/worker/internal/sourcegraphaccounts/notifications_subscriber.go
unknwon: Added in https://github.com/sourcegraph/sourcegraph/pull/63051/commits/9e55151401aa162c3aadd18c0fb7a15ebc1434b9

cmd/worker/internal/sourcegraphaccounts/notifications_subscriber.go
unknwon: I think all of config init steps here equally crucial, and `CredentialsFromJSON` doesn't really do any validation other than expecting valid JSON format. Thus prefer init them in the same order as the fields of `notificationsv1.SubscriberOptions`.

cmd/worker/internal/sourcegraphaccounts/notifications_subscriber.go
unknwon: I ended up restructuring the files a bit in https://github.com/sourcegraph/sourcegraph/pull/63051/commits/f7ab03a3a8ebcad6e69670bf6e95cfd23ddf87c8, PTAL!

I do want anything notifications-related visually and structurally grouped, since this package is generic "sourcegraphaccounts", who knows if we gonna add more workers here 😁

bobheadxi: Much nicer, thanks 😁

bobheadxi (approved):

PR #63056, opened 2024-06-04
chore: Make GetDefinitions API consistent with GetReferences etc.

varungandhi-src:
Implementing precise usages on top of the existing APIs becomes easier
if they're all of the same shape.

## Test plan

Covered by existing tests.

## Changelog


kritzcreek (approved):
LGTM

PR #63059, opened 2024-06-04
chore(codeintel): usagesForSymbol - make start/end non-optional for now

varungandhi-src:
We can add support for optional start/end later
in a backwards-compatible way. Making these non-optional
helps simplify the initial implementation.

## Test plan

Check that `sg start enterprise-codeintel` works locally.

## Changelog


kritzcreek (commented):
LGTM, CI failure looks like a flake

kritzcreek (approved):

PR #63062, opened 2024-06-04
chore: Rename Contains to IsSupersetOf

varungandhi-src:
The name Contains is also commonly used for checking if a
set contains an element in other languages, so it leads to
confusing auto-complete. Additionally, IsSupersetOf sounds
more expensive, which may be the case for this operation.

## Test plan

Covered by existing tests

## Changelog

### DIFF
--- a/internal/collections/set_test.go
+++ b/internal/collections/set_test.go
@@ -65,16 +65,16 @@ func TestSet(t *testing.T) {
 		require.Equal(t, []int{1}, s.Values())
 	})
 
-	t.Run("Contains returns true if set contains the other set", func(t *testing.T) {
-		require.True(t, a.Contains(NewSet(1, 2)))
-		require.True(t, a.Contains(NewSet(1, 2, 3)))
-		require.False(t, a.Contains(b))
+	t.Run("IsSupersetOf returns true if set contains the other set", func(t *testing.T) {
+		require.True(t, a.IsSupersetOf(NewSet(1, 2)))
+		require.True(t, a.IsSupersetOf(NewSet(1, 2, 3)))
+		require.False(t, a.IsSupersetOf(b))
 
 		// set always contains self
-		require.True(t, a.Contains(a))
+		require.True(t, a.IsSupersetOf(a))
 
 		// empty set is always contained
-		require.True(t, a.Contains(NewSet[int]()))
+		require.True(t, a.IsSupersetOf(NewSet[int]()))
 	})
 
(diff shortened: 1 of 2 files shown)


internal/collections/set_test.go
```diff
@@ -65,16 +65,16 @@ func TestSet(t *testing.T) {
...
-	t.Run("Contains returns true if set contains the other set", func(t *testing.T) {
-		require.True(t, a.Contains(NewSet(1, 2)))
-		require.True(t, a.Contains(NewSet(1, 2, 3)))
-		require.False(t, a.Contains(b))
+	t.Run("IsSupersetOf returns true if set contains the other set", func(t *testing.T) {
+		require.True(t, a.IsSupersetOf(NewSet(1, 2)))
+		require.True(t, a.IsSupersetOf(NewSet(1, 2, 3)))
+		require.False(t, a.IsSupersetOf(b))
 
 		// set always contains self
```
burmudar, line 73:
```suggestion
		// a set is a superset of itself
```

internal/collections/set_test.go
```diff
@@ -65,16 +65,16 @@ func TestSet(t *testing.T) {
...
-		require.True(t, a.Contains(a))
+		require.True(t, a.IsSupersetOf(a))
 
 		// empty set is always contained
```
burmudar, line 76:
```suggestion
		// a set is always the superset of an empty set
```

burmudar (approved):

unknwon:
not a native speaker... "A contains B" to me means `A >= B`, "A is superset of B" means `A > B`, which changes the semantic meaning to me. Or does "superset" also implies `>=` relation?

varungandhi-src:
In usual math terminology, "superset" includes the equals case, yes.

Most math textbooks use ⊇ for superset and ⊃ for strict superset (at least one element is different). Sometimes ⊃ is used for superset and ⊋ is used for strict superset. 

[Relevant section on Wikipedia](https://en.wikipedia.org/wiki/Subset#%E2%8A%82_and_%E2%8A%83_symbols)

unknwon:
> In usual math terminology, "superset" includes the equals case, yes.
> 
> Most math textbooks use ⊇ for superset and ⊃ for strict superset (at least one element is different). Sometimes ⊃ is used for superset and ⊋ is used for strict superset.
> 
> [Relevant section on Wikipedia](https://en.wikipedia.org/wiki/Subset#%E2%8A%82_and_%E2%8A%83_symbols)

TIL, thank you!

PR #63066, opened 2024-06-04
gitserver(fs): Remove caching of cloned state

eseliger:
I thought it might be nice to save on a few IOPS from the Stat calls to check if a repo is cloned by caching it for a bit. But turns out that the mutex locking and unlocking is more expensive than the stat operation.

We can come back here later and see if we can optimize it, we never had this in the past so this will be no regression.

Some background on mutex performance: https://stackoverflow.com/questions/57562606/why-does-sync-mutex-largely-drop-performance-when-goroutine-contention-is-more-t

Closes SRC-322

Test plan:

Without the cache, the tests still pass.

### DIFF
--- a/cmd/gitserver/internal/gitserverfs/fs.go
+++ b/cmd/gitserver/internal/gitserverfs/fs.go
@@ -6,7 +6,6 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
-	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sourcegraph/log"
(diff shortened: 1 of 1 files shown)


eseliger:
* **#63066** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/63066?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

ggilmore (approved):
You can try using lock striping as a strategy to reduce contention.

PR #63068, opened 2024-06-04
Reimplement license checks

pjlast:
This PR reintroduces the licensing checks that will block a customer's instance if an invalid license is used.

An instance gets blocked (i.e. only Site Admins can sign in and modify the site config) if the license has been:
1. Revoked
2. Already used on a different Site ID
3. Is expired

If a customer creates a new Sourcegraph instance with a new Site ID, they'll need a new license key as well.

If we receive irregular pings from the same Site ID + License key combination, we'll receive a Slack ping about it as that would require manual investigation.

One change made in this PR as well is to only alert when we receive irregular pings from the exact same Site ID + License key. Previously it only checked Site ID, but due to many customers creating a copy of their production instance for testing, this is not particularly reliable and leads to noise.

## Test plan

Unit tests updated, also lots of manual testing by starting a dotcom instance with `docker run` and doing license checks against that instance.


cmd/frontend/internal/dotcom/productsubscription/license_check_handler.go
```diff
@@ -198,28 +200,20 @@ func NewLicenseCheckHandler(db database.DB) http.Handler {
...
+			logEvent(ctx, db, EventNameAssigned, siteID, license.ID)
 		} else if !strings.EqualFold(*license.SiteID, siteID) {
 			logger.Warn("license being used with multiple site IDs", log.String("previousSiteID", *license.SiteID), log.String("licenseKeyID", license.ID), log.String("subscriptionID", license.ProductSubscriptionID))
 
 			replyWithJSON(w, http.StatusOK, licensing.LicenseCheckResponse{
-				// TODO: revert this to false again in the future, once most customers have a separate
-				// license key per instance
 				Data: &licensing.LicenseCheckResponseData{
-					IsValid: true,
+					IsValid: false,
```
eseliger, line 209: can we add an escape hatch here, for example an env var that will allow us to very quickly turn it off at any time if things misbehave?

pjlast: Ah yes I forgot about this. Will do 👍

cmd/frontend/internal/dotcom/productsubscription/license_check_handler.go
```diff
@@ -198,28 +201,24 @@ func NewLicenseCheckHandler(db database.DB) http.Handler {
...
 				return
 			}
-			logEvent(ctx, db, EventNameAssigned, siteID)
+			logEvent(ctx, db, EventNameAssigned, siteID, license.ID)
 		} else if !strings.EqualFold(*license.SiteID, siteID) {
 			logger.Warn("license being used with multiple site IDs", log.String("previousSiteID", *license.SiteID), log.String("licenseKeyID", license.ID), log.String("subscriptionID", license.ProductSubscriptionID))
 
+			flags := featureflag.FromContext(ctx)
+			// This feature flag allows us to temporarily allow conflicting Site IDs
+			conflictingSiteIDsValid := flags.GetBoolOr("markConflictingSiteIDsAsValid", false)
```
eseliger, line 210: off by default is what we want, correct? just double checking

eseliger (approved):

PR #63069, opened 2024-06-04
feat(sg): respect the context when executing interrupt hooks

burmudar:
During testing I found that sometimes some hooks would just hang and not complete. In this PR we execute all hooks within a timeout context. Ensuring we give _some_ time for hooks to execute but also making sure we eventually exit if some hook is misbehaving.

Additional changes:
- Global timeout for all hook execution is 2 seconds
- We hard exit after 5 intterupts instead of 2
- Hooks are split into two groups: sequential and concurrent. As per their names the hooks are executed differently depending how they were registered.

## Test plan
Tested locally

```
^C⚠️ Interrupt received, executing hook groups for graceful shutdown...
⚠️ Executing 16 'cleanup' hooks for graceful shutdown...
[   repo-updater] INFO repo-updater.repo-updater.grpcserver grpcserver/grpcserver.go:76 Shutting down gRPC server
[   repo-updater] INFO sync_worker workerutil/worker.go:252 Shutting down dequeue loop {"name": "repo_sync_worker", "reason": ""}
worker stopped due to context error: context canceled
gitserver-1 stopped due to context error: context canceled
searcher stopped due to context error: context canceled
gitserver-0 stopped due to context error: context canceled
blobstore stopped due to context error: context canceled
symbols stopped due to context error: context canceled
caddy stopped due to context error: context canceled
repo-updater stopped due to context error: context canceled
embeddings stopped due to context error: context canceled
frontend stopped due to context error: context canceled
zoekt-index-0 stopped due to context error: context canceled
syntax-highlighter stopped due to context error: context canceled
zoekt-web-1 stopped due to context error: context canceled
web stopped due to context error: context canceled
zoekt-web-0 stopped due to context error: context canceled
⚠️ Executing 6 'general' hooks for for graceful shutdown...
❌ failed to run zoekt-index-1.
stderr:
INFO server zoekt-sourcegraph-indexserver/main.go:1017 removing tmp dir {"tmpRoot": "/Users/william/.sourcegraph/zoekt/index-1/.indexserver.tmp"}
2024/06/04 09:15:03 updating index 6 github.com/sourcegraph/sourcegraph@HEAD=e55003da894490122546f876452f651aae65bb55 reason=content-mismatch
INFO server zoekt-sourcegraph-indexserver/main.go:432 updated index {"repo": "github.com/sourcegraph/sourcegraph", "id": 6, "branches": ["HEAD=e55003da894490122546f876452f651aae65bb55"], "duration": "19.21403925s"}
```

## Changelog
- Hard exit sg when 5 intterupt hooks are received
- Respect the context while executing interrupt hooks to ensure we still exit if some hook is misbehaving


eseliger:
> Ensuring we give some time for hooks to execute but also making sure we eventually exit if some hook is misbehaving.

Isn't that why the 2nd sigint from the user terminates immediately using SIGKILLs? 
Should we not trust the hooks for the regular exit mode? 

> We hard exit after 5 intterupts instead of 2

Why is that?

burmudar:
> > Ensuring we give some time for hooks to execute but also making sure we eventually exit if some hook is misbehaving.
> 
> Isn't that why the 2nd sigint from the user terminates immediately using SIGKILLs? Should we not trust the hooks for the regular exit mode?

No - unfortunately not. That would be the more correct way to do it but a hook currently has no knowledge what signal was sent nor how many times a signal has been sent. We execute the same hooks regardless how many signals were sent 😬 So currently we handle it higher up since 99% of the use cases are people smashing `ctrl-c` - once we see that you're probably pressing `ctrl-c` a bunch of times we take it to mean "for the love of god stop"

> 
> > We hard exit after 5 intterupts instead of 2
> 
> Why is that?
This is more as a failsafe. Technically now with do not need to do this since with the context _we will exit_ but if something goes wrong there then ... smashing `ctrl-c` will still make `sg` exit and not hang around

jamesmcnamara (approved):
I like the opt-in for concurrent execution!

dev/sg/interrupt/interrupt.go
```diff
@@ -1,62 +1,146 @@
...
 		<-interrupt
-		std.Out.WriteWarningf("Interrupt received, executing %d hooks for graceful shutdown...", len(hooks))
+		std.Out.WriteWarningf("Interrupt received, executing hook groups for graceful shutdown...")
 		closed = make(chan struct{})
 
 		// prevent additional hooks from registering once we've received an interrupt
 		mux.Lock()
 
 		go func() {
 			// If we receive a second interrupt, forcibly exit.
```
burmudar, line 92:
```suggestion
			// Count the interrupts and exit after 5
```

PR #63072, opened 2024-06-04
chore(codenav): Resolve repo and commit in common code

varungandhi-src:
This PR replaces https://github.com/sourcegraph/sourcegraph/pull/63057. We introduce
a new type which carries information about the repo, the deserialized cursor etc.

## Test plan

Added some property based tests to check that the function doesn't panic
on different kinds of inputs.

## Changelog


internal/codeintel/codenav/transport/graphql/root_resolver.go
```diff
@@ -2,6 +2,8 @@ package graphql
 
 import (
 	"context"
+	"encoding/base64"
+	"encoding/json"
```
kritzcreek, line 6: These are unused afaict

internal/codeintel/resolvers/codenav.go
```diff
@@ -331,6 +401,25 @@ type SymbolComparator struct {
...
 }
 
+func (c *SymbolComparator) ProvenancesForSCIPData() ForEachProvenance[bool] {
+	var out ForEachProvenance[bool]
+	if c == nil || c.Provenance.Equals == nil {
+		out.Precise = true
+		out.Syntactic = true
+		out.SearchBased = true
+	} else {
+		switch p := *c.Provenance.Equals; p {
```
kritzcreek, line 411:
```suggestion
		switch *c.Provenance.Equals {
```

internal/codeintel/resolvers/codenav.go
```diff
@@ -288,35 +297,96 @@ type UsagesForSymbolArgs struct {
...
+		sym := args.Symbol.Name.Equals
+		prov := args.Symbol.Provenance.Equals
+		if sym == nil || prov == nil {
+			return out, errors.New("name.equals and provenance.equals must be specified if SymbolComparator is provided")
+		}
+		if *sym == "" {
+			return out, errors.New("symbol name should be non-empty")
+		}
+		if _, err := scip.ParseSymbol(*sym); err != nil { // parsing is cheap, so don't keep the parsed value
+			return out, errors.Wrap(err, "malformed symbol name")
```
kritzcreek, line 347: If we're parsing the symbol here, I'd say we should put it into the resolved args as well.

kritzcreek (approved):
Very, very nice. Thank you for doing this.

internal/codeintel/codenav/transport/graphql/root_resolver.go
varungandhi-src: They're used for unmarshaling the cursor.

internal/codeintel/resolvers/codenav.go
varungandhi-src: Fixed with latest commit.

internal/codeintel/resolvers/codenav.go
varungandhi-src: Fixed in latest commit.

internal/codeintel/resolvers/codenav.go
```diff
@@ -542,6 +542,7 @@ func resolveWithCache(ctx context.Context, repoStore database.RepoStore, f *Usag
 	}
 	var repoFilter *ResolvedRepositoryFilter
 	if f.Repository != nil && f.Repository.Name.Equals != nil {
+		repoFilter = &ResolvedRepositoryFilter{}
```
varungandhi-src, line 545: This missing initialization before setting a field later was caught by `rapid`. 🎉

internal/codeintel/resolvers/codenav_test.go
```diff
@@ -173,14 +189,19 @@ func TestResolve(t *testing.T) {
 
 	rapid.Check(t, func(t *rapid.T) {
 		args := generator.Draw(t, "args")
+		// NOTE: For some reason, moving the p90Bool.Draw operations inside the hook
+		// functions triggers a bug inside rapid with a buffer overrun or running
+		// out of entropy bits.
```
varungandhi-src, line 194: Not exactly sure what this is, I'll try to minimize it later.

internal/pbt/probability_test.go
```diff
@@ -0,0 +1,25 @@
...
+		hasFalse := false
+		for _, b := range bools {
+			hasTrue = hasTrue || b
+			hasFalse = hasFalse || !b
+			if hasTrue && hasFalse {
+				break
+			}
+		}
+		require.True(t, hasTrue, "failure probability should be 1/2^256")
+		require.True(t, hasFalse, "failure probability should be 1/2^256")
```
varungandhi-src, line 23: @kritzcreek I thought we should have a basic test for this generator, not sure how you feel about this.

internal/codeintel/resolvers/codenav.go
```diff
@@ -395,6 +532,51 @@ func (f *UsagesFilter) DebugString() string {
 	return strings.Join(result, " and ")
 }
 
+func (f *UsagesFilter) Resolve(ctx context.Context, repoStore database.RepoStore) (*ResolvedUsagesFilter, error) {
+	return resolveWithCache(ctx, repoStore, f, map[string]*types.Repo{})
+}
+
+func resolveWithCache(ctx context.Context, repoStore database.RepoStore, f *UsagesFilter, cache map[string]*types.Repo) (*ResolvedUsagesFilter, error) {
```
varungandhi-src, line 539: I know this might seem a bit overkill, but I figured since we need the recursive traversal anyway, adding a cache is not that much extra work.

internal/codeintel/codenav/transport/graphql/root_resolver.go
varungandhi-src: Oops, I misspoke, that's what failed in CI. I swear I was using this earlier, maybe I moved the logic to another file...

internal/codeintel/resolvers/codenav.go
kritzcreek: Yeah, adding the cache seems fine even if we can't produce a filter with more than one repository reference for now. Finally I can filter my usages for `(not (not (not (not "github.com/sourcegraph/sourcegraph"))))`

kritzcreek (approved):
Code looks good in general. Added a few smaller nits, but nothing that should block you from merging once you've fixed the entropy crash

internal/codeintel/codenav/transport/graphql/root_resolver.go
```diff
@@ -205,9 +205,12 @@ func (r *rootResolver) UsagesForSymbol(ctx context.Context, args *resolverstubs.
 	}()
 
 	const maxUsagesCount = 100
-	args.Normalize(maxUsagesCount)
+	resolvedArgs, err := args.Resolve(ctx, r.repoStore, r.gitserverClient, maxUsagesCount)
```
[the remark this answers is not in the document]
varungandhi-src: The scoping in Go doesn't match Rust/ML, so that would be an error since the type would change. It would require introducing another function. However, I can change the variable names so that this is `args` and the other one is `unresolvedArgs`, so the latter becomes more inconvenient to type.

internal/pbt/probability_test.go
varungandhi-src: I removed this and added a Pearson's Chi-Squared test for `WithProbabilities` instead.

internal/pbt/probability.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+const epsilon = 0.00000001
+
+// WithProbabilities runs the generators based on the
+// provided fixed probability distribution. The sum
+// of probabilities should be 1.0.
+func WithProbabilities[T any](possibilities []GeneratorChoice[T]) *rapid.Generator[T] {
+	total := 0.0
+	cumsums := []float64{}
```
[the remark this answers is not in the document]
varungandhi-src: I think this is standard terminology.

https://numpy.org/doc/stable/reference/generated/numpy.cumsum.html
https://pkg.go.dev/gonum.org/v1/gonum/floats#CumSum

But I've renamed this.

internal/pbt/probability.go
```diff
@@ -0,0 +1,49 @@
...
+	for i, p := range possibilities {
+		if p.Chance < 0 {
```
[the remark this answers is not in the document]
varungandhi-src: I'm guessing this is for the `-0.0` case? It doesn't matter if it is `-0.0` here.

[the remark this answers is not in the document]
kritzcreek: Yes, probably doesn't matter here. Just thought it might be an interesting bit of trivia

PR #63074, opened 2024-06-04
SSC proxy: return errors with message and cause

taras-yemets:
See also https://github.com/sourcegraph/self-serve-cody/pull/834
TODO

## Test plan
TODO


cmd/frontend/internal/ssc/ssc_proxy.go
```diff
@@ -303,13 +304,23 @@ func (p *APIProxyHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, toErrorWithMessageAndCause(bodyBytes, http.StatusBadGateway), http.StatusBadGateway)
+		return
+	}
+
+	// For any SSC 4xx errors, ensure the response body contains error of proper shape.
+	if proxyResponse.StatusCode >= 400 {
+		p.Logger.Debug(
+			"received 4xx response from SSC backend",
+			log.String("responseBody", string(bodyBytes)))
+		http.Error(w, toErrorWithMessageAndCause(bodyBytes, proxyResponse.StatusCode), proxyResponse.StatusCode)
```
chrsmith, line 323: Nit: `http.Error(w, toErrorWithMessageAndCause(bodyBytes, proxyResponse.StatusCode), proxyResponse.StatusCode)` seems kinda clunky.

Why not just create a higher-level function like:

```go
func writeJSONErrorResponse(w responeWriter, statusCode int, cause, message string) {
    ...
}
```

Or, keep it as-is, but just rename it to be a little clearer. Like:

```go
errRespBody := renderErrorJSONResponse(...)
```

chrsmith (commented):
Some hottest of hot-takes, but I'll leave some more notes on Slack. (Like you mentioned there, we might not want to put this code here, and just pass through the JSON-error response directly from the SSC backend.)

PR #63075, opened 2024-06-04
fix(appliance): use version for image tags

craigfurman:
Simultaneously fix a bug in the (still unreleased) appliance and change approach to container image metadata. The appliance used to maintain a map of versions to image-tags. The only version hardcoded for the still-in-development appliance was 5.3.9104, since that was the latest at the time development started. These versions were set to 5.3.2, which is incorrect. These versions were obtained from looking at the main branch of deploy-sourcegraph-helm, instead of the relevant release branch.

The change in approach is to stop including the optional sha256 digest of the image in the appliance-generated config. We considered several approaches in
https://linear.app/sourcegraph/issue/REL-22/spike-sketch-out-how-appliance-developers-will-handle-upgrades, please see that issue for more context.

It appears simpler to statelessly determine the image tag (as being equal to the release version string) rather than add moving parts such as calling the release registry, or building a 2-phase release system to bake image checksums into the appliance once they are known post-build. These options introduce complexity and greater possibility of errors than just asking for a version.

Closes
https://linear.app/sourcegraph/issue/REL-135/appliance-knows-what-images-to-pull-for-a-given-release.

## Test plan

Golden tests included.

## Changelog

- Appliance uses requested version as image tag, instead of requiring sha256 digests.

### DIFF
--- a/internal/appliance/config/defaults.go
+++ b/internal/appliance/config/defaults.go
@@ -3,7 +3,6 @@ package config
 import (
 	"fmt"
 
-	"github.com/sourcegraph/sourcegraph/lib/errors"
 	"github.com/sourcegraph/sourcegraph/lib/pointers"
 )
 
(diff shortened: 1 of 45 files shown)


internal/appliance/config/defaults.go
```diff
@@ -148,41 +147,6 @@ func NewDefaultConfig() Sourcegraph {
...
-	if !ok {
-		return "", errors.Newf("no default images found for version %s", sg.Spec.RequestedVersion)
-	}
-	image, ok := images[component]
-	if !ok {
-		return "", errors.Newf("no default image found for service %s", component)
-	}
-	return fmt.Sprintf("%s/%s", sg.Spec.ImageRepository, image), nil
+func GetDefaultImage(sg *Sourcegraph, component string) string {
+	return fmt.Sprintf("%s/%s:%s", sg.Spec.ImageRepository, component, sg.Spec.RequestedVersion)
```
craigfurman, line 151: Simple and stateless! This is the core of the proposal. Image metadata would need no developer input to enable upgrades.

internal/appliance/reconciler/testdata/golden-fixtures/cadvisor/default.yaml
```diff
@@ -45,7 +45,7 @@ resources:
             - args:
                 - --store_container_labels=false
                 - --whitelisted_container_labels=io.kubernetes.container.name,io.kubernetes.pod.name,io.kubernetes.pod.namespace,io.kubernetes.pod.uid
-              image: index.docker.io/sourcegraph/cadvisor:5.3.2@sha256:3860cce1f7ef0278c0d785f66baf69dd2bece19610a2fd6eaa54c03095f2f105
+              image: index.docker.io/sourcegraph/cadvisor:5.3.9104
```
craigfurman, line 48: All of the changes should look like this: the correct patch version being requested, with no SHA. I've read the diff, but it's error-prone so another pair of eyes would be greatly appreciated, even though this is extremely boring 🙏

internal/appliance/config/defaults.go
jdpleiness: This is nice! I'm still a little worried, but I can't explain why so it's probably just FOC 😆 
Either way if we encounter issues with this, it's an easy change 🚀

internal/appliance/reconciler/testdata/golden-fixtures/cadvisor/default.yaml
jdpleiness: Nothing stands out to me either.

jdpleiness (approved):

internal/appliance/config/defaults.go
craigfurman: Oh! What causes the worry? 👀 
Worth hearing it out before merge!

PR #63080, opened 2024-06-04
Svelte: fix editor button styling

camdencheek:
The editor button would wrap before the div would overflow, causing weird wrapping. This fixes that and also makes it a proper button.

Before:
screenshot-2024-06-04_10-23-29@2x

After:
screenshot-2024-06-04_10-24-22@2x

## Test plan

Manual testing + screenshots above.

### DIFF
--- a/client/web-sveltekit/src/lib/repo/open-in-editor/OpenInEditor.svelte
+++ b/client/web-sveltekit/src/lib/repo/open-in-editor/OpenInEditor.svelte
@@ -81,7 +81,7 @@
         {#if editor}
             <Tooltip tooltip={`Open in ${editor.name}`}>
                 <a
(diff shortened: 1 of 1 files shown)


client/web-sveltekit/src/lib/repo/open-in-editor/OpenInEditor.svelte
```diff
@@ -180,10 +183,6 @@
         background-color: var(--color-bg-1);
     }
 
-    .form-label {
-        font-weight: 500;
-    }
-
```
camdencheek, line 186: Unused. Just removing it while I'm in here.

fkling (approved):

PR #63081, opened 2024-06-04
ensure we're pulling the config correctly

chenkc805:
We're pulling in `BLOCKED_PROMPT_PATTERNS` but the config is actually `CODY_GATEWAY_ANTHROPIC_BLOCKED_PROMPT_PATTERNS`, so it it needs `CODY_GATEWAY_ANTHROPIC_` appended to the beginning of the string

### DIFF
--- a/cmd/cody-gateway/shared/config/config.go
+++ b/cmd/cody-gateway/shared/config/config.go
@@ -360,8 +360,8 @@ func (c *Config) loadFlaggingConfig(cfg *FlaggingConfig, envVarPrefix string) {
 	cfg.MaxTokensToSample = c.GetInt(envVarPrefix+"MAX_TOKENS_TO_SAMPLE", "4000", "Maximum permitted value of maxTokensToSample")
 	cfg.MaxTokensToSampleFlaggingLimit = c.GetInt(envVarPrefix+"MAX_TOKENS_TO_SAMPLE_FLAGGING_LIMIT", "4000", "Maximum value of max_tokens_to_sample to allow without flagging.")
 
-	cfg.AllowedPromptPatterns = maybeLoadLowercaseSlice("ALLOWED_PROMPT_PATTERNS", "Allowed prompt patterns")
-	cfg.BlockedPromptPatterns = maybeLoadLowercaseSlice("BLOCKED_PROMPT_PATTERNS", "Patterns to block in prompt.")
+	cfg.AllowedPromptPatterns = maybeLoadLowercaseSlice(envVarPrefix+"ALLOWED_PROMPT_PATTERNS", "Allowed prompt patterns")
+	cfg.BlockedPromptPatterns = maybeLoadLowercaseSlice(envVarPrefix+"BLOCKED_PROMPT_PATTERNS", "Patterns to block in prompt.")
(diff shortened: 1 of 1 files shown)


cmd/cody-gateway/shared/config/config.go
```diff
@@ -360,8 +360,8 @@ func (c *Config) loadFlaggingConfig(cfg *FlaggingConfig, envVarPrefix string) {
 	cfg.MaxTokensToSample = c.GetInt(envVarPrefix+"MAX_TOKENS_TO_SAMPLE", "4000", "Maximum permitted value of maxTokensToSample")
 	cfg.MaxTokensToSampleFlaggingLimit = c.GetInt(envVarPrefix+"MAX_TOKENS_TO_SAMPLE_FLAGGING_LIMIT", "4000", "Maximum value of max_tokens_to_sample to allow without flagging.")
 
-	cfg.AllowedPromptPatterns = maybeLoadLowercaseSlice("ALLOWED_PROMPT_PATTERNS", "Allowed prompt patterns")
-	cfg.BlockedPromptPatterns = maybeLoadLowercaseSlice("BLOCKED_PROMPT_PATTERNS", "Patterns to block in prompt.")
+	cfg.AllowedPromptPatterns = maybeLoadLowercaseSlice(envVarPrefix+"ALLOWED_PROMPT_PATTERNS", "Allowed prompt patterns")
+	cfg.BlockedPromptPatterns = maybeLoadLowercaseSlice(envVarPrefix+"BLOCKED_PROMPT_PATTERNS", "Patterns to block in prompt.")
```
rafax, line 364: note that `envVarPrefix` is prepended on line 355, so this PR would add the prefix twice

chenkc805: Ah yeah thanks. I'm getting lost in lines 354-358, especially with this function call `value := c.GetOptional(envVarPrefix+envVar, description)` so I'm stuck here

chrsmith: Good catch @rafax! @chenkc805 to make this code easier to grok, why don't you do the following:

1. Update `maybeLoadLowercaseSlice` to _no longer_ prefix the supplied env var with `envVarPrefix`.
2. Update all callers of `maybeLoadLowercaseSlide` to add `envVarPrefix+"..."`. (i.e. have the code not "look wrong, but actually be correct".)
3. Update the comment on `AbuseConfig.BlockedPromptPatterns` AND in `flagging.go`'s `abuseConfig.BlockedPromptPatterns` to call out that the string slice is assumed to have already been converted into lower-case. (Since that's what `maybeLoadLowercaseSlice` is doing, but when looking at the code, it's an easy detail to miss!)

chrsmith (commented):
Thanks for the comment @rafax . It looks like the code is "entirely correct", but I would suggest you make some minor refactoring @chenkc805 and merge it. So that this code is a little easier to understand the next time we need to go tinkering with it. (Since clearly it's easy to overlook some subtle details!)

cmd/cody-gateway/shared/config/config.go
chenkc805: Ok did 1 and 2.

For 3, doesn't `maybeLoadLowercaseSlice` already lowercase the input, so we don't have to lowercase the input in the config file. Anyway, updated the comments in 3 to make it a bit clearer.

PR #63082, opened 2024-06-04
Svelte: collapse path when too long

camdencheek:
This applies the `sizeToFit` action to the file header, allowing us to collapse the header in multiple steps:
- First, collapse the actions down to icons
- Next, collapse elements one by one from the front of the path and put them in an overflow dropdown

## Test plan

Mostly manual testing. [Video demo](https://share.cleanshot.com/StdT6Gw9). Specifically tested that jumps in screen size work as expected.


client/web-sveltekit/src/lib/dom.ts
```diff
@@ -327,90 +327,6 @@ export const portal: Action<HTMLElement, { container?: HTMLElement | null } | un
...
- *
- * Using this action should not cause any content flashes because the calculation is done synchronously
- * before painting.
- *
- * @param target The element to measure.
- * @param className The CSS class to add when the content overflows.
- * @param measureClass The CSS class to apply for measuring the size of the content.
- * @returns An action that updates the overflow state of the element.
- */
-export const overflow: Action<HTMLElement, { class: string; measureClass?: string }> = (
```
camdencheek, line 352: This is no longer used. Since I _think_ the `sizeToFit` action is more general (grow/shrink can apply a class name), I went ahead and deleted it.

fkling (approved):
This is amazing :) It could make sense to keep showing a prefix of the path, but we can always iterate on that.

camdencheek:
> This is amazing :)

Thank you 🙂 I was pleased with how well this worked out

> It could make sense to keep showing a prefix of the path

Yeah, agreed. The logic got more annoying that way though, so I figured I should start simple and easy to review.

PR #63084, opened 2024-06-04
fix/msp: flatten custom alert promQL query for GCP

bobheadxi:
The GCP monitoring alert configuration expects, for some reason, a single-line PromQL query only, otherwise the threshold doesn't work. In configuration, however, we may want to write a multi-line query, for ease of readability. This change automatically flattens the PromQL query into a single line and strips extra spaces.

Part of CORE-161

## Test plan

Unit tests


jac (approved):
oh good catch

looking back I tested with single line for promql and multiline for mql 😠

bobheadxi:
> looking back I tested with single line for promql and multiline for mql 😠

To be fair, I think this is a GCP bug 😆

PR #63088, opened 2024-06-04
Svelte: inject into primary `app.html`

camdencheek:
This modifies our pattern for shipping a svelte-enabled HTML document by injecting the svelte-specific things into our primary `app.html` rather than trying to make `app.prod.html` match the behavior of the frontend `app.html`.

I could not figure out a "blessed" way to integrate this into the build system, so I went for the hacky way: render `app.prod.html` in a format that's easily parseable, then parse it in `frontend` and inject it into our `app.html` with the standard go template patterns.

Fixes SRCH-76

## Test plan

Manual testing with `sg start enterprise` and `sg start enterprise-sveltekit`. Manual inspection of the generated HTML and diffing against the previously generated HTML.

### DIFF
--- a/client/web-sveltekit/src/app.prod.html
+++ b/client/web-sveltekit/src/app.prod.html
@@ -1,44 +1,9 @@
-<!DOCTYPE html>
-<html lang="en">
-    <head>
-        <meta charset="utf-8" />
-        <meta http-equiv="x-ua-compatible" content="ie=edge" />
-        <meta http-equiv="Content-Language" content="en" />
-        <meta name="google" content="notranslate" />
-        <meta name="viewport" content="width=device-width, viewport-fit=cover" />
-        <meta name="referrer" content="origin-when-cross-origin" />
-        <meta name="color-scheme" content="light dark" />
-        <title>{{.Title}}</title>
-
-        <script ignore-csp>
-            window.context = {{ .Context }}
-            window.pageError = {{ .Error }}
-        </script>
-
-        %sveltekit.head%
-
-        <script>
-            const userPreference = localStorage.getItem('sourcegraph-theme') || 'System'
-            const theme =
-                (userPreference === 'System' && window.matchMedia('(prefers-color-scheme: dark)').matches) ||
-                userPreference === 'Dark'
-                    ? 'theme-dark'
-                    : 'theme-light'
-            document.documentElement.classList.add(theme)
-        </script>
-    </head>
-    <body data-sveltekit-preload-data data-sveltekit-preload-code="hover">
-        <div style="display: contents">%sveltekit.body%</div>
-        <noscript>
-            <p>
-                Sourcegraph is a web-based code search and navigation tool for dev teams. Search, navigate, and review
-                code. Find answers.
-            </p>
-
-            <br />
-            <br />
-            <br />
-            You need to enable JavaScript to run this app.
-        </noscript>
-    </body>
-</html>
+<!--
+    HACK: Intentionally minimal because this is parsed in frontend which does
+    its own templating. If you change this file, ensure that you also update
+    cmd/frontend/internal/app/ui/sveltekit/sveltekit.go:loadSvelteKitInjections
+-->
+<!-- SPLIT -->
+%sveltekit.head%
+<!-- SPLIT -->
+%sveltekit.body%
--- a/client/web/dev/esbuild/server.ts
+++ b/client/web/dev/esbuild/server.ts
@@ -38,21 +38,6 @@ export const esbuildDevelopmentServer = async (
     // requests go to the upstream.
     const proxyApp = express()
 
-    if (process.env.SVELTEKIT) {
-        // Proxy requests for SvelteKit's assets to the server, not the esbuild server.
-        proxyApp.use(
-            `${assetPathPrefix}/_sk`,
-            createProxyMiddleware({
-                target: {
-                    protocol: 'http:',
-                    host: DEV_SERVER_PROXY_TARGET_ADDR.host,
-                    port: DEV_SERVER_PROXY_TARGET_ADDR.port,
-                },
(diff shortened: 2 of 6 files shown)


client/web-sveltekit/src/app.prod.html
```diff
@@ -1,44 +1,9 @@
...
-            <br />
-            You need to enable JavaScript to run this app.
-        </noscript>
-    </body>
-</html>
+<!--
+    HACK: Intentionally minimal because this is parsed in frontend which does
+    its own templating. If you change this file, ensure that you also update
+    cmd/frontend/internal/app/ui/sveltekit/sveltekit.go:loadSvelteKitInjections
+-->
```
camdencheek, line 5: Instead of trying to reproduce everything that's in the frontend `app.html`, instead we can build this to be easily parseable and inject it into `app.html` using Go's templating that we already use heavily.

client/web-sveltekit/src/app.prod.html
```diff
@@ -1,44 +1,9 @@
...
+<!-- SPLIT -->
+%sveltekit.head%
+<!-- SPLIT -->
+%sveltekit.body%
```
camdencheek, line 9: Instead of trying to reproduce everything that's in the frontend app.html, instead we can build this to be easily parseable and inject it into app.html using Go's templating that we already use heavily.

camdencheek: Related: this PR is an alternative to my [earlier draft](https://github.com/sourcegraph/sourcegraph/pull/63083) in which I manually reconciled the two templates. I think this is a generally better approach since we don't need to try to keep those files in sync.

sg.config.yaml
```diff
@@ -1763,37 +1763,6 @@ commandsets:
     bazelCommands:
       - cody-gateway
 
-  enterprise-bazel-sveltekit:
```
camdencheek, line 1766: Now that injecting the right application logic is fully the responsibility of `frontend`, we don't need these sets anymore. `sg start enterprise` works just fine with the feature flags and switching back and forth.

cmd/frontend/internal/app/ui/app.html
```diff
@@ -28,7 +28,22 @@
...
+    {{.Svelte.Head}}
+    <script ignore-csp>
+        const userPreference = localStorage.getItem('sourcegraph-theme') || 'System'
+        const theme =
+            (userPreference === 'System' && window.matchMedia('(prefers-color-scheme: dark)').matches) ||
+            userPreference === 'Dark'
+                ? 'theme-dark'
+                : 'theme-light'
+        document.documentElement.classList.add(theme)
+    </script>
```
camdencheek, line 42: I don't quite understand why this is needed -- couldn't this just be part of the root layout?

client/web/dev/esbuild/server.ts
```diff
@@ -38,21 +38,6 @@ export const esbuildDevelopmentServer = async (
...
-            createProxyMiddleware({
-                target: {
-                    protocol: 'http:',
-                    host: DEV_SERVER_PROXY_TARGET_ADDR.host,
-                    port: DEV_SERVER_PROXY_TARGET_ADDR.port,
-                },
-                logLevel: 'error',
-            })
-        )
-    }
```
camdencheek, line 54: This seemed to only be used by `sg start enterprise-sveltekit`, which no longer exists now that we use the same `app.html` for everything. I think this could still be used to run a dev server with a local instance, but I think it's probably more intuitive to just run `pnpm dev` with an overriden `SOURCEGRAPH_API_URL`. Am I off base here?

cmd/frontend/internal/app/ui/app.html
fkling: This, together with same basic CSS which doesn't exist yet, would ensure that there is no flash of white when the user has explicitly selected dark mode.
Without this we will only set the theme preference after all code and data is loaded an run.

cmd/frontend/internal/app/ui/handlers.go
```diff
@@ -327,10 +348,6 @@ func serveBasicPage(db database.DB, title func(c *Common, r *http.Request) strin
 		}
 		common.Title = title(common, r)
 
-		if sveltekit.Enabled(r.Context()) {
-			return sveltekit.RenderTemplate(w, common)
-		}
-
```
fkling, line 333: This makes me very happy. I always thought it would be nice to eventually do the switch in `renderTemplate` so that it "just works" everywhere.

sg.config.yaml
fkling: I'm not sure that's true. The SvelteKit app still has to be built. If you've built them before then they might or might not be already in the system, but they wouldn't be updated if you e.g. make changes (which doesn't work reliably with these commands either tbf).
That's really the main difference with these commands: They additionally start the build process for the SvelteKit web app. How the HTML file is generated is irrelevant.

client/web/dev/esbuild/server.ts
fkling: See my comment below. I believe to remember that I added this separately because otherwise the request would be handled by the esbuild web server, which will trigger a rebuild of the react app (unnecessarily) which is rather slow (comparatively).

sg.config.yaml
camdencheek: > (which doesn't work reliably with these commands either tbf).

Ah, gotcha. I was always under the impression that I had to build for these commands to work. Okay, I'll roll back these changes

fkling: It might be fine to remove them. I don't think the rebuilding actually works. The process often terminates. In other words it doesn't really work reliably.

camdencheek: Oh, okay, unreverting then 🙂

camdencheek:
@fkling were there any followups you were looking for from your comments?

PR #63093, opened 2024-06-05
fix(sg): fix cloud eph suggested commands 

burmudar:
- suggested commands didn't have the `ephemeral` subcommand
- update expiry time to show in duration till expiry

## Test plan
Tested locally + CI

## Changelog
- fix suggested cloud ephemeral commands
- show duration till expiry for cloud ephemeral


filiphaftek (approved):
👍

PR #63097, opened 2024-06-05
Syntactic Indexing: add TAR archive indexing mode to scip-syntax CLI

keynmol:
Fixes GRAPH-651
Fixes GRAPH-650

New features:
- `index tar <input>` allows indexing tar archives when `input` is a file, and `stdin` when `input` is literal string `-`
- **BREAKING: Instead of flags controlling the type of input source, we now have subcommands: `index files`, `index workspace`, `index tar`**

Refactoring:
- Tests were improved, with helper functions broken down into composable pieces, and producing 1 snapshot per test, making it easier to manage
- Closures were removed from the indexing code, instead replaced with functions.
- Calls to `.unwrap` were replaced with better error handling code
- Path canonicalisation was replaced with path cleanup + absolutisation - to avoid following symlinks (which is what `canonicalize` would do)

Most of the refactoring was triggered by the changes required to add more tests.

## Test plan

- New integration tests

### DIFF
--- a/docker-images/syntax-highlighter/crates/scip-syntax/BUILD.bazel
+++ b/docker-images/syntax-highlighter/crates/scip-syntax/BUILD.bazel
@@ -94,6 +94,7 @@ rust_test(
     tags = [TAG_PLATFORM_GRAPH],
     deps = all_crate_deps(
         normal = True,
+        normal_dev = True,
     ) + [
         ":scip-syntax",
         ":scip_syntax_lib",
--- a/docker-images/syntax-highlighter/crates/scip-syntax/src/main.rs
+++ b/docker-images/syntax-highlighter/crates/scip-syntax/src/main.rs
@@ -1,6 +1,7 @@
+use std::{path::PathBuf, process};
+
 use clap::{Parser, Subcommand};
 use scip_syntax::index::{index_command, AnalysisMode, IndexMode, IndexOptions};
-use std::{path::PathBuf, process};
 
 #[derive(Parser)]
 #[command(author, version, about, long_about = None)]
--- a/docker-images/syntax-highlighter/crates/scip-syntax/src/main.rs
+++ b/docker-images/syntax-highlighter/crates/scip-syntax/src/main.rs
@@ -1,7 +1,7 @@
 use std::{path::PathBuf, process};
 
 use clap::{Parser, Subcommand};
-use scip_syntax::index::{index_command, AnalysisMode, IndexMode, IndexOptions};
+use scip_syntax::index::{index_command, AnalysisMode, IndexMode, IndexOptions, TarMode};
 
 #[derive(Parser)]
 #[command(author, version, about, long_about = None)]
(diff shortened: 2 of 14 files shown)


docker-images/syntax-highlighter/crates/scip-syntax/BUILD.bazel
```diff
@@ -94,6 +94,7 @@ rust_test(
     tags = [TAG_PLATFORM_GRAPH],
     deps = all_crate_deps(
         normal = True,
+        normal_dev = True,
```
keynmol, line 97: this allows bringing in `dev-dependencies` from Cargo

docker-images/syntax-highlighter/crates/scip-syntax/src/main.rs
```diff
@@ -108,26 +112,39 @@ pub fn main() -> anyhow::Result<()> {
...
+                match tar {
+                    Some(loc) if loc == "-" => IndexMode::TarArchive {
+                        input: scip_syntax::index::TarMode::Stdin,
+                    },
+                    Some(loc) => IndexMode::TarArchive {
+                        input: scip_syntax::index::TarMode::File {
+                            location: PathBuf::from(loc),
+                        },
+                    },
+                    None => match workspace {
```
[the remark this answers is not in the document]
keynmol: Yeah, I think it's about time for this change.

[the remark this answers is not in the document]
keynmol: Done in https://github.com/sourcegraph/sourcegraph/pull/63097/commits/20ebd109be0d56c4166cef57c64d33d5cfd80592

docker-images/syntax-highlighter/crates/scip-syntax/src/main.rs
```diff
@@ -11,48 +11,79 @@ struct Cli {
...
+    /// Index a folder, automatically detecting files
+    /// to be processed by the chosen language
+    Workspace {
         /// Folder to index - will be chosen as project root,
         /// and files will be discovered according to
         /// configured extensions for the selected language
+        /// Has to be absolute path.
         #[arg(long)]
-        workspace: Option<String>,
+        dir: String,
```
kritzcreek, line 51: Could make this a positional argument now

```
scip-syntax index workspace --language java /Users/creek/work/sourcegraph
# instead of
scip-syntax index workspace --language java --dir /Users/creek/work/sourcegraph
```

docker-images/syntax-highlighter/crates/scip-syntax/src/main.rs
```diff
@@ -103,62 +134,44 @@ pub fn main() -> anyhow::Result<()> {
...
-                                IndexMode::Workspace {
-                                    location: location.into(),
-                                }
-                            }
-                        }
+        Commands::Index(index1) => {
+            let result = match index1.index_command {
+                IndexCommand::Files { filenames, options } => {
+                    if filenames.is_empty() {
+                        eprintln!("either specify --workspace or provide a list of files");
```
kritzcreek, line 141: update error for new command

kritzcreek (changes requested):

kritzcreek (approved):
Don't want to block the merge after updating the error message for `--workspace`

LGTM

PR #63099, opened 2024-06-05
Initialise authProviders array in Go code

pjlast:
When a Sourcegraph license is expired or revoked and the instance reverts to free license, then `authz` (for permissions syncing) stops working as well. This means that there are no AuthProviders set up and the front-end code has `null` for these values.

The issue arrises from unitialised arrays in Go being marshalled as `null` instead of `[]`. This PR initialises the array explicitly, instead of relying on the empty value.

## Test plan

Tested locally with a revoked license.

### DIFF
--- a/cmd/frontend/internal/app/jscontext/jscontext.go
+++ b/cmd/frontend/internal/app/jscontext/jscontext.go
@@ -297,7 +297,7 @@ func NewJSContextFromRequest(req *http.Request, db database.DB) JSContext {
 	needsSiteInit := err == nil && !siteInitialized
 
 	// Auth providers
-	var authProviders []authProviderInfo
+	authProviders := []authProviderInfo{}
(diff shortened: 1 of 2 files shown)


eseliger (approved):

cmd/frontend/internal/app/jscontext/jscontext.go
```diff
@@ -297,7 +297,7 @@ func NewJSContextFromRequest(req *http.Request, db database.DB) JSContext {
 	needsSiteInit := err == nil && !siteInitialized
 
 	// Auth providers
-	var authProviders []authProviderInfo
+	authProviders := []authProviderInfo{}
```
eseliger, line 300: Can you leave a comment here that this is a load-bearing allocation? 😅

eseliger (approved):

PR #63100, opened 2024-06-05
Add Support for Counting Tokens for Azure Code and Update in Redis

arafatkatze:
Description:

This PR introduces support for counting tokens within the Azure code and updating these counts in Redis. The token counting logic is embedded directly in the Azure code rather than using a standardized point for all token counting logic.

Reasoning:

	•	Azure does not currently support obtaining token usage from their streaming endpoint, unlike OpenAI.
	•	To enable immediate functionality, the token counting logic is placed within the Azure code itself.
	•	The implementation supports GPT-4o.

Future Considerations:

	•	When Azure eventually adds support for token usage from the streaming endpoint, we will migrate to using Azure’s built-in capabilities.
	•	This will ensure full utilization of Azure OpenAI features as they achieve parity with OpenAI.

Changes:

	•	Added token counting logic to the Azure code.
	•	Updated Redis with the token counts.

Testing:

	•	Verified the implementation works with GPT-4o.

Conclusion:

This is a temporary solution to enable token counting in Azure. We will adapt our approach as Azure enhances its feature set to include token usage from their streaming endpoint.

## Test plan
Tested locally with debugger

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -289,8 +289,8 @@ require (
 	github.com/mroth/weightedrand/v2 v2.0.1
 	github.com/nxadm/tail v1.4.11
 	github.com/oschwald/maxminddb-golang v1.12.0
-	github.com/pkoukk/tiktoken-go v0.1.6
-	github.com/pkoukk/tiktoken-go-loader v0.0.1
+	github.com/pkoukk/tiktoken-go v0.1.7
+	github.com/pkoukk/tiktoken-go-loader v0.0.2-0.20240522064338-c17e8bc0f699
 	github.com/prometheus/statsd_exporter v0.22.7
 	github.com/redis/go-redis/extra/redisotel/v9 v9.0.5
 	github.com/redis/go-redis/v9 v9.5.3
--- a/internal/completions/client/azureopenai/openai.go
+++ b/internal/completions/client/azureopenai/openai.go
@@ -8,12 +8,15 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"strings"
 	"sync"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/ai/azopenai"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/pkoukk/tiktoken-go"
+	tiktoken_loader "github.com/pkoukk/tiktoken-go-loader"
 	"golang.org/x/net/http2"
 
 	"github.com/sourcegraph/sourcegraph/internal/completions/tokenizer"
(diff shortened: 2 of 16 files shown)


schema/site.schema.json
```diff
@@ -2842,6 +2842,38 @@
...
+            "gpt-3.5-turbo-16k-0613",
+            "gpt-4-0314",
+            "gpt-4-32k-0314",
+            "gpt-4-0613",
+            "gpt-4-32k-0613",
+            "gpt-4o",
+            "gpt-4"
+          ]
+        },
+        "azureCompletionModel": {
```
arafatkatze, line 2861: I hate adding extra parameters to the siteconfig and requestparams. However, I had to add these parameters because Azure does not support naming the model directly. Instead, you provide the deployment name. The deployment name can be different from the model name. Because of this discrepancy, I had to introduce extra config parameters to specify the chat model name and the completion model name.

This necessity arises because Azure doesn’t allow retrieval of the model name straightforwardly. The deployment name, which is used in requests, can map to any underlying model. For accurate token counting, the correct model name is required. Although there are ways to retrieve the model name from the deployment, they involve complex logic. Therefore, this approach, despite being somewhat inelegant, is the simplest solution given the constraints.

go.mod
```diff
@@ -283,8 +283,8 @@ require (
 	github.com/mroth/weightedrand/v2 v2.0.1
 	github.com/nxadm/tail v1.4.11
 	github.com/oschwald/maxminddb-golang v1.12.0
-	github.com/pkoukk/tiktoken-go v0.1.6
-	github.com/pkoukk/tiktoken-go-loader v0.0.1
+	github.com/pkoukk/tiktoken-go v0.1.7
+	github.com/pkoukk/tiktoken-go-loader v0.0.2-0.20240522064338-c17e8bc0f699
```
arafatkatze, line 287: Had to update to support gpt4o

slimsag:
Will review this tomorrow, apologies for the delay

internal/completions/client/azureopenai/openai.go
```diff
@@ -230,9 +281,19 @@ func streamAutocomplete(
...
+			inputTokens, err := NumTokensFromAzureOpenAiMessages(requestParams.Messages, requestParams.AzureCompletionModel)
+			if err != nil {
+				logger.Warn("Failed to count input tokens with the token manager %w ", log.Error(err))
+			}
+			outputTokens, err := NumTokensFromAzureOpenAiResponseString(content, requestParams.AzureCompletionModel)
 			if err != nil {
-				logger.Warn("Failed to count tokens with the token manager %w ", log.Error(err))
+				logger.Warn("Failed to count output tokens with the token manager %w ", log.Error(err))
+			}
+			if inputTokens != 0 && outputTokens != 0 {
```
arafatkatze, line 292: Check again

arafatkatze:
Added the validation check also 
image

internal/completions/client/azureopenai/openai.go
```diff
@@ -230,9 +281,20 @@ func streamAutocomplete(
 		// stream is done
 		if errors.Is(err, io.EOF) {
 			tokenManager := tokenusage.NewManager()
-			err = tokenManager.TokenizeAndCalculateUsage(requestParams.Messages, content, tokenizer.AzureModel+"/"+requestParams.Model, "code_completions", tokenusage.AzureOpenAI)
+			inputTokens, err := NumTokensFromAzureOpenAiMessages(requestParams.Messages, requestParams.AzureCompletionModel)
+			if err != nil {
+				logger.Warn("Failed to count input tokens with the token manager %w ", log.Error(err))
+				return err
```
slimsag, line 287: Are you positive we want these cases returning token counting errors? I'm unsure.

On the one hand, if token counting is failing that is bad. But on the other hand, if errors are returned here because token counting is failing... then Cody is entirely 100% broken for users.

Seems to me that we're better logging these errors and simply _not recording token counts_ should there be an issue here.

internal/conf/validate_custom.go
```diff
@@ -54,6 +54,14 @@ func validateCustom(cfg Unified) (problems Problems) {
 		}
 	}
 
+	hasAzure := cfg.Completions != nil && cfg.Completions.Provider == "azure-openai"
+	hasAzureChatModel := cfg.Completions != nil && cfg.Completions.AzureChatModel != ""
+	hasAzureCompletionModel := cfg.Completions != nil && cfg.Completions.AzureCompletionModel != ""
+
+	if hasAzure && !(hasAzureChatModel && hasAzureCompletionModel) {
+		invalid(NewSiteProblem(`when using azure-openai provider its mandatory to set both completions.azureChatModel and completions.azureCompletionModel for proper LLM Token usage`))
+	}
```
slimsag, line 63: You should add a contributed validator instead; e.g. inside an init function in `azureopenai/openai.go`. For example [like this](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@ab305657abfbf28e8645048bd622ba935d8a27c0/-/blob/internal/conf/computed.go?L180-196)

Ideally no validation logic would live in this centralized function, as then it becomes too large.

slimsag:
two minor comments, will be LGTM once those are addressed

internal/conf/validate_custom.go
arafatkatze: Fixed

internal/completions/client/azureopenai/openai.go
arafatkatze: I removed the concept of returning the error and just did the logging.

internal/completions/client/azureopenai/openai.go
```diff
@@ -230,7 +281,15 @@ func streamAutocomplete(
...
-			err = tokenManager.TokenizeAndCalculateUsage(requestParams.Messages, content, tokenizer.AzureModel+"/"+requestParams.Model, "code_completions", tokenusage.AzureOpenAI)
+			inputTokens, err := NumTokensFromAzureOpenAiMessages(requestParams.Messages, requestParams.AzureCompletionModel)
+			if err != nil {
+				logger.Warn("Failed to count input tokens with the token manager %w ", log.Error(err))
+			}
+			outputTokens, err := NumTokensFromAzureOpenAiResponseString(content, requestParams.AzureCompletionModel)
+			if err != nil {
+				logger.Warn("Failed to count output tokens with the token manager %w ", log.Error(err))
+			}
+			err = tokenManager.UpdateTokenCountsFromModelUsage(inputTokens, outputTokens, tokenizer.AzureModel+"/"+requestParams.Model, "code_completions", tokenusage.AzureOpenAI)
```
slimsag, line 292: Can you add a comment above this line:

```
// Note: If we had an error calculating input/output tokens, that is unfortunate, the
// best thing we can do is record zero token usage which would be our hint to look at
// the logs for errors.
```

internal/completions/client/azureopenai/openai.go
```diff
@@ -283,7 +342,15 @@ func streamChat(
...
-			err = tokenManager.TokenizeAndCalculateUsage(requestParams.Messages, content, tokenizer.AzureModel+"/"+requestParams.Model, "chat_completions", tokenusage.AzureOpenAI)
+			inputTokens, err := NumTokensFromAzureOpenAiMessages(requestParams.Messages, requestParams.AzureChatModel)
+			if err != nil {
+				logger.Warn("Failed to count input tokens with the token manager %w ", log.Error(err))
+			}
+			outputTokens, err := NumTokensFromAzureOpenAiResponseString(content, requestParams.AzureChatModel)
+			if err != nil {
+				logger.Warn("Failed to count output tokens with the token manager %w ", log.Error(err))
+			}
+			err = tokenManager.UpdateTokenCountsFromModelUsage(inputTokens, outputTokens, tokenizer.AzureModel+"/"+requestParams.Model, "code_completions", tokenusage.AzureOpenAI)
```
slimsag, line 353: same comment here

slimsag (approved):

PR #63105, opened 2024-06-05
enterprise-portal: update RPC for subscription domain and members

unknwon:
Part of CORE-99

This PR adds new RPCs and fields for add and get/list subscription domain and members, prepare to be used by Cody Analytics related operations.

## Test plan

CI


lib/enterpriseportal/subscriptions/v1/subscriptions.proto
```diff
@@ -63,6 +76,8 @@ message EnterpriseSubscription {
   repeated EnterpriseSubscriptionCondition conditions = 2;
   // Display name of this subscription, e.g. "Acme, Inc."
   string display_name = 3;
+  // The instance domain associated with this subscription, e.g. "acme.sourcegraphcloud.com".
+  string domain = 4;
```
bobheadxi, line 80: Should this be `instance_domain`?

Also, do we anticipate other instance details, like `last_reported_instance_id`? 🤔 If so, I think best practice would dictate this be `EnterpriseSubscriptionInstance instance = 4;`

bobheadxi: Thought I guess the last reported instance ID could live on licenses instead as they do now, to indicate the last reported usage of a particular license

bobheadxi: `instance_domain` would map better if we add a `email_domains` in the future 🤔

lib/enterpriseportal/subscriptions/v1/subscriptions.proto
```diff
@@ -192,6 +207,11 @@ message ListEnterpriseSubscriptionLicensesFilter {
     // Return only product subscriptions whose license key contains this
     // substring.
     string license_key_substring = 4;
+    // Return only product subscriptions contains the member with given SAMS
+    // account ID.
+    string member_sams_account_id = 5;
+    // Return only product subscriptions contains the member with the given role.
+    string member_role = 6;
```
bobheadxi, line 214: I think these two should be a single message, i.e.

```
Member member = 6;

message Member {
   string sams_account_id = 1;
   string role = 2;
}
```

This is because it doesn't make sense to filter subscriptions _only_ for roles: roles are only relevant within the context of filtering for a specific member. For example, is there really a use case to select all subscriptions that have _any_ members with a specific role? This particular filter lookup is _only_ for checking membership for a specific user, and we should tailor the filter for that use case

lib/enterpriseportal/subscriptions/v1/subscriptions.proto
```diff
@@ -224,3 +244,40 @@ message ListEnterpriseSubscriptionLicensesResponse {
   // recently created licenses first.
   repeated EnterpriseSubscriptionLicense licenses = 2;
 }
+
+message UpdateEnterpriseSubscriptionRequest {
+  // The subscription to update.
+  // The field "id" is used to identify the subscription to update.
+  EnterpriseSubscription subscription = 1;
```
bobheadxi, line 251: According to AIP we can annotate the ID on the resource with `[(google.api.field_behavior) = IDENTIFIER];`

lib/enterpriseportal/subscriptions/v1/subscriptions.proto
```diff
@@ -224,3 +244,40 @@ message ListEnterpriseSubscriptionLicensesResponse {
...
+    // Look up a subscription using its external, prefixed UUID-format identifier.
+    string id = 1;
+    // Look up a subscription using its instance domain.
+    string domain = 2;
+  }
+
+  // Look up a member using its SAMS account ID.
+  string member_sams_account_id = 3;
+  // The new roles for the member.
+  repeated string member_roles = 4;
```
bobheadxi, line 277: I think AIP best practice dictates that we should create a `Memberbership` resource message type for these two, which includes the `id` and the `domain`, where we can specify both is used for the identifier of this resource type

lib/enterpriseportal/subscriptions/v1/subscriptions.proto
bobheadxi: lmao AIP kills us again:

> The use of IDENTIFIER indicates that a field within a resource message is used to identify the resource. It must be attached to the name field and must not be attached to any other field (see [fields representing resource names](https://google.aip.dev/122#fields-representing-resource-names)).

https://google.aip.dev/203#identifier

🤦

lib/enterpriseportal/subscriptions/v1/subscriptions.proto
unknwon: > Also, do we anticipate other instance details, like `last_reported_instance_id`? 🤔 If so, I think best practice would dictate this be `EnterpriseSubscriptionInstance instance = 4;`

Honestly a new type feels like an overkill 😂 but on the question, I think `last_reported_instance_id` is a legacy concept just for our data pipeline, and as crazy as it sound (and as we know), it is "unreliable". Thus I doubt we want to introduce that particular field to EP.

unknwon: Updated to `instance_domain` in https://github.com/sourcegraph/sourcegraph/pull/63105/commits/341ab25f141c21d9626bed4105521b22d67c999b

lib/enterpriseportal/subscriptions/v1/subscriptions.proto
unknwon: Updated in https://github.com/sourcegraph/sourcegraph/pull/63105/commits/66278cfa5f5cf948fc6d6ceb22cc89e62e296e14

lib/enterpriseportal/subscriptions/v1/subscriptions.proto
unknwon: >must be attached to the name field and must not be attached to any other field

💀

bobheadxi: A bit too late to change I reckon? 😆 I'd rather stay consistent with SAMS

unknwon: TBH `name` creates more confusing, so I'd stick with `id` here 😁

lib/enterpriseportal/subscriptions/v1/subscriptions.proto
unknwon: Pushed changes in https://github.com/sourcegraph/sourcegraph/pull/63105/commits/cf8d985ec16e049fb3c4729576ee5d4a3a01b22b, PTAL and let me know if that's what you're looking for

lib/enterpriseportal/subscriptions/v1/subscriptions.proto
```diff
@@ -224,3 +257,47 @@ message ListEnterpriseSubscriptionLicensesResponse {
...
+  // The following fields are used to identify the membership to update:
+  //  - subscription.id
+  // Multiple fields are treated as AND-concatenated.
+  EnterpriseSubscription subscription = 1;
+
+  // The list of fields to update, fields are specified relative to the EnterpriseSubscription.
+  // Updatable fields are:
+  //  - subscription.display_name
+  //  - subscription.domain
+  google.protobuf.FieldMask update_mask = 2;
```
bobheadxi, line 272: AIP says relative to EnterpriseSubscription as well, should the examples be `display_name` and `domain`? (omit `subscription.`)

lib/enterpriseportal/subscriptions/v1/subscriptions.proto
```diff
@@ -224,3 +257,47 @@ message ListEnterpriseSubscriptionLicensesResponse {
...
+
+message EnterpriseSubscriptionMembership {
+  // The external, prefixed UUID-format identifier of the subscription.
+  string subscription_id = 1;
+  // The instance domain associated with this subscription, e.g. "acme.sourcegraphcloud.com".
+  string instance_domain = 2;
+  // The SAMS account ID of the member.
+  string member_sams_account_id = 3;
+  // The roles of the member.
+  repeated string member_roles = 4;
```
bobheadxi, line 288: Practices outlined in https://protobuf.dev/programming-guides/api/#group-related-fields would suggest that you can make the `Member` type used in filter top-level, and reuse that here 🤔

lib/enterpriseportal/subscriptions/v1/subscriptions.proto
unknwon: Oh shot, i missed the `*not*` part LOL, thought `should`

lib/enterpriseportal/subscriptions/v1/subscriptions.proto
```diff
@@ -224,3 +257,47 @@ message ListEnterpriseSubscriptionLicensesResponse {
...
+}
+
+message UpdateSubscriptionMembershipRequest {
+  // The membership to update.
+  // The following fields are used to identify the membership to update:
+  //  - membership.subscription_id
+  //  - membership.instance_domain
+  // Multiple fields are treated as AND-concatenated.
```
bobheadxi, line 296: Should we mention this in the individual `subscription_id` and `instance_domain` docstrings too?

bobheadxi: Or tag with the AIP "identifier" annotation

lib/enterpriseportal/subscriptions/v1/subscriptions.proto
bobheadxi: Left some new comments, resolving this one!

lib/enterpriseportal/subscriptions/v1/subscriptions.proto
unknwon: In `ListEnterpriseSubscriptionLicensesFilter`, that `Member` has singular `role`, here it's repeated... we can make that be repeated roles as well, but then that nested within a repeated filter... 😂 who wins?

lib/enterpriseportal/subscriptions/v1/subscriptions.proto
unknwon: Fixed in https://github.com/sourcegraph/sourcegraph/pull/63105/commits/0991b4576bb48ab8f2439e512e412a2b9d952068

lib/enterpriseportal/subscriptions/v1/subscriptions.proto
unknwon: `subscription_id` is an easy decision, but `instance_domain` has gotcha for being an identifier right? When "Update", it is not, but otherwise it is 🤔 So I think documenting on the actual use sites is more clear?

lib/enterpriseportal/subscriptions/v1/subscriptions.proto
bobheadxi: Actually, didn't we say that the filter would look for permissions instead of roles?

bobheadxi: image

bobheadxi: So you're right they can't be shared types, but the other one doesn't align with the design right now

unknwon: Updated in https://github.com/sourcegraph/sourcegraph/pull/63105/commits/25d38d7f328aadb8303b3155b0f6fbfd4def16b6!

lib/enterpriseportal/subscriptions/v1/subscriptions.proto
```diff
@@ -224,3 +257,47 @@ message ListEnterpriseSubscriptionLicensesResponse {
...
+
+message EnterpriseSubscriptionMembership {
+  // The external, prefixed UUID-format identifier of the subscription.
+  string subscription_id = 1;
+  // The instance domain associated with this subscription, e.g. "acme.sourcegraphcloud.com".
+  string instance_domain = 2;
+  // The SAMS account ID of the member.
+  string member_sams_account_id = 3;
+  // The permissions of the member, e.g. "cody_analytics::analytics::read".
+  repeated string member_permissions = 4;
```
bobheadxi, line 288: When _updating_ membership, this should be _role_

bobheadxi: image

unknwon: I totally confused myself LOL 😂

bobheadxi: This is how the RBAC model we have in sg/sg is at least:

1. Code callsites check for _permissions_
2. Users request and are assigned _roles_, which are groups of permissions

This lets us create arbitrary roles with arbitrary permissions, without changing all callsites to require N roles. All that matters is that the roles contain the required permissions to access the desired code paths.

In other words:

1. `UpdateSubscriptionMembershipRequest`: _roles_
2. `ListEnterpriseSubscriptionLicensesFilter`: _permissions_

(aka the opposite of what you have 😆 )
(sorry for the confusion!!!)

bobheadxi: The way I suggested tracking this is that in DB, we will record roles, and hardcode roles <-> permissions mappings in the service:

```
cody_analytics::customer-admin {
  perms: [cody_analytics::analytics::read]
}
```

In the future, we could add future scopes, _for example_ if the team wants to introduce the ability to see what exact user IDs use Cody the most, that might be restricted to `customer-admin` but not a plain old `customer`:

```
cody_analytics::customer-admin {
  perms: [cody_analytics::analytics::read, cody_analytics::user_ids::read]
}
cody_analytics::customer{
  perms: [cody_analytics::analytics::read]
}
```

unknwon: Here we go!! https://github.com/sourcegraph/sourcegraph/pull/63105/commits/b71ecaf72540a793c2aa393bba28668c898b69d6

lib/enterpriseportal/subscriptions/v1/subscriptions.proto
```diff
@@ -192,6 +215,17 @@ message ListEnterpriseSubscriptionLicensesFilter {
...
+    Member member = 5;
+  }
+
+  message Member {
+    // Return only product subscriptions contains the member with given SAMS
+    // account ID.
+    string sams_account_id = 1;
+    // Return only product subscriptions contains the member with the given permission,
+    // e.g. "cody_analytics::analytics::read".
+    string permission = 2;
```
bobheadxi, line 228: should we support `repeated string permission`, to match multiple permissions? Would that be an OR or an AND?

unknwon: If we want `repeated string permission`, I think we should treat as `AND`, to be consistent and no confusion.

bobheadxi: The use case I'm thinking is "can this user _either read or write_ to X" but I suppose it should just unsure that the relevant role has read and write explicitly, i.e. no permission implies another. So maybe we don't need a list here 🤔

bobheadxi (approved):
LGTM overall 😁 Nice!!

lib/enterpriseportal/subscriptions/v1/subscriptions.proto
unknwon: >no permission implies another

I think this makes things a lot simpler, also this is what SAMS token scope spec does. (BTW "read/write" is an "action" in SAMS's term... 😂 😆 )

bobheadxi: On that note, should we use the term `scope` instead of `permission` here?

unknwon: Sounds good!!! 😄

PR #63110, opened 2024-06-05
feat(search): Support Magik language file filter in search

mmanela:
Fixes GRAPH-622 

Adds support to the lang file filter for Magik language (`lang: Magik`). In order to do that we add wrappers around go-enry and update search code to use them.  This provides flexibility for us in the future to support other languages that are not in Linguist as well. 

This PR does NOT update symbols for Magik (that needs the future scip-ctags work) or udpated code insights (will follow up with that)

## Test plan

- Update unit tests
- Manually test lang filter

## Changelog
- Added Magik language to language filters (`lang:magik`)


client/common/src/languages.ts
```diff
@@ -46,6 +46,12 @@ export const POPULAR_LANGUAGES: string[] = [
     'Zig',
 ]
 
+/**
+ * Languages that are not present in the Linguist library
+ * and so are not in go-enry either
+ */
+const LANGUAGES_NOT_IN_LINGUIST = ['Magik']
```
jtibshirani, line 53: Small comment: let's standardize terminology, since we currently use both "languages not in linguist" and "unsupportedByEnry". Maybe we could just say "extra" in both these cases, to keep things concise.

internal/search/types.go
```diff
@@ -421,6 +422,20 @@ func (f *Features) String() string {
 	return flagMap.String()
 }
 
+func (f *Features) UseContentBasedLangFilters(langs []string) bool {
```
jtibshirani, line 425: It feels surprising to disable content based lang filtering if either the include or exclude `lang` is an "extra" one. A user could have a query like `foobar -lang:cpp`, then modify it to `foobar -lang:cpp -lang:magik`, and all of a sudden the meaning of "exclude C++' changes a lot.

Although it's slightly more work, I think we should stick with your original plan to:
1. Just update it here for `search-content-based-lang-filtering: false`
2. Then update Zoekt to recognize `.magik` extensions and record this in the index

mmanela: Thanks @jtibshirani , Ill back these changes out and follow up with the zoekt PR for those parts

client/common/src/languages.ts
mmanela: Thanks for the comment, I will be consistent in language although I don't love Extra since it doesn't have a semantic meaning

jtibshirani (commented):
@mmanela and I chatted offline about adding more tests.

We also decided that this PR will only be scoped to indexed/ unindexed search. In follow up work, we will handle
* Symbol search (both symbols and Rockskip)
* Other places that use `enry` directly, specifically Code Insights

client/common/src/languages.ts
```diff
@@ -46,6 +46,12 @@ export const POPULAR_LANGUAGES: string[] = [
     'Zig',
 ]
 
+/**
+ * Languages that are not present in the go-enry library
+ * and so are not in go-enry either
+ */
+const LANGUAGES_NOT_IN_ENRY = ['Magik']
```
jtibshirani, line 53: I just noticed, why do we have this new constant? Could we just add this to `ALL_LANGUAGES`? Also do we need to add `Pkl` (I don't see it anywhere in this file)?

internal/search/job/jobutil/job_test.go
```diff
@@ -1461,6 +1509,13 @@ func TestToSymbolSearchRequest(t *testing.T) {
 		input:  `repo:go-diff type:symbol HunkNoChunksize lang:Julia -lang:R`,
 		feat:   search.Features{ContentBasedLangFilters: true},
 		output: autogold.Expect(`{"RegexpPattern":"HunkNoChunksize","IsCaseSensitive":false,"IncludePatterns":null,"ExcludePattern":"","IncludeLangs":["Julia"],"ExcludeLangs":["R"]}`),
+	}, {
```
jtibshirani, line 1512: These tests don't really make sense yet because symbol search isn't supported for Magik. Instead, let's update `TestToTextPatternInfo`, which is the main conversion function for searcher (unindexed text search). We can test the conversion works correctly both with and without content-based lang filtering.

client/common/src/languages.ts
mmanela: 1. I didn't add to that list since it has a comment above saying how you can generate directly from a shell command. 
2. I added it now BUT I just checked and it was [just](https://github.com/github-linguist/linguist/pull/6730) added to Linguist. So I filed an issue to remove once go-enry takes that linguist update

internal/search/job/jobutil/job_test.go
mmanela: updated

jtibshirani (approved):

client/common/src/languages.ts
```diff
@@ -46,6 +46,20 @@ export const POPULAR_LANGUAGES: string[] = [
     'Zig',
 ]
 
+/**
+ * Languages that are not present in the go-enry library
+ * and so are not in go-enry either
```
varungandhi-src, line 51: The first mention of go-enry should be Linguist instead.

client/common/src/languages.ts
```diff
@@ -46,6 +46,20 @@ export const POPULAR_LANGUAGES: string[] = [
...
+ */
+const LANGUAGES_NOT_IN_ENRY = [
+    // Not in Linguist and they are not likely to add
+    'Magik',
```
varungandhi-src, line 55: Linking the PR would've been useful here as context.

lib/codeintel/languages/extensions.go
```diff
@@ -2,10 +2,32 @@ package languages
...
+
+// getLanguagesByExtension is a replacement for enry.GetLanguagesByExtension
+// It supports languages that are missing in go-enry
+func GetLanguageExtensions(alias string) []string {
+	if lang, ok := unsupportedByEnryNameToExtensionMap[alias]; ok {
+		return []string{lang}
+	}
+
+	return enry.GetLanguageExtensions(alias)
+}
```
varungandhi-src, line 29: If you look just below this, we have a function which overrides the behavior of `enry`'s `GetLanguagesByExtension`. However, the behavior of this function `GetLanguageExtensions` is not consistent with that, because it doesn't make use of the data in `overrideAmbiguousExtensionsMap`.

Specifically, if `getLanguagesByExtension("foo.json")` returns `"JSON"`, then it is not OK for `GetLanguageExtensions("OASv2-json")` to return `".json"` or `"json"` (https://sourcegraph.com/github.com/github-linguist/linguist/-/blob/lib/linguist/languages.yml?L4750-4760)

lib/codeintel/languages/extensions.go
```diff
@@ -62,13 +84,33 @@ var overrideAmbiguousExtensionsMap = map[string]string{
...
 	// Magik Language
 	".magik": "Magik",
 }
 
+var unsupportedByEnryNameToExtensionMap = reverseMap(unsupportedByEnryExtensionToNameMap)
+
+var unsupportedByEnryAliasMap = map[string]string{
+	// Pkl Configuration Language (https://pkl-lang.org/)
+	"pkl": "Pkl",
+	// Magik Language
```
varungandhi-src, line 102: This comment does not convey any extra information over the line of code below it; it would make sense to either add a link to help the reader or not have a comment at all.

lib/codeintel/languages/extensions.go
```diff
@@ -62,13 +84,33 @@ var overrideAmbiguousExtensionsMap = map[string]string{
 	// for other variants of YAML, hence only 'YAML' is picked by enry.
 }
 
-var unsupportedByEnryExtensionsMap = map[string]string{
+var unsupportedByEnryExtensionToNameMap = map[string]string{
 	// Pkl Configuration Language (https://pkl-lang.org/)
+	// NOTE: Add to linguist on 6/7/24
+	// can remove once go-enry package updates
+	// to that linguist version
```
varungandhi-src, line 91: Leaving this purely as a code comment increases the likelihood that it will be forgotten. Instead, you could've added a Go test that the enry list of languages doesn't have this one by doing a map lookup in https://pkg.go.dev/github.com/go-enry/go-enry/v2@v2.8.8/data#LanguageInfo

In the code, you can add an easy-to-search for marker, like `TODO(id: remove-pkl-special-case)` and mention that in the test code so that when someone sees the test failure, they know what other places need updating.

PR #63112, opened 2024-06-05
Cody Gateway embeddings: powering with generated metadata - take 2

janhartman:
Reverts sourcegraph/sourcegraph#63098 and fixes the problems with conf loading and Fireworks API responses.

## Test plan
Try running locally and then with feature flag.


internal/completions/client/fireworks/fireworks.go
```diff
@@ -94,9 +95,9 @@ func (c *fireworksClient) Complete(
 	if response.Choices[0].Text != "" {
 		// The /completion endpoint returns a text field ...
 		completion = response.Choices[0].Text
-	} else if response.Choices[0].Delta != nil {
+	} else if response.Choices[0].Message != nil {
```
janhartman, line 98: This is the key part - according to the [api docs](https://readme.fireworks.ai/reference/createchatcompletion), we need to access the `message` field in the non-streaming case.

internal/completions/client/fireworks/fireworks.go
```diff
@@ -297,7 +298,10 @@ type message struct {
 // response for a non streaming request
 type fireworksResponse struct {
 	Choices []struct {
-		Text  string `json:"text"`
+		Text    string `json:"text"`
+		Message *struct {
```
janhartman, line 302: By adding this here, we now support streaming and non-streaming with the same struct. I suppose it could be a bit cleaner if we were to separate this into two.

rafax (approved):
LGTM, adding @sourcegraph/cody-prime for the Fireworks client change

cmd/cody-gateway/internal/httpapi/embeddings/metadata.go
```diff
@@ -0,0 +1,55 @@
...
+	return mapper.MapErr(req.Input, c.generateMetadataForChunk)
+}
+
+func (c *metadataClient) generateMetadataForChunk(input *string) (string, error) {
+	promptText := `Here is a section of code.
+Please write a paragraph of documentation for each high-level class, struct, function or similar.
+Be concise, write no more than a few sentences for each entry.
+Return your response in text format. Each entry name should be followed by a newline, then its documentation.
+Respond with nothing else, only the entry names and the documentation. Code: ` +
+		"````" + *input + "```"
```
chrsmith, line 36: This is just a regular markdown code block, right?

```suggestion
		"```" + *input + "```"
```

cmd/cody-gateway/internal/httpapi/embeddings/metadata.go
```diff
@@ -0,0 +1,55 @@
...
+
+	resp, err := c.completionsClient.Complete(c.ctx, types.CompletionsFeatureChat, types.CompletionsVersionLegacy,
+		types.CompletionRequestParameters{
+			Messages: []types.Message{{
+				Speaker: "user",
+				Text:    promptText,
+			}},
+			MaxTokensToSample: 2000,
```
chrsmith, line 44: I assume this is what you want, but all of our completion endpoints receive the "max tokens to sample" as a parameter. So if we are "standardizing" completions having a specific LLM model, token count, etc. then perhaps we should move this out into a dedicated function?

e.g. `func buildCompletionRequestParams(promptText string) types.CompletionRequestParams`?

It just seems like we are hard coding some things here that might better be suited for being package level, e.g. `const maxTokensToSample = 2000` or having this isolated to a dedicated function. (Rather than having these all be fixed in `generateMetadataForChunk`.)

Does that make sense? This isn't a major issue, just a minor code smell. So if none of this resonates with you feel free to ignore.

cmd/cody-gateway/shared/config/config.go
```diff
@@ -297,7 +297,7 @@ func (c *Config) Load() {
 		c.AddError(errors.New("must provide allowed models for Google"))
 	}
 
-	c.AllowedEmbeddingsModels = splitMaybe(c.Get("CODY_GATEWAY_ALLOWED_EMBEDDINGS_MODELS", strings.Join([]string{string(embeddings.ModelNameOpenAIAda), string(embeddings.ModelNameSourcegraphSTMultiQA)}, ","), "The models allowed for embeddings generation."))
+	c.AllowedEmbeddingsModels = splitMaybe(c.Get("CODY_GATEWAY_ALLOWED_EMBEDDINGS_MODELS", strings.Join([]string{string(embeddings.ModelNameOpenAIAda), string(embeddings.ModelNameSourcegraphSTMultiQA), string(embeddings.ModelNameSourcegraphMetadataGen)}, ","), "The models allowed for embeddings generation."))
```
chrsmith, line 300: For readability, consider teasing the model names out into a separate variable. (To just avoid the horizontal scrolling.

```suggestion
	defaultEmbeddingModles := []string{
		string(embeddings.ModelNameOpenAIAda),
		string(embeddings.ModelNameSourcegraphSTMultiQA),
		string(embeddings.ModelNameSourcegraphMetadataGen)}
	configuredEmbeddingModels := c.Get(
		"CODY_GATEWAY_ALLOWED_EMBEDDINGS_MODELS",
		strings.Join(allowedEmbeddingsModels, ","),
		"The models allowed for embeddings generation.")
	c.AllowedEmbeddingsModels = splitMaybe(configuredEmbeddingModels)
```

internal/codygateway/types.go
```diff
@@ -46,6 +46,8 @@ type EmbeddingsRequest struct {
 	Model string `json:"model"`
 	// Input is the list of strings to generate embeddings for.
 	Input []string `json:"input"`
+	// IsQuery is true if the request is used for querying, false if it used for indexing.
+	IsQuery bool `json:"isQuery"`
```
chrsmith, line 50: It seems confusing to have IsQuery=false imply indexing.

What do you think about just having a `Kind string` and match it against `'query'` or `'indexing'`? I'd also suggest we return a 400 to enforce that the caller knows what they are doing. (Instead of defaulting to one or the other if the kind isn't specified.)

What do you think?

internal/completions/client/fireworks/fireworks.go
chrsmith: Good to know. But if that's subtle and very important, consider adding a comment in the code to say just that. (So that the next person stumbling across the code has a hint as to why this needs to be `Message` instead of `Delta`.

internal/completions/client/fireworks/fireworks.go
chrsmith: I would strongly suggest we separate these two. Since otherwise it's confusing as to what data should/should not be present. (e.g. if you have type `fireworksResponse`, it isn't clear what is required for it to be well formed.

It also makes maintaining the API surface area easier, since things are more explicit. (Even at the cost of having two types that are nearly identical, just one being called `fireworksStreamingResponse` or something.

chrsmith (commented):
Overall things look good to me, modulo some minor cleanups and nit picks.

cmd/cody-gateway/internal/httpapi/embeddings/metadata.go
chrsmith: Also, just to make debugging these prompts easier for humans should the need arise, would it make sense to have the code fenced around "```\n"?

cmd/cody-gateway/internal/httpapi/embeddings/metadata.go
janhartman: I understand your concern. However, this is purely experimental for now so I'll leave it as is. If this ends up working well we will likely revisit and refactor this module anyway, otherwise it's going to be deleted soon.

internal/codygateway/types.go
janhartman: I agree with your line of thinking. However, I'll leave it be for now (same reasons as above).

janhartman:
Thanks for the review @chrsmith, great points! I'll be merging this as is now so I can start dogfooding it and then revisit it when I have some experiment results.

internal/codygateway/types.go
chrsmith: Sounds good, we don't need the code to be perfect if we are just experimenting. However, making this type of breaking API change later can be a pain in the ass.

So it's fine to just leave it as-is for now. But you may want to consider adding a comment to make this stand out later as something we'd want to refactor if we do want to "graduate" this code so to speak. e.g.

```
// DESIGN WART: <Here's why this isn't a great design. Maybe we could refactor it this way...>
```
or
```
// TODO: <Refactor this to use 'other design' for reason. This will require updating callers <here>.
```

Just something to consider so that we can at least know what things we might want to refactor sooner rather than later.

PR #63120, opened 2024-06-06
fix(gateway): double provider in telemetry

chenkc805:
On march 7, and we see "anthropic/anthropic" happen on "request blocked" events. This PR fixes that double provider logging error.

Screenshot 2024-06-06 at 1 47 08 AM

## Test plan

check telemetry


chenkc805:
Closing because it's 1) not an easy fix and 2) it's not a big deal

PR #63121, opened 2024-06-06
fix(release): remove generate-notes from github release creation

burmudar:
Generate notes currently generates a changelog all the way from the tag `app-2023` ... which means the content is more than what is allowed by github - hard cap of 125000 characters.

## Test plan
CI

## Changelog
- remove `generate-notes` cli flag when promoting releases
- push only a single tag
- fix token name used in changelog generation


craigfurman (approved):
🚀

craigfurman:
Backported: https://github.com/sourcegraph/sourcegraph/pull/63148

PR #63122, opened 2024-06-06
scim: Fix random suffix added to every user on resync

eseliger:
In the current implementation, we check if there's a user in the database already, and if so we add a random suffix to increase the chances of the insert/update to succeed.

However, we didn't check if the user that already exists is _the same user_. So for a first sync, usernames would look nice and tidy, but then on a second sync from the SCIM provider, every user would get a suffix.

This PR fixes that by adding a check for the user ID.

Test plan:

Added tests, verified SCIM sync doesn't modify usernames anymore locally.


eseliger:
* **#63122** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/63122?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

pjlast (approved):

vdavid:
Oh, good catch, thanks, @eseliger!

PR #63123, opened 2024-06-06
fix(search): Properly handle URLs with file paths that contain route segments

fkling:
Fixes SRCH-484. See that issue for more information about the problem.

Instead of hardcoding the routes that can contain file paths, I'm using the generated list of routes in `routes.ts` and extract any route that ends with a `[...path]` segment (which, so far at least, is used as the file path parameter).

This list is then used to hook into the routing process and encode the file path when any of the routes match. As mentioned in the `reroute` documentation, this does not affect the URL displayed in the browser. Because the file path is now URL encoded we need to decode it before we can process it further (I grepped for all occurrences of `params.path`). 

E.g. for https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/-/blob/[...path]/+page.svelte

| Before | After |
|--------|--------|
|  |  | 

## Test plan

Manual testing, integration test.

### DIFF
--- a/client/web-sveltekit/src/hooks.test.ts
+++ b/client/web-sveltekit/src/hooks.test.ts
@@ -1,0 +1,15 @@
+import { expect, test, describe } from 'vitest'
+
+import { getStaticSuffix_TEST_ONLY as getStaticSuffix } from '$root/client/web-sveltekit/src/hooks'
+
+describe('getStaticSuffix', () => {
+    test.each([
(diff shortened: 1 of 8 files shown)


camdencheek (approved):
Nice! That's not too bad. Thanks for the fix!

PR #63124, opened 2024-06-06
chore(tooling): bump Go version to 1.22.4

Strum355:
Bump for @evict 

## Test plan

CI passes with no complaints

## Changelog

- Bumped version of Go used to build to 1.22.4

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,4 +1,4 @@
-golang 1.22.1
+golang 1.22.4
 nodejs 20.8.1
 fd 8.6.0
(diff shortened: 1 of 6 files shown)


evict (approved):

burmudar (approved):

keegancsmith:
What is the benefit of specifying patch version in go.mod? From a development PoV it doesn't matter, and for our builds it we use the bazel version. I guess it kinda makes sense, and I just need to bite the NixOS not having 1.22.4 yet bullet.

PR #63126, opened 2024-06-06
Svelte: add sentry

camdencheek:
This adds basic sentry support to the Sveltekit webapp. 

Fixes SRCH-90

Additionally, it configures Vite to generate source maps as part of the build so the stack traces in Sentry are readable. I configured it to build public source maps because our source code is already public and this means we don't need to add any special hooks in CI.

## Test plan

Created a test project in Sentry and viewed unhandled exceptions from my local installation. Note that Firefox's "Enhanced Tracking Protection" blocks Sentry, as do many adblockers, so it's likely this won't work for many users. I did, however, test that the application still works find even if Sentry fails to load.


client/web-sveltekit/src/hooks.client.ts
```diff
@@ -0,0 +1,28 @@
+import { handleErrorWithSentry } from '@sentry/sveltekit'
+import * as Sentry from '@sentry/sveltekit'
+
+Sentry.init({
+    dsn: window.context.sentryDSN ?? undefined,
```
fkling, line 5: What happens on the sentry side of things when `dsn` is `undefined`? And when would it be `undefined`? Dotcom, cloud, self-hosted?

fkling (approved):
🥳 very excited about making better use of this

client/web-sveltekit/src/hooks.client.ts
camdencheek: I'll add a comment, but sentry is just disabled if the DSN is undefined

camdencheek: The DSN is settable in site config, and I know it's set by default for dotcom and S2, but I don't think we set it for customer instances. I plan to follow up with the cloud team to ask about that

fkling: >  sentry is just disabled if the DSN is undefined

Good, that's what I hoped. Thank you!

PR #63128, opened 2024-06-06
svelte: Add bottom panel close button

fkling:
Closes SRCH-450

This commit adds a button to the right hand side of the tabs that allows closing the bottom panel.

For this I

- extended the Tabs.svelte component to allow rendering something to the right hand side of the tabs.
- added a new `text` variant to button that renders an unstyled button and shows a background on hover

Note that the icon as shown in the figma design doesn't exist in lucide.

## Test plan

Visual/manual inspection in light and dark mode

## Changelog

- Added bottom panel close button
- Added 'text' variant to Button component
- Added 'actions' slot to Tabs component
- Added Button component stories


client/web-sveltekit/src/lib/wildcard/Button.ts
```diff
@@ -13,6 +13,7 @@ export const BUTTON_VARIANTS = [
     'merged',
     'link',
     'icon',
+    'text',
```
camdencheek, line 16: Oh nice. I've wanted something like this in the past

client/web-sveltekit/src/lib/wildcard/Button.module.scss
```diff
@@ -40,24 +39,17 @@
...
-        outline: 2px solid var(--primary-2);
-        outline-offset: 2px;
-    }
-
+.btn-link,
+.btn-text {
+    &:global(.disabled),
+    &:disabled,
     &[aria-disabled='true'] {
-        cursor: not-allowed;
```
camdencheek, line 53: why get rid of the not-allowed cursor?

camdencheek: Huh. Looks like it's working as expected still in storybook

camdencheek (approved):

client/web-sveltekit/src/lib/Tabs.svelte
```diff
@@ -66,7 +66,12 @@
 </script>
 
 <div class="tabs" data-tabs>
-    <header><TabsHeader {id} tabs={$tabs} selected={$selectedTab} on:select={selectTab} /></header>
+    <header>
+        <TabsHeader {id} tabs={$tabs} selected={$selectedTab} on:select={selectTab} />
+        <div class="actions">
+            <slot name="header-actions" />
+        </div>
```
camdencheek, line 73: Should we use this slot for the last commit info when closed?

client/web-sveltekit/src/lib/wildcard/Button.module.scss
fkling: Sorry, I wanted to leave some comments but didn't have time. I removed this selected because it's not used by us.

fkling:
@taiyab Should we make the button larger? It's currently small. But that will also increase the font size and it seemed a little too large to me.

client/web-sveltekit/src/lib/Tabs.svelte
fkling: We could though I don't know whether that works together with the commit message text-overflow stuff. Something to investigate.

taiyab:
> @taiyab Should we make the button larger? It's currently small. But that will also increase the font size and it seemed a little too large to me.

Nope, no need for now. I'm designing a whole new design system and component library that will take care of this problem properly down the road.

PR #63129, opened 2024-06-06
svelte: Migrate repo sidebar toggle button to use lucide icon

fkling:
See title.

Contributes to SRCH-467

| Before | After |
|--------|--------|
|  |  |

## Test plan

Visual inspection.


taiyab (commented):
The panel one here worked fine.

fkling:
> The panel one here worked fine.

I felt that would be inconsistent then with the bottom panel one. It's the same icon size, so why would it work in one case but not the other?

camdencheek (approved):

PR #63130, opened 2024-06-06
fix(SCIM): don't allow SCIM users to change fields

evict:
Previously it was possible to use the `updateUser` mutation to change user fields for SCIM managed users. We now reject these changes if the user is SCIM managed.

## Test plan
CI tests

### DIFF
--- a/cmd/frontend/graphqlbackend/user.go
+++ b/cmd/frontend/graphqlbackend/user.go
@@ -460,12 +460,17 @@ func (r *schemaResolver) UpdateUser(ctx context.Context, args *updateUserArgs) (
 	}
 
 	update := database.UserUpdate{
-		DisplayName: args.DisplayName,
 		AvatarURL:   args.AvatarURL,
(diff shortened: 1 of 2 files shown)


cmd/frontend/graphqlbackend/user.go
```diff
@@ -460,14 +460,19 @@ func (r *schemaResolver) UpdateUser(ctx context.Context, args *updateUserArgs) (
...
 	}
+
 	user, err := r.db.Users().GetByID(ctx, userID)
 	if err != nil {
 		return nil, errors.Wrap(err, "getting user from the database")
 	}
 
+	if user.SCIMControlled {
+		return nil, errors.New("cannot update externally managed user")
+	}
```
eseliger, line 474: I believe updating the avatar URL is allowed, only displayname and username changes are disabled - at least that is what the UI currently does 🤔

eseliger (approved):
thanks for tests!

PR #63064, opened 2024-06-04
gitserver(client): Reintroduce 500 maximum connections limit

eseliger:
This used to exist in the HTTP world, and we currently have zero safeguards to prevent clients from making one billion requests concurrently.
Until we invest more into server-side rate limiting, or per tenant rate limiting, we reintroduce this limiter, to prevent resource usage spikes.

Test plan:

Added a test suite.

PR #63132, opened 2024-06-06
Revert "gitserver(client): Reintroduce 500 maximum connections limit (#63064)"

eseliger:
This reverts commit 9185da3c3e4282a59529d6f9b90ac945026f0934.

Noticed there are some bad callers in worker and symbols that don't properly return a connection. Will need to investigate and fix that first.

## Test plan

Worked before, CI passes.


camdencheek (approved):
This have any significant impact on the release that just went out?

eseliger:
no, I waited for that to happen 😬

PR #63135, opened 2024-06-06
scim: Fix user updates when SCIM was previously enabled

eseliger:
When the user external account entry for SCIM has been deleted in the past, updating the user again fails as the deleted record is found and then the insertion fails.

This PR fixes that by adjusting the unique index to only consider non-soft-deleted external accounts.

Closes SRC-383

Test plan:

Added a test suite for the DB method - it was untested before :(

### DIFF
--- a/internal/database/external_accounts_test.go
+++ b/internal/database/external_accounts_test.go
@@ -109,6 +109,104 @@ func TestExternalAccounts_AssociateUserAndSave(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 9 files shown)


eseliger:
* **#63135** <a href="https://app.graphite.dev/github/pr/sourcegraph/sourcegraph/63135?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

vdavid (approved):
Nice, looks good, and thanks for the test! :)

PR #63136, opened 2024-06-06
chore(search): Add test case covering hack lang detection behavior

mmanela:
Adds test case to cover the content based behavior for the Hack language 

## Test plan
- Ensure unit test passes 

## Changelog
- Add test case for coverage hack lang detection

### DIFF
--- a/internal/highlight/language_test.go
+++ b/internal/highlight/language_test.go
@@ -307,6 +307,24 @@ func TestGetLanguageFromContent(t *testing.T) {
 			Contents: ``,
 			Expected: "c++",
 		},
+		{
+			Filename: "cpp_hh.hh",
+			Contents: `#pragma once`,
+			Expected: "c++",
+		},
+		{
+			Filename: "cpp_hh.hh",
+			Contents: `#help_index "Bit"`,
+			Expected: "c++",
+		}, {
+			Filename: "cpp_hh.hh",
+			Contents: `/*
+* I am a comment
+*/
+#ifndef __CPU_O3_CPU_HH__
+#define __CPU_O3_CPU_HH__`,
+			Expected: "c++",
+		},
 	}
 
 	for _, testCase := range cases {
@@ -341,6 +359,21 @@ func TestGetLanguageFromPath(t *testing.T) {
 		{
 			Filename: "react.tsx",
 			Expected: "tsx",
+		},
+		{
+			Filename: "hack.hack",
+			Expected: "hack",
+		},
+		{
+			Filename: "cpp.cpp",
+			Expected: "c++",
+		},
+
+		// This resolves to C++ and not hack since
+		// it does not find the <?hh to indicate hack
+		{
+			Filename: "file.hh",
+			Expected: "c++",
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


internal/highlight/language_test.go
```diff
@@ -296,6 +296,17 @@ func TestGetLanguageFromContent(t *testing.T) {
...
 			Expected: "matlab",
 		},
+		{
+			Filename: "hack_hh.hh",
+			Contents: `<?hh`,
+			Expected: "hack",
+		},
+		{
+			Filename: "cpp_hh.hh",
+			// empty file counts as cpp since we only do explicit check to see if it is hack
```
mmanela, line 306:
```suggestion
			// empty file counts as cpp since we only do an explicit check to see if it is hack
```

internal/highlight/language_test.go
```diff
@@ -296,6 +296,17 @@ func TestGetLanguageFromContent(t *testing.T) {
...
+			Filename: "hack_hh.hh",
+			Contents: `<?hh`,
+			Expected: "hack",
+		},
+		{
+			Filename: "cpp_hh.hh",
+			// empty file counts as cpp since we only do an explicit check to see if it is hack
+			Contents: ``,
+			Expected: "c++",
+		},
```
varungandhi-src, line 309: Let's add some more test cases.
1. A file starting with `#pragma once` should be classified as C++, not Hack. (counter-example today: https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/NixOS/nix%24+file:%5Esrc/libstore/globals%5C.hh+lang:Hack&patternType=keyword&sm=0)
2. A file with some comment lines followed by `^ifndef ABC` or `^#include <iostream>` or similar should be classified as C++, not Hack. (counter-example today: https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/harvard-acc/gem5-aladdin%24+file:%5Esrc/cpu/o3/cpu%5C.hh+lang:Hack&patternType=keyword&sm=0)
3. Looks like sometimes Hack gets mixed up with Holy C. Maybe add a test similar to the one above but with `#help_index` (https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/Zeal-Operating-System/ZealOS%24+file:%5Esrc/Kernel/KernelB%5C.HH+lang:Hack&patternType=keyword&sm=0).

It's fine if none of these tests give the right result, but at least the limitation will be documented in the tests.

varungandhi-src (approved):
LGTM

internal/highlight/language_test.go
varungandhi-src: > They all give the right answer

Strictly speaking, the Holy C one should be returning `holyc`, not `c++` :P

mmanela: Ill add a comment there

mmanela: Ah linguist only consider `.hc` to be [HolyC](https://github.com/github-linguist/linguist/blob/2d556f77b88bfdb685bb749ef426f9767be18226/lib/linguist/languages.yml#L2933) 
image

varungandhi-src: Yep, I saw that, but in practice, there are at least some `.hh` files which are Holy C. If you poke around the ZealOS codebase, it also uses many other extensions.

PR #63137, opened 2024-06-06
enterprise-portal: rename "scope" to "permission" for query RPC

unknwon:
Part of CORE-99

After some heavy discussion, we settled on "Roles + Permissions" (that is a whole different large topic), but as far as the RPCs here concerned, using "scopes" here is not appropriate.

## Test plan

CI


lib/enterpriseportal/subscriptions/v1/subscriptions.proto
```diff
@@ -223,9 +223,9 @@ message ListEnterpriseSubscriptionLicensesFilter {
     // Return only product subscriptions contains the member with given SAMS
     // account ID.
     string sams_account_id = 1;
-    // Return only product subscriptions contains the member with the given scope,
+    // Return only product subscriptions contains the member with the given permission,
     // e.g. "cody_analytics::analytics::read".
```
bobheadxi, line 227: Should we remove this example for now, since we discussed using a more Zanzibar-like format?

bobheadxi (approved):
Diff in generated code is crazy for such a small change 😆

PR #63138, opened 2024-06-06
fix(highlighting): Update JSX/TSX highlighting to more properly label tag  delimiters

mmanela:
Update JSX/TSX highlighting to identify tag delimiters properly. 

## Test plan

- Unit tests pass
- Inspect snap file

 
## Changelog
- Update highlighting of JSX/TSX Tags to correctly label delimiters

### DIFF
--- a/client/wildcard/src/global-styles/highlight.scss
+++ b/client/wildcard/src/global-styles/highlight.scss
@@ -592,7 +592,7 @@
     .hl-typed-Tag {
         color: var(--hl-green);
     }
-    .hl-typed-TagDelimited {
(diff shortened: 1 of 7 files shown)


varungandhi-src (approved):
Are you planning to fix the CSS in this PR or a different PR? It still says `TagDelimited` instead of `TagDelimiter` on `main`.

(Also, I noticed you're using the `search` area for highlighting PRs, but it's technically incorrect, because this highlighting does not actually affect search results, which go through a different code path.)

mmanela:
Thanks for calling that out, I can fix in this PR @varungandhi-src

PR #63139, opened 2024-06-06
enterprise-portal: init database schema and handler store

unknwon:
Part of CORE-99

This PR scaffolds the database schema and code structure based on [CORE-99 comment](https://linear.app/sourcegraph/issue/CORE-99/enterprise-portal-design-sams-user-to-subscription-rpcs#comment-8105ac31) with some modifications. See inline comments for more elaborations.
- It uses GORM's ONLY for auto migration, just to kick things off, we may migrate to file-based migration like we are planning for SAMS.
- It then uses the `*pgxpool.Pool` as the DB interface for executing business logic queries.

Additionally, refactored `subscriptionsservice/v1.go` to use a `Store` that provide single interface for accessing data(base), as we have been doing in SAMS and SSC.

## Test plan

Enterprise Portal starts locally, and database is initialized:

CleanShot 2024-06-06 at 17 02 42@2x


lib/managedservicesplatform/runtime/contract/postgresql.go
```diff
@@ -49,7 +49,7 @@ func (c postgreSQLContract) OpenDatabase(ctx context.Context, database string) (
 		if err != nil {
 			return nil, err
 		}
-		return sql.Open("customdsn", stdlib.RegisterConnConfig(config.ConnConfig))
+		return sql.Open("pgx", stdlib.RegisterConnConfig(config.ConnConfig))
```
unknwon, line 52: `"customdsn"` has never worked, this PR is its first user 😂

internal/redislock/mutex.go
unknwon, line 1: Copied from https://github.com/sourcegraph/sourcegraph-accounts/blob/main/backend/internal/redislock/mutex.go

internal/redislock/mutex_test.go
unknwon, line 1: Copied from https://github.com/sourcegraph/sourcegraph-accounts/blob/main/backend/internal/redislock/mutex_test.go and adjusted test setup accordingly.

cmd/enterprise-portal/internal/database/subscriptions.go
```diff
@@ -0,0 +1,25 @@
+package database
+
+// Subscription is a product subscription record.
+type Subscription struct {
+	// ID is the prefixed UUID-format identifier for the subscription.
+	ID string `gorm:"primaryKey"`
+	// InstanceDomain is the instance domain associated with the subscription, e.g.
+	// "acme.sourcegraphcloud.com".
+	InstanceDomain string `gorm:"index"`
+}
```
unknwon, line 10: EP is still not the source of truth for subscriptions yet, so I think start small makes sense for things we actually need. Later PR(s) we will combine data in the handler layer, until EP becomes the source of truth.

We will add full-blown columns during when we move source-of-truth from dotcom to EP.

cmd/enterprise-portal/internal/database/subscriptions.go
```diff
@@ -0,0 +1,25 @@
...
+	// SubscriptionID is the ID of the subscription (aka. object).
+	SubscriptionID string `gorm:"not null"`
+	// SAMSAccountID is the SAMS account ID of the member (aka. subject).
+	SAMSAccountID string `gorm:"not null;index"`
+	// Namespace is the namespace of the relationship of the member to the
+	// subscription.
+	Namespace string `gorm:"not null;index"`
+	// Relation is the relationship of the member to the subscription.
+	Relation string `gorm:"not null"`
+}
```
unknwon, line 25: Per team discussion, we want to adopt Zanzibar concepts, thus making the table schema closer to its terms. Arguably... we may want to just have a `permisisons` table and have columns like

- `namespace string`, e.g. `cody_analytics`
- `subject string`, e.g. `User:<SAMS account ID>`
- `object string`, e.g. `Subscription:<ID>`
- `relation string`, e.g. `customer_admin`

Thoughts?

lib/managedservicesplatform/runtime/contract/postgresql.go
bobheadxi: I think this is a regression from recent PR to use pool, I've definitely used PGDSN for msp-testbed before

cmd/enterprise-portal/internal/database/subscriptions.go
bobheadxi: a time field like `commit_time` sounds like it would be useful, for bookkeeping?

bobheadxi: overall LGTM though - I think another difference is also lack of subject sets https://www.ory.sh/docs/keto/concepts/subjects#subject-sets , but I doubt we'll have a use for that just yet

cmd/enterprise-portal/internal/subscriptionsservice/v1_store.go
```diff
@@ -0,0 +1,38 @@
...
+func NewStoreV1(opts NewStoreV1Options) StoreV1 {
+	return &storeV1{
+		db:       opts.DB,
+		dotcomDB: opts.DotcomDB,
+	}
+}
+
+func (s *storeV1) ListEnterpriseSubscriptionLicenses(ctx context.Context, filters []*subscriptionsv1.ListEnterpriseSubscriptionLicensesFilter, limit int) ([]*dotcomdb.LicenseAttributes, error) {
+	return s.dotcomDB.ListEnterpriseSubscriptionLicenses(ctx, filters, limit)
+}
```
bobheadxi, line 38: nice, thanks 😁

sg.config.yaml
```diff
@@ -412,6 +412,9 @@ commands:
       DOTCOM_INCLUDE_PRODUCTION_LICENSES: 'true'
       # Used for authentication
       SAMS_URL: https://accounts.sgdev.org
+      PGDSN: "user='sourcegraph' password='sourcegraph' host='localhost' port='5432' dbname='{{ .Database }}' search_path='public' application_name='enterprise-portal' client_encoding=UTF8"
```
bobheadxi, line 415: Should we set this above with `DOTCOM_PGDSN_OVERRIDE`? That lets us match whatever `sg.config.overwrite.yaml` configuration devs might have, since we interpolate vars there

cmd/enterprise-portal/internal/database/subscriptions.go
unknwon: IIUC the concept of sets can be added later? We provide a smaller set of "features" to start with. Subject, object, namespace and relation are the pillars.

bobheadxi: yep!

sg.config.yaml
unknwon: Made some changes in https://github.com/sourcegraph/sourcegraph/pull/63139/commits/a6317c3d5a556db13a83e62c2aa38a8d1d0f2467, is that what you're looking for?

bobheadxi: yep!!

cmd/enterprise-portal/internal/database/migrate.go
```diff
@@ -0,0 +1,105 @@
...
+	const releaseBuildVersionExample = "277307_2024-06-06_5.4-9185da3c3e42"
+	// We always run the full auto-migration if the version is not release-build like.
+	if len(currentVersion) < len(releaseBuildVersionExample) ||
+		len(previousVersion) < len(releaseBuildVersionExample) {
+		return false
+	}
+
+	// The release build version is sorted lexicographically, so we can compare it as a string.
+	return previousVersion >= currentVersion
+}
```
bobheadxi, line 105: scary given the naming conventions are outside our control 😆 but hopefully is okay

cmd/enterprise-portal/internal/database/migrate.go
```diff
@@ -0,0 +1,105 @@
...
+	"gorm.io/plugin/opentelemetry/tracing"
+
+	"github.com/sourcegraph/sourcegraph/internal/redislock"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+	"github.com/sourcegraph/sourcegraph/lib/managedservicesplatform/runtime"
+)
+
+// maybeMigrate runs the auto-migration for the database when needed based on
+// the given version.
+func maybeMigrate(ctx context.Context, logger log.Logger, contract runtime.Contract, redisClient *redis.Client, currentVersion string) error {
```
bobheadxi, line 23: Can we put some tracing on this whole setup as well?

bobheadxi (approved):
Nice 💪

cmd/enterprise-portal/internal/database/migrate.go
unknwon: Haha we will see!

cmd/enterprise-portal/internal/database/migrate.go
unknwon: "whole setup" refers to ...?

the GORM portion already enabled tracing:
CleanShot 2024-06-06 at 18 20 49@2x

bobheadxi: Something on top of `maybeMigrate` - otherwise we'll just get a "gorm connect" span with no additional context (did we migrate? did we get the log? how long did it take to claim the redis lock? etc)

unknwon: Gotcha, like this https://github.com/sourcegraph/sourcegraph/pull/63139/commits/484e9bbecc7194c1d56a46c2b96c1825a28e2f1a?

cmd/enterprise-portal/internal/database/migrate.go
```diff
@@ -0,0 +1,116 @@
...
+			versionKey := fmt.Sprintf("%s:db_version", databaseName)
+			if shouldSkipMigration(
+				redisClient.Get(context.Background(), versionKey).Val(),
+				currentVersion,
+			) {
+				logger.Info("skipped auto-migration",
+					log.String("database", databaseName),
+					log.String("currentVersion", currentVersion),
+				)
+				return nil
```
bobheadxi, line 80: add span event + attribute here

cmd/enterprise-portal/internal/database/migrate.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// We want to make sure only one instance of the server is doing auto-migration
+	// at a time.
+	return redislock.OnlyOne(
+		logger,
+		redisClient,
+		fmt.Sprintf("%s:auto-migrate", databaseName),
+		15*time.Second,
+		func() error {
+			versionKey := fmt.Sprintf("%s:db_version", databaseName)
```
bobheadxi, line 71: if this runs, that means we claimed lock right? let's add a span event here

cmd/enterprise-portal/internal/database/migrate.go
```diff
@@ -0,0 +1,116 @@
...
+// the given version.
+func maybeMigrate(ctx context.Context, logger log.Logger, contract runtime.Contract, redisClient *redis.Client, currentVersion string) error {
+	ctx, span := databaseTracer.Start(
+		ctx,
+		"database.maybeMigrate",
+		trace.WithAttributes(
+			attribute.String("currentVersion", currentVersion),
+		),
+	)
+	defer span.End()
```
bobheadxi, line 33: also capture + record error in a defer callback (SetStatus IIRC)

cmd/enterprise-portal/internal/database/migrate.go
bobheadxi: left some comments on enriching the trace a bit more

PR #63141, opened 2024-06-06
Svelte: allow popovers to have external targets

camdencheek:
This updates the `Popover` component to accept a passed-in target rather than setting the target via an action. This allows us to pass in targets not owned by the `Popover` component, which is important for things like the file tree where the logical target is a wrapper around what the consumer has access to.

## Test plan

Manually tested. [Video walkthrough](https://share.cleanshot.com/K29H6Mvw).

### DIFF
--- a/client/web-sveltekit/src/lib/Popover.svelte
+++ b/client/web-sveltekit/src/lib/Popover.svelte
@@ -141,7 +141,6 @@
 <slot {toggle} {registerTrigger} {registerTarget} />
 {#if trigger && isOpen}
     <div
-        class="popover-container"
         use:portal
         use:onClickOutside
         use:registerPopoverContainer
@@ -160,7 +159,7 @@
 {/if}
 
 <style lang="scss">
-    .popover-container {
+    div {
         position: absolute;
         isolation: isolate;
         min-width: 10rem;
--- a/client/web-sveltekit/src/lib/Popover.svelte
+++ b/client/web-sveltekit/src/lib/Popover.svelte
@@ -54,7 +54,9 @@
 
     const registerTrigger: Action<HTMLElement> = node => {
         trigger = node
+    }
 
+    const watchTrigger: Action<HTMLElement, HTMLElement | null> = (_node, trigger) => {
         function handleMouseEnterTrigger(): void {
             clearTimeout(delayTimer)
             delayTimer = setTimeout(() => toggle(true), hoverDelay)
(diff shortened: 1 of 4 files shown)


client/web-sveltekit/src/lib/Popover.svelte
```diff
@@ -1,22 +1,24 @@
...
      * Show the popover when hovering over the trigger.
      */
     export let showOnHover: boolean = false
+    export let placement: Placement = 'bottom'
+    export let offset: OffsetOptions = showOnHover ? 0 : 3
     export let hoverDelay: number = 500
     export let hoverCloseDelay: number = 150
     export let closeOnEsc: boolean = true
+    export let trigger: HTMLElement | null = null
+    export let target: HTMLElement | undefined = undefined
```
camdencheek, line 19: Primary new change: expose `trigger` and `target` so they can be passed in directly rather than only via the `registerTrigger`/`registerTarget` actions.

client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte
```diff
@@ -124,13 +124,12 @@
                     We handle navigation via the TreeView's select event, to preserve the focus state.
                     Using a link here allows us to benefit from data preloading.
                 -->
-                <Popover let:registerTrigger placement="right-end" showOnHover>
+                <Popover trigger={label} placement="right-start" showOnHover offset={{ mainAxis: 8, crossAxis: -32 }}>
```
Comment (camdencheek) [inline client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/FileTree.svelte:127]: This is how the passed-in trigger looks. (the change in placement and offset is not related, but necessary to make this look okay)

client/web-sveltekit/src/lib/Popover.svelte
```diff
@@ -49,47 +51,63 @@
...
+    }
+
+    // Every time trigger changes (either by a props change or a call to registerTrigger)
+    // unwatch the old trigger and watch the new trigger.
+    let oldTrigger: HTMLElement | null
+    $: {
+        oldTrigger && unwatchTrigger(oldTrigger)
+        trigger && watchTrigger(trigger)
+        oldTrigger = trigger
     }
```
fkling, line 111: I hope we can eventually provide a nice API for this use case, which is basically an effect with cleanup logic. Maybe something like

```js
const trackedTrigger = track(trigger => {
    if (trigger) {
        watchTrigger(trigger)
        return () => unwatchTrigger(trigger)
    }
})
$: trackedTrigger.set(trigger)
```

fkling (approved):

client/web-sveltekit/src/lib/Popover.svelte
camdencheek: Agreed! I was honestly kinda surprised that there wasn't something built in to svelte to get the previous value of the variable.

PR #63146, opened 2024-06-07
Svelte: fix tests

camdencheek:
I disabled a few tests to get #62560 merged. This just goes back and fixes the ones I disabled. There are still a handful skipped, but the fixes were non-trivial. 

## Test plan

The unskipped tests pass

### DIFF
--- a/client/web-sveltekit/src/lib/repo/FileHeader.svelte
+++ b/client/web-sveltekit/src/lib/repo/FileHeader.svelte
@@ -63,7 +63,7 @@
 </script>
 
 <div class="header" use:sizeToFit={{ grow, shrink }}>
-    <h2 on:copy={stripSpaces}>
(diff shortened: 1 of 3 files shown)


client/web-sveltekit/src/lib/repo/FileHeader.svelte
```diff
@@ -63,7 +63,7 @@
 </script>
 
 <div class="header" use:sizeToFit={{ grow, shrink }}>
-    <h2 on:copy={stripSpaces}>
+    <h2 on:copy={stripSpaces} data-testid="header-path">
```
fkling, line 66: I suggest to be even more specific hear and include the component name in the test id. Something like

```suggestion
    <h2 on:copy={stripSpaces} data-testid="file-header-path">
```

fkling (approved):

PR #63149, opened 2024-06-07
sg: add some messaging around generating a token

burmudar:
Add some messaging around generating a token as well as when arguments are missing

## Test plan
Tested locally


janhartman (approved):
Thanks! FYI I wouldn't have minded doing this myself either btw 😄

burmudar:
> Thanks! FYI I wouldn't have minded doing this myself either btw 😄

No worries :D I was already in part of this code so thought I quickly do this here :)

PR #63151, opened 2024-06-07
chore: Fix lint error (unparam)

varungandhi-src:
It looks like this is broken on main; got a CI error: https://buildkite.com/sourcegraph/sourcegraph/builds/277526

## Test plan

Covered by existing tests

## Changelog

### DIFF
--- a/cmd/frontend/internal/cli/sysreq.go
+++ b/cmd/frontend/internal/cli/sysreq.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/sourcegraph/sourcegraph/internal/env"
 	"github.com/sourcegraph/sourcegraph/internal/sysreq"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
 )
 
 const skipSysReqsEnvVar = "SRC_SKIP_REQS"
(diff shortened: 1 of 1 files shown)


cmd/frontend/internal/cli/sysreq.go
```diff
@@ -58,5 +58,5 @@ func checkSysReqs(ctx context.Context, w io.Writer) error {
 	if failed != nil {
 		log15.Error("System requirement checks failed (see above for more information).", "failed", failed)
 	}
-	return nil
+	return failed
```
varungandhi-src, line 61: The function doc comment says:

```
// checkSysReqs uses package sysreq to check for the presence of
// system requirements. If any are missing, it prints a message to
// w and returns a non-nil error.
```

burmudar (approved):

cmd/frontend/internal/cli/sysreq.go
varungandhi-src: Oops, `failed` is `[]string`, not `error`

varungandhi-src: Tweaked to return non-nil error when `failed != nil`

PR #63157, opened 2024-06-07
release: never use build number in image family

burmudar:
the executor image and docker mirror image should now follow the following naming convention:

Image family: `sourcegraph-executors-[nightly|internal|'']-<MAJOR>-<MINOR>`
Image name: `sourcegraph-executor-[nightly|internal|'']-<MAJOR>-<MINOR>-<BUILD_NUMBER>`

example:
Image family: `sourcegraph-executors-5-4`
Image name: `sourcegraph-executor-5-4-277666`

#### Nightly
**`nightly` suffix**
Image family: `sourcegraph-executors-nightly-<MAJOR>-<MINOR>`
Image name: `sourcegraph-executor-nightly-<MAJOR>-<MINOR>-<BUILD_NUMBER>`
#### Internal
**`internal` suffix**
Image family: `sourcegraph-executors-internal-<MAJOR>-<MINOR>`
Image name: `sourcegraph-executor-internal-<MAJOR>-<MINOR>-<BUILD_NUMBER>`
#### Public / Promote to public

** No suffix **

Image family: `sourcegraph-executors-<MAJOR>-<MINOR>`
Image name: `sourcegraph-executor-<MAJOR>-<MINOR>-<BUILD_NUMBER>`

>  [!IMPORTANT]
> Should we keep the imagine name stable at `sourcegraph-executor-<MAJOR>-<MINOR>-<BUILD_NUMBER>`
> and only change the family name? 
>
> **Why?**
>
> The Image family dictates the collection of images and that changes each major minor and or release phase so there is really no use in changing the image name too, except at a glance you can see from the name what image family it belongs to?

### DIFF
--- a/dev/ci/internal/ci/executor_operations.go
+++ b/dev/ci/internal/ci/executor_operations.go
@@ -124,7 +124,7 @@ func executorDockerMirrorImageFamilyForConfig(c Config) string {
 		if err != nil {
 			panic("cannot parse version")
 		}
-		imageFamily = fmt.Sprintf("sourcegraph-executors-internal-docker-mirror-%d-%d-%d", ver.Major(), ver.Minor(), ver.Patch())
+		imageFamily = fmt.Sprintf("sourcegraph-executors-internal-docker-mirror-%d-%d", ver.Major(), ver.Minor())
 	}
 	return imageFamily
 }
@@ -146,7 +146,7 @@ func executorImageFamilyForConfig(c Config) string {
 		if err != nil {
 			panic("cannot parse version")
 		}
-		imageFamily = fmt.Sprintf("sourcegraph-executors-internal-%d-%d-%d", ver.Major(), ver.Minor(), ver.Patch())
+		imageFamily = fmt.Sprintf("sourcegraph-executors-internal-%d-%d", ver.Major(), ver.Minor())
 	}
 	return imageFamily
 }
--- a/cmd/executor/docker-mirror/_ami.build.sh
+++ b/cmd/executor/docker-mirror/_ami.build.sh
@@ -19,11 +19,7 @@ GCP_PROJECT="aspect-dev"
 "$gcloud" secrets versions access latest --secret=e2e-builder-sa-key --quiet --project="$GCP_PROJECT" >"workdir/builder-sa-key.json"
 
 ## Setting up packer
-export PKR_VAR_name
-PKR_VAR_name="${IMAGE_FAMILY}"
-if [ "${RELEASE_INTERNAL:-}" == "true" ]; then
-  PKR_VAR_name="${PKR_VAR_name}-${BUILDKITE_BUILD_NUMBER}"
-fi
+export PKR_VAR_name="${IMAGE_FAMILY}-${BUILDKITE_BUILD_NUMBER}"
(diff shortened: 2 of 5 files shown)


eseliger (approved):
the PR description makes sense to me

dev/ci/internal/ci/executor_operations.go
```diff
@@ -146,7 +146,7 @@ func executorImageFamilyForConfig(c Config) string {
 		if err != nil {
 			panic("cannot parse version")
 		}
-		imageFamily = fmt.Sprintf("sourcegraph-executors-internal-%d-%d-%d", ver.Major(), ver.Minor(), ver.Patch())
+		imageFamily = fmt.Sprintf("sourcegraph-executors-internal-%d-%d", ver.Major(), ver.Minor())
```
craigfurman, line 149: do we ever actually want these internal names? 

```suggestion
		imageFamily = fmt.Sprintf("sourcegraph-executors-%d-%d", ver.Major(), ver.Minor())
```

And similar for the docker-mirror. If I understood correctly.

craigfurman (commented):
1 question, because I'm not 100% sure this works 🤔 

I think we'll need to backport this too once we merge it, but I'm not 100% sure! I'd imagine ami.push.sh is run from the release branch, hence why I'd guess yes.

dev/ci/internal/ci/executor_operations.go
craigfurman: I did read the PR description btw, I'm just not getting why 😅

burmudar: Yes we do - **without** internal in the image family name an executor meant for internal releases *only* will be published with the image family name `sourcegraph-executors` which means the terraform in `terraform-aws` and `terraform-google` who look for latest images in the image family `sourcegraph-executors-<MAJOR>-<MINOR>` will pick the image up _after_ every `MAJOR.MINOR` bump.

So ultimately, internal releases are kept to the `sourcegraph-executors-internal` and once you promote to public the image is rebuilt and put into `sourcegraph-executors` which means the released terraform will now pick it up.

This is also has a nice quality which I think the release team can investigate. This technically allows us to test internal releases _with the terraform. If you want to opt into an internal release you just have to update the image family in the terraform - almost like `alpha`, `beta` changes.

Sorry if I am over explaining it! Not my intention at all.

The major caveat not mentioned here are at all, is that you kinda just have to _know_ is that you'll only see the `sourcegraph-executor-MAJOR-MINOR` release once *the tagged release build finishes* which flies completely under the radar - which I think we need to fix somehow.

One idea I just got on how this _can_ be done is part of the promote to public release steps in the release manifest is to run:
```bash
# promote to public 
gcloud compute images create --source-image=sourcegraph-executors-internal-MAJOR-MINOR-BUILD sourcegraph-executors-MAJOR-MINOR-4-BUILD --family sourcegraph-executors-MAJOR-MINOR
```

This stays _more_ true to the spirit of promotion, in that a new image is not built by CI, instead the image used for internal releases just gets copied to a new name.

burmudar: Awesome 👏🏼 

Yup - this should be backported ♻️

PR #63158, opened 2024-06-07
feat(appliance): deploy frontend

craigfurman:
**Add appliance deploy type**

**feat(appliance): deploy frontend**

## Test plan

Golden tests included.

## Changelog

- Appliance (still unreleased) can deploy frontend.


internal/conf/deploy/deploytype.go
```diff
@@ -46,7 +47,7 @@ func Mock(val string) {
 func IsDeployTypeKubernetes(deployType string) bool {
 	switch deployType {
 	// includes older Kubernetes aliases for backwards compatibility
-	case "k8s", "cluster", Kubernetes, Helm, Kustomize, K3s:
+	case "k8s", "cluster", Kubernetes, Helm, Kustomize, K3s, Appliance:
```
craigfurman, line 50: Note. This is also changed in jscontext.ts as instructed by a comment in this file.

internal/appliance/reconciler/frontend_test.go
```diff
@@ -0,0 +1,18 @@
+package reconciler
+
+func (suite *ApplianceTestSuite) TestDeployFrontend() {
+	for _, tc := range []struct {
+		name string
+	}{
+		{name: "frontend/default"},
+		{name: "frontend/with-blobstore"},
+		{name: "frontend/with-ingress"},
```
craigfurman, line 9: Annotated output of:

```
  -deploy-sourcegraph-helm-path ../../deploy-sourcegraph-helm \
  -component frontend \
  -golden-file internal/appliance/reconciler/testdata/golden-fixtures/frontend/with-ingress.yaml \
  -helm-template-extra-args '--set blobstore.enabled=false' | pbcopy
```

Blobstore is disabled for ease of reading, otherwise we'd have to scroll through
all blobstore resources too.


```diff
2c2
< # helm: ServiceAccount/sourcegraph-frontend
---
> # golden: ServiceAccount/sourcegraph-frontend
5a6,8
>   annotations:
>     appliance.sourcegraph.com/configHash: d3eb623947fedba566dfc56adc4733ff5ef1b2887a5cb63d75dbc1df452b0b5c
>   creationTimestamp: "2024-04-19T00:00:00Z"
7,8d9
<     app.kubernetes.io/component: frontend
<     category: rbac
10a12,21
>   namespace: NORMALIZED_FOR_TESTING
>   ownerReferences:
>     - apiVersion: v1
>       blockOwnerDeletion: true
>       controller: true
>       kind: ConfigMap
>       name: sg
>       uid: NORMALIZED_FOR_TESTING
>   resourceVersion: NORMALIZED_FOR_TESTING
>   uid: NORMALIZED_FOR_TESTING
12c23
< # helm: Role/sourcegraph-frontend
---
> # golden: Role/sourcegraph-frontend
15a27,29
>   annotations:
>     appliance.sourcegraph.com/configHash: d3eb623947fedba566dfc56adc4733ff5ef1b2887a5cb63d75dbc1df452b0b5c
>   creationTimestamp: "2024-04-19T00:00:00Z"
17,22d30
<     app.kubernetes.io/component: frontend
<     app.kubernetes.io/instance: release-name
<     app.kubernetes.io/managed-by: Helm
<     app.kubernetes.io/name: sourcegraph
<     app.kubernetes.io/version: 5.3.9104
<     category: rbac
24d31
<     helm.sh/chart: sourcegraph-5.3.9104
25a33,42
>   namespace: NORMALIZED_FOR_TESTING
>   ownerReferences:
>     - apiVersion: v1
>       blockOwnerDeletion: true
>       controller: true
>       kind: ConfigMap
>       name: sg
>       uid: NORMALIZED_FOR_TESTING
>   resourceVersion: NORMALIZED_FOR_TESTING
>   uid: NORMALIZED_FOR_TESTING
45c62
< # helm: RoleBinding/sourcegraph-frontend
---
> # golden: RoleBinding/sourcegraph-frontend
48a66,68
>   annotations:
>     appliance.sourcegraph.com/configHash: d3eb623947fedba566dfc56adc4733ff5ef1b2887a5cb63d75dbc1df452b0b5c
>   creationTimestamp: "2024-04-19T00:00:00Z"
50,55d69
<     app.kubernetes.io/component: frontend
<     app.kubernetes.io/instance: release-name
<     app.kubernetes.io/managed-by: Helm
<     app.kubernetes.io/name: sourcegraph
<     app.kubernetes.io/version: 5.3.9104
<     category: rbac
57d70
<     helm.sh/chart: sourcegraph-5.3.9104
58a72,81
>   namespace: NORMALIZED_FOR_TESTING
>   ownerReferences:
>     - apiVersion: v1
>       blockOwnerDeletion: true
>       controller: true
>       kind: ConfigMap
>       name: sg
>       uid: NORMALIZED_FOR_TESTING
>   resourceVersion: NORMALIZED_FOR_TESTING
>   uid: NORMALIZED_FOR_TESTING
66c89
<     namespace: default
---
>     namespace: NORMALIZED_FOR_TESTING
68c91
< # helm: Service/sourcegraph-frontend-internal
---
> # golden: Service/sourcegraph-frontend-internal
71a95,97
>   annotations:
>     appliance.sourcegraph.com/configHash: d3eb623947fedba566dfc56adc4733ff5ef1b2887a5cb63d75dbc1df452b0b5c
>   creationTimestamp: "2024-04-19T00:00:00Z"
73,74c99,100
<     app: sourcegraph-frontend
<     app.kubernetes.io/component: frontend
---
>     app: sourcegraph-frontend-internal
>     app.kubernetes.io/component: sourcegraph-frontend-internal
76a103,112
>   namespace: NORMALIZED_FOR_TESTING
>   ownerReferences:
>     - apiVersion: v1
>       blockOwnerDeletion: true
>       controller: true
>       kind: ConfigMap
>       name: sg
>       uid: NORMALIZED_FOR_TESTING
>   resourceVersion: NORMALIZED_FOR_TESTING
>   uid: NORMALIZED_FOR_TESTING
77a114,120
>   clusterIP: NORMALIZED_FOR_TESTING
>   clusterIPs:
>     - NORMALIZED_FOR_TESTING
>   internalTrafficPolicy: Cluster
>   ipFamilies:
>     - IPv4
>   ipFamilyPolicy: SingleStack
80a124
>       protocol: TCP
84,85c128
<     app.kubernetes.io/instance: release-name
<     app.kubernetes.io/name: sourcegraph
---
>   sessionAffinity: None
86a130,131
> status:
>   loadBalancer: {}
88c133
< # helm: Service/sourcegraph-frontend
---
> # golden: Service/sourcegraph-frontend
92a138
>     appliance.sourcegraph.com/configHash: d3eb623947fedba566dfc56adc4733ff5ef1b2887a5cb63d75dbc1df452b0b5c
94a141
>   creationTimestamp: "2024-04-19T00:00:00Z"
97c144
<     app.kubernetes.io/component: frontend
---
>     app.kubernetes.io/component: sourcegraph-frontend
99a147,156
>   namespace: NORMALIZED_FOR_TESTING
>   ownerReferences:
>     - apiVersion: v1
>       blockOwnerDeletion: true
>       controller: true
>       kind: ConfigMap
>       name: sg
>       uid: NORMALIZED_FOR_TESTING
>   resourceVersion: NORMALIZED_FOR_TESTING
>   uid: NORMALIZED_FOR_TESTING
100a158,164
>   clusterIP: NORMALIZED_FOR_TESTING
>   clusterIPs:
>     - NORMALIZED_FOR_TESTING
>   internalTrafficPolicy: Cluster
>   ipFamilies:
>     - IPv4
>   ipFamilyPolicy: SingleStack
103a168
>       protocol: TCP
107,108c172
<     app.kubernetes.io/instance: release-name
<     app.kubernetes.io/name: sourcegraph
---
>   sessionAffinity: None
109a174,175
> status:
>   loadBalancer: {}
111c177
< # helm: Deployment/sourcegraph-frontend
---
> # golden: Deployment/sourcegraph-frontend
116c182,184
<     description: Serves the frontend of Sourcegraph via HTTP(S).
---
>     appliance.sourcegraph.com/configHash: d3eb623947fedba566dfc56adc4733ff5ef1b2887a5cb63d75dbc1df452b0b5c
>   creationTimestamp: "2024-04-19T00:00:00Z"
>   generation: 1
118,120c186
<     app.kubernetes.io/component: frontend
<     app.kubernetes.io/instance: release-name
<     app.kubernetes.io/managed-by: Helm
---
>     app.kubernetes.io/component: sourcegraph-frontend
124d189
<     helm.sh/chart: sourcegraph-5.3.9104
125a191,200
>   namespace: NORMALIZED_FOR_TESTING
>   ownerReferences:
>     - apiVersion: v1
>       blockOwnerDeletion: true
>       controller: true
>       kind: ConfigMap
>       name: sg
>       uid: NORMALIZED_FOR_TESTING
>   resourceVersion: NORMALIZED_FOR_TESTING
>   uid: NORMALIZED_FOR_TESTING
127a203
>   progressDeadlineSeconds: 600
133,134d208
<       app.kubernetes.io/instance: release-name
<       app.kubernetes.io/name: sourcegraph
143,145c217,218
<         checksum/auth: cffe7781e068ed7f9b5cff231361a3973a2a3378a28cf047a9a5ad77c832732c
<         checksum/redis: 63b58e05a2640417d599c4aee6d866cb9063e3a9aa452dc08dbfff836b7781b7
<         kubectl.kubernetes.io/default-container: frontend
---
>         kubectl.kubernetes.io/default-container: sourcegraph-frontend
>       creationTimestamp: null
148,149d220
<         app.kubernetes.io/instance: release-name
<         app.kubernetes.io/name: sourcegraph
150a222
>       name: sourcegraph-frontend
152d223
<       affinity: null
158,164c229
<               value: helm
<             - name: GRAFANA_SERVER_URL
<               value: http://grafana:30070
<             - name: JAEGER_SERVER_URL
<               value: http://jaeger-query:16686
<             - name: PROMETHEUS_URL
<               value: http://prometheus:30090
---
>               value: appliance
```

- DEPLOY_TYPE=appliance, instead of helm. See first commit in this PR.
- Unlike Helm, we don't unconditionally set the 3 o11y-endpoint-related env
  vars. This commit only sets them when the relevant integrations are enabled.
  Grafana and Jaeger haven't been implemented in the appliance yet, but I've set
  them to disabled-by-default. We can revisit this. Prometheus is disabled in
  the config for this golden test.

```diff
252a318
>                   apiVersion: v1
256c322
<           image: index.docker.io/sourcegraph/frontend:5.3.9104@sha256:5356f0a0402424ac5ea3b69ab187bc09fd6440e0605dd696f619f1c8119d9bc9
---
>           image: index.docker.io/sourcegraph/frontend:5.3.9104
258a325
>             failureThreshold: 3
263a331,332
>             periodSeconds: 10
>             successThreshold: 1
268a338
>               protocol: TCP
270a341
>               protocol: TCP
272a344
>               protocol: TCP
273a346
>             failureThreshold: 3
278a352
>             successThreshold: 1
293a368
>           terminationMessagePath: /dev/termination-log
297a373
>       dnsPolicy: ClusterFirst
301a378,379
>             - name: DEPLOY_TYPE
>               value: appliance
377,385c455
<             - name: DEPLOY_TYPE
<               value: helm
<             - name: GRAFANA_SERVER_URL
<               value: http://grafana:30070
<             - name: JAEGER_SERVER_URL
<               value: http://jaeger-query:16686
<             - name: PROMETHEUS_URL
<               value: http://prometheus:30090
<           image: index.docker.io/sourcegraph/migrator:5.3.9104@sha256:c5b3af9e0d5734768c8f0960014ae0d193ee4cfe4f9f7cad9f729b93876ba4c2
---
>           image: index.docker.io/sourcegraph/migrator:5.3.9104
400,402c470,479
<           volumeMounts: null
<       nodeSelector: null
<       securityContext: {}
---
>           terminationMessagePath: /dev/termination-log
>           terminationMessagePolicy: FallbackToLogsOnError
>       restartPolicy: Always
>       schedulerName: default-scheduler
>       securityContext:
>         fsGroup: 101
>         fsGroupChangePolicy: OnRootMismatch
>         runAsGroup: 101
>         runAsUser: 100
>       serviceAccount: sourcegraph-frontend
404c481
<       tolerations: null
---
>       terminationGracePeriodSeconds: 30
407a485
> status: {}
409c487
< # helm: Ingress/sourcegraph-frontend
---
> # golden: Ingress/sourcegraph-frontend
414,415c492,494
<     kubernetes.io/ingress.class: nginx
<     nginx.ingress.kubernetes.io/proxy-body-size: 150m
---
>     appliance.sourcegraph.com/configHash: d3eb623947fedba566dfc56adc4733ff5ef1b2887a5cb63d75dbc1df452b0b5c
>   creationTimestamp: "2024-04-19T00:00:00Z"
>   generation: 1
```

- The `kubernetes.io/ingress.class` annotation is deprecated:
  https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/3d49c6def4989983725cb23b2058fd3d42c90ea4/charts/sourcegraph/values.yaml#L347-L349.
  It is superseded by ingress.spec.ingressclassname.
- `nginx.ingress.kubernetes.io/proxy-body-size` is nginx-specific, so IMO it
  doesn't make sense to set it by deafult. Paved-path config for various
  ingresses, including nginx, would make a good appliance feature though. I've
  opened
  https://linear.app/sourcegraph/issue/REL-151/paved-path-appliance-configurations-for-various-ingresses
  as an early placeholder for this.

```diff
417,418d495
<     app: sourcegraph-frontend
<     app.kubernetes.io/component: frontend
420a498,507
>   namespace: NORMALIZED_FOR_TESTING
>   ownerReferences:
>     - apiVersion: v1
>       blockOwnerDeletion: true
>       controller: true
>       kind: ConfigMap
>       name: sg
>       uid: NORMALIZED_FOR_TESTING
>   resourceVersion: NORMALIZED_FOR_TESTING
>   uid: NORMALIZED_FOR_TESTING
431a519,594
> status:
>   loadBalancer: {}
> ---
> # golden: ConfigMap/sg
> apiVersion: v1
> data:
>   spec: |
>     spec:
>       requestedVersion: "5.3.9104"
> 
>       blobstore:
>         disabled: true
> 
>       codeInsights:
>         disabled: true
> 
>       codeIntel:
>         disabled: true
> 
>       frontend:
>         ingress: {}
> 
>       gitServer:
>         disabled: true
> 
>       indexedSearch:
>         disabled: true
> 
>       indexedSearchIndexer:
>         disabled: true
> 
>       pgsql:
>         disabled: true
> 
>       postgresExporter:
>         disabled: true
> 
>       preciseCodeIntel:
>         disabled: true
> 
>       redisCache:
>         disabled: true
> 
>       redisStore:
>         disabled: true
> 
>       repoUpdater:
>         disabled: true
> 
>       searcher:
>         disabled: true
> 
>       symbols:
>         disabled: true
> 
>       syntectServer:
>         disabled: true
> 
>       worker:
>         disabled: true
> 
>       prometheus:
>         disabled: true
> 
>       embeddings:
>         disabled: true
> kind: ConfigMap
> metadata:
>   annotations:
>     appliance.sourcegraph.com/currentVersion: 5.3.9104
>     appliance.sourcegraph.com/managed: "true"
>   creationTimestamp: "2024-04-19T00:00:00Z"
>   name: sg
>   namespace: NORMALIZED_FOR_TESTING
>   resourceVersion: NORMALIZED_FOR_TESTING
>   uid: NORMALIZED_FOR_TESTING
```

internal/appliance/reconciler/frontend_test.go
```diff
@@ -0,0 +1,18 @@
...
+		{name: "frontend/with-ingress-optional-fields"},
```
craigfurman, line 10:
```
diff internal/appliance/reconciler/testdata/golden-fixtures/frontend/with-ingress.yaml internal/appliance/reconciler/testdata/golden-fixtures/frontend/with-ingress-optional-fields.yaml
```

```diff
6c6
<         appliance.sourcegraph.com/configHash: d3eb623947fedba566dfc56adc4733ff5ef1b2887a5cb63d75dbc1df452b0b5c
---
>         appliance.sourcegraph.com/configHash: 7caff941f7756c1a8aa77fb1604c7c1d191868889bbe1ca514eac63a4c1aafc8
326c326,331
<             ingress: {}
---
>             ingress:
>               host: example.com
>               annotations:
>                 foo: bar
>               ingressClassName: an-ingress-class
>               tlsSecret: ingress-tls-secret
386c391
<         appliance.sourcegraph.com/configHash: d3eb623947fedba566dfc56adc4733ff5ef1b2887a5cb63d75dbc1df452b0b5c
---
>         appliance.sourcegraph.com/configHash: 7caff941f7756c1a8aa77fb1604c7c1d191868889bbe1ca514eac63a4c1aafc8
423c428
<         appliance.sourcegraph.com/configHash: d3eb623947fedba566dfc56adc4733ff5ef1b2887a5cb63d75dbc1df452b0b5c
---
>         appliance.sourcegraph.com/configHash: 7caff941f7756c1a8aa77fb1604c7c1d191868889bbe1ca514eac63a4c1aafc8
450c455
<         appliance.sourcegraph.com/configHash: d3eb623947fedba566dfc56adc4733ff5ef1b2887a5cb63d75dbc1df452b0b5c
---
>         appliance.sourcegraph.com/configHash: 7caff941f7756c1a8aa77fb1604c7c1d191868889bbe1ca514eac63a4c1aafc8
469c474
<         appliance.sourcegraph.com/configHash: d3eb623947fedba566dfc56adc4733ff5ef1b2887a5cb63d75dbc1df452b0b5c
---
>         appliance.sourcegraph.com/configHash: 7caff941f7756c1a8aa77fb1604c7c1d191868889bbe1ca514eac63a4c1aafc8
511c516
<         appliance.sourcegraph.com/configHash: d3eb623947fedba566dfc56adc4733ff5ef1b2887a5cb63d75dbc1df452b0b5c
---
>         appliance.sourcegraph.com/configHash: 7caff941f7756c1a8aa77fb1604c7c1d191868889bbe1ca514eac63a4c1aafc8
551c556,557
<         appliance.sourcegraph.com/configHash: d3eb623947fedba566dfc56adc4733ff5ef1b2887a5cb63d75dbc1df452b0b5c
---
>         appliance.sourcegraph.com/configHash: 7caff941f7756c1a8aa77fb1604c7c1d191868889bbe1ca514eac63a4c1aafc8
>         foo: bar
567a574
>       ingressClassName: an-ingress-class
569c576,577
<         - http:
---
>         - host: example.com
>           http:
577a586,589
>       tls:
>         - hosts:
>             - example.com
>           secretName: ingress-tls-secret

```

internal/appliance/reconciler/frontend_test.go
```diff
@@ -0,0 +1,18 @@
...
+		{name: "frontend/with-overrides"},
```
craigfurman, line 11:
```
❯ diff internal/appliance/reconciler/testdata/golden-fixtures/frontend/default.yaml internal/appliance/reconciler/testdata/golden-fixtures/frontend/with-overrides.yaml
```

```diff
6c6
<         appliance.sourcegraph.com/configHash: b5dce290e22d1afb4c9102ac4c245490ab01dd3be13653de391536cfe0e323b0
---
>         appliance.sourcegraph.com/configHash: f6325ffd3262b0ff8bdb406ba80aabad2daea7ecc342353fabd7bbda7ea1f4a9
28c28
<       replicas: 2
---
>       replicas: 9
198,295d197
<           initContainers:
<             - args:
<                 - up
<               env:
<                 - name: DEPLOY_TYPE
<                   value: appliance
<                 - name: PGDATABASE
<                   valueFrom:
<                     secretKeyRef:
<                       key: database
<                       name: pgsql-auth
<                 - name: PGHOST
<                   valueFrom:
<                     secretKeyRef:
<                       key: host
<                       name: pgsql-auth
<                 - name: PGPASSWORD
<                   valueFrom:
<                     secretKeyRef:
<                       key: password
<                       name: pgsql-auth
<                 - name: PGPORT
<                   valueFrom:
<                     secretKeyRef:
<                       key: port
<                       name: pgsql-auth
<                 - name: PGUSER
<                   valueFrom:
<                     secretKeyRef:
<                       key: user
<                       name: pgsql-auth
<                 - name: CODEINTEL_PGDATABASE
<                   valueFrom:
<                     secretKeyRef:
<                       key: database
<                       name: codeintel-db-auth
<                 - name: CODEINTEL_PGHOST
<                   valueFrom:
<                     secretKeyRef:
<                       key: host
<                       name: codeintel-db-auth
<                 - name: CODEINTEL_PGPASSWORD
<                   valueFrom:
<                     secretKeyRef:
<                       key: password
<                       name: codeintel-db-auth
<                 - name: CODEINTEL_PGPORT
<                   valueFrom:
<                     secretKeyRef:
<                       key: port
<                       name: codeintel-db-auth
<                 - name: CODEINTEL_PGUSER
<                   valueFrom:
<                     secretKeyRef:
<                       key: user
<                       name: codeintel-db-auth
<                 - name: CODEINSIGHTS_PGDATABASE
<                   valueFrom:
<                     secretKeyRef:
<                       key: database
<                       name: codeinsights-db-auth
<                 - name: CODEINSIGHTS_PGHOST
<                   valueFrom:
<                     secretKeyRef:
<                       key: host
<                       name: codeinsights-db-auth
<                 - name: CODEINSIGHTS_PGPASSWORD
<                   valueFrom:
<                     secretKeyRef:
<                       key: password
<                       name: codeinsights-db-auth
<                 - name: CODEINSIGHTS_PGPORT
<                   valueFrom:
<                     secretKeyRef:
<                       key: port
<                       name: codeinsights-db-auth
<                 - name: CODEINSIGHTS_PGUSER
<                   valueFrom:
<                     secretKeyRef:
<                       key: user
<                       name: codeinsights-db-auth
<               image: index.docker.io/sourcegraph/migrator:5.3.9104
<               imagePullPolicy: IfNotPresent
<               name: migrator
<               resources:
<                 limits:
<                   cpu: 500m
<                   memory: 100M
<                 requests:
<                   cpu: 100m
<                   memory: 50M
<               securityContext:
<                 allowPrivilegeEscalation: false
<                 readOnlyRootFilesystem: true
<                 runAsGroup: 101
<                 runAsUser: 100
<               terminationMessagePath: /dev/termination-log
<               terminationMessagePolicy: FallbackToLogsOnError
325c227,229
<           frontend: {}
---
>           frontend:
>             replicas: 9
>             migrator: false
385c289
<         appliance.sourcegraph.com/configHash: b5dce290e22d1afb4c9102ac4c245490ab01dd3be13653de391536cfe0e323b0
---
>         appliance.sourcegraph.com/configHash: f6325ffd3262b0ff8bdb406ba80aabad2daea7ecc342353fabd7bbda7ea1f4a9
422c326
<         appliance.sourcegraph.com/configHash: b5dce290e22d1afb4c9102ac4c245490ab01dd3be13653de391536cfe0e323b0
---
>         appliance.sourcegraph.com/configHash: f6325ffd3262b0ff8bdb406ba80aabad2daea7ecc342353fabd7bbda7ea1f4a9
449c353
<         appliance.sourcegraph.com/configHash: b5dce290e22d1afb4c9102ac4c245490ab01dd3be13653de391536cfe0e323b0
---
>         appliance.sourcegraph.com/configHash: f6325ffd3262b0ff8bdb406ba80aabad2daea7ecc342353fabd7bbda7ea1f4a9
468c372
<         appliance.sourcegraph.com/configHash: b5dce290e22d1afb4c9102ac4c245490ab01dd3be13653de391536cfe0e323b0
---
>         appliance.sourcegraph.com/configHash: f6325ffd3262b0ff8bdb406ba80aabad2daea7ecc342353fabd7bbda7ea1f4a9
510c414
<         appliance.sourcegraph.com/configHash: b5dce290e22d1afb4c9102ac4c245490ab01dd3be13653de391536cfe0e323b0
---
>         appliance.sourcegraph.com/configHash: f6325ffd3262b0ff8bdb406ba80aabad2daea7ecc342353fabd7bbda7ea1f4a9

```

craigfurman (commented):
compare-helm, plus some diffs between different configs. Please take a microscope to these. I already have, but 4 eyes are better than 2 🙏

internal/appliance/reconciler/frontend.go
```diff
@@ -0,0 +1,304 @@
...
+	"github.com/sourcegraph/sourcegraph/lib/pointers"
+)
+
+const (
+	pgsqlSecretName          = "pgsql-auth"
+	codeInsightsDBSecretName = "codeinsights-db-auth"
+	codeIntelDBSecretName    = "codeintel-db-auth"
+)
+
+func (r *Reconciler) reconcileFrontend(ctx context.Context, sg *config.Sourcegraph, owner client.Object) error {
```
craigfurman, line 33: Divergence from the helm chart: I've removed support for the deprecated+removed PodDisruptionBudget altogether. I reckon this is probably fine, but noting it. It's not deployed by default in Helm, but it is optional.

internal/appliance/reconciler/frontend.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+	if err := r.reconcileFrontendServiceAccount(ctx, sg, owner); err != nil {
+		return errors.Wrap(err, "reconciling ServiceAccount")
+	}
+	if err := r.reconcileFrontendRole(ctx, sg, owner); err != nil {
+		return errors.Wrap(err, "reconciling Role")
+	}
+	if err := r.reconcileFrontendRoleBinding(ctx, sg, owner); err != nil {
+		return errors.Wrap(err, "reconciling RoleBinding")
+	}
```
craigfurman, line 51: I've diverged a bit from the helm chart here. Helm has a `frontend.privileged` flag, that is true by default. It's bit odd though: when true, it provisions a role that has the same privileges as configured here: https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/main/charts/sourcegraph/templates/frontend/sourcegraph-frontend.Role.yaml

When false, it binds the ClusterRole "view" instead: https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/main/charts/sourcegraph/templates/frontend/sourcegraph-frontend.RoleBinding.yaml

This seems odd to me - as if the privileged flag is backwards. Having a ClusterRole bound would give _more_ read-only privilege, over a similar set of resource kinds.

Other parts of the appliance, e.g. prometheus config, use "privileged" to refer to cluster-level scope and/or privileged container SecurityContext. In that spirit, this commit unconditionally applies the "privileged" frontend config (a namespace-scoped role with read-only access to certain resources), because that doesn't actually appear to _be_ privileged.

craigfurman (commented):
More notes on divergence from helm.

internal/appliance/reconciler/reconcile.go
```diff
@@ -111,6 +111,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	if err := r.reconcileWorker(ctx, &sourcegraph, &applianceSpec); err != nil {
 		return ctrl.Result{}, errors.Newf("failed to reconcile worker: %w", err)
 	}
+	if err := r.reconcileFrontend(ctx, &sourcegraph, &applianceSpec); err != nil {
+		return ctrl.Result{}, errors.Newf("failed to reconcile worker: %w", err)
```
craigfurman, line 115:
```suggestion
		return ctrl.Result{}, errors.Newf("failed to reconcile frontend: %w", err)
```

internal/appliance/reconciler/frontend.go
jdpleiness: Should we flip this around to make sense then? As in make the privileged flag be "actually" privileged and don't set it to privileged by default? I also can't think of any reason off the top of my head `frontend` would _need_ to be privileged 🤔

jdpleiness (approved):
Nothing blocking, just some comments around our weird frontend privilege flags

internal/appliance/reconciler/frontend.go
craigfurman: > Should we flip this around to make sense then?

I assume you mean in the helm chart? Because in this appliance PR, there is no such flag added. I've unconditionally gone from the route _called_ privileged in helm, which actually seems less privileged to me 😅 

We could invert it in helm, but really we could remove it altogether, doing something similar to what the appliance does (after we convince ourselves that won't break anything, of course...)

jdpleiness: >We could invert it in helm, but really we could remove it altogether, doing something similar to what the appliance does (after we convince ourselves that won't break anything, of course...)
Sounds good to me. Looking quickly at the chart, the only thing I can think of is this was _likely_ a bug rather than intentional 🐛

craigfurman: It does look like it, for sure 🙏

PR #63163, opened 2024-06-07
feat(cody-gateway): add Google Gemini stable models to allowed models

abeatrix:
This change adds the stable versions of the Google Gemini 1.5 Flash, Gemini 1.5 Pro, and Gemini Pro models to the list of allowed models in the Cody Gateway configuration and API endpoints.

The changes are made in the following files:
- `cmd/cody-gateway/shared/config/config.go`
- `cmd/frontend/internal/dotcom/productsubscription/codygateway_dotcom_user.go`
- `cmd/frontend/internal/httpapi/completions/chat.go`
- `internal/conf/computed.go`

The new models are added to the allowed model lists and also set as the default chat, fast chat, and completion models when the configuration is not explicitly set.

This change ensures that users are defaulted to use the stable versions of the Google Gemini models with the latest versions are still being available for use in the Cody Gateway application.

## Test plan

Changes are covered by current tests.

### DIFF
--- a/cmd/cody-gateway/shared/config/config.go
+++ b/cmd/cody-gateway/shared/config/config.go
@@ -308,6 +308,9 @@ func (c *Config) Load() {
 			google.Gemini15FlashLatest,
 			google.Gemini15ProLatest,
 			google.GeminiProLatest,
+			google.Gemini15Flash,
+			google.Gemini15Pro,
+			google.GeminiPro,
 		}, ","),
 		"Google models that can to be used."),
 	)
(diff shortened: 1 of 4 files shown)


chrsmith (approved):
I have no idea if this is the complete list, of if you need to add Gemini to a few more places. 🤷 But this LGTM.

When this is rolled out, I believe the last (and final?) step is to just add Gemini to the `shared/model` package's `dotcom.ts` in the `sourcegraph/cody` repo? And then it's totally official?

abeatrix:
> I have no idea if this is the complete list, of if you need to add Gemini to a few more places. 🤷 But this LGTM.
> 
> When this is rolled out, I believe the last (and final?) step is to just add Gemini to the `shared/model` package's `dotcom.ts` in the `sourcegraph/cody` repo? And then it's totally official?

Thanks for the review and YES! I believe it's official after that!!

PR #63169, opened 2024-06-07
chore/cody-gateway: instrument removeUnseenTokens

bobheadxi:
Sync spans seem to hang for quite a while after getting data from the source - this adds instrumentation on the potential cause, `removeUnseenTokens`, so that we can get a bit more detail from the traces:

image

## Test plan

n/a


chrsmith (approved):
:shipit #WCGW

PR #63167, opened 2024-06-07
Svelte: add Cody upsell banner

camdencheek:
Fixes SRCH-492

screenshot-2024-06-07_13-53-54@2x
screenshot-2024-06-07_13-53-47@2x

PR #63170, opened 2024-06-07
Svelte: add footer links for dotcom

camdencheek:
Stacked on #63167 
Completes SRCH-493

## Test plan

https://share.cleanshot.com/LLQCYdTY


client/web-sveltekit/src/routes/search/+page.svelte
```diff
@@ -26,6 +27,51 @@
...
+    function handleLinkClick(name: linkNames): void {
+        TELEMETRY_V2_RECORDER.recordEvent('home.footer.CTA', 'click', { metadata: { type: v2LinkNameTypes[name] } })
+    }
+
+    const footerLinks = true
+        ? [
+              {
+                  name: 'Docs',
+                  href: 'https://sourcegraph.com/docs',
+                  handleClick: () => handleLinkClick('Docs'),
```
fkling, line 50: I kind of want to simplify this so that we don't have to define a separate function for every link. But since the links are actually not rendered here, I'm not sure what the best way would be.

Does the list have to be defined here though? Or can we move it to `SearchHome`? What was your thought process behind putting them here?

client/web-sveltekit/src/routes/search/+page.svelte
```diff
@@ -26,6 +27,51 @@
...
+        Enterprise: 4,
+        Security: 5,
+        Discord: 6,
+    }
+
+    function handleLinkClick(name: linkNames): void {
+        TELEMETRY_V2_RECORDER.recordEvent('home.footer.CTA', 'click', { metadata: { type: v2LinkNameTypes[name] } })
+    }
+
+    const footerLinks = true
```
fkling, line 45: Should this be conditional based on dotcom? I assume hardcoding `true` was only for testing?

fkling (approved):

client/web-sveltekit/src/routes/search/+page.svelte
camdencheek: 🤦 thanks for the catch. That wasn't supposed to make it into the commit

client/web-sveltekit/src/routes/search/+page.svelte
camdencheek: Yeah, I started with a generic "links" element where we pass the data in. But I thought the typing of the telemetry recorder was more sophisticated than it is. 

I refactored this so this is just a "DotcomLinksFooter" so we can isolate the relevant data. In the case that we ever end up using this for self-hosted instances, we can genericize it.

PR #63172, opened 2024-06-07
Cody Gateway: handle streams with trailing newline in Gemini response

abeatrix:
CONTEXT: https://sourcegraph.slack.com/archives/C05ABRRGB0B/p1717790701356599

Fix an issue where a valid Gemini response ends with new lines, causing a false alert.

image

Issue: We are seeing `*errutil.leafError: no Google response found` in Sentry complaining  when I run a command or chat in VS Code using google as provider.

Cause: Currently the code (added my me) would skip to the last line of the stream response and determine if the response is valid or not, which could be an issue because the stream API could ends the response with an empty new line, where our current logic would fail.

Changes included in this PR:

- Modify `parseGoogleTokenUsage` function to find the last non-empty line in the stream, to handle cases where the stream ends with a newline
- Add a test case to cover the scenario where the stream ends with a newline

This change modifies the behavior of the `parseGoogleTokenUsage` function to handle streams with trailing newlines, which would fail even if the response is valid because Gemini adds a new line to the end of their stream.

## Test plan

Make a curl command to the Gemini streaming API to confirm the response ends with a new line:

image

```sh
curl https://generativelanguage.googleapis.com/v1beta/models/gemini-1.5-flash:streamGenerateContent\?alt=sse\&key\=$GOOGLE_API_KEY \
    -H 'Content-Type: application/json' \
    -X POST \
    -d '{
      "contents": [
        {"role":"user",
         "parts":[{
           "text": "Write the first line of a story about a magic backpack."}]},
        {"role": "model",
         "parts":[{
           "text": "In the bustling city of Meadow brook, lived a young girl named Sophie. She was a bright and curious soul with an imaginative mind."}]},
        {"role": "user",
         "parts":[{
           "text": "Can you set it in a quiet village in 1600s France?"}]},
      ]
    }' 2> /dev/null
```

Copied the response to the test file and confirmed our current test would fail with the same error message:

image

This is now handled by the newly added test with the actually stream response returned by calling the Gemini API.

### DIFF
--- a/cmd/cody-gateway/internal/httpapi/completions/google.go
+++ b/cmd/cody-gateway/internal/httpapi/completions/google.go
@@ -191,13 +191,18 @@ func parseGoogleTokenUsage(r io.Reader, logger log.Logger) (promptTokens int, co
 	scanner.Buffer(make([]byte, 0, 4096), maxPayloadSize)
 	scanner.Split(bufio.ScanLines)
 
-	var lastLine []byte
+	var lastNonEmptyLine []byte
+
+	// Find the last non-empty line in the stream.
 	for scanner.Scan() {
-		lastLine = scanner.Bytes()
+		line := scanner.Bytes()
+		if len(bytes.TrimSpace(line)) > 0 {
+			lastNonEmptyLine = line
+		}
 	}
 
-	if bytes.HasPrefix(bytes.TrimSpace(lastLine), []byte("data: ")) {
-		event := lastLine[5:]
+	if bytes.HasPrefix(bytes.TrimSpace(lastNonEmptyLine), []byte("data: ")) {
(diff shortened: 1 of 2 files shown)


chrsmith (approved):
In the interest of getting you unblocked, if you have reproduced the issue and updated the unit tests to now catch the problem. Then let's just get this checked n.

But this `parseGoogleTokenUsage` function seems a little off.

Can't we just call `io.ReadAll` and just read it into memory, and then manipulate the data as a `string`? Needing to deal with scanners is usually tedious and hard to get right. (e.g. `lastNonEmptyLine = line`.

But I'm sure it would take a non-trivial amount of time to rewrite this to simplify things, so let's just get it fixed.

